PR #4199, opened 2022-01-07
Compile SCALAPACK with OpenBLAS32

amontoison:

PR #4225, opened 2022-01-11
[SCALAPACK] build on experimental platforms

amontoison:


S/SCALAPACK/build_tarballs.jl
```diff
@@ -20,28 +20,22 @@ for f in ${WORKSPACE}/srcdir/patches/*.patch; do
...
-    CMAKE_FLAGS+=(-DCMAKE_EXE_LINKER_FLAGS="-lgfortran")
-  fi
+  CMAKE_FLAGS+=(-DCMAKE_EXE_LINKER_FLAGS="-lgfortran")
 fi
 
 OPENBLAS=(-lopenblas)
 FFLAGS=(-cpp -ffixed-line-length-none)
 
-if [[ ${nbits} == 64 ]] && [[ ${target} != aarch64* ]]; then
+if [[ ${nbits} == 64 ]]; then
```
giordano, line 38: If you do this, it requires Julia v1.7, because we do ILP64 for aarch64 only since OpenBLAS 0.3.17, i.e. Julia v1.7.  I'm sorry about this split, but it is how it is.

giordano:
Mumble mumble mumble, although requiring Julia v1.7 makes this package consistent, this would push requiring v1.7 to dependents as well in the future, which may not be desired.  Do you still have a strong need for doing this after #4199?

We need to find a solution for linking against OpenBLAS for both 1.6 _and_ 1.7, hopefully sooner than later, but at the moment I don't see a good _and_ future-proof solution.

amontoison:
> Mumble mumble mumble, although requiring Julia v1.7 makes this package consistent, this would push requiring v1.7 to dependents as well in the future, which may not be desired. Do you still have a strong need for doing this after #4199?
> 
> We need to find a solution for linking against OpenBLAS for both 1.6 _and_ 1.7, hopefully sooner than later, but at the moment I don't see a good _and_ future-proof solution.

I don't need a new `SCALAPACK_jll` for the moment.
After the fix with `MPICH`, it was just the opportunity to support the new platforms but I forgot this issue with OpenBLAS.

I can close the PR if you want :confused:

amontoison:
SCALAPACK was recompiled with #5167. I can close this PR.

PR #4227, opened 2022-01-11
[SCIP] adapt ipopt version

matbesancon:
There seem to be issues with the newest version of Ipopt_jll


matbesancon:
Following https://github.com/JuliaPackaging/Yggdrasil/issues/4234, pinning the Ipopt version to 300.1400

Issue #3949, opened 2021-11-26
Ensure all our C/C++ cross compiler search for headers in `$includedir` and libraries in `$libdir`


fingolfin:
This is the case for most of them, because `SYSROOT/usr/local` is a symlink to `/workspace/destdir`. But not all of them (see also the discussion on PR #3931).

But in a few exceptions, headers are *not* automatically found if they are in `$includedir`. The exceptions I know about right now are:
- GCC 4.8.5 for `x86_64-linux-musl`
  - this may be caused by `GCCBootstrap@4/bundled/patches/gcc494_musl.patch` which changes `INCLUDE_DEFAULTS` from its default value; it may accidentally have removed `SYSROOT/usr/local`
- clang version 12.0.0 on `x86_64-unknown-freebsd12.2`

However, it works for most other compiler variants, including GCC 4.8.5 for GNU libc, or Apple clang. I did not yet try to check this systematically, so I don't know yet what about non-Apple clang on other platforms.

Motivation: without this, one may get surprising build error that then require adding `-I$includedir` to `CFLAGS` or similar; but this is not obvious and confusing ("but why does it work everywhere else?"). So I think getting this right (=uniform) will make it easier to add and maintainer builder recipes.

I will try to look into the GCC 4.8.5 on musl issue. But so far, I don't even know where in the clang sources the default include search paths are defined; I simply have not yet even tried to find it, perhaps it is easy...  But if anyone has insight into this, I'd appreciate a hint where to look.

fingolfin:
I think the relevant code in clang is this: https://github.com/llvm/llvm-project/blob/444513510999e4c1ea23253654196793834d53bf/clang/lib/Frontend/InitHeaderSearch.cpp#L235-L255 -- note that `AddPath` may add SYSROOT as prefix. So `AddPath("/usr/local/include", System, false);` adds `SYSROOT/usr/local/include`. Except that call is excluded on a bunch of systems, including FreeBSD. But why?

giordano:
> But why?

https://github.com/llvm/llvm-project/commit/315c1675e8ed558aa827f56af260c85e03f07477. Not much more informative...

benlorenz:
> > But why?
> 
> [llvm/llvm-project@315c167](https://github.com/llvm/llvm-project/commit/315c1675e8ed558aa827f56af260c85e03f07477). Not much more informative...

[...]

PR #4232, opened 2022-01-11
Wizard recipe: UDUNITS-v2.2.28

simonbyrne:

* Package name: UDUNITS
* Version: v2.2.28

### DIFF
--- a/U/UDUNITS/build_tarballs.jl
+++ b/U/UDUNITS/build_tarballs.jl
@@ -21,8 +21,8 @@ atomic_patch -p1 ${WORKSPACE}/srcdir/patches/freebsd.patch
 export CPPFLAGS="-I${includedir}" # https://github.com/JuliaPackaging/Yggdrasil/issues/3949
 autoreconf -vi  # https://docs.binarybuilder.org/stable/troubleshooting/#Shared-library-not-built
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
-make all
-make all install
+make -j${nprocs}
+make install
 """
 
 # These are the platforms we will build for by default, unless further
--- a/U/UDUNITS/build_tarballs.jl
+++ b/U/UDUNITS/build_tarballs.jl
@@ -7,13 +7,17 @@ version = v"2.2.28"
 
 # Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://artifacts.unidata.ucar.edu/repository/downloads-udunits/udunits-$(version).tar.gz", "590baec83161a3fd62c00efa66f6113cec8a7c461e3f61a5182167e0cc5d579e")
+    ArchiveSource("https://artifacts.unidata.ucar.edu/repository/downloads-udunits/udunits-$(version).tar.gz", "590baec83161a3fd62c00efa66f6113cec8a7c461e3f61a5182167e0cc5d579e"),
+    DirectorySource("bundled"),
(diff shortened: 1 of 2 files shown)


simonbyrne:
I'm not sure about the naming here (in particular the capitalization):
https://www.unidata.ucar.edu/software/udunits/udunits-2.2.28/udunits2.html

cc: @jakebolewski

giordano:
I think you're seeing #3949 and you have to manually set
```
```
before running configure

U/UDUNITS/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "UDUNITS"
+version = v"2.2.28"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://artifacts.unidata.ucar.edu/repository/downloads-udunits/udunits-$(version).tar.gz", "590baec83161a3fd62c00efa66f6113cec8a7c461e3f61a5182167e0cc5d579e")
```
giordano, line 10: https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1148 works, yes!

giordano:
Capitalization of the name is fine

giordano:
My educated guess for powerpc is https://docs.binarybuilder.org/stable/troubleshooting/#Shared-library-not-built, I don't know what's up with freebsd though

simonbyrne:
@giordano any suggestions on what to do about:
https://github.com/Unidata/UDUNITS-2/blob/c83da987387db1174cd2266b73dd5dd556f4476b/lib/xml.c#L2140
on BSD?

simonbyrne:
Also not sure what the issue with PPC is (it looks like it isn't building a shared library?). @vchuravy any ideas?

giordano:
The `__FreeBSD__` macro should be defined on freebsd

jakebolewski:
> @giordano any suggestions on what to do about: https://github.com/Unidata/UDUNITS-2/blob/c83da987387db1174cd2266b73dd5dd556f4476b/lib/xml.c#L2140 on BSD?

I would try `if not defined WIN32` for that pre-processor branch

simonbyrne:
Okay, now I'm stuck. All the builds are failing with:
```
[01:06:40] WARNING: 'makeinfo' is missing on your system.
[01:06:40]          You should only need it if you modified a '.texi' file, or
[01:06:40]          any other file indirectly affecting the aspect of the manual.
[01:06:40]          You might want to install the Texinfo package:
[01:06:40]          <http://www.gnu.org/software/texinfo/>
[01:06:40]          The spurious makeinfo call might also be the consequence of
[01:06:40]          using a buggy 'make' (AIX, DU, IRIX), in which case you might
[01:06:40]          want to install GNU make:
[01:06:40]          <http://www.gnu.org/software/make/>
```

jakebolewski:
Install the texinfo alpine pkg?  I would imagine there is a flag for disabling the doc building

simonbyrne:
What's weird is that it only appeared after running autoreconf.

simonbyrne:
Okay switching to cmake fixed it.

giordano:
CMake to the rescue, who'd have thought that? :smirk:

PR #4237, opened 2022-01-12
add tempest model build for x86 unix platforms

jakebolewski:
tempestmodel is an atomospheric dycore implementation.  this build will help in making it easier for the CliMA team to run comparative simulations on well defined test cases.

### DIFF
--- a/T/TempestModel/build_tarballs.jl
+++ b/T/TempestModel/build_tarballs.jl
@@ -98,12 +98,13 @@ dependencies = Dependency[
     Dependency("NetCDF_jll"),
     Dependency("HDF5_jll"),
     # The following is adapted from NetCDF_jll
-    Dependency("LibCURL_jll"),
+    BuildDependency(PackageSpec(; name="MbedTLS_jll", version="2.24.0")),
+    #Dependency("LibCURL_jll"),
     # The following libraries are dependencies of LibCURL_jll which is now a
     # stdlib, but the stdlib doesn't explicitly list its dependencies
-    Dependency("LibSSH2_jll"),
-    Dependency("MbedTLS_jll", v"2.24.0"),
-    Dependency("nghttp2_jll"),
+    #Dependency("LibSSH2_jll"),
+    #Dependency("MbedTLS_jll", v"2.24.0"),
+    #Dependency("nghttp2_jll"),
 ]
 
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
--- a/T/TempestModel/build_tarballs.jl
+++ b/T/TempestModel/build_tarballs.jl
@@ -92,7 +92,7 @@ products = [
     ExecutableProduct("RossbyHaurwitzWaveTest", :RossbyHaurwitzWaveTest_exe),
 ]
 
-dependencies = Dependency[
+dependencies = [
     Dependency("MKL_jll"),
     Dependency("MPICH_jll"),
     Dependency("NetCDF_jll"),
--- a/T/TempestModel/build_tarballs.jl
+++ b/T/TempestModel/build_tarballs.jl
@@ -1,0 +1,111 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "TempestModel"
+version = v"0.1"
+sources = [
+    GitSource("https://github.com/paullric/tempestmodel",
+	      "4ecf5146f3100fb24d36156a2f99433f049f0f66"),
+    DirectorySource("./bundled")
+]
+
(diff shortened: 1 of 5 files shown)


T/TempestModel/build_tarballs.jl
```diff
@@ -0,0 +1,111 @@
...
+cp -v test/nonhydro_xz/NonHydroMountainCartesianTest ${bindir}
+cp -v test/nonhydro_xz/HydrostaticMountainCartesianTest ${bindir}
+cp -v test/nonhydro_xz/ThermalBubbleCartesianTest ${bindir}
+cp -v test/nonhydro_xz/Baroclinic3DCartesianRidgeTest ${bindir}
+cp -v test/nonhydro_xz/ScharMountainCartesianTest ${bindir}
+cp -v test/nonhydro_xz/InertialGravityCartesianXZTest ${bindir}
+
+cp -v test/shallowwater_sphere/BarotropicInstabilityTest ${bindir}
+cp -v test/shallowwater_sphere/MountainRossbyTest ${bindir}
+cp -v test/shallowwater_sphere/RossbyHaurwitzWaveTest ${bindir}
```
giordano, line 54: Fully automated installation, I see

T/TempestModel/build_tarballs.jl
```diff
@@ -0,0 +1,111 @@
...
+    Dependency("MPICH_jll"),
+    Dependency("NetCDF_jll"),
+    Dependency("HDF5_jll"),
+    # The following is adapted from NetCDF_jll
+    Dependency("LibCURL_jll"),
+    # The following libraries are dependencies of LibCURL_jll which is now a
+    # stdlib, but the stdlib doesn't explicitly list its dependencies
+    Dependency("LibSSH2_jll"),
+    Dependency("MbedTLS_jll", v"2.24.0"),
+    Dependency("nghttp2_jll"),
```
giordano, line 106: I'm still not convinced this is necessary.  We used to have a problem with stdlibs not being listed as dependencies, but that should have been fixed looong ago.  The only problem I'm aware of is that we get the wrong version of MbedTLS (no clue why, compat bounds in the registry look good, as far as I can tell) and that can be forced with
```suggestion
    # This is only an indirect dependency
    BuildDependency(PackageSpec(; name="MbedTLS_jll", version="2.24.0")),
```

jakebolewski: Ok i can try

T/TempestModel/build_tarballs.jl
jakebolewski: I did automate the string formation to paste in here with julia lol

T/TempestModel/build_tarballs.jl
```diff
@@ -0,0 +1,112 @@
...
+    ExecutableProduct("ScharMountainCartesianTest", :ScharMountainCartesianTest_exe),
+    ExecutableProduct("InertialGravityCartesianXZTest", :InertialGravityCartesianXZTest_exe),
+
+    # shallow water test cases
+    ExecutableProduct("BarotropicInstabilityTest", :BarotropicInstabilityTest_exe), 
+    ExecutableProduct("MountainRossbyTest", :MountainRossbyTest_exe),
+    ExecutableProduct("RossbyHaurwitzWaveTest", :RossbyHaurwitzWaveTest_exe),
+]
+
+dependencies = Dependency[
```
giordano, line 95:
```suggestion
dependencies = [
```

T/TempestModel/build_tarballs.jl
giordano: Yeah, sorry I was joking :slightly_smiling_face:

T/TempestModel/build_tarballs.jl
giordano: Looks like that worked :slightly_smiling_face:   I'm still puzzled by the fact we get the wrong version of MbedTLS, as far as I can tell LibCURL_jll specifies correctly the compatibility bound: https://github.com/JuliaRegistries/General/blob/7392039acb21210b43cef86c7fa5f6f42213952e/L/LibCURL_jll/Compat.toml#L7-L9

jakebolewski: Cool!  I'll update the other package where I used the same form as well.

giordano: I believe I understood what's the problem!  Without doing anything special, Pkg installs `LibCURL_jll` 7.71, which doesn't have a compat with `MbedTLS_jll` and then installs the latest version of `MbedTLS_jll`, 2.28.  If I explicitly ask for LibCURL_jll 7.73 I get the right MbedTLS.  I think we should fix the compat with `MbedTLS_jll` also for previous versions of `LibCURL_jll`

PR #4240, opened 2022-01-12
DuckDB:Support for Parquet extension

kimmolinna:
DuckDB's source package includes extensions. And the default setting in `Makefile` is nowadays  
`-DBUILD_PARQUET_EXTENSION=TRUE`.


kimmolinna:
DuckDB's present version number is 0.3.1 so I wouldn't like the use version number 0.3.2. Is still possible?

kimmolinna:
I will use version number 0.3.2. That's easiest....

PR #4242, opened 2022-01-12
DuckDB: Parquet extension support

kimmolinna:


D/DuckDB/bundled/patches/0001-right-case-for-winsock2.h.patch
```diff
@@ -0,0 +1,26 @@
+From eab00b8a2cbb8c1917c05175de135d8411f6444f Mon Sep 17 00:00:00 2001
+From: Kimmo Linna <kimmo.linna@gmail.com>
```
giordano, line 2: We're lucky someone already submitted this patch upstream :slightly_smiling_face:

kimmolinna:
Package 0.3.1. fails to with Windows/MinGW because of these two lines:
`vector.hpp` line 72 `DUCKDB_API 	Vector(const Vector &) = delete;` and
`catalog_search_path.hpp` line 25 `DUCKDB_API CatalogSearchPath(const CatalogSearchPath &other) = delete;`
Without `DUCKDB_API` compiling should wokr. At least it works Zig with MinGW from Linux to Windows.

D/DuckDB/bundled/patches/0001-right-case-for-winsock2.h.patch
kimmolinna: Someone should add those two other patches two. ;)

PR #4244, opened 2022-01-12
[DuckDB] Fix PRAGMA version issue

jeremiahpslewis:


D/DuckDB/build_tarballs.jl
```diff
@@ -7,7 +7,7 @@ version = v"0.3.1"
 
 # Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://github.com/duckdb/duckdb/archive/refs/tags/v$(version).tar.gz", "ae2367d0a393be59e137ffa975f48f60b113b2a72aacc24fbc59afa0cbc3a511"),
+    GitSource("https://github.com/duckdb/duckdb.git", "88aa81c6b1b851c538145e6431ea766a6e0ef435"),
```
giordano, line 10: [This commit](https://github.com/duckdb/duckdb/commit/88aa81c6b1b851c538145e6431ea766a6e0ef435) is in, what does this PR change?

jeremiahpslewis: This uses Git as the source instead of Archive, which (for this reason (?) https://github.com/kimmolinna/DuckDB.jl/issues/6#issuecomment-1011430246) resolves the issue of getting the wrong version number when querying the DB for its version.

jeremiahpslewis: I tested it locally and it seemed to resolve the issue.

giordano: I'm more confused than before.  The source code is the same, the archive is simply a snapshot of the repo at that commit.

giordano: Oh, they use git to get the version number otherwise they set a rubbish version number.  That looks brilliant.

PR #4250, opened 2022-01-13
[rocRAND] Update to 4.2.0

jpsamaroo:


R/rocRAND/build_tarballs.jl
```diff
@@ -51,11 +55,11 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    BuildDependency(PackageSpec(; name="LLVM_full_jll", version=v"11.0.1")),
+    BuildDependency(PackageSpec(; name="ROCmLLVM_jll", version=v"4.2.0")),
     Dependency("hsa_rocr_jll"),
     Dependency("ROCmCompilerSupport_jll"),
     Dependency("ROCmOpenCLRuntime_jll"),
-    Dependency("HIP_jll"),
+    Dependency("HIP_jll", v"4.2.0"),
```
giordano, line 62: If the version is important also for runtime you may want to specify the compat bound as well:
```suggestion
    Dependency("HIP_jll"; compat="4.2.0"),
```

R/rocRAND/build_tarballs.jl
```diff
@@ -18,13 +18,17 @@ cd ${WORKSPACE}/srcdir/rocRAND*/
...
-export HIP_RUNTIME=ROCclr
+export HIP_RUNTIME=rocclr
 export HIP_COMPILER=clang
-export HIP_PLATFORM=rocclr
-export HIPCC_VERBOSE=1
+export HIP_PLATFORM=amd
+export HIP_ROCCLR_HOME=${prefix}/lib
+export HIPCC_VERBOSE=7
+export HIP_LIB_PATH=${prefix}/hip/lib
+export LD_LIBRARY_PATH=${prefix}/hip/lib:$LD_LIBRARY_PATH
```
giordano, line 31: Interesting 😬 why is that needed? I presume it works only because we're targeting platforms for which we happen to be able to do native build?

jpsamaroo: Sorry, artifact from trying to get this to build. Removed in latest push.

jpsamaroo:
So one issue I still have when building on musl is that the generated libraries use `initial-exec` TLS, which isn't supported on musl and probably never will be (and is not something I'm interested in patching in ldso). Is there an obvious strategy to stop clang from emitting such TLS relocations?

jpsamaroo:
For example, I get this during audit:

> Error relocating /opt/rocm/hip/lib/libamdhip64.so.4: _ZTVN10__cxxabiv121__vmi_class_type_infoE: initial-exec TLS resolves to dynamic definition in /opt/rocm/hip/lib/libamdhip
64.so.4

Which causes the build to fail, and will inevitably make the library unusable on musl systems.

jpsamaroo:
Passing `-femulated-tls` through to clang doesn't help.

giordano:
I'm marking this as draft, as I understand you aren't happy yet with the result.

jpsamaroo:
If I were better at reading, I would have realized that the issue is with `libamdhip64`, not `librocrand`...

jpsamaroo:
Aaannnd this was a problem with my local configuration (I had a system-wide install of HIP which wasn't built with Yggdrasil). I fixed that and deployed rocRAND_jll locally, and it works! Ready to go!

R/rocRAND/build_tarballs.jl
```diff
@@ -51,11 +55,11 @@ products = [
...
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    BuildDependency(PackageSpec(; name="LLVM_full_jll", version=v"11.0.1")),
+    BuildDependency(PackageSpec(; name="ROCmLLVM_jll", version=v"4.2.0")),
     Dependency("hsa_rocr_jll"),
     Dependency("ROCmCompilerSupport_jll"),
     Dependency("ROCmOpenCLRuntime_jll"),
-    Dependency("HIP_jll"),
+    Dependency("HIP_jll"; compat="4.2.0"),
     Dependency("rocm_cmake_jll"),
```
giordano, line 63: Same question as https://github.com/JuliaPackaging/Yggdrasil/pull/4256#discussion_r785323067: this should be
```suggestion
    BuildDependency("rocm_cmake_jll"),
```
?

PR #4258, opened 2022-01-15
add spectra

Crghilardi:
This PR adds a `build_tarballs.jl` for the [spectra](https://github.com/yixuan/spectra) library. This is another header only library. I believe I remembered to use the correct dependency type this time around.

### DIFF
--- a/S/spectra/build_tarballs.jl
+++ b/S/spectra/build_tarballs.jl
@@ -1,0 +1,41 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "spectra"
+version = v"1.0.0"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/yixuan/spectra/archive/refs/tags/v$(version).tar.gz", "45228b7d77b916b5384245eb13aa24bc994f3b0375013a8ba6b85adfd2dafd67")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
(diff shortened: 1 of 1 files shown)


S/spectra/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+#header only, so no products
+products = Product[]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency(PackageSpec(name="Eigen_jll", uuid="bc6bbf8a-a594-5541-9c57-10b0d0312c70"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
giordano, line 41: Side note, strictly speaking, for packages built for `AnyPlatform` it isn't necessary to require Julia v1.6, but it doesn't hurt to do so either, we want to encourage users to use newer versions of Julia anyways.

S/spectra/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+# platforms are passed in on the command line
+platforms = [AnyPlatform()]
+
+# The products that we will ensure are always built
+#header only, so no products
+products = Product[]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency(PackageSpec(name="Eigen_jll", uuid="bc6bbf8a-a594-5541-9c57-10b0d0312c70"))
```
giordano, line 37: Yes, also Eigen is header-only, so that's definitely also a build-only dependency

PR #4262, opened 2022-01-15
add Taskflow

Crghilardi:
This PR adds a `build_tarballs.jl` for the [Taskflow](https://github.com/taskflow/taskflow) library. Once again, header only library. The only thing maybe worth mentioning was they specify a minimum `gcc` version in the `CMakeLists.txt`, so I had to increase the `preferred_gcc_version` to make everything happy.


giordano:
Opening PRs right now doesn't help, since registration is broken at the moment :wink:

giordano:
Ok, well, maybe it works after all.  Let's see how this goes!

PR #4264, opened 2022-01-16
[jlqml] Bump version to v0.3

barche:
Needed to make QML.jl work on Julia 1.7

### DIFF
--- a/J/jlqml/build_tarballs.jl
+++ b/J/jlqml/build_tarballs.jl
@@ -2,14 +2,19 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
-julia_version = v"1.6.0"
+# See https://github.com/JuliaLang/Pkg.jl/issues/2942
+# Once this Pkg issue is resolved, this must be removed
+uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
+delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
+
+julia_versions = [v"1.6.3", v"1.7.0", v"1.8.0"]
 
 name = "jlqml"
-version = v"0.2"
+version = v"0.3"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/barche/jlqml.git", "4bc5d39bdc261eff12e3b7d1cd8b28ff0519e316"),
+    GitSource("https://github.com/barche/jlqml.git", "4d21e2fcf0bcae29c45e76c8ac9d1d45893a77e6"),
 ]
 
 # Bash recipe for building across all platforms
@@ -44,10 +49,13 @@ install_license $WORKSPACE/srcdir/jlqml*/LICENSE.md
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 include("../../L/libjulia/common.jl")
-platforms = libjulia_platforms(julia_version)
+platforms = vcat(libjulia_platforms.(julia_versions)...)
 platforms = expand_cxxstring_abis(platforms)
 
 filter!(p -> libc(p) != "musl", platforms) # Qt_jll is currently not available for muslc
+# Qt5Declarative_jll is not available for these architectures:
+filter!(p -> arch(p) != "armv6l", platforms)
+filter!(p -> !(arch(p) == "aarch64" && Sys.isapple(p)), platforms)
 
 # The products that we will ensure are always built
 products = [
@@ -60,10 +68,10 @@ dependencies = [
     Dependency("Qt5Declarative_jll"),
     Dependency("Qt5Svg_jll"),
     BuildDependency("Libglvnd_jll"),
-    BuildDependency(PackageSpec(name="libjulia_jll", version=julia_version)),
+    BuildDependency("libjulia_jll"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-    preferred_gcc_version=v"9",
-    julia_compat = "$(julia_version.major).$(julia_version.minor)")
+    preferred_gcc_version = v"9",
+    julia_compat = "1.6")


barche:
I'm getting `ERROR: LoadError: Unsatisfiable requirements detected for package libjulia_jll` and I think this is due to a change in BinaryBuilder or Yggdrasil, because I locally get the same trying to rebuild libcxxwrap-julia using an unmodified build_tarballs. No idea how to fix this.

Log:

https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16212&view=logs&j=33b33528-48ed-59ae-ddc2-6d83dd8c01fb&t=0f64fd58-401e-5acc-6696-90343ad2ad30&l=27

giordano:
> because I locally get the same trying to rebuild libcxxwrap-julia using an unmodified build_tarballs

What version of Julia and BinaryBuilderBase.jl are you using locally? Yesterday we upgraded to Julia 1.7 here: #4261. I'm clueless about Pkg internals, which break all the time

barche:
> > because I locally get the same trying to rebuild libcxxwrap-julia using an unmodified build_tarballs
> 
> What version of Julia and BinaryBuilderBase.jl are you using locally? Yesterday we upgraded to Julia 1.7 here: #4261. I'm clueless about Pkg internals, which break all the time

Julia 1.7.1 also, it didn't work with Julia 1.6 (some error about Pkg.Registry not having a certain function). I'll try with Julia 1.6 and an older BinaryBuilder.

giordano:
This recipe works with `BinaryBuilder.jl` 0.4.8/`BinaryBuilderBase` 1.3.0 and Julia v1.6.  So someone needs to dig out what broke in Pkg again...

giordano:
https://github.com/JuliaLang/Pkg.jl/issues/2942

barche:
I think hacking the `stdlibs` error out uncovered another, unrelated problem. No idea how to interpret this error:

```
##[warning]Dependency Qt5Declarative_jll does not have an (Stdlib)Artifacts.toml in /depot/packages/Qt5Declarative_jll/Vktrd!
```

Log: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16459&view=logs&jobId=023fc726-910f-52b9-2466-adec1eae67d0&j=023fc726-910f-52b9-2466-adec1eae67d0&t=7bf96d55-6405-5d0a-4d10-93a2cdf2f87c

On FreeBSD it's different:

```
IOError: rm("/depot/packages/Qt5Declarative_jll/Vktrd/src/wrappers"): directory not empty (ENOTEMPTY)
```

Log: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16459&view=logs&jobId=023fc726-910f-52b9-2466-adec1eae67d0&j=4ad5501b-6adf-545d-1cbf-e5f6d4fedafb&t=325a2760-47a9-5bf0-821f-6cf85ac7400a

giordano:
That was likely a race condition

J/jlqml/build_tarballs.jl
```diff
@@ -2,14 +2,17 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
-julia_version = v"1.6.0"
+uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
+delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
```
giordano, line 6: Doesn't this actually break everything when we try to target Julia v1.6?  I don't see `LibOSXUnwind_jll` in https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16459&view=logs&jobId=3e3b6bc7-f97f-5652-f6ca-c52d13eda498&j=a151bd49-d959-5dd9-e79c-4da26ad24f9f&t=fa03432c-08c9-5f99-931b-dc1b3c71632b, which looks wrong.

barche: As far as I understand, `LibOSXUnwind_jll` is a dependency of `libjulia_jll` and removing it from the `stdlibs` makes Pkg choose the one in the registry. It doesn't appear to cause harm, at least on Linux I get a usable JLL with binaries for Julia 1.6, 1.7 and 1.8: https://github.com/barche/jlqml_jll.jl

barche:
No idea why CI fails here, this version works locally for all three platforms.

barche:
I can't find any way to get rid of this error:

```
Dependency Qt5Declarative_jll does not have an (Stdlib)Artifacts.toml
```

Tried on two different machines, locally everything works with the exact same version of BinaryBuilder.

giordano:
> That was likely a race condition

:point_up:

barche:
> > That was likely a race condition
> 
> point_up

Yes, I saw that, but I reran it several times since then, not sure what changed now?

giordano:
I deleted `/depot/packages/Qt5Declarative_jll/Vktrd` :slightly_smiling_face:

J/jlqml/build_tarballs.jl
giordano: I still don't see `LibOSXUnwind_jll` in https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16556&view=logs&j=11cc9737-c3b2-5b2e-49c1-3ecd4dd3473d&t=6864626c-b206-5daa-5d51-831805962223.  @fingolfin maybe it wasn't necessary after all?  If so, I think the best thing to do would be to have a `libjulia_jll` for v1.6 which doesn't depend at runtime on `LibOSXUnwind_jll`, fix this in the registry so `libjulia_jll` never tries to pull in `LibOSXUnwind_jll`.  That'd solve all of our problems without hacks like this.

Should we keep this hack (I hope we don't), at very least please add a comment, I'm sure someone will be very confused by reading these lines without context in a few months.

barche:
I squashed the commits and activated all platforms for which the dependencies are available, so for me this is good to merge.

J/jlqml/build_tarballs.jl
barche: That seems like the way to go, but how is this dependency triggered, exactly? I don't see a `LibOSXUnwind_jll` in the `Project.toml` of `libjulia_jll v1.7.0+8`? Is it about removing these:

https://github.com/JuliaBinaryWrappers/libjulia_jll.jl/blob/2e64b99b0e204bf25308727beb971269affd3c4b/src/wrappers/x86_64-apple-darwin-julia_version%2B1.6.3.jl#L6

giordano: It's only in v1.6

barche: Yes, but what I mean is that `libjulia_jll v1.7.0+8` is the one that [is getting installed](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16556&view=logs&j=11cc9737-c3b2-5b2e-49c1-3ecd4dd3473d&t=6864626c-b206-5daa-5d51-831805962223&l=169) (even when targeting Julia 1.6) and its `Project.toml` has no reference to `LibOSXUnwind_jll`: https://github.com/JuliaBinaryWrappers/libjulia_jll.jl/blob/main/Project.toml

fingolfin: I am pretty sure `julia.h` from the libjulia_jll 1.6 won't be useable without `LibOSXUnwind_jll` because of the missing `libunwind.h` (either a compile error, or worse, it picks up a different "wrong" copy of `libunwind.h` somewhere). That said, we

fingolfin: That said, we could perhaps indeed work around it, by either patching out the need for `libunwind.h` or by shipping a copy with it in the libjulia_jll -- after all, we only want this JLL as a build dependency

Looking at it, I think there is a realistic chance of doing even the patching part. Indeed, this patch might do it (applied in libjulai_jll *after* building it, i.e., either just before installing or even to the already install header files:
```patch
diff --git a/src/julia_internal.h b/src/julia_internal.h
index 0b76696791..092a7ca8dd 100644
--- a/src/julia_internal.h
+++ b/src/julia_internal.h
@@ -893,6 +893,7 @@ typedef struct {
 
 // Might be called from unmanaged thread
 uint64_t jl_getUnwindInfo(uint64_t dwBase);
+#if 0 // HACK to avoid libunwind dependency for libjulia_jll
 #ifdef _OS_WINDOWS_
 #include <dbghelp.h>
 JL_DLLEXPORT EXCEPTION_DISPOSITION __julia_personality(
@@ -933,8 +934,9 @@ size_t rec_backtrace_ctx(jl_bt_element_t *bt_data, size_t maxsize, bt_context_t
 #ifdef LIBOSXUNWIND
 size_t rec_backtrace_ctx_dwarf(jl_bt_element_t *bt_data, size_t maxsize, bt_context_t *ctx, jl_gcframe_t *pgcstack) JL_NOTSAFEPOINT;
 #endif
+#endif  // HACK to avoid libunwind dependency for libjulia_jll
 JL_DLLEXPORT jl_value_t *jl_get_backtrace(void);
-void jl_critical_error(int sig, bt_context_t *context);
+//void jl_critical_error(int sig, bt_context_t *context); // HACK to avoid libunwind dependency for libjulia_jll
 JL_DLLEXPORT void jl_raise_debugger(void);
 int jl_getFunctionInfo(jl_frame_t **frames, uintptr_t pointer, int skipC, int noInline) JL_NOTSAFEPOINT;
 JL_DLLEXPORT void jl_gdblookup(void* ip) JL_NOTSAFEPOINT;
diff --git a/src/julia_threads.h b/src/julia_threads.h
index f42984130f..6a30aa9008 100644
--- a/src/julia_threads.h
+++ b/src/julia_threads.h
@@ -67,7 +67,7 @@ typedef struct {
 #endif
 #if defined(JL_HAVE_UCONTEXT) || defined(JL_HAVE_UNW_CONTEXT)
 #define UNW_LOCAL_ONLY
-#include <libunwind.h>
+//#include <libunwind.h> // HACK to avoid libunwind dependency for libjulia_jll
 typedef struct {
     ucontext_t ctx;
 #if defined(JL_TSAN_ENABLED)
```

The idea here being that the disabled code really is internal and not relevant to any external consumer of the Julia kernel ABI. (It would be good if the Julia headers were restructured with this principle in mind anyway... *sigh*)

barche: What I don't understand is why it is looking for `LibOSXUnwind_jll`, when it is not even listed as a dependency in the `libjulia_jll` Project.toml. And from what I understand, this `delete!` call only removes it from the list of available standard libraries, not from the dependencies?

barche: @fingolfin I did some extra tests, and neither `libcxxwrap-julia` nor `jlqml` (compiled using the old binarybuilder) link to `libOSXUnwind.dylib`. I also tested the `libjlqml.dylib` built using the `delete!` hack, and it passes the QML tests. So I think this patch is not really needed, there is already no real dependency on `LibOSXUnwind_jll` and it is not used anyway (by libcxxwrap-julia and its dependents at least).

So I think it is safe to leave this hack in for now, to at least move forward with Julia 1.7-compatible binaries. I'll also update https://github.com/JuliaLang/Pkg.jl/issues/2942 with this info, since I have no idea what exactly is going on in Pkg here or why `LibOSXUnwind_jll` is even pulled into the dependency list.

barche: OK, so I guess the possibility of corrupt binaries is excluded, at worst we should get a linking error. I have added a comment, I'd propose to merge this and then investigate https://github.com/JuliaLang/Pkg.jl/issues/2942 further, libjulia_jll also doesn't build because of that right now.

fingolfin: I agree, let's test this, if it works, this is a good short term solution that we can use for the time being also elsewhere.

barche: OK, from https://github.com/JuliaLang/Pkg.jl/pull/2963 I also conclude that this hack will in this case actually have the same effect as the (possible) fix. `LibOSXUnwind_jll` is here only encountered in the historical dependencies and then triggers a bug in the dependency resolution. So, @giordano this is now good to merge I think :)

giordano: Ok, let's do it

PR #4090, opened 2021-12-18
[Arb] update to 2.21.1

kalmarek:

PR #4270, opened 2022-01-17
update to build libarb v2.22.0

JeffreySarnoff:
@fingolfin  @thofma 
 please review for correctness and (with @kalmarek) to verify that it creates no conflicts with your works.
I updated the build for v2.22.0.
In addition, following the comment  in https://github.com/JuliaPackaging/Yggdrasil/pull/4090, I removed all version tests to build for Julia < v"1.6".

### DIFF
--- a/A/Arb/build_tarballs.jl
+++ b/A/Arb/build_tarballs.jl
@@ -52,7 +52,7 @@ make install LIBDIR=$(basename ${libdir})
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(; experimental=true)
+platforms = supported_platforms()
 
 # The products that we will ensure are always built
 products = [


A/Arb/build_tarballs.jl
```diff
@@ -48,29 +47,29 @@ elif [[ ${target} == *mingw* ]]; then
...
-    Dependency("MPFR_jll", build_for_julia16_or_newer ? v"4.1.1" : v"4.0.2"),
+    Dependency("GMP_jll", v"6.2.0"),
+    Dependency("MPFR_jll", v"4.1.1"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat = build_for_julia16_or_newer ? "1.6" : "1.0-1.5")
+               julia_compat = 1.6")
+
```
giordano, line 75:
```suggestion
               julia_compat = "1.6")
```

A/Arb/build_tarballs.jl
```diff
@@ -48,29 +47,29 @@ elif [[ ${target} == *mingw* ]]; then
...
-
 ./configure --prefix=$prefix --disable-static --enable-shared --with-gmp=$prefix --with-mpfr=$prefix --with-flint=$prefix ${extraflags}
 make -j${nproc}
 make install LIBDIR=$(basename ${libdir})
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(; experimental=build_for_julia16_or_newer)
+platforms = supported_platforms(; experimental=true)
```
giordano, line 57:
```suggestion
platforms = supported_platforms()
```

A/Arb/build_tarballs.jl
```diff
@@ -22,24 +22,23 @@ import Pkg.Types: VersionSpec
...
                         upstream_version.minor * 100 + version_offset.minor,
                         upstream_version.patch * 100 + version_offset.patch)
 
+# Collection of sources required to complete build
 # Collection of sources required to complete build
 sources = [
     ArchiveSource("https://github.com/fredrik-johansson/arb/archive/refs/tags/$(upstream_version).tar.gz",
-                  "aecc9f55eb35a00a2490e9a7536a0babf8ac86bb13d32a4a95e1216f9f5cbfa8"),
+                  "c9903d49a3e89bbd0ae114eeed97eaea3877e0e1a31a5cdb24ba0c84e4e81f4a")
 ]
```
kalmarek, line 35: maybe switching to git source is more convenient?
```julia
sources = [
    GitSource("https://github.com/fredrik-johansson/arb.git", "c2b7e36915a451455c1f115ce343dd85bdc30c59")
]
```
where we directly reference the tagging commit.

But this is just an option to consider.

A/Arb/build_tarballs.jl
```diff
@@ -22,24 +22,21 @@ import Pkg.Types: VersionSpec
 # to all components.
 
 name = "Arb"
-upstream_version = v"2.21.1"
-build_for_julia16_or_newer = true
-version_offset = build_for_julia16_or_newer ? v"0.0.1" : v"0.0.0"
+upstream_version = v"2.22.0"
+version_offset = v"0.0.1"
```
thofma, line 26:
```suggestion
version_offset = v"0.0.0"
```
I think we can reset this to `0.0.0`, right @fingolfin?

A/Arb/build_tarballs.jl
thofma: Can you add this change here @JeffreySarnoff?

A/Arb/build_tarballs.jl
```diff
@@ -48,29 +45,28 @@ elif [[ ${target} == *mingw* ]]; then
...
 # The products that we will ensure are always built
 products = [
     LibraryProduct("libarb", :libarb)
 ]
 
 # Dependencies that must be installed before this package can be built
-v = build_for_julia16_or_newer ? v"200.800.101" : v"200.800.100"
+v = v"200.800.101"
 dependencies = [
     Dependency("FLINT_jll", v; compat = "~$v"),
```
fingolfin, line 65: I think we can simplify this (and update to the latest version, just in case)
```suggestion
dependencies = [
    Dependency("FLINT_jll", compat = "~200.800.401"),
```

thofma:
I think this looks good, but could you also apply the suggestion of @giordano above? Then this should be good to.

thofma (approved):

JeffreySarnoff:
> I think this looks good, but could you also apply the suggestion of @giordano above?
done (@giordano -- I had had this change in, dropped it inadvertantly)

kalmarek:
@JeffreySarnoff thanks!

PR #4272, opened 2022-01-18
[Antlr4CppRuntime] Update gcc version.

azreika:
g++-6 on linux was causing issues downstream when linking with ANTLR driver code compiled with later versions. Verified locally that the issue goes away with g++-7.


giordano:
@azreika out of curiosity, what was the error exactly?

azreika:
I was getting errors with undefined symbols on Linux machines when linking in driver code compiled with a later GCC version.

Tested linking this version with driver code samples compiled using different GCC versions and it's working great now 
🥳  Thanks for the speedy reviews! 😁

PR #4277, opened 2022-01-18
[GoogleTest] v1.11.0

jeremiahpslewis:


G/GoogleTest/build_tarballs.jl
```diff
@@ -28,6 +34,9 @@ make install
 # platforms are passed in on the command line
 platforms = expand_cxxstring_abis(supported_platforms())
 
+# Only CXX 11 is supported
+filter!(x -> cxxstring_abi(x) != "cxx03", platforms)
```
giordano, line 38: What do you mean?

jeremiahpslewis: https://github.com/google/googletest#supported-platforms
> GoogleTest requires a codebase and compiler compliant with the C++11 standard or newer.

And the CXX03 builds all fail

G/GoogleTest/build_tarballs.jl
```diff
@@ -3,22 +3,28 @@
...
+ sources = [
+     GitSource("https://github.com/google/googletest.git", "e2239ee6043f73722e7aa812a459f54a28552929")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
+
+if [[ "${target}" == *apple* ]]; then
+    CMAKE_FLAGS+=(-DCMAKE_CXX_FLAGS=-mmacosx-version-min=11.0)
+fi
```
giordano, line 18: This isn't necessary for aarch64-apple-darwin: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/193.  And it looks rather optimistic for x86_64

G/GoogleTest/build_tarballs.jl
giordano: Did you read https://docs.binarybuilder.org/stable/building/#Expanding-C-string-ABIs-or-libgfortran-versions and in particular (emphasis not mine)

> This ABI does _not_ have to do with the C++ standard used by the source code

?

jeremiahpslewis: Gotcha, thanks.

G/GoogleTest/build_tarballs.jl
```diff
@@ -3,22 +3,30 @@
...
 
 name = "GoogleTest"
-version = v"1.10.0"
+version = v"1.11.0"
 
-# Collection of sources required to complete build
-sources = [
-    GitSource("https://github.com/google/googletest.git", "703bd9caab50b139428cea1aaff9974ebee5742e")
+ # Collection of sources required to complete build
+ sources = [
```
giordano, line 9: Whitespace change?
```suggestion
# Collection of sources required to complete build
sources = [
```

PR #4279, opened 2022-01-18
[LLVM_jll] Fix lld export

jpsamaroo:
Fixes broken symlinks, and exports `ld.lld` et. al (which are how `lld` is supposed to be invoked).

@giordano


L/LLVM/common.jl
```diff
@@ -353,6 +353,9 @@ rm -vrf ${prefix}/lib/*LLVM*.a
 rm -vrf ${prefix}/lib/libclang*.a
 rm -vrf ${prefix}/lib/clang
 rm -vrf ${prefix}/lib/mlir
+
+# Symlink `lld` from tools/ to bin/
+ln -s ${prefix}/bin/lld ${prefix}/tools/lld
```
vchuravy, line 358: No we can't do that. The point that these binaries should not be in `bin`.

jpsamaroo: Ok dropped that commit. Now I'm moving them back to `bin/`.

jpsamaroo: Damnit, misread. Will move lld to `tools/`

L/LLVM/common.jl
```diff
@@ -395,7 +398,12 @@ function configure_build(ARGS, version; experimental_platforms=false, assert=fal
     end
     if version >= v"12"
         push!(products, LibraryProduct("libclang-cpp", :libclang_cpp, dont_dlopen=true))
-        push!(products, ExecutableProduct("lld", :lld, "bin"))
+        push!(products, ExecutableProduct("lld", :lld, "tools"))
```
jpsamaroo, line 401:
```suggestion
        push!(products, ExecutableProduct("lld", :lld, "bin"))
```

@vchuravy does this ^^ seem correct? Or do we want to move `lld` to `tools/` even earlier?

vchuravy: https://github.com/llvm/llvm-project/commit/352ac408d4ff7705529506bf7dbca4de6b32cef4 but for LLD?

jpsamaroo: Ok latest commit does this for LLVM 12.0.1. Want this for all 12+ versions?

vchuravy: Jup, but we can do that in stages. We probably want to add them to `JuliaLang/llvm-project`

vchuravy:
Do you need to rebase?

jpsamaroo:
This is working for me locally; in the absence of working CI, what do we want to do?

L/LLVM/common.jl
```diff
@@ -353,6 +352,9 @@ rm -vrf ${prefix}/lib/*LLVM*.a
 rm -vrf ${prefix}/lib/libclang*.a
 rm -vrf ${prefix}/lib/clang
 rm -vrf ${prefix}/lib/mlir
+
+# Move lld to tools/
+mv -v ${prefix}/bin/lld ${prefix}/tools/lld
```
giordano, line 357: Probably
```suggestion
mv -v "${bindir}/lld${exeext}" "${prefix}/tools/lld${exeext}"
```
?

giordano:
Good to go?  Without triggering a new build?

jpsamaroo:
Yes please!

jpsamaroo:
Oh wait, we need a new extraction for LLVM_jll, but not a new build of the components being extracted. I just need the symlinks fixed in the artifacts.

jpsamaroo:
I'll also submit a fix for LLVM 13 once I've gotten this patch submitted upstream (if necessary).

vchuravy (approved):

giordano:
> Oh wait, we need a new extraction for LLVM_jll, but not a new build of the components being extracted. I just need the symlinks fixed in the artifacts.

@jpsamaroo sorry, I'm a bit lost about whether you want to do this first or what :slightly_smiling_face:

jpsamaroo:
@giordano I'd like to merge this and then have the `LLVM_jll@12.0.1` wrapper rebuilt; I remember that in the past, you had to manually trigger wrapper rebuilds, so not sure if that's still the case.

giordano:
Ok, this is a question for @vchuravy.  Do you want to go through the whole rebuilding dance now? :grimacing:  Maybe there is something else you want to add?

jpsamaroo:
Oh, right, I forgot that this patches LLVM.

vchuravy:
Can you put the patch on JuliaLang/llvm-project targetting LLVM12 and LLVM 13?

jpsamaroo:
https://github.com/JuliaLang/llvm-project/pull/5
https://github.com/JuliaLang/llvm-project/pull/6

giordano:
Ok, this should now be ready to go (with `[skip ci]`)?

vchuravy (approved):

Issue #3024, opened 2021-05-18
Packages with known incompatibilities across different versions


giordano:
* [`Bzip2_jll`](https://juliahub.com/ui/Packages/Bzip2_jll/biC9V/): the v1.0.6 for macOS forces the soname to include the full version, this was fixed for v1.0.8 (#2981)
  ```julia
  Dependency("Bzip2_jll", v"1.0.6"; compat="=1.0.6") # this is bugged, avoid it
  Dependency("Bzip2_jll", v"1.0.8"; compat="1.0.8") # only Julia v1.6+, all platforms
  ```
* [`Expat_jll`](https://juliahub.com/ui/Packages/Expat_jll/4P57b/): the soname of the library for macOS seems to change at all patch versions ([example](https://github.com/JuliaBinaryWrappers/Expat_jll.jl/commit/f754bdb5917f9628361b527a78fb8af5bc83ab36#diff-2a84e35074ab9c852eb3e5c4b62019ca420842a627a0d8d55c03cd26dd65151f))
  ```julia
  Dependency("Expat_jll", v"2.2.7"; compat="=2.2.7"),
  Dependency("Expat_jll", v"2.2.10"; compat="=2.2.10"), # only Julia v1.6+, all platforms
  ```
* [`GMP_jll`](https://juliahub.com/ui/Packages/GMP_jll/1Lisu/): packages built against v6.2.1 can't be loaded by old Julia versions which link to v6.1.2.  However GMP is backward compatible, so we can use GMP v6.1.2 when building:
  ```julia
  Dependency("GMP_jll", v"6.1.2"; compat="6.1.2")
  ```
* [`ICU_jll`](https://juliahub.com/ui/Packages/ICU_jll/hGiiW/): the soname seems to chagne every major version ([example](https://github.com/JuliaBinaryWrappers/ICU_jll.jl/commit/97483d946bec181a1e118113eb953cde2339035e#diff-f931c317b0588bc69525070a5938c944e354886fcf5451e2f0d993462caf264d))
  ```julia
  Dependency("ICU_jll", v"65.1.0"; compat="65.1")
  Dependency("ICU_jll", v"67.1.0"; compat="67.1") # the most common version, at the moment
  Dependency("ICU_jll", v"68.2.0"; compat="68.2")
  ```
[...]

PR #4282, opened 2022-01-19
[boost] Build boost 1.74

jeremiahpslewis:
Add in a missing boost version I could use for a project, will then follow up with a PR for 1.75, then 1.77, then 1.78

### DIFF
--- a/B/boost/build_tarballs.jl
+++ b/B/boost/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "boost"
-version = v"1.76.0"
+version = v"1.74.0"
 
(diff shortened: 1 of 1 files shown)


giordano:
Side note: boost breaks the ABI in every single minor version.  Unless you have a strong need for package which absolutely needs a specific version I'd recommend _**against**_ adding new versions which serve the only purpose of fragmenting the ecosystem

giordano:
I stand corrected: not every single minor version, but they break the ABI in _**every**_ version, as they embed the full version number in the soname: #3024.  Again, unless this is hugely important I'd _**not**_ add more versions

PR #4289, opened 2022-01-20
[libblastrampoline] Build v4.0.0

staticfloat:
Chewing through version numbers like it's going out of style, let's add LBT v4.0.0 which allows for loading dual-interface libraries such as MKL v2022+.

### DIFF
--- a/L/libblastrampoline/build_tarballs.jl
+++ b/L/libblastrampoline/build_tarballs.jl
@@ -13,6 +13,9 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
+ls -la ${WORKSPACE}
+echo
+ls -la ${WORKSPACE}/srcdir
 cd $WORKSPACE/srcdir/libblastrampoline/src
 
 make -j${nproc} prefix=${prefix} install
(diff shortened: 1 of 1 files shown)


L/libblastrampoline/build_tarballs.jl
```diff
@@ -3,16 +3,19 @@
...
 
 name = "libblastrampoline"
-version = v"3.1.0"
+version = v"4.0.0"
 
 # Collection of sources required to build libblastrampoline
 sources = [
-    GitSource("https://github.com/staticfloat/libblastrampoline",
-              "c6c7bc5d4ae088bd7c519d58e3fb8b686d00db0c")
+    GitSource("https://github.com/JuliaLinearAlgebra/libblastrampoline",
```
giordano, line 10: I'm away from computer, but I'd try 
```suggestion
    GitSource("https://github.com/JuliaLinearAlgebra/libblastrampoline.git",
```
We have some dodgy assumptions about the fact the URL should end with `.git`

giordano: Uhm, no, reading the full log I'm not sure that's the issue. Maybe try cleaning up the cached repository on the build machine?

giordano: It was indeed the missing `.git` in the URL.  We should fix those assumptions, in the meantime, adding the trailing `.git` is a working fix

PR #4293, opened 2022-01-20
[PCRE2] Update source URL and hash

giordano:
The old FTP website is gone, together with the old release tarballs.  Sigh.


giordano:
Of course since there are no release tarballs anymore, the build script as is doesn't work any longer.  Brilliant

PR #4296, opened 2022-01-21
[libblastrampoline] Fix julia compat to v1.8

giordano:
Version 4 of libblastrampoline breaks the ABI compared to v3, and it's used only in julia v1.8: https://github.com/JuliaLang/julia/pull/43877.  We should reflect these constraints in the registry.

CC @staticfloat

### DIFF
--- a/L/libblastrampoline/build_tarballs.jl
+++ b/L/libblastrampoline/build_tarballs.jl
@@ -34,7 +34,7 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat="1.6",
+               julia_compat="1.8",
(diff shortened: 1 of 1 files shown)


L/libblastrampoline/build_tarballs.jl
```diff
@@ -34,7 +34,7 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat="1.6",
+               julia_compat="1.8",
                init_block = """
                @static if VERSION < v"1.7.0-DEV.641"
```
staticfloat, line 39: Then I guess we can get rid of this `init_block` too?

giordano: Good point!

PR #4298, opened 2022-01-21
Update QuantumEspresso to 7.0

mfherbst:


mfherbst:
It doesn't seem to like gfortran3 any more ...

Q/QuantumEspresso/build_tarballs.jl
```diff
@@ -80,4 +80,4 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               preferred_gcc_version=v"5", julia_compat="1.6", preferred_llvm_version=v"11")
+               preferred_gcc_version=v"5", julia_compat="1.6")
```
giordano, line 83: Let's see whether gcc 6 makes any difference:
```suggestion
               preferred_gcc_version=v"6", julia_compat="1.6")
```

mfherbst: yeah, good idea.

mfherbst:
Any clue on that weird real versus complex issue. In particular I'm surprised to only see it for some libgfortran versions on aarch64-apple.

giordano:
For aarch64-apple-darwin [the error](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16421&view=logs&j=177558ba-96a5-5472-3281-acd2b316ecb4&t=592f4724-25c7-5af5-ca54-99f8bf35cfe0&l=10015) is
```
[06:50:23] ld: library not found for -lscalapack
[06:50:23] collect2: error: ld returned 1 exit status
```
which makes sense since at the moment we don't have `SCALAPACK_jll` for that platform.  Same thing for armv6l platforms.  I think we have to filter those platforms out until we build `SCALAPACK_jll` there.

mfherbst:
Strange, because it works ok with the other libgfortran versions.

giordano:
Not sure what "other libgfortran versions" you're referring to.

Q/QuantumEspresso/build_tarballs.jl
```diff
@@ -30,8 +30,10 @@ if [ "${nbits}" == 64 ]; then
     flags+=(--with-libxc=yes --with-libxc-prefix=${prefix})
 fi
 
-if [[ "${target}" == powerpc64le-linux-gnu ]]; then
-    # No scalapack binary available on PowerPC
+if [[    "${target}" == powerpc64le-linux-gnu 
+      || "${target}" == armv6l-linux-musleabihf \
+      || "${target}" == armv6l-linux-gnueabihf ]]; then
```
giordano, line 35:
```suggestion
      || "${target}" == armv6l-linux-* ]]; then
```

mfherbst:
Ah, sorry my bad ... there only is one libgfortran for aarch64-apple-darwin ...

Q/QuantumEspresso/build_tarballs.jl
```diff
@@ -30,8 +30,10 @@ if [ "${nbits}" == 64 ]; then
     flags+=(--with-libxc=yes --with-libxc-prefix=${prefix})
 fi
 
-if [[ "${target}" == powerpc64le-linux-gnu ]]; then
-    # No scalapack binary available on PowerPC
+if [[    "${target}" == powerpc64le-linux-* \
+      || "${target}" == armv6l-linux-* \
+      || "${target}" == aarch64-apple-darwin ]]; then
```
giordano, line 35:
```suggestion
      || "${bb_full_target}" == armv6l-linux-* \
      || "${target}" == aarch64-apple-darwin* ]]; then
```

giordano:
```
[19:47:58] f951: internal compiler error: in doloop_contained_procedure_code, at fortran/frontend-passes.c:2464
```
I don't think there is much hope for aarch64-apple-darwin at the moment

mfherbst:
I agree ... that might sound stupid, but how do I do that without removing the other apple targets as well?

PR #4299, opened 2022-01-21
Update TempestRemap to v2.1.2 bugfix release

jakebolewski:


giordano:
If you closed because of the error, my educated guess is that https://github.com/JuliaPackaging/Yggdrasil/blob/fa063b67726a6f1d2e00a949e730128d94a41a23/T/TempestRemap/build_tarballs.jl#L81 needs to use a `VersionNumber` as version, also because of https://github.com/JuliaLang/Pkg.jl/issues/2938

jakebolewski:
Thanks @giordano, appreciated!

T/TempestRemap/build_tarballs.jl
```diff
@@ -78,7 +76,7 @@ dependencies = [
     Dependency("NetCDF_jll"),
     Dependency("HDF5_jll"),
     # The following is adapted from NetCDF_jll
-    BuildDependency(PackageSpec(; name="MbedTLS_jll", version="2.24.0")),
+    BuildDependency(PackageSpec(; name="MbedTLS_jll", version=VersionNumber("2.24.0"))),
```
giordano, line 79:
```suggestion
    BuildDependency(PackageSpec(; name="MbedTLS_jll", version=v"2.24.0")),
```

PR #4300, opened 2022-01-22
[SDL] Bump versions (help wanted)

Gnimuc:
- SDL2: 2.0.12 => 2.0.20
- SDL2_ttf: 2.0.15 => 2.0.18

Also switch from ArchiveSource to GitSource.

### DIFF
--- a/S/SDL2_ttf/build_tarballs.jl
+++ b/S/SDL2_ttf/build_tarballs.jl
@@ -15,7 +15,7 @@ version = v"2.0.16" # <-- this version number is a lie to build for Julia v1.6
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/SDL2_ttf-*/
+cd $WORKSPACE/srcdir/SDL*/
 
 atomic_patch -p1 ../patches/configure_in-v2.0.15.patch
 atomic_patch -p1 ../patches/Makefile_in_dont_build_programs.patch
--- a/S/SDL2_ttf/build_tarballs.jl
+++ b/S/SDL2_ttf/build_tarballs.jl
@@ -3,12 +3,11 @@
 using BinaryBuilder
 
 name = "SDL2_ttf"
-version = v"2.0.15"
+version = v"2.0.18"
 
 # Collection of sources required to build SDL2_ttf
 sources = [
-    ArchiveSource("https://www.libsdl.org/projects/SDL_ttf/release/SDL2_ttf-$(version).tar.gz",
-                  "a9eceb1ad88c1f1545cd7bd28e7cbc0b2c14191d40238f531a15b01b1b22cd33"),
+    GitSource("https://github.com/libsdl-org/SDL_ttf.git", "3e702ed9bf400b0a72534f144b8bec46ee0416cb"),
     DirectorySource("./bundled"),
 ]
 
--- a/S/SDL2/build_tarballs.jl
+++ b/S/SDL2/build_tarballs.jl
@@ -3,12 +3,11 @@
 using BinaryBuilder
 
 name = "SDL2"
-version = v"2.0.12"
+version = v"2.0.20"
 
 # Collection of sources required to build SDL2
 sources = [
-    ArchiveSource("http://www.libsdl.org/release/SDL2-$(version).tar.gz",
-                  "349268f695c02efbc9b9148a70b85e58cefbbf704abd3e91be654db7f1e2c863"),
+    GitSource("https://github.com/libsdl-org/SDL.git", "b424665e0899769b200231ba943353a5fee1b6b6"),
 ]
 
 # Bash recipe for building across all platforms
@@ -55,4 +54,3 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
-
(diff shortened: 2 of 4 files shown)


Gnimuc:
Looks like the build scripts for SDL_xxx are not trivial to be fixed. I don't have time to dive into the details, so waiting for others to pick this up.

giordano:
Not sure switching to Git sources was a good idea, now Windows builds fail with
```
[05:06:38] configure: error: 
[05:06:38] *** When building from a git clone you should configure and build in a
[05:06:38]     separate directory so you don't clobber SDL_config.h, SDL_revision.h
```

giordano:
And SDL2_image is successful if using the release tarball.

giordano:
And also for SDL2_mixer using the release tarball gets us past the bootstrap failure.  However during the build I get a different issue:
```
[18:58:15] grep: /workspace/destdir/lib/libogg.la: No such file or directory
[18:58:15] sed: /workspace/destdir/lib/libogg.la: No such file or directory
[18:58:15] libtool: link: `/workspace/destdir/lib/libogg.la' is not a valid libtool archive
[18:58:15] make: *** [Makefile:65: build/libSDL2_mixer.la] Error 1
```
@fingolfin we're getting again this issue, again with libogg.

giordano:
Side note: having a single PR which builds multiple package isn't a good idea either.  It makes working through warnings/errors rather cumbersome

giordano:
Another side effect of having a single PR building multiple packages is that now SDL2 works fine (after switching back to the release tarball instead of the git source) but we can't merge it.

I guess there is enough evidence that using the git source was a bad idea.

S/SDL2_ttf/build_tarballs.jl
```diff
@@ -3,20 +3,19 @@
...
 
 name = "SDL2_ttf"
-version = v"2.0.15"
+version = v"2.0.18"
 
 # Collection of sources required to build SDL2_ttf
 sources = [
-    ArchiveSource("https://www.libsdl.org/projects/SDL_ttf/release/SDL2_ttf-$(version).tar.gz",
-                  "a9eceb1ad88c1f1545cd7bd28e7cbc0b2c14191d40238f531a15b01b1b22cd33"),
+    GitSource("https://github.com/libsdl-org/SDL_ttf.git", "3e702ed9bf400b0a72534f144b8bec46ee0416cb"),
```
giordano, line 10: You should got back to the release tarball also here, and refresh the patches

S/SDL2_ttf/build_tarballs.jl
```diff
@@ -3,20 +3,19 @@
...
     DirectorySource("./bundled"),
 ]
 
 version = v"2.0.16" # <-- this version number is a lie to build for Julia v1.6
```
giordano, line 14: Note that this line must be removed

S/SDL2_ttf/build_tarballs.jl
```diff
@@ -3,20 +3,19 @@
...
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/SDL2_ttf-*/
+cd $WORKSPACE/srcdir/SDL*/
```
giordano, line 18:
```suggestion
cd $WORKSPACE/srcdir/SDL2_ttf-*/
```

giordano (changes requested):

fingolfin:
> And also for SDL2_mixer using the release tarball gets us past the bootstrap failure. However during the build I get a different issue:
> 
> ```
> [18:58:15] grep: /workspace/destdir/lib/libogg.la: No such file or directory
> [18:58:15] sed: /workspace/destdir/lib/libogg.la: No such file or directory
> [18:58:15] libtool: link: `/workspace/destdir/lib/libogg.la' is not a valid libtool archive
> [18:58:15] make: *** [Makefile:65: build/libSDL2_mixer.la] Error 1
> ```
> 
> @fingolfin we're getting again this issue, again with libogg.

There is no direct reference to libogg in `SDL2_mixer`, so it must be indirect. And indeed, mpg123 references libogg.la in its artifacts, and was last rebuilt November 2019. I'll submit an update for it now.

S/SDL2_ttf/build_tarballs.jl
fingolfin, line 18: Maybe
```suggestion
cd $WORKSPACE/srcdir/SDL2_ttf*/
```
so it'll work with both git clones and release tarballs.

giordano:
After #4307, #4308, and #4309 now also SDL2_mixer is working.  Thanks Max for the tip.

Yet another issue with bundling multiple packages in a single PR is that SDL2_* packages depend on SDL2 for the new platforms (aarch64-apple-darwin and armv6l).  So we can't possible merge this PR as is.

Gnimuc:
Thanks for all the help! I opened another PR #4310 to update SDL2.

giordano:
SDL2_mixer is good to go, but once again, being together with other packages which are failing makes it impossible to merge it alone.  There are failures for macOS in SDL2_image and you still need to refresh the patches for SDL2_ttf.  I don't think I'm going to do anything more here.

Gnimuc:
I can successfully build SDL2_image's lastest git main branch for x86_64-apple-darwin14 using its CMake build. But it looks like the M1 support is still in progress(https://github.com/libsdl-org/SDL_image/issues/182#issuecomment-909488050). I think we can try this again when new versions are released in the upstream. At this moment, I only need to update SDL2 to fix https://github.com/JuliaMultimedia/SimpleDirectMediaLayer.jl/issues/60.

PR #4307, opened 2022-01-22
Update mpg123

fingolfin:


giordano:
Yes, this package was last built looooong time ago.

giordano:
Lovely, thanks a lot!

fingolfin:
:-) Let's see if it helps! There could very well be another ref...

(This led to me grepping my `~/.julia/artifacts` dir for `.la` file references, and I found a ton, but I think most are outdated... then I realized it is 25 G on one machine and 41 G on another, so now I wiped it and will start over :joy:)

PR #4311, opened 2022-01-23
Restore the hard fail if the registry download fails

DilumAluthge:
This undoes the changes of https://github.com/JuliaPackaging/Yggdrasil/pull/4304

I think having this hard fail is good. It lets us immediately notice when something is wrong; in this case, it quickly notified us that the CSAIL Pkg server was down. So I'd like to restore the hard fail. When e.g. the CSAIL Pkg server has downtime, I'd prefer that we keep the hard fail and just temporarily use a different Pkg server, as in https://github.com/JuliaPackaging/Yggdrasil/pull/4305

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -49,13 +49,7 @@ jobs:
       $(JULIA) --compile=min -O0 -e 'using InteractiveUtils, Pkg, Downloads, Dates
           versioninfo()
           if !isnothing(Pkg.pkg_server())
-              resp = try
-                  Downloads.request("$(Pkg.pkg_server())/registries")
-              catch e
-                  # Let us know the download of the registry went wrong, but do not hard fail
-                  @error "Could not download the registry" exception=(e, catch_backtrace())
-                  exit(0)
-              end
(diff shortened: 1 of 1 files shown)


giordano:
I think `@error` is enough of a warning, without completely blocking any operation

DilumAluthge:
Only if you check the log... I'm guessing most people don't look at the logs if CI is green.

giordano:
Well, I do :slightly_smiling_face:

giordano:
I think there are roughly two classes of users: maintainers and contributors.  Contributors want to check whether their recipes will work, don't care much about the fact the PkgServer may not be working at the moment.  Maintainers should _**always**_ check the log before merging.  And maintainers are also those who can do something about the PkgServer not working, or alerting who needs to be alerted.  I don't see any benefit in breaking everything for contributors, who won't even be able to check whether their recipe works or not.

jeremiahpslewis:
What about adding a package server check as a separate GitHub Actions task? But not cause binary build runs to fail? That way PRs will build, but not go all checks will go green?

PR #4312, opened 2022-01-23
Update build script for SDL2_mixer

Gnimuc:
This is a part of #4300.


giordano:
I didn't notice this didn't change the version number

Gnimuc:
It turns out there is no need to update `SDL2_xxx`s because we only upgraded SDL2 to a new patch version.

PR #4315, opened 2022-01-24
[TempestRemap] linking against libblastrampoline causes segfaults in some utilities so switch to OpenBLAS32

jakebolewski:
Until the build / runtime issues with libblastrampoline are resolved, this is a workaround.

### DIFF
--- a/T/TempestRemap/build_tarballs.jl
+++ b/T/TempestRemap/build_tarballs.jl
@@ -23,8 +23,8 @@ mkdir -p build && cd build
 ../configure \
   --prefix=${prefix} \
   --host=${target} \
-  --with-blas=blastrampoline \
-  --with-lapack=blastrampoline \
+  --with-blas=openblas \
+  --with-lapack=openblas \
   --with-netcdf=${prefix} \
   --with-hdf5=${prefix} \
   --enable-shared \
@@ -72,7 +72,7 @@ products = [
 ]
 
 dependencies = [
-    Dependency("libblastrampoline_jll"),
+    Dependency("OpenBLAS32_jll"),
     Dependency("NetCDF_jll"),
     Dependency("HDF5_jll"),
     # The following is adapted from NetCDF_jll


simonbyrne:
Why not use openblas, since that is bundled with Julia?

jakebolewski:
because then we have to re-write all the symbols

jakebolewski:
If you want to take a crack at it see: https://github.com/JuliaPackaging/Yggdrasil/blob/8be19820afa24ff817185f3c53c4d20786a8add3/A/armadillo/build_tarballs.jl#L23

jakebolewski:
The built in build script already auto-detects MKL (as that is the default BLAS / LAPACK used for non-os deploy builds of tempest remap).

giordano:
If you don't want to remap all symbols then you can use OpenBLAS32_jll

T/TempestRemap/build_tarballs.jl
```diff
@@ -72,13 +69,13 @@ products = [
...
+    Dependency("MKL_jll"),
     Dependency("NetCDF_jll"),
     Dependency("HDF5_jll"),
     # The following is adapted from NetCDF_jll
     BuildDependency(PackageSpec(; name="MbedTLS_jll", version=v"2.24.0")),
 ]
 
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-    julia_compat="1.7",
+    julia_compat="1.6",
```
giordano, line 80: I don't recommend changing the compat bound. This change may make registration to fail (not sure), and in case you want to come back to libblastrampoline later you'll to switch back to 1.7 again

jakebolewski:
Does Julia ship with OpenBLAS32 that or would that be another package to download?

T/TempestRemap/build_tarballs.jl
jakebolewski: ok makes sense

giordano:
OpenBLAS32_jll is another package (another download), with the LP64 interface for all platforms. The openblas which comes with julia uses ILP64 on 64-bit platforms, and LP64 on 32-bit ones. If you're worried about downloading extra stuff, note that mkl is way larger than openblas, and also available on way fewer platforms

jakebolewski:
That seems like a better option and works out of the box:

```
..Histogram of weights
....Column 1: Lower bound on weights
....Column 2: Upper bound on weights
....Column 3: # of weights in that bin
..[[0, 1, 2430]]
------------------------------------------------------------
Writing offline map.. Done
------------------------------------------------------------
Test Summary:                             | Pass  Total
mesh coordinates ClimaCore > TempestRemap |    4      4
     Testing ClimaCoreTempestRemap tests passed 
```

PR #4317, opened 2022-01-25
update hidapi to 0.11.2

laborg:
This updates hidapi to version 0.11.2

### DIFF
--- a/H/hidapi/build_tarballs.jl
+++ b/H/hidapi/build_tarballs.jl
@@ -48,6 +48,6 @@ dependencies = [
     Dependency("libusb_jll")
 ]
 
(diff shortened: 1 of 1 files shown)


H/hidapi/build_tarballs.jl
```diff
@@ -48,6 +48,6 @@ dependencies = [
     Dependency("libusb_jll")
 ]
 
-platforms = [p for p in supported_platforms() if !Sys.isfreebsd(p)]
+platforms = [p for p in supported_platforms(;experimental=true) if !Sys.isfreebsd(p)]
```
giordano, line 51: This can be simplified a bit with
```suggestion
platforms = supported_platforms(;exclude=Sys.isfreebsd)
```

PR #2967, opened 2021-05-11
[Nettle] Upgrade to v3.7.2

giordano:

PR #4322, opened 2022-01-26
After building rr, run the rr test suite

DilumAluthge:
I think that we should run the rr test suite as part of the Ygg build.

The motivation here is that recently, Base Julia CI broke because of an update to the Ygg rr. By running the tests here, we can catch those breakages early, instead of waiting until it hits Base CI.

### DIFF
--- a/R/rr/build_tarballs.jl
+++ b/R/rr/build_tarballs.jl
@@ -19,8 +19,13 @@ cd ${WORKSPACE}/srcdir/rr/
 mkdir build && cd build
 cmake -DCMAKE_BUILD_TYPE=Release \
       -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
-      -Ddisable32bit=ON -DBUILD_TESTS=OFF -DWILL_RUN_TESTS=OFF -Dstaticlibs=ON ..
+      -Ddisable32bit=ON -DBUILD_TESTS=ON -DWILL_RUN_TESTS=ON -Dstaticlibs=ON ..
 make -j${nproc}
+
+# WIP
+# Run the rr test suite
+ctest --output-on-failure -j${nproc}
+
 make install
 """
 


DilumAluthge:
@Keno I think we'll need to add some more build-time (build-only) dependencies for this to work.

@staticfloat or @Keno Could you point me in the right direction?

giordano:
No, you have to use kernel headers, like in https://github.com/Keno/rr/commit/6876b6e085bd2467fa72fa4d1fd47bf07be59161

giordano:
Or like #2967

giordano:
That patches a different file

Keno:
I don't think it really makes sense to run the tests here. We don't do it for anything else. Why is rr special?

DilumAluthge:
The main motivation is to catch breakages before a new rr_jll is deployed to Buildkite.

The alternative option would be to pin rr_jll to a specific commit hash in the Buildkite config. The downside of this approach is that it adds an additional manual step. After merging the Ygg PR and waiting for the General PR to be auto-merged, a human now needs to bump the rr_jll commit in the Buildkite config, make a PR to JuliaCI/julia-buildkite, and wait for tests to finish running on that PR before merging it.

Whereas with this PR, it doesn't add any new manual steps.

DilumAluthge:
@staticfloat Actually you know what I think would be ideal? If we could somehow add rr_jll into our rootfs image for the linux64 tester? Then, the version of rr_jll that is deployed is pinned to the rootfs treehash.

However, this would require that we have some ability to install a JLL package inside a rootfs image and then call it from the command linez

DilumAluthge:
No, scratch that suggestion. I still want to be able to run the `rr_capture.jl` script locally, because it correctly handles all the Distributed worker processes. So we still need to use rr_jll in that script.

DilumAluthge:
> We don't do it for anything else.

For the other JLLs used in Julia (e.g. the OpenBLAS JLLs), we pin those JLLs by commit in the Julia source tree.

Which, I guess, supports the argument that we should not do this PR, and instead we should just pin `rr_jll` by commit in the Julia Buildkite configuration.

staticfloat:
Agreed; Yggdrasil is the wrong place to test.

staticfloat:
I think we're probably going to end up writing a buildkite plugin to deploy JLLs.

PR #3801, opened 2021-10-23
[PETSc] Attempt to rebuild with good include files (and don't rename libpetsc)

Wimmerer:
This is just a first attempt at building PETSc a little differently. It may be totally the wrong way to do it, but it achieves two things: keeps generated include files for each scalar type version, and avoids renaming.

I'm not sure the latter is necessary, but it helps me avoid symlinking/digging into build systems for packages that depend on `-lpetsc`.

PR #4323, opened 2022-01-26
New Recipe: qdldl v0.1.5

bjack205:
This is for the qdldl library for solving quasidefinite symmetric systems. 
Provides support for both `Float32` and `Float64` types, available as separate shared libraries.

### DIFF
--- a/Q/qdldl/build_tarballs.jl
+++ b/Q/qdldl/build_tarballs.jl
@@ -1,0 +1,45 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "qdldl"
+version = v"0.1.5"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/osqp/qdldl/archive/refs/tags/v$(version).tar.gz", "2868b0e61b7424174e9adef3cb87478329f8ab2075211ef28fe477f29e0e5c99")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd qdldl-0.1.5/
+mkdir build_double
+cd build_double/
+cmake -DCMAKE_INSTALL_PREFIX=$prefix/double -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DDFLOAT=0 ..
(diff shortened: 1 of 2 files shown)


Q/qdldl/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libqdldl", :libqdldl64, "double/lib"),
+    LibraryProduct("libqdldl", :libqdldl32, "single/lib")
```
giordano, line 37: Unfortunately it isn't a good idea to have two libraries with same soname in the same process.  It's hugely problematic: #3801

giordano: @imciner2 you seem to be familiar with this library.  Is there a way to build the two distinct variants next to each other with different sonames by default?  This is what FFTW does, which makes it possible for different variants to coexist, but having same soname only causes problems.

bjack205: Bummer. I'll see if I can get the authors to provide an option to add a suffix to the library name.

bjack205:
I've modified this so that it only exports one library for now. I'll see if I can find a way to get add the single-precision library down the road.

giordano:
I renamed the project to match the name used in the [README](https://github.com/osqp/qdldl/blob/f2665f898d88ef7d3b1d8770cad3cff547c846c5/README.md).  Thanks for your contribution!

BTW, I noticed there is a pure-julia implementation at https://github.com/osqp/QDLDL.jl :slightly_smiling_face:

Q/qdldl/build_tarballs.jl
imciner2: The short answer is no, there is no way to build the two variants next to each other right now, and I don't think any other consumer of QDLDL has ever wanted to provide both a float32 and float64 variant, they usually have just used float64 if they provide a library (or at least no one has reached out to us about it). Is there a real need to provide float32 in the C library when there is QDLDL.jl that can handle any type under `AbstractFloat` since it is a pure Julia implementation?

bjack205:
Unfortunately the pure Julia version doesn't seem to be very actively maintained right now. The Julia version is actually faster, as well!

imciner2:
Not all packages will have constant commits, so I don't know why you claim it isn't actively maintained. There are no open issues and the only 2 open PRs were opened within the last month - during which there has been a major holiday season.

PR #4324, opened 2022-01-27
Update P4est to v2.8 and enable MPI support

lchristm:
* Use p4est v2.8 which is the current stable release
* Compile p4est with MPI support (against MicrosoftMPI_jll on Windows and MPICH_jll on all other platforms)
* Compilation with MPI enabled fails for 32 bit Windows, hence that platform is no longer supported

cc @sloede


giordano:
@sloede I presume this is good to go, right?

sloede:
Yes, sorry for not being clear about that.

PR #4325, opened 2022-01-27
[MUMPS] Compile MUMPS without ParMETIS

amontoison:
On OSX, we have a segmentation fault with the error
```
stack_not_16_byte_aligned_error at /usr/lib/system/libdyld.dylib (unknown line)
```
I would like to test without `ParMETIS`.

### DIFF
--- a/M/MUMPS/MUMPS@5/build_tarballs.jl
+++ b/M/MUMPS/MUMPS@5/build_tarballs.jl
@@ -28,8 +28,8 @@ make_args+=(OPTF=-O \
             CDEFS=-DAdd_ \
             LMETISDIR=${prefix} \
             IMETIS="-I${includedir}" \
-            LMETIS="-L${libdir} -lparmetis -lmetis" \
-            ORDERINGSF="-Dpord -Dparmetis" \
+            LMETIS="-L${libdir} -lmetis" \
+            ORDERINGSF="-Dpord -Dmetis" \
             CC="mpicc -fPIC" \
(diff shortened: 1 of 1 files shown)


giordano:
If you want to do a test it may be easier to [build it locally and deploy it to a remote repository](https://docs.binarybuilder.org/stable/building/#Building-a-custom-JLL-package-locally)?  I'm not happy about changing dependencies within the same `X.Y.Z` version number, especially if just for tests, that's a mess in the registry.

amontoison:
Yes, you're right. I never did it but It was easy !
I tested with my fork `amontoison/MUMPS_jll̀` and the error is not related to `ParMETIS` :(

Is it possible that I did something wrong when I renamed the sonames of `libscalapack` on OSX ?
https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SCALAPACK32/build_tarballs.jl#L53-L64

giordano:
> Is it possible that I did something wrong when I renamed the sonames of libscalapack on OSX ?

No, I don't think just renaming the library is problematic (unless there is a serious bug in `install_name_tool`, but we use it all the time).

amontoison:
Ok, thanks!
And by curiosity, did you already encounter this [error](https://github.com/JuliaSmoothOptimizers/MUMPS.jl/runs/4970755578?check_suite_focus=true#step:6:92) with another artifact ?

giordano:
No, this pull request is the only result of the search for `stack_not_16_byte_aligned_error` in this organisation.

PR #4327, opened 2022-01-27
[vegafusion] build script

jeremiahpslewis:
Adds module which can potentially be used together with Vegalite.jl to enable massive dataset processing for dataviz.
Seems cool, want to nudge towards Julia support, so checking out how easily we can build it:

https://medium.com/@jonmmease/announcing-vegafusion-570f62207ba7


V/vegafusion/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+install_license ../LICENSE
+"""
+
+platforms = supported_platforms()
+# Our Rust toolchain for i686 Windows is unusable
+filter!(p -> !Sys.iswindows(p) || arch(p) != "i686", platforms)
+
+# Build fails with error 'BinaryBuilder: Cannot force an architecture via -march'
+# while compiling c code in https://github.com/briansmith/ring dependency
+filter!(p -> arch(p) ∉ ("armv6l", "armv7l", "i686"), platforms)
```
giordano, line 28: Too many Rust packages do this, but they usually force the base architecture (so why doing that to start with?  no clue), if that's the case you can use `lock_microarchitecture=false` to skip the check

V/vegafusion/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+version = v"0.0.4"
+
+sources = [
+    GitSource("https://github.com/vegafusion/vegafusion.git",
+                  "c0927268fc7544878c58a606e01503ba3fed615e"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+export TARGET_CC=$CC
```
giordano, line 13: What is this used for?

V/vegafusion/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
+using BinaryBuilder
+
+name = "vegafusion"
+version = v"0.0.4"
+
+sources = [
+    GitSource("https://github.com/vegafusion/vegafusion.git",
+                  "c0927268fc7544878c58a606e01503ba3fed615e"),
```
giordano, line 8: Indentation....
```suggestion
              "c0927268fc7544878c58a606e01503ba3fed615e"),
```

V/vegafusion/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
+using BinaryBuilder
+
+name = "vegafusion"
```
giordano, line 3:
```suggestion
name = "VegaFusion"
```

V/vegafusion/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+export TARGET_CC=$CC
+
+cd ${WORKSPACE}/srcdir/vegafusion/vegafusion-server/
+cargo build --release
+mkdir -p "${bindir}"
+cp "../target/${rust_target}/release/vegafusion-server${exeext}" "${bindir}/."
```
giordano, line 18: If you want you can simplify with
```suggestion
install -Dm 755 "../target/${rust_target}/release/vegafusion-server${exeext}" "${bindir}/vegafusion-server${exeext}"
```

V/vegafusion/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+install_license ../LICENSE
```
giordano, line 19: Not sure this needed, but it doesn't hurt either.

V/vegafusion/build_tarballs.jl
jeremiahpslewis: I made a number of changes at once, but now I get 

```bash
curl: Can't open './products/vegafusion*x86_64-unknown-freebsd*.tar.gz'!
```

Assume this is related to this fix, but I'll roll back and check. :)

giordano: > Assume this is related to this fix, but I'll roll back and check. :)

Not at all (the failure is even after the tarball is created, how it can be related?), it's due to the fact that the directory name _**must**_ match the `name` of the package, so you have to rename the directory too

V/VegaFusion/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+    GitSource("https://github.com/vegafusion/vegafusion.git",
+              "c0927268fc7544878c58a606e01503ba3fed615e"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/vegafusion/vegafusion-server/
+cargo build --release
+
+mkdir -p "${bindir}"
```
giordano, line 16:
```suggestion
```

giordano:
Side note, could you please _**not**_ tag me (and probably anyone else) in commit messages?  I get multiple annoying notifications because of that.

PR #4333, opened 2022-01-29
[libcint] Update to version 5.1.1

panxl:
Also removed the dependency on OpenBLAS since the library doesn't depend on BLAS anymore since v5.0.0.

### DIFF
--- a/L/libcint/build_tarballs.jl
+++ b/L/libcint/build_tarballs.jl
@@ -36,5 +36,9 @@ products = [
     LibraryProduct("libcint", :libcint)
 ]
 
+# Dependencies that must be installed before this package can be built
+dependencies = [
+]
+
 # Build the tarballs, and possibly a `build.jl` as well.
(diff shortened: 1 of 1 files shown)


L/libcint/build_tarballs.jl
```diff
@@ -39,7 +38,6 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
```
giordano, line 40:
```suggestion
dependencies = Dependency[
```

giordano (commented):
@panxl can we require Julia v1.6?

L/libcint/build_tarballs.jl
panxl: Thanks for the fix!

panxl:
@giordano I think it is reasonable. Is this the right way to add it?

giordano:
Thanks!

PR #4337, opened 2022-01-29
[VMEC] Add NetCDF compat

benjaminfaber:


V/VMEC/build_tarballs.jl
```diff
@@ -66,8 +70,8 @@ dependencies = [
     Dependency("OpenBLAS_jll"),
     Dependency("SCALAPACK_jll"; platforms=filter(!Sys.iswindows, platforms)),
     Dependency("MKL_jll", v"2020.1.216"),
-    Dependency("NetCDF_jll"),
-    Dependency("NetCDFF_jll"),
+    Dependency("NetCDF_jll", v"400.702.400"),
+    Dependency("NetCDFF_jll", v"4.5.3"),
```
giordano, line 74: Note that with this you aren't "pinning" the version, but only selecting the build version: https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.Dependency.  Maybe you wanted to specify also the compat, which is reflected in the generated JLL package?

benjaminfaber: Ah, thanks @giordano, that is what I wanted. One day my PR won't require your intervention and guidance.

giordano: It turns out NetCDF broke the ABI in the 4.7 -> 4.8 transition: https://github.com/JuliaBinaryWrappers/NetCDF_jll.jl/commit/6b1a7db9a5e441cfde9bd794a36c97bfe8338164.  So this compat is not quite right.

giordano: I'm trying to update netcdff and link it to netcdf 4.8 over at #4344, but I'm having quite some troubles with the windows build

PR #4338, opened 2022-01-30
[blaspp] New builder

Wimmerer:
Requirement for LAPACK++, which together are dependencies of SLATE.


Wimmerer:
If `blaspp` is in `$prefix` correctly will `find_package(blaspp)` work in a downstream CMAKE file?

giordano:
It should, yes

PR #4341, opened 2022-01-30
Fix compat bounds of packages depending on NetCDF_jll

giordano:
NetCDF broke the ABI in the 4.7 -> 4.8 transition (the major soversion changed
from 18 to 19), so we always must specify the compat of packages depending on
it.

PR companion to https://github.com/JuliaRegistries/General/pull/53508.

### DIFF
--- a/I/IOAPI/build_tarballs.jl
+++ b/I/IOAPI/build_tarballs.jl
@@ -47,7 +47,11 @@ cp *.h ${includedir} # C header files
 cp fixed_src/* ${includedir} # FORTRAN .EXT (include) files
 
 # Convert static library to dynamic library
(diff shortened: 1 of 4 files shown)


giordano:
There are some failures.  dlopening some libraries fails like
```
ERROR: could not load library "/agent/_work/1/s/I/IOAPI/build/x86_64-linux-gnu-libgfortran5/3uiFOJvw/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libioapi.so"
libmbedtls.so.12: cannot open shared object file: No such file or directory
```
but mbedtls 2.24 should have soname `libmbedtls.so.13`, so I'm not quite sure what's going there.

PR #4346, opened 2022-01-31
Add rocm_smi_lib

Krastanov:
`rocm_smi_lib` provides a library that reads virtual files
in the Linux file system in order to report GPU performance
parameters.

Both the library and the standard Python script that provides
a CLI for the library are included.

On a system with a functional AMD GPU and a Python 3
interpreter it can be tested with:

```julia
run(`python3 $(rocm_smi_lib_jll.rocm_smipy_path)`)
```


Krastanov:
I felt the need to include the python files because:

- they are not python packages, just a single script that provides a CLI interface to the `so` file
- they have hardcoded paths in them which would require to patch them if we package them separately
- that way keeping the versions in sync is trivial
- they happen to not depend on anything else (only standard library)

Krastanov:
And yes, long term we will probably just interface into the `so` library from Julia, without using the python script.

vchuravy:
cc: @jpsamaroo

giordano (changes requested):
The file must be called `build_tarballs.jl`

Krastanov:
That was embarrassing :D Should be fixed now.

giordano:
Can build for more platforms?  Or it's relevant only for x86-64 Linux?

Krastanov:
AMD supports ROCm only on x86-64 Linux. It can probably built on any linux (it is a trivial C library that reads some virtual files), but those virtual files would never exist on another linux.

R/rocm_smi_lib/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/RadeonOpenCompute/rocm_smi_lib/archive/refs/tags/rocm-$(version).tar.gz", "93d19229b5a511021bf836ddc2a9922e744bf8ee52ee0e2829645064301320f4")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/rocm_smi_lib-rocm-*/
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_PREFIX_PATH=${prefix} ..
```
giordano, line 17: You should use the toolchain file https://docs.binarybuilder.org/stable/build_tips/#CMake-builds

Krastanov: oh, noted! I just copied this from `rocminfo_jll`. I will have to look into what these differences mean, as I do not have experience with cmake.

jpsamaroo:
@Krastanov if you want to update this for ROCm 4.2.0, we can pull this JLL into https://github.com/JuliaGPU/AMDGPU.jl/pull/187.

Krastanov:
@jpsamaroo , I will do so by the end of the week.

Krastanov:
Admittedly it is a single character diff, but I would like to check that it actually works on the hardware I have.

Krastanov:
@jpsamaroo , here is the 4.2.0 https://github.com/JuliaPackaging/Yggdrasil/pull/4350

PR #4318, opened 2022-01-25
update to highs wip: errors currently

galabovaa:

PR #4353, opened 2022-02-01
[HiGHS] update to v1.2.0

odow:
Follow-up to https://github.com/JuliaPackaging/Yggdrasil/pull/4318 that I can edit.

There are few things that probably need to be fixed upstream before they make an official 1.2 release.

Closes #4318

### DIFF
--- a/H/HiGHS/build_tarballs.jl
+++ b/H/HiGHS/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.2.0"
 sources = [
     GitSource(
         "https://github.com/ERGO-Code/HiGHS.git",
-        "e0b27f8d89d2277f419841448273fec27563b038",
+        "500cdf47a6330252db4156148f90d99cc8d22cf7",
     ),
     DirectorySource("./bundled"),
 ]
--- a/H/HiGHS/build_tarballs.jl
+++ b/H/HiGHS/build_tarballs.jl
@@ -49,7 +49,6 @@ make install
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = expand_cxxstring_abis(supported_platforms(;experimental=true))
-# platforms = filter!(!Sys.isfreebsd, platforms)
 
 # The products that we will ensure are always built
 products = [
--- a/H/HiGHS/build_tarballs.jl
+++ b/H/HiGHS/build_tarballs.jl
@@ -4,12 +4,12 @@ using BinaryBuilder, Pkg
 
 name = "HiGHS"
 
-version = v"1.1.1"
+version = v"1.2.0"
 
 sources = [
     GitSource(
         "https://github.com/ERGO-Code/HiGHS.git",
-        "31608db682ed3e0fddcdbc5bc899e03fec38c324",
+        "e0b27f8d89d2277f419841448273fec27563b038",
     ),
     DirectorySource("./bundled"),
 ]
@@ -72,6 +72,6 @@ build_tarballs(
     platforms,
     products,
     dependencies;
-    preferred_gcc_version = v"4.9",
+    preferred_gcc_version = v"6",
     julia_compat = "1.6",
 )
--- a/H/HiGHS/build_tarballs.jl
+++ b/H/HiGHS/build_tarballs.jl
@@ -49,7 +49,7 @@ make install
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = expand_cxxstring_abis(supported_platforms(;experimental=true))
-platforms = filter!(!Sys.isfreebsd, platforms)
+# platforms = filter!(!Sys.isfreebsd, platforms)
 
 # The products that we will ensure are always built
 products = [


odow:
cc @lgottwald. The aarch64 issue is fixed. But no luck on the Windows issue.

odow:
There's this upstream issue, which I hadn't seen: https://github.com/ERGO-Code/HiGHS/issues/625

And I'm fairly sure the problem was introduced in https://github.com/ERGO-Code/HiGHS/pull/611

odow:
Per https://gitlab.kitware.com/cmake/cmake/-/issues/22092, CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS is ignored if a `__declspec` is present...

https://github.com/ERGO-Code/HiGHS/blob/7d55c45b08221c6e1ef9b043fe43d9ea2e5aab51/src/parallel/HighsTaskExecutor.h#L46

And that makes sense, because the `globalExecutor` is the only thing exported in `libhighs.dll.a`
```
sandbox:${WORKSPACE}/srcdir/HiGHS/build # nm lib/libhighs.dll.a 

d000077.o:
0000000000000000 i .idata$4
0000000000000000 i .idata$5
0000000000000000 i .idata$7
0000000000000000 I libhighs_dll_iname

d000075.o:
0000000000000000 i .idata$2
0000000000000000 i .idata$4
0000000000000000 i .idata$5
0000000000000000 I _head_libhighs_dll
                 U libhighs_dll_iname

d000076.o:
0000000000000000 i .idata$4
0000000000000000 i .idata$5
0000000000000000 i .idata$6
0000000000000000 i .idata$7
0000000000000000 t .text
0000000000000000 I __imp__ZN17HighsTaskExecutor14globalExecutorE
0000000000000000 I __nm__ZN17HighsTaskExecutor14globalExecutorE
                 U _head_libhighs_dll
```

lgottwald:
That's weird and does not make much sense, because the global executor is in any case required to have a declspec since it is a global variable in a dll and those are not otherwise exported having the export all symbols true or not. Also in the linked issue it states the fix I suggested but maybe it needs to go before the project declaration?
Otherwise there seems to be the linker flag --export-all-symbols that I would try to add to the build for mingw.

H/HiGHS/bundled/patches/cmake.patch
```diff
@@ -0,0 +1,22 @@
+--- check/TestIO.cpp
++++ check/TestIO.cpp
+@@ -63,7 +63,7 @@ TEST_CASE("msgcb", "[highs_io]") {
+     longmsg[sizeof(longmsg) - 2] = '\0';
+     longmsg[sizeof(longmsg) - 1] = '\n';
+     highsLogUser(log_options, HighsLogType::kWarning, longmsg);
+-    REQUIRE(strstr(printedmsg, "HHHH") != NULL);
++    // REQUIRE(strstr(printedmsg, "HHHH") != NULL);
```
odow, line 8: @lgottwald this threw an error about comparing and int and a pointer

H/HiGHS/bundled/patches/cmake.patch
```diff
@@ -0,0 +1,22 @@
...
+   }
+ }
+--- src/parallel/HighsTaskExecutor.h
++++ src/parallel/HighsTaskExecutor.h
+@@ -43,7 +43,7 @@ class HighsTaskExecutor {
+   }
+ #endif
+ 
+-  static LIBHIGHS_EXPORT cache_aligned::shared_ptr<HighsTaskExecutor>
++  static cache_aligned::shared_ptr<HighsTaskExecutor>
```
odow, line 19: @lgottwald was the export actually needed in MSVC? We can safely drop it in MinGW.

lgottwald: Yes it was needed. I only added it because the windows build failed. Apparently the global variable is not affected by the `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` option so that it still needs the export.

lgottwald: It can however be disabled by adding `-Dlibhighs_STATIC_DEFINE` as compile flag as this will cause the export directive to expand to nothing

lgottwald: Could you check if this patch works too?
[mingw.txt](https://github.com/JuliaPackaging/Yggdrasil/files/7983621/mingw.txt)

lgottwald: (edited the patch file, contained some unwanted changes first)

odow: No luck

lgottwald:
Sorry I missed to also add the define for the fast_build option. Also I added (hopefully) a fix for the TestIO.cpp fail but am not 100% sure this fixes the issue
Patch contains both:
[fixes.txt](https://github.com/JuliaPackaging/Yggdrasil/files/7983699/fixes.txt)

odow:
Patches are being upstreamed: https://github.com/ERGO-Code/HiGHS/pull/706/files

odow:
@jajhall: Thanks to the help of @lgottwald things are now compiling, so we're ready for a v1.2 tag :smile:

lgottwald:
We have a slight bit more testing to do in HiGHS but will very soon add the tag

odow:
Seems like our BSD issue is now fixed as well.

giordano:
@odow is this ready?

odow:
Not yet. I'm going to wait for upstream to release an official v.1.2 release so that we stay in-sync. I'll ping you when we're ready.

jajhall:
Looks good, thanks!

odow:
Okay @giordano this is good to go.

odow:
Jeepers. Your reflexes are too quick. Thanks :)

PR #4364, opened 2022-02-04
[RootFS] Update base Alpine Linux image to v3.15 and various utils to more recent versions

giordano:
Companion PR to https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/215


0_RootFS/Rootfs/build_tarballs.jl
```diff
@@ -113,8 +113,8 @@ sources = [
     GitSource("https://github.com/NixOS/patchelf.git",
               "a949ff23315bbb5863627c4655fe216ecbf341a2"),
     # We need a very recent version of meson to build gtk stuffs, so let's just grab the latest
```
giordano, line 115: The funny thing is that this comment is still accurate: we always need to upgrade Meson to the latest version in order to build gtk

PR #4369, opened 2022-02-05
Added ONNXRuntime (cpu)

stemann:


O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,99 @@
...
+# Collection of sources required to complete build
+source_sources = [
+    GitSource("https://github.com/microsoft/onnxruntime.git", "0d9030e79888d1d5828730b254fedc53c7b640c1")
+]
+
+# Bash recipe for building across all platforms
+source_script = raw"""
+cd $WORKSPACE/srcdir
+
+wget https://github.com/Kitware/CMake/releases/download/v3.22.1/cmake-3.22.1-linux-x86_64.sh
```
giordano, line 29: We already have cmake, don't we?

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,109 @@
...
+clone_recursive_path = "onnxruntime"
+clone_recursive_script = """
+rm -rf $(clone_recursive_path)
+git clone $(onnxruntime_source.url) $(clone_recursive_path)
+cd $clone_recursive_path
+git checkout $(onnxruntime_source.hash)
+git submodule update --init --recursive --jobs 8
+cd ..
+"""
+run(`bash -c "$clone_recursive_script"`)
```
giordano, line 19: Please don't do this

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,109 @@
...
+# Bash recipe for building across all platforms
+source_script = raw"""
+cd $WORKSPACE/srcdir
+
+wget https://github.com/Kitware/CMake/releases/download/v3.22.1/cmake-3.22.1-linux-x86_64.sh
+chmod u+x cmake-3.22.1-linux-x86_64.sh 
+./cmake-3.22.1-linux-x86_64.sh --skip-license --include-subdir
+CMAKE=`pwd`/cmake-3.22.1-linux-x86_64/bin/cmake
+
+wget https://github.com/protocolbuffers/protobuf/releases/download/v3.16.1/protoc-3.16.1-linux-x86_64.zip
```
giordano, line 34: What's wrong with https://github.com/JuliaPackaging/Yggdrasil/blob/4ea0a2c9773dbe743abd10f9bd7567a6046ea01b/P/protoc/build_tarballs.jl?

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,109 @@
...
+    mkdir -p \$includedir \$libdir
+    cp -a onnxruntime*/include/* \$includedir
+    if [[ \${target} == *w64* ]]; then
+        chmod 755 onnxruntime*/lib/*
+    fi
+    cp -a onnxruntime*/lib/* \$libdir
+    install_license onnxruntime*/LICENSE
+    """
+    binary_platforms = [platform]
+    build_tarballs(ARGS, name, version, binary_sources, binary_script, binary_platforms, products, dependencies;
```
giordano, line 101: You can't just call build_tarballs multiple times like this. Please stick to the format of all other recipes.

giordano (changes requested):

O/onnxruntime/build_tarballs.jl
stemann: Undone

O/onnxruntime/build_tarballs.jl
stemann: Yeah - I wrote this in early January for BB v0.4 - will update/fix.

O/onnxruntime/build_tarballs.jl
stemann: I thought that was how it was done, e.g. in `CUDA_full@10.2` - when you would like a separate sources/script etc. for a few platforms.

But I guess, I combine the binary_script into the main script and get one `build_tarballs` call.

O/onnxruntime/build_tarballs.jl
stemann: It needs a specific version of `protoc` (3.16): https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16855&view=logs&j=2125d72a-77c5-58f3-9425-89906e3f565e&t=6a65a92e-fb28-56c0-7742-a34594e36aa2

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+    install_license $WORKSPACE/srcdir/onnxruntime/LICENSE
+fi
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+function platform_exclude_filter(p::Platform)
+    libc(p) == "musl" ||
+    p == Platform("i686", "Linux") ||
+    p == Platform("x86_64", "FreeBSD")
```
giordano, line 69:
```suggestion
    Sys.isfreebsd(p)
```

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+    make install
+    install_license $WORKSPACE/srcdir/onnxruntime/LICENSE
+fi
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+function platform_exclude_filter(p::Platform)
+    libc(p) == "musl" ||
+    p == Platform("i686", "Linux") ||
```
giordano, line 68: What's the problem with this one?

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+    p == Platform("x86_64", "FreeBSD")
+end
+platforms = supported_platforms(; exclude=platform_exclude_filter)
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = Product[
```
giordano, line 75:
```suggestion
products = [
```

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+    LibraryProduct(["libonnxruntime", "onnxruntime"], :libonnxruntime)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
```
giordano, line 80:
```suggestion
dependencies = Dependency[
```

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+    cp -a onnxruntime*/include/* \$includedir
+    chmod 755 onnxruntime*/lib/*
+    cp -a onnxruntime*/lib/* \$libdir
+    install_license onnxruntime*/LICENSE
+else
+    if [[ $target == aarch64-apple-darwin ]]; then
+        CMAKE_EXTRA_ARGS+=-DCMAKE_OSX_ARCHITECTURES=arm64
+    fi
+
+    wget https://github.com/protocolbuffers/protobuf/releases/download/v3.16.1/protoc-3.16.1-linux-x86_64.zip
```
giordano, line 42: Still here?

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,86 @@
...
+if [[ $target == *-w64-mingw32 || $target == aarch64-apple-darwin ]]; then
+    if [[ $target == x86_64-w64-mingw32 ]]; then
+        onnxruntime_dist_name=onnxruntime-win-x64-$onnxruntime_version.zip
+    elif [[ $target == i686-w64-mingw32 ]]; then
+        onnxruntime_dist_name=onnxruntime-win-x86-$onnxruntime_version.zip
+    elif [[ $target == aarch64-apple-darwin ]]; then
+        onnxruntime_dist_nameonnxruntime-osx-arm64-$onnxruntime_version.tgz
+    fi
+
+    wget https://github.com/microsoft/onnxruntime/releases/download/v$onnxruntime_version/$onnxruntime_dist_name
```
giordano, line 28: Don't download files during the build, they can't be cached. Add them as additional sources

O/onnxruntime/build_tarballs.jl
stemann: Fixed

O/onnxruntime/build_tarballs.jl
stemann: Fixed

giordano (changes requested):

O/onnxruntime/build_tarballs.jl
stemann: Yes, see https://github.com/JuliaPackaging/Yggdrasil/pull/4369#discussion_r800161663

O/onnxruntime/build_tarballs.jl
stemann: Fixed

O/onnxruntime/build_tarballs.jl
stemann: Fixed

O/onnxruntime/build_tarballs.jl
stemann: Some linker error: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16872&view=logs&jobId=e2698624-9414-515d-1025-6869b251ec37&j=e2698624-9414-515d-1025-6869b251ec37&t=7b855fa8-72f2-562c-496e-00e91cb30c13

O/onnxruntime/build_tarballs.jl
stemann: Fixed

O/onnxruntime/build_tarballs.jl
stemann: Fixed

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,84 @@
...
+        dist_name=onnxruntime-osx-arm64
+    fi
+
+    mkdir -p $includedir $libdir
+    cp -a $dist_name*/include/* $includedir
+    cp -a $dist_name*/lib/* $libdir
+    install_license $dist_name*/LICENSE
+else
+    # Workaround for https://github.com/microsoft/onnxruntime/issues/2152
+    if [[ $target == armv6l-linux-gnueabihf* || $target == armv7l-linux-gnueabihf* ]]; then
```
giordano, line 39:
```suggestion
    if [[ $target == armv-linux-gnueabihf* ]]; then
```

O/onnxruntime/build_tarballs.jl
giordano: There you go: https://github.com/JuliaRegistries/General/pull/54008

O/onnxruntime/build_tarballs.jl
giordano: The script is now _shorter_ and simpler than in [the previous version](https://github.com/JuliaPackaging/Yggdrasil/blob/66f8f17951f3cb6468d6b4492210f80f38fe094f/O/onnxruntime/build_tarballs.jl) when I made this commend

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,83 @@
...
+        chmod 755 $dist_name*/lib/*
+    fi
+
+    mkdir -p $includedir $libdir
+    cp -a $dist_name*/include/* $includedir
+    cp -a $dist_name*/lib/* $libdir
+    install_license $dist_name*/LICENSE
+else
+    if [[ $target == aarch64-apple-darwin* ]]; then
+        cmake_extra_defines=CMAKE_OSX_ARCHITECTURES='arm64'
```
giordano, line 36: We already set [`CMAKE_SYSTEM_PROCESSOR`](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html), is that not enough?  Also, `-arch` is already set correctly:
```console
sandbox:${WORKSPACE} # echo 'int main(){}' |  cc -x c - -###
clang version 12.0.0 (/home/mose/.julia/dev/BinaryBuilderBase/deps/downloads/llvm-project.git d28af7c654d8db0b68c175db5ce212d74fb5e9bc)
Target: arm64-apple-darwin20
Thread model: posix
InstalledDir: /opt/x86_64-linux-musl/bin
 "/opt/x86_64-linux-musl/bin/clang-12" "-cc1" "-triple" "arm64-apple-macosx11.0.0" "-Wundef-prefix=TARGET_OS_" "-Werror=undef-prefix" "-Wdeprecated-objc-isa-usage" "-Werror=deprecated-objc-isa-usage" "-emit-obj" "-mrelax-all" "--mrelax-relocations" "-disable-free" "-disable-llvm-verifier" "-discard-value-names" "-main-file-name" "-" "-mrelocation-model" "pic" "-pic-level" "2" "-mframe-pointer=non-leaf" "-fno-rounding-math" "-munwind-tables" "-fcompatibility-qualified-id-block-type-checking" "-fvisibility-inlines-hidden-static-local-var" "-target-cpu" "apple-a12" "-target-feature" "+neon" "-target-abi" "darwinpcs" "-fallow-half-arguments-and-returns" "-debugger-tuning=lldb" "-nostdinc++" "-resource-dir" "/opt/x86_64-linux-musl/lib/clang/12.0.0" "-isystem" "/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root/usr/include/c++/v1" "-isysroot" "/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root" "-internal-isystem" "/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root/usr/local/include" "-internal-isystem" "/opt/x86_64-linux-musl/lib/clang/12.0.0/include" "-internal-externc-isystem" "/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root/usr/include" "-Wno-unused-command-line-argument" "-fdebug-compilation-dir" "/workspace" "-ferror-limit" "19" "-stack-protector" "1" "-fblocks" "-fencode-extended-block-signature" "-fregister-global-dtors-with-atexit" "-fgnuc-version=4.2.1" "-fmax-type-align=16" "-fcolor-diagnostics" "-o" "/tmp/--3fffb7.o" "-x" "c" "-"
 "/opt/bin/aarch64-apple-darwin20-libgfortran5-cxx11/ld64.aarch64-apple-darwin20" "-dynamic" "-arch" "arm64" "-headerpad_max_install_names" "-macosx_version_min" "11.0.0" "-syslibroot" "/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root" "-o" "a.out" "-L/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/lib" "/tmp/--3fffb7.o" "-lSystem"
```

Should this variable be really necessary (and I'm not yet convinced of that), we should rather put it in [`BuildToolchains.jl`](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/e7286d0d7e0c1b7d724d173529edd12ecb411eba/src/BuildToolchains.jl)

O/onnxruntime/build_tarballs.jl
stemann: Am I waiting on a cache of the General registry?
https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16909&view=logs&j=2125d72a-77c5-58f3-9425-89906e3f565e&t=6a65a92e-fb28-56c0-7742-a34594e36aa2&l=29

giordano: [Yeah](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16909&view=logs&j=96c13d9d-dfd2-51e5-2f76-4a0b3ce09abc&t=09532434-8b0a-5f0e-118d-a3181cde5ec8&l=26)
```
[ Info: PkgServer: https://us-east.pkg.julialang.org/registries.conservative -- last updated: Sun, 06 Feb 2022 14:59:46 GMT (49 minutes, 52 seconds ago)
```

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,85 @@
...
+    make install
+    install_license $WORKSPACE/srcdir/onnxruntime/LICENSE
+fi
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+function platform_exclude_filter(p::Platform)
+    libc(p) == "musl" ||
+    p == Platform("i686", "Linux") || # No binary - and source build fails linking CXX executable onnxruntime_test_all
```
giordano, line 66: You disabled the tests, so maybe this could work now?

O/onnxruntime/build_tarballs.jl
giordano: Did you try without this?

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,85 @@
...
+    Sys.isfreebsd(p)
+end
+platforms = supported_platforms(; exclude=platform_exclude_filter)
+platforms = expand_cxxstring_abis(platforms)
```
giordano, line 70: Uhm, the windows binaries aren't built from source and don't need this, so we only need to expand the C++ string ABI for Linux:
```suggestion
platforms = expand_cxxstring_abis(platforms; skip=!Sys.islinux)
```

O/onnxruntime/build_tarballs.jl
stemann: Seems like it got fixed.

O/onnxruntime/build_tarballs.jl
stemann: Definitely :-)

O/onnxruntime/build_tarballs.jl
stemann: It (also) fails to link `libonnxruntime.so` with `onnxruntime_BUILD_UNIT_TESTS=OFF`: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16907&view=logs&j=e2698624-9414-515d-1025-6869b251ec37&t=7b855fa8-72f2-562c-496e-00e91cb30c13&l=2156

O/onnxruntime/build_tarballs.jl
stemann: Fixed

O/onnxruntime/build_tarballs.jl
stemann: Unfortunately not: https://github.com/JuliaPackaging/Yggdrasil/pull/4369#discussion_r800207354

O/onnxruntime/build_tarballs.jl
stemann: Fixed

O/onnxruntime/build_tarballs.jl
stemann: It fails to link https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=16920&view=logs&j=e9df152d-f6ae-5301-9d40-c17cf2618feb&t=976e9440-1181-57d1-e944-a3ddf8007e43&l=1892

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,86 @@
...
+            CMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+            onnxruntime_BUILD_UNIT_TESTS=OFF \
+            $cmake_extra_defines \
+        --config Release \
+        --parallel $nproc \
+        --path_to_protoc_exe $host_bindir/protoc \
+        --skip_tests \
+        --update
+    cd build/Release
+    make install
```
stemann, line 58: I am a bit split between the recipe that uses the `build.py` script (current) -  - and the recipe that uses cmake directly (cf. f449773c).

Microsoft advertise
* both the `build.py` script for basic CPU builds: https://onnxruntime.ai/docs/build/inferencing.html#build-instructions
* and the direct CMake approach for cross-compiling on Linux: https://onnxruntime.ai/docs/build/inferencing.html#cross-compiling-on-linux 

Any recommendation, @giordano ?

O/onnxruntime/build_tarballs.jl
```diff
@@ -0,0 +1,83 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "onnxruntime"
```
stemann, line 5: This recipe will add `onnxruntime` JLL - and provide the CPU-only `libonnxruntime` LibraryProduct.

As mentioned in https://github.com/jw3126/ONNXRunTime.jl/issues/12, subsequent recipes will add execution providers, cf. https://onnxruntime.ai/docs/build/eps.html

Only some of these EP's are built as shared libraries on top of the CPU-only `libonnxruntime`: https://onnxruntime.ai/docs/build/eps.html#execution-provider-shared-libraries

Is the following sensible?

* `onnxruntime_cuda` to be defined in `O/onnxruntime/onnxruntime_cuda/build_tarballs.jl` will depend on `CUDNN` and provide library products `libonnxruntime` (CUDNN-dependent) and `libonnxruntime_providers_cuda`.
* `onnxruntime_providers_tensorrt` to be defined in `O/onnxruntime/onnxruntime_providers_tensorrt/build_tarballs.jl` will depend on `onnxruntime_cuda` and add `libonnxruntime_providers_tensorrt`.
* `onnxruntime_providers_onednn` to be defined in `O/onnxruntime/onnxruntime_providers_onednn/build_tarballs.jl` will depend on `onnxruntime` and add `libonnxruntime_providers_onednn`.

and so on for e.g. CoreML, AMD MiGraphX.

Or should `onnxruntime_cuda` be `onnxruntime_providers_cuda`?

Might it be an idea to include the CUDA-part in this PR (as it seems the cuda `libonnxruntime` will be independent of the CPU `libonnxruntime`)?

giordano, line 5: Regarding the name, what about `ONNXRuntime`, which matches the case used in [the README](https://github.com/microsoft/onnxruntime/blob/0f5d0a091a37db48593905c33a834597338288e8/README.md)?  Or at least `ONNXRunTime` which matches the cases of [`ONNXRunTime.jl`](https://github.com/jw3126/ONNXRunTime.jl) (which arguably should have been `ONNXRuntime.jl`).  The latter might be a better choice to avoid mixing cases in the middle of the names of two related packages.

In any case remember that the directory name must match _exactly_ the name you use for this variable.

giordano: Look superficially ok, although I'm not very into the GPU side of the libraries.  I know they're pretty complicated to deal with and in most cases I believe we can only build for x86_64-linux-gnu (because `nvcc` isn't a cross-compiler).  If you plan to add other sibling projects you may want to move this now to `O/ONNXRuntime/ONNXRuntime/build_tarballs.jl` (or whatever name you choose to use in the end)

O/onnxruntime/build_tarballs.jl
giordano: Ok, I suspect they're doing something dodgy with `CMAKE_OSX_ARCHITECTURES`, like here https://github.com/microsoft/onnxruntime/blob/481b96d32aecc5fceabc9e9401604d2d31212ef7/tools/ci_build/build.py#L1952-L1953 where they use `platform.machine()` which would use the information about the build machine

O/onnxruntime/build_tarballs.jl
giordano: As mentioned in the comment above, I have the suspicion the python script is trying to outsmart us getting some things wrong.  Directly using CMake may be a better idea

O/onnxruntime/build_tarballs.jl
stemann: Yes, it seems we have to stick with re-packaging binaries for the CUDA-parts for now.

I have renamed to `O/ONNXRuntime/ONNXRuntime/build_tarballs.jl`.

stemann: I have renamed it to `ONNXRuntime`.

O/onnxruntime/build_tarballs.jl
stemann: Done.

And at least a text output definition found for why cross-compiling for `aarch64-apple-darwin` on `x86_64` requires setting `CMAKE_OSX_ARCHITECTURES`: https://github.com/microsoft/onnxruntime/blob/v1.10.0/cmake/CMakeLists.txt#L186

O/onnxruntime/build_tarballs.jl
stemann: Cf. https://github.com/microsoft/onnxruntime/blob/v1.10.0/cmake/CMakeLists.txt#L186

stemann:
I will squash the branch, once we get through the review - I have a manual rebase on hand.

O/ONNXRuntime/ONNXRuntime/build_tarballs.jl
```diff
@@ -0,0 +1,78 @@
...
+using BinaryBuilder, Pkg
+
+name = "ONNXRuntime"
+version = v"1.10.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/microsoft/onnxruntime.git", "0d9030e79888d1d5828730b254fedc53c7b640c1"),
+    ArchiveSource("https://github.com/microsoft/onnxruntime/releases/download/v$version/onnxruntime-win-x64-$version.zip", "a0c6db3cff65bd282f6ba4a57789e619c27e55203321aa08c023019fe9da50d7"),
+    ArchiveSource("https://github.com/microsoft/onnxruntime/releases/download/v$version/onnxruntime-win-x86-$version.zip", "fd1680fa7248ec334efc2564086e9c5e0d6db78337b55ec32e7b666164bdb88c")
```
giordano, line 12: [Trick](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.ArchiveSource):
```suggestion
    ArchiveSource("https://github.com/microsoft/onnxruntime/releases/download/v$version/onnxruntime-win-x64-$version.zip", "a0c6db3cff65bd282f6ba4a57789e619c27e55203321aa08c023019fe9da50d7"; unpack_target="x86_64-w64-mingw32"),
    ArchiveSource("https://github.com/microsoft/onnxruntime/releases/download/v$version/onnxruntime-win-x86-$version.zip", "fd1680fa7248ec334efc2564086e9c5e0d6db78337b55ec32e7b666164bdb88c"; unpack_target="i686-w64-mingw32")
```
So now `${dist_name}` is just `${target}` and you don't have to do the conditional below

O/ONNXRuntime/ONNXRuntime/build_tarballs.jl
```diff
@@ -0,0 +1,78 @@
...
+    elif [[ $target == i686-w64-mingw32* ]]; then
+        dist_name=onnxruntime-win-x86
+    fi
+    chmod 755 $dist_name*/lib/*
+    mkdir -p $includedir $libdir
+    cp -av $dist_name*/include/* $includedir
+    cp -av $dist_name*/lib/* $libdir
+    install_license $dist_name*/LICENSE
+else
+    # Cross-compiling for aarch64-apple-darwin on x86_64 requires setting arch.: https://github.com/microsoft/onnxruntime/blob/v1.10.0/cmake/CMakeLists.txt#L186
```
giordano, line 31: That's dumb, but at least we know why we need this

giordano (commented):
It looks much better now

O/ONNXRuntime/ONNXRuntime/build_tarballs.jl
stemann: Done

O/ONNXRuntime/ONNXRuntime/build_tarballs.jl
stemann: Indeed :-)

PR #4381, opened 2022-02-06
[jemalloc] Try jemalloc 5.3.0 (candidate)

jeremiahpslewis:


J/jemalloc/build_tarballs.jl
```diff
@@ -40,4 +54,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"11.0.0-iains")
```
giordano, line 57: Please, no :-1:

jeremiahpslewis: It's a draft, just wanted to see if it worked!

giordano: `v"11.0.0-iains"` doesn't even make sense for all platforms but `aarch64-apple-darwin` (where you use clang, not even gcc)

jeremiahpslewis: I was looking for a cheap way to trigger MacOS 11.0 and saw a code path which seemed to indicate that might work. But it doesn't resolve the build failure, so it's reverted. :)

PR #4384, opened 2022-02-06
[MozillaCACerts] Update certs

jeremiahpslewis:


M/MozillaCACerts/build_tarballs.jl
```diff
@@ -37,4 +37,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat = "1.6")
```
giordano, line 40: > build for experimental plat compat

What?!?

jeremiahpslewis: Isn't a julia_compat argument necessary for usage by apple silicon?

(Experimental platform compatibility)

giordano: But this is a platform-independent package, you aren't even "building" anything

jeremiahpslewis: Ok, so compat is unnecessary? But reasonable to update to latest certs?

giordano: > Ok, so compat is unnecessary?

Yes

> But reasonable to update to latest certs?

Not sure how they (updating the certificates and changing the compat bound) are related

PR #4402, opened 2022-02-07
[Mineos] Version bump

jeremiahpslewis:


M/Mineos/build_tarballs.jl
```diff
@@ -3,19 +3,32 @@
...
+autoupdate
+autoreconf --install
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+
+# Fix clang error 'error: non-void function * should return a value [-Wreturn-type]'
+if [[ "${target}" != *-freebsd* ]] || [[ "${target}" == *-apple-* ]]; then
+    make CFLAGS="-Wno-return-type $cflags"
+else
+    make
+fi
```
giordano, line 30: Much simpler and doesn't need to have multiple `make` calls
```suggestion
# Fix clang error 'error: non-void function * should return a value [-Wreturn-type]'
if [[ "${target}" != *-freebsd* ]] || [[ "${target}" == *-apple-* ]]; then
    export CFLAGS="-Wno-return-type"
fi
./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
make
```

jeremiahpslewis: Thanks!!

jeremiahpslewis:
@giordano is there hope or should I dump apple silicon? https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=17101&view=logs&jobId=c18725ea-6485-5526-fc57-c82c5921198c&j=c18725ea-6485-5526-fc57-c82c5921198c

giordano:
When everything else fails, [use the search](https://github.com/JuliaPackaging/Yggdrasil/search?q=Rank+mismatch)

jeremiahpslewis:
> When everything else fails, [use the search](https://github.com/JuliaPackaging/Yggdrasil/search?q=Rank+mismatch)

Brilliant!

M/Mineos/build_tarballs.jl
```diff
@@ -3,19 +3,37 @@
 using BinaryBuilder
 
 name = "Mineos"
-version = v"1.0"
+version = v"1.0.2"
```
giordano, line 6: Where is v1.0.2 coming from?  Latest tag [I can see](https://github.com/geodynamics/mineos/tags) is v1.0.1

M/Mineos/build_tarballs.jl
```diff
@@ -3,19 +3,37 @@
 using BinaryBuilder
 
 name = "Mineos"
-version = v"1.0"
+version = v"1.0.2" 
 
 # Collection of sources required to build Mineos
 sources = [
-    "https://github.com/anowacki/mineos.git" =>
-    "e2558b486d7656ef112608a8776643da66dc87cf",
```
giordano, line 11: @anowacki why did you use a fork here?  Would using the upstream repo be ok now?

M/Mineos/build_tarballs.jl
jeremiahpslewis: 1.0.2 is present in the commit message, the version number is also provided in a file, but not an explicit tag

M/Mineos/build_tarballs.jl
anowacki: @giordano The upstream repo has not been maintained for some years, and [my open issue](https://github.com/geodynamics/mineos/issues/10) hasn't received any response.  I can submit a PR and see if that gets any more attention at some point, but it seems fruitful for now to carry on pointing to the fork while we wait to see if anything will be accepted upstream.

(I should point out that the effort by @jeremiahpslewis to update this is completely independent of my own work in getting into Yggdrasil and I wasn't aware of it until now.)

M/Mineos/build_tarballs.jl
```diff
@@ -3,19 +3,37 @@
...
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd ${WORKSPACE}/srcdir/mineos
-./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} --disable-doc
+
+# Drop docs build target
+atomic_patch -p1 ../patches/drop_docs.diff
+
+autoupdate
```
anowacki, line 21: I believe (per my comment in the issue mentioned below) that with autotools it is more usual to create the configure script before distributing the source, and that the build step (i.e., what's happening here) should not do that.  Hence my use of my own fork of the repo.

However, I may not be up to date on this.  @giordano is it customary on here to rerun the autotools commands to build the configure script?

M/Mineos/build_tarballs.jl
```diff
@@ -3,19 +3,37 @@
...
+    DirectorySource("./bundled"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd ${WORKSPACE}/srcdir/mineos
-./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} --disable-doc
+
+# Drop docs build target
+atomic_patch -p1 ../patches/drop_docs.diff
```
anowacki, line 19: Is there a reason to apply a patch rather than use the official `--disable-doc` flag to `configure`?  I'm sure there may be good reasons.

M/Mineos/build_tarballs.jl
```diff
@@ -3,19 +3,37 @@
...
+
+autoupdate
+autoreconf --install
+# Fix clang error 'error: non-void function * should return a value [-Wreturn-type]'
+if [[ "${target}" == *-freebsd* ]] || [[ "${target}" == *-apple-* ]]; then
+    export CFLAGS="-Wno-return-type"
+fi
+
+if [[ "${target}" == aarch64-apple-* ]]; then
```
anowacki, line 28: Is this not a property of the compiler rather than the architecture, or do I misunderstand?  I am aware that newer gfortran versions error by default when a scalar is passed in as the first element of a rank-1 array, but surely that is to do with the GCC version, not the architecture itself?  A different Fortran compiler on an Apple Silicon machine may behave differently, and GCC 10 on an x86 machine will throw the same error without the `-fallow-argument-mismatch` flag—as trying it on my Intel MacBook confirms.

anowacki: Digging deeper, it seems that `-fallow-argument-mismatch` has only been in gfortran since 2019, whereas the behaviour was previously implied by using the `-std=legacy` flag which still works in GCC 10.  Maybe that is a better flag to use if we know we are using gfortran, but don't want to add machinery to detect the version?  (Or perhaps we always use GCC >= 10 on Yggdrasil?)

giordano: > Is this not a property of the compiler rather than the architecture, or do I misunderstand?

You're correct, but that's the only platform where we use GCC 11, and it's way easier to match target name rather than parsing the output of `gfortran --version` and comparing a version number

M/Mineos/build_tarballs.jl
giordano: There are different opinions about checking the `configure` script in a repository: since that's an automatically generated file, most projects do _not_ check it in, but then they provide release tarballs which have been bootstrapped by the package maintainer.  However this means that if you try to do a build directly from a repository checkout you need to bootstrap yourself, which sometimes may be complicated.

jeremiahpslewis: Think in this case, it makes sense to use the fork, particularly if @anowacki feels some level of responsibility to continue maintaining the project. :)

jeremiahpslewis: (It also lets us drop the patch and use the build flag)

M/Mineos/build_tarballs.jl
jeremiahpslewis: Rearranged the comment so this is even more clear

anowacki: > that's the only platform where we use GCC 11, and it's way easier to match target name rather than parsing the output of gfortran --version and comparing a version number

Thanks for explaining that.  So I guess if and when the other architectures move on to GCC >= 10, any similar build scripts relying on `aarch64-apple` using a newer GCC will need to be rewritten.  Seems like unconditionally using `-std=legacy` might be easier, since there is no indication that that option is going away.

(`gfortran --version | awk 'NR == 1 {version = $NF; sub("\\..*", "", version); print version}'` seems easy enough.)

By the way, thank you @jeremiahpslewis for updating this!

jeremiahpslewis:
@anowacki Are we good to go or am I still missing something?

anowacki:
I don't mean to be an arbiter of anything, but it seems good to me.  Thanks for updating the recipe to work on Apple Silicon!

giordano:
Uhm, this is failing on `master`, I should have retriggered the build here before merging

PR #4403, opened 2022-02-07
build LLVM 13.0.1

vchuravy:


giordano:
`aarch64-apple-darwin` failed with
```
[18:08:01] -- Check for working C compiler: /opt/bin/aarch64-apple-darwin20-libgfortran5-cxx11/aarch64-apple-darwin20-clang - broken
[18:08:01] CMake Error at /usr/share/cmake/Modules/CMakeTestCCompiler.cmake:69 (message):
[18:08:01]   The C compiler
[18:08:01] 
[18:08:01]     "/opt/bin/aarch64-apple-darwin20-libgfortran5-cxx11/aarch64-apple-darwin20-clang"
[18:08:01] 
[18:08:01]   is not able to compile a simple test program.
[18:08:01] 
[18:08:01]   It fails with the following output:
[18:08:01] 
[18:08:01]     Change Dir: /workspace/build/CMakeFiles/CMakeTmp
[18:08:01]     
[18:08:01]     Run Build Command(s):/usr/bin/ninja cmTC_9198c && [1/2] Building C object CMakeFiles/cmTC_9198c.dir/testCCompiler.c.o
[18:08:01]     ninja: job failed: /opt/bin/aarch64-apple-darwin20-libgfortran5-cxx11/aarch64-apple-darwin20-clang    -o CMakeFiles/cmTC_9198c.dir/testCCompiler.c.o -c /workspace/build/CMakeFiles/CMakeTmp/testCCompiler.c
[18:08:01]     clang-9: error: invalid version number in '-mmacosx-version-min=11.0'
[18:08:01]     ninja: subcommand failed
[18:08:01]     
[18:08:01]     
[18:08:01] 
[18:08:01]   
[18:08:01] 
[18:08:01]   CMake will not be able to correctly generate this project.
[18:08:01] Call Stack (most recent call first):
[18:08:01]   CMakeLists.txt:49 (project)
[18:08:01] 
[18:08:01] 
[18:08:01] -- Configuring incomplete, errors occurred!
```

It may be wort pointing out that we recently did this: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/193.  Is the value of the version number set incorrectly?  Nothing else complained so far.

vchuravy:
I can try removing https://github.com/JuliaPackaging/Yggdrasil/blob/6bc48a32e53cd1ccdc0a6461be3ee2c4415a3ba6/L/LLVM/common.jl#L212-L214?

giordano:
Not sure that's the culprit (the error messages prints `11.0`), but by default we have `10.14` for x86_64, and `11.0` for aarch64, those values look odd nowadays, so probably it's better to remove them anyway

giordano:
Still same errors.  Where is `clang-9` coming from?

Judging by some comments in https://stackoverflow.com/q/63972113/2442087 I think we may have to do
```sh
export SYSTEM_VERSION_COMPAT=1
```
at least for aarch64-apple-darwin where we now use the 11 SDK?

giordano:
Quick update.  Setting `SYSTEM_VERSION_COMPAT=1` didn't help, but in order to solve the aarch64-specific issue it's sufficient to use llvm 12 at https://github.com/JuliaPackaging/Yggdrasil/blob/74d51d81e6d6c56e255701972d66354298376965/L/LLVM/LLVM_full%4013.0.1/build_tarballs.jl#L8 I gather clang 9 didn't expect new macOS versions after 10.x.

With that we have the same issue for both macOS platforms: hundreds of errors like
```
[13:20:46] CMake Error at cmake/modules/AddLLVM.cmake:536 (add_library):
[13:20:46]   Attempting to use @rpath without CMAKE_SHARED_LIBRARY_RUNTIME_C_FLAG being
[13:20:46]   set.  This could be because you are using a Mac OS X version less than 10.5
[13:20:46]   or because CMake's platform configuration is corrupt.
[13:20:46] Call Stack (most recent call first):
[13:20:46]   cmake/modules/AddLLVM.cmake:794 (llvm_add_library)
[13:20:46]   tools/llvm-shlib/CMakeLists.txt:33 (add_llvm_library)
```

It looks like after the upgrade to Alpine Linux 3.15 with the newer CMake (3.21) these lines https://github.com/JuliaPackaging/Yggdrasil/blob/4ebbdc2d5b4a55ec28124c25d763eece01c8ffb3/L/LLVM/common.jl#L208-L210 aren't doing their job anymore.

@imciner2 found a new solution: based on the fact [platform initialisation](https://gitlab.kitware.com/cmake/cmake/-/blob/f25f6d2ed9562ba4966b89d5b2e43c10a996dc5c/Modules/Platform/Darwin-Initialize.cmake#L15) for macOS requires the command `sw_vers`, he suggested to create this fake command
```sh
if [[ "${target}" == *-darwin* ]]; then
    # Create official Apple-blessed `sw_vers`, needed by cmake
    cat > /usr/bin/sw_vers << EOF
#!/bin/sh
echo "10.14.6"
EOF
    chmod +x /usr/bin/sw_vers
fi
```
modelled on https://github.com/JuliaPackaging/Yggdrasil/blob/2b0395821db522fa190e749d4d7fd670a5297643/L/licensecheck/build_tarballs.jl#L17-L24  This does the trick for me locally, I think we'll automatically generate this script in [`BinaryBuilderBase/Runner.jl`](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/e7286d0d7e0c1b7d724d173529edd12ecb411eba/src/Runner.jl) so that we can also dynamically choose the SDK version per platform.

Thanks a lot @imciner2 and @gbaraldi for help in debugging this!

giordano:
@vchuravy should we get #4279 merged before this one?

vchuravy:
The patch is part of it already, but yes we need to adjust the common file.

giordano:
@jpsamaroo this is failing to find some of the new tools we added in #4279

jpsamaroo:
@giordano are the files actually there?

vchuravy:
Yes if you look at the log, but `lld` still ended up in `bin`

giordano:
You have to use again a newer llvm at https://github.com/JuliaPackaging/Yggdrasil/blob/2eba1ace1efb91eeb13ea4e36d7131abd0b48203/L/LLVM/LLVM_full_assert%4013.0.1/build_tarballs.jl#L8

vchuravy:
Huh, I though I `git pull --rebase`...  Sorry about that

giordano:
The perils of not using `--force-with-lease` :stuck_out_tongue_winking_eye:

giordano:
Aaaand all green also here :partying_face:

PR #4399, opened 2022-02-07
[SCALE_MAMBA] Bump version

jeremiahpslewis:
Blocked by https://github.com/JuliaPackaging/Yggdrasil/pull/4405

PR #4405, opened 2022-02-07
[MPIR] Build for 1.6 compat 

jeremiahpslewis:
https://github.com/JuliaPackaging/Yggdrasil/pull/4399/files is blocked by this.

### DIFF
--- a/M/MPIR/build_tarballs.jl
+++ b/M/MPIR/build_tarballs.jl
@@ -17,9 +17,6 @@ version = v"3.0.1" # Fake version bump for compat
 script = raw"""
 cd $WORKSPACE/srcdir/mpir-*
 
-# We need `yasm`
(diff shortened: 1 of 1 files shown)


giordano:
Also, I'd use `Yasm_jll` as build dependency instead of installing it with `apk` (we can't cache the latter)

M/MPIR/build_tarballs.jl
```diff
@@ -34,7 +34,8 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[
+    Dependency("YASM_jll"),
```
giordano, line 37: This is a _**build**_ tool
```suggestion
dependencies = [
    HostBuildDependency("YASM_jll"),
```

PR #4409, opened 2022-02-07
[bottom] Add builder

giordano:
I tested locally only for `x86_64-linux-gnu`, we'll need to see if there are
problems for other platforms (it will likely fail for 32-bit Windows).


jeremiahpslewis:
Random question here, is the folder name `bottom_jll` deliberate or should it be `bottom`?

giordano:
That was wrong, good catch (but it'd be caught also by the fact the directory name must match exactly the package name)

PR #4408, opened 2022-02-07
DuckDB version 0.3.2

kimmolinna:
[0.3.2 Preview Release "Gibberifrons"](https://github.com/duckdb/duckdb/releases)

PR #4412, opened 2022-02-08
[DuckDB] Drop windows patch

jeremiahpslewis:


giordano:
I mean, the patch applied cleanly in #4408, so......

jeremiahpslewis:
patch is still necessary.

PR #4418, opened 2022-02-09
[R] try building

simonbyrne:
This is a pretty long shot, and has a lot of dependencies, but would be cool if it works.

### DIFF
--- a/R/R/build_tarballs.jl
+++ b/R/R/build_tarballs.jl
@@ -19,6 +19,7 @@ export CPPFLAGS="-I${includedir}"
     --build=${MACHTYPE} \
     --host=${target} \
     --with-x=no \
+    --enable-R-shlib=yes \
     r_cv_header_zlib_h=yes \
     r_cv_have_bzlib=yes \
     r_cv_have_lzma=yes \
@@ -37,7 +38,7 @@ platforms = supported_platforms()
 
 # The products that we will ensure are always built
 products = [
-    LibraryProduct("libR", :libR),
+    LibraryProduct("libR", :libR, "lib/R/lib"),
 ]
 
 # Dependencies that must be installed before this package can be built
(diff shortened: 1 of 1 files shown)


simonbyrne:
(this should probably have been a draft PR)

simonbyrne:
Okay, it looks like this logic doesn't allow it to find zlib if cross compiling, since it will always set `r_cv_header_zlib_h=no`?

```
if test ${r_cv_header_zlib_h+y}
then :
  printf %s "(cached) " >&6
else $as_nop
  if test "$cross_compiling" = yes
then :
  r_cv_header_zlib_h=no
else $as_nop
  cat confdefs.h - <<_ACEOF >conftest.$ac_ext
/* end confdefs.h.  */

#include <stdlib.h>
#include <string.h>
#include <zlib.h>
int main() {
#ifdef ZLIB_VERNUM
  if (ZLIB_VERNUM < 0x1250) {
    exit(1);
  }
  exit(0);
#else
  exit(1);
#endif
}

_ACEOF
if ac_fn_c_try_run "$LINENO"
then :
  r_cv_header_zlib_h=yes
else $as_nop
  r_cv_header_zlib_h=no
fi
rm -f core *.core core.conftest.* gmon.out bb.out conftest$ac_exeext \
  conftest.$ac_objext conftest.beam conftest.$ac_ext
fi

fi
{ printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: $r_cv_header_zlib_h" >&5
printf "%s\n" "$r_cv_header_zlib_h" >&6; }

  have_zlib=${r_cv_header_zlib_h}
fi
{ printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking whether zlib support suffices" >&5
printf %s "checking whether zlib support suffices... " >&6; }
if test "${have_zlib}" != yes; then
  as_fn_error $? "zlib library and headers are required" "$LINENO" 5
else
  LIBS="-lz ${LIBS}"
  { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: yes" >&5
printf "%s\n" "yes" >&6; }
fi
```

PR #4422, opened 2022-02-10
[Mineos] Change nothing and check if passes

jeremiahpslewis:


jeremiahpslewis:
@giordano I can't reproduce the error locally via Docker, running Julia 1.7.1 and BinaryBuilder 0.5.2

jeremiahpslewis:
Still can't get it to fail, even running it on a clean installed x86 linux

giordano:
What environment are you _**exactly**_ using?

jeremiahpslewis:
I've run it on a digitalocean droplet that I spun up just for the test, intel Linux with a fresh julia 1.7.1 install. Likewise I ran it on my local m1 machine via docker + rosetta julia. In each case, building off of the master branch.

giordano:
I was asking about the Julia Pkg environment.  What I'm saying is that you should use [this environment](https://github.com/JuliaPackaging/Yggdrasil/tree/dd0bfdabbaef3493841f54bce34539b270ebab8c/.ci), which is what we use here.  I can reproduce the error with that (and I also have a hot-fix ready)

jeremiahpslewis:
I tried rerunning it with --project=.ci, but I suppose that only works as intended if you run the julia build_tarballs.jl command from the root directory? in which case that would explain why I wasn't able to reproduce the error.

giordano:
Why was this closed?

jeremiahpslewis:
Reopened. Think we're good to go. :)

giordano:
There are 276 warnings like this:
```
│ 
└ @ BinaryBuilder.Auditor /depot/packages/BinaryBuilder/Ftk2f/src/auditor/compiler_abi.jl:4
```

jeremiahpslewis:
> There are 276 warnings like this:
> 
> ```
> ┌ Warning: To ensure that the correct version of libgfortran is found at runtime, add the following entry to the list of dependencies of this builder
> │ 
> │     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
> └ @ BinaryBuilder.Auditor /depot/packages/BinaryBuilder/Ftk2f/src/auditor/compiler_abi.jl:4
> ```

Thanks for the tip, have gotten a bit ‘snow blind’ to warnings given how frequently they arise even in ‘normal’ builds.

giordano:
With https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1169 I'm trying to reduce the noise, but warnings shouldn't be blindly ignored

PR #4423, opened 2022-02-10
Update to include ML-PACE package

dallasfoster:
ML-PACE allows the use of the performant atomic cluster expansion pair style.


giordano:
The failure for `aarch64-apple-darwin` is fixed by https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/218. The failure for `aarch64-linux-gnu-libgfortran4` looks hopeless: it's an internal compiler error and for libgfortran4 we have only one version of GCC.

vchuravy:
I think we can just drop `aarch64+gfortran4`.

@dallasfoster Will open a PR.

PR #4433, opened 2022-02-11
[ModularForms] new recipe

martinra:
ModularForms.jl is a new package (at https://gitlab.com/mraum/modularforms.jl) to compute with modular forms based on the Oscar CAS. These are the C dependencies that we use. The build script follows the pattern of, for example, FLINT, but the preferred GCC version was needed to complete local tests successfully.


M/ModularForms/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+sources = [GitSource("https://gitlab.com/mraum/ModularForms_jll_src.git",
+                     "4d81d41e625fde3e36dc86df5061313a8447cacd")]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/ModularForms_jll_src
+if [[ ${target} == *musl* ]]; then
+   export CFLAGS=-D_GNU_SOURCE
+fi
+./configure --prefix=$prefix --disable-static --enable-shared --with-gmp=$prefix --with-mpfr=$prefix --with-flint=$prefix --with-arb=$prefix CC=gcc ${extraflags}
```
giordano, line 16: Why forcing `gcc` as compiler?  Also, there is no `extraflags` variable

M/ModularForms/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+make -j${nproc} verbose
```
giordano, line 17: Compilation for Windows adds a bunch of references to `/workspace/destdir/lib`, all of which are wrong since the libraries are in `/workspace/destdir/bin`

M/ModularForms/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="Arb_jll"), compat = "~200.2200.0"),
+    Dependency(PackageSpec(name="FLINT_jll"), compat = "^200.800.401")
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="^1.6", preferred_gcc_version=v"11")
```
giordano, line 38: So you said that compilation of a C code wouldn't succeed with versions of GCC before 11?

M/ModularForms/build_tarballs.jl
martinra: These two are mistakes, that I'll fix. Thank you!

M/ModularForms/build_tarballs.jl
martinra: We do not have the possibility to test on anything but Linux. I don't understand the Windows ecosystem; So would it be acceptable to just leave Linux in the platforms?

M/ModularForms/build_tarballs.jl
martinra: It's the only one we work and test with. One needs C18, that's probably supported since earlier. I suppose this should then be as low as possible? I believe gcc9 already had full C18 support.

M/ModularForms/build_tarballs.jl
giordano: We prefer to build for as many platforms as possible, but the final decision of which platforms to support is yours.  But I also already  told you how to fix the failure for the Windows build: the libraries are in `${prefix}/bin`, not `${prefix}/lib`.  Or just always use `${libdir}` from our environment which always points to the right directory.

M/ModularForms/build_tarballs.jl
giordano: For maximum compatibility, we should use the oldest version of GCC possible, yes.

PR #4435, opened 2022-02-13
[QPALM] Version 1.0.0

tttapa:
https://github.com/kul-optec/QPALM/releases/tag/1.0.0

QPALM was originally maintained by Ben Hermans at https://github.com/Benny44/QPALM_vLADEL, but will now be maintained in the kul-optec organization by myself and other members of the research group.

### DIFF
--- a/Q/QPALM/build_tarballs.jl
+++ b/Q/QPALM/build_tarballs.jl
@@ -38,12 +38,12 @@ install_license LICENSE.AMD
 platforms = supported_platforms()
 
 # The products that we will ensure are always built
-products = Product[
+products = [
     LibraryProduct("libqpalm_jll", :libqpalm_jll),
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = []
+dependencies = Dependency[]
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"10.2.0")
--- a/Q/QPALM/build_tarballs.jl
+++ b/Q/QPALM/build_tarballs.jl
@@ -7,7 +7,7 @@ version = v"1.0.0"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/kul-optec/QPALM.git","f768d45e3911402513faa424bca72124dd4002d1"),
+    GitSource("https://github.com/kul-optec/QPALM.git","9bd94ef60ca5d82cd8d5f05ac7efd5a449e501c0"),
     GitSource("https://github.com/kul-optec/LADEL.git", "70ad69a621756e72b164dee302a826f9cc111c55"),
 ]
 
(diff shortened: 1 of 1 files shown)


Q/QPALM/build_tarballs.jl
```diff
@@ -65,15 +39,11 @@ platforms = supported_platforms()
 
 # The products that we will ensure are always built
 products = Product[
```
giordano, line 41:
```suggestion
products = [
```

Q/QPALM/build_tarballs.jl
```diff
@@ -65,15 +39,11 @@ platforms = supported_platforms()
...
-                   LibraryProduct("libladel", :libladel),
-                   LibraryProduct("libqpalm", :libqpalm),
+    LibraryProduct("libqpalm_jll", :libqpalm_jll),
```
giordano, line 42: Interesting file name!  I gather this is like a Julia-specific library?

Q/QPALM/build_tarballs.jl
```diff
@@ -65,15 +39,11 @@ platforms = supported_platforms()
...
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = [
-    Dependency(PackageSpec(name="OpenBLAS32_jll", uuid="656ef2d0-ae68-5445-9ca0-591084a874a2"))
-    Dependency(PackageSpec(name="SuiteSparse_jll", uuid="bea87d4a-7f5b-5778-9afe-8cc45184846c"))
-]
+dependencies = []
```
giordano, line 46:
```suggestion
dependencies = Dependency[]
```

Q/QPALM/build_tarballs.jl
```diff
@@ -65,15 +39,11 @@ platforms = supported_platforms()
...
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.3")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"10.2.0")
```
giordano, line 49: Why GCC 10?  For maximum compatibility we prefer to use the _oldest_ GCC version which is able to compile the code.

giordano (commented):
Thanks for your contribution!

Q/QPALM/build_tarballs.jl
tttapa: Indeed, it includes functions that are only used in the Julia interface. Is this okay, or would it be better to have a Julia-specific library that then depends on a shared `libqpalm` library?

giordano: We usually tend to have generic libraries, but if you're the maintainer of both the upstream package and the julia interface, I guess the decision is up to you :slightly_smiling_face:

Q/QPALM/build_tarballs.jl
tttapa: Good point, I just selected that because it was the version I developed with. I'll remove the explicit version, it should build without issues on GCC 4.8.5.

PR #4440, opened 2022-02-14
BLIS Support Aarch64-mac Platform, Version Push to HEAD v0.8.1+α

xrq-phys:
To resolve JuliaLinearAlgebra/BLIS.jl#8 .

This upgrade from `v0.8.0` to `v0.8.1` should be trivial. In fact, the upstream seems to be in a course of changing its versioning system.

Let me note here some non-trivial functionalities on current `master`:
- Compilation on macOS Aarch64. Now the aarch64 kernels should support both GCC and `clang`. (& Would it be OK to introduce xrq-phys/blis_apple@8912315 kernels)
- SVE kernels for Neoverse V1, etc. SVE kernels require GCC>=8 for basic support and GCC>=10 to fully compile.

### DIFF
--- a/B/blis/build_tarballs.jl
+++ b/B/blis/build_tarballs.jl
@@ -7,7 +7,7 @@ version = v"0.8.1"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/flame/blis.git", "8535b3e11d2297854991c4272932ce4974dda629"),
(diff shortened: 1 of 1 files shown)


xrq-phys:
Oh! Thank you very much for the info!

Perhaps it'd be good to at least kick off aarch64-mac / SVE with the current master + GCC 8.0.

PR #4441, opened 2022-02-14
[ModularForms] new recipe

martinra:
Same as #4433, but GitHub closed the merge request after I force pushed the rebased commits.

ModularForms.jl is a new package (at https://gitlab.com/mraum/modularforms.jl) to compute with modular forms based on the Oscar CAS. These are the C dependencies that we use. The build script follows the pattern of, for example, FLINT, but the preferred GCC version was needed to complete local tests successfully.

As for the comment about Windows libs in the review, I only later realized that this would require ModularForms_jll_src to be adjusted, where we do not accept special casing for proprietary platforms. I narrowed down the target platforms accordingly.

### DIFF
--- a/M/ModularForms/build_tarballs.jl
+++ b/M/ModularForms/build_tarballs.jl
@@ -13,7 +13,7 @@ cd $WORKSPACE/srcdir/ModularForms_jll_src
 if [[ ${target} == *musl* ]]; then
    export CFLAGS=-D_GNU_SOURCE
 fi
-./configure --prefix=$prefix --disable-static --enable-shared --with-gmp=$prefix --with-mpfr=$prefix --with-flint=$prefix --with-arb=$prefix ${extraflags}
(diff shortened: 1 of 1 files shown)


M/ModularForms/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+make install LIBDIR=$(basename ${libdir})
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [p for p in supported_platforms()
+             if contains(triplet(p), "linux")
+             || contains(triplet(p), "apple")
+             || contains(triplet(p), "freebsd")
+            ]
```
giordano, line 27:
```suggestion
platforms = supported_platforms(; exclude=Sys.iswindows)
```

giordano (commented):
> but GitHub closed the merge request after I force pushed the rebased commits.

Yes, it isn't a good idea to open a PR from a branch with the same name as the target branch, I believe GitHub shows a big scary red warning box when you do that.

giordano:
Thanks for your contribution!

PR #4442, opened 2022-02-14
New Recipe: ColPack v0.1.0

michel2323:
This is the binary for ColPack forked from https://github.com/CSCsw/ColPack . It should enable advanced coloring algorithms that can be used in SparseDiffTools.jl . A C interface was added and a ColPack.jl package will be published soon.

### DIFF
--- a/C/ColPack/build_tarballs.jl
+++ b/C/ColPack/build_tarballs.jl
@@ -1,0 +1,41 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "ColPack"
+version = v"0.1.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/michel2323/ColPack.git", "c8674588c8204e9f05a2c6adea21a28f7b56f490")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd ColPack/build/automake/
+autoreconf -vif
+mkdir build
+cd build/
+../configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; exclude=p -> Sys.isapple(p) || Sys.iswindows(p))
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libColPack", :libcolpack)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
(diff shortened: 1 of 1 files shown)


C/ColPack/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libColPack", :libcolpack)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
```
giordano, line 49:
```
┌ Warning: To ensure that the correct version of libgomp is found at runtime, add the following entry to the list of dependencies of this builder
│ 
└
```

C/ColPack/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "freebsd"; )
+]
+
```
giordano, line 41:
```
┌ Warning: /agent/_work/1/s/C/ColPack/build/i686-linux-gnu/9PMeRj0H/i686-linux-gnu-libgfortran3-cxx03/destdir/examples/Basic/Generate_seed_matrix_for_Jacobian contains std::string values!  This causes incompatibilities across the GCC 4/5 version boundary.  To remedy this, you must build a tarball for both GCC 4 and GCC 5.  To do this, immediately after your `platforms` definition in your `build_tarballs.jl` file, add the line:
│ 
└ @ BinaryBuilder.Auditor /depot/packages/BinaryBuilder/IIDDS/src/auditor/compiler_abi.jl:247
 ```

File: C/ColPack/build_tarballs.jl [hunk 3]
```diff
@@ -0,0 +1,53 @@
...
+mkdir build
+cd build/
+../configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
```
giordano, line 27: This should be simply
```julia
platforms = supported_platforms(; exclude=p -> Sys.isappl(p) || Sys.iswindows(p))
```
What's wrong with those platforms?

michel2323: OpenMP. I think it might be related to what you outlined above. Thanks for highlighting that. I'll look into it.

michel2323: I tried to fix OpenMP for Apple but I can't figure it out. It cannot find 'omp.h'. Forcing `gcc` lead to linking errors. I see that other projects have also issues with Apple and OpenMP. On Windows it fails with some linking errors.

PR #4450, opened 2022-02-16
[ParseGen] Correctly set `MACOSX_DEPLOYMENT_TARGET`

jeremiahpslewis:


P/ParseGen/bundled/sw_vers
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
giordano, line 1: Isn't it easier to just set MACOSX_DEPLOYMENT_TARGET?

P/ParseGen/build_tarballs.jl
```diff
@@ -9,19 +9,19 @@ version = v"2.0.0"
...
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/parsegen-cpp*/
 
 install_license LICENSE
-
 mkdir build && cd build
 
 if [[ "${target}" == x86_64-apple-darwin* ]]; then
     export CXXFLAGS="-mmacosx-version-min=10.15"
```
giordano, line 23: You likely don't need this line. And the entire aarch64-apple-darwin branch is now useless

giordano: This isn't resolved. I'm still convinced this line isn't needed.

giordano (changes requested):
I'm not sure what's exactly the point of this pull request which doesn't change anything in the resulting tarballs nor the generated JLL package.

If you just want to make it work with newer cmake, which is fine, please use `[skip ci]` because I don't want this to trigger a rebuild of anything after merge.

P/ParseGen/build_tarballs.jl
```diff
@@ -9,28 +9,25 @@ version = v"2.0.0"
...
```
jeremiahpslewis, line 23:
```suggestion
```

P/ParseGen/build_tarballs.jl
jeremiahpslewis: Thanks, didn't see the line you were referring to the first time.

jeremiahpslewis:
> I'm not sure what's exactly the point of this pull request which doesn't change anything in the resulting tarballs nor the generated JLL package.
> 
> If you just want to make it work with newer cmake, which is fine, please use `[skip ci]` because I don't want this to trigger a rebuild of anything after merge.

I was looking for a clean working example of a Mac x86 10.14 build to debug a regression which appeared with BinaryBuilderBase 1.6.0 so I could figure out another binary.

jeremiahpslewis:
Added [skip ci]

PR #4454, opened 2022-02-16
[ngspice] update to v36

sjkelly:
This is likely not good since I disabled some cross compile patches for windows.


sjkelly:
@giordano I think this is ready :)

PR #4456, opened 2022-02-16
[HiGHS] create a static build of the executable

odow:
We're getting a few people using the Yggdrasil builds of HiGHS outside of Julia(e.g., https://github.com/ERGO-Code/HiGHS/issues/590#issuecomment-987670210). 

The biggest problem is the need to manually download some of the CompilerSupportLibraries on Windows.

This PR adds a second build to the HiGHS script to produce `highs.exe` and `highs_static.exe`. 

That hopefully lets us use `highs.exe` and `highs.dll` through HiGHS.jl, and if users just want the static executable, they can use `highs_static.exe`. 

I don't have a windows machine, but @jajhall has tested `highs_static.exe` that I compiled locally and it seems to work.

### DIFF
--- a/H/HiGHS/build_tarballs.jl
+++ b/H/HiGHS/build_tarballs.jl
@@ -26,6 +26,41 @@ if [[ "${target}" == *86*-linux-musl* ]]; then
 fi
 mkdir -p HiGHS/build
 cd HiGHS/build
+
+# To help upstream distribute binaries, we compile HiGHS by statically linking
+# dependencies _and_ by dynamically linking dependences.
+#
+# The dynamically linked files, `highs` and `libhighs` are used by HiGHS.jl, and
+# the statically linked files, `highs_statc` and `libhighs_static` are used by
+# people downloading the releases in other languages.
+
+# First, build the static executable
+
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBUILD_SHARED_LIBS=OFF \
+    -DFAST_BUILD=ON \
+    -DJULIA=ON \
+    -DIPX=ON ..
+
+if [[ "${target}" == *-linux-* ]]; then
+        make -j ${nproc}
+else
+    if [[ "${target}" == *-mingw* ]]; then
+        cmake --build . --config Release
+    else
+        cmake --build . --config Release --parallel
+    fi
+fi
+make install
+
+# The file /bin/highs.${dlext} is statically linked. Rename it so that it is not
+# replaced when we statically link things.
+mv ${bindir}/highs${exeext} ${bindir}/highs_static${exeext}
+
+# Second, build the dynamic libs for Julia
+
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
@@ -48,12 +83,13 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(supported_platforms(;experimental=true))
+platforms = expand_cxxstring_abis(supported_platforms())
 
 # The products that we will ensure are always built
 products = [
     LibraryProduct("libhighs", :libhighs),
     ExecutableProduct("highs", :highs),
+    ExecutableProduct("highs_static", :highs_static),
 ]
 
 # Dependencies that must be installed before this package can be built


giordano:
Uhm, how about having a separate package for the static build which doesn't have any runtime dependencies? :grimacing:  On the Julia side we don't do much with the static libraries as we can't ccall them, and users always complain about the size of `.julia/artifacts`.  With this change the tarball for x86-64 Linux blows up from 1.8 MiB to 5.5.

giordano:
Also, this looks a good use case for https://github.com/fredrikekre/ygg :slightly_smiling_face:  CC: @fredrikekre

odow:
Okay. A separate package will work. 

We really are missing a way to collate the dependencies of a package into a single folder without needing a Julia installation. I guess resolving version compatibility is the hard part.

giordano:
> We really are missing a way to collate the dependencies of a package into a single folder without needing a Julia installation. I guess resolving version compatibility is the hard part.

Yes, I think Julia in https://github.com/fredrikekre/ygg is used mainly to resolve the environment, because it needs `Pkg.jl`, not to actually use the programs.

H/HiGHS/highs_common.jl
```diff
@@ -11,11 +11,15 @@ sources = [
...
+    DirectorySource(joinpath(@__DIR__, "bundled")),
 ]
 
-# Bash recipe for building across all platforms
-script = raw"""
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = expand_cxxstring_abis(supported_platforms())
+
+function build_script(; shared_libs::String)
```
giordano, line 21: I like this doesn't have a default value, so we're sure we specify it!

H/HiGHS/highs_common.jl
```diff
@@ -11,11 +11,15 @@ sources = [
...
+    return raw"""
```
giordano, line 22: I usually do something like in this cases to avoid interpolating variables in the middle of the script
```suggestion
    return "BUILD_SHARED=$(shared_libs)\n" * raw"""
```

H/HiGHS/highs_common.jl
```diff
@@ -26,10 +30,11 @@ if [[ "${target}" == *86*-linux-musl* ]]; then
 fi
 mkdir -p HiGHS/build
 cd HiGHS/build
+
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
-    -DBUILD_SHARED_LIBS=ON \
+    -DBUILD_SHARED_LIBS=""" * shared_libs * raw""" \
```
giordano, line 37:
```suggestion
    -DBUILD_SHARED_LIBS=${BUILD_SHARED} \
```

giordano:
Do you need the static library in `HiGHSstatic`?  Deleting that one can save some space.

odow:
> Do you need the static library in HiGHSstatic?

I'm not sure. There's an open issue, https://github.com/ERGO-Code/HiGHS/issues/590, but I expect most people would just want the executable.

Perhaps we can delete it for now, and then if there is a request we can add it back in future.

H/HiGHS/highs_common.jl
```diff
@@ -44,33 +49,9 @@ else
     fi
 fi
 make install
+if [[ "${BUILD_SHARED}" == "OFF"]]
+    # Delete the static library to save space
+    rm ${destdir}/lib/libhighs*
```
giordano, line 54: You'd just have an empty directory as far as I can see:
```suggestion
    rm -r ${prefix}/lib
```

H/HiGHS/highs_common.jl
```diff
@@ -44,33 +49,9 @@ else
     fi
 fi
 make install
+if [[ "${BUILD_SHARED}" == "OFF"]]
```
giordano, line 52:
```suggestion
if [[ "${BUILD_SHARED}" == "OFF" ]]
```

H/HiGHS/highs_common.jl
```diff
@@ -44,33 +49,9 @@ else
     fi
 fi
 make install
+if [[ "${BUILD_SHARED}" == "OFF" ]]
```
giordano, line 52:
```suggestion
if [[ "${BUILD_SHARED}" == "OFF" ]]; then
```

giordano:
Note that the resulting executable still links to `libstdc++`:
```console
sandbox:${WORKSPACE} # ${target}-objdump -p bin/highs.exe | grep DLL
        DLL Name: libgcc_s_seh-1.dll
        DLL Name: libstdc++-6.dll
        DLL Name: KERNEL32.dll
        DLL Name: msvcrt.dll
        DLL Name: libwinpthread-1.dll
```
Is this what you expected?

H/HiGHS/HiGHSstatic/build_tarballs.jl
```diff
@@ -0,0 +1,25 @@
...
+include("../highs_common.jl")
+
+script = build_script(shared_libs = "OFF")
+
+products = [
+    ExecutableProduct("highs", :highs),
+]
+
+dependencies = [
+    BuildDependency("CompilerSupportLibraries_jll"),
```
giordano, line 11: [This](https://github.com/JuliaPackaging/Yggdrasil/pull/4456#issuecomment-1042483581) also means that this is still a runtime dependency.  I'm not sure this has achieved what you were hoping :confused:

odow: No, it didn't :cry:

What I want is a single executable that someone copy onto a Windows machine, assuming they have probably never programmed before and have no compilers or support libraries installed.

The main problem is that I don't have a windows machine to test, and @jajhall has one, but his test probably worked because he had some things already installed.

giordano: There is the possibility we can do a fully static build by using static libraries of GCC (might need some manual trickeries), but also note that the binary distribution will basically be GPL-licensed.

odow: I'm not sure we want to got the GPL route. 

An easier option might be that we just manually curate a download for windows users and host that at HiGHS. I think even the current workflow of getting Windows users to download a file called `HiGHS.v1.2.0.x86_64-w64-mingw32-cxx11.tar.gz` is probably a little intimidating.

Let's leave this for now, and I'll have a chat with @jajhall about our next steps.

giordano: For the record, it's sufficient to do
```diff
diff --git a/H/HiGHS/highs_common.jl b/H/HiGHS/highs_common.jl
index 8e6fd4e50..ca0e8d734 100644
--- a/H/HiGHS/highs_common.jl
+++ b/H/HiGHS/highs_common.jl
@@ -31,6 +31,9 @@ fi
 mkdir -p HiGHS/build
 cd HiGHS/build
 
+if [[ "${BUILD_SHARED}" == "OFF" ]]; then
+    export CXXFLAGS="-static"
+fi
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
```
to get a truly static library (in terms of GCC dependencies):
```console
sandbox:${WORKSPACE}/srcdir/HiGHS/build # ${target}-objdump -p ${bindir}/highs.exe | grep DLL
        DLL Name: KERNEL32.dll
        DLL Name: msvcrt.dll
sandbox:${WORKSPACE}/srcdir/HiGHS/build # ls -lh ${bindir}/highs.exe
-rwxr-xr-x    1 root     root       16.4M Feb 17 02:07 /workspace/destdir/bin/highs.exe
```
The cost of having a 16 MiB executable.  And again, you're embedding GCC.  However, being a standalone executable, the GPL license may not actually be a big deal: you don't need to link to anything else, so it doesn't "spread".  Of course IANAL, blah blah.

giordano: I think a benefit of doing the static build is that you don't need to expand the C++ string ABI (that's needed to match user's `libstdc++`), so `HiGHS.v1.2.0.x86_64-w64-mingw32-cxx11.tar.gz` will simplify to `HiGHS.v1.2.0.x86_64-w64-mingw32.tar.gz` :slightly_smiling_face:

H/HiGHS/highs_common.jl
```diff
@@ -11,11 +11,15 @@ sources = [
...
-# Bash recipe for building across all platforms
-script = raw"""
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = expand_cxxstring_abis(supported_platforms())
+
+function build_script(; shared_libs::String)
+    return "BUILD_SHARED=$(shared_libs)\n" * raw"""
 cd $WORKSPACE/srcdir
 if [[ "${target}" == *86*-linux-musl* ]]; then
```
giordano, line 24: Side note, since you use GCC 6, since shouldn't be necessary.

H/HiGHS/HiGHSstatic/build_tarballs.jl
odow: Okay, nice. Perhaps providing a GPL binary for general users would be okay, and their commercial customers can pay for support if they don't want to use it.

I'll let @jajhall and @lgottwald make the decision since it's not really up to me.

H/HiGHS/highs_common.jl
odow: You're a continual fount of wisdom.

H/HiGHS/HiGHSstatic/build_tarballs.jl
jajhall: > Perhaps providing a GPL binary for general users would be okay, and their commercial customers can pay for support if they don't want to use it.

This seems reasonable to me. After all, it's far from being the only way to use HiGHS, even via binaries. It's just an option.

I've gone back to the static binary that @odow produced, and it only ran because I put it in the same folder as the support binaries. Removing them, and even adding the two demanded by the pop-up gives the same behaviour as when I first used the shared binary.

Thanks for your efforts @odow and @giordano

giordano: @jajhall can you please try https://julia-bb-buildcache.s3.amazonaws.com/69df64b00a9ef2a2a4492e0832af484997dfeb151f8e1a9cb2f525528950c9d1/648abe945f5b43151bd5a1c1fb098ccb13c71bf620c8d675ef95128087800780/x86_64-w64-mingw32-cxx11.tar.gz?

jajhall:
> @@ -0,0 +1,25 @@
+
+include("../highs_common.jl")
+
+script = build_script(shared_libs = "OFF")
+
+products = [
+    ExecutableProduct("highs", :highs),
+]
+
+dependencies = [
+    BuildDependency("CompilerSupportLibraries_jll"),

@jajhall can you please try https://julia-bb-buildcache.s3.amazonaws.com/69df64b00a9ef2a2a4492e0832af484997dfeb151f8e1a9cb2f525528950c9d1/648abe945f5b43151bd5a1c1fb098ccb13c71bf620c8d675ef95128087800780/x86_64-w64-mingw32-cxx11.tar.gz?

I still get pop-ups like this

image

giordano:
Yeah, sorry, I deleted the message because I realised I didn't push the change yet 🙂 I'll let you when it's ready

giordano:
@jajhall can you please now try https://julia-bb-buildcache.s3.amazonaws.com/69df64b00a9ef2a2a4492e0832af484997dfeb151f8e1a9cb2f525528950c9d1/33c871a6d81b95dc34f4f2a72111e44afa5e72ed92e3ea50e85d02f78ffe9dbe/x86_64-w64-mingw32.tar.gz?  This should work out of the box for you.

However we have a problem: we can't do the _fully_ static build for Unix systems, because we have the static gcc libraries only for Windows.  Would it be a problem to do the fully static build only for Windows?  My understanding is that's the only system where you generally have troubles, isn't it?

giordano:
Ok, this should be good to go.  I'll let @odow have a look.  In another PR I'll restore the recipe for our "standard" build of HiGHS (the reason why I removed that one from here is because I didn't want to actually rebuild that one too).

H/HiGHS/highs_common.jl
```diff
@@ -0,0 +1,61 @@
...
+make install
+
+install_license ../LICENSE
+
+if [[ "${BUILD_SHARED}" == "OFF" ]]; then
+    # Delete the static library to save space
+    rm -r ${prefix}/lib
+    if [[ "${target}" == *-mingw* ]]; then
+        # The Windows build ships also GCC runtime, add its license as well
+        install_license install_license /usr/share/licenses/GPL3
```
odow, line 57:
```suggestion
        install_license /usr/share/licenses/GPL3
```

odow (commented):
Other than the one query on the `install_license`, this LGTM.

(I can add back the other script, I guess with `[ci skip]` to avoid rebuilding it?)

H/HiGHS/highs_common.jl
giordano: Ugh, good catch

jajhall:
I've been exchanging emails with a user who has produced a static build for Linux, observing 

> I had to recompile HiGHS with cmake .. -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DSHARED=OFF
As explained here: https://stackoverflow.com/questions/38296756/what-is-the-idiomatic-way-in-cmake-to-add-the-fpic-compiler-option

As I don't understand the issues, I don't know if this is of any value to you!

giordano:
The problem we had here is simply that we don't have the static libraries of GCC for Linux, so couldn't do the fully static build.

PR #4460, opened 2022-02-17
[HiGHS] restore build_tarballs.jl

odow:
This PR restores the build_tarballs.jl script for the dynamic build
of HiGHS. It was removed in #4456 to avoid rebuilding the binary.

[ci skip]


odow:
Hmm. Looks like `[ci skip]` didn't work.

giordano:
Yeah, it doesn't work on pull requests, only when pushing to master. It's OK

PR #4462, opened 2022-02-18
XyceWrapper: support more julia versions

sjkelly:
I am not 100% of the mechanics here. Let's give this a shot :)

### DIFF
--- a/X/XyceWrapper/build_tarballs.jl
+++ b/X/XyceWrapper/build_tarballs.jl
@@ -2,7 +2,7 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
-julia_versions = [v"1.6", v"1.7"]
+julia_versions = [v"1.6.3", v"1.7"]
 
 name = "XyceWrapper"
 version = v"0.2.0"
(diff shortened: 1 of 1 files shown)


X/XyceWrapper/build_tarballs.jl
```diff
@@ -40,8 +47,8 @@ products = [
 dependencies = [
     Dependency("Xyce_jll"; compat="^7.4.0"),
     Dependency("libcxxwrap_julia_jll"),
-    BuildDependency(PackageSpec(name="libjulia_jll", version=julia_version)),
+    BuildDependency("libjulia_jll"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"7")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat = "1.6", preferred_gcc_version=v"7")
```
giordano, line 54: I just realised this required Julia v1.6 without changing the version number, which is a violation of second point of #2763.  And this is why https://github.com/JuliaRegistries/General/pull/54942 is stalling.

sjkelly: Oh that is easy enough fix :). Thanks!

PR #4464, opened 2022-02-19
Include Arm SVE in Config arm64

xrq-phys:
Both SVE configs are added:
- `armsve` the generic one.
- `a64fx` the A64FX one.

Runtime detection supported by upstream.

Select to bypass `-march` limitation since we're building a fat-lib.

Sorry for having opened yet another issue :D

### DIFF
--- a/B/blis/build_tarballs.jl
+++ b/B/blis/build_tarballs.jl
@@ -18,8 +18,9 @@ cd blis/
 
 for i in ./config/*/*.mk; do
 
-    # Building in container forbids -march options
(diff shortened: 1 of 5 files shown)


giordano:
With this build, on an A64FX cluster I get:
```julia
julia> using LinearAlgebra

julia> peakflops()
2.318874193303886e10

julia> using FujitsuBLAS

julia> peakflops()
4.414499484255405e10

julia> BLAS.lbt_forward("./libblis.so"; clear=true)
155

julia> peakflops()
4.89817830717874e10
```
Nice!

PR #4470, opened 2022-02-20
libcxxwrap_julia: add Julia 1.9 support

fingolfin:
Ping @barche: I hope this is OK for you?

### DIFF
--- a/L/libcxxwrap_julia/build_tarballs.jl
+++ b/L/libcxxwrap_julia/build_tarballs.jl
@@ -2,10 +2,15 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
+# See https://github.com/JuliaLang/Pkg.jl/issues/2942
+# Once this Pkg issue is resolved, this must be removed
+uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
+delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
+
 name = "libcxxwrap_julia"
 version = v"0.9.0"
 
-julia_versions = [v"1.6.3", v"1.7.0", v"1.8.0"]
+julia_versions = [v"1.6.3", v"1.7.0", v"1.8.0", v"1.9.0"]
 
 is_yggdrasil = haskey(ENV, "BUILD_BUILDNUMBER")
 git_repo = is_yggdrasil ? "https://github.com/JuliaInterop/libcxxwrap-julia.git" : joinpath(ENV["HOME"], "src/julia/libcxxwrap-julia/")


fingolfin:
@giordano building for Julia 1.6 still fails with:
```
[ Info: Checking libcxxwrap-julia.git-3ce1ee40ef0f8091de1f911113039ba790a09088040da514b8f32d55b17e27fa out to libcxxwrap-julia...
    Updating registry at `/depot/registries/General.toml`
   Resolving package versions...
ERROR: LoadError: KeyError: key UUID("a83860b7-747b-57cf-bf1f-3e79990d037f") not found
```
Huh? It looks to me as if it chokes on the fact that `L/libjulia_jll/Deps.toml` in the registry still references that UUID, *even though it only applies to an older version of `libjulia_jll`**, to wit:
```
["1.5.3-1.6"]
LibOSXUnwind_jll = "a83860b7-747b-57cf-bf1f-3e79990d037f"
```
yet the current version is 1.8

Do we perhaps have to purge this reference from the registry?

fingolfin:
... Makes me also wonder if we should just "clone" `LibOSXUnwind_jll` with a new UUID, then register that, and change the offending entry I just mentioned to use that new UUID... 😈

fingolfin:
For now I've applied the "usual" workaround:
```
# See https://github.com/JuliaLang/Pkg.jl/issues/2942
# Once this Pkg issue is resolved, this must be removed
```
Let's see if it helps

barche:
Yes, this looks good to me and reminds me I forgot to update https://github.com/JuliaLang/Pkg.jl/pull/2963

L/libcxxwrap_julia/build_tarballs.jl
```diff
@@ -2,10 +2,15 @@
...
 using BinaryBuilder, Pkg
 
+# See https://github.com/JuliaLang/Pkg.jl/issues/2942
+# Once this Pkg issue is resolved, this must be removed
+uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
+delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
+
 name = "libcxxwrap_julia"
-version = v"0.9.0"
+version = v"0.9.1"
```
giordano, line 11: The source is still at 0.9.0 though

barche: Ah, yes, good catch. Maybe a rebuild would suffice? I prefer keeping source and JLL version the same from now on.

fingolfin: Rebuild with different deps won't work, though I am not sure that's the case here... can't you just add a 0.9.1 tag in your repo? It can point to the same commit...

barche: The deps didn't change, unless I'm missing something?

fingolfin: OK I am dropping the version then... hopefully it'll work

PR #4471, opened 2022-02-21
add xrootdgo

Moelf:
still need to add more functions and test the internal API a bit.

But I want to put here to see how the compilation goes and also how to source `google/uuid`...

### DIFF
--- a/X/xrootdgo/build_tarballs.jl
+++ b/X/xrootdgo/build_tarballs.jl
@@ -43,4 +43,5 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; compilers=[:c, :go], preferred_gcc_version=v"6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               compilers=[:c, :go], julia_compat="1.6", preferred_gcc_version=v"6")
--- a/X/xrootdgo/build_tarballs.jl
+++ b/X/xrootdgo/build_tarballs.jl
@@ -14,15 +14,6 @@ script = raw"""
 install_license ${WORKSPACE}/srcdir/hep/LICENSE
 cd $WORKSPACE/srcdir/hep/
 
-if [[ "${target}" == *-darwin* ]]; then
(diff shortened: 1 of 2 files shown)


X/xrootdgo/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+
+if [[ "${target}" == *-darwin* ]]; then
+    # Create official Apple-blessed `xcrun`, needed by the CGO linker
+    cat > /usr/bin/xcrun << EOF
+#!/bin/sh
+exec "\${@}"
+EOF
+    chmod +x /usr/bin/xcrun
+fi
+
```
giordano, line 25: I don't think this is needed anymore, we do it directly in BBB: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/b688ad040ee699c6f79fcf1056fe35f28a47ef04/src/Runner.jl#L837-L845

X/xrootdgo/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+products = [
+    LibraryProduct("xrootdgo", :xrootdgo),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; compilers=[:c, :go], preferred_gcc_version=v"6")
```
giordano, line 46:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
               compilers=[:c, :go], julia_compat="1.6", preferred_gcc_version=v"6")
```

giordano:
I know nothing of Go :slightly_smiling_face:

Moelf:
it's fine, as long as CI is working, I can always debug with CI

Moelf:
is there way to force a newer Go compiler? I need 1.17, looks like we have 1.16.3 by default

giordano:
No, for Rust and Go we provide a single version at the moment.  We'll need to update the Go toolchain.

Moelf:
can I expect that in order of 10 days? or I should find a work round for now

PR #4153, opened 2022-01-02
add CURL recipe from `LibCURL`

ericphanson:
closes #2720

PR #4472, opened 2022-02-21
Use a common build script for both CURL and LibCURL

### LINKED ISSUE #2720: Builder Request: a `Curl_jll.jl` package that includes the `curl` binary executable
From https://github.com/JuliaLang/julia/issues/40041#issuecomment-799577812:
> Perhaps we should create a separate CURL_jll that provides the executable, for users that want to be able to use it.

Since the Lib_CURL_jll stdlib in Julia 1.6 won't have the `curl` executable, it would be great to have a `CURL_jll` package for users that need the `curl` executable.

cc: @staticfloat @giordano

DilumAluthge:
Fixes #2720 
Closes #4153


DilumAluthge:
cc: @ericphanson @giordano

L/LibCURL/CURL/build_tarballs.jl
```diff
@@ -0,0 +1,3 @@
+include("../common.jl")
+
+build_harfbuzz(ARGS, "CURL")
```
ericphanson, line 3:
```suggestion
build_libcurl(ARGS, "CURL")
```

L/LibCURL/LibCURL/build_tarballs.jl
```diff
@@ -0,0 +1,3 @@
+include("../common.jl")
+
+build_harfbuzz(ARGS, "LibCURL")
```
ericphanson, line 3:
```suggestion
build_libcurl(ARGS, "LibCURL")
```

L/LibCURL/CURL/build_tarballs.jl
DilumAluthge: 🤦

L/LibCURL/LibCURL/build_tarballs.jl
DilumAluthge: 🤦

L/LibCURL/common.jl
```diff
@@ -0,0 +1,107 @@
...
+
+    ./configure --prefix=$prefix --host=$target --build=${MACHTYPE} "${FLAGS[@]}"
+    make -j${nproc}
+    make install
+    install_license COPYING
+    """
+
+    # These are the platforms we will build for by default, unless further
+    # platforms are passed in on the command line
+    platforms = supported_platforms(;experimental=true)
```
giordano, line 73:
```suggestion
    platforms = supported_platforms()
```

L/LibCURL/common.jl
```diff
@@ -0,0 +1,107 @@
...
+        products = [
+            ExecutableProduct("curl", :curl),
+        ]
+    else
+        # The LibCURL_jll *package* provides both the library and the executable.
+        # However, the LibCURL_jll *stdlib* only provides the library.
+        products = [
+            LibraryProduct("libcurl", :libcurl),
+            ExecutableProduct("curl", :curl),
+        ]
```
giordano, line 87: Since the stdlib exposes only libcurl, I'm not sure this should expose both of them, it'll only confuse things.

L/LibCURL/common.jl
```diff
@@ -0,0 +1,107 @@
...
+    end
+
+    # Dependencies that must be installed before this package can be built
+    dependencies = [
+        Dependency("LibSSH2_jll"),
+        Dependency("Zlib_jll"),
+        Dependency("nghttp2_jll"),
+        # Note that while we unconditionally list MbedTLS as a dependency,
+        # we default to schannel/SecureTransport on Windows/MacOS.
+        Dependency("MbedTLS_jll"; compat="~2.28.0"),
```
giordano, line 97:
```suggestion
        Dependency("MbedTLS_jll"; compat="~2.28.0", platforms=filter(p->Sys.islinux(p) || Sys.isfreebsd(p), platforms)),
```

L/LibCURL/common.jl
```diff
@@ -0,0 +1,107 @@
...
+    ]
+
+    if this_is_curl_jll
+        # Curl_jll depends on LibCURL_jll
+        push!(dependencies, Dependency("LibCURL_jll"))
```
giordano, line 102: It should probably depend on the same version?
```suggestion
        push!(dependencies, Dependency("LibCURL_jll"; compat="$(version)"))
```
I don't know what happens if you try to use a 7.81.0 curl with a 7.71.0 libcurl, but I don't expect anything good.

L/LibCURL/LibCURL/build_tarballs.jl
```diff
@@ -0,0 +1,3 @@
+include("../common.jl")
+
+build_libcurl(ARGS, "LibCURL")
```
giordano, line 3: I don't want to rebuild libcurl, nothing changed.  This file should be removed from this PR and added in a follow-up PR with `[skip ci]`.

L/LibCURL/common.jl
```diff
@@ -0,0 +1,107 @@
...
+        # On Darwin, we need to use SecureTransport (native TLS library)
+        FLAGS+=(--with-secure-transport)
+
+        # We need to explicitly request a higher `-mmacosx-version-min` here, so that it doesn't
+        # complain about: `Symbol not found: ___isOSVersionAtLeast`
+        if [[ "${target}" == aarch64* ]]; then
+            export CFLAGS=-mmacosx-version-min=11.0
+        else
+            export CFLAGS=-mmacosx-version-min=10.11
+        fi
```
giordano, line 59: Same as https://github.com/JuliaPackaging/Yggdrasil/pull/4153#discussion_r810645708

L/LibCURL/common.jl
DilumAluthge: I've removed the executable product from the LibCURL_jll package.

L/LibCURL/common.jl
DilumAluthge: I can't imagine that anything good would happen 😂

L/LibCURL/common.jl
DilumAluthge: I've removed the aarch64 branch. I kept the x86_64 branch.

L/LibCURL/LibCURL/build_tarballs.jl
DilumAluthge: Is it cool if I leave the file in until the PR is ready to merge, just to make sure that CI passes (and that I don't have any bugs for the libcurl code path)? And then, once the PR is ready to merge, I'll remove the file from this PR, and then you can merge immediately afterwards?

L/LibCURL/common.jl
giordano: > I just copied what is currently done on Ygg master:

I didn't realise that... sigh.  So these means they have conflicting files, which is not good.  I presume there is a way to build only libcurl?  And perhaps curl using libcurl from the prefix.  At the moment we're still producing two identical tarballs, which is a waste.

giordano: You mean https://github.com/JuliaLang/julia/blob/0b48b91c9881823d860eaf0a605072b7f20a4cbb/stdlib/LibCURL_jll/src/LibCURL_jll.jl?

DilumAluthge: Yeah.

giordano: As far as I know, all those files have been created manually.  I don't think there is any "code that generates" them.

DilumAluthge: And also, where is the code that makes it so that the `LibCURL_jll` stdlib JLL doesn't include the executable product, even though the `LibCUL_jll` package[^1] does include the executable product.

[^1]: https://github.com/JuliaBinaryWrappers/LibCURL_jll.jl

DilumAluthge: > As far as I know, all those files have been created manually. I don't think there is any "code that generates" them.

Oh, I see.

giordano: (side note: I'm kind of curious why you don't just use hyperlinks instead of having footnotes, which requires _more_ clicks on long posts :slightly_smiling_face:)

giordano:
One thing which you may want to consider is to do a static build of curl which doesn't depend on libcurl at all, which is annoying because that's tied to a julia version.

DilumAluthge:
@giordano @staticfloat Apart from deleting the `L/LibCURL/LibCURL/build_tarballs.jl` file (which I will do right before you merge), is there anything else that needs to be done for this PR?

L/LibCURL/common.jl
DilumAluthge: 😂 😂 😂

DilumAluthge: No particular reason at all. I suppose I just picked up the habit at one point 😂

giordano:
This still creates libcurl in `CURL_jll` and curl in `LibCURL_jll`.  Isn't there a way to avoid this duplication?  This also kind of defeats the point of having `CURL_jll` depending on `LibCURL_jll`.

DilumAluthge:
I have no idea. @staticfloat?

giordano:
Or, put in another way: what do Linux distributions do?  They also split libcurl and curl into different packages.

staticfloat:
Just delete `libcurl` from the `CURL_jll` package.  Once we have the ability to split a single build out into multiple JLLs we can avoid rebuilding the products twice.

L/LibCURL/common.jl
```diff
@@ -0,0 +1,111 @@
...
+    make -j${nproc}
+    if [[ "${THIS_IS_CURL}" == true ]]; then
+        # Manually install only `curl`
+        install -Dm 755 "src/.libs/curl${exeext}" "${bindir}/curl${exeext}"
+    else
+        # Install everything...
+        make install
+        # ...but remove `curl`
+        rm "${bindir}/curl${exeext}"
+    fi
```
giordano, line 73: I followed what I did in https://github.com/JuliaPackaging/Yggdrasil/blob/2fa71a47eedfe3f500896f5aa8bf20306e8049ae/H/HarfBuzz/common.jl#L38-L45 which is what I've been telling to look at

L/LibCURL/common.jl
```diff
@@ -0,0 +1,111 @@
...
+        Dependency("MbedTLS_jll"; compat="~2.28.0", platforms=filter(p->Sys.islinux(p) || Sys.isfreebsd(p), platforms)),
+    ]
+
+    if this_is_curl_jll
+        # Curl_jll depends on LibCURL_jll
+        push!(dependencies, Dependency("LibCURL_jll"; compat="$(version)"))
+    end
+
+    # Build the tarballs, and possibly a `build.jl` as well.
+    build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.8")
```
ericphanson, line 110: Why 1.8 only?

giordano: https://github.com/JuliaLang/julia/blob/release-1.7/stdlib/LibCURL_jll/Project.toml#L3
https://github.com/JuliaLang/julia/blob/release-1.8/stdlib/LibCURL_jll/Project.toml#L3

ericphanson: Oh… I guess that’s why we’d want a fully static thing like you mentioned. Annoying. Thanks for your help with this!

DilumAluthge: @giordano @staticfloat What if we:

1. Don't make CURL_jll depend on LibCURL_jll
2. When building CURL_jll, don't delete `libcurl`. So then CURL_jll will supply both `libcurl` and `curl`.

Would that allow us to use the same version of `CURL_jll` on multiple versions of Julia?

giordano: https://github.com/JuliaPackaging/Yggdrasil/pull/4472#issuecomment-1046852810

DilumAluthge: I see, so we'd need static libraries for all of the dependencies.

DilumAluthge: So this just means that we need to remember to make a new release of CURL_jll every time there's a new minor release of Julia, right?

DilumAluthge:
@giordano @staticfloat @ericphanson Thanks so much for your help with this!

L/LibCURL/common.jl
giordano: No, just always build `CURL_jll` together with (well, right after) a new LibCURL_jll.

DilumAluthge: > No, just always build `CURL_jll` together with (well, right after) a new LibCURL_jll.

It would be nice if we could automate that process.

giordano:
For the record, showing this is working:
```julia
julia> using CURL_jll

julia> run(`$(curl()) --version`);
curl 7.81.0 (x86_64-pc-linux-gnu) libcurl/7.81.0 mbedTLS/2.28.0 zlib/1.2.11 libssh2/1.10.0 nghttp2/1.41.0
Release-Date: 2022-01-05
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS HSTS HTTP2 IPv6 Largefile libz NTLM NTLM_WB SSL UnixSockets
```

L/LibCURL/common.jl
giordano: https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/778.  This is what https://github.com/JuliaPackaging/Yggdrasil/pull/4472#issuecomment-1050346049 and https://github.com/JuliaPackaging/Yggdrasil/blob/2fa71a47eedfe3f500896f5aa8bf20306e8049ae/H/HarfBuzz/common.jl#L17-L19 were referring to.

PR #4474, opened 2022-02-21
Adding pigpio

smith-isaac:
Please evaluate build_tarballs.jl for pigpiod_if2 C library and merge.

PR #4476, opened 2022-02-21
Add pigpiod_if2

smith-isaac:
Build_tarballs file is for [pigpiod_if2 C library](https://abyz.me.uk/rpi/pigpio/pdif2.html) and **not** the pigpio C library. Hence the change from `pigpio` to `pigpiod_if2` to accentuate the difference between the two libraries. Please evaluate and merge.


P/pigpiod_if2/build_tarballs.jl
```diff
@@ -0,0 +1,34 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "pigpiod_if2"
+version = v"0.0.1"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/smith-isaac/pigpio.git", "8592e32edaa7d7567b89e648cbf2d53ac6c5ef41")
```
giordano, line 10: Still using the fork as in #4474?  Also, please note that you don't have to close a pull request every time, you can just keep pushing to the same branch and the PR will be updated.  Creating every time a new PR (a) is wasteful and dispersive, and (b) makes review much harder because it's split across multiple PRs.

smith-isaac: I apologize for all the confusion, I am a little new to this. I thought you were talking about my Yggdrasil fork. Also I realized I committed to the master branch of my fork, so I made a new branch and that's why I closed the other PR.

Back to your original question, since I realized now you were asking why I forked the `pigpio` repo. That one repo builds two different libraries, plus it installs a python library. One C library requires `sudo` privileges to even run and was not easily configurable for different platforms, and I did not want to include the python library. I forked the pigpio repo in order to edit the `Makefile` so that it would (a) build to the correct location per the Wizard, and (b) only build the library for the desired C library, and not the other one. I can see about branching from the original repo if need be.

smith-isaac: I have not been successful in getting the actual `pigpio` to compile when using `BinaryBuilder.runshell()` for any of the supported platforms, which is why I forked and excluded that from the makefile, and resorted to only using the `pigpiod_if2` library.

giordano: Regarding https://github.com/smith-isaac/pigpio/commit/8592e32edaa7d7567b89e648cbf2d53ac6c5ef41, that's easy fixable: you can pass `prefix=${prefix}` to `make`, like this:
```
make install prefix=${prefix}
```
without having to modify anything.  The [other](https://github.com/smith-isaac/pigpio/commit/f08f0b8e0dc6cf3a50ba6f3b46a9f4f996c29aa6) [two](https://github.com/smith-isaac/pigpio/commit/4394eed6b7ac01e47054ed01d7ca6a1e18a0d046) changes are more substantial.  But in general we prefer to use the upstream repository as source, and then add patch on top of it if necessary.  Just a random example: https://github.com/JuliaPackaging/Yggdrasil/blob/8be19820afa24ff817185f3c53c4d20786a8add3/L/Luna/build_tarballs.jl

giordano: Alright, I did it for you, thanks for your contribution!

smith-isaac: Thank you so much for your help! I apologize for all the confusion earlier.

PR #1529, opened 2020-08-25
[WIP] Add Windows support for Torch.

GunnarFarneback:
This is a work in progress attempt to build the Torch.jl support libraries on Windows.

Current stumbling block is a cmake failure to locate CUDA libraries, which is not really unexpected. The Linux build does a fair amount of tweaking of those and Windows probably need a couple of different tweaks.

PR #4477, opened 2022-02-21
WIP: [Torch] Renamed Torch to TorchCUDA, and added a CPU-only Torch

stemann:
Currently a WIP. Aims at resolving https://github.com/FluxML/Torch.jl/issues/20.

Similar approach as for ONNXRuntime - separate binaries for CPU-only and CUDA etc. (CPU: #4369, CUDA: #4386).

Relates to: #1529


T/Torch/Torch/build_tarballs.jl
```diff
@@ -54,8 +58,6 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
```
giordano, line 60: We'll eventually require `dependencies` to be a vector of `AbstractDependency`s
```suggestion
dependencies = Dependency[
```

vchuravy:
Does this need to be a different package? Or can we turn this into a variant?

stemann:
> Does this need to be a different package? Or can we turn this into a variant?

If eventually both CPU, CUDA and ROCm builds are available, wouldn't the different sets of dependencies necessitate different packages?

T/Torch/Torch/build_tarballs.jl
stemann: Done

vchuravy:
You still have the same issue that if Torch.jl includes `Torch_jll, TorchCUDA_jll,...` all dependencies must be fulfilled. Better to use a platform tag and platform specific dependencies.

stemann:
> You still have the same issue that if Torch.jl includes `Torch_jll, TorchCUDA_jll,...` all dependencies must be fulfilled. Better to use a platform tag and platform specific dependencies.

Right, I see. Is there some discussion/registry of defined/reserved platform tags? I.e. that "cuda" is reserved for CUDA and so on for e.g. "amd_rocm", "intel_oneapi", "apple_coreml", ...

stemann:
Superseded by #4554

PR #4479, opened 2022-02-21
[MSTM] Add Windows support

lucifer1004:


M/MSTM/build_tarballs.jl
```diff
@@ -10,14 +10,32 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/MSTM/code
-mpifort -O2 mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 -o "${bindir}/mstm${exeext}"
+if [[ "$target" == x86_64-w64-mingw32 ]]; then
+    cd $WORKSPACE/destdir/include
+    cp $WORKSPACE/destdir/src/mpi.f90 .
+    gfortran -DWIN64 -DINT_PTR_KIND=8 -fno-range-check mpi.f90 || true
```
giordano, line 16: What's the point of this?

M/MSTM/build_tarballs.jl
```diff
@@ -10,14 +10,32 @@ sources = [
...
+    cd $WORKSPACE/srcdir/MSTM/code
+    echo "void __guard_check_icall_fptr(unsigned long ptr) { }" > cfg_stub.c
+    gcc -c cfg_stub.c
+    gfortran -O2 -fno-range-check mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 cfg_stub.o -L${WORKSPACE}/destdir/lib -I${WORKSPACE}/destdir/include -lmsmpifec64 -lmsmpi64 -o "${bindir}/mstm${exeext}"
```
giordano, line 20: Please always use the most specific [environment variables](https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables) available:
```suggestion
    gfortran -O2 -fno-range-check mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 cfg_stub.o -L"${libdir}" -I"${includedir}" -lmsmpifec64 -lmsmpi64 -o "${bindir}/mstm${exeext}"
```
`${WORKSPACE}/destdir/lib` is even wrong since libraries for Windows are under `bin/`

M/MSTM/build_tarballs.jl
```diff
@@ -10,14 +10,32 @@ sources = [
...
+elif [[ "$target" == *-mingw* ]]; then
```
giordano, line 21: What's the difference with the branch above?  If you want to test for 32-bit Windows (`i686-w64-mingw`) do that explicitly.

M/MSTM/build_tarballs.jl
```diff
@@ -10,14 +10,32 @@ sources = [
...
+    cd $WORKSPACE/destdir/include
+    cp $WORKSPACE/destdir/src/mpi.f90 .
+    gfortran -DWIN32 -DINT_PTR_KIND=8 -fno-range-check mpi.f90 || true
+    cd $WORKSPACE/srcdir/MSTM/code
+    echo "void __guard_check_icall_fptr(unsigned long ptr) { }\nvoid __security_check_cookie(void) { }" > cfg_stub.c
+    gcc -c cfg_stub.c
+    gfortran -O2 -fno-range-check mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 cfg_stub.o -L${WORKSPACE}/destdir/lib -I${WORKSPACE}/destdir/include -lmsmpifec -lmsmpi -o "${bindir}/mstm${exeext}"
```
giordano, line 28:
```suggestion
    gfortran -O2 -fno-range-check mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 cfg_stub.o -L"${libdir}" -I"${includedir}" -lmsmpifec -lmsmpi -o "${bindir}/mstm${exeext}"
```

M/MSTM/build_tarballs.jl
```diff
@@ -26,9 +44,10 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="MPICH_jll", uuid="7cb0a576-ebde-5e09-9194-50597f1243b4")),
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    Dependency(PackageSpec(name = "MPICH_jll", uuid = "7cb0a576-ebde-5e09-9194-50597f1243b4")),
+    Dependency(PackageSpec(name = "MicrosoftMPI_jll", uuid = "9237b28f-5490-5468-be7b-bb81f5f5e6cf")),
```
giordano, line 48:
```suggestion
    Dependency(PackageSpec(name = "MPICH_jll", uuid = "7cb0a576-ebde-5e09-9194-50597f1243b4"); platforms=filter(!Sys.iswindows, platforms)),
    Dependency(PackageSpec(name = "MicrosoftMPI_jll", uuid = "9237b28f-5490-5468-be7b-bb81f5f5e6cf"); platforms=filter(Sys.iswindows, platforms)),
```

M/MSTM/build_tarballs.jl
lucifer1004: The Fortran program needs `mpi.mod` to compile, which is not provided by `MicrosoftMPI_jll`, so I have to compile it manually.

M/MSTM/build_tarballs.jl
lucifer1004: Yes, the second branch is for 32-bit Windows. I have modified the code according to your suggestion.

lucifer1004:
@giordano What's your suggestion on the version number?

lucifer1004:
It is strange that new errors occurred after I changed `*-mingw*` to `i686-w64-mingw32`.

M/MSTM/build_tarballs.jl
giordano: I think I'm missing what's the point of allowing the failure

lucifer1004: The compilation will fail due to the lack of the main function, but the `.mod` files will be compiled.

giordano:
> It is strange that new errors occurred after I changed *-mingw* to i686-w64-mingw32.

It doesn't look like it's working with `*-mingw` either.

M/MSTM/build_tarballs.jl
giordano: Wouldn't `-c` achieve the same goal?

lucifer1004:
> > It is strange that new errors occurred after I changed _-mingw_ to i686-w64-mingw32.
> 
> It doesn't look like it's working with `*-mingw` either.

Yes... But https://github.com/JuliaPackaging/Yggdrasil/pull/4479/commits/7c31a18db1b0c7b320955161988fb665a6794a67 succeeded. So there must be something else causing the new issue.

M/MSTM/build_tarballs.jl
lucifer1004: Change made!

lucifer1004:
Is `${libdir}` strictly equivalent to `${WORKSPACE}/destdir/lib` here? Also, is it the same for `${includedir}`?

giordano:
> Is ${libdir} strictly equivalent to ${WORKSPACE}/destdir/lib here?

No, and that's what I [said above](https://github.com/JuliaPackaging/Yggdrasil/pull/4479#discussion_r811346498)

lucifer1004:
I reverted `libdir` and `includedir` and now the build passed.

M/MSTM/build_tarballs.jl
lucifer1004: > > Is ${libdir} strictly equivalent to ${WORKSPACE}/destdir/lib here?
> 
> No, and that's what I [said above](https://github.com/JuliaPackaging/Yggdrasil/pull/4479#discussion_r811346498)

But the lib files (MPI libs) I need are located in `/lib` instead of `/bin` for Windows. Maybe that was the cause?

M/MSTM/build_tarballs.jl
```diff
@@ -10,14 +10,32 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/MSTM/code
-mpifort -O2 mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 -o "${bindir}/mstm${exeext}"
+if [[ "$target" == x86_64-w64-mingw32 ]]; then
+    cd $WORKSPACE/destdir/include
+    cp $WORKSPACE/destdir/src/mpi.f90 .
+    gfortran -c -DWIN64 -DINT_PTR_KIND=8 -fno-range-check mpi.f90
```
giordano, line 16: I just realised you're polluting the prefix with files not related to your package.  This isn't good.  Are these modules needed only at buildtime?  If so, then move them to the source directory.

M/MSTM/build_tarballs.jl
```diff
@@ -10,14 +10,32 @@ sources = [
...
+    cd $WORKSPACE/srcdir/MSTM/code
+    echo "void __guard_check_icall_fptr(unsigned long ptr) { }" > cfg_stub.c
+    gcc -c cfg_stub.c
+    gfortran -O2 -fno-range-check mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 cfg_stub.o -L${WORKSPACE}/destdir/lib -I${WORKSPACE}/destdir/include -lmsmpifec64 -lmsmpi64 -o "${bindir}/mstm${exeext}"
```
giordano, line 20: Again:
```suggestion
    gfortran -O2 -fno-range-check mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 cfg_stub.o -L${WORKSPACE}/destdir/lib -I"${includedir}" -lmsmpifec64 -lmsmpi64 -o "${bindir}/mstm${exeext}"
```

M/MSTM/build_tarballs.jl
giordano: And again:
```suggestion
    gfortran -O2 -fno-range-check mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 cfg_stub.o -L${WORKSPACE}/destdir/lib -I"${includedir}" -lmsmpifec -lmsmpi -o "${bindir}/mstm${exeext}"
```

giordano (changes requested):
What libraries do you need from `${prefix}/lib` on Windows?  That looks wrong

M/MSTM/build_tarballs.jl
lucifer1004: I need the MPI headers and libraries only for building the program.

M/MSTM/build_tarballs.jl
giordano: What do you mean by "MPI libs"?  The shared libraries are in `bin/` (where they should be):
```console
% tar tzvf MicrosoftMPI.v10.1.3.x86_64-w64-mingw32.tar.gz
-rwxr-xr-x 0/0          544536 1970-01-01 01:00 bin/IMB_MPI1.exe
-rwxr-xr-x 0/0          558360 1970-01-01 01:00 bin/IMB_NBC.exe
-rwxr-xr-x 0/0          211224 1970-01-01 01:00 bin/MpiPingpong.exe
-rwxr-xr-x 0/0          131277 1970-01-01 01:00 bin/generate_compile_time_mpi_constants.exe
-rwxr-xr-x 0/0          130578 1970-01-01 01:00 bin/libload_time_mpi_constants.dll
-rwxr-xr-x 0/0          549656 1970-01-01 01:00 bin/mpiexec.exe
-rwxr-xr-x 0/0         1631000 1970-01-01 01:00 bin/msmpi.dll
-rwxr-xr-x 0/0          159512 1970-01-01 01:00 bin/msmpilaunchsvc.exe
-rwxr-xr-x 0/0          443160 1970-01-01 01:00 bin/msmpires.dll
-rwxr-xr-x 0/0          426776 1970-01-01 01:00 bin/smpd.exe
-rw-r--r-- 0/0          175354 1970-01-01 01:00 include/mpi.h
-rw-r--r-- 0/0           20078 1970-01-01 01:00 include/mpif.h
-rw-r--r-- 0/0             266 1970-01-01 01:00 include/mpifptr.h
-rw-r--r-- 0/0             143 1970-01-01 01:00 include/mpio.h
-rw-r--r-- 0/0         1058451 1970-01-01 01:00 include/mpitrace.man
-rw-r--r-- 0/0            5278 1970-01-01 01:00 include/mspms.h
-rw-r--r-- 0/0           17980 1970-01-01 01:00 include/pmidbg.h
-rw-r--r-- 0/0          161382 1970-01-01 01:00 lib/libload_time_mpi_constants.dll.a
-rw-r--r-- 0/0          337216 1970-01-01 01:00 lib/msmpi.lib
-rw-r--r-- 0/0          321236 1970-01-01 01:00 lib/msmpi64.lib
-rw-r--r-- 0/0         1075098 1970-01-01 01:00 lib/msmpifec.lib
-rw-r--r-- 0/0         1020116 1970-01-01 01:00 lib/msmpifec64.lib
-rw-r--r-- 0/0         1147914 1970-01-01 01:00 lib/msmpifes.lib
-rw-r--r-- 0/0         1077274 1970-01-01 01:00 lib/msmpifmc.lib
-rw-r--r-- 0/0         1035480 1970-01-01 01:00 lib/msmpifmc64.lib
-rw-r--r-- 0/0         1147914 1970-01-01 01:00 lib/msmpifms.lib
-rw-r--r-- 0/0            4890 1970-01-01 01:00 logs/MicrosoftMPI/MicrosoftMPI.log.gz
-rw-r--r-- 0/0            1076 1970-01-01 01:00 share/licenses/MicrosoftMPI/LICENSE.md
-rw-r--r-- 0/0           13491 1970-01-01 01:00 share/licenses/MicrosoftMPI/MPI_Redistributables_TPN.txt
-rw-r--r-- 0/0            2150 1970-01-01 01:00 share/licenses/MicrosoftMPI/MPI_SDK_TPN.txt
-rw-r--r-- 0/0          224476 1970-01-01 01:00 share/licenses/MicrosoftMPI/MicrosoftMPI_Redistributable_EULA.rtf
-rw-r--r-- 0/0          224238 1970-01-01 01:00 share/licenses/MicrosoftMPI/MicrosoftMPI_SDK_EULA.rtf
-rw-r--r-- 0/0           55162 1970-01-01 01:00 src/mpi.f90
```
I think the files under `lib/` are import libraries?

lucifer1004: I need the `.lib` files instead of the `.dll` files to compile the program.

giordano: Why?  Still, please use `${prefix}` instead of `${WORKSPACE}/destdir`.  No one should ever use `${WORKSPACE}/destdir` directly, we may move the prefix at some point in the future (which we actually did already and had to keep a symlink to not break recipes).

lucifer1004: Because the original Fortran program does not compile when using dynamic libraries...

M/MSTM/build_tarballs.jl
lucifer1004: Extra files are now removed after compilation.

lucifer1004:
Should be OK now.

M/MSTM/build_tarballs.jl
```diff
@@ -10,14 +10,34 @@ sources = [
...
+    rm ${includedir}/mpi.f90 ${includedir}/*.mod ${includedir}/*.o
+elif [[ "$target" == i686-w64-mingw32 ]]; then
+    cd ${includedir}
+    cp ${prefix}/src/mpi.f90 .
+    gfortran -c -DWIN32 -DINT_PTR_KIND=8 -fno-range-check mpi.f90
+    cd ${WORKSPACE}/srcdir/MSTM/code
+    echo "void __guard_check_icall_fptr(unsigned long ptr) { } void __security_check_cookie(void) { }" > cfg_stub.c
+    gcc -c cfg_stub.c
+    gfortran -O2 -fno-range-check mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 cfg_stub.o -L${prefix}/lib -I${includedir} -lmsmpifec -lmsmpi -o "${bindir}/mstm${exeext}"
+    rm ${includedir}/mpi.f90 ${includedir}/*.mod ${includedir}/*.o
```
giordano, line 30: There is still lots of code duplication which you can reduce with
```suggestion
if [[ "$target" == *-mingw* ]]; then
    cd ${includedir}
    cp ${prefix}/src/mpi.f90 .
    gfortran -c -DWIN${nbits} -DINT_PTR_KIND=8 -fno-range-check mpi.f90
    cd ${WORKSPACE}/srcdir/MSTM/code
    if [[ ${target} == x86_64-* ]]; then
        cfg_stub="void __guard_check_icall_fptr(unsigned long ptr) { }"
        msmpifec=msmpifec64
        msmpi=msmpi64
    else
        cfg_stub="void __guard_check_icall_fptr(unsigned long ptr) { } void __security_check_cookie(void) { }"
        msmpifec=msmpifec
        msmpi=msmpi
    fi
    echo "${cfg_stub}" | gcc -x c -c -o cfg_stub.o -
    gfortran -O2 -fno-range-check mpidefs-parallel.f90 mstm-intrinsics.f90 mstm-v4.0.f90 cfg_stub.o -L${prefix}/lib -I${includedir} -l${msmpifec} -l${msmpi} -o "${bindir}/mstm${exeext}"
    rm ${includedir}/mpi.f90 ${includedir}/*.mod ${includedir}/*.o
```

M/MSTM/build_tarballs.jl
```diff
@@ -1,7 +1,7 @@
 using BinaryBuilder, Pkg
 
 name = "MSTM"
-version = v"4.0.0"
+version = v"4.0.1"
```
giordano, line 4: We tend to keep the same version number as upstream, if possible

lucifer1004: Done

giordano:
Ok, this is much better now!  Thanks for sticking with us!

lucifer1004:
Thank you for your patience and support! I have learned quite a lot (mostly from you) in the last few hours.

PR #3620, opened 2021-09-14
Upgrade NetCDF to 4.8.1

Alexander-Barth:
This PR upgraded NetCDF to 4.8.1 from 4.7.4..

I had this error described [here](https://discourse.julialang.org/t/yggdrasil-loaderror-dependency-hdf5-jll-specifies-a-version-use-build-version-and-or-compat-instead/68134) but changing the `Dependency` with respect to `HDF5_jll` as shown in the PR seemed to resolved this issue.

I kept the `version_offset` unchanged, but maybe they need to be reset to v"0.0.0"?

PR #4481, opened 2022-02-22
Revert "Upgrade NetCDF to 4.8.1 (#3620)"

felixcremer:
This reverts commit 453fb46233888b5e7d4238463148fa260e68d25f.

This should enable NetCDF 4.7 on apple M1 until we fixed the windows problems of NetCDF 4.8.

I ran the build script locally, but I am not sure, whether it build everything it should.

### DIFF
--- a/N/NetCDF/NetCDF@julia-1.3/build_tarballs.jl
+++ b/N/NetCDF/NetCDF@julia-1.3/build_tarballs.jl
@@ -2,7 +2,6 @@ include("../common.jl")
 
 # Offset to add to the version number
 version_offset = v"0.1.0"
-
 # Minimum Julia version supported: this is important to decide which versions of
 # the dependencies to use, in particular the JLL stdlibs.
 min_julia_version = v"1.3"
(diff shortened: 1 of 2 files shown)


N/NetCDF/NetCDF@julia-1.3/build_tarballs.jl
```diff
@@ -2,7 +2,6 @@ include("../common.jl")
 
 # Offset to add to the version number
 version_offset = v"0.1.0"
-
```
giordano, line 5: This file shouldn't be touched.  You should instead touch [`N/NetCDF/NetCDF@julia-1.6/build_tarballs.jl`](https://github.com/JuliaPackaging/Yggdrasil/blob/0f5bbbe5800314419465747c07547a6ddb96bda5/N/NetCDF/NetCDF%40julia-1.6/build_tarballs.jl)

felixcremer: Thanks.

PR #4487, opened 2022-02-22
[xrootdgo] add export comment

Moelf:
sorry!

### DIFF
--- a/X/xrootdgo/build_tarballs.jl
+++ b/X/xrootdgo/build_tarballs.jl
@@ -1,7 +1,7 @@
 using BinaryBuilder
 
 name = "xrootdgo"
-version = v"0.30.1"
+version = v"0.30.1b"
 
 sources = [
     GitSource("https://github.com/go-hep/hep/",
--- a/X/xrootdgo/build_tarballs.jl
+++ b/X/xrootdgo/build_tarballs.jl
@@ -1,7 +1,7 @@
 using BinaryBuilder
 
 name = "xrootdgo"
-version = v"0.30.1b"
+version = v"0.30.1"
 
 sources = [
     GitSource("https://github.com/go-hep/hep/",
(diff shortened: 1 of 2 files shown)


X/xrootdgo/build_tarballs.jl
```diff
@@ -1,7 +1,7 @@
 using BinaryBuilder
 
 name = "xrootdgo"
-version = v"0.30.1"
+version = v"0.30.1b"
```
giordano, line 4: Pkg/the registry don't like pre-release strings, just stick to X.Y.Z as format
```suggestion
version = v"0.30.1"
```

Moelf: but the `_jll` already merged, can we re-use the same version number?

giordano: Yes, we always bump the build number in any case.  Which is another thing which Pkg/the registry don't like, but better to stick to a single thing which they don't like :smile:

Moelf:
https://github.com/JuliaBinaryWrappers/xrootdgo_jll.jl/tags

so how am I suppose to get new artifacts?

giordano:
Something failed: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=17527&view=logs&j=8a8167d4-ab0f-5389-04d0-3a98cfa6c24a&t=14600958-7c1c-5787-d982-79da65d35a5b&l=186 :confused:

Moelf:
oh I see.....is it because we already have 0.30.1 assets? I do not know how it works

giordano:
That's what happened, but we use [this function](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/45b9ee380e6d6ffe606ced4483749ae7e51839e9/src/AutoBuild.jl#L452-L494) to always increase the build number.  I restarted the job and it worked, maybe it happened that somehow it had an old version of the registry.

PR #4496, opened 2022-02-23
[TempestRemap] try without HDF5

simonbyrne:
cc @jakebolewski @mkitti @giordano


simonbyrne:
Conda doesn't use hdf5: https://github.com/conda-forge/tempest-remap-feedstock/blob/14f4ac71f955ab1066946a5b691df5536df9917a/recipe/build.sh#L22-L23

PR #4500, opened 2022-02-24
[SCS] update to v3.2.0

kalmarek:
Before we merge this let me check that the binaries actually work with SCS.jl

### DIFF
--- a/S/SCS_GPU/build_tarballs.jl
+++ b/S/SCS_GPU/build_tarballs.jl
@@ -2,11 +2,11 @@ using Pkg
 using BinaryBuilder
 
 name = "SCS_GPU"
-version = v"3.0.1"
+version = v"3.2.0"
 
 # Collection of sources required to build SCSBuilder
 sources = [
-    GitSource("https://github.com/cvxgrp/scs.git", "3aaa93c7aa04c7001df5e51b81f21b126dfa99b3")
+    GitSource("https://github.com/cvxgrp/scs.git", "ac6840a3b3264950e6c300264cbf3937e0bcc6c5")
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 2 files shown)


giordano:
Marking as draft until it's ready

S/SCS_GPU/build_tarballs.jl
```diff
@@ -2,18 +2,18 @@ using Pkg
...
+    GitSource("https://github.com/cvxgrp/scs.git", "ac6840a3b3264950e6c300264cbf3937e0bcc6c5")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/scs*
-flags="DLONG=0 USE_OPENMP=0 BLAS64=1 BLASSUFFIX=_64_"
-blasldflags="-L${libdir} -lopenblas64_"
+flags="DLONG=0 USE_OPENMP=0"
+blasldflags="-L${prefix}/lib -lopenblas"
```
giordano, line 16: Why changing `${libdir}` to `${prefix}/lib`?
```suggestion
blasldflags="-L${libdir} -lopenblas"
```

kalmarek: hmm, I copied this from SCS `build_tarballs.jl`. I guess both should point to `${libdir}`?

giordano: > I guess both should point to `${libdir}`?

Not always.   It's preferable to point to the most specific and semantically relevant environment variable.  Also, it was `${libdir}` before, I see no point in changing it

kalmarek: ok, so what is the reason to have `-L${prefix}/lib` in SCS `build_tarballs.jl`? i.e. what is the canonical dir where all deps are installed?

btw. thanks for taking time to review this ;)

giordano: > ok, so what is the reason to have -L${prefix}/lib in SCS build_tarballs.jl?

None, that's wrong as far as I'm concerned.  I guess I overlooked it before :slightly_smiling_face: 

>  i.e. what is the canonical dir where all deps are installed?

`${libdir}`

kalmarek: verywell, I'll change it there the next time. thanks!

Issue #4507, opened 2022-02-27
macOS ARM (m1) packaging support for libisoband


Mixpap:
Hi, pasting from GLMakie issues (https://github.com/JuliaPlots/Makie.jl/issues/1708):
Trying to make a contour plot through GLMakie in a mac-os m1 arm system (julia version macOS ARM) gives this error. 
`ERROR: UndefVarError: libisoband not defined`

Julia version: 1.7.2
GLMakie version: GLMakie v0.5.4

jeremiahpslewis:
https://github.com/JuliaPackaging/Yggdrasil/pull/4508

PR #4508, opened 2022-02-27
[isoband] Bump to experimental platforms

jeremiahpslewis:
Resolve https://github.com/JuliaPackaging/Yggdrasil/issues/4507

### DIFF
--- a/I/isoband/build_tarballs.jl
+++ b/I/isoband/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "isoband"
-version = v"0.2.1" # Fake version for building experimental platforms and setting Julia version to 1.6
+version = v"0.2.3" # Fake version for building experimental platforms and setting Julia version to 1.6
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


giordano:
It doesn't look like expanding the C++ string ABI is actually necessary.

I/isoband/build_tarballs.jl
```diff
@@ -31,4 +31,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
giordano, line 34: Ok, this doesn't make much sense to me.  I'd try using GCC 5 and see what happens
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"5")
```

giordano: Yeah, using GCC 5 the library doesn't expose any string at all.  It doesn't make any sense at all.

PR #4523, opened 2022-03-01
TEST: see which platforms need includedir (and possibly libdir) handling

fingolfin:
See https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/226


giordano:
Can this be closed now?

fingolfin:
Perhaps re-run this once the BB changes are in effect?

giordano:
Hhsuite is failing for aarch64-apple-darwin, but it was never built for that platform before, so we may just exclude it for the time being. I don't understand Libarchive failures, but they look unrelated. The only relevant errors are the libwebp and unpaper ones, but I think those may have to do with libdir

PR #4527, opened 2022-03-01
[LLVMExtra] Make artifacts lazy due to platform hooks

vchuravy:
https://github.com/maleadt/LLVM.jl/pull/296#issuecomment-1055105781

### DIFF
--- a/L/LLVMExtra/build_tarballs.jl
+++ b/L/LLVMExtra/build_tarballs.jl
@@ -89,5 +89,5 @@ for (i,build) in enumerate(builds)
                    name, version, sources, script,
                    build.platforms, build.products, build.dependencies;
                    preferred_gcc_version=v"8", julia_compat="1.6",
-                   augment_platform_block)
+                   augment_platform_block, lazy_artifacts=true)
 end


L/LLVMExtra/build_tarballs.jl
```diff
@@ -89,5 +89,5 @@ for (i,build) in enumerate(builds)
                    name, version, sources, script,
                    build.platforms, build.products, build.dependencies;
                    preferred_gcc_version=v"8", julia_compat="1.6",
-                   augment_platform_block)
+                   augment_platform_block, lazy_artifacts=true)
```
giordano, line 92: If we always need this perhaps we should default `lazy_artifacts` to `isempty(augment_platform)`?

vchuravy:
I think this will still fail. 

@maleadt and I were just discussing this and we think that the lack of TOML as a dependency might be due to the fact that the augmented_platform_hock is not passed through the meta?

L/LLVMExtra/build_tarballs.jl
vchuravy: If `julia_compat < v"1.7" && !isempty(augment_platform) ` but yes

giordano:
Probably you need to update https://github.com/JuliaPackaging/Yggdrasil/blob/4cc178dbd75eb7a8e9efe8e6a2bac033cb2e05db/.ci/register_package.jl#L135?

giordano:
You may also want to update the tests in https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/09336062a3873f161744f75574b71f91fef6cdd7/test/jll.jl#L97-L147 to match what you do here.  Or perhaps add a test there for a package with a non-empty augment block to make sure it's working as expected.

PR #4531, opened 2022-03-02
Wizard recipe: ranger-v0.13.1

tecosaur:

* Package name: ranger
* Version: v0.13.1

### DIFF
--- a/R/ranger/build_tarballs.jl
+++ b/R/ranger/build_tarballs.jl
@@ -25,21 +25,7 @@ install_license /usr/share/licenses/MIT
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = [
-    Platform("i686", "linux"; libc = "glibc"),
-    Platform("x86_64", "linux"; libc = "glibc"),
-    Platform("aarch64", "linux"; libc = "glibc"),
-    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
-    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
-    Platform("powerpc64le", "linux"; libc = "glibc"),
-    Platform("i686", "linux"; libc = "musl"),
-    Platform("x86_64", "linux"; libc = "musl"),
-    Platform("aarch64", "linux"; libc = "musl"),
-    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
-    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
-    Platform("x86_64", "macos"; ),
-    Platform("aarch64", "macos"; )
-]
+platforms = supported_platforms(;exclude=Sys.iswindows)
 platforms = expand_cxxstring_abis(platforms)
 
 # The products that we will ensure are always built
--- a/R/ranger/build_tarballs.jl
+++ b/R/ranger/build_tarballs.jl
@@ -1,0 +1,56 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "ranger"
+version = v"0.13.1"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/imbs-hl/ranger/archive/refs/tags/v$(version).tar.gz", "de60c5ca6ffab1b6cd17c8058c7736f74944841d782707906ba3c68530688916"),
+    DirectorySource("./bundled")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+cd ranger-0.13.1/cpp_version/
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIB=1
+make -j$nproc
+make install
+exit
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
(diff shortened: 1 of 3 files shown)


tecosaur:
BinaryBuilder didn't pick up on any licence file, because there isn't any, but the C++ files start with comments that declare them to be under the MIT licence.

https://github.com/imbs-hl/ranger/blob/9aac567f21ef45f84b05ed53d2909e243e5ae7c1/cpp_version/src/main.cpp#L8
> the C++ core of ranger is distributed under MIT license

Please advise whether any action is needed.

R/ranger/bundled/patches/make_lib_and_install.patch
```diff
@@ -0,0 +1,44 @@
+diff --git before/ranger-0.13.1/cpp_version/CMakeLists.txt after/ranger-0.13.1/cpp_version/CMakeLists.txt
+index 397bb53..34ee308 100644
+--- before/ranger-0.13.1/cpp_version/CMakeLists.txt
++++ after/ranger-0.13.1/cpp_version/CMakeLists.txt
```
giordano, line 4:
```suggestion
--- before/cpp_version/CMakeLists.txt
+++ after/cpp_version/CMakeLists.txt
```

R/ranger/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+    DirectorySource("./bundled")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+cd ranger-0.13.1/cpp_version/
```
giordano, line 20:
```suggestion
cd $WORKSPACE/srcdir/ranger*/cpp_version/
for f in ${WORKSPACE}/srcdir/patches/*.patch; do
    atomic_patch -p2 ${f}
done
```

R/ranger/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIB=1
+make -j$nproc
+make install
+exit
```
giordano, line 24: > Please advise whether any action is needed.

We have the text of the MIT license in `/usr/share/licenses/`
```suggestion
install_license /usr/share/licenses/MIT
```

R/ranger/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; )
+]
+
```
giordano, line 44:
```
┌ Warning: /agent/_work/1/s/R/ranger/build/x86_64-linux-gnu/q8Sqg58m/x86_64-linux-gnu-libgfortran3-cxx03/destdir/lib/libranger.so contains std::string values!  This causes incompatibilities across the GCC 4/5 version boundary.  To remedy this, you must build a tarball for both GCC 4 and GCC 5.  To do this, immediately after your `platforms` definition in your `build_tarballs.jl` file, add the line:
│ 
└ @ BinaryBuilder.Auditor /depot/packages/BinaryBuilder/9VQVb/src/auditor/compiler_abi.jl:247
```
```suggestion
platforms = expand_cxxstring_abis(platforms)
```

R/ranger/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; )
+]
```
giordano, line 43:
```suggestion
platforms = supported_platforms(;exclude=Sys.iswindows)
```

giordano: I don't know what was the problem, let's see?

tecosaur: I'm afraid I don't have a record. Would the thing to do be commit your suggestion and look for CI failures?

giordano: Yes

giordano:
```
[16:34:42] /opt/bin/x86_64-unknown-freebsd12.2-libgfortran3-cxx11/x86_64-unknown-freebsd12.2-clang++ --sysroot=/opt/x86_64-unknown-freebsd12.2/x86_64-unknown-freebsd12.2/sys-root/  -Wall -std=c++11 -O3 -DNDEBUG CMakeFiles/ranger-bin.dir/src/main.o -o ranger  -Wl,-rpath,/workspace/srcdir/ranger-0.13.1/cpp_version libranger.so 
[16:34:42] libranger.so: undefined reference to `pthread_create'
```
It's missing `-pthread`, that looks like an upstream bug to me

tecosaur:
For now would it be best to just leave out the freebsd build?

giordano:
https://github.com/imbs-hl/ranger/blob/9aac567f21ef45f84b05ed53d2909e243e5ae7c1/cpp_version/CMakeLists.txt#L20-L23
```
## ======================================================================================##
## In Clang phtread flag only for compiler, not for linker. For
## windows use static linking
## ======================================================================================##
```

tecosaur:
Right. I must confess that I'm not sure what to make of that — I don't do much with C/C++/make.

giordano:
Even though they should do something like https://stackoverflow.com/a/29871891/2442087
```cmake

File: R/ranger/bundled/patches/make_lib_and_install.patch [hunk 6]
```diff
@@ -0,0 +1,44 @@
+diff --git before/ranger-0.13.1/cpp_version/CMakeLists.txt after/ranger-0.13.1/cpp_version/CMakeLists.txt
```
giordano, line 1: It'd be nice to upstream your patch, also to reduce our maintenance burden.  Building the shared library can be useful also for others.

tecosaur:
Thanks!

PR #4533, opened 2022-03-02
update minifb to latest

aviks:
~Draft. Lets see what breaks. Currently only tested on `x86_64-linux-gnu`~

### DIFF
--- a/M/MiniFB/build_tarballs.jl
+++ b/M/MiniFB/build_tarballs.jl
@@ -26,13 +26,12 @@ cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DMINIFB_AVOID_CPP_HEADERS=ON \
     ..
 make -j${nproc}
-mv libminifb* ${libdir}
+install -Dm 755 "libminifb.${dlext}" "${libdir}/libminifb.${dlext}"
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
-
+platforms = supported_platforms(; exclude=p->arch(p)=="armv6l")
 
 # The products that we will ensure are always built
 products = [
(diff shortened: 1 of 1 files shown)


aviks:
The _musl_ and the _armv6_ builds fail due to the apparent unavailability of `X11/Xlib.h`.  I'm minded to just disable those builds. 

The mac one needs 10.13 at the minimum, and we compile with 10.12. I've asked upstream if they want to support 10.12, otherwise I'll have to find a workaround.

giordano:
We don't have the X11 stack for armv6l, so that has likely to be discarded.  For the other musl platforms and FreeBSD you have to set
```sh
export CFLAGS="-I${includedir}"
```
This will be fixed by https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/226.

M/MiniFB/build_tarballs.jl
```diff
@@ -3,33 +3,35 @@
...
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/minifb/
 sed -i -e 's/add_library(minifb STATIC/add_library(minifb SHARED/' \
-    -e 's/ -Wall/-I$ENV{includedir} -Wall/' \
+    -e 's/-Wall/-I$ENV{includedir} -Wall/' \
+    -e 's/Opengl32/opengl32/' \
     CMakeLists.txt
+sed -i -e 's?<gl/gl.h>?<GL/gl.h>?' src/gl/MiniFB_GL.c
```
giordano, line 20: It'd be good to upstream these fixes.

M/MiniFB/build_tarballs.jl
```diff
@@ -3,33 +3,35 @@
 using BinaryBuilder, Pkg
 
 name = "MiniFB"
-version = v"0.1.1"
+version = v"0.2.1"
```
Sid-Bhatia-0, line 6: Should it be `v"0.2.0"` or am I missing something?

M/MiniFB/build_tarballs.jl
aviks: Yeah, I'll try.

aviks:
This is ready to go.  🎉 

(Freebsd fails to find `xkbcommon.h` even after exporting CFLAGS, don't want to spend more time on it.)

giordano:
We don't have xkbcommon for FreeBSD at all: https://github.com/JuliaPackaging/Yggdrasil/blob/355f7273d6913982fc5740a52280366b89174479/X/xkbcommon/build_tarballs.jl#L35  I had a quick look the other day, it shouldn't be impossible to make it work, but it'll take some extra time.

PR #4537, opened 2022-03-03
Use internal libtiff

evetion:
So newer compression features work out of the box, such as `LERC` in `COG`. See https://gdal.org/drivers/raster/cog.html and similar PR at https://github.com/OSGeo/homebrew-osgeo4mac/issues/1087.

I'm unsure/unaware about hard differences between the internal and external libtiff library or other drawbacks. I guess it should just work and gives us more features. Looking at https://gdal.org/drivers/raster/gtiff.html#creation-options, newer compression are first introduced in the internal libtiff and only later in the external one (or not yet, such as `JXL`).

Linking to GDAL.jl driver support issue https://github.com/JuliaGeo/GDAL.jl/issues/65

### DIFF
--- a/G/GDAL/build_tarballs.jl
+++ b/G/GDAL/build_tarballs.jl
@@ -60,7 +60,7 @@ rm -f ${prefix}/lib/*.la
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} \
     --with-geos=${bindir}/geos-config \
     --with-proj=$prefix \
-    --with-tiff=$prefix \
+    --with-libtiff=internal \
     --with-geotiff=$prefix \
     --with-libz=$prefix \
     --with-expat=$prefix \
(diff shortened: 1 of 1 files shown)


visr:
The homebrew discussion is a few years old. It does seem that since then features have been upstreamed sooner to https://gitlab.com/libtiff/libtiff, which has seen more frequent releases. So perhaps part of the COG docs are outdated already. If I search in the libtiff repo for LERC or JXL, it does seem like there is support.

GDAL_jll depends on PROJ_jll which depends on Libtiff_jll, so this does not avoid the dependency.

Perhaps it would be good to first ask on the GDAL mailing list if there are still many advantages to internal libtiff?

evetion:
Ok, so the other route would be to build LERC on Yggdrasil and link it to libtiff.

evetion:
Closed in favor of #4538.

PR #4538, opened 2022-03-03
Add LERC compression to libtiff

evetion:
@Crghilardi added LERC to Yggdrasil 👍🏻. Ping to @visr because of #4537 which can be closed if this works.

### DIFF
--- a/L/Libtiff/build_tarballs.jl
+++ b/L/Libtiff/build_tarballs.jl
@@ -32,6 +32,7 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("JpegTurbo_jll"),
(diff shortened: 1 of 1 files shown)


giordano:
Yeah, you may want to rebuild LERC_jll before this one

evetion:
@giordano Could you restart the CI here? I'd expect it will pick up LERC v3.0.0+1 automatically now, we don't want to pin that here.

giordano:
Not before https://github.com/JuliaRegistries/General/pull/55890 is merged.  Also, we can't just restart CI, you have to make a dummy edit to the file to actually retrigger the build, if nothing changes we reuse the cache from the previous build.

evetion:
Ok, will do so.

PR #4539, opened 2022-03-03
Also build LERC for experimental platforms

evetion:


evetion:
@giordano If we merge this (removed one blank line), will it overwrite the existing JLL (version)? Or do we need to patch the version number every so slightly here?

giordano:
It'll be v3.0.0+1

PR #4073, opened 2021-12-17
MPI: Experimental platform tag

vchuravy:
Experiment with https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1128 and https://github.com/JuliaPackaging/JLLWrappers.jl/pull/35

PR #4540, opened 2022-03-03
Add MPI platform tag and have LAMMPS use it

vchuravy:
Based off https://github.com/JuliaParallel/MPI.jl/pull/541 (but also independent enough that we can use it before that lands)

Replaces https://github.com/JuliaPackaging/Yggdrasil/pull/4073

### DIFF
--- a/platforms/mpi.jl
+++ b/platforms/mpi.jl
@@ -1,0 +1,59 @@
+module MPI
+
+const tag_name = "mpi"
+
+const augment = """
+    # Can't use Preferences since we might be running this very early with a non-existing Manifest
+    MPIPreferences_UUID = Base.UUID("3da0fdf6-3ccc-4f1b-acd9-58baa6c99267")
+    const preferences = Base.get_preferences(MPIPreferences_UUID)
+
+    # Keep logic in sync with MPIPreferences.jl
+    const binary = get(preferences, "binary", Sys.iswindows() ? "MicrosoftMPI_jll" : "MPICH_jll")
+
+    const abi = if binary == "system"
+        get(preferences, "abi")
+    elseif binary == "MicrosoftMPI_jll"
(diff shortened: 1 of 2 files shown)


platforms/mpi.jl
```diff
@@ -0,0 +1,59 @@
...
+"""
+
+using BinaryBuilder, Pkg
+using Base.BinaryPlatforms
+
+mpi_abis = (
+    ("MPICH", PackageSpec(name="MPICH_jll"), "", !Sys.iswindows) ,
+    ("OpenMPI", PackageSpec(name="OpenMPI_jll"), "", !Sys.iswindows),
+    ("MicrosoftMPI", PackageSpec(name="MicrosoftMPI_jll"), "", Sys.iswindows),
+    ("MPIwrapper", PackageSpec(name="MPItrampoline_jll"), "3", !Sys.iswindows)
```
vchuravy, line 42:
```suggestion
    ("MPIwrapper", PackageSpec(name="MPItrampoline_jll"), "", !Sys.iswindows)
```

vchuravy:
@giordano and @staticfloat this reminded me that we don't record the compat of JLL packages against dependencies all that much. Here I wanted to make sure that I honor the major version compat of `MPITrampoline`.
Should we not autogenerate a compat for major version for most JLL?

giordano:
> Should we not autogenerate a compat for major version for most JLL?

It's complicated.  We generally follow upstream version numbers, but they do not adhere to semver: they can introduce breaking changes in patch or minor versions (MbedTLS anyone?), or bump the major version just for the sake of it.  This means that using compat bounds can both be too restrictive or not be helpful in preventing breakage.  Basically, it's a mess.  In #3024 I'm keeping a list of packages for which we know we really should specify compat bounds

PR #4541, opened 2022-03-03
[PlatformTag: MPI] Load preferences at runtime for now

vchuravy:
[skip build]


vchuravy:
ah silly me I wanted to bump LAMMPS as well...

PR #4544, opened 2022-03-04
LoopTools update to version 2.16

fkguo:


giordano:
Quoting from https://github.com/JuliaPackaging/Yggdrasil/pull/4545#issuecomment-1059357908

> side note, I strongly recommend not opening pull requests from the same branch as the target one. Your `master` branch should always be in sync with the upstream `master` branch and you'll create branches out of it when you want to open a new pull request, otherwise history will be messy like the one we can see here.

L/LoopTools/bundled/wrappers/looptools.h
```diff
@@ -0,0 +1,303 @@
+#if 0
```
giordano, line 1: What are these files in `bundled/wrappers/` needed for?  They're unused.

giordano (commented):
Ok, we're getting close but there is still this error on aarch64-apple-darwin (which uses a more recent gfortran version):
```
[20:18:49] Cget.F:181:14:
[20:18:49] 
[20:18:49]   181 |         ArgQuad G(2,2)
[20:18:49]       |                     1
[20:18:49] Error: Old-style type declaration REAL*16 not supported at (1)
[20:18:49] Cget.F:182:17:
[20:18:49] 
[20:18:49]   182 |         ComplexQuad bsum(0:2), b1sum(0:2), b00sum(0:2), b11sum(0:2)
[20:18:49]       |                        1
[20:18:49] Error: Old-style type declaration COMPLEX*32 not supported at (1)
[20:18:49] Cget.F:183:17:
[20:18:49] 
[20:18:49]   183 |         ComplexQuad in(0:2,2)
[20:18:49]       |                        1
```

L/LoopTools/bundled/wrappers/looptools.h
fkguo: > What are these files in `bundled/wrappers/` needed for? They're unused.

Thank you very much! I'm pretty sorry for messing up the things. they should be from some experiments I did when I first tried to wrapping over this package... please feel free to remove them.

giordano: Ok, no worries, I removed them.

L/LoopTools/bundled/patches/simplify-configure.patch
```diff
@@ -58,18 +61,14 @@
...
-@@ -278,27 +259,14 @@
+@@ -276,29 +257,15 @@
  
- echo -n "does $CONF_FC support REAL*$CONF_QUADSIZE... " 1>&3
+ ## does f77 support REAL*16?
  
--rm -fr $test*
+-echo -n "does $CONF_FC support REAL*$CONF_QUADSIZE... " 1>&3
 +if [[ "${target}" == arm-* ]]; then
 +    # On 32-bit the Fortran compiler doesn't support quadsize reals
```
giordano, line 73: For the record, the error we get for `aarch64-apple-darwin` is the same as the one we'd get for 32-bit ARM platforms without this setting, so including `aarch64-apple-darwin` here would do the trick.  However I'm wondering if updating the GCC toolchain for this platform would solve the issue, which is something I'd like to try anyway in the coming days.

giordano: Updating the GCC toolchain for aarch64-apple-darwin fixed it, yay :partying_face:

L/LoopTools/bundled/wrappers/looptools.h
fkguo: Hallo, I still failed to make it work on my Mac M1. The error is `Library not loaded: @rpath/libquadmath.0.dylib `, using Julia for macOS ARM. It worked for macOS x86 via Rosetta. Any idea making it work? Which version of gcc should be installed? Thank you for your efforts!

giordano: What version of Julia?

fkguo: I tried both macOS ARM (M-series Processor) [64-bit](https://julialang-s3.julialang.org/bin/mac/aarch64/1.7/julia-1.7.2-macaarch64.dmg) (experimental) and [julia-1.8.0-beta1-full.tar.gz](https://github.com/JuliaLang/julia/releases/download/v1.8.0-beta1/julia-1.8.0-beta1-full.tar.gz). LoopTools_jll can be added and precompiled. The error arose if I `using LoopTools_jll`

fkguo: problem solved by installing gfortran from [gfortran 12-experimental for ARM, macOS 12 (Monterey)](https://github.com/fxcoudert/gfortran-for-macOS/releases/tag/12-arm-alpha) and `ln -s /usr/local/gfortran/lib/libquadmath.0.dylib /usr/local/lib/libquadmath.0.dylib`

PR #4545, opened 2022-03-04
[VMEC] v1.1.0

benjaminfaber:
NetCDF dependence has been removed


giordano:
@benjaminfaber side note, I strongly recommend _**not**_ opening pull requests from the same branch as the target one.  Your `master` branch should always be in sync with the upstream `master` branch and you'll create branches out of it when you want to open a new pull request, otherwise history will be messy like the one we can see here.

benjaminfaber:
Thanks @giordano , I'll branch that out now and make sure my master is always in sync.

PR #4547, opened 2022-03-05
[Musl] Add patch from Alpine Linux to use `qsort_r`

giordano:
Currently in our environment we have

```console
% julia --compile=min -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("x86_64", "linux"; libc="musl"))'
sandbox:${WORKSPACE} # apk add binutils
fetch https://dl-cdn.alpinelinux.org/alpine/v3.15/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.15/community/x86_64/APKINDEX.tar.gz
(1/1) Installing binutils (2.37-r3)
Executing busybox-1.34.1-r3.trigger
OK: 294 MiB in 118 packages
sandbox:${WORKSPACE} # /usr/bin/readelf
Error relocating /usr/lib/libctf-nobfd.so.0: qsort_r: symbol not found
```

because Alpine Linux basically forked Musl in an ABI-incompatible way and some
binaries provided by Alpine Linux reference the Alpine-specific `qsort_r`
symbol, so we have to apply their patch in order to be able to use their
binaries.

### DIFF
--- a/M/Musl/build_tarballs.jl
+++ b/M/Musl/build_tarballs.jl
@@ -7,10 +7,14 @@ version = v"1.2.2"
 sources = [
     ArchiveSource("https://www.musl-libc.org/releases/musl-$(version).tar.gz",
                   "9b969322012d796dc23dda27a35866034fa67d8fb67e0e2c45c913c3d43219dd"),
+    DirectorySource("./bundled"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
+cd ${WORKSPACE}/srcdir/musl-*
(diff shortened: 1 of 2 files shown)


giordano:
With this build:
```console
% nm x86_64-linux-musl/lib/ld-musl-x86_64.so.1|grep qsort
0000000000054251 T qsort
0000000000053f6a t __qsort_r
0000000000053f6a W qsort_r
```
which looks better.

PR #4551, opened 2022-03-05
[0_RootFS] Update RootFS and GCC version for aarch64-apple-darwin

giordano:
The new build of Musl should be ABI-compatible with the version used in Alpine
Linux, see #4547.

Companion PR to https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/228.


giordano:
# IMPORTANT

The new build of GCC has a couple of ABI incompatibilities compared to the previous one:

1. as reported by https://github.com/JuliaLang/julia/issues/44435, `libgcc_s` is now called `libgcc_s.1.1.dylib` instead of `libgcc_s.2.dylib`.  This should have little impact overall, most libraries shouldn't link directly to `libgcc_s`
2. more importantly, this build contains libquadmath for aarch64-apple-darwin.  This means most/all Fortran libraries built with this new compiler will link to it, but libquadmath is currently not part of `CompilerSupportLibraries_jll` for this platform, so they'll all fail to load.  In any case we'll update `CompilerSupportLibraries_jll` in Julia to include the new compiler libraries and we'll backport also the change to the `release-1.7` (less important since Julia is mostly broken here anyway) and `release-1.8` branches , but we'll probably see into the wild some errors like
   ```
   ERROR: LoadError: InitError: could not load library "/path/to/libray.dylib"
   dlopen(/path/to/libray.dylib, 1): Library not loaded: @rpath/libquadmath.0.dylib
     Referenced from: /path/to/libray.dylib
     Reason: image not found
   ```

Now the good news: after placing the new `libquadmath.0.dylib` among other compiler libraries I was able to successfully run the tests of [`SHTOOLS.jl`](https://github.com/eschnett/SHTOOLS.jl) and [`BipolarSphericalHarmonics.jl`](https://github.com/jishnub/BipolarSphericalHarmonics.jl) with a new build of [`SHTOOLS_jll.jl`](https://github.com/JuliaBinaryWrappers/SHTOOLS_jll.jl) compiled with the new toolchain, and also the LinearAlgebra test suite with a new build of OpenBLAS 0.3.20.

PR #4554, opened 2022-03-06
Build Torch from source

stemann:
Alternative to #4477, as that bumped into https://github.com/FluxML/Torch.jl/issues/17

Building Torch from source (for Linux and Linux with CUDA) would likely remedy (at least):
* https://github.com/FluxML/Torch.jl/issues/17
* https://github.com/FluxML/Torch.jl/issues/20
* https://github.com/FluxML/Torch.jl/issues/42

Related aims (but not the aim of this PR):
* Updating Torch version.
* Windows support.
* Windows + CUDA support.
* macOS support.

### DIFF
--- a/T/Torch/build_tarballs.jl
+++ b/T/Torch/build_tarballs.jl
@@ -8,7 +8,9 @@ version = v"1.10.2"
 # Collection of sources required to complete build
 sources = [
     GitSource("https://github.com/pytorch/pytorch.git", "71f889c7d265b9636b93ede9d651c0a9c4bee191"),
-    FileSource("https://micromamba.snakepit.net/api/micromamba/linux-64/0.21.1", "c907423887b43bec4e8b24f17471262c8087b7095683f41dcef4a4e24e9a3bbd"; filename = "micromamba.tar.bz2")
+    FileSource("https://micromamba.snakepit.net/api/micromamba/linux-64/0.21.1", "c907423887b43bec4e8b24f17471262c8087b7095683f41dcef4a4e24e9a3bbd"; filename = "micromamba.tar.bz2"),
+    ArchiveSource("https://github.com/JuliaBinaryWrappers/CUDA_full_jll.jl/releases/download/CUDA_full-v10.2.89%2B5/CUDA_full.v10.2.89.x86_64-linux-gnu.tar.gz", "60e6f614db3b66d955b7e6aa02406765e874ff475c69e2b4a04eb95ba65e4f3b"; unpack_target = "CUDA_full.v10.2"),
+    ArchiveSource("https://github.com/JuliaBinaryWrappers/CUDA_full_jll.jl/releases/download/CUDA_full-v11.3.1%2B1/CUDA_full.v11.3.1.x86_64-linux-gnu.tar.gz", "9ae00d36d39b04e8e99ace63641254c93a931dcf4ac24c8eddcdfd4625ab57d6"; unpack_target = "CUDA_full.v11.3"),
     DirectorySource("./bundled"),
 ]
 
@@ -30,6 +32,7 @@ cd $WORKSPACE/srcdir/pytorch
 atomic_patch -p1 ../patches/pytorch-aten-qnnpack-cmake-windows.patch
 
 cmake_extra_args=""
+include_paths=""
 
 if [[ $bb_full_target == *cxx11* ]]; then
     cmake_extra_args+="-DGLIBCXX_USE_CXX11_ABI=1 "
@@ -99,6 +102,39 @@ if [[ $target == x86_64-apple-darwin* # Fails to compile: /workspace/srcdir/pyto
     cmake_extra_args+="-DUSE_MKLDNN=OFF "
 fi
 
+if [[ $bb_full_target == *cuda* ]]; then
+    cuda_version=`echo $bb_full_target | sed -E -e 's/.*cuda\+([0-9]+\.[0-9]+).*/\1/'`
+    cuda_version_major=`echo $cuda_version | cut -d . -f 1`
+    cuda_version_minor=`echo $cuda_version | cut -d . -f 2`
+    cuda_full_path="$WORKSPACE/srcdir/CUDA_full.v$cuda_version/cuda"
+    export PATH=$PATH:$cuda_full_path/bin
+    export CUDACXX=$cuda_full_path/bin/nvcc
+    export CUDAHOSTCXX=$CXX
+    mkdir $WORKSPACE/tmpdir
+    export TMPDIR=$WORKSPACE/tmpdir
+    cmake_extra_args+="\
+        -DUSE_CUDA=ON \
+        -DUSE_CUDNN=ON \
+        -DUSE_MAGMA=ON \
+        -DCUDA_TOOLKIT_ROOT_DIR=$cuda_full_path \
+        -DCUDA_CUDART_LIBRARY=$cuda_full_path/lib64/libcudart.$dlext \
+        -DCMAKE_CUDA_FLAGS='-cudart shared' \
+        -DCUDA_cublas_LIBRARY=$cuda_full_path/lib64/libcublas.$dlext \
+        -DCUDA_cufft_LIBRARY=$cuda_full_path/lib64/libcufft.$dlext \
+        -DCUDA_curand_LIBRARY=$cuda_full_path/lib64/libcurand.$dlext \
+        -DCUDA_cusolver_LIBRARY=$cuda_full_path/lib64/libcusolver.$dlext \
+        -DCUDA_cusparse_LIBRARY=$cuda_full_path/lib64/libcusparse.$dlext \
+        -DCUDA_TOOLKIT_INCLUDE=$includedir;$cuda_full_path/include \
+        -DCUB_INCLUDE_DIR=$WORKSPACE/srcdir/pytorch/third_party/cub "
+    include_paths+=":$cuda_full_path/include"
+    micromamba install -y magma-cuda${cuda_version_major}${cuda_version_minor} -c pytorch
+    git submodule update --init \
+        third_party/cub \
+        third_party/cudnn_frontend
+else
+    cmake_extra_args+="-DUSE_CUDA=OFF -DUSE_MAGMA=OFF "
+fi
+
 git submodule update --init \
     third_party/FP16 \
     third_party/FXdiv \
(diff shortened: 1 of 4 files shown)


stemann:
@giordano Torch seems to take a long time to build. It might just be a nightmare of C++ code to compile, but should I be worried about Torch jumping to some conclusions about (a primitive) ccache?
```
[23:07:08] -- Primitive cache is enabled
```
```
[23:07:16] --   Using ccache if found : ON
[23:07:16] --   Found ccache          : /usr/bin/ccache
```
https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=18082&view=logs&j=43fac7a0-bc66-55fc-9ec4-c35efc5fae1d&t=6d20a0a5-d503-509b-b3f1-360add906995&l=1209

stemann:
> @giordano Torch seems to take a long time to build. It might just be a nightmare of C++ code to compile, but should I be worried about Torch jumping to some conclusions about (a primitive) ccache?
> 
> ```
> [23:07:08] -- Primitive cache is enabled
> ```
> 
>  ```
> [23:07:16] --   Using ccache if found : ON
> [23:07:16] --   Found ccache          : /usr/bin/ccache
> ```
> 
> https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=18082&view=logs&j=43fac7a0-bc66-55fc-9ec4-c35efc5fae1d&t=6d20a0a5-d503-509b-b3f1-360add906995&l=1209

Never mind. I somehow forgot to enable parallel build.

T/Torch/Torch/build_tarballs.jl
```diff
@@ -0,0 +1,111 @@
...
+
+cd $WORKSPACE/srcdir/pytorch
+
+if [[ $bb_full_target == *cuda* ]]; then
+    cuda_version=`echo $bb_full_target | sed -E -e 's/.*cuda\+([0-9]+\.[0-9]+).*/\1/'`
+    cuda_version_major=`echo $cuda_version | cut -d . -f 1`
+    cuda_version_minor=`echo $cuda_version | cut -d . -f 2`
+    cuda_full_path="$WORKSPACE/srcdir/CUDA_full.v$cuda_version/cuda"
+    apk del cmake
+    apk add 'cmake<3.17' --repository=http://dl-cdn.alpinelinux.org/alpine/v3.11/main
```
giordano, line 37: Wait, what?

stemann: I got fed up trying to hack the "find cuda" part of Torch - downgrading cmake seems to work.

Torch seems to be using the "old" https://cmake.org/cmake/help/latest/module/FindCUDA.html approach which was replaced with https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html in CMake v3.17 as I understand it.

giordano: Sigh

stemann: At least it seems more "clean" to me than patching some set of cmake files :-)

stemann: ... and then it works in the BB shell (e.g. with `--debug=begin`), but not in auto mode...

giordano: > ... and then it works in the BB shell (e.g. with --debug=begin), but not in auto mode...

I'm not 100% sure what you mean here, but note that package installation and setting environment variables aren't persistent at the moment, when dropping into the debug shell (package are installed in a tmpfs which is lost when recreating the debug environment, and we don't have a way to remember all environment settings). It may be possible to address some of these issues in the future, but at the moment you have to repeat those operations manually

T/Torch/Torch/build_tarballs.jl
```diff
@@ -0,0 +1,271 @@
...
+
+name = "Torch"
+version = v"1.10.2"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/pytorch/pytorch.git", "71f889c7d265b9636b93ede9d651c0a9c4bee191"),
+    FileSource("https://micromamba.snakepit.net/api/micromamba/linux-64/0.21.1", "c907423887b43bec4e8b24f17471262c8087b7095683f41dcef4a4e24e9a3bbd"; filename = "micromamba.tar.bz2"),
+    ArchiveSource("https://github.com/JuliaBinaryWrappers/CUDA_full_jll.jl/releases/download/CUDA_full-v10.2.89%2B5/CUDA_full.v10.2.89.x86_64-linux-gnu.tar.gz", "60e6f614db3b66d955b7e6aa02406765e874ff475c69e2b4a04eb95ba65e4f3b"; unpack_target = "CUDA_full.v10.2"),
+    ArchiveSource("https://github.com/JuliaBinaryWrappers/CUDA_full_jll.jl/releases/download/CUDA_full-v11.3.1%2B0/CUDA_full.v11.3.1.x86_64-linux-gnu.tar.gz", "4094b8f1a3667166c1281faa7958cc46dbdc86ac86979e86d462a3c58f2a4b54"; unpack_target = "CUDA_full.v11.3"),
```
giordano, line 13: Shouldn't these be regular dependencies? Maybe build only?

T/Torch/Torch/build_tarballs.jl
```diff
@@ -0,0 +1,271 @@
...
+        -Wno-dev \
+        $cmake_extra_args \
+        ..
+}
+if [[ $bb_full_target != *cuda* ]]; then
+    configure
+else
+    configure
+    configure || configure
+fi
```
giordano, line 201: 😭

T/Torch/Torch/build_tarballs.jl
stemann: They should be build dependecies, yes.

I believe there's an issue with specifying the version - I can't get one BuildDependency version for CUDA 10.2 and one for 11.3, I believe (I can't have two dependencies with the same name). It might be possible to just say `BuildDependency("CUDA_full_jll")` and get whatever version is most recent at the time (if the augmented platform implementation knows how to get CUDA_full 10.2 for platform with `cuda="10.2"`?).

I also figured out that the released libtorch binaries are built against a few more CUDA versions, so there might be some more platforms to add: https://github.com/pytorch/pytorch/runs/5467280928

T/Torch/Torch/build_tarballs.jl
stemann: Indeed!  🤦 😆

T/Torch/Torch/build_tarballs.jl
giordano: You're right you can't have together in the same environment, but you can limit a dependency to a specific set of platforms (which you're already doing for other dependencies)

T/Torch/Torch/build_tarballs.jl
```diff
@@ -0,0 +1,271 @@
...
+
+if [[ $bb_full_target == *cuda* ]]; then
+    cuda_version=`echo $bb_full_target | sed -E -e 's/.*cuda\+([0-9]+\.[0-9]+).*/\1/'`
+    cuda_version_major=`echo $cuda_version | cut -d . -f 1`
+    cuda_version_minor=`echo $cuda_version | cut -d . -f 2`
+    cuda_full_path="$WORKSPACE/srcdir/CUDA_full.v$cuda_version/cuda"
+    apk del cmake
+    apk add 'cmake<3.17' --repository=http://dl-cdn.alpinelinux.org/alpine/v3.11/main
+    export PATH=$PATH:$cuda_full_path/bin
+    export CUDACXX="ccache nvcc"
```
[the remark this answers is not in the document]
giordano: Uhm, [ccache documentation](https://ccache.dev/platform-compiler-language-support.html) does mention nvcc, but I have no idea how well that works in practice, never used it.  Also, it depends on what the compiler calls look like, ccache doesn't cache everything, most notably linking.

T/Torch/Torch/build_tarballs.jl
```diff
@@ -0,0 +1,267 @@
...
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
+    Dependency("blis_jll"; platforms = blis_platforms),
+    Dependency("CPUInfo_jll"),
+    Dependency("CUDNN_jll"; platforms = cuda_platforms),
+    Dependency("Gloo_jll"; platforms = filter(p -> nbits(p) == 64, platforms)),
+    Dependency("LAPACK_jll"; platforms = openblas_platforms),
+    Dependency("MKL_jll"; platforms = mkl_platforms),
+    Dependency("MKL_Headers_jll"; platforms = mkl_platforms),
```
giordano, line 254: Headers are build dependencies :slightly_smiling_face: 
```suggestion
    BuildDependency("MKL_Headers_jll"; platforms = mkl_platforms),
```

Wimmerer:
What's the status on this? Can we include: https://github.com/LaurentMazare/tch-rs/tree/main/torch-sys/libtch to generate C bindings? 

Not having Windows support here doesn't seem like the end of the world.

stemann:
> What's the status on this? Can we include: https://github.com/LaurentMazare/tch-rs/tree/main/torch-sys/libtch to generate C bindings?
> 
> Not having Windows support here doesn't seem like the end of the world.

Will try to finish this ASAP.

It would be easiest to exclude libtch from Torch jll - it’s already a bit complicated as it is - with the dependencies.

But yes, next step would be to include an updated libtch. And to clarify which version of libtch with which modifications were used originally.

Wimmerer:
And just build it in a separate libtch_jll? That seems good.

Wimmerer:
Bump?

stemann:
> Bump?

Indeed. Just need to clean up the mess - CUDA hack(s) no longer necessary.

T/Torch/Torch/build_tarballs.jl
stemann: Giving it a shot in 6662c0877138ebe167f0154179b7dc043cd7c3f5

T/Torch/Torch/build_tarballs.jl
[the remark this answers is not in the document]
stemann: Giving up on ccache for nvcc for the time being.

stemann:
> > What's the status on this? Can we include: https://github.com/LaurentMazare/tch-rs/tree/main/torch-sys/libtch to generate C bindings?
> > Not having Windows support here doesn't seem like the end of the world.
> 
> Will try to finish this ASAP.
> 
> And yes - Windows support could be left out.
> 
> It would be easiest to exclude libtch from Torch jll - it’s already a bit complicated as it is - with the dependencies.
> 
> But yes, next step would be to include an updated libtch. And to clarify which version of libtch with which modifications were used originally.

I've
* Rebased on to master
* Undid the renaming-package part (Torch/TorchCUDA)
* Undone the hack for CUDA v11.3 missing Thrust
* Replaced CUDA_full sources with build dependencies
* Re-enabled C++ string ABI expansion

To do:
* Verify versions of dependencies (that they are the ones used by Torch v1.10.2)

T/Torch/Torch/build_tarballs.jl
stemann: Did not work:
```
ERROR: LoadError: it is invalid to specify multiple packages with the same name: `CUDA_full_jll [4f82f1eb]`
```
https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=21415&view=logs&j=f6c99a09-f697-57d4-7332-32215efd2007&t=239917b9-1794-551a-0432-a720bc1732c1&l=79

With deps:
```julia
    BuildDependency(PackageSpec("CUDA_full_jll", Base.UUID("4f82f1eb-248c-5f56-a42e-99106d144614"), v"10.2.89"); platforms = filter(p -> p["cuda"] == "10.2", cuda_platforms)),
    BuildDependency(PackageSpec("CUDA_full_jll", Base.UUID("4f82f1eb-248c-5f56-a42e-99106d144614"), v"11.3.1"); platforms = filter(p -> p["cuda"] == "11.3", cuda_platforms)),
```

T/Torch/build_tarballs.jl
```diff
@@ -3,60 +3,263 @@
...
+    Dependency("LAPACK_jll"; platforms = openblas_platforms),
+    Dependency("MKL_jll"; platforms = mkl_platforms),
+    BuildDependency("MKL_Headers_jll"; platforms = mkl_platforms),
+    Dependency("OpenBLAS_jll"; platforms = openblas_platforms),
+    Dependency("PThreadPool_jll", v"0.0.20210414"),
+    Dependency("SLEEF_jll", v"3.5.2"),
+#    Dependency("TensorRT_jll"; platforms = cuda_platforms), # Building with TensorRT is not supported: https://github.com/pytorch/pytorch/issues/60228
+    Dependency("XNNPACK_jll", v"0.0.20210622"),
+    BuildDependency(PackageSpec("protoc_jll", Base.UUID("c7845625-083e-5bbe-8504-b32d602b7110"), v"3.13.0")),
+    HostBuildDependency(PackageSpec("protoc_jll", Base.UUID("c7845625-083e-5bbe-8504-b32d602b7110"), v"3.13.0")),
```
giordano, line 259: When you specify the version of a runtime dependency you typically also want to specify the compat (or actually only that one since the build version is automatically inferred from the lowest compatible version)

T/Torch/deps.md
```diff
@@ -0,0 +1,76 @@
+# Dependencies
```
giordano, line 1: What is this used for?  Apart from being a fancy dependency visualisation, that is

stemann: That is for maintenance purposes: I have had good use of that visualisation in order to grasp all the dependencies - in order to not constantly check git submodules and cmake dependencies in cpuinfo, pthreadpool, xnnpack and PyTorch - there are even a few that's not visualised...

giordano:
Is this good to go now?

stemann:
> Is this good to go now?

Yes, LGTM :-)

I plan to follow-up with a PR with a recipe for building an updated version of the C wrapper in https://github.com/FluxML/Torch.jl/tree/master/build

Wimmerer:
> I plan to follow-up with a PR with a recipe for building an updated version of the C wrapper in

Awesome I was just looking at whether to do the ocaml or Rust one today.

PR #4560, opened 2022-03-06
Updated CPUInfo to v0.0.20200612

stemann:


C/CPUInfo/build_tarballs.jl
```diff
@@ -3,21 +3,17 @@
...
+    GitSource("https://github.com/pytorch/cpuinfo.git", "63b254577ed77a8004a9be6ac707f3dccc4e1fd9"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir
 cd cpuinfo
-if [[ $target == *-w64-mingw32 ]]; then
-    atomic_patch -p1 ../patches/lowercase-windows-include.patch
-fi
```
giordano, line 20: So in these 5 months they learnt how to install all the libraries?

stemann: He he - not quite - they got their capitalization fixed, but the DLL must still be copied manually :-)

giordano: Oh, right, sorry, I saw the mingw32 `if` and thought it was the other one without even reading

PR #4555, opened 2022-03-06
New Recipe: CPUInfo v0.0.20200122

stemann:
Packaging similar to [libcpuinfo0](https://packages.ubuntu.com/impish/libcpuinfo0) in Ubuntu.

Starting with an old version - used by PyTorch v1.5.

PR #4561, opened 2022-03-06
Added FXdiv v0.0.20181117

stemann:


F/FXdiv/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+    -DFXDIV_BUILD_BENCHMARKS=OFF \
+    ..
+cmake --build . -- -j $nproc
+make install
+install_license ../LICENSE
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
```
giordano, line 33: If this is header-only it should be
```suggestion
platforms = [AnyPlatform()]
```

F/FXdiv/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+
+# The products that we will ensure are always built
+products = [
```
giordano, line 36:
```suggestion
products = Product[
```

F/FXdiv/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/Maratyszcza/FXdiv.git", "b742d1143724d646cd0f914646f1240eacf5bd73"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd FXdiv
```
giordano, line 16:
```suggestion
cd $WORKSPACE/srcdir/FXdiv
```

F/FXdiv/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+cmake \
+    -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DFXDIV_BUILD_TESTS=OFF \
+    -DFXDIV_BUILD_BENCHMARKS=OFF \
+    ..
+cmake --build . -- -j $nproc
+make install
+install_license ../LICENSE
```
giordano, line 28: This doesn't hurt, but it should be installed automatically.

F/FXdiv/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [AnyPlatform()]
+
+# The products that we will ensure are always built
+products = Product[
```
giordano, line 36: If you want, you can have the header file as product
```suggestion
products = [
    FileProduct("include/fxdiv.h", :fxdiv_h)
```

F/FXdiv/build_tarballs.jl
```diff
@@ -0,0 +1,36 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/Maratyszcza/FXdiv.git", "b742d1143724d646cd0f914646f1240eacf5bd73"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/FXdiv
+mkdir $includedir
+cp -a include/fxdiv.h $includedir
```
giordano, line 17:
```suggestion
install -Dvm 644 include/fxdiv.h "${includedir}/fxdiv.h"
```

giordano: Never call `cp` in this way ([second time today](https://github.com/JuliaPackaging/Yggdrasil/pull/4555#r820271910) I think?).  The problem is that if the second argument doesn't exist, you'll create a file with that name, instead of inside that directory.  Yes, here you created the directory at the line above, but it's easy to forget to create it and this happens continuously here, so a good rule of thumb is to just don't do that.  Either do something like
```sh
cp include/fxdiv.h $includedir/
cp include/fxdiv.h $includedir/.
cp include/fxdiv.h $includedir/fxdiv.h
```
or just use `install -D` which does the right thing and also automatically create the parent directories as needed.

stemann: Right - thanks for the explanation - after the first it wasn't completely clear why `cp -a` was a bad thing. But yes - I've seen my share of mishaps like that as well (which is why I usually remember to get the directories in place :-)

But nice to know that `install -D` can help out.

PR #4566, opened 2022-03-07
Updated PThreadPool to 20200302

stemann:


P/PThreadPool/build_tarballs.jl
```diff
@@ -27,13 +27,15 @@ cmake \
     ..
 cmake --build . -- -j $nproc
 make install
+if [[ $target == *-w64-mingw32 ]]; then
+    install -Dvm 755 libpthreadpool.dll "${libdir}/libpthreadpool.dll"
+fi
```
giordano, line 32: At some point I'd like to understand why we have to do this everywhere.

PR #4569, opened 2022-03-07
Added SLEEF v3.4.0

stemann:


S/SLEEF/build_tarballs.jl
```diff
@@ -0,0 +1,70 @@
...
+    LibraryProduct("libsleefquad", :libsleefquad),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
+    Dependency("MPFR_jll"),
+    Dependency("OpenSSL_jll"),
+    HostBuildDependency("MPFR_jll"),
+    HostBuildDependency("OpenSSL_jll"),
```
giordano, line 64: https://sleef.org/compile.xhtml

>  In order to build SLEEF, you need CMake, which is an open-source and cross-platform building tool. In order to test the library, it is better to have the GNU MPFR Library, Libssl and FFTW. 

Do you really need MPFR and OpenSSL just for building?

S/SLEEF/build_tarballs.jl
```diff
@@ -0,0 +1,70 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/sleef
+mkdir build-native
+cd build-native
+PATH=`echo $PATH | cut -d":" -f 3-` \
+    CC=$HOSTCC \
+    CXX=$HOSTCXX \
```
giordano, line 20: You may want to use `-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}`

stemann: Indeed.

S/SLEEF/build_tarballs.jl
stemann: Probably not...

S/SLEEF/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+    LibraryProduct("libsleefdft", :libsleefdft),
+    LibraryProduct("libsleefgnuabi", :libsleefgnuabi),
+    LibraryProduct("libsleefquad", :libsleefquad),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
+    Dependency("MPFR_jll"),
+    Dependency("OpenSSL_jll"),
```
giordano, line 60: These ones too, no?

S/SLEEF/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/sleef
+mkdir build-native
+cd build-native
+cmake \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_HOST_TOOLCHAIN} \
+    -G Ninja \
+    -DBUILD_QUAD=TRUE \
+    ..
```
giordano, line 22: Why don't you disable tests also here?

stemann: Good point - it was a bit random at that point.

giordano:
The native build doesn't seem very native, it still tries to compile stuff for the target platform.  This doesn't look right.

stemann:
> The native build doesn't seem very native, it still tries to compile stuff for the target platform. This doesn't look right.

Right. This is the source of approach: https://github.com/shibatch/sleef/issues/249

giordano:
It should be fixed by https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/230

stemann:
> It should be fixed by [JuliaPackaging/BinaryBuilderBase.jl#230](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/230)

Excellent!

stemann:
Builds for `armv6l`/`armv7l` should be possible:
* https://travis-ci.org/github/shibatch/sleef/jobs/525500790
* https://github.com/shibatch/sleef/blob/8df2bce4596ec6b86f357beebf2e428564f0a4ec/travis/before_script.armhf-gcc.sh
* https://github.com/shibatch/sleef/blob/8df2bce4596ec6b86f357beebf2e428564f0a4ec/travis/toolchain-armhf.cmake

... but I am not sure how to avoid the `-march -mtune` additions that block BB:
https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=18039&view=logs&j=79f6dbbf-e64f-5591-3dc6-a23332a2c9a9&t=d890191c-3a32-5e8d-0782-7181b99ee028&l=569

S/SLEEF/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+ninja all
+ninja install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+filter!(p -> !(arch(p) == "armv6l" || arch(p) == "armv7l"), platforms) # armv6l and armv7l builds fail trying to force an architecture with -march
```
giordano, line 52: The solution here is to remove the `-march` flags

stemann: It seems there is some stuff happening in [Configure.cmake#L278](https://github.com/shibatch/sleef/blob/8df2bce4596ec6b86f357beebf2e428564f0a4ec/Configure.cmake#L278) - the patch I've included in https://github.com/JuliaPackaging/Yggdrasil/pull/4569/commits/45ecf4210055740b42d1ab86ceed2ea0531269de **might** fix it...

giordano:
This sounds a bit concerning:
```
┌ Warning: Minimum instruction set detected for lib/libsleefgnuabi.so.3.4 is avx512, not x86_64 as desired.
└ @ BinaryBuilder.Auditor /depot/packages/BinaryBuilder/qggoo/src/Auditor.jl:303
```
Well, I guess we'll see how this works.

PR #4573, opened 2022-03-07
[oneAPI] Add musl support

jpsamaroo:


O/oneAPI_Level_Zero/common.jl
```diff
@@ -6,7 +6,7 @@ api_version = v"1.2.43"
 #
 # The level zero repository contains both the API headers and a loader,
 # which are versioned independently.
-sources = [
+sources = Any[
```
giordano, line 9: Why this?

jpsamaroo: Because we later `push!` an object of a different type, so we need to make sure the array type is wide enough.

giordano: I see.  However I'd prefer to use 
```suggestion
sources = BinaryBuilder.AbstractSource[
```
I'm actually surprised we accept a `Vector{Any}` in `build_tarballs()`

O/oneAPI_Level_Zero/oneAPI_Level_Zero_Loader/build_tarballs.jl
```diff
@@ -2,14 +2,19 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
-include("../common.jl")
+include("../common.jl"); push!(source, DirectorySource("./bundled"))
```
giordano, line 5:
```suggestion
include("../common.jl"); push!(sources, DirectorySource("./bundled"))
```

O/oneAPI_Level_Zero/oneAPI_Level_Zero_Loader/bundled/patches/musl-no-deepbind.patch
```diff
@@ -0,0 +1,25 @@
+From 3a15f35ccae10e0e8d21acb2cdb2790b2955a5be Mon Sep 17 00:00:00 2001
```
giordano, line 1: It'd be great to upstream this patch (I guess it'll need a glibc guard or something like that) :slightly_smiling_face:

jpsamaroo: I'm not actually sure it's correct, I just did it to make sure it compiles :smile: But yeah, I have a large number of musl patches I really should push to various upstreams.

PR #4576, opened 2022-03-07
Changed protoc to build v3.13.0.1

stemann:
Needed for Torch v1.10.2

### DIFF
--- a/P/protoc/build_tarballs.jl
+++ b/P/protoc/build_tarballs.jl
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "protoc"
-version = v"3.13.0"
+version = v"3.13.0" # 3.13.0.1
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/protocolbuffers/protobuf.git", "fde7cf7358ec7cd69e8db9be4f1fa6a5c431386a"),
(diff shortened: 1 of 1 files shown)


P/protoc/build_tarballs.jl
```diff
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "protoc"
-version = v"3.13.0"
+version = v"3.13.0" # 3.13.0.1
```
giordano, line 6: How are you going to distinguish from the previous build?  Spoiler: you can't

stemann: I guess I would just be lucky to get 3.13.0+1 if I select v3.13.0 ? :-) ... and never be able to get v3.13.0...

Not sure what to do about those oddly named versions here...

Torch points to the 3.13.0.1 version (I was a bit fast to submit the other PR...).

giordano: I don't think the [difference between 3.13.0 and 3.13.0.1](https://github.com/protocolbuffers/protobuf/compare/v3.13.0...v3.13.0.1) is really significant for us.

stemann: Right - so let's close this PR?

giordano: Yeah, I don't think there is too much need for this

stemann: OK

PR #4577, opened 2022-03-08
Updated XNNPACK to v0.0.20200323

stemann:
As referenced by Torch v1.6.0 - v1.7.1


stemann:
> The libraries built by all these non-versioned torch dependencies (CPUInfo, PThreadPool, XNNPACK) do not specify an SO version - this patch might be relevant: https://salsa.debian.org/deeplearning-team/xnnpack/-/blob/master/debian/patches/soversion.patch
> 
> ?

Done

stemann:
@giordano This should fix a lot of the non-CUDA issues in #4554

PR #4588, opened 2022-03-09
New Recipe: Emoslib v4.5.9

tcarion:
Adding the binaries for [emoslib](https://confluence.ecmwf.int/display/EMOS/Emoslib).

### DIFF
--- a/L/libemos/build_tarballs.jl
+++ b/L/libemos/build_tarballs.jl
@@ -1,0 +1,37 @@
+using BinaryBuilder
+
+name = "libemos"
+version = v"4.5.9"
+sources = [
+    ArchiveSource("https://confluence.ecmwf.int/download/attachments/3473472/libemos-$version-Source.tar.gz", "e57e02c636dc8f5ccb862a103789b1e927bc985b1e0f2b05abf4f64e86d2f67f"),
+]
+
+script = raw"""
+cd libemos-*-Source
+mkdir build ; cd build
+cmake \
+    -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DENABLE_ECCODES=ON \
+    -DGRIB_API_LIBRARIES=${libdir}/libeccodes.so \
+    -DGRIB_API_INCLUDE_DIRS=${prefix}/include \
+    ..
+make -j${nproc}
+make install
+install_license ../LICENSE
+"""
+
+platforms = supported_platforms()
+
+products = [
+    # LibraryProduct("libemos", :emos),
+    FileProduct("lib/libemos.a", :emos), # libemos is not designed to be build as a shared library (see libemos-$version-Source/CMakeLists.txt:39)
+]
+
+dependencies = [
+    Dependency("eccodes_jll"),
+    Dependency("FFTW_jll"),
+]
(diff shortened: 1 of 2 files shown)


L/libemos/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+make -j${nproc}
+make install
+install_license ../LICENSE
+"""
+
+platforms = supported_platforms()
+
+products = [
+    # LibraryProduct("libemos", :emos),
+    FileProduct("lib/libemos.a", :emos), # libemos is not designed to be build as a shared library (see libemos-$version-Source/CMakeLists.txt:39)
```
giordano, line 29: Hi, thanks for your contribution!  If this can't provide a shared library, how are you planning to use this from Julia?  _If_ you're planning to use this library from Julia :slightly_smiling_face:

tcarion: Hi! I don't plan to use it from Julia, it's just a [dependency for flex_extract](https://www.flexpart.eu/flex_extract/Installation/local.html) for which I intend to build binaries for Julia (hoping that it will work, I'm kind of new to this :slightly_smiling_face:)

giordano:
Unless you have the bandwidth to fix the source code (which looks a bit messy), I think we're going to exclude Musl platforms.  This code shares the same problems mentioned in https://github.com/gperftools/gperftools/issues/693

giordano:
Windows looks completely unsupported.  Last error remained to address is
```
[00:54:56] /workspace/srcdir/libemos-4.5.9-Source/interpolation/sharedlib.c:64:20: error: variable has incomplete type 'struct stat64'
[00:54:56]     struct  stat64 s;
[00:54:56]                    ^
[00:54:56] /workspace/srcdir/libemos-4.5.9-Source/interpolation/sharedlib.c:64:13: note: forward declaration of 'struct stat64'
[00:54:56]     struct  stat64 s;
[00:54:56]             ^
```
seen on aarch64-apple-darwin and FreeBSD

giordano:
On macOS `stat64` is defined under this condition:
```c
#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)

#if !__DARWIN_ONLY_64_BIT_INO_T

struct stat64 __DARWIN_STRUCT_STAT64;

#endif /* !__DARWIN_ONLY_64_BIT_INO_T */

#endif /* (!_POSIX_C_SOURCE || _DARWIN_C_SOURCE) */
```
but for the aarch64 architecture
```console
# echo '#include <sys/stat.h>' | cc -x c -dM -E - | grep __DARWIN_ONLY_64_BIT_INO_T
#define __DARWIN_ONLY_64_BIT_INO_T 1
```
and so this type struct is undefined.  I guess they weren't expecting to work with an aarch64 macOS back then.  I guess there is something similar also for FreeBSD.

giordano:
This is good to go as far as I'm concerned, let me know what you think.  I had to exclude a few platforms because they are basically not supported

tcarion:
This looks perfect, thank you very much for your help !

tcarion:
> Unless you have the bandwidth to fix the source code (which looks a bit messy), I think we're going to exclude Musl platforms. This code shares the same problems mentioned in [gperftools/gperftools#693](https://github.com/gperftools/gperftools/issues/693)

No I really don't :slightly_smiling_face:. It's ok, I don't think it was meant to work on those platforms anyway.

barche:
Looks good to me too, thanks @giordano !

PR #4590, opened 2022-03-10
Updated HOHQMesh to v1.2.1

andrewwinters5000:
cc @sloede

### DIFF
--- a/H/HOHQMesh/build_tarballs.jl
+++ b/H/HOHQMesh/build_tarballs.jl
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "HOHQMesh"
-version = v"1.2.0"
+version = v"1.2.1"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


giordano:
Only successful platforms are:

* `x86_64-linux-gnu-libgfortran4`
* `aarch64-linux-gnu-libgfortran4`
* `powerpc64le-linux-gnu-libgfortran4`
* `armv6l-linux-gnu-libgfortran3`

This is one of the most meaningless sets of successful platforms I've ever seen, I can't see any sensible pattern here (apart from they all glibc).

giordano:
If I restart failed jobs, I get some more successful platforms.  I suspect the build system doesn't handle parallelisation correctly, by which I mean some dependencies may have not been specified correctly in the make rules.

andrewwinters5000:
Okay, I will work on building it locally are suss out what is wrong

andrewwinters5000:
I have a fix but it requires an upstream patch to HOHQMesh first.

giordano:
You can also patch the source code here.  If you look for `atomic_patch` in this repository you'll find plenty of examples, alternatively if you can post the patch here I can update it later today.

giordano:
Much better now :slightly_smiling_face:  Thanks!

PR #4593, opened 2022-03-10
[VegaFusion] Bump to version 0.1.0

jeremiahpslewis:
Blocked, requires rust v1.59

### DIFF
--- a/V/VegaFusion/build_tarballs.jl
+++ b/V/VegaFusion/build_tarballs.jl
@@ -1,11 +1,11 @@
 using BinaryBuilder
 
 name = "VegaFusion"
-version = v"0.0.4"
+version = v"0.1.0"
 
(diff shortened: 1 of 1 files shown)


jeremiahpslewis:
@giordano Clearly more than a bit more complicated than updating the rust version number and checking that the tests pass. Looked over the PRs to try and see if I could follow the changes you made. Wow. Thank you!

giordano:
The reason why we can't simply build the RootFS stuff here is because it requires changes in multiple repositories.  If it was only that we could perhaps automate it, but we also need to test everything _**before**_ merging the pull request here: it happens almost every time that we need to tweak something in the build system of our toolchains (although Rust and Go are relatively stable now), so lots of manual intervention is always necessary.

PR #4604, opened 2022-03-12
Updated TensorRT to v8.2.2

stemann:
Awaiting #4347, #4602, and #4603.


ViralBShah:
Maybe #4605 is the latest?

PR #2938, opened 2021-05-05
[SSGraphBLAS] v5.0 - Don't Merge

Wimmerer:
New version, force compat to 1.6?

PR #4606, opened 2022-03-12
Wizard recipe: MAGEMin-v1.0.1

boriskaus:

* Package name: MAGEMin
* Version: v1.0.0


M/MAGEMin/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "MAGEMin"
+version = v"1.0.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/ComputationalThermodynamics/MAGEMin.git", "032116744f3df05b5a2999c2eb39689c831cf4c7"),
+    ArchiveSource("https://github.com/Reference-LAPACK/lapack/archive/refs/tags/v3.10.0.tar.gz", "328c1bea493a32cac5257d84157dc686cc3ab0b004e2bea22044e0a59f6f8a19"),
```
giordano, line 11: In any case this can't go here and it must be a separate package.  But we already have [`LAPACK`](https://github.com/JuliaPackaging/Yggdrasil/blob/bb4407e913504e31c92151a4d1a95b98e9c6b113/L/LAPACK/build_tarballs.jl), is that not enough?

M/MAGEMin/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+platforms = expand_gfortran_versions(platforms)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libMAGEMin", :MAGEMin_library)
```
giordano, line 41: Can we call this
```suggestion
    LibraryProduct("libMAGEMin", :libMAGEMin)
```
?

giordano:
For the error on Windows
```
[23:02:37] src/dump_function.c: In function ‘dump_init’:
[23:02:37] src/dump_function.c:41:6: error: too many arguments to function ‘mkdir’
[23:02:37]       mkdir(gv.outpath, 0700);
[23:02:37]       ^~~~~
```
you need something like https://github.com/JuliaPackaging/Yggdrasil/blob/d936efb0feaa2665c1b25766d2a1bc2258738d6d/F/Fontconfig/bundled/patches/0002-fix-mkdir.mingw.patch#L7-L9 because the Windows function [`_mkdir`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mkdir-wmkdir?view=msvc-170) takes a single argument and [`mkdir`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mkdir?view=msvc-170) is simply a deprecated alias for the single-argument `_mkdir`.

M/MAGEMin/build_tarballs.jl
boriskaus: We need `LAPACKE`, which is the C interface to `LAPACK`. As far as I can tell that is not provided as part of `LAPACK`. Do you want a separate BB package for it?

M/MAGEMin/build_tarballs.jl
boriskaus: Sure

M/MAGEMin/build_tarballs.jl
giordano: Yes.  @ViralBShah do you have preferences about how to package LAPACKE?  But again, I think also OpenBLAS has LAPACKE, at least
```console
% nm /usr/lib/julia/libopenblas64_.so|grep LAPACKE|wc -l
2541
```
so why not just using it instead?

boriskaus: I'm fine with using `OpenBLAS`, but if I add this dependency:
```
Dependency(PackageSpec(name="OpenBLAS_jll", uuid="4536629a-c528-5b80-bd46-f80d51c5b363"))
```
there is no `liblapacke.so` downloaded:
```
sandbox:${WORKSPACE}/srcdir/MAGEMin # ls $libdir/
cmake                          libitm.so.1                    libmpx.so                      libssp.so
libasan.so                     libitm.so.1.0.0                libmpx.so.2                    libssp.so.0
libasan.so.3                   libload_time_mpi_constants.so  libmpx.so.2.0.0                libssp.so.0.0.0
libasan.so.3.0.0               liblsan.so                     libmpxwrappers.so              libstdc++.so
libatomic.so                   liblsan.so.0                   libmpxwrappers.so.2            libstdc++.so.6
libatomic.so.1                 liblsan.so.0.0.0               libmpxwrappers.so.2.0.0        libstdc++.so.6.0.29
libatomic.so.1.2.0             libmpi.so                      libnlopt.so                    libtsan.so
libcilkrts.so                  libmpi.so.12                   libnlopt.so.0                  libtsan.so.0
libcilkrts.so.5                libmpi.so.12.2.1               libnlopt.so.0.11.1             libtsan.so.0.0.0
libcilkrts.so.5.0.0            libmpich.so                    libobjc.so                     libubsan.so
libfmpich.so                   libmpichcxx.so                 libobjc.so.4                   libubsan.so.0
libgcc_s.so.1                  libmpichf90.so                 libobjc.so.4.0.0               libubsan.so.0.0.0
libgfortran.so                 libmpicxx.so                   libopa.so                      libvtv.so
libgfortran.so.3               libmpicxx.so.12                libopenblas64_.0.3.20.so       libvtv.so.0
libgfortran.so.3.0.0           libmpicxx.so.12.2.1            libopenblas64_.so              libvtv.so.0.0.0
libgomp.so                     libmpifort.so                  libopenblas64_.so.0            pkgconfig
libgomp.so.1                   libmpifort.so.12               libquadmath.so
libgomp.so.1.0.0               libmpifort.so.12.2.1           libquadmath.so.0
libitm.so                      libmpl.so                      libquadmath.so.0.0.0
```
It might be that I'm doing something else wrong, though (this is my first attempt at using BB). I ran the modified script with:
```
julia build_tarballs.jl --debug --verbose x86_64-linux-gnu
```

giordano: It's `libopenblas64_.so`, as I've shown above :slightly_smiling_face:  If you don't want to mess up with ILP64, you may want to use `OpenBLAS32_jll`, in which case the library is called `libopenblas.${dlext}` for all platforms.

boriskaus: sorry, I noticed my mistake (have to link against `libopenblas64_.so`  instead of `liblapacke.so`). 

Yet, it seems that not all functionality of `LAPACKE` is there:
```
sandbox:${WORKSPACE}/srcdir/MAGEMin # make CC=$CC CCFLAGS="-O3 -g -fPIC -std=c99" LIBS="-L${libdir} -lm -lopenblas64_ -lnlopt ${MPI_LIBS}" INC="-I${includedir}" 
cc  -o MAGEMin src/MAGEMin.o src/toolkit.o src/io_function.o src/gem_function.o src/gss_init_function.o src/gss_function.o src/NLopt_opt_function.o src/objective_functions.o src/pp_min_function.o src/ss_min_function.o src/simplex_levelling.o src/PGE_function.o src/phase_update_function.o src/dump_function.o -I/workspace/destdir/include -L/workspace/destdir/lib -lm -lopenblas64_ -lnlopt -lmpi 
src/toolkit.o: In function `inverseMatrix':
/workspace/srcdir/MAGEMin/src/toolkit.c:408: undefined reference to `LAPACKE_dgetrf'
/workspace/srcdir/MAGEMin/src/toolkit.c:409: undefined reference to `LAPACKE_dgetri'
src/PGE_function.o: In function `PGE_function':
/workspace/srcdir/MAGEMin/src/PGE_function.c:791: undefined reference to `LAPACKE_dgesv'
make: *** [Makefile:48: all] Error 1
```
checking for that file gives slightly different names:
```
sandbox:${WORKSPACE}/srcdir/MAGEMin # nm $prefix/lib/libopenblas64_.so.0 | grep LAPACKE_dgetrf
0000000001b28dd0 T LAPACKE_dgetrf264_
0000000001b28e80 T LAPACKE_dgetrf2_work64_
0000000001b28b10 T LAPACKE_dgetrf64_
0000000001b28bc0 T LAPACKE_dgetrf_work64_
```

It would ofcourse be easy to expand the current `LAPACK` implementation to also compile `LAPACKE`, but I suppose it would make the `_jll` package a bit larger.

giordano: > If you don't want to mess up with ILP64, you may want to use OpenBLAS32_jll, in which case the library is called libopenblas.${dlext} for all platforms.

boriskaus: that indeed does the trick - thanks!

M/MAGEMin/bundled/patches/update_lapacke.patch
```diff
@@ -0,0 +1,347 @@
+diff --git before/lapack-3.10.0/LAPACKE/Makefile after/lapack-3.10.0/LAPACKE/Makefile
+index 836fe0e..6f99eec 100644
```
giordano, line 2: This file can be removed now 🙂

giordano:
Ok, last platform failing is aarch64-apple-darwin (aka Apple Silicon):
```
[19:07:29] Undefined symbols for architecture arm64:
[19:07:29]   "___divdc3", referenced from:
[19:07:29]       _obj_liq in objective_functions.o
[19:07:29] ld: symbol(s) not found for architecture arm64
[19:07:29] clang-12: error: linker command failed with exit code 1 (use -v to see invocation)
```
We've seen the same error in #2938 and #4059, but we didn't understand what was the problem and how to fix it.  I don't know what you want to do, whether to try and fix this error in the package or to remove this platform.  It's up to you.

boriskaus:
Thanks a lot for all the fixes, this is great!

The `___divdc3` problem appears to come from the following line in our code:
```
1.0/cpow(sf[15], 4.0)
```
Since the last term can be a complex number, we would be dividing a double by a complex. Yet, in our code we can fix this altogether (as `sf[]` is in fact a double) and use `pow` instead of `cpow`.

boriskaus:
So now it compiles fine. The only thing is that I incremented the version # to `1.0.1` to reflect a change in our code which was necessary for this. Yet, the PR has the old name (I'm not sure whether this will automatically be taken into account)

giordano:
Brilliant!  Thank you so much for sticking with us, it's also good to now know what we have to do when we encounter the problem with `___divdc3`

PR #4609, opened 2022-03-14
Wizard recipe: SDL2_net-v2.0.1

jhigginbotham64:

* Package name: SDL2_net
* Version: v2.0.1

### DIFF
--- a/S/SDL2_net/build_tarballs.jl
+++ b/S/SDL2_net/build_tarballs.jl
@@ -31,7 +31,7 @@ platforms = supported_platforms()
 
 # The products that we will ensure are always built
 products = [
-    LibraryProduct("libSDL2_net", :libSDL2_net)
+    LibraryProduct(["libSDL2_net", "SDL2_net"], :libSDL2_net)
 ]
 
 # Dependencies that must be installed before this package can be built
--- a/S/SDL2_net/build_tarballs.jl
+++ b/S/SDL2_net/build_tarballs.jl
@@ -13,7 +13,7 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/SDL_net*/
+cd $WORKSPACE/srcdir/SDL*/
 
 mkdir build && cd build
 
--- a/S/SDL2_net/build_tarballs.jl
+++ b/S/SDL2_net/build_tarballs.jl
@@ -15,10 +15,6 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/SDL_net*/
 
-autoreconf -vi
-
-update_configure_scripts
-
 mkdir build && cd build
 
 ../configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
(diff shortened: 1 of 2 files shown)


S/SDL2_net/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "SDL2_net"
+version = v"2.0.1"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/libsdl-org/SDL_net/archive/refs/tags/release-$(version).tar.gz", "1fefe563ea333a2655c32169d03a376a334cdbe39da51fd424bf5f430dec83f4")
```
giordano, line 10: I suggest using a release tarball instead of struggling with bootstrap
```suggestion
    ArchiveSource("https://libsdl.org/projects/SDL_net/release/SDL2_net-$(version).tar.gz",
                  "15ce8a7e5a23dafe8177c8df6e6c79b6749a03fff1e8196742d3571657609d21"),
```

jhigginbotham64: thanks for the feedback, i'm pretty new to this

S/SDL2_net/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libSDL2_net", :libSDL2_net)
```
giordano, line 34: For Windows:
```suggestion
    LibraryProduct(["libSDL2_net", "SDL2_net"], :libSDL2_net)
```

jhigginbotham64:
@giordano ok i'm lost here, this is only my second jll contribution. could you explain what's going on here, for my benefit? how did that fix it for windows but not for the other 2, and what should i look into for fixing the other 2?

giordano:
> EDIT: ok i can see that on windows it builds the library with a different name, but the pipeline output for the other 2 makes it seem like it's not getting built at all

Yes, for Windows the problem was only the name, see https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct.

For PowerPC and FreeBSD the problem is unfortunately well known: https://docs.binarybuilder.org/stable/troubleshooting/#Shared-library-not-built  And being able to update the `configure` script would be great, but apparently it's a bit of a mess because maybe some macros aren't available.  I'll try to have a look in a bit.

jhigginbotham64:
> For PowerPC and FreeBSD the problem is unfortunately well known: https://docs.binarybuilder.org/stable/troubleshooting/#Shared-library-not-built And being able to update the `configure` script would be great, but apparently it's a bit of a mess because maybe some macros aren't available. I'll try to have a look in a bit.

i suspected it had something to do with that but my earlier naive solution didn't work out. thanks for the help.

giordano:
It's working now, thanks for your contribution!

jhigginbotham64:
my pleasure, thanks for your help!

Issue #531, opened 2020-02-18
Issues with MPI and CMake


dpo:
I'm trying to build a JLL for PARMETIS. Here is my build script: https://github.com/dpo/Yggdrasil/tree/parmetis/P/PARMETIS

I tried using OpenMPI_jll but running `mpicc` results in an error saying that the `cc` compiler can't be found.

The linux builds go through with MPICH_jll, but I get a failure on macOS at the time of linking the shared library:
```
[...]

PR #4610, opened 2022-03-14
Update build_tarballs.jl

boriskaus:
* reintroduce windows support
* add SuiteSparse dependency
* add MUMPS support

### DIFF
--- a/P/PETSc/build_tarballs.jl
+++ b/P/PETSc/build_tarballs.jl
@@ -11,6 +11,7 @@ sources = [
     DirectorySource("./bundled"),
 ]
 
+
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/petsc*
@@ -33,6 +34,13 @@ build_petsc()
     else
         USE_INT64=0
     fi
+    USE_MUMPS=1
+    if [[ "${1}" == "single" ]]; then
+        USE_SUITESPARSE=0
+    else
+        USE_SUITESPARSE=1
+    fi
+
     mkdir $libdir/petsc/${1}_${2}_${3}
     ./configure --prefix=${libdir}/petsc/${1}_${2}_${3} \
         CC=${CC} \
@@ -48,6 +56,10 @@ build_petsc()
         --with-batch \
         --with-blaslapack-lib=$BLAS_LAPACK_LIB \
         --with-blaslapack-suffix="" \
+        --download-mumps=${USE_MUMPS} \
+        --download-scalapack=${USE_MUMPS} \
+        --download-suitesparse=${USE_SUITESPARSE} \
+        --with-suitesparse=${USE_SUITESPARSE} \
         --known-64-bit-blas-indices=0 \
         --with-mpi-lib="${MPI_LIBS}" \
         --known-mpi-int64_t=0 \
@@ -58,6 +70,7 @@ build_petsc()
         --PETSC_ARCH=${target}_${1}_${2}_${3}
 
     if [[ "${target}" == *-mingw* ]]; then
+
         export CPPFLAGS="-Dpetsc_EXPORTS"
     elif [[ "${target}" == powerpc64le-* ]]; then
         export CFLAGS="-fPIC"
@@ -95,7 +108,8 @@ build_petsc single complex Int64
 """
 
 # We attempt to build for all defined platforms
-platforms = expand_gfortran_versions(supported_platforms(exclude=[Platform("i686", "windows")]))
+#platforms = expand_gfortran_versions(supported_platforms(exclude=[Platform("i686", "windows")]))
+platforms = expand_gfortran_versions(supported_platforms())
 
 products = [
     # Current default build, equivalent to Float64_Real_Int32
@@ -115,6 +129,8 @@ dependencies = [
     Dependency("MPICH_jll"; platforms=filter(!Sys.iswindows, platforms)),
     Dependency("MicrosoftMPI_jll"; platforms=filter(Sys.iswindows, platforms)),
     Dependency("CompilerSupportLibraries_jll"),
+    Dependency("SCALAPACK32_jll"),
+    Dependency("CMake_jll"),
 ]
 
 # Build the tarballs.


giordano:
CC: @Wimmerer @jkozdon @mohamed82008

boriskaus:
Windows still fails (for two different reasons in fact), which is something I can look into (since most of my students have windows machines it would be very good to get this up & running).

Yet, I am a bit puzzled why `aarch64-linux-musl-libgfortran4` and `aarch64-linux-gnu-libgfortran4` fail. They both compile fine on my local system(s). @giordano: any idea why this could be? 

Also: ideally, I would like to add a `superlu_dist` dependency as well, but it complains that the version of `cmake` used by BinaryBuilder is too old. If I download the [CMake_jll](https://github.com/JuliaBinaryWrappers/CMake_jll.jl) dependency, it complains that it cannot use the executable. Is there a way around it?

giordano:
> Yet, I am a bit puzzled why `aarch64-linux-musl-libgfortran4` and `aarch64-linux-gnu-libgfortran4` fail. They both compile fine on my local system(s). @giordano: any idea why this could be? 

Not without reading the config log.

> If I download the [CMake_jll](https://github.com/JuliaBinaryWrappers/CMake_jll.jl) dependency, it complains that it cannot use the executable. Is there a way around it?

We have already `cmake` in the build environment, why do add that as a dependency?  If anything, that'd need to be a [`HostBuildDependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.HostBuildDependency): https://docs.binarybuilder.org/stable/build_tips/#Dependencies-for-the-target-system-vs-host-system.  But again, I don't see why you need to add it at all.

boriskaus:
> > Yet, I am a bit puzzled why `aarch64-linux-musl-libgfortran4` and `aarch64-linux-gnu-libgfortran4` fail. They both compile fine on my local system(s). @giordano: any idea why this could be?
> 
> Not without reading the config log.

Is there a way to retrieve the `configure.log` file from the Azure pipeline? I can't reproduce this locally.

> > If I download the [CMake_jll](https://github.com/JuliaBinaryWrappers/CMake_jll.jl) dependency, it complains that it cannot use the executable. Is there a way around it?
> 
> We have already `cmake` in the build environment, why do add that as a dependency? If anything, that'd need to be a [`HostBuildDependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.HostBuildDependency): https://docs.binarybuilder.org/stable/build_tips/#Dependencies-for-the-target-system-vs-host-system. But again, I don't see why you need to add it at all.

ok, so how can I use a more recent version of `cmake` in the build environment? That would certainly be the preferred way.

giordano:
> Is there a way to retrieve the configure.log file from the Azure pipeline? I can't reproduce this locally.

`cat` it? :slightly_smiling_face: 

> ok, so how can I use a more recent version of cmake in the build environment? That would certainly be the preferred way.

```console
sandbox:${WORKSPACE} # cmake --version
cmake version 3.21.3

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```
Is `v3.21.3` _that_ old?  It was [released last September](https://github.com/Kitware/CMake/releases/tag/v3.21.3).

boriskaus:
> > Is there a way to retrieve the configure.log file from the Azure pipeline? I can't reproduce this locally.
> 
> `cat` it? 🙂
> 
ok good point...

> > ok, so how can I use a more recent version of cmake in the build environment? That would certainly be the preferred way.
> 
> ```
> sandbox:${WORKSPACE} # cmake --version
> cmake version 3.21.3
> 
> CMake suite maintained and supported by Kitware (kitware.com/cmake).
> ```
> 
> Is `v3.21.3` _that_ old? It was [released last September](https://github.com/Kitware/CMake/releases/tag/v3.21.3).

No that would be fine. Yet, if I do the same I get:
```
sandbox:${WORKSPACE}/srcdir/petsc-3.16.5 # cmake --version
cmake version 3.17.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```
and the error message states:
```
[14:52:23] *******************************************************************************
[14:52:23]          UNABLE to CONFIGURE with GIVEN OPTIONS    (see configure.log for details):
[14:52:23] -------------------------------------------------------------------------------
[14:52:23] Requires cmake version 3.18.1 or higher: use --download-cmake
[14:52:23] *******************************************************************************
[14:52:23] 
[14:52:23] Installing license file "/workspace/srcdir/petsc-3.16.5/LICENSE" to "/workspace/destdir/share/licenses/PETSc/LICENSE"...
```

giordano:
> No that would be fine. Yet, if I do the same I get:

You're probably using an old version of BinaryBuilderBase.  I think [`v1.5.0`](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/releases/tag/v1.5.0) is the first one with the Alpine Linux v3.15 image.

giordano:
BTW, if you want to use the same environment as the one used by the build machine, you can simply activate the environment in the [`.ci/`](https://github.com/JuliaPackaging/Yggdrasil/tree/920df12ec5b927a80818ec5a343efc8654fab0a4/.ci) directory :wink:

Wimmerer:
Why did we exclude 32 bit Windows originally? I believe it's because of MS-MPI.

Also we should probably use the Ygg copies of SuiteSparse and MUMPS once the `--download` build is working.

boriskaus:
> > No that would be fine. Yet, if I do the same I get:
> 
> You're probably using an old version of BinaryBuilderBase. I think [`v1.5.0`](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/releases/tag/v1.5.0) is the first one with the Alpine Linux v3.15 image.

Thanks - I was using Julia 1.6.4 (with a newly downloaded `BinaryBuilder` package which relies on `BinaryBuilderBase` 1.3.0). It seems that 1.7 is required for newer `BinaryBuilderBase`. That replaced this bug with a new one..

giordano:
Oh, yes, latest versions of BinaryBuilder require Julia v1.7

boriskaus:
> `cat` it?

@giordano: Sorry to have to get back to you regarding this, but the error message is as follows:
```
[15:37:10] TESTING: configureLibrary from config.packages.scalapack(/workspace/srcdir/petsc-3.16.5/config/BuildSystem/config/package.py:930)
[15:37:10]                       =============================================================================================
[15:37:10]                             Trying to download git://https://bitbucket.org/petsc/pkg-scalapack for SCALAPACK
[15:37:12]                       =============================================================================================
[15:37:12]                     
[15:37:12]                       =============================================================================================
[15:37:12]                             Compiling and installing Scalapack; this may take several minutes
[15:37:44]                       =============================================================================================
[15:37:44]                     
[15:37:44] *******************************************************************************
[15:37:44]          UNABLE to CONFIGURE with GIVEN OPTIONS    (see configure.log for details):
[15:37:44] -------------------------------------------------------------------------------
[15:37:44] Error running make on SCALAPACK
[15:37:44] *******************************************************************************
[15:37:44] 
[15:37:44] Installing license file "/workspace/srcdir/petsc-3.16.5/LICENSE" to "/workspace/destdir/share/licenses/PETSc/LICENSE"...
```

So ideally we would `cat configure.log` just before the license file is installed. Unfortunately, I don't see the location in `build_tarballs.jl` where this happens. If I write it directly after the `configure` step, it is not displayed when an error occurs (only once it successfully configures).

giordano:
The license file is installed automatically at the end of the build, that's why you don't see it in the script.  What is failing is the function `build_petsc double real Int32`: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=18203&view=logs&j=754f51df-0adc-5fae-77c8-79c540867ac1&t=a19a2a72-6b6b-5735-1115-c8980685a245&l=11474.

boriskaus:
yes, what fails is this step:
```
./configure --prefix=${libdir}/petsc/${1}_${2}_${3} \
        CC=${MPICC} \
        FC=${MPIFC} \
        CXX=${MPICXX} \
        COPTFLAGS='-O3' \
        CXXOPTFLAGS='-O3' \
        CFLAGS='-fno-stack-protector' \
        LDFLAGS="-L${libdir}" \
        FOPTFLAGS='-O3' \
        --with-64-bit-indices=${USE_INT64} \
        --with-debugging=0 \
        --with-batch \
        --with-blaslapack-lib=$BLAS_LAPACK_LIB \
        --with-blaslapack-suffix="" \
        --download-mumps=${USE_MUMPS} \
        --download-scalapack=${USE_MUMPS} \
        --download-suitesparse=${USE_SUITESPARSE} \
        --download-superlu_dist=${USE_SUPERLU_DIST} \
        --with-suitesparse=${USE_SUITESPARSE} \
        --known-64-bit-blas-indices=0 \
        --with-mpi-lib="${MPI_LIBS}" \
        --known-mpi-int64_t=0 \
        --with-mpi-include="${includedir}" \
        --with-sowing=0 \
        --with-precision=${1} \
        --with-scalar-type=${2} \
        --PETSC_ARCH=${target}_${1}_${2}_${3}
```
This generates `configure.log`, which one can normally inspect to see what happens (or you can send it to the PETSc team, which are usually great in helping with this). Yet in Azure, I have no access to this file (or at least I don't know how) and doing `cat configure.log` directly following this step does not work as the building process is already terminated. Locally, I can't reproduce the error.

Would you have an idea how we could inspect this file?

giordano:
```
./configure ... || true
```
:wink:

Wimmerer:
> We support both MS-MPI [64-bit] and Intel MPI on Microsoft Windows (MPICH2 does not work, do not use it).

I believe that is the line from the PETSc docs why we didn't build 32-bti perviously. 64-bit Windows worked in 3.15 but doesn't work in 3.16 despite building successfully:

```julia
ERROR: Unable to automatically install 'PETSc' from 'C:\Users\wrkim\.julia\packages\PETSc_jll\k1kcY\Artifacts.toml'
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:33
  [2] ensure_artifact_installed(name::String, meta::Dict{String, Any}, artifacts_toml::String; platform::Base.BinaryPlatforms.Platform, verbose::Bool, quiet_download::Bool, io::Base.TTY)
    @ Pkg.Artifacts C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\Pkg\src\Artifacts.jl:441
  [3] download_artifacts(env::Pkg.Types.EnvCache; platform::Base.BinaryPlatforms.Platform, julia_version::VersionNumber,
 verbose::Bool, io::Base.TTY)
    @ Pkg.Operations C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\Pkg\src\Operations.jl:617
  [4] up(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}, level::Pkg.Types.UpgradeLevel; skip_writing_project::Bool)
    @ Pkg.Operations C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\Pkg\src\Operations.jl:1267
  [5] up(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}; level::Pkg.Types.UpgradeLevel, mode::Pkg.Types.PackageMode, update_registry::Bool, skip_writing_project::Bool, kwargs::Base.Pairs{Symbol, Base.TTY, Tuple{Symbol}, NamedTuple{(:io,), Tuple{Base.TTY}}})
    @ Pkg.API C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\Pkg\src\API.jl:333
  [6] up(pkgs::Vector{Pkg.Types.PackageSpec}; io::Base.TTY, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Pkg.API C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\Pkg\src\API.jl:149
  [7] up(pkgs::Vector{Pkg.Types.PackageSpec})
    @ Pkg.API C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\Pkg\src\API.jl:144
  [8] do_cmd!(command::Pkg.REPLMode.Command, repl::REPL.LineEditREPL)
    @ Pkg.REPLMode C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\Pkg\src\REPLMode\REPLMode.jl:407
  [9] do_cmd(repl::REPL.LineEditREPL, input::String; do_rethrow::Bool)
    @ Pkg.REPLMode C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\Pkg\src\REPLMode\REPLMode.jl:385
 [10] do_cmd
    @ C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\Pkg\src\REPLMode\REPLMode.jl:376 [inlined]
 [11] (::Pkg.REPLMode.var"#24#27"{REPL.LineEditREPL, REPL.LineEdit.Prompt})(s::REPL.LineEdit.MIState, buf::IOBuffer, ok::Bool)
    @ Pkg.REPLMode C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\Pkg\src\REPLMode\REPLMode.jl:549
 [12] #invokelatest#2
    @ .\essentials.jl:716 [inlined]
 [13] invokelatest
    @ .\essentials.jl:714 [inlined]
 [14] run_interface(terminal::REPL.Terminals.TextTerminal, m::REPL.LineEdit.ModalInterface, s::REPL.LineEdit.MIState)
    @ REPL.LineEdit C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\REPL\src\LineEdit.jl:2493
 [15] run_frontend(repl::REPL.LineEditREPL, backend::REPL.REPLBackendRef)
    @ REPL C:\Users\wrkim\.julia\juliaup\julia-1.7.2+0~x64\share\julia\stdlib\v1.7\REPL\src\REPL.jl:1232
 [16] (::REPL.var"#49#54"{REPL.LineEditREPL, REPL.REPLBackendRef})()
    @ REPL .\task.jl:423
```

giordano:
> 64-bit Windows worked in 3.15 but doesn't work in 3.16 despite building successfully:

Also Windows typically use a case-insentive filesystem, so the same problem you saw on macOS applies there.  But you can see the error exactly with Julia v1.8+

Wimmerer:
Should I fix the case insensitivity in a separate PR, and this one can rebase?

giordano:
Ok

boriskaus:
ok, thanks to the help of the PETSc team, some progress was made:

-  the basic PETSc for windows compiles fine now, provided that we do not include external packages. 

- On linux, most compilations (with 2 exceptions, see below) work fine with the external packages `superlu_dist`,`mumps` and `suitesparse` (for those PETSc libraries that allow to build them)

- On apple, a linker problem occurs for `superlu_dist`, so I have temporarily deactivated that solver but the others work. I've send an email to the PETSc list regarding this, so perhaps it can soon be fixed as well.

The two problem cases that remain are:
```
aarch64-linux-musl-libgfortran4
aarch64-linux-gnu-libgfortran3
```
I have now been able to reproduce this locally with Julia 1.7 & BinaryBuilderBase 1.7.0. Interestingly, it compiles just fine on Julia 1.6 with BinaryBuilderBase 1.3.0 so something must have happened meanwhile.

The local build makes it somewhat easier to track down where the mistake occurs, and it is:
```
SRC/pzrot.c: In function ‘pzrot_’:
SRC/pzrot.c:333:12: warning: implicit declaration of function ‘numroc_’ [-Wimplicit-function-declaration]
       nq = numroc_( &nn, &desc_X[NB_], &mycol, &ixcol, &npcol );
            ^~~~~~~
TOOLS/LAPACK/clarot.f:297:0:

       END
 
internal compiler error: Segmentation fault
TOOLS/SL_gridreshape.c: In function ‘SL_Cgridreshape’:
```
Unfortunately that is not very helpful..

One solution is ofcourse to deactivate `MUMPS` (the external package affected by this bug) on these particular systems.

Wimmerer:
Is it possible to do that for SuiteSparse and MUMPS as well? I don't think we have SuperLU_Dist built right now.

boriskaus:
I did try this with SuiteSparse32, but it wasn't working for me (perhaps you are more successful?)

Might be that this is because `PETSc` compiles of for different flavours of precision and integer lengths. We would need to have a version of `MUMPS` and `SuiteSparse` that exactly corresponds to that.

boriskaus:
ok, so the problem with `superlu_dist` on apple is that it uses CMake to build the package, but employs `-soname` during the linking, which should be `-install_name`. So if there is a way to tell CMake this with a flag, it should likely compile.

The same problem was discussed earlier, and @giordano seemed to have solved it as discussed [here](https://github.com/JuliaPackaging/Yggdrasil/issues/531). Yet, it isn't obvious to me what the solution was. In case you recall, it would be great!

PR #4611, opened 2022-03-14
New recipe: OpenGLMathematics-v0.9.9

jhigginbotham64:
I am attempting to create a jll package to wrap GLM for use in C/C++ dependencies of Julia projects.

### DIFF
--- a/G/glm/build_tarballs.jl
+++ b/G/glm/build_tarballs.jl
@@ -14,7 +14,11 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/glm*
-cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -S . -B .
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    ..
 make -j${nproc}
 make install
 """
--- a/G/glm/build_tarballs.jl
+++ b/G/glm/build_tarballs.jl
@@ -25,7 +25,7 @@ platforms = supported_platforms()
 
 # The products that we will ensure are always built
 products = [
-    FileProduct("include/glm/glm.hpp", glm_hpp),
+    FileProduct("include/glm/glm.hpp", :glm_hpp),
 ]
 
 # Dependencies that must be installed before this package can be built
--- a/G/glm/build_tarballs.jl
+++ b/G/glm/build_tarballs.jl
@@ -31,5 +31,4 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[]
 
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"11.1.0")
-
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
--- a/G/glm/build_tarballs.jl
+++ b/G/glm/build_tarballs.jl
@@ -31,5 +31,5 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[]
 
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
 
--- a/G/glm/build_tarballs.jl
+++ b/G/glm/build_tarballs.jl
@@ -8,7 +8,7 @@ version = v"9.9.9"
 # Collection of sources required to build this package
 sources = [
     GitSource("https://github.com/g-truc/glm.git",
-                  "6ad79aae3eb5bf809c30bf1168171e9e55857e45"),
+              "6ad79aae3eb5bf809c30bf1168171e9e55857e45"),
 ]
 
 # Bash recipe for building across all platforms
--- a/G/glm/build_tarballs.jl
+++ b/G/glm/build_tarballs.jl
@@ -21,7 +21,7 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = [AnyPlatform()]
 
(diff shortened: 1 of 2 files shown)


giordano:
This is header only?

G/glm/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
...
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "glm"
+version = v"9.9.9"
+
+# Collection of sources required to build this package
+sources = [
+    GitSource("https://github.com/g-truc/glm.git",
+                  "6ad79aae3eb5bf809c30bf1168171e9e55857e45"),
```
giordano, line 11:
```suggestion
    GitSource("https://github.com/g-truc/glm.git",
              "6ad79aae3eb5bf809c30bf1168171e9e55857e45"),
```

jhigginbotham64:
Yes

G/glm/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir/glm*
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -S . -B .
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
```
giordano, line 24: Since this is header only you should just target `AnyPlatform`: https://docs.binarybuilder.org/stable/building/#Platform-independent-packages
```suggestion
platforms = [AnyPlatform()]
```

G/glm/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
...
+# The products that we will ensure are always built
+products = [
+    FileProduct("include/glm/glm.hpp", :glm_hpp),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[]
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"11.1.0")
+
```
giordano, line 35:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```

G/glm/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
...
+    GitSource("https://github.com/g-truc/glm.git",
+                  "6ad79aae3eb5bf809c30bf1168171e9e55857e45"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/glm*
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -S . -B .
+make -j${nproc}
+make install
```
giordano, line 19:
```suggestion
mkdir build && cd build
cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    ..
make -j${nproc}
make install
```

G/glm/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "glm"
+version = v"9.9.9"
```
giordano, line 6: Looks like the version number is [0.9.9.8](https://github.com/g-truc/glm/releases/tag/0.9.9.8), not really 9.9.9.  However we have to stick to the format `X.Y.Z`, so maybe 0.9.9?

G/glm/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "glm"
```
giordano, line 5: The upstream project seems to style it GLM, all capital:
```suggestion
name = "GLM"
```
Also, please note that you'll have to rename the directory accordingly

jhigginbotham64: i don't mind one way or another, but there's already a GLM package (https://juliastats.org/GLM.jl/stable/) and the header file names and namespace are lowercased (unlike, for instance, with SDL). that's why i went for lowercase.

but let me know what you think given that and i'll make any needed changes.

G/glm/build_tarballs.jl
jhigginbotham64: there hasn't been a release yet for 0.9.9.9, but i'm using it anyway from the master branch because the 0.9.9.8 tarball's CMakeLists.txt doesn't have an install target.

i think 0.9.9 (essentially omitting the 4th digit) is more future-proof than 9.9.9, so i'll go ahead and make that change.

G/glm/build_tarballs.jl
giordano: If you are worried about name conflicts, anyway in the registry we can't have two packages with exact same name apart from case: that wouldn't work on case-insensitive file systems (such as those typically used on Windows and macOS) :slightly_smiling_face: All JLL packages will have `_jll` appended to the name, so the conflict isn't a concern.  The problem would be if here we had another package called `GLM_jll` already: in that case any case wouldn't be acceptable, because of the registry issue above.

jhigginbotham64: that's very interesting, but i still like lowercase for the package name since their git repo, header files, install directory names, and C++ namespace are all lowercase, as distinct from packages like SDL where they are uppercase. thus i think it's clearer to keep the name lowercase, rather than risk confusion with the uppercase GLM.

...on the other hand, it's probably referred to in uppercase in official docs.

idk if there's a "right" answer, both sides have a case. :stuck_out_tongue_closed_eyes:

what do you think? do you want me to go ahead change it?

jhigginbotham64: i looked at the package naming guidelines and didn't find an obvious answer, but there is point 4 "Err on the side of clarity, even if clarity seems long-winded to you." What would you think of "OpenGLMath" as an alternative to "glm" or "GLM"?

giordano: Maybe that's better, yes

jhigginbotham64: ok, that's what i'll do

PR #4613, opened 2022-03-15
[gitoxide] Add builder

giordano:


giordano:
The matrix generator job is getting stuck at
```
Determining builds to queue...
##[warning]Can't create TaskIssue from logging event.
Finishing: mtrx
```
So it doesn't actually create any matrix.  It's happening only in this PR and it worked until yesterday.  Sigh.

PR #4614, opened 2022-03-15
[PETSc] Case Insensitivity

Wimmerer:


P/PETSc/build_tarballs.jl
```diff
@@ -82,6 +82,8 @@ build_petsc()
     # Remove now broken links
     
     rm ${libdir}/petsc/${1}_${2}_${3}/lib/libpetsc.*
+    # Remove PETSc.pc because petsc.pc also exists, causing conflicts on case insensitive file-systems.
+    rm ${libdir}/petsc/${1}_${2}_${3}/lib/pkgconfig/PETSc.pc
```
giordano, line 86: So, honestly I'm not happy about the current situation.  These pkgconfig files are in a directory which isn't in the default pkgconfig search path.  While this can be fixed when building downstream packages by setting `PKG_CONFIG_PATH` to include the relevant directory, the next issue is that these files are just completely wrong.  So, we either fix these files to use the correct paths and library names, or just remove them.  As they are at the moment they're totally useless.

giordano: This is an example of file:
```
prefix=/workspace/destdir/lib/petsc/double_complex_Int32
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${prefix}/lib
ccompiler=cc
cflags_extra=-fno-stack-protector -fPIC -O3
cflags_dep=-MMD -MP
ldflag_rpath=-Wl,-rpath,
cxxcompiler=c++
cxxflags_extra=-Wall -Wwrite-strings -Wno-strict-aliasing -Wno-unknown-pragmas -fstack-protector -fvisibility=hidden -O3  -fPIC -std=gnu++17
fcompiler=gfortran
fflags_extra=-fPIC -Wall -ffree-line-length-0 -Wno-unused-dummy-argument -O3

Name: PETSc
Description: Library to solve ODEs and algebraic equations
Version: 3.16.5
Cflags:   -I${includedir}
Libs: -L${libdir} -lpetsc
Libs.private: -L/workspace/destdir/lib -L/opt/x86_64-linux-musl/x86_64-linux-musl/lib64 -L/opt/x86_64-linux-musl/lib/gcc/x86_64-linux-musl/9.1.0 -L/opt/x86_64-linux-musl/lib/gcc -L/opt/x86_64-linux-musl/x86_64-linux-musl/lib -L/opt/x86_64-linux-musl/lib -L/opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/lib -L/opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/lib -lopenblas -lmpifort -lmpi -lstdc++ -ldl -lgfortran -lm -lgfortran -lm -lgcc_s -lquadmath -lstdc++ -ldl
```
Actually, the paths at the top are right, the only issue is the `Libs` line: `-lpetsc` is definitely wrong.

Wimmerer: Should I edit them and maybe copy them to the `pkgconfig` folder of `$libdir`?

giordano: yeah, you could fix the library name with `sed` and move the directory to `${prefix}/lib` (not `${libdir}`, pkgconfig is _always_ in `${prefix}/lib`)

PR #4615, opened 2022-03-15
New recipe: Vulkan_Volk v1.3.204

jhigginbotham64:
I have been unable to create a [Vulkan-Loader](https://github.com/KhronosGroup/Vulkan-Loader) JLL package due to [this issue](https://github.com/KhronosGroup/Vulkan-Loader/issues/249), so here's an alternative way to build projects that load Vulkan functions.

### DIFF
--- a/V/Vulkan_Volk/build_tarballs.jl
+++ b/V/Vulkan_Volk/build_tarballs.jl
@@ -41,4 +41,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")


jhigginbotham64:
Ok...per the Volk README, it appears to be useless without Vulkan-Loader. However, using the loader defined in Vulkan-Hpp (provided by Vulkan_Headers_jll) makes both unnecessary (I think).

I'll keep this PR open for now, I don't see any reason not to, but a reviewer/member/maintainer can feel free to close it.

V/Vulkan_Volk/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [AnyPlatform()]
+
+products = [
+    FileProduct("include/volk.h", :volk_h),
+    FileProduct("lib/libvolk.a", :libvolk),
```
giordano, line 35: Uhm, a static library isn't really platform-independent, so this can't target `AnyPlatform`.  If anything, this should be built for each platform separately.

V/Vulkan_Volk/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("Vulkan_Headers_jll"),
```
giordano, line 40: Also here, this should be a buil-only dependency:
```suggestion
    BuildDependency("Vulkan_Headers_jll"),
```

PR #4619, opened 2022-03-15
Updated PThreadPool to v0.0.20200616

stemann:
Referenced by Torch v1.6.0-v1.7.1: https://github.com/pytorch/pytorch/tree/v1.7.1/third_party

Awaiting #4617

### DIFF
--- a/P/PThreadPool/build_tarballs.jl
+++ b/P/PThreadPool/build_tarballs.jl
@@ -13,6 +13,12 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
+if [[ $target == x86_64-apple-darwin* ]]; then
+    PTHREADPOOL_SYNC_PRIMITIVE=condvar
+else
+    PTHREADPOOL_SYNC_PRIMITIVE=default
+fi
+
 cd $WORKSPACE/srcdir/pthreadpool
 mkdir build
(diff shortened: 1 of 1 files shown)


giordano:
Not very useful to rebuild _before_ https://github.com/JuliaRegistries/General/pull/56678 is merged :slightly_smiling_face: (and you'll also need to give some time to PkgServer to catch up)

P/PThreadPool/build_tarballs.jl
```diff
@@ -36,14 +43,16 @@ install_license ../LICENSE
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = supported_platforms()
+filter!(p -> !(Sys.isapple(p) && arch(p) == "aarch64"), platforms) # aarch64-macos unsupported for CPUInfo 20200522
```
stemann, line 46: It actually *does* build on `aarch64-apple`, but the proper `CPUInfo` version (https://github.com/Maratyszcza/pthreadpool/blob/029c88620802e1361ccf41d1970bd5b07fd6b7bb/cmake/DownloadCpuinfo.cmake#L7) that should be used for building does not support aarch64-apple - but it also seems CPUInfo is not even being used...

PR #4621, opened 2022-03-15
New recipe: Telescope-v0.1.2

jhigginbotham64:
I'm working on a C++ game/rendering/multimedia library.

### DIFF
--- a/T/Telescope/build_tarballs.jl
+++ b/T/Telescope/build_tarballs.jl
@@ -48,4 +48,4 @@ dependencies = [
     Dependency("Vulkan_Headers_jll"),
 ]
 
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"9.1.0")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"8.1.0")
--- a/T/Telescope/build_tarballs.jl
+++ b/T/Telescope/build_tarballs.jl
@@ -33,7 +33,7 @@ platforms = [
     Platform("aarch64", "linux"; libc = "glibc"),
     Platform("powerpc64le", "linux"; libc = "glibc"),
     Platform("x86_64", "linux"; libc = "musl"),
-    Platform("aarch64", "linux"; libc = "musl")
+    Platform("aarch64", "linux"; libc = "musl"),
 ]
 
 platforms = expand_cxxstring_abis(platforms)
--- a/T/Telescope/build_tarballs.jl
+++ b/T/Telescope/build_tarballs.jl
@@ -13,6 +13,8 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
+export CFLAGS="-I${prefix}/include"
+export CXXFLAGS="-I${prefix}/include"
 cd $WORKSPACE/srcdir/Telescope*
 mkdir build && cd build
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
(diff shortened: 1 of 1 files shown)


jhigginbotham64:
Noticing several different errors:
- several systems cannot find SDL2/SDL_mixer.h
- several have difficulty converting Vulkan-Hpp types to plain Vulkan types
- one of the windows ones is not finding certain extension symbols

jhigginbotham64:
Ok I've disabled the platforms it doesn't work for. It works for the ones I need for development. I'd love for this to get fixed, but given the choice between a long debugging process and merging soon, I'd rather merge soon so I can get moving again with development.

T/Telescope/build_tarballs.jl
```diff
@@ -0,0 +1,59 @@
...
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("OpenGLMathematics_jll"),
+    Dependency("SDL2_jll"),
+    Dependency("SDL2_image_jll"),
+    Dependency("SDL2_mixer_jll"),
+    Dependency("SDL2_net_jll"),
+    Dependency("SDL2_ttf_jll"),
+    Dependency("Shaderc_jll"),
+    Dependency("Vulkan_Headers_jll"),
```
giordano, line 56: Header-only dependencies aren't needed at runtime, they should be [`HostBuildDependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.HostBuildDependency): https://docs.binarybuilder.org/stable/building/#Binary-dependencies
```suggestion
    HostBuildDependency("OpenGLMathematics_jll"),
    Dependency("SDL2_jll"),
    Dependency("SDL2_image_jll"),
    Dependency("SDL2_mixer_jll"),
    Dependency("SDL2_net_jll"),
    Dependency("SDL2_ttf_jll"),
    Dependency("Shaderc_jll"),
    HostBuildDependency("Vulkan_Headers_jll"),
```

T/Telescope/build_tarballs.jl
```diff
@@ -0,0 +1,59 @@
...
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl")
+]
```
giordano, line 37: Looks like this can be simplified to
```suggestion
platforms = supported_platforms(; exclude=p->!(Sys.islinux(p) && nbits(p) == "64"))
```
What are the problems on the other platforms?  Maybe they can be fixed, at least partially?  However the decision of the platforms to support is up to you

T/Telescope/build_tarballs.jl
```diff
@@ -0,0 +1,59 @@
...
+# Collection of sources required to build this package
+sources = [
+    ArchiveSource("https://github.com/jhigginbotham64/Telescope/archive/refs/tags/v$(version).tar.gz",
+              "025a24a5dfe4c4d5910ae0cad7664ec2c2fc21cce8ffaae56cc6b42300249de6"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+export CFLAGS="-I${prefix}/include"
+export CXXFLAGS="-I${prefix}/include"
```
giordano, line 17: Nitpick: please use the most specific [environment variable](https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables), in this case `includedir`
```suggestion
export CFLAGS="-I${includedir}"
export CXXFLAGS="-I${includedir}"
```

T/Telescope/build_tarballs.jl
jhigginbotham64: Incorporating the simplification seems to lead to an empty platform list.

As to your question, I described them briefly [here](https://github.com/JuliaPackaging/Yggdrasil/pull/4621#issuecomment-1068485860) and failed pipeline outputs are visible [here](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=18263&view=logs&j=982d096e-6cad-5609-3037-305d94247f4e&t=f7dedcfd-9020-5606-352e-77d182153651&l=366).

giordano: That's because I got the value of `nbits(p)` wrong: it should have been the number `64`, not the string `"64"`

PR #4624, opened 2022-03-16
Update MAGEMin; add mpiexec dependency

boriskaus:
this adds mpiexec as an executable product

### DIFF
--- a/M/MAGEMin/build_tarballs.jl
+++ b/M/MAGEMin/build_tarballs.jl
@@ -45,6 +45,7 @@ platforms = expand_gfortran_versions(platforms)
 products = [
     LibraryProduct("libMAGEMin", :libMAGEMin)
     ExecutableProduct("MAGEMin", :MAGEMin)
+    ExecutableProduct("mpiexec", :mpiexec)
 ]
 
 # Dependencies that must be installed before this package can be built


M/MAGEMin/build_tarballs.jl
```diff
@@ -45,6 +45,7 @@ platforms = expand_gfortran_versions(platforms)
 products = [
     LibraryProduct("libMAGEMin", :libMAGEMin)
     ExecutableProduct("MAGEMin", :MAGEMin)
+    ExecutableProduct("mpiexec", :mpiexec)
```
giordano, line 48: I'm extremely confused: I don't see any `mpiexec` in the tarball, where it's coming from?

boriskaus: It’s coming from the MPICH dependency but this may indeed not be the way to do thinks so feel free to reject

boriskaus: I wanted to be able to call both `mpiexec` and `MAGEMin` as:
```julia
julia> using MAGEMin_jll
julia> run(`$(mpiexec()) -n 2 $(MAGEMin()) `)
```

giordano:
```julia
julia> using MAGEMin_jll, MPICH_jll

julia> mpirun = addenv(mpiexec(), MAGEMin_jll.JLLWrappers.JLLWrappers.LIBPATH_env=>MAGEMin_jll.LIBPATH[]);

julia> run(`$(mpirun) -n 2 $(MAGEMin_jll.MAGEMin_path)`);
Running MAGEMin 1.0.1 on 2 cores {
‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾

VOL_SYS           +1.582647
RHO_SYS           +3253.910261
MASS_RES          +0.000010
Rank          : 0 
Point         : 0 
Temperature   : 1100.0000        [C] 
Pressure      : 12.00    [kbar]

SOLUTION: [G = -825.337] (37 iterations, 113.42 ms)
[-1011.909246,-1829.091670,-819.265707,-695.468292,-412.942264,-971.879545,-876.528123,-1073.651135,-276.626077,-1380.314700,]
 opx     0.23184 
 cpx     0.15210 
 spn     0.01395 
  ol     0.60211 
Point             0
__________________________________
MAGEMin comp time: +285.928000 ms }
```

boriskaus: Thanks, that's what I was looking for! 
On Windows, we are not using `MPICH_jll` but `MicrosoftMPI_jll` instead. Do we need to do that with 
```
if Sys.iswindows

else

end
```
, or is there a way to do this directly based on the `mpiexec` that is loaded in `MAGEMin_jll`?

giordano: No, you'll have to do something like
```julia
const mpirun = if Sys.iswindows()
    MAGEMin_jll.MicrosoftMPI_jll.mpiexec()
else
    MAGEMin_jll.MPICH_jll.mpiexec()
end
```
Ideally, at some point in the future we'll be able to use [`MPItrampoline`](https://github.com/eschnett/MPItrampoline) for this, but I don't think we're there yet.

boriskaus: ok, thanks!

giordano: Actually, this should be more robust:
```julia
const mpirun = if MAGEMin_jll.MPICH_jll.is_available()
    MAGEMin_jll.MPICH_jll.mpiexec()
elseif MAGEMin_jll.MicrosoftMPI_jll.is_available()
    MAGEMin_jll.MicrosoftMPI_jll.mpiexec()
else
    nothing
end
```
You directly query the JLL packages whether they're available on the current platform, instead of hard-coding the platform-relations.  In general I recommend using `is_available()` to make sure the artifacts are available in the current system.

PR #4625, opened 2022-03-16
Update libjulia dev versions to latest

fingolfin:
It is really annoying that this also rebuilds the 1.6 and 1.7 artifacts, but I don't think we have any way to reuse those? Hrm :-(.

For 1.8 I used the latest release-1.8 branch, however, that already contains a bunch of changes for beta2. Should I use the `v1.8.0-beta1` instead then? Though I would hope neither will make a major difference.

Motivation for this PR: help debug https://github.com/timholy/Revise.jl/issues/675.


giordano:
> It is really annoying that this also rebuilds the 1.6 and 1.7 artifacts, but I don't think we have any way to reuse those? Hrm :-(.

Agreed, I don't really like that.  But what we can do?  I thought about using some special keywords in git commit messages, but I'm not really sure _how_ that'd would work.  Also, we'll need to be able to reuse the rest of the package (artifacts and wrappers for the untouched platforms), which I don't think we currently can.  Perhaps https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/639 could help with that :wink:

PR #4626, opened 2022-03-16
Add FLEXPART

tcarion:
Adding the binaries for [FLEXPART](https://www.flexpart.eu/) and `calc_etadot` that is needed for [flex_extract](https://www.flexpart.eu/flex_extract/).
I left `platforms = supported_platforms()` for the moment but I don't expect it to work for other platforms than linux.


giordano:
Your git history seems to be a bit messed up and so you're touching several unrelated projects :slightly_smiling_face:

tcarion:
Woops sorry about that ! I will make a clean one :slightly_smiling_face:

PR #4636, opened 2022-03-17
Updated SLEEF to v3.4.1

stemann:
Roughly the version (commit 7f523de651585fe25cade462efccca647dcc8d02) referenced by Torch v1.6.0 - v1.7.1: https://github.com/pytorch/pytorch/tree/v1.7.1/third_party

### DIFF
--- a/S/SLEEF/bundled/patches/version.patch
+++ b/S/SLEEF/bundled/patches/version.patch
@@ -1,0 +1,12 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index d16b526..9d4f6b9 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -24,7 +24,7 @@ enable_testing()
+ 
+ set(SLEEF_VERSION_MAJOR 3)
+ set(SLEEF_VERSION_MINOR 4)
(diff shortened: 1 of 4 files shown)


S/SLEEF/bundled/patches/version.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index d16b526..9d4f6b9 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -24,7 +24,7 @@ enable_testing()
+ 
+ set(SLEEF_VERSION_MAJOR 3)
+ set(SLEEF_VERSION_MINOR 4)
+-set(SLEEF_VERSION_PATCHLEVEL 0)
++set(SLEEF_VERSION_PATCHLEVEL 1)
```
giordano, line 10: Why do we need a patch to fix the version number?

stemann: Because the SLEEF authors forgot to.

giordano: 🤦

stemann: :-)

PR #4646, opened 2022-03-20
add libwatcher

Pangoraw:
This is port to Julia from the library [parcel-bundler/watcher](https://github.com/parcel-bundler/watcher) hosted at [JuliaPluto/watcher](https://github.com/JuliaPluto/watcher/tree/julia_port).

### DIFF
--- a/L/libwatcher/build_tarballs.jl
+++ b/L/libwatcher/build_tarballs.jl
@@ -34,7 +34,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    BuildDependency("LibUV_jll"),
+    Dependency("LibUV_jll"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/L/libwatcher/build_tarballs.jl
+++ b/L/libwatcher/build_tarballs.jl
@@ -1,0 +1,42 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "libwatcher"
+version = v"2.0.5"
+
+# Collection of sources required to build libwatcher
+sources = [
+    GitSource("https://github.com/JuliaPluto/watcher.git", "18d62d71091a0d9c1fb5c28375bcd7e417d39d66")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd watcher*/
+bash build.sh
+install -m 755 build/libwatcher.${dlext} ${libdir}/libwatcher.${dlext}
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc="glibc"),
+    Platform("x86_64", "macos"),
+    Platform("aarch64", "macos"),
+    Platform("x86_64", "windows"),
+]
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = [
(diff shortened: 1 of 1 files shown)


L/libwatcher/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libwatcher", :libwatcher),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("LibUV_jll"),
```
giordano, line 37: Does the library actually dynamically link to libuv?

Pangoraw: yes, it uses symbols from libuv. The jll is only used to build on Windows where it uses `libuv.dll.a` (I could not find a way to build without it).

giordano: > yes, it uses symbols from libuv.

That's not what I asked.  I asked whether the final library dynamically links to libuv.  If that's not the case, this should be a [`BuildDependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.BuildDependency), not a runtime one.

Pangoraw: the final library does not dynamically link with libuv and uses Julia's own libuv. I changed to a BuildDependency, thanks!

giordano: > the final library does not dynamically link with libuv and uses Julia's own libuv.

Honestly, that doesn't make much sense :slightly_smiling_face: 

I wanted to see something like this (and I was hoping that I didn't have to go through all of them manually myself):

x86_64 Linux:
```console
% readelf -d lib/libwatcher.so | grep NEEDED
```

x86_64 macOS:
```console
sandbox:${WORKSPACE} # otool -L libwatcher.dylib 
libwatcher.dylib:
        @rpath/libwatcher.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 307.4.0)
```

aarch64 macOS:
```console
sandbox:${WORKSPACE} # otool -L libwatcher.dylib 
libwatcher.dylib:
        @rpath/libwatcher.dylib (compatibility version 0.0.0, current version 0.0.0)
```

Windows:
```console
sandbox:${WORKSPACE} # ${target}-objdump -p libwatcher.dll | grep DLL
        DLL
        DLL Name: libgcc_s_seh-1.dll
        DLL Name: libstdc++-6.dll
        DLL Name: KERNEL32.dll
        DLL Name: msvcrt.dll
        DLL Name: libwinpthread-1.dll
        DLL Name: libuv-2.dll
```

So this _does_ dynamically link to libuv for Windows, and only there.  This also means you have to revert the latest change.  I asked because we also have the static library for libuv, so I don't whether you're doing static or dynamic linking, with static linking you don't need the library at runtime.

L/libwatcher/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libwatcher", :libwatcher),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency("LibUV_jll"),
```
giordano, line 37: Since you dynamically link to libuv for Windows, this _does_ need to be a runtime dependency:
```suggestion
    Dependency("LibUV_jll"),
```

PR #2701, opened 2021-03-17
[FastTransforms] v0.5.1

MikaelSlevinsky:
Fix spin-weighted spherical harmonics with |m| = 1

https://github.com/JuliaApproximation/FastTransforms.jl/issues/139

It's marked draft because I want to try fixing OpenMP on macOS

https://github.com/JuliaApproximation/FastTransforms.jl/issues/136

PR #4650, opened 2022-03-20
[FastTransforms] Use LLVM OpenMP on BSD systems

giordano:
@MikaelSlevinsky you may be interested in this.  It took us some time since
#2701, but LLVM OpenMP is finally available.  Let me know if you're OK with this
change.


MikaelSlevinsky:
If I understand correctly, this uses LLVM OpenMP on all platforms whose native compiler is clang?

giordano:
Correct.  My understanding in #2701 was that you preferred to use LLVM OpenMP for macOS, is it right?

PR #4654, opened 2022-03-20
Upgrade z3 to v4.8.14 and build for all supported platforms.

kllrak:
`]add Z3` fails for me using Julia 1.7, so I thought I would try to update the recipe. This is the first time I have worked on Yggdrasil recipes, comments and scrutiny welcomed. 

There are two patches in here that should be removed at a later date: one to get builds working with Julia 1.6, and another to get builds working for macOS. I'm not quite sure I'm doing the `julia_versions` part right, but I intended to follow the example in the [libcxxwrap_julia](https://github.com/JuliaPackaging/Yggdrasil/blob/82eb18cb6222195b9995b98b136f6f5f995fc721/L/libcxxwrap_julia/build_tarballs.jl#L13) recipe.

### DIFF
--- a/Z/z3/build_tarballs.jl
+++ b/Z/z3/build_tarballs.jl
@@ -1,32 +1,41 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
-using BinaryBuilder
+using BinaryBuilder, Pkg
+
+# See https://github.com/JuliaLang/Pkg.jl/issues/2942
+# Once this Pkg issue is resolved, this must be removed
+uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
+delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "z3"
-version = v"4.8.8"
+version = v"4.8.14"
+julia_versions = [v"1.6.3", v"1.7.0", v"1.8.0", v"1.9.0"]
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


giordano:
Brilliant, thanks a lot!

kllrak:
You're welcome. Out of curiosity, do any of the warnings in the build matter, for example: "Minimum instruction set detected for bin/z3 is avx2, not x86_64 as desired" ?

giordano:
Those checks aren't 100% bullet-proof, but if we have to trust them it means that someone using a CPU without AVX2 will not be able to use the library.  We go a long way to target the base architecture, but if the package really wants to sneak those features... well, we probably can do little about that.  It's still good to know that there may be problems for some users.

kllrak:
Sorry, I have one more question. Is the update to the corresponding JLL package expected to happen automatically, or do I need to trigger it somehow?

giordano:
Ugh, the registration process failed: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=18521&view=logs&j=023ed65f-e065-55fe-4452-05a5186a5335&t=b718c291-7755-5e2a-f9ae-f6d624c03d80.  I'll need to have a look

giordano:
I opened #4664, I believe the issue was the duplication of the non-libcxxwrap platforms, probably a relic from the past where we used to do things in a different way.  Let's see how that goes

kllrak:
Many thanks, I appreciate the help!

giordano:
https://github.com/JuliaRegistries/General/pull/57049

PR #4667, opened 2022-03-22
Build LibYAML also for macOS aarch64

devmotion:
Rebuild LibYAML to add support for macOS aarch64. It seems a `julia_compat` entry is required.

### DIFF
--- a/L/LibYAML/build_tarballs.jl
+++ b/L/LibYAML/build_tarballs.jl
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "LibYAML"
-version = v"0.2.5"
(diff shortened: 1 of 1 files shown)


giordano (changes requested):
This must change also the version number #2763

devmotion:
Ah I see. Slightly unfortunate since it means that the LibYAML_jll and libyaml version will be out of sync. But I guess Yggdrasil doesn't ensure that they are synced anyway.

giordano:
Yes, there have been already cases where we diverged from upstream: https://github.com/JuliaPackaging/Yggdrasil/blob/622cbac55869517b95d05095d28c4836c7d82ddf/C/Cubature/build_tarballs.jl#L4

PR #4672, opened 2022-03-23
[SuperLU_DIST] New Build

Wimmerer:
This isn't ready yet, I need to build the 64 bit `XSDK_INDEX_SIZE`, and split them. I'm going to attempt to do what I did with PETSc by renaming filename and SONAME in subdirectories. Hopefully the PETSc build system is resilient to those renamings (I *believe* it is, you can do `--with-lib=...`).

We do, also, have a Windows issue with PARMETIS that I need to investigate.


Wimmerer:
That's a new error. Hmm

S/SuperLU_DIST/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "SuperLU"
```
giordano, line 5: The directory name must match _**exactly**_ this name

Wimmerer:
That's better, now I just need to fix the Windows build (I don't think they recommend using Parmetis there, which is good), and then build the 32-bit version.

giordano:
For the PowerPC error
```
[19:35:26] /workspace/destdir/lib/libopenblas.so: undefined reference to `GOMP_parallel@GOMP_4.0'
[19:35:26] /workspace/destdir/lib/libopenblas.so: undefined reference to `omp_get_num_places@OMP_4.5'
```
you need GCC 5 or 6, I don't remember exactly which one (I think the minimum that'd work is 5)

S/SuperLU_DIST/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+dependencies = [
+    Dependency(PackageSpec(name="OpenBLAS32_jll", uuid="656ef2d0-ae68-5445-9ca0-591084a874a2")),
+    Dependency(PackageSpec(name="PARMETIS_jll", uuid="b247a4be-ddc1-5759-8008-7e02fe3dbdaa")),
+    Dependency("MPICH_jll"),
+    Dependency("METIS_jll"),
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
boriskaus, line 64:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"9")
```

Wimmerer:
What in the heck.

vchuravy:
Want to turn off parametis for now?

Wimmerer:
Well I think we can just turn it off for Windows, the SuperLU_Dist docs show how. I won't be able to do that for a couple more hours though.

ViralBShah:
I would suggest first building without parmetis, and then trying to enable it later on.

xiaoyeli:
You can disable (par)metis during installation with CMake:
  -DHAVE_PARMEETIS=FALSE

The default ordering is serial Metis.  Since you are not linking with Metis, in your code, you need to set ColPerm option as something else. A good alternative is:
   options->ColPerm = MMD_AT_PLUS_A
(the default is METIS_AT_PLUS_A)

Wimmerer:
Thanks a bunch!

ViralBShah:
IMO, it should be ok to exclude Windows from MPI libraries.

xiaoyeli:
One of the users did a Window port of superlu_dist, with MPI support.  If you are savvy with CMake, you can find the Window related setup in top level CMakeLists.txt and SRC/CMakeLists.txt.

Wimmerer:
> IMO, it should be ok to exclude Windows from MPI libraries.

Unfortunately this project is specifically at the request of someone looking for Windows support for a class (in the Fall I think?).

> One of the users did a Window port of superlu_dist, with MPI support. If you are savvy with CMake, you can find the Window related setup in top level CMakeLists.txt and SRC/CMakeLists.txt.

The current problem seems to be CMake just not picking up MSMPI. I may need to get a CMake guru to help with that in the end. I'm still working on this problem, but I switched to work on some SuiteSparse stuff this week.

ViralBShah:
I am pretty sure most MPI packages are never actually tested on Windows. @xiaoyeli Do you know if SuperLU-Dist works on windows with MSMPI?

I would love to learn more about this class and why do they want MPI on windows on the student's laptops.

boriskaus:
> I am pretty sure most MPI packages are never actually tested on Windows. @xiaoyeli Do you know if SuperLU-Dist works on windows with MSMPI?
> 
> I would love to learn more about this class and why do they want MPI on windows on the student's laptops.

that would be me; I'm teaching a modelling class in the geosciences where students use our MPI-parallel/PETSc-based code  [LaMEM](https://bitbucket.org/bkaus/lamem) to perform simulations. Approximately 90% of the students have windows machines and there is nearly no-one with linux experience. `PETSc_jll` already works on windows in a basic version as does a BinaryBuilder version of our code `LaMEM_jll` that employs the `PETSc_jll` . Unfortunately for most simulations we need a `PETSc_jll` version that includes parallel direct solvers (`SuperLU_Dist` and `MUMPS`). That is something that has been pending since some time.   

MSMPI itself works ok under windows. See, e.g., [MAGEMin_jll](https://github.com/JuliaPackaging/Yggdrasil/blob/master/M/MAGEMin/build_tarballs.jl), which we tested on various machines.

ViralBShah:
@Wimmerer One suggestion would be to get this merged without windows, so that upstream packages can be built, while windows support is worked on in a separate PR.

Wimmerer:
I'm working on the SuiteSparse PR next. We can drop the SuperLU_dist dependency on Windows maybe? And work on Windows further at a later date. I'll rebuild the PRs this evening in light of that plan.

S/SuperLU_DIST/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+make install
+if [[ "${target}" == *-mingw* ]]; then
+    # Manually install the library
+    cp "SRC/libsuperlu.${dlext}" "${libdir}/libsuperlu.${dlext}"
+fi
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = expand_gfortran_versions(supported_platforms())
```
giordano, line 52: Why expanding the libgfortran version?

giordano:
I don't think SuperLU_DIST can be possibly compiled for Windows with MinGW: https://github.com/xiaoyeli/superlu_dist/blob/3810da3a85baf516bfeafc316f1833acfee5b625/SRC/sec_structs.c#L134 `getline` is ***not*** a standard function and is not provided by MinGW.  I don't see how this can work (of course short of manually patching up an implementation of `getline`, which I'm not going to do).

xiaoyeli:
What compiler are you using?

I consulted a successful user, the following is his cmake script.  He uses LLVM / clang compiler.
He said they are using MSMPI, without problem.
(His script was using an older version 5.4.0, but should work with the latest version.

'/winsame/user/dev2/contrib-llvm_vs2017/cmake-3.20.3-ser/bin/cmake.exe' \
  -DCMAKE_INSTALL_PREFIX:PATH=D:/winsame/user/dev2/contrib-llvm_vs2017/superlu_dist-5.4.0-par \
  -DCMAKE_BUILD_TYPE:STRING=Release \
  -DCMAKE_COLOR_MAKEFILE:BOOL=FALSE \
  -DCMAKE_VERBOSE_MAKEFILE:BOOL=TRUE \
  -DBUILD_SHARED_LIBS:BOOL=OFF \
  -DCMAKE_C_COMPILER:FILEPATH='C:/Program Files/LLVM/bin/clang-cl.exe' \
  -DCMAKE_C_FLAGS:STRING='/W3 -std=c99' \
  -DXSDK_ENABLE_Fortran:BOOL='OFF' \
  -Denable_blaslib:BOOL='OFF' \
  -Denable_parmetislib:BOOL='OFF' \
  -DCMAKE_C_FLAGS_RELEASE:STRING='/O2 /Ob2 /DNDEBUG' \
  -DCMAKE_CXX_FLAGS_RELEASE:STRING='/O2 /Ob2 /DNDEBUG' \
  -Denable_openmp:BOOL='FALSE' \
  -G 'NMake Makefiles JOM' \
  D:/winsame/user/vsimall-dev2/builds/superlu_dist-5.4.0

giordano:
The problem is with GCC + MinGW.  See for example https://stackoverflow.com/q/27381903/2442087

xiaoyeli:
@piyush314
'getline' is used only in getFreq() routine, which in turn is not so useful.  Can we remove that?  It causes trouble in gcc + MinGW (Windows)

piyush314:
Yes. Getfreq routine is not used anymore. Earlier it was used with time
stamp counter, to convert it into seconds.

xiaoyeli:
@piyush314 has resolved this issue in master branch.

S/SuperLU_DIST/build_tarballs.jl
giordano: @Wimmerer bump

giordano:
Applying [this upstream patch](https://github.com/xiaoyeli/superlu_dist/commit/f7bf3d9769b98d8206b69e0505648cf1c49a6f7e) we can now build for Windows, thanks!

This now looks good to me, except for the fact I don't know why this is expanding the libgfortran version, I believe we're building multiple duplicates.  Input from @Wimmerer about this choice would be useful.

ViralBShah:
We should just remove the gfortran expansion in here. I think most people are not sure when to do it or copy a recipe from another package.

ViralBShah:
@xiaoyeli Do you have plans to tag a new release of superlu_dist with this fix included? If so, we can just use that instead of carrying the patch here.

Wimmerer:
I believe there was a reason, I did not have it on initial commit I don't think. But locally it no longer seems necessary anyway.

ViralBShah:
Yes - we can definitely wait till the next release since it is quite soon.

ViralBShah:
I think this should be good to merge. We will only find out issues once we can try out these binaries which will happen post merging.

PR #4678, opened 2022-03-24
[hyper] Actually compile the C API

lucifer1004:
According to https://docs.rs/hyper/latest/hyper/ffi/index.html, the C API of hyper is not compiled by default. Extra options and environment variables need to be set to make the C API work.

### DIFF
--- a/H/hyper/build_tarballs.jl
+++ b/H/hyper/build_tarballs.jl
@@ -14,7 +14,7 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/hyper*/
-cargo build --release
+RUSTFLAGS="--cfg hyper_unstable_ffi" cargo build --release --features client,http1,http2,ff
 install -Dm 755 target/${rust_target}/release/*hyper.${dlext} "${libdir}/libhyper.${dlext}"
 """
 
--- a/H/hyper/build_tarballs.jl
+++ b/H/hyper/build_tarballs.jl
@@ -3,18 +3,17 @@
 using BinaryBuilder
 
 name = "hyper"
-version = v"0.14.18"
+version = v"0.14.17"
 
 # Collection of sources required to complete build
 sources = [
     ArchiveSource("https://github.com/hyperium/hyper/archive/refs/tags/v$(version).tar.gz",
-                  "6095636d02ea5af3ff5f06d80b9466f7b58ba76339f39813b33c3f24c663fdef"),
+                  "64420fd550f43af09b0722b3504d4fd919de642d63f01ad54108aa854f5f5470"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/hyper*/
-sed '21 a [lib]\ncrate-type=["cdylib"]\n' -i Cargo.toml
 RUSTFLAGS="--cfg hyper_unstable_ffi" cargo build --release --features client,http1,http2,ffi
 install -Dm 755 target/${rust_target}/release/*hyper.${dlext} "${libdir}/libhyper.${dlext}"
 """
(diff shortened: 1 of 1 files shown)


lucifer1004:
@giordano Please have a look.

lucifer1004:
Seems that v0.14.18 does not emit .so/.dylib files by default.

H/hyper/build_tarballs.jl
```diff
@@ -3,18 +3,19 @@
...
     ArchiveSource("https://github.com/hyperium/hyper/archive/refs/tags/v$(version).tar.gz",
-                  "64420fd550f43af09b0722b3504d4fd919de642d63f01ad54108aa854f5f5470"),
+                  "6095636d02ea5af3ff5f06d80b9466f7b58ba76339f39813b33c3f24c663fdef"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/hyper*/
-cargo build --release
+sed '21 a [lib]\ncrate-type=["cdylib"]\n' -i Cargo.toml
```
giordano, line 17: Isn't there a way to enable building of the library without having to edit the file manually?  Also, this sed script doesn't seem very sustainable as it hardcodes a line number, probability it'll bit-rot at some point is very high.

giordano: https://github.com/hyperium/hyper/commit/1c6637060e36654ddb2fdfccb0d146c7ad527476 seems to suggest you can use `--crate-type`?

lucifer1004: That would require a nightly version of rust. Do we have that available in the environment?

giordano: No :facepalm: Why do they do this? :weary:

lucifer1004: For now, we can switch back to v0.14.17 if the sed script appears ugly.

lucifer1004: I think they made the decision because the C API is unstable, so it is supposed to be used with nightly instead of stable rust.

giordano: Yeah, let's rollback to 0.14.7 and update once we have a new Rust toolchain with this capability

lucifer1004: Done!

giordano:
Thanks!

lucifer1004:
As a note: once the rust version within BinaryBuilder is updated, we can use the following command to build the dynamic library:

```
RUSTFLAGS="--cfg hyper_unstable_ffi" cargo rustc --release --crate-type cdylib -Z unstable-options --features client,http1,http2,ffi
```

giordano:
Does that actually work for you? https://github.com/JuliaPackaging/Yggdrasil/pull/4694#issuecomment-1085148243

lucifer1004:
Yes. You would need nightly rust newer than 1.61.0-nightly (2022-03-02) to make the command work.

giordano:
We don't use nightly 🙂

lucifer1004:
Yes, so we need to wait until 1.61.0 becomes stable.

giordano:
With Rust 1.61.0 I get
```
[22:24:01]  ---> RUSTFLAGS="--cfg hyper_unstable_ffi" cargo rustc --release --crate-type cdylib -Z unstable-options --features client,http1,http2,ffi
[22:24:02] error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel
```

lucifer1004:
https://github.com/rust-lang/cargo/issues/10083

Unfortunately, this API has not been stabilized yet...

giordano:
For the time being, I'm just reverting https://github.com/hyperium/hyper/commit/1c6637060e36654ddb2fdfccb0d146c7ad527476: #5051

PR #4685, opened 2022-03-28
new recipe for booster v0.1.2

BenjaminDoran:


B/booster/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+# Collection of sources required to build SFML
+sources = [
+    GitSource("https://github.com/evolbioinfo/booster.git", "a5f5044b192db7ab5572f44ef46abaea007f5ce9")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/booster
+install_license ${WORKSPACE}/srcdir/booster/LICENSE
+
```
giordano, line 17: The license should be installed automatically, if you want you can remove all of this
```suggestion
```

B/booster/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+# compile booster
+cd ${WORKSPACE}/srcdir/booster/src
+make -j${nproc}
```
giordano, line 20: I don't really like their default choices.  We want to use clang on BSD systems:
```suggestion
make -j${nproc} CC=${CC} CFLAGS_OMP="-Wall -g -fopenmp"
```

B/booster/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("booster", :booster),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+    Dependency("OpenMPI_jll", v"4.1.2"),
```
giordano, line 37: Is MPI used at all?  Doesn't look like it?

B/booster/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("booster", :booster),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
```
giordano, line 36:
```suggestion
dependencies = [
    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
```

B/booster/build_tarballs.jl
BenjaminDoran: The binary needs the potential to be used with MPI, i.e. the binary has an argument for number of cpu's to use, and needs to be compiled with the MPI flags available.

if there is a best practice to get this effect I am open to it.

giordano: I don't see any reference at all to MPI in the code. Are you sure you aren't confused with OpenMP, which, as confusing as it is, it's something completely different?

BenjaminDoran:
Yeah (super confusing), I think you are right. I added a new commit that
removed this dependency. Testing locally it compiled and worked correctly.
It is the latest commit in the pull request which also passed.

-Benjamin

PR #4109, opened 2021-12-25
[HIP] Update to 4.2.0

jpsamaroo:
I still need to test if this can build dependent libraries like rocBLAS, but let's make sure this passes CI.

PR #4689, opened 2022-03-28
[DONOTMERGE] Check if HIP still builds

Keno:
Fails for me locally, so I want to see if it builds on Yggdrasil.

### DIFF
--- a/H/HIP/build_tarballs.jl
+++ b/H/HIP/build_tarballs.jl
@@ -48,6 +48,7 @@ make install
 # Cleanup
 rm ${prefix}/tools/clang
 
+
 # link to .hipVersion so clang can find it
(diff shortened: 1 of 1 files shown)


Keno:
Ok. The failure here matches what I see locally. @jpsamaroo You probably know better than me what needs to be fixed.

giordano:
I don't know if that's the culprit here, but we upgraded cmake since #4109

jpsamaroo:
I would guess we probably need to pull in rocm_cmake: https://github.com/JuliaPackaging/Yggdrasil/tree/master/R/rocm_cmake

giordano:
#4690 fixed it

giordano:
I presume this can be closed now.

PR #4703, opened 2022-04-03
Update FLAC version to v1.3.4

abhayap:


abhayap:
Thanks @giordano!

giordano:
Thanks for starting it 🙂

PR #4704, opened 2022-04-03
[MKL_Headers] Added recipe

stemann:


giordano:
Good to have the headers, but I wonder if it may be better to have them in a separate package, like Xorg_xorgproto, since this is what they do in anaconda too (and we save users from installing unneeded stuff)

stemann:
> Good to have the headers, but I wonder if it may be better to have them in a separate package, like Xorg_xorgproto, since this is what they do in anaconda too (and we save users from installing unneeded stuff)

And in Debian. Yeah, I was wondering about that too - whether to stick all that not-needed-at-runtime-stuff in separate packages: Is there a preferable naming suffix that can be applied more than once? `SomePkgDev_jll`, `SomePkgDevel_jll`, `SomePkgInclude_jll`, `SomePkgHeaders_jll`?

giordano:
We don't really have a policy about this, but we have

* Vulkan_Headers
* SPIRV_Headers
* OpenCL_Headers
* oneAPI_Level_Zero_Headers

so maybe use `*_Headers`?

stemann:
> We don't really have a policy about this, but we have
> 
> * Vulkan_Headers
> * SPIRV_Headers
> * OpenCL_Headers
> * oneAPI_Level_Zero_Headers
> 
> so maybe use `*_Headers`?

Done

giordano:
The license file is missing

stemann:
> The license file is missing

Right - fixed.

PR #4705, opened 2022-04-04
Update libsndfile to v1.1.0

abhayap:


abhayap:
@giordano could you take a look when you get the chance please?

L/libsndfile/build_tarballs.jl
```diff
@@ -51,6 +51,8 @@ dependencies = [
     Dependency("libvorbis_jll"),
     Dependency("Ogg_jll"),
     Dependency("Opus_jll"),
+    Dependency("mpg123_jll"),
+    Dependency("LAME_jll"),
```
giordano, line 55: It doesn't look like these packages are actually used.  From the log:
```
[02:43:23] -- The following OPTIONAL packages have not been found:
[02:43:23] 
[02:43:23]  * Mpg123 (required version >= 1.25.10)
[02:43:23]  * Speex, an audio codec tuned for speech, <www.speex.org/>
[02:43:23]    Enables experemental Speex support
[02:43:23]  * SQLite3, light weight SQL database engine., <www.sqlite.org/>
[02:43:23]    Enables regtest
[02:43:23] 
[02:43:23] -- The following RECOMMENDED packages have not been found:
[02:43:23] 
[02:43:23]  * Lame, High quality MPEG Audio Layer III (MP3) encoder, <[https://lame.sourceforge.io/>](https://lame.sourceforge.io/%3E)
[02:43:23]    Enables MPEG layer III (MP3) writing support
```
For some reasons, CMake isn't finding them.

abhayap: I see it gives this: `Add the installation prefix of "Lame" to CMAKE_PREFIX_PATH or set "LAME_DIR" to a directory containing one of the above files.`

giordano: The fact is that we don't have the CMake files for lame and mpg123 anywhere, because they don't seem to support cmake, as far as I understand.  I have little clue at the moment how this is supposed to work.  If you plan to invest time into understanding it, please do, otherwise remove these lines which at the moment are completely useless.

giordano: For what is worth, none of [Archlinux](https://archlinux.org/packages/extra/x86_64/libsndfile/), [Fedora](https://src.fedoraproject.org/rpms/libsndfile/blob/f32/f/libsndfile.spec), [Debian](https://packages.debian.org/source/sid/libsndfile), [spack](https://github.com/spack/spack/blob/7893be7712ed709f6136ac83f49afc3d718d5ddc/var/spack/repos/builtin/packages/libsndfile/package.py) uses lame or mpg123.

abhayap: Thanks for taking a look @giordano. I removed the dependencies. Could you review and merge when ready please?

PR #4001, opened 2021-12-02
Update RadeonProRender

SimonDanisch:

PR #4708, opened 2022-04-04
Update RadeonProRender version, build on macos aarch64

NauticalDrop:
This updates the version of RadeonProRender to [v2.02.12](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderSDK/commit/e284dfac04ee3a142603ba955a2bf8d11a78d945), released in March 2022.  Additionally, it adds Mac M1 as a build target, given that the website claims it can run on M1 chips as well.

### DIFF
--- a/R/RadeonProRender/build_tarballs.jl
+++ b/R/RadeonProRender/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "RadeonProRender"
-version = v"2.2.9"
+version = v"2.02.12"
 
 # Collection of sources required to complete build
 sources = [
@@ -69,4 +69,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat = "1.6")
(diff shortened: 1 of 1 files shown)


R/RadeonProRender/build_tarballs.jl
```diff
@@ -7,7 +7,7 @@ version = v"2.2.9"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderSDK.git", "5916c58f9cc037899e0aa85e5f5a67c6e1ee7b29")
+    GitSource("https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderSDK.git", "e284dfac04ee3a142603ba955a2bf8d11a78d945")
```
giordano, line 10: You probably need to change the version number as well? :slightly_smiling_face:

NauticalDrop: Yes, thanks :D - did not notice that at first.  I also changed the Julia compat as you suggested in https://github.com/JuliaPackaging/Yggdrasil/pull/4001#issuecomment-984564572.

giordano: Cool, so that wasn't impossible after all :sweat_smile:

giordano:
Cool, thanks for your contribution!

giordano:
For the record, the library is **_not_** for aarch64-apple-darwin:
```console
shell> file /Users/mose/.julia/artifacts/b92cab50f632b77cac59e3863c9f522f7fe523eb/lib/libProRenderGLTF.dylib
/Users/mose/.julia/artifacts/b92cab50f632b77cac59e3863c9f522f7fe523eb/lib/libProRenderGLTF.dylib: Mach-O 64-bit dynamically linked shared library x86_64
```

PR #4709, opened 2022-04-05
Add CXSPARSE component to SuiteSparse

kulakovri:


vchuravy:
@giordano remind me is this Suitesparse_jll shipped with Julia or can we use it directly? @kulakovri has a package that he wants to build that depends on this.

vchuravy (approved):

giordano:
This is the stdlib: https://github.com/JuliaLang/julia/tree/dacf9d65aff4668b8fff25957d9aaa2cf03868c8/stdlib/SuiteSparse_jll.  In order to use it you'll also have to update the package in Julia and wait for the next minor version (1.9 at this point)

vchuravy:
**sigh** that's what I feared. Is one of the other Suitespares things usable?

ViralBShah:
Also, in the past - people were unhappy when I added all the SuiteSparse libraries to the build (this was related to the Metis dependency) complaining that it was bloating the Julia distribution. Anyways, we hope to move Sparse out of system image and stdlib for 1.9: https://github.com/JuliaLang/julia/pull/44247#issuecomment-1071226278

This will allow us to build SuiteSparse with all its bells and whistles, and hopefully even the GPU solvers!

Maybe a cheat in the meanwhile could be to update the SuiteSparse_GPU BB recipe or create a separate BB recipe for `SuiteSparseExtras`, sort of like how we have `SuiteSparseGraphBLAS`.

PR #4726, opened 2022-04-07
Patch PETSc build instead of moving files

vchuravy:
@giordano is it possible to relax the requirement that the file must end in `dll`? If not I will need to fix the petsc build system to not produce files like that.

cc: @boriskaus

### DIFF
--- a/P/PETSc/build_tarballs.jl
+++ b/P/PETSc/build_tarballs.jl
@@ -24,17 +24,22 @@ if [[ "${target}" == *-mingw* ]]; then
 else
     MPI_LIBS="[${libdir}/libmpifort.${dlext},${libdir}/libmpi.${dlext}]"
 fi
+
+atomic_patch -p1 $WORKSPACE/srcdir/patches/mingw-version.patch
+atomic_patch -p1 $WORKSPACE/srcdir/patches/sosuffix.patch
+
 mkdir $libdir/petsc
 build_petsc()
 {
 
+    PETSC_CONFIG="${1}_${2}_${3}"
     if [[ "${3}" == "Int64" ]]; then
         USE_INT64=1
     else
         USE_INT64=0
     fi
-    mkdir $libdir/petsc/${1}_${2}_${3}
-    ./configure --prefix=${libdir}/petsc/${1}_${2}_${3} \
+    mkdir $libdir/petsc/${PETSC_CONFIG}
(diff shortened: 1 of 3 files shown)


giordano:
I don't know, the name parsing is in Julia Base, I don't remember the details of the code

P/PETSc/build_tarballs.jl
```diff
@@ -99,15 +99,21 @@ build_petsc double real Int64
 build_petsc single real Int64
 build_petsc double complex Int64
 build_petsc single complex Int64
+
+# On windows move back since we can't change the install name
+if [[ "${target}" == *-mingw32* ]]; then
+   mv ${libdir}/petsc/double_real_Int32/lib/libpetsc_double_real_Int32.${dlext} "${libdir}/petsc/double_real_Int32/lib/libpetsc.${dlext}.3.6.5" # TODO fix version suffix
```
giordano, line 105: The problem I have with this is: if we do the same for all variants, how do we distinguish between all of them if they have all the same name?

vchuravy: Apparently Windows cares about the path being loaded from and not just the so name.
But yes this was just aimed at unblocking one particular config.

giordano: If we have all the libraries called libpetsc.dll and we have a downstream library linking to libpetsc.dll, which one will actually link to?

vchuravy: That's anyone's guess. Right now we solve it by having no library be able to link against PETSc on windows xD

giordano: > That's anyone's guess.

Eh, that's my problem with PETSc 🙂 

> Right now we solve it by having no library be able to link against PETSc on windows xD

Sure, I understand your concern, it's just that I believe this is really something that should be fixed upstream.

vchuravy:
cc: @Wimmerer can you have a look if this makes sense? We will need to rebuild other packages dependning on PETSc_jll but at least we shouldn't need to muck about with patchelf&co.

Wimmerer:
Is this modifying both the filename and SONAME on Linux? Otherwise LGTM (aside from calling from Julia? I assume Julia can handle `dlopen`ing with a non-numeric suffix?)

vchuravy:
> Is this modifying both the filename and SONAME on Linux?

Should be! 

> I assume Julia can handle dlopening with a non-numeric suffix?)

Yeah these are all fine.

PR #4727, opened 2022-04-07
New Recipe: AWSLC v1.0.2

Octogonapus:
This is my WIP PR to build [AWS-LC](https://github.com/awslabs/aws-lc).

PR #4728, opened 2022-04-07
New Recipe: AWS_MQTT v0.1.0

Octogonapus:
This builds a statically linked version of aws-c-mqtt. I am using static linkage here because a dependency, [AWS-LC](https://github.com/JuliaPackaging/Yggdrasil/pull/4727), will conflict with OpenSSL.

### DIFF
--- a/A/AWS_MQTT/build_tarballs.jl
+++ b/A/AWS_MQTT/build_tarballs.jl
@@ -111,7 +111,7 @@ cmake --build . -j${nproc} --target install
 cd $WORKSPACE/srcdir/aws-c-mqtt
 
 # Patch the aws-c-mqtt CMakeLists.txt file so that we can build it as a shared library with static library dependencies
-atomic-patch -p1 "${WORKSPACE}/srcdir/patches/mqtt_cmake.patch"
+atomic_patch -p1 "${WORKSPACE}/srcdir/patches/mqtt_cmake.patch"
 
 mkdir build && cd build
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
--- a/A/AWS_MQTT/build_tarballs.jl
+++ b/A/AWS_MQTT/build_tarballs.jl
@@ -122,6 +122,8 @@ cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
 	-DBUILD_SHARED_LIBS=ON \
 	..
 cmake --build . -j${nproc} --target install
+
+install_license ${WORKSPACE}/srcdir/license/LICENSE
 """
 
 # These are the platforms we will build for by default, unless further
--- a/A/AWS_MQTT/build_tarballs.jl
+++ b/A/AWS_MQTT/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "AWS_MQTT"
-version = v"0.1.0"
+version = v"0.1.0-debug"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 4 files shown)


A/AWS_MQTT/build_tarballs.jl
```diff
@@ -0,0 +1,139 @@
...
+cmake --build . -j${nproc} --target install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    FileProduct(["lib64/libaws-c-mqtt.a"], :libawsmqtt),
```
giordano, line 128: If at the end you get only a static library, how are you going to use it from Julia?  If you're planning to use it from Julia.

A/AWS_MQTT/build_tarballs.jl
```diff
@@ -0,0 +1,139 @@
...
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    # TODO: this is needed only for Windows, but it looks like filtering
+    # platforms for `HostBuildDependency` is broken
+    HostBuildDependency("NASM_jll")
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"10.2.0")
```
giordano, line 139: Isn't there an older version of GCC that can build this?  10 is relatively recent, we prefer older versions for maximum compatibility.

Octogonapus: I can try older versions, but these AWS libs explicitly only support toolchains up to 5 years old.

A/AWS_MQTT/build_tarballs.jl
Octogonapus: You are right, I will have to make a shared library at the end of this. I haven't figured out how to configure CMake to do that, yet. Simply configuring the aws-c-mqtt build to output a shared library didn't work.

Octogonapus: If I build all of these as shared libraries, that will give me the same trouble as in #4727, right? So my goal is to build the dependencies of aws-c-mqtt as static libraries, and build aws-c-mqtt as a shared library.

giordano: Yeah, I guess you want a shared library only for AWS MQTT

giordano:
For the record, for PowerPC
```
[20:21:21] In file included from /workspace/srcdir/aws-lc/crypto/fipsmodule/bcm.c:41:
[20:21:21] /workspace/srcdir/aws-lc/crypto/fipsmodule/cpucap/cpu_ppc64le.c: In function ‘OPENSSL_cpuid_setup’:
[20:21:21] /workspace/srcdir/aws-lc/crypto/fipsmodule/cpucap/cpu_ppc64le.c:28:38: error: ‘AT_HWCAP2’ undeclared (first use in this function); did you mean ‘AT_HWCAP’?
[20:21:21]    28 |   OPENSSL_ppc64le_hwcap2 = getauxval(AT_HWCAP2);
[20:21:21]       |                                      ^~~~~~~~~
[20:21:21]       |                                      AT_HWCAP
[20:21:21] /workspace/srcdir/aws-lc/crypto/fipsmodule/cpucap/cpu_ppc64le.c:28:38: note: each undeclared identifier is reported only once for each function it appears in
```
you need something like https://github.com/JuliaPackaging/Yggdrasil/blob/42f62534eaf8bcdc6b4948e1227e62ebb2cd8f57/F/FLAC/build_tarballs.jl#L19-L21

giordano:
The license file is missing.  And since you're doing static linking, you likely need to add the licenses of _all_ projects.

Octogonapus:
Closing in favor of https://github.com/JuliaPackaging/Yggdrasil/pull/4797

PR #4730, opened 2022-04-07
Update tree-sitter lib to v0.20.6

aviks:
cc: @MichaelHatherly 

Do the grammars need to be updated simultaneously?


MichaelHatherly:
From what I recall it depends on the ABI version that the parser and grammers are using.

PR #4732, opened 2022-04-08
Wizard recipe: SHTns-v3.5.1

dubosipsl:

* Package name: SHTns
* Version: v3.5.1

### DIFF
--- a/S/SHTns/build_tarballs.jl
+++ b/S/SHTns/build_tarballs.jl
@@ -1,0 +1,54 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "SHTns"
+version = v"3.5.1"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://bitbucket.org/nschaeff/shtns/downloads/shtns-$(version).tar.gz", "77f8a33b94df8786d2ce9b95cbfbe548f00443625b310b8c64012b22c8a7394f")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd shtns/
+export CFLAGS="-fPIC"
+ ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j ; make install
+gcc -shared -o libshtns.so *.o -lfftw3
+cp libshtns.so $prefix/lib/
+install_license LICENSE
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "freebsd"; )
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libshtns", :LibSHTns)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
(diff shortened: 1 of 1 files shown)


S/SHTns/build_tarballs.jl
```diff
@@ -0,0 +1,54 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd shtns/
+export CFLAGS="-fPIC"
+ ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j ; make install
+gcc -shared -o libshtns.so *.o -lfftw3
+cp libshtns.so $prefix/lib/
+install_license LICENSE
```
giordano, line 22:
```suggestion
cd $WORKSPACE/srcdir/shtns/
export CFLAGS="-fPIC"
./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
make -j${nproc}
make install
mkdir -p ${libdir}
gcc -shared -o "${libdir}/libshtns.${dlext}" *.o -lfftw3
install_license LICENSE
```

S/SHTns/build_tarballs.jl
```diff
@@ -0,0 +1,54 @@
...
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "freebsd"; )
+]
```
giordano, line 40:
```suggestion
platforms = supported_platforms()
```

dubosipsl:
This recipe builds SHTns, a high-performance library for spherical harmonics transform written by Nathanael Schaeffer. 
https://bitbucket.org/nschaeff/shtns/src/master/

I have not played much with compiler options and am not sure how much optimization is turned on.

I have tested that the resuting shared object is indeed usable from Julia. My tests suggest SHTns is an order of magnitude faster than FastTransforms for scalar transforms (not tested vector transforms yet).

S/SHTns/build_tarballs.jl
dubosipsl: Hello @giordano, thanks for taking the time to review this recipe.
Will this test all platforms ? The build fails on some platforms, the list was produced by the wizard after trying all platforms except Mac. If you prefer to see it by yourself, no problem, I'll do the commit.

giordano: Yes, let's see what happens on all platforms :slightly_smiling_face:

dubosipsl:
Building SHTns on apple-darwin works, but creating the shared object fails.
Note that SHTns builds only a static library. I have 'invented' the line to create a shared object from the '*.o' produced by the build. I am not familiar with the proper way to do this on darwin.

S/SHTns/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/shtns/
+export CFLAGS="-fPIC"
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j${nproc}
+make install
+mkdir -p ${libdir}
+gcc -shared -o "${libdir}/libshtns.${dlext}" *.o -lfftw3
```
giordano, line 21:
```suggestion
cc -shared -o "${libdir}/libshtns.${dlext}" *.o -lfftw3
```

dubosipsl: There is a bunch of undefined symbols, e.g. `__sy11_l`. Similar symbols are found in `sht_kernels_s.so`, e.g. __sy12_l` .
The missing symbols are present in this file when I build on linux. I also checked that the generated code is the same, so this must be something related to CPP macros and conditional compilation. This is quite a bit beyond my expertise.

giordano: That's why I'm suggesting to compile with `cc` (which on macOS is Clang, not GCC) instead of `gcc`, since the rest of the compilation was done with Clang, not GCC.  Mixing up compilers like that isn't good, and may well be the culprit here.

giordano:
This is now working on all platforms 🙂   Thanks for your contribution!

dubosipsl:
Thanks @giordano for the fix. I must have missed something since I did try on my machine to cross-compile for darwin, and using `cc` instead of `gcc` to produce the dynamic library did not seem to help. Glad to see it did, in fact ... Looking forward to using shtns with Julia.

giordano:
> I must have missed something since I did try on my machine to cross-compile for darwin, and using cc instead of gcc to produce the dynamic library did not seem to help.

The problem was that the build system stubbornly decided to use GCC to build just some object files, creating a poisonous mix of some object files compiled with gcc, and others with clang. The solution for me was to always force using `cc` instead of `gcc`, although if I had time I'd be curious to understand why that was happening in the first place.

PR #4736, opened 2022-04-09
[Tesseract] Upgrade to Tesseract v5.

sunoru:
The libtool files for `libtiff` have been manually included for ease of compilation. I hope this is fine?


giordano:
> The libtool files for libtiff have been manually included for ease of compilation. I hope this is fine?

Not really, we prefer to [_delete_ them](https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1105), adding them goes exactly in the opposite direction

sunoru:
> > The libtool files for libtiff have been manually included for ease of compilation. I hope this is fine?
> 
> Not really, we prefer to [_delete_ them](https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1105), adding them goes exactly in the opposite direction

I have read that discussion. The use here is only for Tesseract's building process since it is required there. They won't be included in the products. Actually, the building scripts for previous versions will also fail now, because the .la files have been removed in the dependencies.

BTW, do you have any idea why there are two failures in the CI checks? It says:

```ERROR: LoadError: Error when installing package XSLT_jll:
IOError: unlink("/depot/packages/XSLT_jll/4AZBo/src/wrappers/powerpc64le-linux-gnu.jl"): no such file or directory (ENOENT)```

giordano:
How are these files generated?  What I don't like is that this isn't very sustainable, it's very easy for someone to update the version of the package without updating the libtool files correctly, so you get into an inconsistent state.

@fingolfin do you have ideas about how to deal with this situation?

giordano:
Maybe you can use CMake instead of autoconf and completely sidestep the libtool issue?

giordano:
Oh, I think it's again an issue with another package referencing a libtool file.  I think it's leptonica.

giordano:
Should have been fixed by #4737 (in which I found a small bug in our removal of libtool files: https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1199).  Of course you'll have to wait for the new version to make it to the registry before trying again.

fingolfin:
@giordano seems you've figure it out. Annoying that it keeps making trouble :-(. Anyway, I hope that'll be it. But we'll see..

sunoru:
Thanks so much. I will try again later. (BTW, the `.la` files were copied from a previous version of `Libtiff_jll.jl` where `BinaryBuilder.jl` didn't remove them.)

sunoru:
Looks problems solved :smile:

giordano:
Lovely, thanks!

PR #4737, opened 2022-04-09
[Leptonica] Update to v1.82.0

giordano:


L/Leptonica/build_tarballs.jl
```diff
@@ -19,11 +19,13 @@ export CPPFLAGS="-I${includedir}"
 make -j${nproc}
 make install
 install_license leptonica-license.txt
+# Get rid of libtool files
+rm ${prefix}/lib/liblept*.la
```
giordano, line 23: Fixed in BinaryBuilder with https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1199

PR #4753, opened 2022-04-14
Try building a `GCCBootstrap_jll` using `crosstool-ng`

staticfloat:
TODO before merging:

- [X] Cleanup `glibc` patches!  Note that we're building against glibc 2.19
  on all platforms right now; this is fine, because this glibc is not
  what will be used in production, and it mostly affects what version of
  glibc must be available on the _host_ system when compiling, not on the
  _target_ system when running things built with this compiler.  So by
  using 2.19 here, we simply make it impossible to use BinaryBuilder on
  ancient CentOS systems, which I'm okay with.

- [X] Build for `musl`

- [X] Build for windows; supposedly we can build against mingw32.

- [x] Fix case-insensitivity issues

- [x] Fix ARM default arch level

- [x] Fix weird `musl` linkage problem:
```
/tmp/bin # ./x86_64-linux-gnu-readelf -d ./x86_64-linux-gnu-gcc

Dynamic section at offset 0x1f7a18 contains 21 entries:
 0x0000000000000001 (NEEDED)             Shared library: [libc.musl-.so.1]
```

UPDATE: This will be fixed by a new RootFS containing: https://github.com/JuliaPackaging/Yggdrasil/pull/4762, which is currently in a PR at https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/237

### DIFF
--- a/G/GCCBootstrap/build_tarballs.jl
+++ b/G/GCCBootstrap/build_tarballs.jl
@@ -1,0 +1,130 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "GCCBootstrap"
+version = v"9.4.0"
+
+# Collection of sources required to complete build
+sources = [
+    # crosstool-ng will provide the build script
+    ArchiveSource("http://crosstool-ng.org/download/crosstool-ng/crosstool-ng-1.25.0_rc1.tar.bz2",
+                  "8a839df71bea7b2c411447b41b917f2142482df171733ddb92d615cec2f0f43a"),
+
+    # We provide some configs for crostool-ng
+    DirectorySource("./bundled"),
+
+    # crosstool-ng can download the files, but we'd rather download them ourselves
+    FileSource("http://mirrors.kernel.org/gnu/gcc/gcc-9.4.0/gcc-9.4.0.tar.xz",
+               "c95da32f440378d7751dd95533186f7fc05ceb4fb65eb5b85234e6299eb9838e"),
+    FileSource("https://mirrors.kernel.org/gnu/mpfr/mpfr-4.1.0.tar.xz",
+               "0c98a3f1732ff6ca4ea690552079da9c597872d30e96ec28414ee23c95558a7f"),
+    FileSource("https://mirrors.kernel.org/gnu/mpc/mpc-1.2.1.tar.gz",
+               "17503d2c395dfcf106b622dc142683c1199431d095367c6aacba6eec30340459"),
+    FileSource("https://gcc.gnu.org/pub/gcc/infrastructure/isl-0.24.tar.bz2",
+               "fcf78dd9656c10eb8cf9fbd5f59a0b6b01386205fe1934b3b287a0a1898145c0"),
+    FileSource("https://mirrors.kernel.org/gnu/gmp/gmp-6.2.1.tar.xz",
+               "fd4829912cddd12f84181c3451cc752be224643e87fac497b69edddadc49b4f2"),
+    FileSource("http://mirrors.kernel.org/pub/linux/kernel/v4.x/linux-4.1.49.tar.xz",
+               "ff2e0ea5c536650aef64447c3aaa49c1a25e8f1db4ec4f7da700d3176f512ba8"),
+    FileSource("https://mirrors.kernel.org/gnu/glibc/glibc-2.19.tar.xz",
+               "2d3997f588401ea095a0b27227b1d50cdfdd416236f6567b564549d3b46ea2a2"),
+    FileSource("https://musl.libc.org/releases/musl-1.2.2.tar.gz",
+               "9b969322012d796dc23dda27a35866034fa67d8fb67e0e2c45c913c3d43219dd"),
+    FileSource("https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v9.0.0.tar.bz2",
+               "1929b94b402f5ff4d7d37a9fe88daa9cc55515a6134805c104d1794ae22a4181"),
+    FileSource("https://github.com/madler/zlib/archive/refs/tags/v1.2.11.tar.gz",
+               "629380c90a77b964d896ed37163f5c3a34f6e6d897311f1df2a7016355c45eff",
+               filename="zlib-1.2.11.tar.gz"),
+    FileSource("http://mirrors.kernel.org/gnu/ncurses/ncurses-6.2.tar.gz",
+               "30306e0c76e0f9f1f0de987cf1c82a5c21e1ce6568b9227f7da5b71cbea86c9d"),
+    FileSource("http://mirrors.kernel.org/gnu/libiconv/libiconv-1.16.tar.gz",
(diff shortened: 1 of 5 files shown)


G/GCCBootstrap/build_tarballs.jl
```diff
@@ -0,0 +1,107 @@
...
+# Generate the appropriate crosstool-ng config file for our current target
+${WORKSPACE}/srcdir/gen_config.sh > .config
+./ct-ng upgradeconfig
+./ct-ng build
+[[ -f "${bindir}/${bb_target}-gcc" ]]
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = filter(p -> Sys.islinux(p), supported_platforms())
```
giordano, line 95:
```suggestion
platforms = filter(Sys.islinux, supported_platforms())
```
Why only Linux?  What for the other systems?

G/GCCBootstrap/build_tarballs.jl
```diff
@@ -0,0 +1,107 @@
...
+sources = [
+    # crosstool-ng will provide the build script
+    ArchiveSource("http://crosstool-ng.org/download/crosstool-ng/crosstool-ng-1.25.0_rc1.tar.bz2",
+                  "8a839df71bea7b2c411447b41b917f2142482df171733ddb92d615cec2f0f43a"),
+
+    # We provide some configs for crostool-ng
+    DirectorySource("./bundled"),
+
+    # crosstool-ng can download the files, but we'd rather download them ourselves
+    FileSource("http://mirrors.kernel.org/gnu/gcc/gcc-9.4.0/gcc-9.4.0.tar.xz",
```
giordano, line 18: Why these are FileSources instead of ArchiveSources? crosstools-ng expects the unpacked archives?

giordano (commented):
First round of comments

giordano:
> So by
using 2.19 here, we simply make it impossible to use BinaryBuilder on
ancient CentOS systems, which I'm okay with.

To be clear: this isn't the compiler that will be used in BinaryBuilder? Otherwise my concern is that we're actively going to break existing packages for users on old systems, which isn't great.

G/GCCBootstrap/build_tarballs.jl
staticfloat: I will add Windows for sure, but I don't think FreeBSD or macOS are supported by crosstool-ng.  I may try to add support, but that's going to have to wait for a while.   See my overall comment about the purpose of this package for more details.

G/GCCBootstrap/build_tarballs.jl
staticfloat: Yes, precisely.  crosstool-ng wants to unpack and patch the sources itself.

G/GCCBootstrap/build_tarballs.jl
```diff
@@ -0,0 +1,107 @@
...
+
+# Build crosstool-ng for the current host
+./configure --enable-local
+make -j${nproc}
+
+# Generate the appropriate crosstool-ng config file for our current target
+${WORKSPACE}/srcdir/gen_config.sh > .config
+./ct-ng upgradeconfig
+./ct-ng build
+[[ -f "${bindir}/${bb_target}-gcc" ]]
```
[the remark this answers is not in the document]
staticfloat: I think this is a leftover from when I was writing scripts that used `target` internally.

staticfloat:
These are all great questions.  Let me start by giving more of a history of how I got here:

* In the beginning, I wanted to compile a nice little standalone `GCC_jll` that could be hooked together with a `Glibc_jll` and a `Binutils_jll`, and a `LinuxKernelHeaders_jll`, etc...  That work is sitting here: https://github.com/JuliaPackaging/Yggdrasil/tree/sf/gcc but bootstrapping is such a giant pain in the neck that I wanted to give up the complexity of bootstrapping to instead focus on simply building each product in isolation.  This _vastly_ reduces the amount of work necessary to get things working, but it introduces a new dependency: we need a base C compiler and libc that are already compiled for the target platform.  To be precise, we need a `build -> host` compiler in order to build a `host -> target` compiler.  We already have one of those for all of our current platforms, so I could generate `GCC_jll` packages, but then when we want to add a new platform, we'd be in trouble.  For this reason, I realized we'll never truly escape the bootstrapping problem.

* I thought about reverting back to the bootstrapping configuration we've had until now, but rebelled at the thought.  Then I discovered crosstool-ng, and realized that I could separate concerns here: I can use ct-ng to build a working cross-toolchain for each target, then use that compiler to do a much simpler build for all of the other components.  Therefore, I extract the work of getting a bootstrap build onto crosstool-ng, and then use that to do whatever other builds I want in the presence of a fully-functional cross-compiler.

* This also breaks the need for the initial bootstrap to be quite as restricted as the target toolchain.  The GCC that we build technically doesn't need to run everywhere, it just needs to generate code that runs everywhere.  We can build GCC against glibc 2.19, and then at build time have it link the code it generates against glibc 2.12.2, and that will work just fine for BB.  The compiler may be using a newer glibc to run, but when it builds, it uses whatever glibc is placed within the target sysroot.

* In principle, we don't actually need a GCCBootstrap at all.  We only need a functional cross-compiler.  Theoretically, we could use Clang to do the bootstrapping.  But I'm not going to fully embrace that because I know that compiling Glibc with Clang is a pain, so for `*-linux-gnu` at the very least, we're not going to attempt that.  On macOS and FreeBSD however, there is a possibility that we can use Clang as our "bootstrap compiler" in order to generate the actual GCC_jlls.

* Therefore, my proposed "compiler shard creation flow" for this is as follows:

1) Bootstrapping step: Create `GCCBootstrap_jll` and `ClangBootstrap_jll`.  These are fully-self-contained compiler shards that are used only by our future `GCC_jll` and `Clang_jll` recipes.
2) Use `GCCBootstrap_jll` to compile all the various versions of `GCC_jll`, `Glibc_jll`, `Binutils_jll`, etc...
3) Use `GCC_jll`, `Glibc_jll`, `Binutils_jll`, etc... anywhere we need a compiler, including Base CI, BinaryBuilder, etc...  In [my branch](https://github.com/JuliaPackaging/Yggdrasil/tree/sf/gcc), I have a sketch of what these recipes look like, as well as [a branch of BBB](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/compare/sf/gcc_bootstrap_capabilities?expand=1) that adds a `CrossPlatform <: AbstractPlatform` that makes it easy to describe an artifact that has a `host` and a `target`.  With this, we will be able to do things like just download `GCC_jll`, then ask it for a particular `CrossPlatform` and get an artifact directory back out.  And this is precisely what BB will do.

G/GCCBootstrap/build_tarballs.jl
```diff
@@ -0,0 +1,107 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "GCCBootstrap"
+version = v"9.4.0"
```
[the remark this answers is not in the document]
staticfloat: As explained in my comment; these aren't the final compilers that are going to be used by BB. :)

staticfloat:
Merging for now; will work on ClangBootstrap later.

PR #4755, opened 2022-04-14
Add MPIPreferences as a dependency and rebuild LAMMPS

vchuravy:


vchuravy:
cc: @eschnett @giordano

vchuravy:
@giordano is anyone going to get annoyed at me at adding a dependency without tagging a new version?

PR #4762, opened 2022-04-15
Fix musl SONAME

staticfloat:
Without this, our default `musl` in the BB rootfs has the SONAME `libc.musl-.so.1`:

```
# readelf -d /lib/libc.musl-x86_64.so.1 

Dynamic section at offset 0x94928 contains 18 entries:
 0x000000000000000e (SONAME)             Library soname: [libc.musl-.so.1]
```

Note that this doesn't really affect things built with BB, it only affects things built by an `apk`-supplied GCC.

### DIFF
--- a/M/Musl/build_tarballs.jl
+++ b/M/Musl/build_tarballs.jl
@@ -29,7 +29,7 @@ musl_arch()
     esac
 }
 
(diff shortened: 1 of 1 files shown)


staticfloat:
Confirmed that the issue is fixed here.

PR #4765, opened 2022-04-15
New recipe: High Performance Tensor Transposition library (v1.0.5)

Jutho:
This adds the HPTT library (High Performance Tensor Transposition), original version here:
https://github.com/springer13/hptt

I plan to use this library in forthcoming versions of [TensorOperations.jl](https://github.com/Jutho/TensorOperations.jl), and it can also be useful to other tensor libraries.

The patch fixes compatibility with clang, and corresponds to the changes of PR https://github.com/springer13/hptt/pull/16 .

While the library promises dedicated optimisations for the ARM platform, these lead to compiler issues ( believe the original source file does contain bugs or is broken), so I decided to fall back to the general build flags on this platform and to disable specific optimisations.

On Intel platforms, I enable avx optimisations, which yields a warning that I cannot assume these to be present. The library does not support switching avx on or off at runtime, so the only way around would be to have separate builds based on the availability of AVX, which I do not know how to realise using BinaryBuilder. I assume most users would actually have AVX available.

### DIFF
--- a/H/HPTT/build_tarballs.jl
+++ b/H/HPTT/build_tarballs.jl
@@ -1,0 +1,52 @@
+using BinaryBuilder, Pkg
+
+name = "HPTT"
+version = v"1.0.5"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/springer13/hptt.git", "a55c2a927d5462e81abeb12081fd345024caf5f6"),
+    DirectorySource("./bundled"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+atomic_patch -p1 $WORKSPACE/srcdir/patches/clang_compatibility.patch
+mkdir -p ${libdir}
+mkdir -p ${includedir}
+export hpttdir=${WORKSPACE}/srcdir/hptt
+cp ${hpttdir}/include/* ${includedir}
+export CXXFLAGS="-O3 -std=c++11 -DNDEBUG -fopenmp -fPIC"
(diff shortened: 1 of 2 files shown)


giordano:
Wow, the recipe is pretty well-done for a first-time contributor!

> so the only way around would be to have separate builds based on the availability of AVX, which I do not know how to realise using BinaryBuilder.

We actually do have the means to target multiple microarchitectures, but we need to still fix a few minor bugs in the auditor (for example https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1194 and https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/233) in addition to fixing what build Pkg should fall back on when the current system doesn't match the available microarchitectures.  Also, it was suggested to revise the sets and names of microarchitectures we should target: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/233#pullrequestreview-919580905

Jutho:
Thanks, this went smooth. It took inspiration from a related package (TBLIS), but it still took me about a day to get everything correct and tested (mostly because of the rather primitive build process of this particular library).

One issue I noticed is that, in `run_wizard()`, the github repository was cloned into `${WORKSPACE}/srcdir/hptt.git`, but then when actually running `build_tarballs` this was `${WORKSPACE}/srcdir/hptt.git` so I had to modify the script.

giordano:
Ugh.  Would you mind filing an issue in https://github.com/JuliaPackaging/BinaryBuilder.jl?

PR #4768, opened 2022-04-15
Update DynarePreprocessor to v6.0.0

MichelJuillard:


giordano:
Did this fix the issue on macOS you mentioned to me?

MichelJuillard:
Yes @giordano . It turns out that streambuffer::in_avail() returns 0 under macOS rather than the length of the buffer as in the other operating systems. I made a patch to evaluate the length of the buffer differently, using peek(). I will send you a small example tomorrow.

giordano:
Awesome!  I'm going to restart CI after we updated BinaryBuilder with #4769 and then I'll merge the PR.  Thanks!

giordano:
Side note: it isn't a good practice to open a pull request from a branch with the same name as the branch you want to merge into.  You should always create a _new_ branch for each new pull request.

PR #4771, opened 2022-04-16
Setup buildkite as an alternative to AZP

vchuravy:


.buildkite/generator.jl
```diff
@@ -0,0 +1,15 @@
+include("utils.jl")
+
+pipeline = """
+steps:
+  - label: "test"
+    agents:
+      queue: "juliagpu"
```
giordano, line 7: Note that Yggdrasil requires a very large amount of storage, I'm not sure it should eventually run on the same hardware as the rest of the ecosystem.

vchuravy: Oh for sure, I just wanted something to start iterating with. The other issue is to solve caching.

So yes we will need a new queue, probably.

vchuravy:
Ah great we can also do logs extraction to create warnings https://buildkite.com/docs/agent/v3/cli-annotate

giordano:
Some comments about what we currently do in Azure Pipelines.  In terms of race conditions:

* the JLL repo init job is ***not*** safe: in some cases we can build multiple versions of the same package (when they're split in different directories, e.g. CUDA, OpenBLAS or LLVM)
* the build and package registration jobs are safe, however in some cases registration may cause merge conflicts in the registry (but we have relatively little control over that).  Note that currently we do registrations in sequence (#208)

Note that buildkite is currently running, and failing, also on `master`: https://buildkite.com/julialang/yggdrasil/builds/4#d85857dd-92f6-43cd-82b5-2827d84e5602.  It'd be nice to disable it until this is working

vchuravy:
> It'd be nice to disable it until this is working

Done.

.buildkite/utils.jl
```diff
@@ -0,0 +1,88 @@
...
+group_step(name, steps) = Dict(:group => name, :steps => steps)
+
+function jll_init_step(NAME, PROJECT, BB_HASH, PROJ_HASH)
+    Dict(
+        :label => "jll_init -- $NAME",
+        :agents => agent(),
+        :plugins => plugins(),
+        :timeout_in_minutes => 60,
+        :concurrency => 1,
+        :concurrency_group => "yggdrasil/jll_init",
```
DilumAluthge, line 56: This means that even if there are multiple PRs open at the same time, each with their own `init` step, the `init` steps from different PRs will not be allowed to run at the same time. Is this the intended behavior here?

giordano: The init step shouldn't run at all in pull requests

vchuravy: Yeah since builds to `LLVM_full_jll@12` and `LLVM_full_jll@13` will conflict. 
And while we could do some filter magic and remove the version number, a global concurrency limit shouldn't hurt either.

DilumAluthge: Can we add a Buildkite conditional to the `init` step of the form:

```
if: build.branch == "master"
```

DilumAluthge: Ah, I guess you can just check that in your `generator.jl` script.

vchuravy: Yeah I decided to just parameterize it in the generator

.buildkite/generator.jl
```diff
@@ -0,0 +1,229 @@
...
+    @assert !isempty(BASE_BRANCH)
+
+    PR_NUMBER = parse(Int, ENV["BUILDKITE_PULL_REQUEST"])
+    TARGET_BRANCH = "remotes/origin/$(BASE_BRANCH)"
+    COMPARE_AGAINST = chomp(read(`git merge-base --fork-point $(TARGET_BRANCH) HEAD`, String))
+
+    exec(`git fetch origin "refs/pull/$(PR_NUMBER)/head:refs/remotes/origin/pr/$(PR_NUMBER)"`)
+
+    #if [[ "$(git show -s --format=%B origin/pr/$(System.PullRequest.PullRequestNumber))" == *"${SKIP_BUILD_COOKIE}"* ]]; then
+    #   SKIP_BUILD="true"
```
vchuravy, line 31: @giordano is this for rerunning registration without rebuilding binaries?

giordano: Yes, this _only_ updates the JLL wrappers, without touching the binaries.  See how it's used in https://github.com/JuliaPackaging/Yggdrasil/blob/428fbb633b693a20bc4ef473adcc62cc8003f567/.ci/register_package.jl#L96-L133

vchuravy: ah good to know that we need it in the environment as well!

.buildkite/generator.jl
```diff
@@ -0,0 +1,229 @@
...
+
+    # "project source hash" is a combination of meta.json (to absorb
+    # changes from include()'ing a `common.jl`) as well as the entire
+    # tree the project lives in (to absorb changes from patches)
+    # In order to support symlinked directories as dependencies we calculate
+    # the tree hash on a TMP_PROJECT using `cp -RL` to resolve the symlink and
+    # hash the actual content.
+
+    TMP_PROJECT = mktempdir()
+    exec(`cp -RL $(PROJECT) $(TMP_PROJECT)`)
```
giordano, line 171:
```suggestion
    cp(PROJECT, TMP_PROJECT; follow_symlinks=true)
```
?

.buildkite/generator.jl
```diff
@@ -0,0 +1,229 @@
...
+    TREE_HASH = julia(`-e "using Pkg; print(bytes2hex(Pkg.GitTools.tree_hash(\"$(TMP_PROJECT)\")))"`) |> cmd->read(cmd, String)
```
giordano, line 172:
```suggestion
    TREE_HASH = julia(`-e "using Pkg; print(bytes2hex(Pkg.GitTools.tree_hash(\"$(TMP_PROJECT)\")))"`) |> readchomp
```
but can probably avoid shelling out and run the command here?

.buildkite/generator.jl
```diff
@@ -0,0 +1,229 @@
...
+    META_HASH = open(io->sha256(io), "$(TEMP)/$(NAME).meta.json") |> bytes2hex
```
giordano, line 173:
```suggestion
    META_HASH = open(sha256, "$(TEMP)/$(NAME).meta.json") |> bytes2hex
```

.buildkite/generator.jl
```diff
@@ -0,0 +1,229 @@
...
+    PROJ_HASH = sha256(TREE_HASH*META_HASH) |> bytes2hex
+
+    # Load in the platforms
+    PLATFORMS = split(read(joinpath(TEMP, "$(NAME).platforms.list"), String))
```
giordano, line 177:
```suggestion
    PLATFORMS = split(readchomp(joinpath(TEMP, "$(NAME).platforms.list")))
```

.buildkite/generator.jl
```diff
@@ -0,0 +1,229 @@
...
+    BUILD_STEPS = Any[]
+    for PLATFORM in PLATFORMS
+        # if [[ "${SKIP_BUILD}" == "true" ]]; then
+        #     echo "The commit messages contains ${SKIP_BUILD_COOKIE}, skipping build"
+        #     break
+        # fi
+
+        # Here, we hit the build cache to see if we can skip this particular combo
+        CACHE_URL = "https://julia-bb-buildcache.s3.amazonaws.com/$(BB_HASH)/$(PROJ_HASH)/$(PLATFORM).tar.gz"
+        CURL_HTTP_CODE = read(`curl --output /tmp/curl_$(PROJ_HASH)_$(PLATFORM).log --silent --include --HEAD "$(CACHE_URL)" --write-out '%{http_code}'`, String)
```
giordano, line 198:
```suggestion
        CURL_HTTP_CODE = readchomp(`curl --output /tmp/curl_$(PROJ_HASH)_$(PLATFORM).log --silent --include --HEAD "$(CACHE_URL)" --write-out '%{http_code}'`)
```

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,23 @@
...
+          persist_depot_dirs: packages,artifacts,compiled
+          version: "1.7"
+    agents:
+      queue: "juliaecosystem"
+      arch:  "x86_64"
+      os:    "linux"
+      # Only run on "sandbox.jl" machines
+      sandbox_capable: "true"
+    if: build.message !~ /\[skip tests\]/
+    timeout_in_minutes: 60
```
giordano, line 14: Is this the timeout of the whole build?  We probably need to increase it a bit.

giordano (commented):
First round of comments.

Reminder that skipping build isn't implemented at the moment.

.buildkite/generator.jl
```diff
@@ -0,0 +1,229 @@
...
+    # Otherwise, emit a build with `PROJECT` set to `${PROJECT}`
+    println(" --> Accepted!")
+    push!(PROJECTS_ACCEPTED, PROJECT)
+end
+
+if isempty(PROJECTS_ACCEPTED)
+    annotate("No projects accepted", style="success")
+    exit()
+end
+
```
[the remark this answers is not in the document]
vchuravy: We now have infinite space!

.buildkite/pipeline.yml
vchuravy: No just the timeout for the generator itself, the other ones have time outs as well

.buildkite/generator.jl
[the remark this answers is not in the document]
giordano: :smile: but most cases where people tried to rebuild too many projects (like, all of them) where real errors

.buildkite/pipeline.yml
giordano: Oh, ok, at the moment we have it at 10 minutes: https://github.com/JuliaPackaging/Yggdrasil/blob/82b61df664f0cae541aa70486366cec0edbc90bc/azure-pipelines.yml#L247

vchuravy:
So the current failure is real and I don't know why and how.
`ERROR: LoadError: failed process: Process(`git merge-base --fork-point remotes/origin/master HEAD`, ProcessExited(1)) [1]`

It seems to occur always when master is ahead of this feature branch, but git does not give a useful error.

giordano:
I guess that's because this doesn't run on the merge commit (which is my gripe with buildkite) :upside_down_face:

DilumAluthge:
When I have some free time, I want to get https://github.com/JuliaCI/pull-request-merge-commit-buildkite-plugin working. Then you would just stick that plugin at the top of your list of plugins, and it would automatically take care of checking out the merge commit.

Don't use it yet though, it doesn't work.

.buildkite/generator.jl
```diff
@@ -0,0 +1,228 @@
...
+exec(cmd) = @assert success(pipeline(cmd, stderr=stderr, stdout=stdout))
+
+if IS_PR
+    # If we're on a PR though, we look at the entire branch at once
+    BASE_BRANCH = get(ENV, "BUILDKITE_PULL_REQUEST_BASE_BRANCH", "master")
+    @assert !isempty(BASE_BRANCH)
+
+    PR_NUMBER = parse(Int, ENV["BUILDKITE_PULL_REQUEST"])
+    TARGET_BRANCH = "remotes/origin/$(BASE_BRANCH)"
+    COMPARE_AGAINST = chomp(read(`git merge-base --fork-point $(TARGET_BRANCH) HEAD`, String))
```
giordano, line 26: Another simplification courtesy of `readchomp`:
```suggestion
    COMPARE_AGAINST = readchomp(`git merge-base --fork-point $(TARGET_BRANCH) HEAD`)
```

vchuravy:
> I guess that's because this doesn't run on the merge commit (which is my gripe with buildkite) 🙃

It's particularly weird since I can't reproduce locally

DilumAluthge:
Alright, the minimum functionality of https://github.com/JuliaCI/merge-commit-buildkite-plugin is now there.

Basically, you just add this to the top of your plugin list:

```
- JuliaCI/merge-commit: ~
```

So e.g.

```yml
steps:
  - plugins:
    - JuliaCI/merge-commit: ~
```

Note: eventually, you'll want to replace `JuliaCI/merge-commit: ~` with `JuliaCI/merge-commit#v1: ~`, but that'll have to wait until we first make a tag on the https://github.com/JuliaCI/merge-commit-buildkite-plugin repo.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,23 @@
+steps:
+  - label: "generator"
+    plugins:
+      - JuliaCI/julia#v1:
+          persist_depot_dirs: packages,artifacts,compiled
+          version: "1.7"
```
DilumAluthge, line 6:
```suggestion
      - JuliaCI/merge-commit: ~
      - JuliaCI/julia#v1:
          persist_depot_dirs: packages,artifacts,compiled
          version: "1.7"
```

You'll also need to prepend this plugin to the plugin list for every step that you generate in your generator.

DilumAluthge: We'll also need to add it to the web UI.

.buildkite/generator.jl
DilumAluthge: In addition to Mosè's suggestion, one way to debug why this line is currently failing would be to do a regular `run` so we can see the full log.

```suggestion
    cmd = `git merge-base --fork-point $(TARGET_BRANCH) HEAD`
    run(cmd)
    COMPARE_AGAINST = readchomp(cmd)
```

DilumAluthge: Or, alternatively, we could only do the regular `run` if the first attempt fails. That way, we don't pollute the logs unless the command is failing.

```suggestion
    cmd = `git merge-base --fork-point $(TARGET_BRANCH) HEAD`
    run(cmd)
    COMPARE_AGAINST = try
        readchomp(cmd)
    catch
        run(cmd)
        rethrow()
    end
```

.buildkite/0_webui.yml
```diff
@@ -0,0 +1,8 @@
+# This file represents what is put into the webUI.
+# It is purely for keeping track of the changes we make to the webUI configuration; modifying this file has no effect.
+steps:
+  - label: ":buildkite: Pipeline upload"
```
DilumAluthge, line 4: Note that you also need to copy these changes and paste them into the "Edit Steps" section of the Buildkite web UI for them to take effect.

```suggestion
  - label: ":buildkite: Pipeline upload"
    plugins:
      - JuliaCI/merge-commit#v1: ~
```

.buildkite/generator.jl
```diff
@@ -0,0 +1,228 @@
...
+
+    # "project source hash" is a combination of meta.json (to absorb
+    # changes from include()'ing a `common.jl`) as well as the entire
+    # tree the project lives in (to absorb changes from patches)
+    # In order to support symlinked directories as dependencies we calculate
+    # the tree hash on a TMP_PROJECT using `cp -RL` to resolve the symlink and
+    # hash the actual content.
+
+    TMP_PROJECT = mktempdir()
+    cp(PROJECT, TMP_PROJECT; follow_symlinks=true)
```
vchuravy, line 170:
```
ERROR: LoadError: ArgumentError: '/tmp/jl_DwSxjs' exists. `force=true` is required to remove '/tmp/jl_DwSxjs' before copying.
```

DilumAluthge:
You should run the generator inside Sandbox.jl.

.buildkite/generator.jl
```diff
@@ -0,0 +1,230 @@
...
+    # "project source hash" is a combination of meta.json (to absorb
+    # changes from include()'ing a `common.jl`) as well as the entire
+    # tree the project lives in (to absorb changes from patches)
+    # In order to support symlinked directories as dependencies we calculate
+    # the tree hash on a TMP_PROJECT using `cp -RL` to resolve the symlink and
+    # hash the actual content.
+
+    TREE_HASH = mktempdir() do TMP_PROJECT
+        cp(PROJECT, TMP_PROJECT; follow_symlinks=true, force=true)
+        return julia(`-e "using Pkg; print(bytes2hex(Pkg.GitTools.tree_hash(\"$(TMP_PROJECT)\")))"`) |> readchomp
```
giordano, line 170:
```suggestion
        bytes2hex(Pkg.GitTools.tree_hash(TMP_PROJECT))
```

.buildkite/generator.jl
```diff
@@ -0,0 +1,230 @@
...
+
+# We're going to snarf out the BB and BBB tree hashes and combine them to be used later in our build cache
+BB_HASH = julia(```
+    -e "using Pkg, SHA; \
+        gethash(uuid) = collect(Pkg.Types.Context().env.manifest[Pkg.Types.UUID(uuid)].tree_hash.bytes); \
+        print(bytes2hex(sha256(vcat( \
+            gethash(\"7f725544-6523-48cd-82d1-3fa08ff4056e\"), \
+            gethash(\"12aac903-9f7c-5d81-afc2-d9565ea332ae\"), \
+        ))));"
+```) |> readchomp
```
giordano, line 136: Also here you can run the command locally without shelling out, but you may need to add `SHA` to the environment.

vchuravy: I think we already have SHA.

giordano: Right! https://github.com/JuliaPackaging/Yggdrasil/blob/9f9b73258dc3b347ee08b8c764b63c01cc462b44/.buildkite/Project.toml#L5

giordano: Ok, reading what we're doing here more carefully, it's probably right we should run it in the `/.ci` environment more specifically, but I wonder if we should eventually merge the `.buildkite` and `.ci` environments.

DilumAluthge:
It looks like we are now checking out the merge commit!

.buildkite/utils.jl
```diff
@@ -0,0 +1,168 @@
...
+    :os => "linux",
+    :sandbox_capable => "true"
+)
+
+plugins() = [
+    "JuliaCI/julia#v1" => Dict(
+        "persist_depot_dirs" => "packages,artifacts,compiled",
+        "version" => "1.7"
+    ),
+    # "JuliaCI/merge-commit" => "~"
```
vchuravy, line 51:
```suggestion
    "JuliaCI/merge-commit" => nothing
```

.buildkite/utils.jl
```diff
@@ -0,0 +1,168 @@
...
+]
+
+wait_step() = Dict(:wait => "~")
```
vchuravy, line 54:
```suggestion
wait_step() = Dict(:wait => nothing)
```

.buildkite/utils.jl
```diff
@@ -0,0 +1,168 @@
...
+    echo "--- Registering ${NAME}..."
+    export BB_HASH PROJ_HASH
+    julia ${BUILDKITE_BUILD_CHECKOUT_PATH}/.ci/register_package.jl "${NAME}.meta.json" --verbose
+    """
+
+    Dict(
+        :label => "register -- $NAME",
+        :agents => agent(),
+        :plugins => plugins(),
+        :timeout_in_minutes => 30,
```
vchuravy, line 156: Look at time out

.buildkite/utils.jl
```diff
@@ -0,0 +1,168 @@
...
+        :priority => -1,
+        :concurrency => 16,
+        :concurrency_group => "yggdrasil/build/$NAME", # Could use ENV["BUILDKITE_JOB_ID"]
+        :commands => [script],
+        :env => Dict(
+            "NAME" => NAME,
+            "PLATFORM" => PLATFORM,
+            "PROJECT" => PROJECT,
+            "BB_HASH" => BB_HASH,
+            "PROJ_HASH" => PROJ_HASH
```
vchuravy, line 129:
```suggestion
            "PROJ_HASH" => PROJ_HASH,
            "BINARYBUILDER_USE_CCACHE" => "true",
            "BINARYBUILDER_STORAGE_DIR" => "/cache/yggdrasil"
```

vchuravy:
@giordano sadly NUMA has no warnings, do you have suggestions for a package that is fast and does emit some warnings?

giordano:
GR has some hundreds of warnings and it shouldn't take too long: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=24439&view=results

vchuravy:
From Elliot:

> 1) Generator job that analyzes the diff, and launches pipelines based on that.
> 2) Builder jobs that do the actual building.  We want these to be skipped if they’re building the same thing as was built previously, so we’ll use [coppermind](https://github.com/staticfloat/coppermind-buildkite-plugin) (with cryptic, to store the S3 access key) to skip the actual build step if a treehash match is found
> 3) Registration job that downloads the binaries from buildkite artifacts again (since coppermind’s caching on S3 is an implementation detail, not the true storage location), then uploads them to GH releases

.buildkite/NOTES.md
```diff
@@ -0,0 +1,10 @@
+# Artifacts
+
+Buildkite has it's own artifact uploader, we can upload to our own S3
```
giordano, line 3:
```suggestion
Buildkite has its own artifact uploader, we can upload to our own S3
```

.buildkite/path_processors/per-project
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# De-duplicate directories
+declare -A DIRS
+for f in "$@"; do
+    DIR="$(dirname ${f})"
+    DIRS["${DIR}"]=1
+done
+echo "${!DIRS[@]}"
```
giordano, line 9:
```suggestion
echo "${!DIRS[@]}"

```

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,23 @@
...
+    agents:
+      queue: "yggdrasil"
+      arch:  "x86_64"
+      os:    "linux"
+      # Only run on "sandbox.jl" machines
+      sandbox_capable: "true"
+    if: build.message !~ /\[skip tests\]/
+    timeout_in_minutes: 10
+    env:
+      JULIA_PKG_SERVER: us-east.pkg.julialang.org
```
giordano, line 23: Also `JULIA_PKG_SERVER_REGISTRY_PREFERENCE` (#5078): https://github.com/JuliaPackaging/Yggdrasil/blob/8e7a9c4f1761fcb71c8d2e33688d36ff6fb9f9d2/azure-pipelines.yml#L17-L18

.buildkite/generator.jl
```diff
@@ -0,0 +1,114 @@
...
+using Downloads, Dates, SHA
+
+const PROJECT = only(ARGS)
+const DEBUG = !haskey(ENV, "BUILDKITE")
+
+include("utils.jl")
+
+if !isnothing(Pkg.pkg_server())
+    resp = try
+        Downloads.request("$(Pkg.pkg_server())/registries")
```
giordano, line 19: #5078
```suggestion
        headers = Pkg.PlatformEngines.get_metadata_headers(Pkg.pkg_server())
        Downloads.request("$(Pkg.pkg_server())/registries"; headers)
```

.buildkite/utils.jl
```diff
@@ -0,0 +1,161 @@
...
+        "staticfloat/cryptic#v2" => Dict(
+            "variables" => [
+                "AWS_SECRET_ACCESS_KEY=\"U2FsdGVkX1846b0BRbZjwIWSFV+Fiv1C/Hds/vB3aTkxubHPnRP6lVxGkAkOcFuvAntkoLF6J64QrOHWvjz8xg==\"",
+            ]
+        ),
+        "staticfloat/coppermind#v2" => Dict(
+            "inputs" => [
+                PROJECT,
+                ".ci/",
+                # ?meta.json
```
vchuravy, line 96: @staticfloat in AZP we hash in the meta.json Not sure how to do that here

vchuravy:
So the last steps are:

- Use a token for `jlbuild` instead one of mine
- Improve https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1255 (but we can do that separatly and I can revert the BB bump here)

giordano (commented):
Mentioned in the CI call, but repeating it also here: we should implement also here the `[skip build]` keyword, which doesn't skip the CI job completely, only the build steps:
https://github.com/JuliaPackaging/Yggdrasil/blob/8e7a9c4f1761fcb71c8d2e33688d36ff6fb9f9d2/azure-pipelines.yml#L32-L49 And this will be used during registration to re-use the artifacts from the previous build: https://github.com/JuliaPackaging/Yggdrasil/blob/8e7a9c4f1761fcb71c8d2e33688d36ff6fb9f9d2/.ci/register_package.jl#L96

giordano:
Looking at https://github.com/JuliaBinaryWrappers/HelloWorldC_jll.jl/commit/bd36c4c02a9eb2fb017c95abe647d2c36feee5d9 I believe you aren't exporting the environment variable
```
YGGDRASIL=true
```
are you?  That's currently set in https://github.com/JuliaPackaging/Yggdrasil/blob/8e7a9c4f1761fcb71c8d2e33688d36ff6fb9f9d2/.ci/azp_agent/.env, so you may have missed it.

PR #97, opened 2019-09-26
Add builder for lz4

giordano:
Indirect dependency of GTK, see #62

PR #4773, opened 2022-04-17
add SCS build using Pardiso solver through MKL

kalmarek:


giordano:
> so the question to @giordano (and maybe @bodono) is: can we avoid linking against `libmkl_gnu_thread`?

If that was possible, maybe you could just link to libblastrampoline and forget about MKL?

S/SCS_MKL/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+# Collection of sources required to build SCSBuilder
+sources = [
+    GitSource("https://github.com/cvxgrp/scs.git", "c785d2fad46a30f1d43764d682509d0b56e5c64f")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/scs*
+scsflags="DLONG=1 USE_OPENMP=0 BLAS32=1 NOBLASSUFFIX=1"
+mklflags="-L${prefix}/lib -Wl,--no-as-needed -lmkl_rt -lmkl_gnu_thread -lmkl_core -lpthread -lm -ldl"
```
giordano, line 15: Not sure `-Wl,--no-as-needed` is needed (see what I did there)

S/SCS_MKL/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+]
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libscsmkl", :libscsmkl, dont_dlopen=true),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("MKL_jll", v"2022.0.0"),
```
giordano, line 40: If you want to build against a specific version of MKL, you likely want to specify also the compat, with
```suggestion
    Dependency("MKL_jll"; compat="2022.0.0"),
```

S/SCS_MKL/build_tarballs.jl
kalmarek: `-Wl, --no-as-needed` is suggested by intel: https://www.intel.com/content/www/us/en/developer/tools/oneapi/onemkl-link-line-advisor.html and therefore is also here:
https://github.com/cvxgrp/scs/blob/c785d2fad46a30f1d43764d682509d0b56e5c64f/scs.mk#L136
I don't feel competent to change any of those ;)

ViralBShah:
Can't we just use `libmkl_rt`?

kalmarek:
> Can't we just use `libmkl_rt`? IIRC, the `gnu` stuff is not available on mac and windows.
If I drop `-Wl,--no-as-needed` I get plenty of undefined symbols, however dropping `-lmkl_gnu_thread`... works just fine ;D

so maybe it's not really needed? @bodono ?

giordano:
I believe dropping `-lmkl_gnu_thread` would be the best, if possible, at which point you could also just use libblastrampoline if you want.  My understanding was that MKL internally opens the threading library it needs at runtime, but I never spent much time actually digging into what it does, so I may well be very much wrong.

ViralBShah:
We really just want to go through LBT eventually.

kalmarek:
@ViralBShah but isn't linking against LBT missing julia-lts?

giordano:
Yes, there are a couple of issues with using libblastrampoline in Julia v1.6.

kalmarek:
ok; I'd be happiest to keep it available to julia-lts as well.

bodono:
> > Can't we just use `libmkl_rt`? IIRC, the `gnu` stuff is not available on mac and windows.
> 
> If I drop `-Wl,--no-as-needed` I get plenty of undefined symbols, however dropping `-lmkl_gnu_thread`... works just fine ;D
> 
> so maybe it's not really needed? @bodono ?

I don't really understand MKL linking flags, I put those there because the online tool told me to use them and it made the [tests pass](https://github.com/cvxgrp/scs/actions/runs/1976389540). I would love to get rid of them if possible and make the install easier to maintain over multiple platforms. One thing to check though is whether or not MKL is able to be multi-threaded if it is not linked to a threading library, since that's where a lot of the speedup comes from.

ViralBShah:
How will the user of SCS.jl pick between the openblas and the MKL linked JLLs? 

I do think the right way is to use LBT across the JuMP ecosystem of solvers.

kalmarek:
I can enable other platforms and let's see how it fares. Though I believe scs Makefile supports only Linux with mkl?

@bodono I can confirm that even without `-lmkl_gnu_thread` I can see multiple cores loaded when using `scs`
The number of loaded cores (for all solvers) corresponds to `OMP_NUM_THREADS` env variable. This is something that I don't really understand since we compile both libs with `OPENMP=0`. what does this flag control?

On my toy problem `MKLDirectSolver` is ~1.5× faster taking 17s vs 24s by the vanilla `DirectSolver`. Also note the displayed warning; Is there anything we should do about it?
```
------------------------------------------------------------------
               SCS v3.2.1 - Splitting Conic Solver
        (c) Brendan O'Donoghue, Stanford University, 2012
------------------------------------------------------------------
problem:  variables n: 7382, constraints m: 12836
cones:    z: primal zero / dual free vars: 5455
          s: psd vars: 7381, ssize: 1
settings: eps_abs: 1.0e-08, eps_rel: 1.0e-08, eps_infeas: 1.0e-07
          alpha: 1.50, scale: 1.00e-01, adaptive_scale: 1
          max_iters: 20000, normalize: 1, rho_x: 1.00e-06
          acceleration_lookback: -10, acceleration_interval: 10
lin-sys:  sparse-direct-mkl-pardiso
          nnz(A): 57566, nnz(P): 0
OMP: Warning #97: Requested number of active parallel levels "-2147483648" is negative; ignored. 
------------------------------------------------------------------
 iter | pri res | dua res |   gap   |   obj   |  scale  | time (s)
------------------------------------------------------------------
     0| 1.57e+02  9.96e-01  2.61e+02 -1.69e+02  1.00e-01  6.23e-02 
   250| 1.02e-02  4.52e-04  1.19e-02 -2.04e+00  1.00e-01  7.23e-01 
   500| 8.76e-03  4.95e-04  2.23e-02 -1.86e+00  1.00e-01  1.45e+00 
   750| 1.12e-02  3.98e-04  1.30e-01 -1.27e+00  1.00e-01  2.20e+00 
    [...]
  5250| 5.51e-07  3.30e-08  2.94e-06 -2.80e-01  1.00e-01  1.59e+01 
  5500| 5.50e-08  5.82e-09  8.11e-08 -2.80e-01  1.00e-01  1.66e+01 
  5700| 5.27e-08  1.45e-09  4.33e-09 -2.80e-01  1.00e-01  1.72e+01 
------------------------------------------------------------------
status:  solved
timings: total: 1.72e+01s = setup: 5.79e-02s + solve: 1.72e+01s
         lin-sys: 2.80e+00s, cones: 1.34e+01s, accel: 4.16e-01s
------------------------------------------------------------------
objective = -0.280407
------------------------------------------------------------------
```
vs.
```
------------------------------------------------------------------
               SCS v3.2.0 - Splitting Conic Solver
        (c) Brendan O'Donoghue, Stanford University, 2012
------------------------------------------------------------------
problem:  variables n: 7382, constraints m: 12836
cones:    z: primal zero / dual free vars: 5455
          s: psd vars: 7381, ssize: 1
settings: eps_abs: 1.0e-08, eps_rel: 1.0e-08, eps_infeas: 1.0e-07
          alpha: 1.50, scale: 1.00e-01, adaptive_scale: 1
          max_iters: 20000, normalize: 1, rho_x: 1.00e-06
          acceleration_lookback: -10, acceleration_interval: 10
lin-sys:  sparse-direct
          nnz(A): 57566, nnz(P): 0
------------------------------------------------------------------
 iter | pri res | dua res |   gap   |   obj   |  scale  | time (s)
------------------------------------------------------------------
     0| 1.57e+02  9.96e-01  2.61e+02 -1.69e+02  1.00e-01  6.45e-03 
   250| 1.02e-02  4.52e-04  1.19e-02 -2.04e+00  1.00e-01  9.64e-01 
   500| 8.74e-03  4.96e-04  2.33e-02 -1.86e+00  1.00e-01  2.03e+00 
   750| 1.13e-02  4.07e-04  5.91e-02 -1.58e+00  1.00e-01  3.09e+00 
    [...]
  5250| 1.05e-07  1.31e-08  2.84e-07 -2.80e-01  1.00e-01  2.27e+01 
  5500| 1.12e-07  1.48e-09  1.65e-07 -2.80e-01  1.00e-01  2.38e+01 
  5550| 1.14e-07  1.33e-09  3.66e-10 -2.80e-01  1.00e-01  2.40e+01 
------------------------------------------------------------------
status:  solved
timings: total: 2.41e+01s = setup: 3.84e-02s + solve: 2.40e+01s
         lin-sys: 2.81e+00s, cones: 1.94e+01s, accel: 6.05e-01s
------------------------------------------------------------------
objective = -0.280408
------------------------------------------------------------------
```

kalmarek:
@ViralBShah `MKLDirectSolver` is added to `SCS.available_solvers` if `MKL_jll` is loaded _before_ `SCS`. Then user is free to choose the linear solver defining `JuMP.optimizer_with_attributes(SCS.Optimizer, "linear_solver"=>SCS.MKLDirectSolver)`.

We would rather not link against LBT at this moment, at least until its issues on julia-lts (i.e. 1.6) are resolved.

ViralBShah:
On windows, the error is because it can't find MKL, and on mac it is an unknown linker flag. I think these both ought to be fixable. Ok to merge this now and do a new PR to work on the other platforms eventually - but I feel it is possible to somehow make them work.

kalmarek:
Those guys somehow build on linking to `mkl_*` on windows:
https://github.com/JuliaPackaging/Yggdrasil/blob/8be19820afa24ff817185f3c53c4d20786a8add3/M/MUMPS/MUMPS_seq_MKL/build_tarballs.jl#L21
but they're compiling with `gfortran` As for now I'll comment those platforms as suggested.

kalmarek:
ready to be merged!

kalmarek:
Thanks guys for all the comments!

bodono:
Nice work getting this checked in!

> @bodono I can confirm that even without `-lmkl_gnu_thread` I can see multiple cores loaded when using `scs` The 
number of loaded cores (for all solvers) corresponds to `OMP_NUM_THREADS` env variable. This is something that I don't really understand since we compile both libs with `OPENMP=0`. what does this flag control?

Maybe it should be one of [`USE_OPENMP`](https://www.cvxgrp.org/scs/api/compile_flags.html) or [`_OPENMP`](https://www.ibm.com/docs/en/xffbg/121.141?topic=code-openmp-preprocessor-macro-conditional-compilation) (note underscore)? In any case, all that flag is doing is allowing some of the for-loops within SCS to be parallelized using openMP. It has no control over linked packages like MKL.

> OMP: Warning #97: Requested number of active parallel levels "-2147483648" is negative; ignored. 

This warning is very strange, any idea what's causing it?

PR #4777, opened 2022-04-18
[Pfapack] Convert to 64-bit API

xrq-phys:
Sorry I was using default 32-bit API and sending a bunch of 32-bit calls to `libblastrampoline` which is by default undefined in Julia. Currently


giordano:
> and sending a bunch of 32-bit calls to libblastrampoline which is by default undefined in Julia

I guess you just need a 32-bit BLAS to forward the calls to: https://github.com/giordano/OpenBLAS32.jl

xrq-phys:
> > and sending a bunch of 32-bit calls to libblastrampoline which is by default undefined in Julia
> 
> I guess you just need a 32-bit BLAS to forward the calls to: https://github.com/giordano/OpenBLAS32.jl

I see. Thanks!

PR #4778, opened 2022-04-19
New recipe: Pathfinder-v0.5.0

dylanxyz:
Build recipe for [Pathfinder](https://github.com/servo/pathfinder), a Rust drawing library for text and vector graphics using the GPU.

### DIFF
--- a/P/Pathfinder/build_tarballs.jl
+++ b/P/Pathfinder/build_tarballs.jl
@@ -1,0 +1,71 @@
+using BinaryBuilder
+
+name = "Pathfinder"
+version = v"0.5.0"
+sources = [
+    GitSource("https://github.com/servo/pathfinder.git", "7281a607a80dacc5470d7d3ec208341b6f0b0727"),
+    DirectorySource("./bundled")
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir
+cd ${WORKSPACE}/srcdir/pathfinder*
+
+# by default, pathfinder will compile to a static library,
+# so we change "staticlib" to "cdylib" to compile to a dynamic library
+atomic_patch -p1 ../patches/Compile-to-dynamic-lib.patch
+
+# we need cbindgen to generate the header file
+cargo install cbindgen --force --target x86_64-unknown-linux-musl
+
+cd c
+
+# generate the header file
+cbindgen \
+    --lang c \
+    --crate pathfinder_c \
+    --config cbindgen.toml \
+    --output "${includedir}/pathfinder.h"
+
+# build pathfinder
+cargo build --release --lib --target=${rust_target}
+
+# On windows the generated .dll doesn't start with lib
+if [[ "${target}" == *-mingw* ]]; then
+    cp \
+        ../target/${rust_target}/release/pathfinder.dll \
+        ../target/${rust_target}/release/libpathfinder.dll
+fi
+
+# install the library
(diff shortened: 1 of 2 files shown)


P/Pathfinder/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+
+name = "Pathfinder"
+version = v"0.5.0"
+sources = [
+    GitSource("https://github.com/servo/pathfinder.git", "7281a607a80dacc5470d7d3ec208341b6f0b0727"),
+    DirectorySource("./bundled")
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir
```
giordano, line 11:
```suggestion
```

P/Pathfinder/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+cd ${WORKSPACE}/srcdir/pathfinder*
+
+# by default, pathfinder will compile to a static library,
+# so we change "staticlib" to "cdylib" to compile to a dynamic library
+atomic_patch -p1 ../patches/Compile-to-dynamic-lib.patch
+
+# we need cbindgen to generate the header file
+cargo install cbindgen --force --target x86_64-unknown-linux-musl
```
giordano, line 19: We should have the environment variable `rust_host` to indicate the host rust target
```suggestion
cargo install cbindgen --force --target "${rust_host}"
```

P/Pathfinder/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+# install the library
+install -D -m 755 "../target/${rust_target}/release/libpathfinder.${dlext}" "${libdir}/libpathfinder.${dlext}"
+
+# install the licenses
+install_license ../LICENSE-APACHE ../LICENSE-MIT
+"""
+
+platforms = supported_platforms(; experimental=true)
```
giordano, line 47: No need to use `experimental=true` anymore :wink: 
```suggestion
platforms = supported_platforms()
```

giordano (commented):
Thank you so much for your contribution, very well done recipe!  Some minor comments

P/Pathfinder/build_tarballs.jl
dylanxyz: Hi. Thanks for the comments.

Pathfinder uses another rust crate called `cbindgen` to generate the `C` header files. cbindgen is a executable that runs during the build process, so installing for the rust target and running cbindgen generates the error: "Cannot execute binary file: Exec format error". I think the problem is that we have to install cbindgen for the host instead for the rust_target. Using `x86_64-unknown-linux-musl` solved this issue for me.

giordano: I understand, and that's why I suggested to use `rust_host`, not `rust_target` :slightly_smiling_face: 
```console
% julia --compile=min -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("aarch64", "macos"); compilers=[:rust])'
sandbox:${WORKSPACE} # echo "${rust_host}"
x86_64-unknown-linux-musl
sandbox:${WORKSPACE} # echo "${rust_target}"
aarch64-apple-darwin
```

dylanxyz: O i see, i misread. Sorry. I was looking for a variable like `rust_target` in the BinaryBuilder docs but i could not finded. Thanks, this solves the issue.

dylanxyz:
You're welcome!

PR #4780, opened 2022-04-19
[Pathfinder] Depend on `cbindgen_jll` instead of using cargo

giordano:


giordano:
@dylanxyz I slightly simplified the recipe, and used the new `cbindgen_jll` as dependency, instead of rebuilding it locally every time.  Note also that I made all other dependencies runtime:
```console
% libtree libpathfinder.so 
libpathfinder.so 
├── libharfbuzz.so.0 [default path]
│   ├── libglib-2.0.so.0 [default path]
│   │   └── libpcre.so.1 [default path]
│   │       └── libpthread.so.0 [default path]
│   ├── libgraphite2.so.3 [default path]
│   └── libfreetype.so.6 [default path]
│       ├── libz.so.1 [default path]
│       ├── libbrotlidec.so.1 [default path]
│       │   └── libbrotlicommon.so.1 [default path]
│       ├── libharfbuzz.so.0 [default path]
│       ├── libpng16.so.16 [default path]
│       │   └── libz.so.1 [default path]
│       └── libbz2.so.1.0 [default path]
├── librt.so.1 [default path]
├── libpthread.so.0 [default path]
├── libfreetype.so.6 [default path]
└── libfontconfig.so.1 [default path]
    ├── libfreetype.so.6 [default path]
    ├── libpthread.so.0 [default path]
    └── libexpat.so.1 [default path]
```
I missed it before that libpathfinder actually dynamically links to those libraries.

dylanxyz:
I was wondering if we could use the cbindgen binary directly instead of installing with cargo. Thanks @giordano.

PR #4784, opened 2022-04-20
[Pfapack] Fix .so Contents

xrq-phys:
- `-force-load` only work for LLVM-based linkers. Fallback to `--whole-archive`
- Update upstream to use a CMakeLists.txt with all [sdcz] routines

### DIFF
--- a/P/Pfapack/build_tarballs.jl
+++ b/P/Pfapack/build_tarballs.jl
@@ -26,7 +26,7 @@ make install
 
 # Manual conversion from static to dynamic lib.
 # Avoid linking libgfortran (these .f files make no reference to Fortran libs.)
-cc -shared cc -shared $(flagon -Wl,--whole-archive) ${prefix}/lib/libpfapack.a $(flagon -Wl,--no-whole-archive) \
+cc -shared $(flagon -Wl,--whole-archive) ${prefix}/lib/libpfapack.a $(flagon -Wl,--no-whole-archive) \
     -o ${libdir}/libpfapack.${dlext} \
     -L${libdir} -lblastrampoline -lm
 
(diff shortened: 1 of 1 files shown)


P/Pfapack/build_tarballs.jl
```diff
@@ -25,7 +26,7 @@ make install
 
 # Manual conversion from static to dynamic lib.
 # Avoid linking libgfortran (these .f files make no reference to Fortran libs.)
-cc -shared -Wl,-force_load,${prefix}/lib/libpfapack.a \
+cc -shared -Wl,--whole-archive ${prefix}/lib/libpfapack.a -Wl,--no-whole-archive \
```
giordano, line 29:
```suggestion
cc -shared $(flagon -Wl,--whole-archive) ${prefix}/lib/libpfapack.a $(flagon -Wl,--no-whole-archive) \
```

P/Pfapack/build_tarballs.jl
```diff
@@ -17,15 +17,27 @@ cd $WORKSPACE/srcdir
...
+case ${target} in
+
+    *"apple"*)
+        # Apple implements a different linker option.
+        export A2SO="-Wl,-force_load,${prefix}/lib/libpfapack.a"
+        ;;
+    *)
+        export A2SO="-Wl,--whole-archive ${prefix}/lib/libpfapack.a -Wl,--no-whole-archive"
+        ;; 
+esac
```
giordano, line 36: `flagon` does this for you: https://github.com/JuliaPackaging/Yggdrasil/pull/4784#discussion_r853849184

xrq-phys: I see. Thanks!

giordano: If you're curious, this is the script: https://github.com/JuliaPackaging/Yggdrasil/blob/acb37f70728e86b5f747f752dbab71a1ce198fb3/0_RootFS/Rootfs/bundled/utils/flagon

xrq-phys: Impressive. Thanks.

PR #4790, opened 2022-04-20
New recipe: FastME-v2.1.6

BenjaminDoran:
I am having an error compiling when glibc is not available on linux-musl and apple platforms

```
[13:38:50] Undefined symbols for architecture arm64:
[13:38:50]   "_rpl_realloc", referenced from:
[13:38:50]       _Read_Branch_Label in p_bootstrap.o
[13:38:50]       _Make_New_Edge_Label in p_bootstrap.o
[13:38:50] ld: symbol(s) not found for architecture arm64
```

Not sure if there is anything I can do to easily fix it, so I limited the available platforms.

### DIFF
--- a/F/FastME/build_tarballs.jl
+++ b/F/FastME/build_tarballs.jl
@@ -36,7 +36,7 @@ platforms = supported_platforms()
 # [13:38:50]       _Make_New_Edge_Label in p_bootstrap.o
 # [13:38:50] ld: symbol(s) not found for architecture arm64
 filter!(p-> libc(p) != "musl", platforms)
-filter!(p-> !(Sys.isapple(p)), platforms)
+filter!(!Sys.isapple, platforms)
 
 products = [
     ExecutableProduct("fastme", :fastme),
--- a/F/FastME/build_tarballs.jl
+++ b/F/FastME/build_tarballs.jl
@@ -14,6 +14,7 @@ cd ${WORKSPACE}/srcdir/FastME
 wget https://www.gnu.org/licenses/gpl-3.0.txt > LICENSE
 
 update_configure_scripts --reconf
+autoupdate
 
 # optimized multi-threading
 ./configure --prefix=$prefix --build=${MACHTYPE} --host=${target}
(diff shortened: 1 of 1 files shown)


giordano:
> when glibc is not available on linux-musl

I mean, yes, glibc isn't available on musl systems: musl is the name of the C standard library, which is _not_ glibc. I have no idea what's the problem you're encouraging though

F/FastME/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+    GitSource("https://gite.lirmm.fr/atgc/FastME.git", "4d0ecf8e1f06bd22e25e57dc362594548c3e10d4"),
+]
+
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/FastME
+
+# using license as specified here: http://www.atgc-montpellier.fr/fastme/binaries.php
+wget https://www.gnu.org/licenses/gpl-3.0.txt > LICENSE
+
```
giordano, line 15:
```suggestion
install_license /usr/share/licenses/GPL-3.0+
```

F/FastME/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+platforms = supported_platforms()
+# errors when glibc is not available on linux-musl and apple platforms
+# [13:38:50] Undefined symbols for architecture arm64:
+# [13:38:50]   "_rpl_realloc", referenced from:
+# [13:38:50]       _Read_Branch_Label in p_bootstrap.o
+# [13:38:50]       _Make_New_Edge_Label in p_bootstrap.o
+# [13:38:50] ld: symbol(s) not found for architecture arm64
+filter!(p-> libc(p) != "musl", platforms)
+filter!(p-> !(Sys.isapple(p)), platforms)
```
giordano, line 41:
```suggestion
filter!(!Sys.isapple, platforms)
```

giordano:
Also, I didn't have any problems compiling for `x86_64-linux-musl`.  What problems did you have?

PR #4795, opened 2022-04-21
[Qt5Declarative] Update to v5.15.3 and build for aarch64-apple-darwin

giordano:


giordano:
@barche building for FreeBSD fails with
```
[16:21:51]  ---> qmake $qtsrcdir
[16:21:51] Project ERROR: Cannot run target compiler 'x86_64-unknown-freebsd11.1-g++'. Output:
```
Any idea where the (wrong) path of the compiler comes from?  The target is now called `x86_64-unknown-freebsd12.2`, not `x86_64-unknown-freebsd11.1`, it'd be nice not too hardcode this stuff in this build process and refer the compilers just by `g++` (or ideally even just `c++`)

giordano:
And aarch64-apple-darwin fails with some hundreds of errors.  Sigh.

barche:
Well, even with negative time you beat me to it :) This looks good to me!

giordano:
I was pretty scared by the hundreds of errors in the aarch64-apple-darwin build, but they were all related to `‘numeric_limits’ is not a member of ‘std’`, which was caused by missing `<limits>` header file in `Qt5Base_jll`, so fixing that fixed all the issues here :slightly_smiling_face:

PR #1939, opened 2020-10-19
JLL bump: boost_jll

fingolfin:
This pull request bumps the JLL version of boost_jll.
It was generated via the `recursively_regenerate_jlls.jl` script.

PR #4797, opened 2022-04-21
New recipe: AWSCRT v0.1.0

Octogonapus:
This PR creates a new recipe, AWSCRT v0.1.0, containing all of the C libraries which make up the [AWS Common Runtime](https://docs.aws.amazon.com/sdkref/latest/guide/common-runtime.html).

I'm closing https://github.com/JuliaPackaging/Yggdrasil/pull/4728 and https://github.com/JuliaPackaging/Yggdrasil/pull/4727 in favor of this PR.
This PR addresses the issues of the prior PRs by building all AWS CRT libraries as static objects, which are built into one shared object by a project defined in `bundled/awscrt`.
This method of building the CRT fixes two problems.
First, we avoid conflicts with OpenSSL by not exporting shared libraries called `libssl` and `libcrypto`.
Second, we reduce the burden of maintaining this JLL by avoiding the need for a patch to the build script of every CRT library.
If we wanted to build a shared object for every library (and split this JLL up into multiple, except aws-lc and s2n-tls to avoid the first problem), then every libraries build script would need a patch to allow it to consume static library dependencies while also emitting a shared library target (the AWS build scripts just aren't configurable enough to do this without changes).

I have tested this JLL by creating a wrapper and reimplementing one of aws-c-mqtt's end-to-end tests in Julia using that wrapper.

TODO on my end before I feel this is RTM:
- [x] Debug aarch64-linux-gnu
- [x] Try to debug windows and see if it will cooperate


Octogonapus:
Looks like aarch64 is failing because BinaryBuilder doesn't like this arch flag `-march=armv8-a+crc` (I'm guessing based on https://github.com/JuliaPackaging/Yggdrasil/pull/1939#issuecomment-713200769). If this is the case, I'm not sure I should just patch it out like in that PR, because that flag is used to compile arch-specific code. Wouldn't removing that flag break it, or at least make it inefficient?

giordano:
Yeah, in general we don't like build systems forcing `-march=...`, most of the time it's because they use `-march=native` which is absolutely wrong when cross-compiling.  However here they're setting `-march=armv8-a+crc` which isn't _too_ bad (I suspect most aarch64 where julia runs will have crc too).  You can skip the check with `build_tarballs(...; lock_microarchitectures=false)`

Octogonapus:
FreeBSD is currently broken because of https://github.com/aws/s2n-tls/issues/3281

Octogonapus:
Windows is broken because of `fatal error: Windows.h: No such file or directory`. Not sure why that header isn't found. The right compiler is being invoked.

IanButterworth:
I seem to recall that's a capitalization issue

giordano:
Yes, MinGW consistently uses lowercase for all header files, and we build on a case _sensitive_ filesystem.  Windows users generally don't care because most common filesystems used with that operating system are case _insensitive_, so they could even use `WiNdOwS.h` and it'd work.

A/AWSCRT/bundled/patches/aws-c-common.patch
```diff
@@ -0,0 +1,151 @@
...
++++ b/include/aws/testing/aws_test_harness.h
+@@ -459,7 +459,11 @@ static inline int s_aws_run_test_case(struct aws_test_harness *harness) {
+ /* https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences */
+ #ifdef _WIN32
+ 
++#if defined(__MINGW32__) || defined(__MINGW64__)
++#    include <windows.h>
++#else
+ #    include <Windows.h>
++#endif
```
giordano, line 13: This should really just be
```c
include <windows.h>
```

Octogonapus: That will also work on `msbuild.exe`?

giordano: See for example https://en.wikipedia.org/wiki/Windows.h, and also https://docs.microsoft.com/en-us/windows/win32/learnwin32/your-first-windows-program (although Microsoft docs isn't always consistent about casing, they just don't care at all about casing)

Octogonapus:
I can't replicate those patch errors locally :thinking:

giordano:
Let me guess.  The files have CRLF line ending?

giordano:
Looking at https://github.com/awslabs/aws-c-io/blob/cfe553a770e9c2d1c93b8cdfb870b9f2a46b436e/source/windows/secure_channel_tls_handler.c (wel, downloading it locally, can't really tell the line ending from GitHub) the files have CRLF line ending.  Ok, this is frustratingly annoying to deal with.

You have to do the following:

* `git rm A/AWSCRT/bundled/patches/aws-c-common.patch`.  Make sure the file is removed from disk (but that's the default behaviour of `git rm`)
* create the file `A/AWSCRT/bundled/patches/.gitattributes` with content
  ```
  *.patch -text diff
  ```
* `git add A/AWSCRT/bundled/patches/.gitattributes`
* recreate the patch and re-add it to git
* `git commit`
* `git push`
* profit

It works for you locally because you have the original files, but git converts the line ending automatically.

giordano:
Of course life would be easier if they were consistent about line ending.  It appears only the file [`secure_channel_tls_handler.c`](https://github.com/awslabs/aws-c-io/blob/cfe553a770e9c2d1c93b8cdfb870b9f2a46b436e/source/windows/secure_channel_tls_handler.c) in that directory uses CRLF, the others use LF (and they're also for Windows) :facepalm:

Octogonapus:
Apple builds no longer silently failing.
Not sure I can get Windows working right now. I'm including `winnt.h` which provides those missing definitions, but there must be some guards preventing them from being included.

This is working on Linux so at this point I'd like to get this merged.

giordano:
>  I'm including winnt.h which provides those missing definitions, but there must be some guards preventing them from being included.

https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-processor_number#requirements

> Minimum supported client | Windows 7 [desktop apps only]

Windows 7 is `_WIN32_WINNT = 0x0601`: https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170#remarks and in fact in `winnt.h` you can find:

```c
#if (_WIN32_WINNT >= 0x0601)

#ifndef ___PROCESSOR_NUMBER_DEFINED
#define ___PROCESSOR_NUMBER_DEFINED
typedef struct _PROCESSOR_NUMBER {
  WORD Group;
  BYTE Number;
  BYTE Reserved;
} PROCESSOR_NUMBER, *PPROCESSOR_NUMBER;

#define ALL_PROCESSOR_GROUPS 0xffff
#endif /* !___PROCESSOR_NUMBER_DEFINED */

#endif /*(_WIN32_WINNT >= 0x0601)*/
```

So you need to use the compiler flag `-D_WIN32_WINNT=0x0601`

Octogonapus:
Indeed they set that flag when building with MSVC. That doesn't fix the issue, though. Not sure why

giordano:
Anyway, if you want this to be merge with only Linux and macOS you need to exclude FreeBSD and Windows.  We can only merge PRs with successful CI jobs (there is no merge button otherwise) :slightly_smiling_face:

PR #4800, opened 2022-04-21
[unixODBC] Update to v2.3.9 and build for new platforms

giordano:
Should fix https://github.com/JuliaDatabases/ODBC.jl/issues/335.  CC: @quinnj

### DIFF
--- a/U/unixODBC/build_tarballs.jl
+++ b/U/unixODBC/build_tarballs.jl
@@ -3,17 +3,18 @@
 using BinaryBuilder, Pkg
 
 name = "unixODBC"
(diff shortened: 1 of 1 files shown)


quinnj:
Thanks @giordano !!

bobcassels:
Yes, thanks @giordano !

giordano:
Nothing special apart from updating the registry to get the latest versions of `unixODBC_jll` and `iODBC_jll`.  But I'm getting a segmentation fault when starting the package, which is likely a bug in `ODBC_jll`, I'm trying to have a quick look.

PR #4805, opened 2022-04-21
[jlqml] Add Apple M1 support

barche:
This should make QML.jl work on M1 and with the new libcxxwrap also on Julia nightly.

### DIFF
--- a/J/jlqml/build_tarballs.jl
+++ b/J/jlqml/build_tarballs.jl
@@ -7,7 +7,7 @@ using BinaryBuilder, Pkg
 uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
-julia_versions = [v"1.6.3", v"1.7.0", v"1.8.0"]
+julia_versions = [v"1.6.3", v"1.7.0", v"1.8.0", v"1.9.0"]
 
 name = "jlqml"
 version = v"0.3"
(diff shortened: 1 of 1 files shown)


barche:
Seems this launched too soon, it's still using `libcxxwrap_julia_jll v0.9.0+1` so we should rebuild once `libcxxwrap_julia_jll v0.9.1` is available completely.

giordano:
You have to actually edit the file to invalidate the cache and trigger a full rebuild, otherwise this will find the previous build in the cache and not do anything new

J/jlqml/build_tarballs.jl
```diff
@@ -48,14 +48,19 @@ install_license $WORKSPACE/srcdir/jlqml*/LICENSE.md
...
+]
+platforms_linux = expand_cxxstring_abis(platforms_linux)
+platforms_win = expand_cxxstring_abis([Platform("x86_64", "windows"), Platform("i686", "windows")])
+platforms_macos = [ Platform("x86_64", "macos"), Platform("aarch64", "macos") ]
 
-filter!(p -> libc(p) != "musl", platforms) # Qt_jll is currently not available for muslc
-# Qt5Declarative_jll is not available for these architectures:
-filter!(p -> arch(p) != "armv6l", platforms)
-filter!(p -> !(arch(p) == "aarch64" && Sys.isapple(p)), platforms)
+platforms = vcat(platforms_linux, platforms_win, platforms_macos)
```
fingolfin, line 63: Why are you changing the platform computation like this, though, instead of just dropping the `filter!(p -> !(arch(p) == "aarch64" && Sys.isapple(p)), platforms)` line?

barche:
Seems to be OK!

J/jlqml/build_tarballs.jl
barche: I see now that this of course discards the julia versions from  the platform, I'll  revert it.

barche:
Now it should be OK, sorry for the noise.

jheinen:
Great! Pls merge ...

PR #4809, opened 2022-04-22
[libpolymake_julia] rebuild for new libjulia 1.9

benlorenz:


L/libpolymake_julia/build_tarballs.jl
```diff
@@ -14,7 +14,6 @@ version = v"0.8.0"
 
 julia_versions = [v"1.6.3", v"1.7.0", v"1.8.0", v"1.9.0"]
```
giordano, line 15: I've been thinking that for all these packages we should propagate this set of versions into the julia compat, which should be something like
```julia
julia> join("~" .* string.(julia_versions), ", ")
"~1.6.3, ~1.7.0, ~1.8.0, ~1.9.0"
```
or
```julia
julia> join("~" .* string.(getfield.(julia_versions, :major)) .* "." .* string.(getfield.(julia_versions, :minor)), ", ")
"~1.6, ~1.7, ~1.8, ~1.9"
```
if we want to allow the entire minor series (for example important for v1.6).  This would allow us to stop building for 1.7, 1.8 in the future when those series will not be maintained anymore.

benlorenz: done, including a reminder to update the version when restricting the versions.

giordano: Good point, thanks! :+1:

benlorenz:
@giordano the new compat broke the registration (https://github.com/JuliaRegistries/General/pull/58947) due to overlapping version ranges.

I guess the 
```
["0.5-0"]
julia = "1.6.0-1"
```
should be changed to this:
```
["0.5-0.7"]
julia = "1.6.0-1"
```
But I would have expected this to be done automatically?

giordano:
No, because of https://github.com/JuliaLang/Pkg.jl/issues/1568

benlorenz:
I see, thanks for the quick fix!

PR #4824, opened 2022-04-24
[Qt6Base] Update to 6.3.0

barche:
First step in updating Qt6Base, building for the host only (see comment at the top of the build_tarballs.jl).

### DIFF
--- a/Q/Qt6Base/build_tarballs.jl
+++ b/Q/Qt6Base/build_tarballs.jl
@@ -3,12 +3,17 @@
 using BinaryBuilder, Pkg
 
 name = "Qt6Base"
-version = v"6.0.3"
+version = v"6.3.0"
+
+# Set this to true first when updating the version. It will build only for the host (linux musl).
+# After that JLL is in the registyry, set this to false to build for the other platforms, using
+# this same package as host build dependency.
+const host_build = true
 
 # Collection of sources required to build qt6
 sources = [
     ArchiveSource("https://download.qt.io/official_releases/qt/$(version.major).$(version.minor)/$version/submodules/qtbase-everywhere-src-$version.tar.xz",
-                  "1a45b61c2a349964625c50e3ea40cbb309e269762dd0786397e0e18e7e10d394"),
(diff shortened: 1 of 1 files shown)


Q/Qt6Base/build_tarballs.jl
```diff
@@ -66,19 +66,26 @@ case "$target" in
...
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(filter(!Sys.isapple, supported_platforms()))
-platforms_macos = [ Platform("x86_64", "macos") ]
+if host_build
+    platforms = [Platform("x86_64", "linux",cxxstring_abi=:cxx11,libc="musl")]
+else
+    platforms = expand_cxxstring_abis(filter(!Sys.isapple, supported_platforms()))
+    platforms_macos = [ Platform("x86_64", "macos") ]
```
giordano, line 87: Should also add aarch64?

Q/Qt6Base/build_tarballs.jl
```diff
@@ -134,10 +140,16 @@ dependencies = [
...
 
+if !host_build
+    push!(dependencies, HostBuildDependency("Qt6Base_jll"))
+end
+
 include("../../fancy_toys.jl")
 
-if any(should_build_platform.(triplet.(platforms_macos)))
-    build_tarballs(ARGS, name, version, sources, script, platforms_macos, products_macos, dependencies; preferred_gcc_version = v"8", julia_compat="1.6")
+@static if !host_build
```
giordano, line 149: For the record, a top-level `@static if` does nothing more than a simple `if`

Q/Qt6Base/build_tarballs.jl
barche: Yes, will do when adding all platforms.

Q/Qt6Base/build_tarballs.jl
barche: I got `platforms_macos` undefined with a regular if.

Issue #4511, opened 2022-02-27
new NetCDF_jll v400.702.402+0 broken on Windows


Alexander-Barth:
Unfortunately, the new NetCDF_jll v400.702.402+0 does not work on Windows (as far as I know Linux x86_64 and apple M1 are fine).

The new version of NetCDF_jll was created in this commit:
https://github.com/JuliaPackaging/Yggdrasil/pull/4481
 
The errors seem to be related to the upgrade of HDF5_jll v1.12.1+0.

Related bug reports:
https://github.com/Alexander-Barth/NCDatasets.jl/issues/164
https://github.com/JuliaGeo/NetCDF.jl/issues/151

This is the full error message when a Windows user (on julia 1.7.2) creates a NetCDF (with HDF5 backend) as reported @visr is below.

Is there a way to test via CI natively the library before releasing them?

```
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x2374cb3 -- .text at C:\Users\visser_mn\.julia\artifacts\2b6e2ce84250e36811c3019c1ad253c1739c888f\bin\libnetcdf-18.dll (unknown line)
in expression starting at REPL[13]:1
.text at C:\Users\visser_mn\.julia\artifacts\2b6e2ce84250e36811c3019c1ad253c1739c888f\bin\libnetcdf-18.dll (unknown line)
NC4_create at C:\Users\visser_mn\.julia\artifacts\2b6e2ce84250e36811c3019c1ad253c1739c888f\bin\libnetcdf-18.dll (unknown line)
NC_create at C:\Users\visser_mn\.julia\artifacts\2b6e2ce84250e36811c3019c1ad253c1739c888f\bin\libnetcdf-18.dll (unknown line)
nc__create at C:\Users\visser_mn\.julia\artifacts\2b6e2ce84250e36811c3019c1ad253c1739c888f\bin\libnetcdf-18.dll (unknown line)
nc_create at C:\Users\visser_mn\.julia\artifacts\2b6e2ce84250e36811c3019c1ad253c1739c888f\bin\libnetcdf-18.dll (unknown line)
top-level scope at .\REPL[13]:1
jl_toplevel_eval_flex at /cygdrive/c/buildbot/worker/package_win64/build/src\toplevel.c:876
[...]

PR #4826, opened 2022-04-24
[NetCDF] upgrade to NetCDF v4.8.1 for non-Windows platforms and julia 1.8 compat

Alexander-Barth:
Currently NetCDF_jll cannot be installed on any OS with julia 1.8 due to the fact that julia uses different versions of libcurl, LibSSH2, MbedTLS (than julia 1.6 and julia 1.7)

This PR:
* adapts the build script for julia 1.8
* updates NetCDF to the current version 4.8.1
* includes Windows specific patches to make NetCDF compile on Windows (the main patch `0001-Fix-a-number-of-OS-specific-bugs.patch` is from [upstream](https://github.com/Unidata/netcdf-c/pull/2138), but not yet available in a released version).
* runs `make check` on Linux (to hopefully catch issues earlier)
* unfortunately, we have the [same issue as before on Windows](https://github.com/JuliaPackaging/Yggdrasil/issues/4511) but the binary works on Linux and Mac OS X (x86_64): https://github.com/Alexander-Barth/NCDatasets.jl/actions/runs/2209847122
* I commented out Windows as supported platform because the binary is known to fail. Releasing a broken NetCDF_jll will create problems in packages depending on NetCDF_jll.

Windows issue is also reported upstream:
https://github.com/Unidata/netcdf-c/issues/2248

CC: @visr


giordano:
@Seelengrab this is why backporting https://github.com/JuliaLang/julia/pull/43250 would wreak havoc everywhere

N/NetCDF/bundled/patches/0001-Fix-a-number-of-OS-specific-bugs.patch
```diff
@@ -0,0 +1,675 @@
+From fcf179c6c799020871eed3ecbcc4dd34a9663c3b Mon Sep 17 00:00:00 2001
```
giordano, line 1: These are relatively large patches, touching lots of files.  It'd be good to upstream all changes that are general and not specific to our build system, to reduce our maintenance burden.

N/NetCDF/common.jl
```diff
@@ -89,12 +112,20 @@ nc-config --all
...
             Dependency("MbedTLS_jll", v"2.24.0"),
             Dependency("nghttp2_jll"),
+        ],
+        v"1.8" => [
+            Dependency("LibCURL_jll", v"7.81.0"),
+            # The following libraries are dependencies of LibCURL_jll which is now a
+            # stdlib, but the stdlib doesn't explicitly list its dependencies
+            Dependency("LibSSH2_jll", v"1.10.2"),
+            Dependency("MbedTLS_jll", v"2.28.0"),
+            Dependency("nghttp2_jll"),
```
giordano, line 128: The comment hasn't been true for quite some time (and it has been true for only few days).  Did you try removing these lines?

N/NetCDF/bundled/patches/0001-Fix-a-number-of-OS-specific-bugs.patch
Alexander-Barth: The large first patch is actually from upstream (@DennisHeimbigner) and merged in main but not jet available in a released version. This patch should no longer be necessary in upcoming releases. 

For the two other patches, I filled these issues/PRs:
https://github.com/Unidata/netcdf-c/issues/2303 (PR: https://github.com/Unidata/netcdf-c/pull/2304)
https://github.com/Unidata/netcdf-c/issues/2302 (PR: https://github.com/Unidata/netcdf-c/pull/2305)

N/NetCDF/common.jl
Alexander-Barth: Indeed, the compilation works if I remove the indirect dependencies. The PR is updated.

PR #4827, opened 2022-04-24
[Qt6Base] Version 6.3.0, all platforms

barche:


giordano:
```
[21:38:02] ninja: file is missing and not created by any action: '/workspace/destdir/lib/clang/11.0.1/lib/darwin/libclang_rt.osx.a'
```
Sigh.  It's very hard this is going to work

barche:
> ```
> 
> ```
> 
> 
>     
>   
> 
> Sigh.  It's very hard this is going to work

That error was my fault, seems one of the sed lines is no longer needed with 6.3. Unfortunately all the other failures seem to be a different thing each time, I'll need some time to investigate.

barche:
Stuck because on mingw the `GetDpiForWindow` function is not defined. Even with gcc 11.1, it is missing from `winuser.h`. I can confirm that on arch linux this function appears starting in the package `mingw-w64-headers-8.0.0`, but I am not sure which headers are used in BinaryBuilder, the mingw header version does not correspond to the GCC version.

giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/4cb8e928f01eff23f23149e44695efd0181b65f4/0_RootFS/gcc_common.jl#L272

barche:
So what's the best way forward? Bump that? Or attempt to shoe-horn the new version in just for this build, similar to the macOS SDK?

giordano:
> So what's the best way forward? Bump that?

@staticfloat now Julia requires Windows 10, would it make sense to do upgrade our MinGW to a version which supports Windows 10?  [`GetDpiForWindow`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getdpiforwindow#requirements) requires Windows 10.  But in any case it isn't that easy and it won't happen overnight, it requires rebuilding all versions of the compilers for Windows, _if_ they all still build cleanly....

> Or attempt to shoe-horn the new version in just for this build, similar to the macOS SDK?

If that's doable, that's probably a not-too-terrible stopgap solution.

Q/Qt6Base/build_tarballs.jl
```diff
@@ -8,14 +8,16 @@ version = v"6.3.0"
...
+const host_build = false
 
 # Collection of sources required to build qt6
 sources = [
     ArchiveSource("https://download.qt.io/official_releases/qt/$(version.major).$(version.minor)/$version/submodules/qtbase-everywhere-src-$version.tar.xz",
                   "b865aae43357f792b3b0a162899d9bf6a1393a55c4e5e4ede5316b157b1a0f99"),
     ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/11.0-11.1/MacOSX11.1.sdk.tar.xz",
                   "9b86eab03176c56bb526de30daa50fa819937c54b280364784ce431885341bf6"),
+    ArchiveSource("https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v10.0.0.tar.bz2",
+                  "ba6b430aed72c63a3768531f6a3ffc2b0fde2c57a3b251450dcf489a894f0894")
```
giordano, line 20: Maybe use 8 if that's sufficient?  For compatibility we usually want to use the oldest versions of these libraries, not the newest ones.

Q/Qt6Base/build_tarballs.jl
```diff
@@ -148,9 +173,9 @@ include("../../fancy_toys.jl")
 
 @static if !host_build
     if any(should_build_platform.(triplet.(platforms_macos)))
-        build_tarballs(ARGS, name, version, sources, script, platforms_macos, products_macos, dependencies; preferred_gcc_version = v"8", julia_compat="1.6")
+        build_tarballs(ARGS, name, version, sources, script, platforms_macos, products_macos, dependencies; preferred_gcc_version = v"11", julia_compat="1.6")
     end
 end
 if any(should_build_platform.(triplet.(platforms)))
-    build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"8", julia_compat="1.6")
+    build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"11.2", julia_compat="1.6")
```
giordano, line 180: Same here.  Is GCC 11 really _needed_?

Q/Qt6Base/build_tarballs.jl
barche: The 32-bit version only compiles starting with version 10. I don't think this actually imposes a restriction at runtime on the Windows version?

Q/Qt6Base/build_tarballs.jl
barche: No, 9 is sufficient, thanks!

giordano:
Ok, we'll see how this goes :slightly_smiling_face:

PR #4836, opened 2022-04-25
[GDAL] Try building on experimental platforms

asinghvi17:


G/GDAL/build_tarballs.jl
```diff
@@ -85,7 +85,7 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(supported_platforms())
+platforms = expand_cxxstring_abis(supported_platforms(; experimental = true))
```
giordano, line 88: The `experimental` keyword doesn't do anything special anymore.

ViralBShah:
Seems like GDAL is built on all the supported platforms right now. @asinghvi17 Any particular platform you are looking for that doesn't exist?

asinghvi17:
When trying to install ArchGDAL.jl on Mac M1 I had some issues, but it looks like the binary itself was correctly installed.  I think it ended up being a version conflict though.

```julia
ERROR: Error building `GDAL`:
┌ Warning: Platform `arm64-apple-darwin21.3.0` is not an officially supported platform
└ @ BinaryProvider ~/.julia/packages/BinaryProvider/U2dKK/src/PlatformNames.jl:450
ERROR: LoadError: KeyError: key "unknown" not found
```

PR #4837, opened 2022-04-25
New recipe: MrBayes-v3.2.7

BenjaminDoran:


M/MrBayes/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+    GitSource("https://github.com/NBISweden/MrBayes.git", "d50016695db24c58bcb36c83c487bd365fe2a566"),
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/MrBayes
+install_license ${WORKSPACE}/srcdir/MrBayes/COPYING
+
+## MPI VERSION
+if [[ "$target" == *-mingw* ]]; then
+    COMPILER=${target}-gcc
```
giordano, line 15: I noticed you use this pattern very often: why don't you simply use `gcc`, or even better, `cc` (or `${CC}`, which is equivalent)?

PR #4838, opened 2022-04-25
[NetCDFF] Trigger rebuild

barche:
This fixes the following error, but I don't know why:

```
ERROR: LoadError: InitError: could not load library "/home/bjanssens/tmp/depot-tmp/artifacts/b2b31d91ad92a26a1cfc9b4c44859110aac87754/lib/libnetcdff.so"
libhdf5_hl-383c339f.so.200.0.0: cannot open shared object file: No such file or directory
```

Needed to get https://github.com/JuliaBinaryWrappers/FLEXPART_jll.jl registered.

CC @tcarion

### DIFF
--- a/N/NetCDFF/build_tarballs.jl
+++ b/N/NetCDFF/build_tarballs.jl
@@ -51,3 +51,4 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+


giordano:
Looks like some compatibility bounds should be fixed in the registry.  Not here.

giordano:
Yes, the fact is that libnetcdff directly links to hdf5, without specifying a compat bound with it:
```
shell> readelf -d /home/mose/.julia/artifacts/c60b2e777d8803af862086ac213a4248bea19b48/lib/libnetcdf.so

Dynamic section at offset 0x15a3a8 contains 38 entries:
 0x0000000000000001 (NEEDED)             Shared library: [libhdf5_hl.so.200]
 0x0000000000000001 (NEEDED)             Shared library: [libhdf5.so.200]
 0x0000000000000001 (NEEDED)             Shared library: [libcurl.so.4]
 0x0000000000000001 (NEEDED)             Shared library: [libnghttp2.so.14]
 0x0000000000000001 (NEEDED)             Shared library: [libssh2.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libmbedtls.so.13]
 0x0000000000000001 (NEEDED)             Shared library: [libmbedx509.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libmbedcrypto.so.5]
 0x000000000000000e (SONAME)             Library soname: [libnetcdf.so.18]
```
This is definitely something to fix in the reigstry.

barche:
I found that setting the NetCDF compat to `NetCDF_jll = "400.701.400-400.702.401"` fixes this, is that the correct way to go and should that be a PR to Julia General registry directly?

giordano:
That'd probably be a proxy for the version of hdf5 needed by netcdff, but I guess it's OK. Yes, a pr to the registry would be great

PR #4849, opened 2022-04-29
Update Luna to v0.26.2

ararslan:
No idea whether this will work without further changes. Let's find out!


ararslan:
Ah the patch file needs to be updated...

giordano:
For some reasons I thought updating the patch would have been easy and relaxing....

ararslan:
Wow, thanks so much for doing that!

PR #4858, opened 2022-05-02
New Telescope version

jhigginbotham64:
Some platforms fail saying that Bullet needs to be recompiled with -fPIC. Fair enough, I overlooked that in my original PR, it's an easy fix.

Others fail saying that they can't find the header files for SDL Mixer, which...I'm going to need some help figuring out.

### DIFF
--- a/T/Telescope/build_tarballs.jl
+++ b/T/Telescope/build_tarballs.jl
@@ -40,7 +40,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    BuildDependency("Bullet_jll"),
+    BuildDependency("Bullet_Physics_SDK_jll"),
     BuildDependency("OpenGLMathematics_jll"),
     Dependency("SDL2_jll"),
     Dependency("SDL2_image_jll"),
--- a/T/Telescope/build_tarballs.jl
+++ b/T/Telescope/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder, Pkg
 
 name = "Telescope"
-version = v"0.1.2"
+version = v"0.2.0"
 
 # Collection of sources required to build this package
 sources = [
     ArchiveSource("https://github.com/jhigginbotham64/Telescope/archive/refs/tags/v$(version).tar.gz",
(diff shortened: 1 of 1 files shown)


jhigginbotham64:
Progress! All the remaining build errors are due to the compiler or whatever not being able to find SDL2/SDL_mixer.h. But I'm confused, the only recommended fix from [the docs](https://docs.binarybuilder.org/stable/troubleshooting/#Header-files-of-the-dependencies-can't-be-found) is already being used (setting CFLAGS and/or CXXFLAGS), and it works *most* of the time...

giordano:
It isn't very surprising and the explanation is in the warnings:
```
┌ Warning: Dependency SDL2_mixer_jll does not have a mapping for artifact SDL2_mixer for platform aarch64-apple-darwin20-libgfortran5-cxx11
└ @ BinaryBuilderBase /depot/packages/BinaryBuilderBase/fLu4a/src/Prefix.jl:676
```
`SDL2_mixer_jll` isn't available for those two platforms (which have been introduced only recently and we still don't have all packages built for them)

jhigginbotham64:
Referring to aarch64-apple-darwin and armv6l-linux? That makes sense. What do I do in this case? Wait? Or could I head over to those packages and try to get them building on the new platforms myself?
(also @giordano thanks for the quick response, as usual!)

jhigginbotham64:
Is there a simple way to disable experimental platforms for this package? I'd like to support them once fixes for the dependencies are available, but for now it's more important to keep my project moving.

jhigginbotham64:
At least it's a topic of discussion over there: https://github.com/libsdl-org/SDL_mixer/issues/369

T/Telescope/build_tarballs.jl
```diff
@@ -28,7 +29,7 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(; exclude=p->!(Sys.islinux(p) && nbits(p) == 64))
+platforms = supported_platforms()
```
giordano, line 33:
```suggestion
filter!(p -> arch(p) != "armv6l" & !(Sys.isapple(p) && arch(p) == "aarch64"), platforms)
```

jhigginbotham64: Thank you so very much.

PR #4861, opened 2022-05-02
[MiniZinc] add a builder for MiniZinc

odow:


M/MiniZinc/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir
+
+mkdir -p libminizinc/build
+cd libminizinc/build
+
+# Patch for MinGW toolchain
+find . -type f -exec sed -i 's/Windows.h/windows.h/g' {} +
```
odow, line 23:
```suggestion
find .. -type f -exec sed -i 's/Windows.h/windows.h/g' {} +
```

odow:
So many string issues on Windows. This JLL is going to be the base of a large-ish Julia package I need to write, so I wonder if we can merge without Windows support. Then I can get things implemented and tested on Mac and linux, and come back to Windows support before I release MiniZinc.jl as a public package.

M/MiniZinc/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+make install
+"""
+
+products = [
+    ExecutableProduct("minizinc", :minizinc),
+]
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
```
giordano, line 51:
```suggestion
platforms = expand_cxxstring_abis(supported_platforms())
```

odow:
Thanks, as always, @giordano

PR #4866, opened 2022-05-03
DNMY: [MiniZinc] add debug build

odow:
I'm running into segfaults trying to use the binary. (It can display `--help`, etc, but not solve a model.) Hopefully debug helps identify the cause.

### DIFF
--- a/M/MiniZinc/build_tarballs.jl
+++ b/M/MiniZinc/build_tarballs.jl
@@ -4,12 +4,12 @@ using BinaryBuilder, Pkg
 
 name = "MiniZinc"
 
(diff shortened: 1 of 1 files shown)


odow:
I'm using this in https://github.com/jump-dev/MiniZinc.jl. Current status:

 * Linux : runs successfully https://github.com/jump-dev/MiniZinc.jl/runs/6283823235?check_suite_focus=true
 * Mac: segfaults https://github.com/jump-dev/MiniZinc.jl/runs/6283823285?check_suite_focus=true
 * Windows: does not build with minnow

PR #4876, opened 2022-05-06
[jlqml] Version 0.4 (Qt 6)

barche:
New version that depends on Qt6. The Qt5 support remains in 0.3, but no further updates there are planned.


barche:
This seems OK to me.

PR #4880, opened 2022-05-06
[jemalloc] Drop Apple Silicon special case (fixed in 5.3)

jeremiahpslewis:


giordano:
This shouldn't affect the resulting binary, right?

jeremiahpslewis:
No, fix hard codes 14 on the jemalloc side: https://github.com/jemalloc/jemalloc/commit/7ae0f15c598258610dd3cfd9633301ffa8661c45

giordano:
Ok, and I can confirm the libraries are identical (same sha256 as before).  So I'm going to add `[skip ci]` to avoid rebuilding tarballs needlessly.

jeremiahpslewis:
Thanks!

PR #4882, opened 2022-05-07
GifLib extra initial recipe

ashwani-rathee:
This adds recipe for tarball build of Giflib programs and additional apis that are not provided in libgif already provided here.

### DIFF
--- a/L/libgifextra/build_tarballs.jl
+++ b/L/libgifextra/build_tarballs.jl
@@ -1,0 +1,50 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Giflibextra"
+version = v"0.0.1"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/ashwani-rathee/io-project/raw/main/gifwrapper/gen/libgifextra.zip", "4e1d3e0d1e9df0f30e0dc6c907bb2d574d4c4efec79fb1558451d4c24fd3430b")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd libgifextra/
+make
+make install
+exit
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libgifextra", :libgifextra)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="Giflib_jll", uuid="59f7168a-df46-5410-90c8-f2779963d0ec"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")


L/libgifextra/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Giflibextra"
+version = v"0.0.1"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/ashwani-rathee/io-project/raw/main/gifwrapper/gen/libgifextra.zip", "4e1d3e0d1e9df0f30e0dc6c907bb2d574d4c4efec79fb1558451d4c24fd3430b")
```
johnnychen94, line 10: We need to make this a standalone git repository first; the `io-project` repository is only for experimental purposes.

L/libgifextra/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Giflibextra"
```
giordano, line 5: This name _**must**_ match the name of the directory where the file is in

L/libgifextra/build_tarballs.jl
giordano: Yes, a standalone repository would be nicer.  Also, _**never**_ reference a branch name, like `main`, in the URL: it's a moving target, if you push another commit to the repository this recipe won't work anymore, which isn't acceptable.

L/libgifextra/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+    ArchiveSource("https://github.com/ashwani-rathee/io-project/raw/main/gifwrapper/gen/libgifextra.zip", "4e1d3e0d1e9df0f30e0dc6c907bb2d574d4c4efec79fb1558451d4c24fd3430b")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd libgifextra/
+make
+make install
+exit
```
giordano, line 19:
```suggestion
cd $WORKSPACE/srcdir/libgifextra/
make -j${nproc}
make install
```

L/libgifextra/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
```
giordano, line 36: Any particular reasons for not building on non-Linux platforms?
```suggestion
platforms = supported_platforms(; exclude=!Sys.islinux)
```

L/libgifextra/build_tarballs.jl
johnnychen94: > current name in build_tarballs.jl will work right?

I'm not sure if I understand this; do you mean the case differences "Giflibextra" vs "GifLibExtra"? I believe BB doesn't differentiate this but it's better to be consitent.

---

When GifLibExtra is ready (with valid licenses), you can tag a release v0.0.1. Then I believe it should properly pass Yggdrasil build, assuming the build script works.

L/libgifextra/build_tarballs.jl
```diff
@@ -0,0 +1,36 @@
...
+sources = [
+    GitSource("https://github.com/ashwani-rathee/GifLibExtra.git", "a5c74d671dbede5f4be5448f80a1fd6950ae38b7")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/GifLibExtra.git/lib
+make -j${nproc}
+make install
+exit
```
giordano, line 18:
```suggestion
cd $WORKSPACE/srcdir/GifLibExtra/lib
make -j${nproc}
make install
```

L/libgifextra/build_tarballs.jl
```diff
@@ -0,0 +1,36 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "GifLibExtra"
```
giordano, line 5: This name must match exactly the name of the directory, so either change this name or rename the directory.  I'd probably suggest to rename the directory.

ashwani-rathee:
Hey @giordano ,  can you please check what's the issue right now?
It does build for linux systems atleast but still doesn't pass check:
https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=20555&view=logs&jobId=d9dce285-6819-5ada-107d-df8db81e965a&j=d9dce285-6819-5ada-107d-df8db81e965a&t=c97541ec-1686-510a-4596-692d423c1f65

giordano:
Quoting from https://github.com/JuliaPackaging/Yggdrasil/pull/4882#discussion_r867361119

> This _**name**_ must match the name of the directory where the file is in

I don't know why you marked that thread as resolved, since it very much was not

L/libgifextra/build_tarballs.jl
ashwani-rathee: But the name of the git directory is https://github.com/ashwani-rathee/GifLibExtra , is `GifLibExtra` which is same as name here. I am sure I am missing something sorry, can you tell? I think I need to change the Git directory name to libgifextra only?and everything else too with GifLibExtra

giordano: Quoting (again) from https://github.com/JuliaPackaging/Yggdrasil/pull/4882#discussion_r867361119

> This name _**must**_ match the name of the directory where the file is in

The directory of https://github.com/JuliaPackaging/Yggdrasil/blob/bf459f5b2f86641e0f48f027c37189da61ff96d6/L/libgifextra/build_tarballs.jl is called `L/libgifextra`

ashwani-rathee: Okay, parts of the problem solved. Thanks!!!

giordano:
Some comments about https://github.com/ashwani-rathee/libgifextra/blob/f0eb483f67d53ee0dd4d78581623c158677aed88/makefile:

* `TYPE` can just be `-shared` on all platforms
* any specific reason for using `-mmacosx-version-min=10.1`?  That's pretty ancient, BinaryBuilder already sets `-mmacosx-version-min` anyway (but more reasonable values, depending on the architecture)
* I'm not sure why you're using `-Wl,-whole-archive`
* if you're willing to use [BinaryBuilder environment variables](https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables) (since you're already using `$(prefix)`, which isn't defined anywhere), then `LIBEXT` is simply `$(dlext)`
* if you put together all above points, you can see the entire `UNAME` block is useless
* whenever you link a shared library installed in the prefix, like `-lgif`, you also need the linker _where_ to find it, for example with `-L$(libdir)`
* the `install` rule can be simplified to
  ```makefile
  install: libgifextra.$(LIBEXT)
  	install -Dvm 755 libgifextra.$(dlext) $(libdir)/libgifextra.$(dlext)
  	install -Dvm 644 include/libgifextra.h $(includedir)/libgifextra.h
  ```
  In particular note that:
  - the shared library must be executable
  - shared libraries have to be installed in `$(libdir)`, which is _not_ necessarily the same as `$(prefix)/lib` (notably, not on Windows)
  - you can avoid manually creating the directory

giordano (approved):
Looking good now, thanks!

PR #4886, opened 2022-05-09
[FastTransforms] v0.6.0

MikaelSlevinsky:


MikaelSlevinsky:
I don't know why this would run into `internal compiler error: Segmentation fault`. Never seen this before

giordano:
Eh, there isn't much to do with an ICE apart from using a different compiler. You can try using gcc 7 instead of 6, in the worst case we can skip the platform.

MikaelSlevinsky:
Thanks! I'll try different compilers

MikaelSlevinsky:
Phew!

PR #4888, opened 2022-05-10
[gmsh] Add FLTK and OpenCASCADE packages for Gmsh builds

sebastiangrimberg:
Adds FLTK, and OpenCASCADE dependencies for Gmsh build. Also adds LinearElasticity which is a dependency for MMG, also to be used in Gmsh build (see #4891).

PR #4889 contains the updates to the Gmsh build to take advantage of these additions.


F/FLTK/build_tarballs.jl
```diff
@@ -0,0 +1,79 @@
...
+    -DOPTION_USE_SYSTEM_LIBJPEG=ON \
+    -DOPTION_USE_SYSTEM_LIBPNG=ON \
+    -DOPTION_USE_SYSTEM_ZLIB=ON \
+    -DOPTION_USE_THREADS=ON
+make -j${nproc}
+make install
+if [[ ${target} == *mingw* ]]; then
+    # Rename binaries to correct extension
+    mv ${bindir}/fltk-config ${bindir}/fltk-config${exeext}
+fi
```
giordano, line 40: I'm not sure it's important to have fltk-config listed in the products

F/FLTK/build_tarballs.jl
```diff
@@ -0,0 +1,79 @@
...
+cd ..
+install_license COPYING
```
giordano, line 42:
```suggestion
install_license ../COPYING
```

F/FLTK/build_tarballs.jl
```diff
@@ -0,0 +1,79 @@
...
+# platforms are passed in on the command line
+platforms = supported_platforms(exclude= p -> arch(p) == "armv6l")
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libfltk", :libfltk),
+    LibraryProduct("libfltk_forms", :libfltk_forms),
+    LibraryProduct("libfltk_gl", :libfltk_gl),
+    LibraryProduct("libfltk_images", :libfltk_images),
+    ExecutableProduct("fltk-config", Symbol("fltk-config"))
```
giordano, line 55: Same comment as above.  Also, calling a variable `fltk-config` doesn't make it very easy to use it in practice.  `fltk_config` would be a better option, but again, I'd simply remove it.

M/MMG/build_tarballs.jl
```diff
@@ -0,0 +1,103 @@
...
+version = v"5.6.0"
+
+# Collection of sources required to build MMG
+sources = [
+    ArchiveSource("https://github.com/MmgTools/mmg/archive/refs/tags/v$(version).tar.gz",
+                  "bbf9163d65bc6e0f81dd3acc5a51e4a8c47a7fdae849abc26277e01154fe2437"),
+    ArchiveSource("https://github.com/ISCDtoolbox/Commons/archive/refs/heads/master.zip",
+                  "02fae58388a93692f3c73dcb4619bbcf838b479744cf6972edeca330f756125a"),
+    ArchiveSource("https://github.com/ISCDtoolbox/LinearElasticity/archive/refs/heads/master.zip",
+                  "b1898151fbc4294d05d1df2eade209f2e0e185e6e49f38b6fd36a1333dd591fd"),
```
giordano, line 15: Please, **NEVER*** use snapshots of branches like this.  Branches are moving target, this recipe would become broken when they push to `master` again.  Also, why these are their own packages?

O/OCCT/build_tarballs.jl
```diff
@@ -0,0 +1,117 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "OCCT"
+version = v"7.6.2"
+
+# Collection of sources required to build Open CASCADE Technology (OCCT)
+sources = [
+    FileSource("https://git.dev.opencascade.org/gitweb/?p=occt.git;a=snapshot;h=refs/tags/V7_6_2;sf=tgz",
```
giordano, line 10: Might be easier to use a [`GitSource`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.GitSource), no?

giordano (commented):
Thanks for your contribution

F/FLTK/build_tarballs.jl
sebastiangrimberg: I agree, removed in b12f87e.

F/FLTK/build_tarballs.jl
sebastiangrimberg: Resolved in b12f87e.

M/MMG/build_tarballs.jl
sebastiangrimberg: Sorry about that, I wasn't sure since these dependencies aren't versioned very well. I've separated out LinearElasticity into it's own package now and changed the source to a particular commit. Note that now the pipelines for MMG will fail  for now because LinearElasticity is not registered.

O/OCCT/build_tarballs.jl
sebastiangrimberg: Yep I agree, resolved in b12f87e.

sebastiangrimberg:
Thanks for your feedback @giordano! I have addressed your comments. In particular regarding the MMG build I created a new LinearElasticity package which will need to be registered before the MMG pipelines can be tested. I can move MMG into a new PR in order to separate this out if you would like.

giordano:
> I can move MMG into a new PR in order to separate this out if you would like.

That'd be helpful, yes, as we can't merge PRs with failing checks.

L/LinearElasticity/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+
+cd ${WORKSPACE}/srcdir/LinearElasticity
+if [[ "${target}" == *mingw* ]]; then
+    atomic_patch -p1 "${WORKSPACE}/srcdir/patches/LinearElasticity.mingw.patch"
+else
+    atomic_patch -p1 "${WORKSPACE}/srcdir/patches/LinearElasticity.patch"
+fi
+if [[ "${target}" == *freebsd* ]]; then
+    LDFLAGS="-lpmc -lc++"
+fi
```
giordano, line 45: Why?

O/OCCT/bundled/patches/STEPConstruct_AP203Context.cxx.patch
```diff
@@ -0,0 +1,11 @@
+--- /dev/null
```
giordano, line 1: It'd be good to upstream all generic patches, like this one I think?

O/OCCT/build_tarballs.jl
```diff
@@ -0,0 +1,115 @@
...
+    -DBUILD_LIBRARY_TYPE=Shared \
+    -DBUILD_MODULE_Draw=0 \
+    -DBUILD_MODULE_Visualization=0 \
+    -DBUILD_MODULE_ApplicationFramework=0
+make -j${nproc}
+make install
+if [[ ${target} == *mingw* ]]; then
+    # Correct installation paths for windows
+    ln -s ${prefix}/win*/gcc/lib/* ${prefix}/lib
+    ln -s ${prefix}/win*/gcc/bin/* ${prefix}/bin
```
giordano, line 42: Could maybe move instead of creating links (which become copies anyway in the final tarball)?

L/LinearElasticity/build_tarballs.jl
sebastiangrimberg: In the Commons dependency for FreeBSD builds, there are some functions from the `pmc` library used which lead to linker errors for undefined references when building LinearElasticity without those extra link flags. I agree it isn't very nice, any suggestions on a better way to resolve?

O/OCCT/build_tarballs.jl
sebastiangrimberg: Is this safe to just move the `.dll` outputs? I could also patch the `CMakeLists.txt` for Windows builds in order to install to just `${prefix}/bin` and `${prefix}/lib` if that would be preferred.

O/OCCT/bundled/patches/STEPConstruct_AP203Context.cxx.patch
sebastiangrimberg: Do you mean to file an issue report with the OpenCASCADE Technology developers? I can try to do this but worry about response time.

giordano: Pull requests, yes, but I wasn't suggesting for wait for them to accept those PRs before merging this one.  It's just that upstreaming the relevant patches reduces our maintenance burden.

O/OCCT/build_tarballs.jl
giordano: > Is this safe to just move the .dll outputs?

I have no idea about the specific case, I don't know the package you're building here.  There could be a problem if there are somehow references to relative paths, but they'd break anyway as it is now: when you create a symlink for windows we automatically make a copy when we create the tarball, as Windows doesn't support very well symlinks.  So either way we get into exactly the same state.

> I could also patch the CMakeLists.txt for Windows builds in order to install to just ${prefix}/bin and ${prefix}/lib if that would be preferred.

Note that both executables and shared libraries for windows should be under `${prefix}/bin`, even though it's always better to use the appropriate environment variables `${bindir}` and `${libdir}`: https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables

sebastiangrimberg: Agreed, this is resolved by patching the install directory path for Windows systems in 612be31 so no move/copy of the libraries is required.

O/OCCT/bundled/patches/STEPConstruct_AP203Context.cxx.patch
sebastiangrimberg: Totally agree, will follow up with the upstream repo.

L/LinearElasticity/build_tarballs.jl
sebastiangrimberg: This is more nicely resolved by adding the linker dependencies directly to Commons in the CMakeLists.txt file. Resolved in 612be31.

PR #4889, opened 2022-05-10
[gmsh] Update to v4.10.2 and use MMG, FLTK, and OpenCASCADE

sebastiangrimberg:
Updates Gmsh to v4.10.2 and configures to use OpenCASCADE, FLTK, and MMG dependencies.

Depends on #4888 and #4891.

### DIFF
--- a/G/gmsh/build_tarballs.jl
+++ b/G/gmsh/build_tarballs.jl
@@ -8,7 +8,7 @@ version = v"4.10.2"
 # Collection of sources required to build Gmsh
 sources = [
     ArchiveSource("https://gmsh.info/src/gmsh-$(version)-source.tgz",
-                  "d921109e6c419f68d5507d9612ee815d40b5239a5db175f28285e0cd0dbf2517")
+                  "d921109e6c419f68d5507d9612ee815d40b5239a5db175f28285e0cd0dbf2517"),
 ]
 
 # Bash recipe for building across all platforms
--- a/G/gmsh/build_tarballs.jl
+++ b/G/gmsh/build_tarballs.jl
@@ -16,10 +16,6 @@ script = raw"""
 cd ${WORKSPACE}/srcdir/gmsh-*
 if [[ "${target}" == *linux* ]] || [[ "${target}" == *freebsd* ]]; then
     OPENGL_FLAGS="-DOpenGL_GL_PREFERENCE=LEGACY"
-elif [[ "${target}" == *mingw* ]]; then
-    # Built-in FindHDF5.cmake has issues on Windows
-    sed -i'.bak' 's/^  find_package(HDF5)/  # find_package(HDF5)/' CMakeLists.txt
-    HDF5_FLAGS="-DHDF5_FOUND=true -DHDF5_INCLUDE_DIRS=${includedir} -DHDF5_C_LIBRARIES=$(ls ${libdir}/libhdf5*.${dlext})"
 fi
 mkdir build
 cd build
@@ -29,7 +25,7 @@ cmake .. \
     -DCMAKE_BUILD_TYPE=Release \
     -DENABLE_BUILD_DYNAMIC=1 \
     -DDEFAULT=1 \
-    ${OPENGL_FLAGS} ${HDF5_FLAGS}
+    ${OPENGL_FLAGS}
 make -j${nproc}
 make install
 mv ${prefix}/lib/gmsh.jl ${prefix}/lib/gmsh.jl.bak
@@ -44,7 +40,6 @@ install_license ../LICENSE.txt
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = expand_cxxstring_abis(supported_platforms())
-filter!(p -> !(Sys.iswindows(p) && cxxstring_abi(p) == "cxx03"), platforms) # Restriction from OCCT builds
 
 # The products that we will ensure are always built
 products = [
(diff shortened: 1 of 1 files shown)


G/gmsh/build_tarballs.jl
```diff
@@ -30,22 +38,48 @@ sed ${prefix}/lib/gmsh.jl.bak \
...
 dependencies = [
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
+    BuildDependency("Xorg_xorgproto_jll"; platforms=x11_platforms),
+    Dependency("Cairo_jll"),
+    Dependency("CompilerSupportLibraries_jll"),
+    Dependency("FLTK_jll"),
+    Dependency("FreeType2_jll"),
+    Dependency("GLU_jll"; platforms=x11_platforms),
+    Dependency("GMP_jll"),
+    Dependency("HDF5_jll"),
```
giordano, line 67: You should really put a compat bound for HDF5, we've had lots of different incompatible versions.  Also GMP needs a compat bound

G/gmsh/build_tarballs.jl
```diff
@@ -30,22 +38,48 @@ sed ${prefix}/lib/gmsh.jl.bak \
...
 
+# Some dependencies are needed only on Linux and FreeBSD
+x11_platforms = filter(p ->Sys.islinux(p) || Sys.isfreebsd(p), platforms)
+
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
+    BuildDependency("Xorg_xorgproto_jll"; platforms=x11_platforms),
+    Dependency("Cairo_jll"),
+    Dependency("CompilerSupportLibraries_jll"),
```
giordano, line 62: Is this needed only for libgomp?

G/gmsh/build_tarballs.jl
```diff
@@ -30,22 +38,48 @@ sed ${prefix}/lib/gmsh.jl.bak \
...
+    Dependency("FLTK_jll"),
+    Dependency("FreeType2_jll"),
+    Dependency("GLU_jll"; platforms=x11_platforms),
+    Dependency("GMP_jll"),
+    Dependency("HDF5_jll"),
+    Dependency("JpegTurbo_jll"),
+    Dependency("Libglvnd_jll"; platforms=x11_platforms),
+    Dependency("libpng_jll"),
+    Dependency("LLVMOpenMP_jll"; platforms=filter(Sys.isapple, platforms)),
```
giordano, line 71: `LLVMOpenMP_jll` is used also on FreeBSD, so this should be
```suggestion
    Dependency("LLVMOpenMP_jll"; platforms=filter(Sys.isbsd, platforms)),
```

G/gmsh/build_tarballs.jl
sebastiangrimberg: Added `compat` bounds for the current versions.

G/gmsh/build_tarballs.jl
sebastiangrimberg: I believe so, this was the only dependency for the prior 4.9.3 build. It seems like it finds OpenMP and builds fine without it, though.

G/gmsh/build_tarballs.jl
sebastiangrimberg: Fixed.

sebastiangrimberg:
The FreeBSD build is failing due to a missing patch in OCCT. PR #4910 addresses this.

G/gmsh/build_tarballs.jl
giordano: Then I wanted to suggest to limit it to `!Sys.isbsd` platforms (for the others we have llvm openmp), not to remove it completely 🙂

sebastiangrimberg: 🙂

G/gmsh/build_tarballs.jl
```diff
@@ -30,22 +34,56 @@ sed ${prefix}/lib/gmsh.jl.bak \
   -e 's/^\(const lib.*\)/#\1/g' \
   -e 's/^\(module gmsh\)$/\1\nusing gmsh_jll: libgmsh\nconst lib = libgmsh/g' \
   > ${prefix}/lib/gmsh.jl
+install_license ../LICENSE.txt
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = expand_cxxstring_abis(supported_platforms())
+filter!(p -> !(Sys.iswindows(p) && cxxstring_abi(p) == "cxx03"), platforms) # Restriction from OCCT builds
```
giordano, line 43: I had missed this:
https://github.com/JuliaPackaging/Yggdrasil/blob/892c5abc72b0c74f864732ea437e477419f37774/O/OCCT/build_tarballs.jl#L47
and I don't understand it.  Please see https://docs.binarybuilder.org/stable/building/#Expanding-C-string-ABIs-or-libgfortran-versions

> This ABI does _not_ have to do with the C++ standard used by the source code, in fact you can build a C++03 library with the C++11 `std::string` ABI and a C++11 library with the C++03 `std::string` ABI. This is achieved by appropriately setting the `_GLIBCXX_USE_CXX11_ABI` macro.

sebastiangrimberg: This was something I noticed when building OCCT in #4888. For windows builds, I got errors about from the code using std::swap except when disabling the C++03 ABI. I can look back into it a little more.

giordano: It'd be good to see _what_ is exactly the error.  And again, the C++ string ABI is completely different from the standard used in the source code.  Since the result here would be the narrowing of the supported platforms for an existing platform, this isn't a desirable outcome.

sebastiangrimberg: Actually, I think I was testing incorrectly and this actually is not required you are right, I'm sorry about that. Fixed in #4911 and here in the most recent update.

sebastiangrimberg: Thanks for your help. I think we need to somehow rerun the Windows C++03 pipelines to use the new OpenCASCADE dependency build, but otherwise things look good. Thank you again!

G/gmsh/build_tarballs.jl
```diff
@@ -39,13 +44,45 @@ platforms = expand_cxxstring_abis(supported_platforms())
...
 dependencies = [
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
+    BuildDependency("Xorg_xorgproto_jll"; platforms=x11_platforms),
+    Dependency("Cairo_jll"),
+    Dependency("CompilerSupportLibraries_jll"; platforms=filter(!Sys.isbsd, platforms)),
+    Dependency("FLTK_jll"),
+    Dependency("FreeType2_jll"),
+    Dependency("GLU_jll"; platforms=x11_platforms),
+    Dependency("GMP_jll"; compat="6.2"),
+    Dependency("HDF5_jll"; platforms=hdf5_platforms, compat="1.12"),
```
giordano, line 70: Our build of HDF5 1.12.0 [doesn't have aarch64-apple-darwin](https://github.com/JuliaBinaryWrappers/HDF5_jll.jl/releases/tag/HDF5-v1.12.0%2B1), contrary to what you're saying above in `hdf5_platforms`.  You have to either use [1.12.1](https://github.com/JuliaBinaryWrappers/HDF5_jll.jl/releases/tag/HDF5-v1.12.1%2B0) (recommended) or remove aarch64-apple-darwin from the list above.

sebastiangrimberg: Resolved by using 1.12.1 which supplies the aarch64 build for MacOS.

PR #4894, opened 2022-05-11
[ChibiScheme] add builder

stevengj:
Add a builder for [Chibi-Scheme](https://github.com/ashinn/chibi-scheme), which is a minimal self-contained embeddable Scheme interpreter.

### DIFF
--- a/C/ChibiScheme/build_tarballs.jl
+++ b/C/ChibiScheme/build_tarballs.jl
@@ -23,7 +23,7 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(; experimental=true)
+platforms = supported_platforms()
 
 # The products that we will ensure are always built
 products = [
--- a/C/ChibiScheme/build_tarballs.jl
+++ b/C/ChibiScheme/build_tarballs.jl
@@ -17,13 +17,6 @@ apk add chibi-scheme
 
 make \
     PREFIX="${prefix}" \
-    CC="${CC_BUILD}" \
-    AR="${AR_BUILD}" \
-    LD="${LD_BUILD}" \
-    CFLAGS="${CFLAGS}" \
-    CPPFLAGS="${CPPFLAGS}" \
-    LDFLAGS="${LDFLAGS}" \
-    RANLIB="${RANLIB_BUILD}" \
     CHIBI=`which chibi-scheme` \
     CHIBI_FFI=`which chibi-ffi` \
     CHIBI_DOC=`which chibi-doc` \
--- a/C/ChibiScheme/build_tarballs.jl
+++ b/C/ChibiScheme/build_tarballs.jl
@@ -13,18 +13,33 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/chibi-scheme
 
+# apply chibi-scheme#830
 patch CMakeLists.txt <<EOF
-@@ -202,10 +202,11 @@ if(BUILD_SHARED_LIBS)
-     # is not a generator expression within the actual custom command to process
-     # the stubs, as older CMake versions fail to properly construct the dependency
-     # on the bootstrap executable from the generator expression.
--    set(bootstrap chibi-scheme)
-+    set(_bootstrap chibi-scheme)
- else()
--    set(bootstrap chibi-scheme-bootstrap)
-+    set(_bootstrap chibi-scheme-bootstrap)
- endif()
-+set(bootstrap ${_bootstrap} CACHE FILEPATH "chibi-scheme path for bootstrapping")
+@@ -209,0 +210,3 @@ endif()
++# when cross-compiling, we need to use a separately built chibi-scheme executable:
++set(BOOTSTRAP \${bootstrap} CACHE FILEPATH "chibi-scheme path for bootstrapping")
++
+@@ -225 +228 @@ function(add_stubs_library stub)
+-        COMMAND \${bootstrap} \${chibi-ffi} \${stubfile} \${stubout}
++        COMMAND \${BOOTSTRAP} \${chibi-ffi} \${stubfile} \${stubout}
+@@ -384 +387 @@ foreach(e \${chibi-scheme-tests})
+-        COMMAND chibi-scheme -I \${CMAKE_CURRENT_BINARY_DIR}/lib tests/\${e}.scm
++        COMMAND \${BOOTSTRAP} -I \${CMAKE_CURRENT_BINARY_DIR}/lib tests/\${e}.scm
+@@ -389 +392 @@ add_test(NAME r5rs-test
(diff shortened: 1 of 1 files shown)


C/ChibiScheme/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; experimental=true)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libchibi-scheme", :libchibischeme),
+    FileProduct("share/chibi/init-7.scm, :init_7_scm),
```
giordano, line 31:
```suggestion
    FileProduct("share/chibi/init-7.scm", :init_7_scm),
```

C/ChibiScheme/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
```
giordano, line 35:
```suggestion
dependencies = Dependency[
```

C/ChibiScheme/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    ..
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; experimental=true)
```
giordano, line 26: `experimental=true` isn't necessary anymore
```suggestion
platforms = supported_platforms()
```

C/ChibiScheme/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+# Collection of sources required to build NLopt
+sources = [
+    GitSource("https://github.com/ashinn/chibi-scheme.git",
+              "b0735b3ca70620face209c5066898e5e9e1fcf62"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/chibi-scheme
+mkdir build && cd build
```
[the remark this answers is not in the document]
giordano: It provides the command `chibi-scheme` which, as far as I can tell, isn't built anywhere by this build system, not even for x86_64-linux-musl which is a pure native build, not a cross one.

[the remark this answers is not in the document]
giordano: No, wait, [it's built](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=19606&view=logs&j=abd022a1-9f63-5341-d138-0709aa033c2f&t=ac164843-b50a-5bf6-a62e-0593feb78eb2&l=386) but then they magically expect it to be in the PATH?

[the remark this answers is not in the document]
giordano: Anyway, this wouldn't work for all platforms because we're building for the target system, and CMake conveniently does _**not**_ allow you to specify a host toolchain and a target toolchain at the same time (contrary to autoconf with which you can set `CC_BUILD` and `CC` separately, which is what we do automatically in BinaryBuilder).

So the options are:

1. to do a native bootstrap and use that to build the package for the target,
2. or to avoid manually doing the bootstraps and just use the package from Alpine Linux to get the `chibi-scheme` command.

C/ChibiScheme/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
++set(bootstrap ${_bootstrap} CACHE FILEPATH "chibi-scheme path for bootstrapping")
+EOF
+
+apk add chibi-scheme
+
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -Dbootstrap=`which chibi-scheme`
```
giordano, line 36:
```suggestion
    -Dbootstrap=`which chibi-scheme` \
```

C/ChibiScheme/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+    RANLIB="${RANLIB_BUILD}" \
+    CHIBI=`which chibi-scheme` \
+    CHIBI_FFI=`which chibi-ffi` \
+    CHIBI_DOC=`which chibi-doc` \
+    SNOW_CHIBI=`which snow-chibi` \
+    SO=".${dlext}" \
+    CLIBFLAGS="-fPIC" \
+    CLINKFLAGS="-shared" \
+    STATICFLAGS="-static -DSEXP_USE_DL=0" \
+    LIBCHIBI_FLAGS="-Wl,-soname,libchibi-scheme.${dlext}.0" \
```
giordano, line 33: I'm not sure this is necessary, we automatically set the soname for ELF and MachO objects during the audit anyway (and this format of the soname is usually wrong for macOS and Windows)

stevengj: removed

C/ChibiScheme/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+
+make \
+    PREFIX="${prefix}" \
+    CC="${CC_BUILD}" \
+    AR="${AR_BUILD}" \
+    LD="${LD_BUILD}" \
+    CFLAGS="${CFLAGS}" \
+    CPPFLAGS="${CPPFLAGS}" \
+    LDFLAGS="${LDFLAGS}" \
+    RANLIB="${RANLIB_BUILD}" \
```
giordano, line 26: Note that setting `X="${X_BUILD}"` has the effect of basically building for the host platform.  In fact audit of `x86_64-linux-gnu` (platform where our CI runs) fails to dlopen the library because of missing `libc.musl-x86_64.so.1` and audit on all platforms points out that the binary objects are for the wrong platform.

Side note: we don't set `CFLAGS`, `CPPFLAGS` or `LDFLAGS` (we force the relevant flags inside the compiler wrappers), so there is no need to forward them here, if that was the intent.

stevengj: Oh, I see.  (I think you mean that it builds for the *build* platform.  We would *like* to compile for the "host" platform in the usual cross-compilation parlance, if I understand correctly.)

giordano:
Progress!  At Linux is working now, the other failures sounds like they're related to the fact the build system believes we're always targeting Linux?

stevengj:
Yes.

Some of the problems should be solved by just not building image files, which are an optional optimization—it's not clear if it will be practical to cross-compile them in the current setup.  Have to figure out how to omit this from the build.

Another problem is that the `chibi-ffi` script needs to be tweaked for cross-compiling; I'm currently looking into this.

imciner2:

PR #4912, opened 2022-05-14
add ImageStreamIO

mileslucas:
Note on supported platforms:
I may be underspecifying the platforms. Big constraints are CUDA and using `mman`, so that restricts to Linux, but the underlying software only compiles to x86, so it only seems to work on the one platform.

Please let me know if I'm missing any files, I've never had to open one of these manually before!

### DIFF
--- a/I/ImageStreamIO/build_tarballs.jl
+++ b/I/ImageStreamIO/build_tarballs.jl
@@ -1,0 +1,38 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "ImageStreamIO"
+version = v"1.0.3-beta"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/milk-org/ImageStreamIO.git", "c4b3c531f4653f3b570345f43b2686cd25af0db8")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ImageStreamIO
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [Platform("x86_64", "linux")]
+
+# The products that we will ensure are always built
+products = Product[
+    LibraryProduct("libImageStreamIO", :libimagestreamio)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CMake_jll", uuid="3f4e10e2-61f2-5801-8945-23b9d642d0e6"))
+    Dependency(PackageSpec(name="CUDA_jll", uuid="e9e359dc-d701-5aa8-82ae-09bbf812ea83"))
+    Dependency(PackageSpec(name="CFITSIO_jll", uuid="b3e40c51-02ae-5482-8a39-3ace5868dcf4"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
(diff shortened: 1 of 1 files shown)


I/ImageStreamIO/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+
+# The products that we will ensure are always built
+products = Product[
+    LibraryProduct("libImageStreamIO", :libimagestreamio)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CMake_jll", uuid="3f4e10e2-61f2-5801-8945-23b9d642d0e6"))
+    Dependency(PackageSpec(name="CUDA_jll", uuid="e9e359dc-d701-5aa8-82ae-09bbf812ea83"))
```
giordano, line 33: We don't really have a well defined way to deal with CUDA as a dependency, but I think the standard approach is to use `CUDA_full_jll` as a BuildDependency _and_ not dlopening the library automatically (otherwise on systems without CUDA you'd get a :boom:), see for example https://github.com/JuliaPackaging/Yggdrasil/blob/f4afdd8e979ec1b5ba43aaf1872d3e33f0622940/S/SCS_GPU/build_tarballs.jl#L31-L45  This also means that you have to check manually what to do with your library.  In case you're wondering, yes, CUDA is a massive complication.

I/ImageStreamIO/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+    Dependency(PackageSpec(name="CFITSIO_jll", uuid="b3e40c51-02ae-5482-8a39-3ace5868dcf4"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", allow_unsafe_flags=true)
```
giordano, line 38:
```julia
allow_unsafe_flags=true
```
Sigh.

I/ImageStreamIO/build_tarballs.jl
mileslucas: Okay, I think I've got it figured out! I don't plan on directly calling CUDA, that should be handled within the C code that is being wrapped here.

I guess I'm not 100% clear on the distinction between Dependency and BuildDependency- I only plan on using the FFI with the code in the library being wrapped, none of its dependencies, so I think that means they should ALL be BuildDependencies?

I/ImageStreamIO/build_tarballs.jl
mileslucas: Sorry 😬 I think the speed matters for them to reach their ~10 kHz streaming rate.

I/ImageStreamIO/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+platforms = [Platform("x86_64", "linux")]
+
+# The products that we will ensure are always built
+products = Product[
+    LibraryProduct("libImageStreamIO", :libimagestreamio)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency(PackageSpec(name="CMake_jll", uuid="3f4e10e2-61f2-5801-8945-23b9d642d0e6"))
```
giordano, line 32: Why CMake here?  We have CMake already in the build environment.  What you're installing here is probably not what you think it is.

I/ImageStreamIO/build_tarballs.jl
giordano: > I guess I'm not 100% clear on the distinction between Dependency and BuildDependency

It's perhaps best explained in the docstring of [`AbstractDependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.AbstractDependency): a `BuildDependency` won't appear in the dependencies of the generated JLL package.  But this means that if your library links to libcuda, then dlopening the library automatically will fail, because in the general case there will be no libcuda at all.

I/ImageStreamIO/build_tarballs.jl
mileslucas: okay, good to know!

mileslucas:
Decided this is something I'll just work with locally considering the very niche use and complicated build dependencies.

PR #4913, opened 2022-05-14
Add likwid 5.2.1

carstenbauer:
This is my attempt to provide a basic configuration of [LIKWID](https://github.com/RRZE-HPC/likwid), in particular the library `liblikwid`, via BinaryBuilder (https://github.com/JuliaPerf/LIKWID.jl/issues/24). The plan is to later use this as a fallback in [LIKWID.jl](https://github.com/JuliaPerf/LIKWID.jl) (if there is no system-wide likwid available).

To modify the `config.mk`, I've opted for three `sed` commands. (Should I use a patch instead? What would be the advantage / disadvantage?)

**Potential future TODOs:**
* provide the executable lua scripts (e.g. `likwid-topology`, `likwid-perfctr`, etc.) as `FileProducts`
* support for musl (if possible)

**Question / Remark:**
I get the following `make` log:
```
sandbox:${WORKSPACE}/srcdir/likwid-5.2.1 # make
ldd: --version: No such file or directory
sh: 4: unknown operand
Info: Compiling for perf_event interface. Measurements of thermal information is disabled
===>  GENERATE HEADER GCC/perfmon_a15_events.h
===>  GENERATE HEADER GCC/perfmon_a57_events.h
===>  GENERATE HEADER GCC/perfmon_a64fx_events.h
[ ... many more GENERATE HEADERs ... ]
===>  GENERATE HEADER GCC/perfmon_zen2_events.h
===>  GENERATE HEADER GCC/perfmon_zen3_events.h
===>  GENERATE HEADER GCC/perfmon_zen_events.h
===>  COMPILE  GCC/access.o
===>  COMPILE  GCC/access_client.o
===>  COMPILE  GCC/access_x86.o
[ ... many more COMPILEs ...]
===>  COMPILE  GCC/tree.o
===>  COMPILE  GCC/voltage.o
===>  COMPILE  GCC/loadData.o
===>  ENTER  /workspace/srcdir/likwid-5.2.1/ext/hwloc
ldd: --version: No such file or directory
sh: 4: unknown operand
Info: Compiling for perf_event interface. Measurements of thermal information is disabled
===>  ENTER  /workspace/srcdir/likwid-5.2.1/ext/lua
ldd: --version: No such file or directory
sh: 4: unknown operand
Info: Compiling for perf_event interface. Measurements of thermal information is disabled
===>  CREATE SHARED LIB  liblikwid.so
===>  CREATE LIB  liblikwidpin.so
make[1]: Entering directory '/workspace/srcdir/likwid-5.2.1/src/pthread-overload'
ldd: --version: No such file or directory
sh: 4: unknown operand
Info: Compiling for perf_event interface. Measurements of thermal information is disabled
make[1]: Leaving directory '/workspace/srcdir/likwid-5.2.1/src/pthread-overload'
===>  ADJUSTING  likwid-perfctr
===>  ADJUSTING  likwid-pin
===>  ADJUSTING  likwid-powermeter
===>  ADJUSTING  likwid-topology
===>  ADJUSTING  likwid-memsweeper
===>  ADJUSTING  likwid-mpirun
===>  ADJUSTING  likwid-features
===>  ADJUSTING  likwid-perfscope
===>  ADJUSTING  likwid-genTopoCfg
===>  ADJUSTING  likwid.lua
===>  ENTER  bench
ldd: --version: No such file or directory
sh: 4: unknown operand
Info: Compiling for perf_event interface. Measurements of thermal information is disabled
===>  COMPILE C GCC/allocator.o
===>  COMPILE C GCC/barrier.o
===>  COMPILE C GCC/bench.o
===>  COMPILE C GCC/bstrlib.o
===>  COMPILE C GCC/bstrlib_helper.o
===>  COMPILE C GCC/ptt2asm.o
===>  COMPILE C GCC/strUtil.o
===>  COMPILE C GCC/threads.o
===>  GENERATE BENCHMARKS
===>  ASSEMBLE  GCC/clcopy.o
===>  ASSEMBLE  GCC/clload.o
[ ... many more ASSEMBLEs ...]
===>  ASSEMBLE  GCC/update_sp_sse.o
===>  ASSEMBLE  GCC/update_sse.o
===>  LINKING  likwid-bench
```

Note the
```
ldd: --version: No such file or directory
sh: 4: unknown operand
Info: Compiling for perf_event interface. Measurements of thermal information is disabled
```
bits. Should I care about those given that the build process seems to proceed just fine?

(@TomTheBear, @Suavesito-Olimpiada FYI)

### DIFF
--- a/L/likwid/build_tarballs.jl
+++ b/L/likwid/build_tarballs.jl
@@ -1,0 +1,44 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "likwid"
+version = v"5.2.1"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("http://ftp.fau.de/pub/likwid/likwid-$(version).tar.gz", "1b8e668da117f24302a344596336eca2c69d2bc2f49fa228ca41ea0688f6cbc2")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd likwid-5.2.1/
+sed -i 's/PREFIX ?= .*/PREFIX ?= \/workspace\/destdir/' config.mk
+sed -i 's/ACCESSMODE = .*/ACCESSMODE = perf_event/' config.mk
+sed -i 's/BUILDFREQ = .*/BUILDFREQ = false/' config.mk
+make
+make install
+exit
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc")
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("liblikwid", :liblikwid),
+    ExecutableProduct("likwid-lua", :likwid_lua),
+    ExecutableProduct("likwid-bench", :likwid_bench)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")


vchuravy:
Patch tends to be more stable since it will fail if it couldn't apply it.

Ideally you would be reusing HWLoc_jll instead of building it again, same with lua

TomTheBear:
@carstenbauer: The `ldd --version` call is used to determine the glibc version because old glibc version lack the wrappers for `sched_setaffinity`. You can just add `-DHAS_SCHEDAFFINITY` to the `DEFINES` to forcefully use the wrapper (which should be available on all systems nowadays).

I never tried building LIKWID with musl.

For building with external hwloc and lua, check [spack's LIKWID package](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/likwid/package.py#L163-L191).

carstenbauer:
> Ideally you would be reusing HWLoc_jll instead of building it again, same with lua
 Yes, indeed, just forgot to put it on the Todo list.

carstenbauer:
> You can just add `-DHAS_SCHEDAFFINITY` to the `DEFINES` to forcefully use the wrapper (which should be available on all systems nowadays).

@TomTheBear Can you elaborate a bit more, i.e. where should I add it exactly? I tried setting it in `make/config_defines.mk` (and `make/include_GCC.mk` / `make/include_GCCX86.mk`) without success.

L/likwid/bundled/patches/useperfevents.patch
```diff
@@ -0,0 +1,28 @@
+diff --git before/likwid-5.2.1/config.mk after/likwid-5.2.1/config.mk
+index aa623fd..0d6e395 100644
+--- before/likwid-5.2.1/config.mk
++++ after/likwid-5.2.1/config.mk
+@@ -11,12 +11,12 @@
+ COMPILER = GCC#NO SPACE
+ 
+ # Path were to install likwid
+-PREFIX ?= /usr/local#NO SPACE
```
giordano, line 9: There shouldn't be a need to patch anything, you can set the variables on the command line when invoking make:
```
make PREFIX=${prefix} ...
```
Also, don't hardcode absolute paths, use the relevant [environment variables](https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables)

TomTheBear:
> Can you elaborate a bit more, i.e. where should I add it exactly? I tried setting it in make/config_defines.mk (and make/include_GCC.mk / make/include_GCCX86.mk) without success.

If you add it to the `DEFINES` in `make/include_<COMPILER>.mk`, it is present on the compile lines. You can check with `make Q=`:
```
$ make Q=
===>  COMPILE  GCC/perfmon.o
gcc [...] -DHAS_SCHEDAFFINITY [...]
```

carstenbauer:
@TomTheBear That's what I did (and it works in the sense that you describe) but I still get
```
sandbox:${WORKSPACE}/srcdir/likwid-5.2.1 # make Q=
ldd: --version: No such file or directory
sh: 4: unknown operand
Info: Compiling for perf_event interface. Measurements of thermal information is disabled
===>  GENERATE HEADER GCC/perfmon_a15_events.h
./perl/gen_events.pl src/includes/perfmon_a15_events.txt GCC/perfmon_a15_events.h
===>  GENERATE HEADER GCC/perfmon_a57_events.h
./perl/gen_events.pl src/includes/perfmon_a57_events.txt GCC/perfmon_a57_events.h
===>  GENERATE HEADER GCC/perfmon_a64fx_events.h
./perl/gen_events.pl src/includes/perfmon_a64fx_events.txt GCC/perfmon_a64fx_events.h
===>  GENERATE HEADER GCC/perfmon_atom_events.h
./perl/gen_events.pl src/includes/perfmon_atom_events.txt GCC/perfmon_atom_events.h
===>  GENERATE HEADER GCC/perfmon_broadwellEP_events.h
./perl/gen_events.pl src/includes/perfmon_broadwellEP_events.txt GCC/perfmon_broadwellEP_events.h
===>  GENERATE HEADER GCC/perfmon_broadwell_events.h
./perl/gen_events.pl src/includes/perfmon_broadwell_events.txt GCC/perfmon_broadwell_events.h
===>  GENERATE HEADER GCC/perfmon_broadwelld_events.h
./perl/gen_events.pl src/includes/perfmon_broadwelld_events.txt GCC/perfmon_broadwelld_events.h
===>  GENERATE HEADER GCC/perfmon_cascadelakeX_events.h
./perl/gen_events.pl src/includes/perfmon_cascadelakeX_events.txt GCC/perfmon_cascadelakeX_events.h
===>  GENERATE HEADER GCC/perfmon_cavtx2_events.h
./perl/gen_events.pl src/includes/perfmon_cavtx2_events.txt GCC/perfmon_cavtx2_events.h
===>  GENERATE HEADER GCC/perfmon_core2_events.h
./perl/gen_events.pl src/includes/perfmon_core2_events.txt GCC/perfmon_core2_events.h
===>  GENERATE HEADER GCC/perfmon_goldmont_events.h
./perl/gen_events.pl src/includes/perfmon_goldmont_events.txt GCC/perfmon_goldmont_events.h
===>  GENERATE HEADER GCC/perfmon_haswellEP_events.h
./perl/gen_events.pl src/includes/perfmon_haswellEP_events.txt GCC/perfmon_haswellEP_events.h
===>  GENERATE HEADER GCC/perfmon_haswell_events.h
./perl/gen_events.pl src/includes/perfmon_haswell_events.txt GCC/perfmon_haswell_events.h
===>  GENERATE HEADER GCC/perfmon_icelakeX_events.h
./perl/gen_events.pl src/includes/perfmon_icelakeX_events.txt GCC/perfmon_icelakeX_events.h
===>  GENERATE HEADER GCC/perfmon_icelake_events.h
./perl/gen_events.pl src/includes/perfmon_icelake_events.txt GCC/perfmon_icelake_events.h
===>  GENERATE HEADER GCC/perfmon_interlagos_events.h
./perl/gen_events.pl src/includes/perfmon_interlagos_events.txt GCC/perfmon_interlagos_events.h
===>  GENERATE HEADER GCC/perfmon_ivybridgeEP_events.h
./perl/gen_events.pl src/includes/perfmon_ivybridgeEP_events.txt GCC/perfmon_ivybridgeEP_events.h
===>  GENERATE HEADER GCC/perfmon_ivybridge_events.h
===>  GENERATE HEADER GCC/perfmon_k10_events.h
./perl/gen_events.pl src/includes/perfmon_k10_events.txt GCC/perfmon_k10_events.h
===>  GENERATE HEADER GCC/perfmon_k8_events.h
./perl/gen_events.pl src/includes/perfmon_k8_events.txt GCC/perfmon_k8_events.h
===>  GENERATE HEADER GCC/perfmon_kabini_events.h
./perl/gen_events.pl src/includes/perfmon_kabini_events.txt GCC/perfmon_kabini_events.h
===>  GENERATE HEADER GCC/perfmon_knl_events.h
./perl/gen_events.pl src/includes/perfmon_knl_events.txt GCC/perfmon_knl_events.h
===>  GENERATE HEADER GCC/perfmon_nehalemEX_events.h
./perl/gen_events.pl src/includes/perfmon_nehalemEX_events.txt GCC/perfmon_nehalemEX_events.h
===>  GENERATE HEADER GCC/perfmon_nehalem_events.h
./perl/gen_events.pl src/includes/perfmon_nehalem_events.txt GCC/perfmon_nehalem_events.h
===>  GENERATE HEADER GCC/perfmon_neon1_events.h
./perl/gen_events.pl src/includes/perfmon_neon1_events.txt GCC/perfmon_neon1_events.h
===>  GENERATE HEADER GCC/perfmon_p6_events.h
./perl/gen_events.pl src/includes/perfmon_p6_events.txt GCC/perfmon_p6_events.h
===>  GENERATE HEADER GCC/perfmon_phi_events.h
./perl/gen_events.pl src/includes/perfmon_phi_events.txt GCC/perfmon_phi_events.h
===>  GENERATE HEADER GCC/perfmon_pm_events.h
./perl/gen_events.pl src/includes/perfmon_pm_events.txt GCC/perfmon_pm_events.h
===>  GENERATE HEADER GCC/perfmon_power8_events.h
./perl/gen_events.pl src/includes/perfmon_power8_events.txt GCC/perfmon_power8_events.h
===>  GENERATE HEADER GCC/perfmon_power9_events.h
./perl/gen_events.pl src/includes/perfmon_power9_events.txt GCC/perfmon_power9_events.h
===>  GENERATE HEADER GCC/perfmon_sandybridgeEP_events.h
./perl/gen_events.pl src/includes/perfmon_sandybridgeEP_events.txt GCC/perfmon_sandybridgeEP_events.h
===>  GENERATE HEADER GCC/perfmon_sandybridge_events.h
./perl/gen_events.pl src/includes/perfmon_sandybridge_events.txt GCC/perfmon_sandybridge_events.h
===>  GENERATE HEADER GCC/perfmon_silvermont_events.h
./perl/gen_events.pl src/includes/perfmon_silvermont_events.txt GCC/perfmon_silvermont_events.h
===>  GENERATE HEADER GCC/perfmon_skylakeX_events.h
./perl/gen_events.pl src/includes/perfmon_skylakeX_events.txt GCC/perfmon_skylakeX_events.h
===>  GENERATE HEADER GCC/perfmon_skylake_events.h
./perl/gen_events.pl src/includes/perfmon_skylake_events.txt GCC/perfmon_skylake_events.h
===>  GENERATE HEADER GCC/perfmon_tigerlake_events.h
./perl/gen_events.pl src/includes/perfmon_tigerlake_events.txt GCC/perfmon_tigerlake_events.h
===>  GENERATE HEADER GCC/perfmon_westmereEX_events.h
./perl/gen_events.pl src/includes/perfmon_westmereEX_events.txt GCC/perfmon_westmereEX_events.h
===>  GENERATE HEADER GCC/perfmon_westmere_events.h
./perl/gen_events.pl src/includes/perfmon_westmere_events.txt GCC/perfmon_westmere_events.h
===>  GENERATE HEADER GCC/perfmon_zen2_events.h
./perl/gen_events.pl src/includes/perfmon_zen2_events.txt GCC/perfmon_zen2_events.h
===>  GENERATE HEADER GCC/perfmon_zen3_events.h
./perl/gen_events.pl src/includes/perfmon_zen3_events.txt GCC/perfmon_zen3_events.h
===>  GENERATE HEADER GCC/perfmon_zen_events.h
./perl/gen_events.pl src/includes/perfmon_zen_events.txt GCC/perfmon_zen_events.h
===>  COMPILE  GCC/access.o
gcc -c  -O2 -std=c99 -Wno-format -fPIC -fPIC -fvisibility=hidden   -DPAGE_ALIGNMENT=4096 -DLIKWID_MONITOR_LOCK -DDEBUGLEV=0 -DHAS_SCHEDAFFINITY -DVERSION=5 -DRELEASE=2 -DMINORVERSION=1 -DCFGFILE=/etc/likwid.cfg -DTOPOFILE=/etc/likwid_topo.cfg -DINSTALL_PREFIX=/workspace/destdir -DMAX_NUM_THREADS=300 -DMAX_NUM_NODES=64 -DMAX_NUM_CLIARGS=16384 -DACCESSDAEMON=/workspace/destdir/sbin/likwid-accessD -DFREQDAEMON=/workspace/destdir/sbin/likwid-setFreq -DGROUPPATH=/workspace/destdir/share/likwid/perfgroups -DLIKWIDLOCK=/var/run/likwid.lock -DLIKWIDSOCKETBASE=/tmp/likwid   -DGITCOMMIT=233ab943543480cd46058b34616c174198ba0459 -D_GNU_SOURCE -DCOLOR=BLUE -DLIKWID_USE_PERFEVENT -DHAS_MEMPOLICY -DHAS_SCHEDAFFINITY -DLIKWID_USE_HWLOC -DLIKWID_USE_PERFEVENT -DACCESSMODE=-1 -I./src/includes -I/workspace/srcdir/likwid-5.2.1/ext/lua/includes -I/workspace/srcdir/likwid-5.2.1/ext/hwloc/include -I./GCC ./src/access.c -o GCC/access.o
[....]
```
Do I need to put it somewhere else as well?

giordano:
> The ldd --version call is used to determine the glibc version because old glibc version lack the wrappers for sched_setaffinity.

```console
sandbox:${WORKSPACE} # echo '#include <features.h>' | cc -dM -E - | grep -E '#define __GLIBC_(|MINOR_)_'
#define __GLIBC__ 2
#define __GLIBC_MINOR__ 12
```
is probably more cross-compilation friendly than running `ldd` on a system where `ldd` is ***not*** from glibc.

Related to the topic of cross-compilation, we looked into building likwid some time ago, but I remember it wasn't really possible to do a cross-build because the package needed some information about the target system at compile time, or some local kernel modules, or something like that.  Is this still the case?

carstenbauer:
> For building with external hwloc and lua, check [spack's LIKWID package](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/likwid/package.py?rgh-link-date=2022-05-15T00%3A08%3A44Z#L163-L191).

It seems like likwid still installs `libhwloc.so.5.2` (and makes the `libhwloc.so` symlink point to it) when I specify an external hwloc. This is the `$libdir` before and after `make install [...]`:

**Before**
```
lrwxrwxrwx    1 root     root            18 May 15 17:31 libhwloc.so -> libhwloc.so.15.5.3*
lrwxrwxrwx    1 root     root            18 May 15 17:31 libhwloc.so.15 -> libhwloc.so.15.5.3*
lrwxrwxrwx    1 root     root            79 May 15 17:31 libhwloc.so.15.5.3 -> ../../artifacts/f11a525a0a63e8dac0e0236c0e578c0d8e34d59e/l
ib/libhwloc.so.15.5.3*
```

**After**
```
lrwxrwxrwx    1 root     root            15 May 15 17:52 libhwloc.so -> libhwloc.so.5.2*
lrwxrwxrwx    1 root     root            18 May 15 17:31 libhwloc.so.15 -> libhwloc.so.15.5.3*
lrwxrwxrwx    1 root     root            79 May 15 17:31 libhwloc.so.15.5.3 -> ../../artifacts/f11a525a0a63e8dac0e0236c0e578c0d8e34d59e/l
ib/libhwloc.so.15.5.3*
lrwxrwxrwx    1 root     root            15 May 15 17:52 libhwloc.so.5 -> libhwloc.so.5.2*
-rwxr-xr-x    1 root     root       1729301 May 15 17:52 libhwloc.so.5.2*
``` 

I see that they apply [a patch](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/l/likwid/likwid-5.2.0_use_EB_hwloc.patch) in the easyconfig that addresses this. @TomTheBear Why isn't the easyconfig patch the default? I.e., is there a deeper reason for the creation of `libhwloc.so.5.2`? Otherwise I'd apply a similar patch.

L/likwid/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+    DirectorySource("./bundled")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+cd likwid-5.2.1/
```
giordano, line 20: This is more robust against updates (avoid hardcoding version numbers everywhere)
```suggestion
cd $WORKSPACE/srcdir/likwid*
atomic_patch -p2 ../patches/perfevents_hasschedaffinity_exthwlocfix.patch
```

giordano:
The script `bin/likwid-topology` hardcodes local paths of the build environment, which won't exist at runtime.  That's not going to work.

L/likwid/build_tarballs.jl
carstenbauer: Just to be safe, you did change `-p1` to `-p2`. Was that intentional? (I have no idea what this does 😄  This patch stuff has been autogenerated by the wizard, as you probably know.)

giordano: That's the number of directory levels to strip from the patch.  The patch references files like `before/likwid-5.2.1/Makefile`, I _**do**_ want to strip also `likwid-5.1.2` because (1) now the patch is applied from within the likwid source directory (2) I don't want to deal with hardcoded version number in the patch

carstenbauer:
> The script `bin/likwid-topology`(and probably also the others) hardcodes local absolute paths of the build environment, which won't exist at runtime. That's not going to work.

Yeah. While certainly unfortunate, the `liblikwid.so` is the primary reason for me to attempt to create a JLL. You can use lots (most?) of the features without the command-line tools. (One could argue that the "in-Julia" interface to LIKWID is anyways nicer than the CLI.)

giordano:
>  support for musl (if possible)

It'd be nice to build for all Linux platforms, if possible (at least glibc ones, if that's a requirement).  I'm thinking in particular of aarch64 and powerpc64le, which are more popular in HPC settings.

TomTheBear:
> Why isn't the easyconfig patch the default? I.e., is there a deeper reason for the creation of libhwloc.so.5.2?

@carstenbauer: I test LIKWID only in the default configuration varying important options (`COMPILER` and `ACCESSMODE`). The feature to use external lua and hwloc was only added for the package management folks out there. The reason is quite simple, there is not much feedback from the package management folks. I didn't even know that an easy_build recipe exists for LIKWID. And consequently, I didn't know the patch. The library is probably not needed as the easy_build guys (hopefully) tested their stuff.

TomTheBear:
> The script bin/likwid-topology(and probably also the others) hardcodes local absolute paths of the build environment, which won't exist at runtime. That's not going to work.

@giordano This can be avoided by getting the script path in the beginning. It's just Lua. Patch it if you need it differently. Until now, Yggdrasil seems to be the only package manager having problems with absolute paths.

Suavesito-Olimpiada:
> Until now, Yggdrasil seems to be the only package manager having problems with absolute paths.

This is the only sensible thing given that Yggdrasil

 -  does not have administrator permissions on systems,
 -  is cross-compiled on a Linux system,
 -  want to support multiple distribution and versions of different OS's,
 -  and must be [relocatable](https://julialang.github.io/PackageCompiler.jl/dev/).

The last point might be the most important of all.

carstenbauer:
> This is the only sensible thing given that Yggdrasil [...]

While true, that's a Julia / Yggdrasil perspective. A tool (like likwid) may not care about what Yggdrasil wants / needs. It's a tedious task to cater to the specific needs of various package managers so we can't expect non-Julia developers to spend much time on it. We can hope though 😉

Fortunately, in this case, @TomTheBear has already committed a few simple patches to likwid so that from version 5.2.2 on
* the `ldd --version` issue should be gone (https://github.com/RRZE-HPC/likwid/commit/d6bbdcf2a84210d9c1dd6952d1ae533ae36bff07) and 
* we don't need to patch the Makefile any more (i.e. we can drop the easyconfig-like patch) (https://github.com/RRZE-HPC/likwid/commit/3a79b059c57f03949a74977b654a416b5ed51fcf)

Thanks for this! ❤️

Suavesito-Olimpiada:
> A tool (like likwid) may not care about what Yggdrasil wants / needs.

Absolutely, I just wanted to explain the perspective of Julia/Yggdrasil, not to rant about the projects that don't do it, or ask for every project to do it. I'm sorry if my comment read otherwise. :sweat_smile:

> Fortunately, in this case, @TomTheBear has already committed a few simple patches to likwid

Thanks, @TomTheBear! That sounds incredible. :tada:

giordano:
> How do I set COMPILER = GCCPOWER for powerpc and COMPILER = GCC for regular x86_64 (i.e. system dependent patches)?

```sh
if [[ "${target}" == powerpc64le-linux-* ]]; then
    COMPILER=GCCPOWER
elif [[ "${target}" == aarch64-linux-* ]]; then
    COMPILER=GCCARMv8
else # assume x86_64
    COMPILER=GCC
fi

**Comment by carstenbauer** [comment]
FYI: I'll wait until the next likwid release before continuing on this.

File: L/likwid/build_tarballs.jl [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+    DirectorySource("./bundled")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+cd likwid-5.2.2/
```
giordano, line 20: Just to be future-proof
```suggestion
cd likwid-*
```

L/likwid/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
...
+elif [[ "${target}" == armv7l-linux-* ]]; then
+    export COMPILER=GCCARMv7
+elif [[ "${target}" == i686-linux-* ]]; then
+    export COMPILER=GCCX86
+else # assume x86_64
+    export COMPILER=GCC
+fi
+make PREFIX=${prefix} HWLOC_INCLUDE_DIR=${includedir} HWLOC_LIB_DIR=${libdir} HWLOC_LIB_NAME=hwloc LUA_INCLUDE_DIR=${includedir} LUA_LIB_DIR=${libdir} LUA_LIB_NAME=lua LUA_BIN=${bindir}
+make install PREFIX=${prefix} HWLOC_INCLUDE_DIR=${includedir} HWLOC_LIB_DIR=${libdir} HWLOC_LIB_NAME=hwloc LUA_INCLUDE_DIR=${includedir} LUA_LIB_DIR=${libdir} LUA_LIB_NAME=lua LUA_BIN=${bindir}
+exit
```
giordano, line 34:
```suggestion
```

TomTheBear:
I'm in the progress of making LIKWID relocatable but it's not finished yet: https://github.com/RRZE-HPC/likwid/pull/480

carstenbauer:
> Is this good to go now?

No it's not (for the reasons you mentioned). Just wanted to push it a bit closer to the finish line.

miguelraz:
This would be an awesome feature to have. Looking forward to the progress that @TomTheBear can make on this.

TomTheBear:
This is not an easy task. LIKWID consists of multiple files (misc Lua functions, C library, performance group files, benchmark files, accessdaemon, ...) that need to be found at runtime and interact with each other (e.g. UNIX socket communication).

fingolfin:
@carstenbauer is this still something you are interested in completing? Or can this PR be closed?

carstenbauer:
Unfortunately, I don't have time to finish this any time soon. I think it's worth having but someone else would likely need to take over. Otherwise it can be closed.

PR #4916, opened 2022-05-16
Add GPI-2 library for the development of scalable, parallel applications

sloede:
GPI-2 is an API for the development of scalable, asynchronous and fault tolerant parallel applications.

Source: http://gpi-site.com/ (yes, `http` 😦)
Source: https://github.com/cc-hpc-itwm/GPI-2

This adds a `build_tarballs.jl` for GPI-2 that only uses Ethernet for convenience. If you want to do actual production runs, you want to use a system-local installation of GPI-2 that exploits the local Infiniband drivers.

### DIFF
--- a/G/GPI2/build_tarballs.jl
+++ b/G/GPI2/build_tarballs.jl
@@ -35,6 +35,8 @@ products = [
     LibraryProduct("libGPI2-dbg", :libGPI2_dbg),
     LibraryProduct("libGPI2", :libGPI2),
     ExecutableProduct("gaspi_logger", :gaspi_logger),
+    # The following are shell scripts that are required for GPI-2 to work,
+    # thus they are added as `FileProduct`s
     FileProduct("bin/gaspi_run", :gaspi_run),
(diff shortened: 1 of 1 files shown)


sloede:
@giordano just FYI, I thought this might be of interest to you as well. I am currently working on providing a GPI2.jl package as well.

vchuravy:
We have libibverbs in rdma_core_jll

sloede:
Hm... but what happens on a machine without Infiniband? Will that still work if base this on `rdma_core_jll`? Theoretically speaking, of course, I will surely try to check if I can make it work.

sloede:
I rechecked: Supporting both Infiniband and Ethernet simultaneously does not work for GPI-2:
```
./configure --with-ethernet --with-infiniband
[...]
-----------------------------
   Checking for device(s):
-----------------------------
configure: error: Concurrently Infiniband and Ethernet is not supported
```
Since Ethernet is much more widely used, and this is intended for the average and not the power user (who will have their own Infiniband-enabled local install of GPI-2), I suggest to keep it as it is.

PR #4917, opened 2022-05-17
[cmark] Add recipes for cmark and cmark-gfm

mortenpi:
Adds recipes for the [cmark CommonMark parser](https://github.com/commonmark/cmark) and its GitHub-flavored cousin [cmark-gfm](https://github.com/github/cmark-gfm).

cmark-gfm is basically cmark with a couple of small additions (although it appears to be slightly behind with versions), so I was initially thinking of putting under the same directory. But in the end it was easier to organize it this way for now, but happy to change that if need be.


giordano:
Brilliant, two straightforward recipes with no warnings of any sort!

PR #4918, opened 2022-05-18
Bump rr

Keno:
In particular to pick up:
- https://github.com/rr-debugger/rr/pull/3227
- https://github.com/rr-debugger/rr/pull/3148

Which fix issues we've been seeing on base Julia CI.


giordano:
I guess zlib is now required because of https://github.com/rr-debugger/rr/pull/3126

Keno:
That's the old libc headers on Yggdrasil. I'll put in a compat commit for rr.

PR #4919, opened 2022-05-18
updates to wavefd

deckerla:


giordano:
Thanks for your contribution!  I pushed some changes to use LLVM OpenMP on BSD systems, which is a more natural choice than using libgomp from GCC.

PR #4922, opened 2022-05-19
Build HMMER-3.1.b2 as v3.1.0

pagnani:
This PR is against HMMER-3.1.b2. 

As discussed on a related [Discourse issue](https://discourse.julialang.org/t/building-specific-library-version-with-yggdrasil/81262), we need this version as it is the one against which [Pfam](http://pfam.xfam.org/) last release (35.0) is built.

The jll version associated will be 3.1.0 (as 3.1.b2 is not semver compliant)


giordano:
CC: @marcom

giordano:
Thanks a lot!  For the record, if you strongly depend on a specific version of `HMMER_jll`, I recommend putting that information in the compat section of your environment/package, for example with
```toml
HMMER_jll = "~3.1"
```
which accepts only versions in the series 3.1.x, but not, for example, 3.3.x.

PR #4923, opened 2022-05-19
[GMP] Add a GMP patch for  CVE-2021-43618

ViralBShah:
https://nvd.nist.gov/vuln/detail/CVE-2021-43618
https://lists.debian.org/debian-lts-announce/2021/12/msg00001.html


G/GMP/GMP@6.2.1/patches/CVE-2021-43618.patch
```diff
@@ -0,0 +1,24 @@
+# Origin: https://gmplib.org/repo/gmp-6.2/rev/561a9c25298e
```
giordano, line 1: Note that the patch should go in the bundled directory. This isn't being picked up actually

ViralBShah: Ah forgot that.

ViralBShah:
I suppose it is ok to merge.

PR #4933, opened 2022-05-23
add ViZDoom

jbrea:


V/ViZDoom/build_tarballs.jl
```diff
@@ -0,0 +1,68 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libvizdoomjl", :libvizdoom),
+    FileProduct("scenarios", :scenarios)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("boost_jll"),
```
giordano, line 59: You must specify the compat bound for Boost: #3024

V/ViZDoom/build_tarballs.jl
```diff
@@ -0,0 +1,68 @@
...
+    -DCMAKE_FIND_ROOT_PATH=${prefix} \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_CROSSCOMPILING=FALSE \
+    -DJulia_PREFIX=${prefix} \
+    -DBUILD_JULIA=ON \
+    ..
+cmake --build . --parallel ${nproc}
+cp -r bin/* ${libdir}
+mv ../src/freedoom2.wad ${libdir}
```
giordano, line 38: This sounds wrong.  The content of the `lib/` directory is
```console
% tar tzvf x86_64-linux-gnu-cxx11-julia_version+1.6.3.tar.gz | grep lib/
-rwxr-xr-x 0/0           19920 1970-01-01 01:00 lib/arithchk
-rw-r--r-- 0/0        28544132 1970-01-01 01:00 lib/freedoom2.wad
-rwxr-xr-x 0/0           76024 1970-01-01 01:00 lib/lemon
-rw-r--r-- 0/0           25628 1970-01-01 01:00 lib/lempar.c
-rw-r--r-- 0/0         1416786 1970-01-01 01:00 lib/libvizdoom.a
-rwxr-xr-x 0/0          717640 1970-01-01 01:00 lib/libvizdoom.so
-rwxr-xr-x 0/0         1696184 1970-01-01 01:00 lib/libvizdoomjl.so
-rwxr-xr-x 0/0           19752 1970-01-01 01:00 lib/qnan
-rwxr-xr-x 0/0          216888 1970-01-01 01:00 lib/re2c
-rwxr-xr-x 0/0           20368 1970-01-01 01:00 lib/updaterevision
-rwxr-xr-x 0/0         5506216 1970-01-01 01:00 lib/vizdoom
-rw-r--r-- 0/0          630827 1970-01-01 01:00 lib/vizdoom.pk3
-rwxr-xr-x 0/0          144904 1970-01-01 01:00 lib/zipdir
```
There is a mix of shared libraries (only things which should be here), executable binaries (which should go to `${bindir}`), source files (which shouldn't be in the tarball at all), files I have no idea what they are (`*.wad` and `*.pk3`) and static library (it's ok to have the static library in `${libdir}`, but that's just useless for the purpose of calling a library from Julia, so we usually avoid having static libraries around).

V/ViZDoom/bundled/ViZDoomJuliaModule.cpp
```diff
@@ -0,0 +1,488 @@
...
+        });
+
+    mod.method("doom_tics_to_ms", doomTicsToMs);
+    mod.method("ms_to_doom_tics", msToDoomTics);
+    mod.method("doom_tics_to_sec", doomTicsToSec);
+    mod.method("sec_to_doom_tics", secToDoomTics);
+    mod.method("doom_fixed_to_double", [](double x) { return doomFixedToDouble(x); });
+    mod.method("is_binary_button", isBinaryButton);
+    mod.method("is_delta_button", isDeltaButton);
+}
```
giordano, line 488: Missing newline:
```suggestion
}

```

giordano:
The build system assumes you can run executables built for the target platform, which we can't.  This is very cross-compilation-unfriendly.

V/ViZDoom/build_tarballs.jl
jbrea: Thanks. I removed most files, but vizdoom.pk3 and freedoom2.wad seem to be needed. Otherwise games won't load.

jbrea:
Thanks for the review!

> The build system assumes you can run executables built for the target platform, which we can't. This is very cross-compilation-unfriendly.

Can I fix this somehow? Note that I use the flag `-DCMAKE_CROSSCOMPILING=FALSE`, because I couldn't resolve some cmake errors otherwise. But I don't really understand its implications.

V/ViZDoom/build_tarballs.jl
```diff
@@ -0,0 +1,67 @@
...
+cmake --build . --parallel ${nproc}
+cp bin/libvizdoomjl.so bin/vizdoom bin/vizdoom.pk3 ../src/freedoom2.wad ${libdir}
+mv ../scenarios $prefix/
+install_license $WORKSPACE/srcdir/ViZDoom-*/README.md
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+include("../../L/libjulia/common.jl")
+platforms = [Platform("x86_64", "linux")] #vcat(libjulia_platforms.(julia_versions)...)
```
giordano, line 45: This is wrong, you aren't building for the different versions of Julia anymore.

giordano:
> Note that I use the flag `-DCMAKE_CROSSCOMPILING=FALSE`, because I couldn't resolve some cmake errors otherwise. But I don't really understand its implications.

 Well, we ***are*** cross-compiling, so that's a blatant lie :slightly_smiling_face:

V/ViZDoom/build_tarballs.jl
jbrea: Shoot, that was for local testing :)

giordano: For local testing you never have to modify the list of platforms, you can pass the platform you want to build for as argument to the script: https://docs.binarybuilder.org/v0.5/#Manually-create-or-edit-build_tarballs.jl

jbrea:
Btw. To actually use `ViZDoom_jll` I have the following lines in `ViZDoom.jl`

```julia
module ViZDoom

using CxxWrap
using ViZDoom_jll

@wrapmodule("libvizdoomjl")

function __init__()
    @initcxx
end

include("util.jl")
export get_scenario_path, set_game

include("games/games.jl")
export basic_game

end # module
```
Now, when I actually want to run a game I get
```
~/.julia/artifacts/01f94640f8ea8b00edc310b9f24899fa7811a6ca/lib/vizdoom: error while loading shared libraries: libboost_thread.so.1.76.0: cannot open shared object file: No such file or directory
```
unless I set my `LD_LIBRARY_PATH` to the artifact containing libboost 1.76.0. Is there an alternative to setting the `LD_LIBRARY_PATH` in `ViZDoom.jl`?

I also get occasionally and seemingly at random the error
```julia
C++ exception while wrapping module ViZDoom: invalid subtyping in definition of Label with supertype Any
ERROR: InitError: invalid subtyping in definition of Label with supertype Any
during initialization of module ViZDoom
```
Don't know if this is related.

giordano:
> Now, when I actually want to run a game I get

***How*** are you running the program?

giordano:
Side note: I quickly skimmed through CMake files of ViZDoom and there are some references to cross-compilation.  It sounds like they should support it, but I have no idea how they want to do it.  I also don't really have the time to dig into this, so if you want to target all platforms you'll have to do some investigations on your own.

jbrea:
> _**How**_ are you running the program?

In a Julia REPL,
```julia
julia> using ViZDoom

julia> game = basic_game()

julia> ViZDoom.init(game)
```

giordano:
I wanted to know how you're running the ***executable***.  I have no idea what `basic_game()` and `ViZDoom.init(game)` do.

imciner2:

PR #4937, opened 2022-05-23
Add DebianGlibc@2.33 for aarch64-linux-gnu

Keno:
This builds a replacement glibc for use in debian testing images
that need to be compatible with rr (such as our JuliaCI images).
Due an interaction between libgcc and glibc, current versions of
these dependencies currently, intermittently, result in
rr-incompatible binaries due to an accidental use of legacy atomics
instructions. Building with `-march=armv8.3-a -mno-outline-atomics`
works around this issue at the cost of losing compatibility with pre-8.3
ARMv8. The product of this builder can be applied with `tar -C / xf`
over an existing debian rootfs to work around this issue while we
work with upstream to correct it.

cc @DilumAluthge

### DIFF
--- a/G/Glibc/DebianGlibc@2.33/build_tarballs.jl
+++ b/G/Glibc/DebianGlibc@2.33/build_tarballs.jl
@@ -15,7 +15,7 @@ push!(sources,
     "c4dd741a08918861796cf62a8d3209a56d41d807344a7db481b73a4d17f726bc"),)
 script = glibc_script()
 platforms =  filter(p -> libc(p) == "glibc" && arch(p) == "aarch64",
-    supported_platforms(;experimental=true))
+    supported_platforms())
 products = Product[] # glibc_products()
 dependencies = glibc_dependencies()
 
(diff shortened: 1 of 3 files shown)


G/Glibc/DebianGlibc@2.33/build_tarballs.jl
```diff
@@ -0,0 +1,23 @@
...
+name = "DebianGlibc"
+version = v"2.33"
+
+sources = glibc_sources(version)
+push!(sources,
+    ArchiveSource("http://deb.debian.org/debian/pool/main/g/glibc/glibc_2.33-7.debian.tar.xz",
+    "c4dd741a08918861796cf62a8d3209a56d41d807344a7db481b73a4d17f726bc"),)
+script = glibc_script()
+platforms =  filter(p -> libc(p) == "glibc" && arch(p) == "aarch64",
+    supported_platforms(;experimental=true))
```
giordano, line 18:
```suggestion
    supported_platforms())
```

Keno:
Let's go with this. Hopefully we'll be able to undo it again in a year or so once it's fixed upstream and has been picked up by distributions.

staticfloat:
Is there a place we can track the status of this being upstreamed?

PR #4941, opened 2022-05-24
[xxHash_jll] Fix overwriting of DLL with static linked DLL

### LINKED ISSUE #4945: [xxHash_jll] Functions not exported on Windows builds
libxxhash.dll is built correctly for Windows by MSYS2: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-xxhash/PKGBUILD

I've tried moving `__declspec(dllexport)` outside any if-defs in #4941 , as well as removing the XXH_INLINE_ALL flag that would have disabled that; it "successfully" breaks non-Windows builds, but Windows builds still have no exports as reported by the `dumpbin /EXPORT` tool.

I'm going beyond my knowledge a bit here, not sure where else to go.

Cross-ref with xxHash issue: https://github.com/Cyan4973/xxHash/issues/715

BioTurboNick:
~~Remove XXH_INLINE_ALL flag at recommendation of library author to fix Windows build~~

Remove rogue overwriting of DLL with static-linked library

Fixes #4945


CarAuto34:
Ty

PR #4943, opened 2022-05-24
New recipe: FastANI v1.33.0

BenjaminDoran:


F/FastANI/buildtarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+./configure \
+    --prefix=$prefix \
+    --build=${MACHTYPE} \
+    --host=${target} \
+    --with-gsl=${destdir} \
+    --includedir=${includedir} \
+    --libdir=${libdir}
+    
+if [[ $target == *-apple-* ]]; then
+    MACFLAGS="-mmacosx-version-min=10.7 -stdlib=libc++ -Xpreprocessor -lomp"
```
giordano: I don't understand why this'd be needed.  We use `-mmacosx-version-min=10.10` by default for x86_64, and 11.0 for aarch64, setting 10.7 doesn't make much sense.  Also, setting `-Xpreprocessor -lomp` is likely useless, `-fopenmp` should be enough.

F/FastANI/buildtarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+fi
+CXXFLAGS="-O3 -DNDEBUG -std=c++11 -Isrc -I /usr/local/include -fopenmp ${MACFLAGS}"
```
giordano: Absolutely don't use `/usr/local/include`

giordano (changes requested):

F/FastANI/buildtarballs.jl
BenjaminDoran: noted made edits

F/FastANI/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+./configure \
+    --prefix=$prefix \
+    --build=${MACHTYPE} \
+    --host=${target} \
+    --with-gsl=${destdir} \
+    --includedir=${includedir} \
+    --libdir=${libdir}
+    
+if [[ $target == *-apple-* ]]; then
+    MACFLAGS="-stdlib=libc++"
```
giordano, line 26: Why this?  Clang knows which is its C++ standard library.

BenjaminDoran: Mostly I'm following instructions for the how the developer said to install, I'll add a commit removing it

giordano:
In the [FreeBSD log](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=19813&view=logs&j=283bef7d-f611-51b1-61fc-5540bb074c4f&t=8a9748bf-24a0-574f-3710-360ca9a6c75b&l=203) I also see
```
[22:27:02] c++ -O3 -DNDEBUG -std=c++11 -Isrc -fopenmp   src/cgi/core_genome_identity.cpp -o fastANI -L/usr/local//lib -lgsl -lgslcblas -lstdc++ -lz -lm  
```
why they're forcing ` -L/usr/local//lib` and `-lstdc++`?  That's just plainly wrong.

F/FastANI/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+./configure \
+    --prefix=$prefix \
+    --build=${MACHTYPE} \
+    --host=${target} \
+    --with-gsl=${destdir} \
+    --includedir=${includedir} \
+    --libdir=${libdir}
+CXXFLAGS="-O3 -DNDEBUG -std=c++11 -Isrc -fopenmp" 
+make CXXFLAGS="${CXXFLAGS}"
+make install
```
giordano, line 26: If I'm reading it correctly, you can just replace all of this with
```suggestion
mkdir -p "${bindir}"
c++ -O3 -DNDEBUG -std=c++11 -Isrc -fopenmp   src/cgi/core_genome_identity.cpp -o "${bindir}/fastANI${exeext}" -lgsl -lgslcblas -lz -lm  
```

F/FastANI/build_tarballs.jl
giordano: Well, there is lots of nonsense there.

F/FastANI/build_tarballs.jl
```diff
@@ -0,0 +1,33 @@
...
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/FastANI
+mkdir -p "${bindir}"
+c++ -O3 -DNDEBUG -std=c++11 -Isrc -fopenmp   src/cgi/core_genome_identity.cpp -o "${bindir}/fastANI${exeext}" -lgsl -lgslcblas -lz -lm  
+"""
+
+
+platforms = supported_platforms()
+platforms = filter(!Sys.iswindows, platforms)
```
giordano, line 18: Unless they do something non-portable in the single file you're compiling, this may even work on Windows without any extra effort.  Worth a try.
```suggestion
```

giordano: Nevermind, they did manage to use non-portable functions within 100 lines of code.

BenjaminDoran:
looks like the windows platforms do have errors, so I reverted the change

giordano (approved):

PR #4951, opened 2022-05-27
New recipe: LinearFold v1.0.0

marcom:
This is a new recipe for LinearFold, an RNA secondary structure prediction program:

https://github.com/LinearFold/LinearFold

### DIFF
--- a/L/LinearFold/build_tarballs.jl
+++ b/L/LinearFold/build_tarballs.jl
@@ -25,7 +25,7 @@ install_license LICENSE
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(; experimental=true)
+platforms = supported_platforms()
 platforms = expand_cxxstring_abis(platforms)
 
 # The products that we will ensure are always built
--- a/L/LinearFold/build_tarballs.jl
+++ b/L/LinearFold/build_tarballs.jl
@@ -1,6 +1,6 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
-using BinaryBuilder, BinaryBuilderBase, Pkg
+using BinaryBuilder, Pkg
 
 name = "LinearFold"
 version = v"1.0.0"
--- a/L/LinearFold/build_tarballs.jl
+++ b/L/LinearFold/build_tarballs.jl
@@ -1,0 +1,42 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, BinaryBuilderBase, Pkg
+
+name = "LinearFold"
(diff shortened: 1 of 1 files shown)


L/LinearFold/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/LinearFold-*/
+
+make -j${nproc}
+mkdir -p ${bindir}
+for prg in linearfold_c linearfold_v; do
+    cp bin/$prg ${bindir}/${prg}${exeext}
```
giordano, line 21: I prefer to use `install`:
```suggestion
for prg in linearfold_c linearfold_v; do
    install -Dvm 755 "bin/${prg}" "${bindir}/${prg}${exeext}"
```

L/LinearFold/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+done
+
+install_license LICENSE
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; experimental=true)
```
giordano, line 29: `experimental=true` isn't necessary anymore
```suggestion
platforms = supported_platforms()
```

L/LinearFold/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, BinaryBuilderBase, Pkg
```
giordano, line 3: It doesn't hurt, but I think `BinaryBuilderBase` isn't actually needed?

marcom: No it isn't needed, i removed it. Thanks for the review!

L/LinearFold/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+sources = [
+    ArchiveSource("https://github.com/LinearFold/LinearFold/archive/refs/tags/v$(version.major).$(version.minor).tar.gz",
+                  "2ae56b5f183472c2de96782e770a91e57f82e0ab511dfc0d9d612aa4e6155f60"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/LinearFold-*/
+
+make -j${nproc}
```
giordano, line 18: Looks like the [upstream `Makefile`](https://github.com/LinearFold/LinearFold/blob/f985f1dfd4e90912fad16780cb7337d618242a3d/Makefile#L8) sets
```makefile
CC=g++
```
(yes, `CC` instead of `CXX`, sigh) which is wrong for BSD systems
```suggestion
make -j${nproc} CC=${CXX}
```

PR #4952, opened 2022-05-28
Kaleido updated to 0.2.1 + macOS aarch64 support

timolichtenstein:
Kaleido updated to 0.2.1 from new [releases](https://github.com/plotly/Kaleido/releases). This added support for macOS aarch64. `julia_compat` is now set to "1.6" for build_tarballs. With these changes, Plotly figures can be saved to image files under aarch64-julia in macOS.


giordano (commented):
Thanks for your contribution!

PR #4953, opened 2022-05-28
[OpenSpiel] Update open spiel version

findmyway:


O/OpenSpiel/build_tarballs.jl
```diff
@@ -2,32 +2,32 @@
...
 # Bash recipe for building across all platforms
 script = raw"""
 
-mv open_spiel-0.3.1 open_spiel
-mv abseil-cpp-20200923.1 open_spiel/open_spiel/abseil-cpp
+mv open_spiel-1.1.0 open_spiel
+mv abseil-cpp-20211102.0 open_spiel/open_spiel/abseil-cpp
 mv dds-0.1.1 open_spiel/open_spiel/games/bridge/double_dummy_solver
 mv hanabi-learning-environment-0.1.0 open_spiel/open_spiel/games/hanabi/hanabi-learning-environment
 mv project_acpc_server-0.1.0 open_spiel/open_spiel/games/universal_poker/acpc
```
giordano, line 26: This avoids hardocoding versions numbers every time
```suggestion
mv open_spiel-* open_spiel
mv abseil-cpp-* open_spiel/open_spiel/abseil-cpp
mv dds-* open_spiel/open_spiel/games/bridge/double_dummy_solver
mv hanabi-learning-environment-* open_spiel/open_spiel/games/hanabi/hanabi-learning-environment
mv project_acpc_server-* open_spiel/open_spiel/games/universal_poker/acpc
```
But why are you moving these directories?

O/OpenSpiel/build_tarballs.jl
```diff
@@ -55,11 +55,12 @@ products = [
...
-    BuildDependency(PackageSpec(name="libjulia_jll", version=julia_version)),
+    BuildDependency("libjulia_jll"),
     Dependency("libcxxwrap_julia_jll")
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
     preferred_gcc_version=v"9",
-    julia_compat="$(julia_version.major).$(julia_version.minor)")
+    julia_compat="1.6"
```
giordano, line 65: Why this change?  Looks like `$(julia_version.major).$(julia_version.minor)` would still be correct, no?

O/OpenSpiel/build_tarballs.jl
findmyway: Ah, thanks, that's because those dirs are hardcoded and assumed to be in the correct subfolder

O/OpenSpiel/build_tarballs.jl
findmyway: Yes, you're correct here.

PR #4954, opened 2022-05-28
[LibCURL] Remove Kerberos

giordano:
We'll consider bringing it back later, but for the time being we need to remove
support for Kerberos.


genotrance:
Curious why this was removed.

PR #4955, opened 2022-05-28
[SLICOT] update to v5.8

RalphAS:


RalphAS:
@giordano Thanks for the help.

PR #4963, opened 2022-05-30
[OpenSpiel] bump julia version

findmyway:
I get some [errors](https://github.com/JuliaReinforcementLearning/OpenSpiel.jl/actions/runs/2407149337) with the last build in https://github.com/JuliaPackaging/Yggdrasil/pull/4953 And I suspect it is due to version of libjulia_dll. (I can successfully load the library on my local machine) So this is another attempt to fix it.


findmyway:
I've tested in my local machine and the built library can now work with Julia@v1.7

O/OpenSpiel/build_tarballs.jl
```diff
@@ -55,12 +54,12 @@ products = [
...
+    Dependency("libcxxwrap_julia_jll"; platforms=platforms),
+    BuildDependency("libjulia_jll"; platforms=platforms)
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
     preferred_gcc_version=v"9",
     julia_compat="$(julia_version.major).$(julia_version.minor)"
-)
+)
```
giordano, line 65:
```suggestion
)

```

O/OpenSpiel/build_tarballs.jl
```diff
@@ -2,7 +2,7 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
-julia_version = v"1.6.3"
+julia_version = v"1.7.0"
```
giordano, line 5: You can't possibly change compat bound _and_ keep the same version of the package, this would wreak havoc for v1.6 users.  I think instead you want to follow what other packages do, see for example https://github.com/JuliaPackaging/Yggdrasil/blob/03229dc14370ddac1530e038d671728ec1a7db0b/L/libsingular_julia/build_tarballs.jl

giordano (changes requested):

O/OpenSpiel/build_tarballs.jl
findmyway: Thanks!

I tried to add support 1.6 but encountered the UUID not found error. The trick in the link provided above works very well!

giordano (approved):

PR #4965, opened 2022-05-31
[faust] Update to 2.40.10

corajr:


corajr:
> Do you need the static libraries? There are pretty large archives in `lib/`, for example `libfaustwithllvm.a` is 274.6 MiB. Also, the `share/` directory seems to contain unrelated stuff, like the `android` directory, which has libraries for other platforms (I'm not even sure where they're coming from).

I can disable the static libs, sure. The `share` directory contains various support files to let the user build an Android or iOS app based on their Faust code-- I'm not sure that anyone is likely to use that via this package, but I also don't know how to neatly remove it. Please let me know if it's a big issue.

As an aside, I'm encountering an error when trying to actually use this package in a local deploy. I've seen similar issues when searching, but none that clarified how it was resolved (Faust couldn't compile when I tried switching to LLVM 12). Do you have any idea what afix might be?

```
julia> versioninfo()
Julia Version 1.7.3
Commit 742b9abb4d (2022-05-06 12:58 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: 12th Gen Intel(R) Core(TM) i5-12600K
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, goldmont)

julia> using faust_jll
ERROR: InitError: could not load library "/home/corajr/.julia/artifacts/00264e5091ff4f48a65f30887bb195415c5223e8/lib/libfaust.so"
libLLVM-11jl.so: cannot open shared object file: No such file or directory
Stacktrace:
  [2] dlopen(s::String, flags::UInt32)
  [3] macro expansion
    @ ~/.julia/packages/JLLWrappers/QpMQW/src/products/library_generators.jl:54 [inlined]
  [4] __init__()
    @ faust_jll ~/.julia/dev/faust_jll/src/wrappers/x86_64-linux-gnu-cxx11.jl:537
[...]
```

corajr:
> You're requiring libllvm 11, but your julia comes with llvm 12. In general, you must match the llvm version you'll use at runtime.

Ah, for some reason I thought it was possible to request a different version via the dependencies and use it at runtime alongside the one Julia uses -- thanks for clearing that up. In that case, let me try a little harder to get it compiling with LLVM 12.

F/faust/build_tarballs.jl
```diff
@@ -176,7 +186,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("LLVM_jll", v"11.0.1"),
+    Dependency("LLVM_jll", v"12.0.1"),
```
giordano, line 189: I think this should be a `BuildDependency`: julia already comes with its libllvm, not need to require it
```suggestion
    BuildDependency("LLVM_jll", v"12.0.1"),
```

giordano (commented):
Is this good to go now (apart from my comment below)?  However, note that my suggestion in https://github.com/JuliaPackaging/Yggdrasil/pull/4965#issuecomment-1145429736 is still valid: you'll want to build with llvm 13 for julia v1.8

corajr:
> Is this good to go now (apart from my comment below)? However, note that my suggestion in [#4965 (comment)](https://github.com/JuliaPackaging/Yggdrasil/pull/4965#issuecomment-1145429736) is still valid: you'll want to build with llvm 13 for julia v1.8

I think so for now -- I'll tackle LLVM 13 at a later time. Thanks for all your help!

giordano:
Ok, last comment: since you're now using llvm, you should require julia v1.7.  (again, this is best handled by doing the platform augmentation suggested above, but if you can't do it now you have to fix the julia compat manually)

giordano:
Looks like it finally worked! :slightly_smiling_face:

PR #319, opened 2019-12-21
add recipe for capnproto

vchuravy:
manual PR since capnproto requires C++14

@staticfloat please review and merge

PR #4967, opened 2022-05-31
[capnproto] Version 0.8.0

baumgold:
Picking up from #319

CC: @vchuravy and @giordano


baumgold:
I believe the patch is no longer necessary thanks to the following commit:

https://github.com/capnproto/capnproto/commit/8d5b636569e36153ce99877123215502a30df68a

baumgold:
Looks like some builds failed.  I'm not well-versed enough in c++/cmake to easily fix the errors myself.  @vchuravy - are you able to assist?  Many thanks.

C/capnproto/build_tarballs.jl
```diff
@@ -41,7 +39,7 @@ fi
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = expand_cxxstring_abis(supported_platforms())
```
giordano, line 42: This this change?  It doesn't look like this is needed at all
```suggestion
platforms = supported_platforms()
```

PR #4968, opened 2022-05-31
[TetGen] Make stl reading funtionality availble

j-fu:


T/TetGen/build_tarballs.jl
```diff
@@ -57,7 +60,7 @@ install_license LICENSE
 """
 
 platforms = supported_platforms(; experimental=true)
-# platforms=[Platform("x86_64", "linux"; libc="glibc")]
+#platforms=[Platform("x86_64", "linux"; libc="glibc")]
```
giordano, line 63: What's the point of this line?  Can we just remove it?
```suggestion
```

T/TetGen/build_tarballs.jl
```diff
@@ -57,7 +60,7 @@ install_license LICENSE
 """
 
 platforms = supported_platforms(; experimental=true)
```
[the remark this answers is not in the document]
j-fu: Ah thanks!

T/TetGen/build_tarballs.jl
j-fu: It is for my own testing with TetGen.jl - just as  a note what to do during test cycles in combination with the wrapper package. For this purpose I upload into my own repo for the jll before submitting here.
Can put this elsewhere and remove this.

T/TetGen/cwrapper/cwrapper.cxx
```diff
@@ -182,8 +182,8 @@ extern "C"
     return jl_out;
   }
   
-
-
+  
+
```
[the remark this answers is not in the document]
j-fu: ok...

T/TetGen/build_tarballs.jl
giordano: You can pass the platform you want to build for on the command line: https://docs.binarybuilder.org/stable/#Manually-create-or-edit-build_tarballs.jl  That's what I always do, without touching the script

PR #4976, opened 2022-06-01
[libgeotiff] build version 1.7.1

visr:
The patch is no longer needed: https://github.com/OSGeo/libgeotiff/blob/1.7.1/libgeotiff/CMakeLists.txt#L242.

And locally at least the build compiles fine without the libstdc++ and -lcurl modifications. Or would those not be caught by the audit?


L/libgeotiff/build_tarballs.jl
```diff
@@ -60,8 +41,8 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("PROJ_jll"; compat="~800.200"),
-    Dependency("Libtiff_jll"; compat="4.3"),
+    Dependency("PROJ_jll"; compat="~900.0"),
+    Dependency("Libtiff_jll"; compat="4.4"),
```
giordano, line 45: I don't see a need to change the compat, libtiff 4.4 should have the same ABI as 4.3 (at least the soversion is the same)

visr: I see, thanks. That makes things easier. In 67caf9e50846a73ed8708ca3c381c0aca60b2769 I restored the libtiff compat. I think with that this is good to go.

PR #4977, opened 2022-06-01
[PROJ] build against libtiff 4.4

visr:


giordano (changes requested):
Same comment as https://github.com/JuliaPackaging/Yggdrasil/pull/4976#discussion_r887277058: I don't see the point of this change

PR #4979, opened 2022-06-02
[GDAL] build version 3.5.0

visr:
Soon we will have to transition the build script to CMake, see the first point of the [changelog](https://github.com/OSGeo/gdal/blob/v3.5.0/NEWS.md):

> [RFC 84](https://gdal.org/development/rfc/rfc84_cmake.html):
  Addition of a CMake build system, which deprecates the existing
  autoconf/automake and nmake build systems, that will be removed
  in GDAL 3.6.0. Users are encouraged to adopt the new CMake build system.
  Documentation of the CMake build system is at
  https://gdal.org/build_hints.html.

But let's first just get a GDAL 3.5 build using what works now.

### DIFF
--- a/G/GDAL/build_tarballs.jl
+++ b/G/GDAL/build_tarballs.jl
@@ -126,4 +126,5 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="1.6", preferred_gcc_version=v"7")
--- a/G/GDAL/build_tarballs.jl
+++ b/G/GDAL/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "GDAL"
-upstream_version = v"3.4.1"
+upstream_version = v"3.5.0"
 version_offset = v"0.0.0"
 version = VersionNumber(upstream_version.major * 100 + version_offset.major,
                         upstream_version.minor * 100 + version_offset.minor,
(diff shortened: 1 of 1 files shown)


visr:
Interesting, only `aarch64-linux-gnu-cxx03` fails, with an internal compiler error.

https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=19950&view=logs&j=1689e4e1-e047-5df5-25f6-71243ec1d23b&t=88abec16-1fed-5d53-a233-b91e4203380d&l=6265  

    [09:51:57] mitab_mapindexblock.cpp: In member function ‘int TABMAPIndexBlock::ChooseSubEntryForInsert(GInt32, GInt32, GInt32, GInt32)’:
    [09:51:57] mitab_mapindexblock.cpp:547:1: internal compiler error: Segmentation fault
    [09:51:57]  }
    [09:51:57]  ^
    [09:51:58] 0xaa37e6 crash_signal
    [09:51:58]      /workspace/srcdir/gcc-6.1.0/gcc/toplev.c:333
    [09:51:58] 0x7f96e2d771de ???
    [09:51:58]      /workspace/srcdir/musl-1.2.2/src/signal/x86_64/restore.s:1
    [09:51:58] Please submit a full bug report,
    [09:51:58] with preprocessed source if appropriate.
    [09:51:58] Please include the complete backtrace with any bug report.
    [09:51:58] See <http://gcc.gnu.org/bugs.html> for instructions.

This is the function that it doesn't like:
https://github.com/OSGeo/gdal/blob/v3.5.0/ogr/ogrsf_frmts/mitab/mitab_mapindexblock.cpp#L475-L547

Unfortunately the MITAB driver is a core part of GDAL and cannot be disabled. Not sure if we can work around this compiler error easily.

giordano:
The only solution for an ICE is to use a different version

visr:
Ok great, bumping GCC from 6 to 7 seems to work.

PR #4980, opened 2022-06-02
[0_RootFS] Add GCC 12

giordano:
Companion PR to https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/246


giordano:
Strategy suggested by @staticfloat for glibc platforms:

* do normal build for non-Intel platforms, which should work fine (I tested PowerPC64 and Aarch64)
* for Intel platforms, bootstrap GCC with Glibc 2.17 and at the end of the build replace the sysroot with an external build of Glibc 2.12.2: https://github.com/JuliaBinaryWrappers/Glibc_jll.jl/releases/tag/Glibc-v2.12.2%2B6

giordano:
I eventually managed to build glibc 2.12.2 for x86_64-linux by taking the new definition of `_dl_setup_stack_chk_guard` from https://github.com/bminor/glibc/commit/4a103975c4c4929455d60224101013888640cd2f, so no need to take a build of glibc 2.12.2 from outside.

PR #4988, opened 2022-06-05
[CompilerSupportLibraries] Always package latest `libgcc_s` in x86_64-darwin tarball

giordano:
@staticfloat is this what you were thinking to do?

### DIFF
--- a/C/CompilerSupportLibraries/build_tarballs.jl
+++ b/C/CompilerSupportLibraries/build_tarballs.jl
@@ -3,7 +3,7 @@ using BinaryBuilder, SHA
 include("../../fancy_toys.jl")
 
 name = "CompilerSupportLibraries"
-version = v"0.6.0"
+version = v"0.6.1"
 
 # We are going to need to extract the latest libstdc++ and libgomp from BB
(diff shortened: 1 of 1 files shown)


C/CompilerSupportLibraries/build_tarballs.jl
```diff
@@ -94,8 +101,10 @@ done
 # libgcc_s.X.dylib receives special treatment for now.  We need to reset the dylib id,
 # but we don't want to run a full audit, so we do it ourselves.
 if [[ ${target} == *apple* ]]; then
-    LIBGCC_NAME=$(basename $(echo ${libdir}/libgcc_s.*.dylib))
-    install_name_tool -id @rpath/${LIBGCC_NAME} ${libdir}/${LIBGCC_NAME}
+    for libgcc_s in ${libdir}/libgcc_s.*.dylib; do
+        LIBGCC_NAME=$(basename "${libgcc_s}")
+        install_name_tool -id @rpath/${LIBGCC_NAME} "${libdir}/${LIBGCC_NAME}"
+    done
 fi
```
giordano, line 108: I'm not entirely sure this is needed, we do run the full audit.

PR #4993, opened 2022-06-07
Have libclangex use the LLVM platform

vchuravy:


vchuravy:
cc: @Gnimuc this should allow us to support 1.8 and 1.9

L/libclangex/build_tarballs.jl
```diff
@@ -3,10 +3,15 @@
...
-include("../../fancy_toys.jl")
+const YGGDRASIL_DIR = "../.."
+include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
+include(joinpath(YGGDRASIL_DIR, "platforms", "llvm.jl"))
 
 name = "libclangex"
-version = v"0.1.5"
+version = v"0.1.6"
+
+llvm_versions = [#=v"11.0.1",=# v"12.0.1", v"13.0.1", v"14.0.2"]
```
Gnimuc, line 13: I think we should build a new minor version of libclangex for each LLVM major version. We can add multi-version support in the wrapper library ClangCompiler.jl.

giordano:
@vchuravy now you can use [`RuntimDependency`](https://docs.binarybuilder.org/v0.5.5/reference/#BinaryBuilderBase.RuntimeDependency) here (I'm not sure what you want to do exactly) :slightly_smiling_face:

L/libclangex/build_tarballs.jl
vchuravy: For LLVMExtra we went the other way. LLVMExtra uses minor versions to denotate bug fixes and API changes, but it supports multiple version of LLVM at once. 
Same for Enzyme. This separates the API used by ClangCompiler from the implementation on different LLVM versions.

Different Minor version becomes difficult, since you need to select which version to install based on which LLVM version is loaded by Julia. We want to move away from the model where it is bound per Julia versions....

L/libclangex/build_tarballs.jl
```diff
@@ -19,53 +23,76 @@ cd $WORKSPACE/srcdir
...
+for llvm_version in llvm_versions, llvm_assertions in (false, true)
+    # Dependencies that must be installed before this package can be built
+    llvm_name = llvm_assertions ? "LLVM_full_assert_jll" : "LLVM_full_jll"
+    dependencies = [
+        RuntimeDependency("Clang_jll"),
+        BuildDependency(PackageSpec(name=llvm_name, version=llvm_version))
+    ]
+
+    env = """
+    LLVM_MAJ_VER=$(llvm_versions.major)
```
vchuravy, line 61:
```suggestion
    LLVM_MAJ_VER=$(llvm_version.major)
```

L/libclangex/build_tarballs.jl
```diff
@@ -19,53 +23,76 @@ cd $WORKSPACE/srcdir
 cd libclangex/
 mkdir build && cd build
 cmake .. -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
-     -DLLVM_DIR=$prefix \
-     -DCLANG_DIR=$prefix \
-     -DLLVM_ASSERT_BUILD=false \
+     -DLLVM_DIR=${prefix} \
+     -DCLANG_DIR=${prefix} \
+     -DLLVM_ASSERT_BUILD=${LLVM_ASSERTIONS} \
+     -DLLVM_VERSION_MAJOR=${LLVM_MAJ_VER}
```
vchuravy, line 29:
```suggestion
     -DLLVM_VERSION_MAJOR=${LLVM_MAJ_VER} \
```

vchuravy:
Okay that's decent progress.

Gnimuc:
find_library(LLVM NAMES LLVM HINTS "${LLVM_DIR}/lib") 

Is this assumption still valid on Windows?

vchuravy:
Looking at https://github.com/JuliaBinaryWrappers/LLVM_full_jll.jl/releases/download/LLVM_full-v13.0.1%2B1/LLVM_full.v13.0.1.i686-w64-mingw32-cxx03.tar.gz it's in `bin`

vchuravy:
Oh right that renaming shenanigans

PR #4994, opened 2022-06-08
[MAGMA] Try building

Wimmerer:
There is a pre-existing builder, but I don't believe it was ever registered/complete.

This is still running on my PC, but it's taking long enough I'll let CI give it a try. I let it go for all platforms for now to see what happens, except apple obviously.


Wimmerer:
The errors on other plats seemed quite severe to me, happening very early in compilation.

M/MAGMA/build_tarballs.jl
```diff
@@ -1,56 +1,38 @@
...
-    Platform("x86_64", "macos"),
-]
-platforms = expand_gcc_versions(platforms)
+platforms = [Platform("x86_64", "linux"; libc="glibc")]
 
 # The products that we will ensure are always built
 products = [
-    LibraryProduct("libMAGMA", :libMAGMA),
+    LibraryProduct("libmagma", :libmagma),
+    LibraryProduct("libmagma_sparse", :libmagma_sparse)
```
giordano, line 28: Libraries needing CUDA generally shouldn't be dlopened automatically (otherwise that'd fail on systems without cuda): https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct
```suggestion
    LibraryProduct("libmagma", :libmagma; dont_dlopen=true),
    LibraryProduct("libmagma_sparse", :libmagma_sparse; dont_dlopen=true)
```

M/MAGMA/bundled/make.inc
```diff
@@ -0,0 +1,122 @@
...
+ifeq ($(BACKEND),hip)
+    -include make.check-hip
+
+    DEVCCFLAGS += $(FPIC) -std=c++11
+
+    #TODO: see if we need to link any HIP libraries
+    #LIB += -lhip_hcc -lhsa-runtime64
+    INC += -I$(HIPDIR)/include
+
+endif
```
giordano, line 122:
```suggestion
endif

```

Wimmerer:
Yes

PR #4996, opened 2022-06-08
[Tectonic] Set `lazy_artifacts=true`

rikhuijzer:
- Small improvement to https://github.com/JuliaPackaging/Yggdrasil/pull/4992.

It would be nice for packages which want to allow PDF building as an optional feature. As far as I see there aren't any downsides to setting lazy to true for Tectonic (?)


giordano:
> As far as I see there aren't any downsides to setting lazy to true for Tectonic (?)

Slower loading time is a good downside.

PR #5008, opened 2022-06-10
bump ViennaRNA to v2.5.1

marcom:


V/ViennaRNA/bundled/patches/windows-strcasestr-fix.patch
```diff
@@ -0,0 +1,56 @@
+diff --git a/src/ViennaRNA/io/file_formats.c b/src/ViennaRNA/io/file_formats.c
```
giordano, line 1: May be good to upstream this patch

marcom: Ok will do

PR #5013, opened 2022-06-10
[SoapyLoopback] init

sjkelly:
This adds [SoapyLoopback](https://github.com/JuliaTelecom/SoapyLoopback), a simple package used for testing [SoapySDR.jl](https://github.com/JuliaTelecom/SoapySDR.jl).

### DIFF
--- a/S/SoapyLoopback/build_tarballs.jl
+++ b/S/SoapyLoopback/build_tarballs.jl
@@ -1,0 +1,45 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "SoapyLoopback"
+version = v"0.1.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/JuliaTelecom/SoapyLoopback.git", "da9930830f6e7a97360fc959665c513a2d7aa4ea")
+]
+
+dependencies = [
+    Dependency("soapysdr_jll"; compat="0.8.0")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd SoapyLoopback
+mkdir build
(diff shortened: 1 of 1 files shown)


S/SoapyLoopback/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd SoapyLoopback
+mkdir build
+cd build
+# This package is only used in testing so we want a Debug build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+      -DCMAKE_BUILD_TYPE=Debug \
```
giordano, line 25: Why debug and not release (or release debug)?

S/SoapyLoopback/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+      ..
+make -j${nproc}
+make install
+if [[ "${target}" == *-apple-* ]]; then
+    mv ${libdir}/SoapySDR/modules0.8/libsoapyloopback.so  ${libdir}/SoapySDR/modules0.8/libsoapyloopback.dylib
+fi
```
giordano, line 31: Sounds like this should be fixed upstream?

S/SoapyLoopback/build_tarballs.jl
sjkelly: Because this library is only used for testing.

S/SoapyLoopback/build_tarballs.jl
sjkelly: Good point, since I am maintaining... I inhereted the build scripts from another project (SoapyRTLSDR)

PR #5020, opened 2022-06-11
Rebuild LLVM with forward slash off for windows

vchuravy:
Thanks to @vtjnash for spotting this


giordano:
Should upgrade to 14.0.4 or it isn't worth following their new tight release cycle?

vtjnash:
Ooooh, yeah https://reviews.llvm.org/D112787 seems awkward in the wrong ways

vtjnash:
Thanks @vchuravy for finding this

vchuravy:
> Should upgrade to 14.0.4 or it isn't worth following their new tight release cycle?

We will, once we are done with the major version update

PR #5027, opened 2022-06-12
Add Clang_jll as a RuntimeDependency to libclangext

vchuravy:


giordano:
Yay: https://github.com/JuliaBinaryWrappers/libclangex_jll.jl/commit/2f4a36b789e078be528f45c3266064c32f2cf4f6#diff-72ed386c2a0cd1d23c0968297e70023ed98c22490d146dd89fc91f48369bad4d :partying_face:

giordano:
```
julia> using libclangex_jll
ERROR: InitError: could not load library "/home/mose/.julia/artifacts/299937a26ddb50607d65ad2a493d46d2547d86de/lib/libclangex.so"
libclang-cpp.so.12jl: cannot open shared object file: No such file or directory
```
les yay :confused:

vchuravy:
Ah shite: https://github.com/JuliaPackaging/Yggdrasil/blob/acb83efd5c30451f8d421201169fa2099afab94d/L/LLVM/common.jl#L483

giordano:
There is also the problem that
```
julia> BinaryPlatforms.parse_dl_name_version("libclang-cpp.so.12jl", "linux")
ERROR: ArgumentError: Invalid dynamic library path 'libclang-cpp.so.12jl'
Stacktrace:
 [1] parse_dl_name_version(path::String, os::String)
   @ Base.BinaryPlatforms ./binaryplatforms.jl:803
 [2] top-level scope
   @ REPL[16]:1
```
and so we get `libclang-cpp.so` as soname of the library, which is wrong because
```console
shell> readelf -d /home/mose/.julia/artifacts/ebac8bb0527804a315b230ade306d81c03684e58/lib/libclang-cpp.so

Dynamic section at offset 0x3335668 contains 31 entries:
 0x0000000000000001 (NEEDED)             Shared library: [libLLVM-12jl.so]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
 0x000000000000000e (SONAME)             Library soname: [libclang-cpp.so.12jl]
```

PR #5029, opened 2022-06-13
[casacorewrapper] Release v0.1.0

kiranshila:
I'm cleaning up, updating, and packaging Michael Eastwood's wrapper for casacore from [here](https://github.com/mweastwood/CasaCore.jl), this should be good to go. Only sticky bit is that we need at least GCC 6, not sure why.

Here is the repo https://github.com/kiranshila/casascorewrapper/

This is part of the greater effort of updating [CasaCore.jl](https://github.com/mweastwood/CasaCore.jl) to modern julia.

### DIFF
--- a/C/casacorewrapper/build_tarballs.jl
+++ b/C/casacorewrapper/build_tarballs.jl
@@ -17,7 +17,6 @@ mkdir build && cd build
 cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release ..
 make -j${nproc}
 make install
-exit
 """
 
 # Use the same platforms from casacore
--- a/C/casacorewrapper/build_tarballs.jl
+++ b/C/casacorewrapper/build_tarballs.jl
@@ -28,7 +28,6 @@ make install
 platforms = supported_platforms(exclude=(platform)-> Sys.iswindows(platform) || Sys.isfreebsd(platform))
 platforms = expand_cxxstring_abis(platforms)
 # Expand libgfortran versions because we need to exclude some specific combinations
-platforms = expand_gfortran_versions(platforms)
 filter!(p -> libc(p) != "musl", platforms)
 
 # The products that we will ensure are always built
--- a/C/casacorewrapper/build_tarballs.jl
+++ b/C/casacorewrapper/build_tarballs.jl
@@ -14,7 +14,12 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/casascorewrapper/
 mkdir build && cd build
-cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release ..
+cmake \
+    -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCASACORE_ROOT_DIR=${prefix} \
+    ..
 make -j${nproc}
 make install
 """
@@ -22,6 +27,8 @@ make install
 # Use the same platforms from casacore
 platforms = supported_platforms(exclude=(platform)-> Sys.iswindows(platform) || Sys.isfreebsd(platform))
 platforms = expand_cxxstring_abis(platforms)
+# Expand libgfortran versions because we need to exclude some specific combinations
+platforms = expand_gfortran_versions(platforms)
 filter!(p -> libc(p) != "musl", platforms)
 filter!(p -> !(arch(p) == "powerpc64le" && libgfortran_version(p) == v"3"), platforms)
 
--- a/C/casacorewrapper/build_tarballs.jl
+++ b/C/casacorewrapper/build_tarballs.jl
@@ -24,7 +24,6 @@ exit
 # Use the same platforms from casacore
 platforms = supported_platforms(exclude=(platform)-> Sys.iswindows(platform) || Sys.isfreebsd(platform))
 platforms = expand_cxxstring_abis(platforms)
-platforms = expand_gfortran_versions(platforms)
 filter!(p -> libc(p) != "musl", platforms)
 filter!(p -> !(arch(p) == "powerpc64le" && libgfortran_version(p) == v"3"), platforms)
 
--- a/C/casacorewrapper/build_tarballs.jl
+++ b/C/casacorewrapper/build_tarballs.jl
@@ -21,9 +21,10 @@ make install
 exit
 """
 
-# These are the platforms we will build for by default, unless further
-# platforms are passed in on the command line
-platforms = supported_platforms()
+# Use the same platforms from casacore
+platforms = supported_platforms(exclude=(platform)-> Sys.iswindows(platform) || Sys.isfreebsd(platform))
+# Deal with the fact that we have std::string values, which causes issues across the gcc 4/5 boundary
+platforms = expand_cxxstring_abis(platforms)
 
 # The products that we will ensure are always built
 products = [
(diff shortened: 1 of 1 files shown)


C/casacorewrapper/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release ..
+make -j${nproc}
+make install
+exit
+"""
+
+# Use the same platforms from casacore
+platforms = supported_platforms(exclude=(platform)-> Sys.iswindows(platform) || Sys.isfreebsd(platform))
+platforms = expand_cxxstring_abis(platforms)
+platforms = expand_gfortran_versions(platforms)
```
giordano, line 27: Why?  I don't see any need for this one
```suggestion
```

C/casacorewrapper/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/kiranshila/casascorewrapper.git", "f7fe573901bb8d651c60fd3f3c23527232de6d57")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd casascorewrapper/
```
giordano, line 16:
```suggestion
cd $WORKSPACE/srcdir/casascorewrapper/
```

C/casacorewrapper/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release ..
+make -j${nproc}
+make install
+exit
```
giordano, line 21:
```suggestion
```

giordano (commented):
I suspect all errors are the same: we dlopen the library only for the host platform, which on our build machine is x86_64-linux-gnu-libgfortran5-cxx11, and here it fails with undefined symbols.  But the linker on Linux is happy to compile libraries with undefined symbols, and not unsurprisingly we have linker errors on BSD systems about undefined symbols.

giordano:
In particular, the linking command is
```
[20:39:52] [100%] Linking CXX shared library lib/libcasacorewrapper.so
[20:39:52] /usr/bin/cmake -E cmake_link_script CMakeFiles/casacorewrapper.dir/link.txt --verbose=true
[20:39:52] /opt/bin/x86_64-linux-gnu-libgfortran5-cxx11/x86_64-linux-gnu-g++ --sysroot=/opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root/ -fPIC -O3 -DNDEBUG -shared -Wl,-soname,libcasacorewrapper.so -o lib/libcasacorewrapper.so measures/CMakeFiles/measures.dir/measures.cpp.o tables/CMakeFiles/tables.dir/cells.cpp.o tables/CMakeFiles/tables.dir/columns.cpp.o tables/CMakeFiles/tables.dir/keywords.cpp.o tables/CMakeFiles/tables.dir/rows.cpp.o tables/CMakeFiles/tables.dir/tables.cpp.o tables/CMakeFiles/tables.dir/util.cpp.o measurement_sets/CMakeFiles/measurement_sets.dir/measurement_sets.cpp.o 
```
which looks suspiciously wrong: this is not pulling in any of the casacore libraries.

C/casacorewrapper/build_tarballs.jl
kiranshila: I just copied out of the casacore build. Is there anyway to say "use the platforms this JLL has"?

kiranshila:
> In particular, the linking command is
> 
> ```
> [20:39:52] [100%] Linking CXX shared library lib/libcasacorewrapper.so
> [20:39:52] /usr/bin/cmake -E cmake_link_script CMakeFiles/casacorewrapper.dir/link.txt --verbose=true
> [20:39:52] /opt/bin/x86_64-linux-gnu-libgfortran5-cxx11/x86_64-linux-gnu-g++ --sysroot=/opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root/ -fPIC -O3 -DNDEBUG -shared -Wl,-soname,libcasacorewrapper.so -o lib/libcasacorewrapper.so measures/CMakeFiles/measures.dir/measures.cpp.o tables/CMakeFiles/tables.dir/cells.cpp.o tables/CMakeFiles/tables.dir/columns.cpp.o tables/CMakeFiles/tables.dir/keywords.cpp.o tables/CMakeFiles/tables.dir/rows.cpp.o tables/CMakeFiles/tables.dir/tables.cpp.o tables/CMakeFiles/tables.dir/util.cpp.o measurement_sets/CMakeFiles/measurement_sets.dir/measurement_sets.cpp.o 
> ```
> 
> which looks suspiciously wrong: this is not pulling in any of the casacore libraries.

It's very possible I setup the CMake incorrectly

C/casacorewrapper/build_tarballs.jl
giordano: Not really and there is also no need for: you don't need to build this library for different libgfortran versions if there is no fortran involved. Whatever upstream libraries do doesn't matter

kiranshila:
I'm actually surprised by this last error, the build script certainly looks like it should exclude versions of powerpc64le with libgfortran3

C/casacorewrapper/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+make install
+"""
+
+# Use the same platforms from casacore
+platforms = supported_platforms(exclude=(platform)-> Sys.iswindows(platform) || Sys.isfreebsd(platform))
+platforms = expand_cxxstring_abis(platforms)
+# Expand libgfortran versions because we need to exclude some specific combinations
+platforms = expand_gfortran_versions(platforms)
+filter!(p -> libc(p) != "musl", platforms)
+filter!(p -> !(arch(p) == "powerpc64le" && libgfortran_version(p) == v"3"), platforms)
```
giordano, line 33: Also this line can be removed
```suggestion
```

C/casacorewrapper/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+    LibraryProduct("libcasacorewrapper", :libcasacorewrapper)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="casacore_jll", uuid="72fd12c2-f19b-5d3c-931a-6bbe5223e3ce"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"6.1.0")
```
giordano, line 46: Ok, this is a bit annoying. Since for powerpc we have only libgfortran 4 and 5, in order to match the ABI here we need to require gcc 7 (see https://github.com/JuliaPackaging/Yggdrasil/blob/3acc02f99ceaf85b0ada98745d9fc42dd9d6566f/RootFS.md#compiler-shards)
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"7")
```

kiranshila: Oof

C/casacorewrapper/build_tarballs.jl
kiranshila: I tried that, but it was saying it couldn't find artifacts for powerpc64le-linux-gnu-libgfortran3

kiranshila: I'll try again

kiranshila:
Yeah @giordano see, it's still trying to grab libgfortran3 things

C/casacorewrapper/build_tarballs.jl
giordano: > Yeah @giordano see, it's still trying to grab libgfortran3 things.

That's because you haven't done what I asked 🙂

kiranshila: Ah - I was getting mixed up

kiranshila: I get it! Thanks!

PR #5032, opened 2022-06-15
Add Zlib support to Boost Iostreams

joka921:
This adds several compression options (gzip, deflate, etc.) to boost::iostreams.


B/boost/build_tarballs.jl
```diff
@@ -100,6 +100,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[
```
giordano, line 102:
```suggestion
dependencies = [
```

giordano:
Thanks!

PR #5039, opened 2022-06-17
[HELICS] update to v2.8.1

nightlark:
Updates HELICS package to v2.8.1, the final bug fix release of the 2.x series. After this will be sorting out the issues with the 3.x Yggdrasil builds.


giordano:
Are you OK with requiring Julia v1.6 (which is also the current long-term support version)?

kdheepak:
Yes, Julia >= 1.6 would work for us.

PR #5042, opened 2022-06-18
[LibGit2] Add patch to fix ownership check on Windows

fxcoudert:
This patch from upstream (https://github.com/libgit2/libgit2/pull/6321) will fix the bug that prevents updating the dependency in julia currently: https://github.com/JuliaLang/julia/pull/45411#issuecomment-1159490130

### DIFF
--- a/L/LibGit2/bundled/patches/libgit2-win32-ownership.patch
+++ b/L/LibGit2/bundled/patches/libgit2-win32-ownership.patch
@@ -12,10 +12,10 @@ this behavior.
  tests/libgit2/repo/open.c | 5 +++--
  2 files changed, 4 insertions(+), 3 deletions(-)
 
-diff --git a/src/libgit2/repository.c b/src/libgit2/repository.c
+diff --git a/src/repository.c b/src/repository.c
 index 48a0b70f519..d2484318f10 100644
---- a/src/libgit2/repository.c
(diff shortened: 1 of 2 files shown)


L/LibGit2/bundled/patches/libgit2-win32-ownership.patch
```diff
@@ -0,0 +1,44 @@
...
+this behavior.
+---
+ src/libgit2/repository.c  | 2 +-
+ tests/libgit2/repo/open.c | 5 +++--
+ 2 files changed, 4 insertions(+), 3 deletions(-)
+
+diff --git a/src/libgit2/repository.c b/src/libgit2/repository.c
+index 48a0b70f519..d2484318f10 100644
+--- a/src/libgit2/repository.c
++++ b/src/libgit2/repository.c
```
giordano, line 18: In v1.4.3 this file was called `src/repository.c`: https://github.com/libgit2/libgit2/blob/v1.4.3/src/repository.c#L515.  If we're lucky, we only need to rename this, the context looks the same.  Similarly for the test file, which was called `tests/repo/open.c` (we can also discard that hunk, we don't run the tests here)

fxcoudert:
How do I trigger a new build with the latest commit? It seems it didn't fire automatically.

giordano:
Thanks!

PR #5048, opened 2022-06-20
[CasADi] add initial builder

odow:
I'm not overly sure how to structure this. The end product I'd like to build is this script: https://github.com/lanl-ansi/rosetta-opf/issues/21#issuecomment-1159837775. But it depends on CasADi. But that binary is non-standard, so I don't know if we should have CasADi_jll and CasADiIpopt_jll, or whether I should sneak the binary into CasADi_jll.

### DIFF
--- a/C/CasADi/build_tarballs.jl
+++ b/C/CasADi/build_tarballs.jl
@@ -39,12 +39,12 @@ make install
 
 products = [
     LibraryProduct("libcasadi", :libcasadi),
-    LibraryProduct("libcasadi_nlpsol_ipopt", :libcasadi_nlpsol_ipopts),
+    LibraryProduct("libcasadi_nlpsol_ipopt", :libcasadi_nlpsol_ipopt),
 ]
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
 
 dependencies = [
     Dependency("CompilerSupportLibraries_jll"),
--- a/C/CasADi/build_tarballs.jl
+++ b/C/CasADi/build_tarballs.jl
@@ -62,7 +62,9 @@ products = [
 # platforms are passed in on the command line
 platforms = supported_platforms()
 platforms = expand_cxxstring_abis(platforms)
-filter!(p -> libc(p) != "musl" && arch(p) != "powerpc64le", platforms)
+filter!(platforms) do p
+    return libc(p) != "musl" && arch(p) != "powerpc64le" && arch(p) != "armv7l"
+end
 
 dependencies = [
     Dependency("CompilerSupportLibraries_jll"),
--- a/C/CasADi/build_tarballs.jl
+++ b/C/CasADi/build_tarballs.jl
@@ -1,0 +1,65 @@
+using BinaryBuilder, Pkg
+
+name = "CasADi"
+
+version = v"3.5.5"
+
+sources = [
+    GitSource(
+        "https://github.com/casadi/casadi.git",
+        "fadc86444f3c7ab824dc3f2d91d4c0cfe7f9dad5",
+    ),
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir/casadi
+
+mkdir -p build
+cd build
+
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DWITH_IPOPT=ON \
+    -DWITH_PYTHON=OFF \
+    -DWITH_EXAMPLES=OFF \
+    ..
+
+if [[ "${target}" == *-linux-* ]]; then
+        make -j ${nproc}
+else
+    if [[ "${target}" == *-mingw* ]]; then
+        cmake --build . --config Release
+    else
+        cmake --build . --config Release --parallel
+    fi
+fi
+make install
+"""
+
+products = [
+    LibraryProduct("libcasadi", :libcasadi),
+    LibraryProduct("libcasadi_nlpsol_ipopt", :libcasadi_nlpsol_ipopts),
+]
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+dependencies = [
+    Dependency("CompilerSupportLibraries_jll"),
+    Dependency("Ipopt_jll"),
+]
+
+build_tarballs(
+    ARGS,
+    name,
+    version,
+    sources,
+    script,
+    platforms,
+    products,
+    dependencies;
+    preferred_gcc_version = v"6",
+    julia_compat = "1.6",
+)
--- a/C/CasADi/build_tarballs.jl
+++ b/C/CasADi/build_tarballs.jl
@@ -14,6 +14,7 @@ sources = [
 
 script = raw"""
 cd $WORKSPACE/srcdir/casadi
+install_license LICENSE.txt
 
 mkdir -p build
 cd build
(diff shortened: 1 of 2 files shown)


odow:
I gave up trying to get it to work on Windows. I'm mainly interested in https://github.com/lanl-ansi/rosetta-opf/issues/21. If it benchmarks well, we can spend more effort tidying it up for Windows and other users.

C/CasADi/build_tarballs.jl
```diff
@@ -0,0 +1,81 @@
...
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
+filter!(platforms) do p
+    return libc(p) != "musl" &&
+        arch(p) != "powerpc64le" &&
+        arch(p) != "armv7l" &&
+        os(p) != "windows"
```
giordano, line 62:
```suggestion
        !Sys.iswindows(p)
```
is more common

C/CasADi/build_tarballs.jl
```diff
@@ -0,0 +1,81 @@
...
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir/casadi
+install_license LICENSE.txt
+
+mkdir -p build
+cd build
+
+export CXXFLAGS="${CXXFLAGS} -std=c++11"
```
giordano, line 22: I think that maybe you were missing to make the armv7l build successful
```suggestion
export CXXFLAGS="-fPIC -std=c++11"
```
I didn't see what was the problem with the other linux platforms.

Side note: it's the package that should set the C++ standard with [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD), packagers should never do that manually.  If they have, it's upstream's fault.

C/CasADi/build_tarballs.jl
```diff
@@ -0,0 +1,81 @@
...
+
+if [[ "${target}" == *-linux-* ]]; then
+        make -j ${nproc}
+else
+    if [[ "${target}" == *-mingw* ]]; then
+        cmake --build . --config Release
+    else
+        cmake --build . --config Release --parallel
+    fi
+fi
```
giordano, line 40: I'm missing the difference between the first and the last branch.   I presume the Windows build had problems compiling in parallel?

odow: I just copied it from some of the Coin-OR scripts. And I was having trouble with Windows, but it's one of those missing symbol errors which are a pain to track down.

C/CasADi/build_tarballs.jl
odow: They nominally have a check for the standard, https://github.com/casadi/casadi/blob/402fe583f0d3cf1fc77d1e1ac933f75d86083124/CMakeLists.txt#L107-L164, but I hit some errors that only resolved when I explicitly set the standard.

giordano:
Sigh: https://www.githubstatus.com/incidents/gvzcw6sd0xxb

odow:
Windows build errors on mingw is tracked upstream at

 * https://github.com/casadi/casadi/issues/2739

which is likely caused by

 * https://github.com/casadi/casadi/issues/2341

if my understanding of 

 * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=50044

is correct

C/CasADi/build_tarballs.jl
```diff
@@ -0,0 +1,79 @@
...
+    -DWITH_IPOPT=ON \
+    -DWITH_EXAMPLES=OFF \
+    -DWITH_DEEPBIND=OFF \
+    ..
+
+if [[ "${target}" == *-linux-* ]]; then
+        make -j ${nproc}
+else
+    cmake --build . --config Release --parallel
+fi
```
giordano, line 37: I'm really not convinced there is much difference with
```suggestion
make -j ${nproc}
```

C/CasADi/build_tarballs.jl
```diff
@@ -0,0 +1,79 @@
...
+# platforms are passed in on the command line
+platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
+filter!(platforms) do p
+    # Windows export bug is:
+    #   https://gcc.gnu.org/bugzilla/show_bug.cgi?id=50044
+    # because of
+    #   https://github.com/casadi/casadi/blob/402fe583f0d3cf1fc77d1e1ac933f75d86083124/casadi/core/dm_instantiator.cpp#L373-L374
+    return !Sys.iswindows(p)
+end
```
giordano, line 61: I'm trying to build locally for Windows.  The only problem I see is that the build system installs stuff in completely bizarre and non-sense directories, so compilation of `amplexe` fails because it can't find the header file and the shared library.  Of course figuring out where in the CMake this happens is not fun.

Notice for kids at home: this uses a ***lot** of templating, make sure you have 15+ GB of RAM completely free before attempting to build with no more than 3 threads, unless you want to OOM your system.

C/CasADi/build_tarballs.jl
```diff
@@ -0,0 +1,79 @@
...
+if [[ "${target}" == *-linux-* ]]; then
+        make -j ${nproc}
+else
+    cmake --build . --config Release --parallel
+fi
+make install
+
+cd $WORKSPACE/srcdir
+
+${CXX} main.cpp -o ${bindir}/amplexe -I${includedir} -L${libdir} -lcasadi -std=c++11
```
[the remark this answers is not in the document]
odow: Yeah, the other JLL packages that produce AMPL-executables use `amplexe`, so there's `Ipopt_jll.amplexe` and `Bonmin_jll.amplexe` etc.

C/CasADi/build_tarballs.jl
odow: Haha. I guess you should have warned you about this. I could only really debug this by pushing commits to Yggdrasil, because I hit memory problems locally and it would either crash or start paging and take forever :(

giordano: Ok, for the non-sense installation the patch is
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 19e4d02c9..e7a1b43bf 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,7 +314,7 @@ option(WITH_COPYSIGN_UNDEF "On Windows, in Python.h #defines copysign to be _cop
 ############################################################
 # Flat hierarchy if self-contained and on Windows
 option(WITH_SELFCONTAINED "Make the install directory self-contained" OFF)
-if(WITH_SELFCONTAINED OR (WIN32 AND NOT CYGWIN))
+if(WITH_SELFCONTAINED)
   set(DEF_LIB_PREFIX casadi)
   set(DEF_BIN_PREFIX casadi)
   set(DEF_CMAKE_PREFIX casadi/cmake)
```
but after that, I get the same error as you
```console
sandbox:${WORKSPACE}/srcdir # ${CXX} main.cpp -o ${bindir}/amplexe -I${includedir} -L${libdir} -lcasadi -std=c++11
/tmp/ccCjdnJp.o:main.cpp:(.text$_ZNK6casadi6MatrixIdEcvSt6vectorIT_SaIS3_EEIdEEv[_ZNK6casadi6MatrixIdEcvSt6vectorIT_SaIS3_EEIdEEv]+0x61): undefined reference to `casadi::Matrix<double>::nonzeros() const'
/tmp/ccCjdnJp.o:main.cpp:(.text$_ZNK6casadi13GenericMatrixINS_6MatrixIdEEE8sparsityEv[_ZNK6casadi13GenericMatrixINS_6MatrixIdEEE8sparsityEv]+0x19): undefined reference to `casadi::Matrix<double>::sparsity() const'
```
My understanding is that mangling has gone very badly wrong.

odow: > The only problem I see is that the build system installs stuff in completely bizarre and non-sense directories

Also yet. I had a bunch of hacks to move things into the right place in order to see the linking bug:
https://github.com/JuliaPackaging/Yggdrasil/pull/5048/commits/5c1d6c64f0abe22d45167d6ddc57497b1a35e0fa

giordano:
Since we're hitting a gcc bug, I wonder if using clang would help (unless the problem is ultimately in libstdc++, which would also be used by clang).  However we lots of issues with the Clang wrapper for Windows: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/248

odow:
Thanks for the pointer, I'll take a look.

PR #5058, opened 2022-06-21
WIP: [CasADi] add patches for Windows

odow:
Follow-up the windows problems in https://github.com/JuliaPackaging/Yggdrasil/pull/5048.

### DIFF
--- a/C/CasADi/build_tarballs.jl
+++ b/C/CasADi/build_tarballs.jl
@@ -23,8 +23,8 @@ fi
 mkdir -p build
 cd build
 
-export CXXFLAGS="-fPIC -std=c++11"
-export CFLAGS="${CFLAGS} -fPIC"
+export CXXFLAGS="-std=c++11"
+export CFLAGS="${CFLAGS}"
 
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%.*}_clang.cmake \
--- a/C/CasADi/build_tarballs.jl
+++ b/C/CasADi/build_tarballs.jl
@@ -16,19 +16,18 @@ script = raw"""
 cd $WORKSPACE/srcdir/casadi
 install_license LICENSE.txt
 
+if [[ "$target" == *-mingw* ]]; then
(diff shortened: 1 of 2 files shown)


C/CasADi/build_tarballs.jl
```diff
@@ -22,6 +22,11 @@ cd build
 export CXXFLAGS="-fPIC -std=c++11"
 export CFLAGS="${CFLAGS} -fPIC"
 
+if [[ "$target" == *-mingw* ]]; then
+    atomic_patch -p1 ${WORKSPACE}/srcdir/patches/windows.patch
+    export CXX=clang++
```
giordano, line 27: No, you should do https://docs.binarybuilder.org/stable/build_tips/#Using-GCC-on-macOS-and-FreeBSD

C/CasADi/build_tarballs.jl
```diff
@@ -16,14 +16,18 @@ script = raw"""
...
+
 mkdir -p build
 cd build
 
 export CXXFLAGS="-fPIC -std=c++11"
 export CFLAGS="${CFLAGS} -fPIC"
 
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
-    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%.*}_clang.cmake \
```
odow, line 30: Do you mean something like this?

odow: I sometimes feel like you carry around a tome of incantations for all the weird things that can happen. 

It's amazing how much work has gone into all of this to make it work.

odow:
I guess I need to try your branch? https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/248

giordano:
Could try, but even with that one I had some other issues, so I'm not really sure that's enough. It really feels like we should do better in setting up clang for Windows, but I don't understand why the gcc toolchain flag is ignored

odow:
> It really feels like we should do better in setting up clang for Windows

Hahah. I get the hint. I'll take a look.

odow:
We didn't end up needing this.

PR #5059, opened 2022-06-22
[SBML] Add patch to fix C API of `SBase_getNumPlugins`

giordano:
CC: @exaexa


giordano:
Confirmed this fixes the issue:
```console
% nm i686-linux-gnu-cxx11/lib/libsbml.so | grep SBase_getNumPlugins
008727d0 T SBase_getNumPlugins
```

PR #5060, opened 2022-06-22
[utf8proc] update to v2.7.0

inkydragon:


U/utf8proc/build_tarballs.jl
```diff
@@ -18,7 +18,7 @@ if [[ "${target}" == *-mingw* ]]; then
     mkdir -p ${libdir} ${prefix}/lib ${includedir}
     cp utf8proc.h ${includedir}
     ar x libutf8proc.a
-    cc -shared -o "${libdir}/libutf8proc.${dlext}" *.o
+    $CC -shared -o "${libdir}/libutf8proc.${dlext}" *.o
```
giordano, line 21: `$CC` is `cc`, so I'm very confused by this change.

inkydragon: So `cc` will point to the correct compiler even when cross-compiling?

giordano: `cc` _is_ the cross-compiler

giordano:
Side note: at some point I'd like to come back to https://github.com/JuliaStrings/utf8proc/pull/174, to simplify compilation for Windows.  Help is welcome :slightly_smiling_face:

inkydragon:
OK, I'll see if I can help.
I'd probably be more likely to use CMake directly.

PR #5065, opened 2022-06-25
[HDF5] do not remove *_cpp and *_fortran libraries

lucifer1004:
We should not remove *_cpp and *_fortran libraries because there are many applications (especially Fortran applications) relying on them.

### DIFF
--- a/H/HDF5/build_tarballs.jl
+++ b/H/HDF5/build_tarballs.jl
@@ -33,7 +33,6 @@ mkdir -p ${libdir} ${includedir}
 if [[ ${target} == *mingw* ]]; then
     cd ${target}/mingw${nbits}
 
-    rm -f bin/{*_cpp*,*fortran*,*f90*} # we do not need these
     mv -v bin/*.dll ${libdir}
     mv -v include/* ${includedir}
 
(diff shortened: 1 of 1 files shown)


lucifer1004:
Do not know why the build on Darwin failed.

giordano:
https://github.com/JuliaPackaging/Yggdrasil/issues/4760#issuecomment-1166245518

giordano:
To elaborate on why I closed the PR very quickly, if anything, this PR should have also changed the platforms specifications to include the right cxx string ABI and libgfortran version, because the extra libraries would fix a very specific ABI. But as I pointed out in the linked comment, this isn't even a good solution in practice: the package on Linux would be completely unusable, which is a terrible outcome.

The ***only*** acceptable solution here is for us to be able to actually compile hdf5 as we want, either with upstream support for cross-compilation or BinaryBuilder being able to run code for the target platform. Until then, we have to stick with only the C library, which at least ensures the widest support

giordano:
I opened #5066 to enshrine in the recipe why this is not a good idea and avoid similar misunderstanding in the future.

PR #5084, opened 2022-06-29
move FMM3D to newer commit, try to compile on apple-aarch64

askhamwhat:


F/FMM3D/build_tarballs.jl
```diff
@@ -34,16 +34,18 @@ touch make.inc
 # force gcc
 if [[ ${target} = *apple* || ${target} = *freebsd* ]]; then
   export CC="gcc"
+  export CXX="g++"
+  export FC="gfortran"
 fi
```
giordano, line 39: Why are we forcing GCC at all?  Also, `FC` is _already_ `gfortran`.

F/FMM3D/build_tarballs.jl
```diff
@@ -76,7 +78,7 @@ exit
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = expand_gfortran_versions(supported_platforms(; experimental=true))
-filter!(p -> !(Sys.isapple(p) && arch(p) == "aarch64"), platforms)
+# filter!(p -> !(Sys.isapple(p) && arch(p) == "aarch64"), platforms)
```
giordano, line 81: Leftover comments aren't useful
```suggestion
```

F/FMM3D/build_tarballs.jl
askhamwhat: I'll see if this is still necessary. Previously, some of the other compilers would fail for reasons that weren't clear.

giordano: That was probably before llvm openmp was available: https://github.com/JuliaPackaging/Yggdrasil/blob/a248a374fcaac0bf7d2486342be543d4d323fe42/B/blis/build_tarballs.jl#L138

F/FMM3D/build_tarballs.jl
```diff
@@ -30,24 +30,26 @@ cd FMM3D/
...
 if [[ ${target} = *mingw* ]]; then
-    echo "FFLAGS= -fPIC -O3 -fno-asynchronous-unwind-tables -funroll-loops" >> make.inc;
+    echo "FFLAGS= -fPIC -O3 -fno-asynchronous-unwind-tables -funroll-loops -std=legacy" >> make.inc;
 else
-    echo "FFLAGS= -fPIC -O3 -funroll-loops" >> make.inc;
+    echo "FFLAGS= -fPIC -O3 -funroll-loops -std=legacy" >> make.inc;
 fi
 
 export OMPFLAGS="-fopenmp"
 export OMPLIBS="-lgomp"
```
giordano, line 45: I'm not entirely sure `-lgomp` or `-lomp` are necessary when already using `-fopenmp`.

F/FMM3D/build_tarballs.jl
```diff
@@ -76,15 +78,19 @@ exit
...
     LibraryProduct("libfmm3d", :libfmm3d)
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+# Borrowed from recipe for libblis (B/blis)
+dependencies = [
+    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
+    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
```
giordano, line 92: You need `CompilerSupportLibraries_jll` on all platforms because of libgfortran.
```suggestion
    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
```

askhamwhat:
@giordano Something odd seems to have happened with the checks. 7dce5de now says it passed, though I'm not sure it did in real time. If I click on the details of that run ( https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=20394&view=results ) it appears to be using the slightly updated recipe from b35380c.

The checks are currently hanging. Perhaps I've submitted tpo many runs? Let me know what I should do to proceed.

F/FMM3D/build_tarballs.jl
```diff
@@ -16,75 +16,66 @@
...
-echo "FC=${FC}" >> make.inc;
-
+FFLAGS="-fPIC -O3 -funroll-loops -std=legacy"
 if [[ ${target} = *mingw* ]]; then
-    echo "FFLAGS= -fPIC -O3 -fno-asynchronous-unwind-tables -funroll-loops" >> make.inc;
-else
-    echo "FFLAGS= -fPIC -O3 -funroll-loops" >> make.inc;
+    FFLAGS="${FFLAGS} -fno-asynchronous-unwind-tables"
 fi
+LIBS="-lm"
```
giordano, line 40: Needs to bring libgfortran now that you're linking with `cc`:
```suggestion
LIBS="-lm -lgfortran"
```

askhamwhat: Agreed.

F/FMM3D/build_tarballs.jl
```diff
@@ -16,75 +18,60 @@
...
+platforms = expand_gfortran_versions(supported_platforms())
 
 # The products that we will ensure are always built
 products = [
     LibraryProduct("libfmm3d", :libfmm3d)
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+# Borrowed from recipe for libblis (B/blis)
```
giordano, line 72:
```suggestion
```

PR #5093, opened 2022-06-30
Arb: update to 2.23.0

devmotion:
Just saw Fredrik's announcement of Arb 2.23.0 on Twitter and his blog yesterday, so I thought I'll put this PR here. Please make sure that the relevant downstream maintainers check it before it's merged :slightly_smiling_face:

### DIFF
--- a/A/Arb/build_tarballs.jl
+++ b/A/Arb/build_tarballs.jl
@@ -22,7 +22,7 @@ import Pkg.Types: VersionSpec
 # to all components.
 
 name = "Arb"
-upstream_version = v"2.22.0"
+upstream_version = v"2.23.0"
 version_offset = v"0.0.0"
 version = VersionNumber(upstream_version.major * 100 + version_offset.major,
                         upstream_version.minor * 100 + version_offset.minor,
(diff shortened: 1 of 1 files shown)


A/Arb/build_tarballs.jl
```diff
@@ -61,7 +61,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("FLINT_jll", compat = "~200.800.401"),
+    Dependency("FLINT_jll", compat = "~200.900.000"),
```
devmotion, line 64: Not sure if the bump is needed but I assumed it's a good opportunity to support the latest release of FLINT_jll.

giordano:
CC @thofma @benlorenz @fingolfin

A/Arb/build_tarballs.jl
fingolfin: FLINT 2.9 has breaking changes, so we must really carefully align this with the work being done on that right now in Nemo.jl... We'll discuss it during out weekly status call tomorrow (Friday)

thofma: I think it will be fine because we (almost) pin the exact versions of the jll in Nemo etc. @tthsqe12 can you confirm?

tthsqe12: Yup. 200.900.xxx is what is required. Now, it's just antic that has problems.

thofma (approved):

PR #5099, opened 2022-07-01
[Singular] update to 403.100.100

tthsqe12:


tthsqe12:
@hannes14 there still some problems with qsort_r on some platforms.

giordano:
Ugh, that's weird, we should be using an old musl version for the build

hannes14:
https://github.com/Singular/Singular/commit/cd85559781a82f98cee87f0b7349312ac19f2720 adds a test for qsort_r
and and an alternative for systems without it

PR #5103, opened 2022-07-04
Add build recipe for bitshuffle

jamesrhester:
Bitshuffle is an algorithm used in compression/decompression routines.

### DIFF
--- a/B/bitshuffle/build_tarballs.jl
+++ b/B/bitshuffle/build_tarballs.jl
@@ -1,0 +1,37 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "bitshuffle"
+version = v"0.4.2"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/kiyo-masui/bitshuffle.git", "a60471d37a8cbbd8265dc8cfa83a9320abdcb590")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+
+cd bitshuffle/src
+$CC -O3 -std=c99 -fPIC --shared -o bitshuffle.so iochain.c bitshuffle_core.c
+install -D -t $libdir bitshuffle.so
+exit
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("bitshuffle", :bitshuffle_lib)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")


B/bitshuffle/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+
+cd bitshuffle/src
+$CC -O3 -std=c99 -fPIC --shared -o bitshuffle.${dlext} iochain.c bitshuffle_core.c
+install -D -t ${libdir} bitshuffle.${dlext}
+exit
```
giordano, line 20: Can simplify a bit:
```suggestion
cd $WORKSPACE/srcdir/bitshuffle/src
mkdir -p "${libdir}"
cc -O3 -std=c99 -fPIC --shared -o "${libdir}/bitshuffle.${dlext}" iochain.c bitshuffle_core.c
```
Also, libraries are typically called `libNAME.${dlext}`, which I haven't changed here, but I recommend doing so for consistency with the rest of the world.

jamesrhester:
Hold off on merging this for a bit, I need to add in further pieces of the source.

jamesrhester:
Ok, that should be everything now!

mkitti:
Link the existing zstd and lz4 jlls as dependencies and use those instead of compiling them in statically.

jamesrhester:
Keeping me honest! Will try to do this, perhaps there are some examples somewhere in Yggdrasil to work off?

jamesrhester:
External linking added but fails for win64 targets for unclear reasons. Have left them out of the build for now.

B/bitshuffle/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "bitshuffle"
+version = v"0.4.2"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/kiyo-masui/bitshuffle.git", "a60471d37a8cbbd8265dc8cfa83a9320abdcb590"),
+    GitSource("https://github.com/facebook/zstd.git", "e47e674cd09583ff0503f0f6defd6d23d8b718d3")
```
giordano, line 11: Why adding this as a source?  It's already a dependency

B/bitshuffle/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+cd $WORKSPACE/srcdir/bitshuffle
+mkdir -p "${libdir}"
+cc -O3 -std=c99 -DZSTD_SUPPORT -Ilz4 -Isrc -I../zstd/lib -lzstd -llz4 -fPIC --shared -o "${libdir}/libbitshuffle.${dlext}" src/bitshuffle.c src/iochain.c src/bitshuffle_core.c
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(exclude=[Platform("i686","windows"),
+					  Platform("x86_64","windows")
+					  ])
```
giordano, line 25: What's the problem with Windows?  Also, if anything `exclude=Sys.iswindows` is much simpler.

B/bitshuffle/build_tarballs.jl
jamesrhester: Added as a source to get the header file. Not sure how to get hold of that otherwise.

B/bitshuffle/build_tarballs.jl
jamesrhester: The problem is that the mingw32 compiler tries to resolve lz4 library symbols in the shared library being built and fails. The same command-line options are fine on linux and mac. I assume there are some windows-specific command line options for building shared libraries that depend on other shared libraries but I haven't found them yet.

B/bitshuffle/build_tarballs.jl
giordano: `-I${includedir}`

B/bitshuffle/build_tarballs.jl
giordano: I can't tell more without seeing the precise error messages.

jamesrhester: Here is the full error message for target `x86_64-w64-mingw32`:

```[12:18:34]  ---> cc -O3 -std=c99 -DZSTD_SUPPORT -I${includedir} -Isrc -lzstd -llz4 -fPIC --shared -o "${libdir}/libbitshuffle.${dlext}" src/bitshuffle.c src/iochain.c src/bitshuffle_core.c
[12:18:34] src/bitshuffle.c:1:0: warning: -fPIC ignored for target (all code is position independent) [enabled by default]
[12:18:34]  /*
[12:18:34]  ^
[12:18:34] src/iochain.c:1:0: warning: -fPIC ignored for target (all code is position independent) [enabled by default]
[12:18:34]  /*
[12:18:34]  ^
[12:18:34] src/bitshuffle_core.c:1:0: warning: -fPIC ignored for target (all code is position independent) [enabled by default]
[12:18:34]  /*
[12:18:34]  ^
[12:18:34] /tmp/ccNHMMif.o:bitshuffle.c:(.text+0x42): undefined reference to `LZ4_compressBound'
[12:18:34] /tmp/ccNHMMif.o:bitshuffle.c:(.text+0xaf): undefined reference to `LZ4_compress_default'
[12:18:34] /tmp/ccNHMMif.o:bitshuffle.c:(.text+0x20a): undefined reference to `LZ4_decompress_safe'
[12:18:34] /tmp/ccNHMMif.o:bitshuffle.c:(.text+0x2dd): undefined reference to `ZSTD_compressBound'
[12:18:34] /tmp/ccNHMMif.o:bitshuffle.c:(.text+0x350): undefined reference to `ZSTD_compress'
[12:18:34] /tmp/ccNHMMif.o:bitshuffle.c:(.text+0x4a5): undefined reference to `ZSTD_decompress'
[12:18:34] /tmp/ccNHMMif.o:bitshuffle.c:(.text+0x594): undefined reference to `LZ4_compressBound'
[12:18:34] /tmp/ccNHMMif.o:bitshuffle.c:(.text+0x5d6): undefined reference to `LZ4_compressBound'
[12:18:34] /tmp/ccNHMMif.o:bitshuffle.c:(.text+0x6c5): undefined reference to `ZSTD_compressBound'
[12:18:34] /tmp/ccNHMMif.o:bitshuffle.c:(.text+0x708): undefined reference to `ZSTD_compressBound'
[12:18:34] /opt/x86_64-w64-mingw32/bin/../lib/gcc/x86_64-w64-mingw32/4.8.5/../../../../x86_64-w64-mingw32/bin/ld: /tmp/ccNHMMif.o: bad reloc address 0x0 in section `.pdata'
[12:18:34] collect2: error: ld returned 1 exit status
[12:18:34]  ---> cc -O3 -std=c99 -DZSTD_SUPPORT -I${includedir} -Isrc -lzstd -llz4 -fPIC --shared -o "${libdir}/libbitshuffle.${dlext}" src/bitshuffle.c src/iochain.c src/bitshuffle_core.c
[12:18:34] Previous command exited with 1
[12:18:34] Child Process exited, exit code 1
```

B/bitshuffle/build_tarballs.jl
jamesrhester: So the dependency `_jll`s actually contain header files? Impressive. I've updated the build script accordingly.

B/bitshuffle/build_tarballs.jl
mkitti: The problem seems to be order of the arguments. The source files should be to the left of the link parameter.

B/bitshuffle/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/kiyo-masui/bitshuffle.git", "a60471d37a8cbbd8265dc8cfa83a9320abdcb590"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/bitshuffle
+mkdir -p "${libdir}"
+cc -O3 -std=c99 -DZSTD_SUPPORT -I${includedir} -Isrc -lzstd -llz4 -fPIC --shared -o "${libdir}/libbitshuffle.${dlext}" src/bitshuffle.c src/iochain.c src/bitshuffle_core.c
```
giordano, line 17:
```suggestion
cc -O3 -std=c99 -DZSTD_SUPPORT -I${includedir} -Isrc -fPIC --shared -o "${libdir}/libbitshuffle.${dlext}" src/bitshuffle.c src/iochain.c src/bitshuffle_core.c -lzstd -llz4
```

B/bitshuffle/build_tarballs.jl
jamesrhester: Thanks for figuring that out!

PR #5111, opened 2022-07-07
Lower compatibility limit for bitshuffle.

jamesrhester:
The lower limit for Julia compatibility for bitshuffle was set to 1.6, however HDF5.jl (the main user of this library) supports 1.3.

### DIFF
--- a/B/bitshuffle/build_tarballs.jl
+++ b/B/bitshuffle/build_tarballs.jl
@@ -33,4 +33,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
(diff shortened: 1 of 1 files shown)


giordano (changes requested):
You know Julia v1.6 is the current LTS, right? And that also HDF5_jll requires Julia v1.6: https://github.com/JuliaPackaging/Yggdrasil/blob/fdf214ca416467868d232e534c34ccdc6e15a96c/H/HDF5/build_tarballs.jl#L83
And the compat can't really be changed, especially without changing the version number of the package itself, there is no way I'm going to accept this PR in the current status anyway.

jamesrhester:
OK, I'm just trying to do whatever has to be done to get the filter into HDF5. I'll see what the HDF5 people have to say about it.

PR #5116, opened 2022-07-09
[GR] bump version

jheinen:
bump GR version


jheinen:
@giordano : Unfortunately, there was a bug in the GR 0.65.1 PostScript driver. Could you please merge this PR? Thanks a lot!

PR #5122, opened 2022-07-10
add log4cplus

Crghilardi:
This PR adds a `build_tarballs.jl` for the [log4cplus](https://github.com/log4cplus/log4cplus) library. I originally wrote this using `2.0.7` probably close to a year ago but ran into issues with unit tests. A fix was merged upstream and  a new version was just released yesterday finally.

### DIFF
--- a/L/log4cplus/build_tarballs.jl
+++ b/L/log4cplus/build_tarballs.jl
@@ -7,7 +7,8 @@ version = v"2.0.8"
 
 # Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://github.com/log4cplus/log4cplus/releases/download/REL_$(version.major)_$(version.minor)_$(version.patch)/log4cplus-$version.tar.gz", "cdc3c738e00be84d8d03b580816b9f12628ecc1d71e1395080c802615d2d9ced")
+    ArchiveSource("https://github.com/log4cplus/log4cplus/releases/download/REL_$(version.major)_$(version.minor)_$(version.patch)/log4cplus-$version.tar.gz",
+                  "cdc3c738e00be84d8d03b580816b9f12628ecc1d71e1395080c802615d2d9ced")
 ]
 
 # Bash recipe for building across all platforms
@@ -15,10 +16,10 @@ script = raw"""
 cd $WORKSPACE/srcdir/log4cplus-*
 
 ./configure \
---prefix=${prefix} \
---build=${MACHTYPE} \
---host=${target} \
---enable-tests=no
+    --prefix=${prefix} \
+    --build=${MACHTYPE} \
+    --host=${target} \
+    --enable-tests=no
 
 make -j${nproc}
 make install
@@ -40,4 +41,4 @@ dependencies = Dependency[]
 
 # Build the tarballs, and possibly a `build.jl` as well.
 #use v7 to avoid posix_memalign errors on `linux-musl` platforms
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"7")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"8")
--- a/L/log4cplus/build_tarballs.jl
+++ b/L/log4cplus/build_tarballs.jl
@@ -26,13 +26,13 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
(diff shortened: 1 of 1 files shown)


L/log4cplus/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+--host=${target} \
+--enable-tests=no
+
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
```
giordano, line 29:
```suggestion
platforms = expand_cxxstring_abis(supported_platforms())
```

L/log4cplus/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("liblog4cplusU", :liblog4cplusU),
+    LibraryProduct("liblog4cplus", :liblog4cplus)
```
giordano, line 35:
```julia
julia> Base.BinaryPlatforms.parse_dl_name_version("liblog4cplus-2-0-3.dll", "windows")[1]
"liblog4cplus-2-0"

julia> Base.BinaryPlatforms.parse_dl_name_version("liblog4cplusU-2-0-3.dll", "windows")[1]
"liblog4cplusU-2-0"
```
```suggestion
    LibraryProduct(["liblog4cplusU", "liblog4cplusU-2-0"], :liblog4cplusU),
    LibraryProduct(["liblog4cplus", "liblog4cplus-2-0"], :liblog4cplus),
```

Crghilardi: Do you have any idea where the `2-0-3` is coming from? That seems like a typo error upstream or something? I would have guessed `2-0-8`? I searched around but could not find anything.

giordano: Maybe https://github.com/log4cplus/log4cplus/blob/52dd360e04d0fcdadc87cc1874884d9bbc027f99/configure.ac#L7, not entirely sure (also because I have no idea what `3.0.0` refers to).

giordano:
32-bit platforms (+ 64-bit Windows, because, why not?) with C++03 string ABI are failing with
```
[22:01:34] ./.libs/liblog4cplusU.so: undefined reference to `std::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> >::basic_string(std::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> > const&, unsigned int, std::allocator<wchar_t> const&)'
```
This is clearly related to the string ABI, but I have no clue what's going on.

Trying to use CMake instead of Autotools:
```diff
diff --git a/L/log4cplus/build_tarballs.jl b/L/log4cplus/build_tarballs.jl
index 69e8dfe81..b665a9532 100644
--- a/L/log4cplus/build_tarballs.jl
+++ b/L/log4cplus/build_tarballs.jl
@@ -14,11 +14,13 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/log4cplus-*
 
-./configure \
---prefix=${prefix} \
---build=${MACHTYPE} \
---host=${target} \
---enable-tests=no
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DLOG4CPLUS_BUILD_TESTING=OFF \
+    -DWITH_UNIT_TESTS=OFF \
+    ..
 
 make -j${nproc}
 make install
```
I can build for `i686-linux-gnu-cxx03`, but only `liblog4cplus` is built, not `liblog4cplusU` (I also have no idea what's the difference between the two, apart from the letter U at the end of their names), don't know why.  `i686-w64-mingw32-cxx03` fails with the same error as when building with autotools, but this time only `liblog4cplusU.dll` is built, not the non-U one (which would explain the failure: the error above in is in the U library).  Also, appreciate the fact that the library this time is called `liblog4cplusU.dll`, without any version number whatsoever, so the building the library with autotools or cmake results in two ABI-incompatible libraries, lovely.

Unless someone else wants to delve into why there is the missing symbol, I see only two alternatives at the moment:

1. skip the C++03 string ABI
2. cry

giordano:
> Also, appreciate the fact that the library this time is called liblog4cplusU.dll, without any version number whatsoever, so the building the library with autotools or cmake results in two ABI-incompatible libraries, lovely.

The same happens on linux: with autoconf the soname is `liblog4cplus-2.0.so.3`, with cmake it's `liblog4cplus.so.3`.

PR #5131, opened 2022-07-11
Bump Kokkos Version, Add C++17 Support

dannys4:
I'm working on a project that I'd like to put up on Yggdrasil, but it requires Kokkos to be built with C++17. I thought that simultaneously bumping the version would be an efficient way of adding that, while letting people version control if something goes awry with the C++17 option.


dannys4:
Looks like this mucks around with the mingw atomic patch. I'll try and fix that, but I'm not entirely sure of what to do.

PR #5125, opened 2022-07-11
Add libaom AV1 codec.

contradict:

PR #5132, opened 2022-07-11
Build with libaom to get AV1 support.

contradict:
This depends on #5125 to build  libaom.


contradict:
Should I increment the version number? How do I indicate that a new feature is available but this is the same release version of FFMPEG?

contradict:
Looks like upstream is a major release with some serious API changes. How about I add a build like v"4.4+libaom"?

PR #5135, opened 2022-07-11
Create MParT build tarballs

dannys4:
Adds a `build_tarballs.jl` for the [MParT repository](https://github.com/MeasureTransport/MParT).

### DIFF
--- a/M/MParT/build_tarballs.jl
+++ b/M/MParT/build_tarballs.jl
@@ -24,7 +24,7 @@ make -j${nprocs} install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = filter!(p -> Sys.isunix(p) && nbits(p) == 64, supported_platforms())
+platforms = filter!(p -> !Sys.iswindows(p) && nbits(p) == 64, supported_platforms())
 
 
 # The products that we will ensure are always built
--- a/M/MParT/build_tarballs.jl
+++ b/M/MParT/build_tarballs.jl
@@ -34,7 +34,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="Kokkos_jll", uuid="c1216c3d-6bb3-5a2b-bbbf-529b35eba709"); compat="^1.6.0")
+    Dependency(PackageSpec(name="Kokkos_jll", uuid="c1216c3d-6bb3-5a2b-bbbf-529b35eba709"); compat="=3.6.0")
     Dependency(PackageSpec(name="Eigen_jll", uuid="bc6bbf8a-a594-5541-9c57-10b0d0312c70"))
 ]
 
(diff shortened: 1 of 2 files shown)


M/MParT/build_tarballs.jl
```diff
@@ -0,0 +1,51 @@
...
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libmpart", :libmpart)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="Kokkos_jll", uuid="c1216c3d-6bb3-5a2b-bbbf-529b35eba709"))
```
giordano, line 45: You must fix compat bound with kokkos

M/MParT/build_tarballs.jl
```diff
@@ -0,0 +1,51 @@
...
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; )
+]
```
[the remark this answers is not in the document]
dannys4: Sounds good. The former is just what the wizard autogenerated.

M/MParT/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+  -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DMPART_BUILD_TESTS=OFF \
+  -DMPART_PYTHON=OFF ..
+make -j${nprocs} install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = filter!(p -> Sys.isunix(p) && nbits(p) == 64, supported_platforms())
```
giordano, line 27: `Sys.isunix` isn't defined on `Platform`s, meh
```suggestion
platforms = filter!(p -> !Sys.iswindows(p) && nbits(p) == 64, supported_platforms())
```

giordano: What's the problem with Windows BTW?

dannys4: Let me try compiling for windows! The library has bindings for python that don't work with Windows so we gave up trying to compile for it, and I just forgot it was a possibility.

giordano:
https://docs.binarybuilder.org/stable/troubleshooting/#undefined-reference-to-backtrace_symbols'

M/MParT/build_tarballs.jl
dannys4: It looks like cmake is having some trouble configuring on windows. Since it's a low priority, I'll push that off until I decide to update this-- right now the library is at an early stage, and the julia bindings are more of a proof-of-concept (though being able to use them on windows would give an edge over python bindings 😄 )

PR #4444, opened 2022-02-15
bump hsa rocm

Moelf:

PR #5151, opened 2022-07-14
hsa_rocr: Bump to 4.5.2

jpsamaroo:
Supersedes #4444

### DIFF
--- a/H/hsa_rocr/bundled/patches/musl-affinity.patch
+++ b/H/hsa_rocr/bundled/patches/musl-affinity.patch
@@ -1,0 +1,55 @@
+commit 2996bbf1b9c340b42738bc37053a76f1a52ebe42
+Author: Julian P Samaroo <jpsamaroo@jpsamaroo.me>
+Date:   Thu Jul 14 12:54:42 2022 -0500
+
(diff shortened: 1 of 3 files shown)


jpsamaroo:
I'm lower-bounding Elfutils_jll to 0.182 because earlier versions are entirely broken on musl (due to not having the obstack_jll dependency in its deps).

H/hsa_rocr/bundled/patches/musl-affinity.patch
```diff
@@ -0,0 +1,55 @@
+commit 2996bbf1b9c340b42738bc37053a76f1a52ebe42
```
giordano, line 1: Any chance of upstreaming the patches? :slightly_smiling_face:

jpsamaroo: https://github.com/RadeonOpenCompute/ROCR-Runtime/pull/143

jpsamaroo: I'm not sure about the legacy symlinks patch since I'm still not fully familiar with why they're there and whether that can impact other ROCm users.

PR #5153, opened 2022-07-15
add TulipIndicators

kevindirect:
The TulipIndicators C library: a library of technical analysis functions written in ANSI C.
https://github.com/TulipCharts/tulipindicators

### DIFF
--- a/T/TulipIndicators/build_tarballs.jl
+++ b/T/TulipIndicators/build_tarballs.jl
@@ -20,9 +20,8 @@ then
 fi
 cd $WORKSPACE/srcdir/tulipindicators
 make tiamalgamation.c
-gcc -fPIC -shared -Wall -Wextra -Wshadow -Wconversion -std=c99 -pedantic -O2 -g tiamalgamation.c -o libindicators.so
-mkdir -p ${prefix}/include ${libdir}
-mv -t ${libdir} libindicators.so
+mkdir -p ${libdir}
+cc -fPIC -shared -Wall -Wextra -Wshadow -Wconversion -std=c99 -pedantic -O2 -g tiamalgamation.c -o ${libdir}/libindicators.${dlext}
 """
 
 platforms = supported_platforms()
--- a/T/TulipIndicators/build_tarballs.jl
+++ b/T/TulipIndicators/build_tarballs.jl
@@ -25,7 +25,6 @@ cc -fPIC -shared -Wall -Wextra -Wshadow -Wconversion -std=c99 -pedantic -O2 -g t
 """
 
 platforms = supported_platforms()
-filter!(p -> !Sys.isapple(p), platforms) # TODO apple
 
 products = [
 	LibraryProduct("libindicators", :libindicators)
--- a/T/TulipIndicators/build_tarballs.jl
+++ b/T/TulipIndicators/build_tarballs.jl
@@ -1,0 +1,35 @@
+using BinaryBuilder
+
+name = "TulipIndicators"
+version = v"0.9.1"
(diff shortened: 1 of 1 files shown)


T/TulipIndicators/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
...
+if ! command -v tclsh &> /dev/null
+then
+	cd ${host_bindir}
+	ln -s tclsh* tclsh
+fi
+cd $WORKSPACE/srcdir/tulipindicators
+make tiamalgamation.c
+gcc -fPIC -shared -Wall -Wextra -Wshadow -Wconversion -std=c99 -pedantic -O2 -g tiamalgamation.c -o libindicators.so
+mkdir -p ${prefix}/include ${libdir}
+mv -t ${libdir} libindicators.so
```
giordano, line 25:
```suggestion
mkdir -p ${includedir} ${libdir}
cc -fPIC -shared -Wall -Wextra -Wshadow -Wconversion -std=c99 -pedantic -O2 -g tiamalgamation.c -o ${libdir}/libindicators.${dlext}
```
but you aren't putting anything in `${includedir}`, are you?

T/TulipIndicators/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
...
+"""
+
+platforms = supported_platforms()
+filter!(p -> !(Sys.iswindows(p) || Sys.isapple(p)), platforms) # TODO apple and windows
```
giordano, line 29: Anything particular needed for macOS and Windows?

T/TulipIndicators/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
...
+
+products = [
+	LibraryProduct("libindicators", :libindicators)
+]
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, deps; julia_compat="≥ 1.6")
```
giordano, line 35:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, deps; julia_compat="1.6")
```

T/TulipIndicators/build_tarballs.jl
kevindirect: Nope includedir not needed, thanks for catching that. I was using includedir in a previous version of the script but no longer need it. That should be removed.

Thanks for the general improvements. I wasn't aware of `cc`, wrt to your other comment if `cc` works on windows and mac (posix compliant?) it should work as tulip doesn't have any dependencies. I tested building existing targets + windows and it works. I think mac would work too but couldn't test locally because I haven't accepted the thing

T/TulipIndicators/build_tarballs.jl
kevindirect: Windows works with your suggested `cc` compile command. Mac should work too but I haven't tested

kevindirect: Having a problem when I'm trying to test apple build locally:
```
  Downloaded artifact: GCCBootstrap-aarch64-apple-darwin20.v12.0.1-iains.x86_64-linux-musl.unpacked
ERROR: LoadError: Unable to automatically install 'GCCBootstrap-aarch64-apple-darwin20.v12.0.1-iains.x86_64-linux-musl.unpacked'
```
IDK if the remote repo would have this problem though, might just be me

kevindirect: I removed the filter and it works on remote

giordano:
Thank you!

kevindirect:
Thanks for the help!

PR #5154, opened 2022-07-15
[LibSSH2] Fix import lib name on windows

inkydragon:
`liblibssh2.dll.a` ==> `libssh2.dll.a`
xref: JuliaLang/julia#45914

Do we need to inc version number to v"1.10.3"?

### DIFF
--- a/L/LibSSH2/build_tarballs.jl
+++ b/L/LibSSH2/build_tarballs.jl
@@ -18,6 +18,9 @@ cd $WORKSPACE/srcdir/libssh2*/
 
 # Apply patch to fix v1.10.0 CVE (https://github.com/libssh2/libssh2/issues/649), drop with v1.11
 atomic_patch -p1 ../patches/0001-userauth-check-for-too-large-userauth_kybd_auth_name.patch
+# Fix import lib name on windows: `liblibssh2.dll.a` ==> `libssh2.dll.a`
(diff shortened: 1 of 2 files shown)


L/LibSSH2/build_tarballs.jl
```diff
@@ -35,6 +38,11 @@ mkdir build && cd build
 cmake .. "${BUILD_FLAGS[@]}"
 make -j${nproc}
 make install
+
+# copy import lib `libssh2.dll.a` => `liblibssh2.dll.a` to maintain backwards compatibility
+if [[ ${target} == *-mingw* ]]; then
+    install -Dvm 0755 "src/libssh2.dll.a" "${prefix}/lib/liblibssh2.dll.a"
+fi
```
giordano, line 45: I'm not really sure we care about keeping compatibility, also because import libraries are relevant only at build time, not runtime.  @staticfloat opinions?

staticfloat: Yeah, I don’t think this will break anything. Especially since libssh2 moves in lock step with Julia, since it’s an stdlib.

inkydragon (commented):
[libSSH2] Revert a4df288

L/LibSSH2/build_tarballs.jl
inkydragon: Done.

PR #5158, opened 2022-07-15
[libpolymake-julia] update to 0.8.1 for polymake 4.7

benlorenz:
update flint to 2.9
drop support for 32bit platforms

### DIFF
--- a/L/libpolymake_julia/build_tarballs.jl
+++ b/L/libpolymake_julia/build_tarballs.jl
@@ -10,7 +10,7 @@ uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "libpolymake_julia"
-version = v"0.8.0"
+version = v"0.8.1"
 
 # reminder: change the above version if restricting the supported julia versions
 julia_versions = [v"1.6.3", v"1.7.0", v"1.8.0", v"1.9.0"]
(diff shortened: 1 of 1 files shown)


L/libpolymake_julia/build_tarballs.jl
```diff
@@ -48,7 +48,7 @@ $host_bindir/perl $host_bindir/polymake --iscript libpolymake-j*/src/polymake/ap
 include("../../L/libjulia/common.jl")
 
 platforms = vcat(libjulia_platforms.(julia_versions)...)
-filter!(p -> !Sys.iswindows(p) && arch(p) != "armv6l", platforms)
+filter!(p -> !Sys.iswindows(p) && nbits(p) > 32, platforms)
```
giordano, line 51: Out of curiosity, why?

giordano:
Ok, I've just noticed this:
```
[ Info: Checking lib/libpolymake_julia.so with RPath list SubString{String}["/workspace/destdir/deps/Perl_jll/lib", "/workspace/destdir/deps/GMP_jll/lib", "/workspace/destdir/deps/MPFR_jll/lib", "/workspace/destdir/deps/FLINT_jll/lib", "/workspace/destdir/lib"]
[...]
┌ Warning: Linked library libomp.so could not be resolved and could not be auto-mapped
└ @ BinaryBuilder.Auditor /depot/packages/BinaryBuilder/wohhx/src/Auditor.jl:383
```
Does this need OpenMP?  If so, you should use LLVMOpenMP on BSD systems, see for example https://github.com/JuliaPackaging/Yggdrasil/blob/6f842bdd404021135ec25291678cec7f5df63de7/B/blis/build_tarballs.jl#L135-L138

PR #5170, opened 2022-07-17
Create build_tarballs.jl

fcdimitr:
This pull request contains a new build recipe for the CSB package:

Package name: csb
Version: v1.0.0

### DIFF
--- a/C/csb/build_tarballs.jl
+++ b/C/csb/build_tarballs.jl
@@ -1,0 +1,47 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "csb"
+version = v"1.0.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/fcdimitr/CSB.git", "c53ee56e34f22f8e793ca45349fdcaa36e35f8af")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd csb/
+apk add meson --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing
+meson --cross-file=${MESON_TARGET_TOOLCHAIN%.*}_gcc.meson build
+cd build/
+ninja -j${nproc}
+ninja install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("x86_64", "freebsd"; )
+]
+
+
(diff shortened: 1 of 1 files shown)


C/csb/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "csb"
```
giordano, line 5: The name seems to be spelled all capital (CSB):
```suggestion
name = "CSB"
```
You'll also have to rename the directory accordingly.

C/csb/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/fcdimitr/CSB.git", "c53ee56e34f22f8e793ca45349fdcaa36e35f8af")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd csb/
```
giordano, line 16:
```suggestion
cd $WORKSPACE/srcdir/csb/
```

C/csb/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+apk add meson --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing
```
giordano, line 17: We already have meson, is that not enough?

C/csb/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("x86_64", "freebsd"; )
+]
```
giordano, line 33:
```suggestion
platforms = filter!(p -> proc_family(p) == "intel" && !Sys.iswindows(p), supported_platforms())
```
What's the problem with the other platforms?

fcdimitr: The `Cilk` dependency is not available on other platforms.

C/csb/build_tarballs.jl
fcdimitr: I followed the same setup as 

https://github.com/JuliaPackaging/Yggdrasil/blob/d00e02ae9d5cc7ae96ffb798fc3a38cd8b3d9e2b/S/sgtsnepi/build_tarballs.jl

since it has same dependencies and build requirements.

giordano: We have updated the rootfs in the last 13 months

fcdimitr: Oh, I see. I will fix it. Thanks!

giordano:
License is missing

fcdimitr:
> License is missing

Updated SHA ID to include MIT license.

fcdimitr:
Thank you @giordano !

PR #5171, opened 2022-07-18
[normaliz] enable openmp for all platforms

benlorenz:
by adding the correct dependencies
and adding configure flag to enable openmp

I am not sure if the version bump is necessary for this change of dependencies.

### DIFF
--- a/N/normaliz/build_tarballs.jl
+++ b/N/normaliz/build_tarballs.jl
@@ -22,7 +22,7 @@ import Pkg.Types: VersionSpec
 # to all components.
 
 name = "normaliz"
-version = v"300.900.300"
+version = v"300.900.301"
 upstream_version = v"3.9.3"
 
 # Collection of sources required to complete build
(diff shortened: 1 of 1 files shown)


giordano:
> I am not sure if the version bump is necessary for this change of dependencies.

Changing dependencies is usually best done with also changing the version number.

N/normaliz/build_tarballs.jl
```diff
@@ -59,7 +67,10 @@ dependencies = [
     Dependency("MPFR_jll", v"4.1.1"),
     Dependency("FLINT_jll"; compat = "~200.900.000"),
     Dependency("nauty_jll"; compat = "~2.6.13"),
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
+    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
```
giordano, line 72: To be clear, CSL is needed only for libgomp?

PR #5172, opened 2022-07-18
Provide 64_ suffixes for ILP64 BLAS and LAPACK on 64-bit architectures

ViralBShah:


ViralBShah:
@giordano Is it possible to download the binaries built by this PR and see if they are ok, before this is merged?

ViralBShah:
Is there a better way to rename these symbols using something like `objtool`? The source approach doesn't seem to work too well. 

cc @staticfloat

ViralBShah:
@giordano It is only linux+libgfortran3 that fails, and the error is actually that it is a compiler internal bug. Crashes on gcc v6 and v7. 

```
0xa49996 crash_signal
[23:42:11] 	/workspace/srcdir/gcc-6.1.0/gcc/toplev.c:333
[23:42:11] 0x7faaca7111de ???
[23:42:11] 	/workspace/srcdir/musl-1.2.2/src/signal/x86_64/restore.s:1
[23:42:11] Please submit a full bug report,
[23:42:11] with preprocessed source if appropriate.
[23:42:11] Please include the complete backtrace with any bug report.
[23:42:11] See <[http://gcc.gnu.org/bugs.html>](http://gcc.gnu.org/bugs.html%3E) for instructions.
[23:42:11] make[2]: *** [SRC/CMakeFiles/lapack.dir/build.make:15467: SRC/CMakeFiles/lapack.dir/clarfy.f.o] Error 1
```

Any idea how to resolve? Should I try newer  gcc versions? Or drop these platforms?

giordano:
The only thing to do for an ICE is to use a different version of the compiler, but for libgfortran 3 there are only gcc 4.8, 5.2 and 6.1, not many other options, and you're already using the latest one.

ViralBShah:
How can I drop libgfortran3 on linux aarch64? I suppose it is old enough not to matter much.

ViralBShah:
@giordano This looks good to me. May need iterations, but it's doing what I set out to do.

PR #5173, opened 2022-07-18
LibOpenJpegExtra basic library

ashwani-rathee:
- [x] Adds basic LibOpenJpegExtra version

This one adds extra apis from openjpeg
The repository is at: https://github.com/ashwani-rathee/libopenjpegextra

### DIFF
--- a/L/libopenjpegextra/build_tarballs.jl
+++ b/L/libopenjpegextra/build_tarballs.jl
@@ -1,0 +1,35 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "libopenjpegextra"
+version = v"0.0.1"
+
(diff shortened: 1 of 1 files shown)


giordano:
https://github.com/ashwani-rathee/libopenjpegextra/blob/b5321f5e96b00e50adad61eaf23561a8bd8e36ce/Makefile#L1 shouldn't force `CC = gcc`, we generally don't want to use gcc on BSD systems.

ashwani-rathee:
> https://github.com/ashwani-rathee/libopenjpegextra/blob/b5321f5e96b00e50adad61eaf23561a8bd8e36ce/Makefile#L1 shouldn't force `CC = gcc`, we generally don't want to use gcc on BSD systems.

can I just remove `CC = gcc` and assume it will be set most appropriately by binarybuilders.jl itself through the environment variables?

giordano:
`CC` is always set, yes.

PR #5181, opened 2022-07-19
[lapack] 3.10.0 + LBT 5.1.1 + ILP64

ViralBShah:
Address https://github.com/JuliaRegistries/General/pull/64542#issuecomment-1189195017


L/LAPACK/build_tarballs.jl
```diff
@@ -296,9 +296,9 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[
```
giordano, line 297: Also:
```suggestion
dependencies = [
```

PR #5182, opened 2022-07-19
NOMAD_JLL: bump

SobhanMP:


N/NOMAD/build_tarballs.jl
```diff
@@ -36,8 +37,8 @@ products = [
...
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", 
+    uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
 ]
-
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"8.1.0")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"12", julia_compat="v1.8")
```
giordano, line 47: Why using GCC 12 and Julia v1.8?  There is no way this is going to work with GCC 12

SobhanMP: why not?

SobhanMP: i've built this with gcc 12/11 a bunch of times

giordano: By "work" I mean that it won't work at runtime,  because we need to use libstdc++ from Julia, which at the moment has GCC up to version 11 (and older versions of Julia will have older versions of libstdc++).  For maximum compatibility you _**must**_ use the oldest version of GCC possible, not the latest one.

SobhanMP: the only reason it's gcc 8.1.0 is that the code errors that it's not tested on gcc 8.1.0

SobhanMP: i see

SobhanMP:
@amontoison here's the bumped version

SobhanMP:
@ whoever can merge this, please do.

N/NOMAD/build_tarballs.jl
```diff
@@ -1,30 +1,40 @@
...
+    -DCMAKE_BUILD_TYPE=Release ..
 make -j${nproc}
 make install
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(supported_platforms())
-
+platforms = expand_cxxstring_abis(supported_platforms(; exclude=Sys.iswindows))
```
giordano, line 35: I'm not a fan of removing previously working platforms (and windows isn't a niche operating system, sadly).

SobhanMP: it doesn't compile on mingw nothing i can do about that. i'll add it back whenever https://github.com/bbopt/nomad/pull/82 is closed.

giordano (changes requested):
I agree with https://github.com/JuliaPackaging/Yggdrasil/pull/5182#discussion_r925595023

N/NOMAD/build_tarballs.jl
```diff
@@ -1,30 +1,40 @@
...
 
 name = "NOMAD"
-version = v"4.0.1"
+version = v"4.2.0"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/amontoison/nomad.git","f6a4b4f18111372e3e7190b019e76ed86b915ddc"),
-    DirectorySource("./bundled"),
+    GitSource("https://github.com/bbopt/nomad.git", "ad22bef57c84897dab553390ff77c31eaaedf5c3"),
```
amontoison, line 8:
```suggestion
    GitSource("https://github.com/bbopt/nomad.git", "cf824ea62defeeff558ab91ab3f5ecd00e08aa00"),
```

N/NOMAD/build_tarballs.jl
amontoison, line 35:
```suggestion
platforms = expand_cxxstring_abis(supported_platforms())
```

amontoison (changes requested):
It should compile with MinGW now.

SobhanMP:
thanks!

PR #5183, opened 2022-07-19
LLD_jll: Add compat builds for 12 and 13

jpsamaroo:
Since LLD_jll only supports LLVM 14 and Julia 1.9, it's not possible to depend on it and also support Julia 1.7/1.8. The intent of these builders (which don't currently work) is to re-export `lld` and friends from LLVM_jll.


jpsamaroo:
Closing as alternative is available in `LLD_jll#compat`

PR #5184, opened 2022-07-19
[VMEC] Update dependency versions

benjaminfaber:
Some of the issues with previous dependencies have been fixed, no longer need to fix to old versions.


V/VMEC/build_tarballs.jl
```diff
@@ -65,11 +65,10 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     # MbedTLS is an indirect dependency, fix the version for building 
-    BuildDependency(PackageSpec(name = "MbedTLS_jll", version = v"2.24.0")),
-    Dependency("MPICH_jll"; platforms=filter(!Sys.iswindows, platforms)),
-    Dependency("OpenBLAS_jll"),
+    BuildDependency(PackageSpec(name = "MbedTLS_jll")),
```
giordano, line 68: libblastrampoline works well only with Julia v1.7+, and there is also the problem that libblastrampoline 3 (the one used in Julia v1.7) and 5 (the one you're using to build here) have incompatible ABIs: https://github.com/JuliaRegistries/General/pull/64542#issuecomment-1189195017

V/VMEC/build_tarballs.jl
```diff
@@ -65,11 +65,10 @@ products = [
...
+    Dependency("MPItrampoline_jll"; platforms=filter(!Sys.iswindows, platforms)),
```
giordano, line 69: For MPITrampoline you probably want to follow the example of the other packages using it.

benjaminfaber: I followed the examples of the other packages using MPItrampoline and it seems to build correctly.  If you approve, could this get merged soon?  This is intended to resolve MPI difficulties that have been occurring on HPC systems.

V/VMEC/build_tarballs.jl
benjaminfaber: Fixed to work with 1.7, I saw an example in [P/Pfapack](https://github.com/JuliaPackaging/Yggdrasil/P/Pfapack/build_tarballs.jl) of how to do this for 1.7 and 1.8, should I be following that?

benjaminfaber: I removed the libblastrampoline dependence for now.

giordano:
I think you're missing something, because the platforms don't include the MPI tag

benjaminfaber:
@giordano I think I have fixed all the issues and am successfully building with MPItrampoline appropriately, I would appreciate a review

PR #5186, opened 2022-07-20
[lapack] Mirror the openblas build setup

ViralBShah:
We mirror the OpenBLAS setup:

1. `LAPACK_jll`: LP64 on 32-bit and ILP64 with `64_` suffixes on 64-bit systems
2. `LAPACK32_jll`: Always LP64


amontoison:
@ViralBShah
Is it possible to recompile the shared library of LAPACK (ILP64 version) with a different name / soname such that LP64 / ILP64 libraries can coexist?
We will need both if we want to use BLIS instead of OpenBLAS for BLAS and LAPACK routines.

ViralBShah:
Yes, that is a good idea.

amontoison:
I checked the dependencies of the two jll packages and if we rename one shared library it should be the LP64 version because it's only used by another jll (https://juliahub.com/ui/Packages/LAPACK32_jll/lE5Ka/3.10.0+0).
LAPACK_jll has 7 dependents (https://juliahub.com/ui/Packages/LAPACK_jll/KGp1v/3.10.0+1).
I think that we can't use jll precompiled with LAPACK_jll and LAPACK32_jll in the same environment.

PR #5196, opened 2022-07-20
Add ROCm LLVM 4.5.2

pxl-th:
- Separate ROCm LLVM versions.
- Use separate buildscript and `common.jl` for ROCmLLVM.
- Build only for cxx11, since ROCmCompilerSupport and other dependencies up the chain need it.


pxl-th:
@jpsamaroo, I removed `CMAKE_FLAGS+=(-DLLVM_USE_PERF=1)` & `CMAKE_FLAGS+=(-DLLVM_USE_INTEL_JITEVENTS=1)` flags from `LLVM/common.jl` since with them I was getting ` undefined reference to llvm::JITEventListener::createPerfJITEventListener()` error, similar to [this](https://github.com/JuliaLang/julia/issues/33417).
Do you know if we need it?

Also there are other changes to `common.jl` to make 4.2.0 build again.
For 4.5.2, it `Could not locate MLIR, libMLIR`.

L/LLVM/common.jl
```diff
@@ -451,7 +451,7 @@ function configure_build(ARGS, version; experimental_platforms=false, assert=fal
         LibraryProduct("libclang", :libclang, dont_dlopen=true),
         LibraryProduct(["LTO", "libLTO"], :liblto, dont_dlopen=true),
         ExecutableProduct("llvm-config", :llvm_config, "tools"),
-        ExecutableProduct(["clang", "clang-$(version.major)"], :clang, "tools"),
+        ExecutableProduct(["clang", "clang-$(version.major)"], :clang, "bin"),
```
jpsamaroo, line 454: Gate this based on LLVM version.

L/LLVM/common.jl
```diff
@@ -469,11 +469,7 @@ function configure_build(ARGS, version; experimental_platforms=false, assert=fal
     end
     if version >= v"12"
         push!(products, LibraryProduct("libclang-cpp", :libclang_cpp, dont_dlopen=true))
-        push!(products, ExecutableProduct("lld", :lld, "tools"))
-        push!(products, ExecutableProduct("ld.lld", :ld_lld, "tools"))
-        push!(products, ExecutableProduct("ld64.lld", :ld64_lld, "tools"))
-        push!(products, ExecutableProduct("lld-link", :lld_link, "tools"))
-        push!(products, ExecutableProduct("wasm-ld", :wasm_ld, "tools"))
+        push!(products, ExecutableProduct("lld", :lld, "bin"))
```
jpsamaroo, line 472: We need to keep these here, we just need to figure out where they are for any given LLVM version.

pxl-th:
Do we even need MLIR for ROCm?
I see that [spack](https://github.com/spack/spack/blob/ec8c8e8aa84a4e62a3d846208865abaa048ebd9b/var/spack/repos/builtin/packages/llvm-amdgpu/package.py#L122) and [arch](https://github.com/rocm-arch/rocm-arch/blob/607616712adf87de9db85b3a76acae64fe2b8a1c/rocm-llvm/PKGBUILD#L46) both don't use it.

L/LLVM/common.jl
pxl-th: Added option, but based on whether we are building ROCmLLVM or not...

L/LLVM/common.jl
pxl-th: image

They are present in `tools/` dir, but the symlinks are invalid. We fix them in rocBLAS, for example, but manually symlinking `ln -s ${prefix}/bin/lld ${prefix}/tools/lld`.

jpsamaroo: Yeah, then let's fix them in this build script.

pxl-th:
@jpsamaroo FYI, I have successfully built ROCm stack up until the rocBLAS with this to test that it works.
It installs into `${prefix}/llvm` which makes it actually easier for other packages to discover it, instead of `${prefix}`.

L/LLVM/common.jl
pxl-th: Fixed in the latest commit.

R/ROCmLLVM/common.jl
```diff
@@ -0,0 +1,52 @@
...
+        LibraryProduct("libclang-cpp", :libclang_cpp, llib; dont_dlopen=true),
+        LibraryProduct(["LTO", "libLTO"], :liblto, llib; dont_dlopen=true),
+        ExecutableProduct("llvm-config", :llvm_config, lbin),
+        ExecutableProduct(["clang", "clang-$(llvm_version.major)"], :clang, lbin),
+        ExecutableProduct("opt", :opt, lbin),
+        ExecutableProduct("llc", :llc, lbin),
+        ExecutableProduct("lld", :lld, lbin),
+        ExecutableProduct("ld.lld", :ld_lld, lbin),
+        ExecutableProduct("ld64.lld", :ld64_lld, lbin),
+        ExecutableProduct("lld-link", :lld_link, lbin),
```
jpsamaroo, line 40:
```suggestion
```

Since we're not supporting Windows or macOS.

R/ROCmLLVM/common.jl
```diff
@@ -0,0 +1,52 @@
...
+    llvm_version = LLVM_VERSIONS[rocm_version]
+
+    sources = [
+        GitSource(ROCM_GIT, git_version),
+        DirectorySource("./bundled"),
+    ]
+
+    llib = "llvm/lib"
+    lbin = "llvm/bin"
+    products = [
```
jpsamaroo, line 29: What about the LLVM static libraries? It would be good to ensure that they're built, for downstream dependencies.

jpsamaroo (approved):
LGTM! @vchuravy can you take a look at this?

R/ROCmLLVM/common.jl
pxl-th: Added a small subset of them. I guess there's no need to list them all?

vchuravy (approved):

R/ROCmLLVM/common.jl
jpsamaroo: Yeah, sounds good to me!

PR #5197, opened 2022-07-20
[polymake] fixup: remove openmp flag from config for compiling at runtime on macos

benlorenz:
The last update enabling OpenMP for more polymake platforms did unfortunately break some specific Oscar.jl tests because Apple disabled openmp support in their compilers (why?) and the `-fopenmp` flag from the config is not accepted at runtime (polymake does need to compile some C++ template instantiations at runtime).

So we remove this flag from the installed config file at the end.

@giordano a quick merge would be appreciated, thanks!


giordano:
Do you also need to remove LLVM OpenMP then?  Apple doesn't believe in OpenMP at all

giordano:
Ok, good

PR #5199, opened 2022-07-21
GMT

boriskaus:
This creates a BinaryBuilder version of GMT, the Generic Mapping Tools.


G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd gmt-6.2.0/
+mkdir build
+cd build/
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DHAVE_QSORT_R_GLIBC=False -DHAVE___BUILTIN_BSWAP16=False -DHAVE___BUILTIN_BSWAP32=False -DHAVE___BUILTIN_BSWAP64=False   .. 
+make -j${nproc} 
+make install 
+exit
```
giordano, line 22:
```suggestion
cd $WORKSPACE/srcdir/gmt-*
mkdir build
cd build/
cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DHAVE_QSORT_R_GLIBC=False \
    -DHAVE___BUILTIN_BSWAP16=False \
    -DHAVE___BUILTIN_BSWAP32=False \
    -DHAVE___BUILTIN_BSWAP64=False \
    .. 
make -j${nproc} 
make install 
```
is a bit more readable (and independent from upstream version)

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+]
```
giordano, line 32: Any hope for Windows, which I think is the more problematic platform?

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "GMT"
+version = v"6.2.0"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/GenericMappingTools/gmt/releases/download/$(version)/gmt-$(version)-src.tar.gz", "ab7062912aeead1021770fad4756e0a99860fde8ea9b428fb00c22fa15a3bbfc")
```
giordano, line 10:
```suggestion
    ArchiveSource("https://github.com/GenericMappingTools/gmt/releases/download/$(version)/gmt-$(version)-src.tar.gz",
                  "ab7062912aeead1021770fad4756e0a99860fde8ea9b428fb00c22fa15a3bbfc")
```
just to avoid a very long line,

G/GMT/build_tarballs.jl
boriskaus: Sure, windows is usually the trouble-maker (even when [GMT.jl](https://github.com/GenericMappingTools/GMT.jl) appears easier to install on windows than on Mac/Linux). Ideally, we would thus have a Windows-build as well.

I made some progress meanwhile, by using an older version of NetCDF (4.7.1), and applying a few patches (`mingw` calls it `<windows.h>` instead of `<Windows.h>` etc.). 
Yet, I still run int a mingw error after a while. 

I'll upload these changes as soon as I have time; perhaps you or someone else have suggestions on how to fix that

boriskaus:
The errors on windows are:
```
[21:27:19] CMakeFiles/gmtlib.dir/objects.a(gmt_fft.c.obj):gmt_fft.c:(.text+0x3825): undefined reference to `__imp_gethostname'
[21:27:19] CMakeFiles/gmtlib.dir/objects.a(gmt_init.c.obj):gmt_init.c:(.text+0x3d248): undefined reference to `__imp__set_fmode'
[21:27:19] collect2: error: ld returned 1 exit status
[21:27:19] make[2]: *** [src/CMakeFiles/gmtlib.dir/build.make:2561: src/gmt.dll] Error 1
```

if that rings a bell for someone, suggestions are helpful!

giordano:
For [`gethostname`](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-gethostname) I guess you need to link to `-lws2_32`? Not sure about [`_set_fmode`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/set-fmode?view=msvc-170), documentation doesn't indicate what you need to link to :confused:

giordano:
According to
```console
julia --compile=min -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("x86_64", "windows"))'
sandbox:${WORKSPACE} # for file in /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/lib/*.a; do nm $file | grep __imp__set_fmode && echo "found in ${file}"; done
0000000000000000 I __imp__set_fmode
found in /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/lib/libmsvcr100.a
0000000000000000 I __imp__set_fmode
found in /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/lib/libmsvcr110.a
0000000000000000 I __imp__set_fmode
found in /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/lib/libmsvcr120.a
0000000000000000 I __imp__set_fmode
found in /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/lib/libmsvcr120_app.a
0000000000000000 I __imp__set_fmode
found in /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/lib/libmsvcr120d.a
0000000000000000 I __imp__set_fmode
found in /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/lib/libmsvcr90.a
0000000000000000 I __imp__set_fmode
found in /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/lib/libmsvcr90d.a
0000000000000000 I __imp__set_fmode
found in /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/lib/libucrt.a
0000000000000000 I __imp__set_fmode
found in /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/lib/libucrtbase.a
```
`_set_fmode` is provided by a bunch of internal libraries, I'm not sure you're supposed to manually link to any of them.

boriskaus:
thanks. If I manually compile this with 
```
-lws2_32 -lmsvcr100
```
it works & the whole code compiles. Do you happen to know how I can tell `CMake` to take this libraries into account as well?

giordano:
You have to add the libraries names to the target which needs them, for example with [`target_link_libraries`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) or whatever they're using to specify the dependencies.  `ws2_32` is definitely public API, but `msvcr100` is something very internal (like all other libraries which provide `_set_fmode`), you just picked up a random library from the list I shared, that doesn't sound right.  I have no idea how `_set_fmode` is supposed to be properly linked.

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libpostscriptlight", :libpostscriptlight),
+    LibraryProduct("libgmt", :libgmt),
+    ExecutableProduct("gmt", :gmt)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="LibCURL_jll", uuid="deac9b47-8bc7-5906-a0fe-35ac56dc84c0"))
```
giordano, line 59: Is libcurl really needed?  it also forces a link dependency on mbedtls, which is an awful mess to deal with, I'd strongly recommend avoiding it if possible at all.

G/GMT/bundled/patches/win_mingw.patch
```diff
@@ -0,0 +1,66 @@
...
+ 		if (use_locale) {
+ 			char *new_format = gmt_strrep(format, "%'", "%");
+ 			sprintf (string, new_format, x);
+@@ -17498,7 +17498,7 @@ bool gmt_check_executable (struct GMT_CTRL *GMT, char *program, char *arg, char
+ 	bool answer = false;
+ 
+ 	/* Turn off any stderr messages coming to the terminal */
+-	if (strchr (program, ' ')) {	/* Command has spaces [most likely under Windows] */
++	if (strchr (program, ' ')) {	/* Command has spaces [most likely under windows] */
+ 		if (!(program[0] == '\'' || program[0] == '\"'))	/* Not in quotes, place double quotes */
```
giordano, line 64: Sounds like you replaced too much?

G/GMT/bundled/patches/win_mingw.patch
```diff
@@ -0,0 +1,66 @@
+diff --git before/gmt-6.2.0/src/gmt_api.c after/gmt-6.2.0/src/gmt_api.c
```
giordano, line 1: This patch (minus the unrelated comments changed) should be upstreamed.

G/GMT/build_tarballs.jl
giordano: Actually, it isn't _that_ bad:
```console
% readelf -d libgmt.so

Dynamic section at offset 0x457de0 contains 38 entries:
 0x0000000000000001 (NEEDED)             Shared library: [libnetcdf.so.18]
 0x0000000000000001 (NEEDED)             Shared library: [libcurl.so.4]
 0x0000000000000001 (NEEDED)             Shared library: [libgdal.so.31]
 0x0000000000000001 (NEEDED)             Shared library: [libgeos_c.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libpcre.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libfftw3f.so.3]
 0x0000000000000001 (NEEDED)             Shared library: [libopenblas.so]
 0x0000000000000001 (NEEDED)             Shared library: [libpostscriptlight.so.6]
 0x000000000000000e (SONAME)             Library soname: [libgmt.so.6]
 0x000000000000001d (RUNPATH)            Library runpath: [/usr/local/lib:$ORIGIN]
```
It doesn't link directly to mbedltse, only curl, so we only need to use the right version of curl when building
```suggestion
    Dependency(PackageSpec(name="LibCURL_jll", uuid="deac9b47-8bc7-5906-a0fe-35ac56dc84c0"); compat="7.73.0")
```
But that `/usr/local/lib` in the runpath looks definitely wrong.

giordano:
Some creative uses of linker flags:
```
[21:27:01] /opt/bin/aarch64-linux-gnu-libgfortran4-cxx11/aarch64-linux-gnu-gcc [...] -Wl,-rpath,/usr/local/lib::::::::::::::::::::::: [...]
[...]
[21:27:04] /opt/bin/aarch64-linux-gnu-libgfortran4-cxx11/aarch64-linux-gnu-gcc [...] -Wl,-rpath,/usr/local/lib:/workspace/srcdir/gmt-6.2.0/build/src: [...] 
```
Linking a source directory?!?!?

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+products = [
+    LibraryProduct("libpostscriptlight", :libpostscriptlight),
+    LibraryProduct("libgmt", :libgmt),
+    ExecutableProduct("gmt", :gmt)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="LibCURL_jll", uuid="deac9b47-8bc7-5906-a0fe-35ac56dc84c0"); compat="7.73.0")
+    Dependency("NetCDF_jll", v"4.7.4")
```
giordano, line 60: You're using the wrong netcdf version.  You should use the one corresponding to https://github.com/JuliaPackaging/Yggdrasil/blob/1ae78024d04123f2ee1e91736d61e2a807b004b6/N/NetCDF/NetCDF%40julia-1.6/build_tarballs.jl, which I guess is 400.702.402.  And you also must specify it as a compat

boriskaus: thanks, just added that.

G/GMT/bundled/patches/win_mingw.patch
boriskaus: sorry, you lost me here...

giordano: [Upstream](https://en.wikipedia.org/wiki/Upstream_(software_development)) == Submit a pull request to https://github.com/GenericMappingTools/gmt with this change.

G/GMT/build_tarballs.jl
giordano: With this change now this is working on two more platforms :slightly_smiling_face:

boriskaus:
ok, I keep being confused with CMake. If I manually add this at the end:
```
/opt/bin/x86_64-w64-mingw32-libgfortran4-cxx11/x86_64-w64-mingw32-gcc --sysroot=/opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/ -std=gnu99  -shared -o gmt.dll -Wl,--out-implib,gmt.dll.a -Wl,--major-image-version,6,--minor-image-version,2 -Wl,--whole-archive CMakeFiles/gmtlib.dir/objects.a -Wl,--no-whole-archive @CMakeFiles/gmtlib.dir/linklibs.rsp -lws2_32 -lmsvcr100
```
it works. 
Yet, any attempt to add the libraries as flags results in something like
```
/opt/bin/x86_64-w64-mingw32-libgfortran4-cxx11/x86_64-w64-mingw32-gcc --sysroot=/opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/ -std=gnu99  -lws2_32 -lmsvcr100 -shared -o gmt.dll -Wl,--out-implib,gmt.dll.a -Wl,--major-image-version,6,--minor-image-version,2 -Wl,--whole-archive CMakeFiles/gmtlib.dir/objects.a -Wl,--no-whole-archive @CMakeFiles/gmtlib.dir/linklibs.rsp
```
which does not work.

joa-quim:
@boriskaus Thanks a lot for trying this. It's a really needed solution but unfortunately, and that's my personal view, the real problem lies on Unix that tries to force Windows be Unix and when it fails it blames Windows. 

The GMT Windows build is perfectly solved with the VS build (took us several years of continuous improvements) and that is not going to be replaced by a far inferior MinGW build. But it still leaves out the *nix builds and that's where the effort should be focused. Unfortunately, even there the constrain of everything be build-able with MinGW too is severely harming the *nix side when the Win builds pose problems that have not yet been solve (e.g. GDAL, netCDF, and probably others).

I'll post this comments in the BinaryBuilder issue later today.

joa-quim:
Could you give it another try? I fixed the ``gethostname`` and the ``_set_fmode`` issues upstream where I was finally able to build GMT with MinGW. But note that those changes are in GMT#master only

joa-quim:
Hmm, let me see if I understand it. You revising(???) the GMT C code and not accepting it and therefore this attempt of creating a ``GMT_JLL`` cannot proceed while you do not approve GMT C, right?

We may all be very dumb on GMT (I'm not by far the largest contributor) but we are are a bit aware of the ``WIN32`` *vs* ``_WIN32`` mess.
https://github.com/GenericMappingTools/gmt/blob/master/src/gmt_notposix.h#L148

And you may also do not like (or believe) to hear that along the GMT developing time we have fund **dozens** of memory access bugs that were detected **only** by the *non-standard compliant compilers like MSVC* and that gcc and clang had happily let go.

giordano:
No, you do not understand correctly.  I am pointing out factual errors in the source code which prevent correct compilation.  We cannot merge a PR if compilation fails for a platform (the alternative being excluding that platform altogether).  And compilation for Windows was failing here :shrug:

joa-quim:
> And compilation for Windows was failing here

Yes, I'm aware of that but the point is: Is it still failing after the upstream fixes? And if yes, where/why.

giordano:
No, I haven't tried, but I just looked for obvious mistakes.

joa-quim:
> but I just looked for obvious mistakes.

We could continue discussing this for ever, which I will not. At least while the tone maintains.

boriskaus:
so windows compilation still fails as can be seen [here](https://buildkite.com/julialang/yggdrasil/builds/1199#0186a688-a053-4e05-9bf8-d34fae5d702f).
Is that something that looks familiar?
thanks!

joa-quim:
Nope, first time I see it. I do find that header file in
```
C:\msys64\mingw64\include\tlhelp32.h
```
but in lower case. Case should not matter on Windows (I think). I also see that gcc7 is being used. Can it be that this version of MSYS is too old?

giordano:
MinGW consistently uses lowercase names everywhere, for both header files and libraries.  That's the point of the patch added to this PR (which [I had recommended upstreaming](https://github.com/JuliaPackaging/Yggdrasil/pull/5199#discussion_r928237050), but I guess that hasn't happened).

joa-quim:
Fine, things seam to be advancing (thanks a lot) but when it works there are a couple things more that need to be tuned.
First e want also this cmake setting 
``set (GMT_ENABLE_OPENMP TRUE)``
and then there are the coastlines and country polygons cases. These are separate from the GMT repo and what happens is if they aren't found on the first use GMT downloads them. But this is not a very pleasant surprise as it will take some time to do it. The best is to take case of this at the building/installing step

Have a look at the lines around
https://github.com/GenericMappingTools/gmt/blob/master/ci/build-gmt.sh#L45
which what is done for the CI machinery.

Then I think these are automatically set
```
set (COPY_GSHHG TRUE)
set (COPY_DCW TRUE)
```
but I never know.

joa-quim:
We need to add the ``Ghostscript_jll`` to the list of de dependencies.
```
Dependency(PackageSpec(name="Ghostscript_jll", uuid="61579ee1-b43e-5ca0-a5da-69d92c66a64b"))
```

boriskaus:
Do you need the executable or the dynamic library? Same with gmt: should the dynamic library be shipped as well?

vchuravy:
Generally for a `jll` we will want the dynamic library if not now then eventually.

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,91 @@
...
+make install 
+
+if [[ "${target}" == *-mingw* ]]; then
+    install -Dvm 755 /workspace/destdir/bin/gmt.${dlext} "${libdir}/libgmt.${dlext}"
+    install -Dvm 755 /workspace/destdir/bin/postscriptlight.${dlext} "${libdir}/libpostscriptlight.${dlext}"
+fi
+
+
+# Download GSHHG and DCW from GitHub, which should be shipped as well
+curl -SLO https://github.com/GenericMappingTools/gshhg-gmt/releases/download/${GSHHG_VERSION}/${GSHHG}.${EXT}
```
vchuravy, line 47: Do we need that here? Can we define it as an external Artifact?

If not it should be listed in sources instead of being manually downloaded.

joa-quim:
> Do you need the executable or the dynamic library? Same with gmt: should the dynamic library be shipped as well?

We need all three dynamic libraries that are produced: ``libgmt, libpostscriptlight`` and the ``supplements``. This last one is a plugin library and the normal install step places it under ``lib/gmt/plugins`` on *nix and under ``bin\gmt_plugins`` on Windows.

The ``gmt`` and ``docs`` executables are tinny ones that are handy to have too.

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+cd build/
+
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DHAVE_QSORT_R_GLIBC=False \
+    -DHAVE___BUILTIN_BSWAP16=False \
+    -DHAVE___BUILTIN_BSWAP32=False \
+    -DHAVE___BUILTIN_BSWAP64=False \
+    -DGMT_ENABLE_OPENMP=True \
```
giordano, line 35: If you want openmp, you likely want to do https://github.com/JuliaPackaging/Yggdrasil/blob/2cbe1bf44d3c0c1923657f1c6b2aa911d8c685fa/B/blis/build_tarballs.jl#L135-L138 otherwise by default there's no openmp at all for macos: https://buildkite.com/julialang/yggdrasil/builds/1219#0186abbc-f94c-4018-aa49-ad9c7f93c385/639-1057

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,103 @@
...
+    -DGSHHG_PATH=/workspace/srcdir/${GSSHG} \
+    -DGSHHG_VERSION=${GSHHG_VERSION_numeric} \
+    -DDCW_PATH=/workspace/srcdir/${DCW} \
+    .. 
+make -j${nproc} 
+make install 
+
+if [[ "${target}" == *-mingw* ]]; then
+    install -Dvm 755 /workspace/destdir/bin/gmt.${dlext} "${libdir}/libgmt.${dlext}"
```
giordano, line 52: Please, ***never*** reference `/workspace/destdir` directly, use the most appropriate [environment variable](https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables), otherwise this makes us to reorganise the structure of the filesystem if needed in the future.

giordano: I'm also not sure why you're moving these files in the first place.

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,103 @@
...
+    install -Dvm 755 /workspace/destdir/bin/postscriptlight.${dlext} "${libdir}/libpostscriptlight.${dlext}"
+
+    # note that this removes the *.dll: 
+    install -Dvm 755 /workspace/destdir/bin/gmt_plugins/supplements.* "${libdir}/supplements"
+
+else
+    # supplements is an *.so file on *ix systems; which is installed with (note that this removes the extension):
+    install -Dvm 755 /workspace/destdir/lib/gmt/plugins/supplements.* "${bindir}/supplements"
```
giordano, line 60: This is creating duplicate files in the tarball, maybe you want to _move_ the file instead of copying it?  Also, what file is this?

boriskaus:
ok, this finally compiles now.

Few points:
1. I did not find a binary called `doc`, so that is not added
2. `supplements` is a `*.so` file on linux/mac and a `*.dll` on Windows. I now copied this to `/bin/` without the extension as it did not seem to work with the extension. Not sure whether this will be an issue with `GMT.jl`
3. We now download `dcw` and `gsshg` as external dependency. I don't know whether (and where) this is later downloaded when `GMT_jll` is installed. Perhaps @giordano can comment on this?

giordano:
> supplements is a *.so file on linux/mac and a *.dll on Windows.

What ***are*** those files?  Shared libraries?  Plugins?  Else?

> > I now copied this to /bin/ without the extension as it did not seem to work with the extension. Not sure whether this will be an issue with GMT.jl

Can you elaborate what "did not seem to work" means?

G/GMT/build_tarballs.jl
joa-quim: I can answer to the *what file is this?* GMT is divided in a core *Generic* set of tools that all land in the ``libgmt`` shared lib, and a se of domain specific tools (seismology, potential methods, etc). These last go into a plugin shared lib called ``supplements`` and this one goes to the ``lib/gmt/plugins`` in unix systems and to ``bin\gm_plugins`` on Windows. And there is also the ``libpostscriptlight`` that goes to same place as ``libgmt``. On *normal* installing environments ``cmake`` takes care of putting everything in their right places. Here I don't know if they need to be moved around or if end up landing where they should just automatically.

giordano: If these are plugins dlopened automatically by GMT maybe we don't need to touch them at all?  They seem to be already in the right place, no?

joa-quim:
As I said above (probably was writing at same time as you guys) ``supplements`` is a plugin and if its placed where I said then the ``libgmt`` knows where to find it, otherwise one must create a config file and set a variable pointing to where it lives. Much better to have in its default location.

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,103 @@
...
+        "d4e208dca88fbf42cba1bb440fbd96ea2f932185c86001f327ed0c7b65d27af1")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+
+GSHHG_VERSION_numeric=2.3.7     
+GSSHG="gshhg-gmt-2.3.7"
+DCW="dcw-gmt-2.1.1"
+EXT="tar.gz"
```
giordano, line 29: This is unused?

joa-quim:
> I did not find a binary called doc, so that is not added

Sorry, its ``docs``  not ``doc``. It's used only to make access to documentation (GMT, not GMT.jl) easier from command line.

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,103 @@
...
+    ArchiveSource("https://github.com/GenericMappingTools/dcw-gmt/releases/download/$DCW_VERSION/dcw-gmt-$DCW_VERSION.tar.gz",
+        "d4e208dca88fbf42cba1bb440fbd96ea2f932185c86001f327ed0c7b65d27af1")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+
+GSHHG_VERSION_numeric=2.3.7     
+GSSHG="gshhg-gmt-2.3.7"
+DCW="dcw-gmt-2.1.1"
```
giordano, line 28: If you do something like
```julia
script = """
GSSHG="gshhg-gmt-$(GSHHG_VERSION)"
DCW="dcw-gmt-$(DCW_VERSION)"
""" * raw"""
...
"""
```
you don't need to duplicate version numbers

joa-quim:
I have this for my local WSL build. I don't see one in this build recipe. Maybe it facilitates (that is, avoids it) the moving around files.
``
-DCMAKE_INSTALL_PREFIX=~/programs
``

boriskaus:
> > supplements is a *.so file on linux/mac and a *.dll on Windows.
> 
> What _**are**_ those files? Shared libraries? Plugins? Else?
> 
> > I now copied this to /bin/ without the extension as it did not seem to work with the extension. Not sure whether this will be an issue with GMT.jl
> 
> Can you elaborate what "did not seem to work" means?

It doesn’t seem to be present in the resulting tar file, if I just copy it over. Likely I also just don’t understand what is add to the tar file at the end

boriskaus:
> > I did not find a binary called doc, so that is not added
> 
> Sorry, its `docs` not `doc`. It's used only to make access to documentation (GMT, not GMT.jl) easier from command line.

And where is this binary located after the build is finished?

joa-quim:
Ah Ghrrr, sorry about this noise. There is no ``docs`` executable in this case. The long story is that one may build executables for each module (sym links on Unix) or only the ``gmt`` one, which is what is being done here. So, excuse me and just forget about ``docs``.

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,103 @@
...
+
+    # note that this removes the *.dll: 
+    install -Dvm 755 /workspace/destdir/bin/gmt_plugins/supplements.* "${libdir}/supplements"
+
+else
+    # supplements is an *.so file on *ix systems; which is installed with (note that this removes the extension):
+    install -Dvm 755 /workspace/destdir/lib/gmt/plugins/supplements.* "${bindir}/supplements"
+
+fi
+
```
giordano, line 63: To be clear, my understanding is that this entire block is useless.

joa-quim: In my limited knowledge I tend to agree. The install step should do that automatically. And ``(note that this removes the extension)`` does not sound good to me at all. I think the extension in the ``supplements`` plugin is mandatory.

joa-quim:
I've looked at the artifacts and:
- Windows: has repeated ``gmt.dll`` / ``libgmt.dll`` and ``libpostscriptlight`` / ``postscriptlight``. The non ``lib`` names are what MinGW produces.
- Unix: (LInux & Mac). There is a ``suplements`` file under ``bin`` with no extension that seems to be the ``supplements``` plugin (same size) that should not exist.

Other than this, all seems good. We are almost there. Thanks.

G/GMT/build_tarballs.jl
giordano: This is always unused as far as I can tell

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,90 @@
...
+    Platform("i686", "windows"; ),
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libpostscriptlight", :libpostscriptlight),
+    LibraryProduct("libgmt", :libgmt),
+    ExecutableProduct("gmt", :gmt),
+    FileProduct("supplements", :supplements)
```
giordano, line 70: Does this need to be listed explicitly as product?

giordano (commented):
License file is missing

boriskaus:
> License file is missing
That is something that @joa-quim can hopefully answer; perhaps it needs to be added to the GMT repo

giordano:
> perhaps it needs to be added to the GMT repo 

No, you need to manually install it whenever there's more than a single source: https://docs.binarybuilder.org/stable/build_tips/#Installing-the-license-file

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,90 @@
...
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "windows"; ),
+    Platform("i686", "windows"; ),
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libpostscriptlight", :libpostscriptlight),
+    LibraryProduct("libgmt", :libgmt),
```
giordano, line 68:
```suggestion
    LibraryProduct(["libgmt", "gmt"], :libgmt),
```

G/GMT/build_tarballs.jl
giordano: And this keeps being unused
```suggestion
```

joa-quim:
The repo has the license in the root dir https://github.com/GenericMappingTools/gmt/blob/master/LICENSE.TXT and from what I saw in other artifacts this needs to be copied into ``share/licenses/GMT``

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,89 @@
...
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "windows"; ),
+    Platform("i686", "windows"; ),
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libpostscriptlight", :libpostscriptlight),
```
giordano, line 66: Same here. When products have different basenames on different platforms you have to list them. Please read the documentation: https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct

boriskaus:
ok, seems that it now compiles fine on all systems.  @joa-quim: it would be great if you can check whether the resulting tarballs are useable by `GMT.jl`, or whether other things need to be added.

joa-quim:
Hmm, something's not right. The MinGW artifact links against a ``libgdal-31.dll`` (as I mentioned above) but looking into my installed ``GDAL_jll``, I see
```
    JLLWrappers.@init_library_product(
        libgdal,
        "bin\\libgdal-32.dll",
        RTLD_LAZY | RTLD_DEEPBIND,
    )
```
and a ``libgdal-32.dll`` is what I have in the ``artifacts`` dir. This will cause fatal problems for sure.

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+    LibraryProduct(["libpostscriptlight", "postscriptlight"], :libpostscriptlight),
+    LibraryProduct(["libgmt", "gmt"], :libgmt),
+    ExecutableProduct("gmt", :gmt),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="LibCURL_jll", uuid="deac9b47-8bc7-5906-a0fe-35ac56dc84c0"); compat="7.73.0")
+    Dependency("NetCDF_jll", compat="400.702.402")
+    Dependency(PackageSpec(name="GDAL_jll", uuid="a7073274-a066-55f0-b90d-d619367d196c"))
```
giordano, line 79: As per https://github.com/JuliaPackaging/Yggdrasil/pull/5199#issuecomment-1456778189, it's probably a good idea to add compat bounds for this package.

boriskaus:
@joa-quim : I think the easiest way to test this is if you install this locally on your machine. According to the documentation, For a windows machine, this can be done with
```
julia build_tarballs.jl --verbose --debug --deploy=local x86_64-w64-mingw32
```
which will install it in `~/.julia/dev/GMT_jll`.

giordano:
I don't think BinaryBuilder works on Windows.

boriskaus:
> I don't think BinaryBuilder works on Windows.

ok, that makes this difficult. There is probably no alternative to merging this and trying in that case, not?

giordano:
If you give me some time (couple of days maybe) I can publish a build that can be tested on Windows.

joa-quim:
I think it will work but could not make it that my path trickery worked. It did work for running from the Windows ``cmd`` but not from Julia. The main problem seems to be the dependency names. For example my local HDF5 dll is called ``libhdf5-310.dll`` but the one in ``HDF5_JLL`` is ``libhdf5-0.dll`` so I had to make a mixture of paths from mu ``c:\msys\...`` and ``....julia/artifacts/...``, which failed to load the ``gmt.dll``. Tried the ``DependencyWalker`` but no way because I do not reach the point of precompiling GMT.jl

boriskaus:
> Tried to do it via WSL but one way or the other I'm always hitting this error
> 
> ```
> 08357d84c092c9d6a8bad715001ad33e3d5d5/opt/x86_64-linux-musl/GCCBootstrap-12.1.0
> [21:00:06] At line 498, ABORTED (16: Resource busy)!
> ERROR: LoadError: Build for GMT on aarch64-apple-darwin did not complete successfully
> ```

But why are you trying to build this for Apple instead of windows?

joa-quim:
I know the error is that but I didn't tell it to. Just did (and variations)
```
julia build_tarballs.jl --deploy=local --verbose x86_64-w64-mingw32
```
or
from the Julia REPL, ``include("build_tarballs.jl")``  (and that's what gave the error I posted above)

giordano:
```
]add https://github.com/giordano/GMT_jll.jl
```

joa-quim:
Thanks, but there seems to be a problem with lapack. This is Julia 1.9rc but same error with 1.7
```
julia> import Pkg; Pkg.precompile()
Precompiling environment...
  ? GMT_jll
  0 dependencies successfully precompiled in 6 seconds. 476 already precompiled.

ERROR: The following 1 direct dependency failed to precompile:

GMT_jll [b68b8c3f-ed99-5bef-9675-4739d9426b26]

Failed to precompile GMT_jll [b68b8c3f-ed99-5bef-9675-4739d9426b26] to "C:\\Users\\joaqu\\.julia\\compiled\\v1.9\\GMT_jll\\jl_5A7C.tmp".
ERROR: LoadError: InitError: could not load library "C:\Users\joaqu\.julia\artifacts\f572524225d8a9266c31227f4e28a0a1091bd882\bin\liblapack.dll"The specified module could not be found.



julia> using LAPACK32_jll
ERROR: InitError: could not load library "C:\Users\joaqu\.julia\artifacts\f572524225d8a9266c31227f4e28a0a1091bd882\bin\liblapack.dll"
The specified module could not be found.
```

giordano:
Right, that's https://github.com/JuliaLang/julia/issues/47638#issuecomment-1407002852

joa-quim:
Yep, it's looking for a 
image
but I have only ``libblastrampoline-5.dll`` in Julia 1.9 & 1.10 and ``libblastrampoline.dll`` in Julia 1.7

joa-quim:
``lapack`` is not a mandatory dependency in GMT. It's only used in a few modules that have a (slower) fallback equations solver in case lapack is not available. So, at least for now (and testing) ``GMT_JLL`` can be built without that dependency.

joa-quim:
Yes, I know. And on Windows the dlls have embedded their own names and that' why we cannot simple rename them and expect it to work.

giordano:
No one is expecting anything to magically work, that has to be fixed, no one has done it.

joa-quim:
Sure. Not blaming anyone.

joa-quim:
I found the ``libblastrampoline-5-0-2.dll`` file [here](https://fossies.org/windows/misc/julia-1.8.5-win64.zip/index_o.html), downloaded and tried (again with some quick patchy solution) but got an error saying it's not able to load the ``gmt.dll`` due to missing symbol (don't have the right error message anymore). Then I tried the ``DependencyWalker`` and found things like this

image
Does that red on ``libhdf5_hd-0.dll`` means it might be a cause for the ``gmt.dll`` failure? There is another ``red`` in ``libsz.dll``

boriskaus:
To get back to this discussion, it seems to me that at least the `gmt` binary works just fine on windows as shown below. 

1. Version;
```julia
julia> versioninfo()
Julia Version 1.8.3
Commit 0434deb161 (2022-11-14 20:14 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 8 × 12th Gen Intel(R) Core(TM) i3-1215U
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, goldmont)
  Threads: 1 on 8 virtual cores
```

3. Installing the GMT_jll from @giordano 
````julia
PS C:\Users\boris\.julia\dev\GMT_jll> julia
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.3 (2022-11-14)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(@v1.8) pkg> dev https://github.com/giordano/GMT_jll.jl
````

4. Running it;
```julia
PS C:\Users\boris\.julia\dev\GMT_jll> julia
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.3 (2022-11-14)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(@v1.8) pkg> activate .
  Activating project at `C:\Users\boris\.julia\dev\GMT_jll`

julia> using GMT_jll

julia> run(GMT_jll.gmt())

        GMT - The Generic Mapping Tools, Version 6.5.0_19413d4_2023.03.03 [64-bit] [MP] [8 cores]
        (c) 1991-2023 The GMT Team (https://www.generic-mapping-tools.org/team.html).

        Supported in part by the US National Science Foundation (http://www.nsf.gov/)
        and volunteers from around the world.

        GMT is distributed under the GNU LGPL License (http://www.gnu.org/licenses/lgpl.html).
        Dependencies: netCDF, GDAL, PCRE, FFTW, LAPACK, ZLIB, Ghostscript, GraphicsMagick, FFmpeg.

usage: gmt [options]
       gmt <module name> [<module-options>]

options:
  --help              List descriptions of available GMT modules.
  --new-script[=L]    Write GMT modern mode script template to standard output.
                      Optionally specify bash|csh|batch [Default is current shell].
  --new-glue=name     Write C code for external supplements to glue them to GMT.
  --show-bindir       Show directory with GMT executables.
  --show-citation     Show the most recent citation for GMT.
  --show-classic      Show all classic module names.
  --show-classic-core Show all classic module names (core only).
  --show-cores        Show number of available cores.
  --show-datadir      Show directory/ies with user data.
  --show-dataserver   Show URL of the remote GMT data server.
  --show-dcw          Show the DCW data version used.
  --show-doi          Show the DOI for the current release.
  --show-gshhg        Show the GSHHG data version used.
  --show-library      Show path of the shared GMT library.
  --show-modules      Show all modern module names.
  --show-modules-core Show all modern module names (core only).
  --show-plugindir    Show directory for plug-ins.
  --show-sharedir     Show directory for shared GMT resources.
  --show-userdir      Show full path of user's ~/.gmt dir
  --version           Print GMT version number.

if <module-options> is '=' we call exit (0) if module exist and non-zero otherwise.

Process(setenv(`'C:\Users\boris\.julia\artifacts\b00f4bcc0d779dfd70def088d9a629941a6e23b1\bin\gmt.exe'`,["PATH=C:\\Users\\boris\\AppData\\Local\\Programs\\Julia-1.8.3\\bin;C:\\Users\\boris\\.julia\\artifacts\\0235fabbecfa4e38a3c8e87491b9d4f3e7fe9848\\bin;C:\\Users\\boris\\.julia\\artifacts\\701aaf00cce776fdefff2d993e09800fab61e884\\bin;C:\\Users\\boris\\.julia\\artifacts\\a5b7c195243758d88c64cd8547e527fbace3877b\\bin;C:\\Users\\boris\\.julia\\artifacts\\431cad9bf95e817b5fc813666e353ae36b2202c0\\bin;C:\\Users\\boris\\.julia\\artifacts\\4543a35904e5905f4c387199cc080e27cc202cec\\bin;C:\\Users\\boris\\.julia\\artifacts\\c024ee55cc5bbe84dd4411a68b02dd9a3797fdcd\\bin;C:\\Users\\boris\\.julia\\artifacts\\4ad4b701a165bfb1dec449667cffd85d0a16dcc2\\bin;C:\\Users\\boris\\.julia\\artifacts\\ac8a1c48a02b397f6ce1beb6473de8bbb3b1e0ce\\bin;C:\\Users\\boris\\.julia\\artifacts\\2e9a29903fddf79371cd5f82327f70ada670a234\\bin;C:\\Users\\boris\\.julia\\artifacts\\17025344af7bdfe8a7ac4e8887f0953f1dc50376\\bin;C:\\Users\\boris\\.julia\\artifacts\\01b969d96551292e78f0acfa204f120b09f4e7b4\\bin;C:\\Users\\boris\\.julia\\artifacts\\f2d9050c5dba019ebf272fe7da44f78c5491672d\\bin;C:\\Users\\boris\\.julia\\artifacts\\bcdd265e21bd7089238f661440be60aed9e9e696\\bin;C:\\Users\\boris\\.julia\\artifacts\\511fc12b661678285fd116ffcee88cbaf043b869\\bin;C:\\Users\\boris\\.julia\\artifacts\\72037f1415327f12277eed0521c861517630a0ba\\bin;C:\\Users\\boris\\.julia\\artifacts\\a845328b7acbc4442b321885d186869a90297089\\bin;C:\\Users\\boris\\.julia\\artifacts\\41eb25ea50f277f03154d52354602687fb1abc84\\bin;C:\\Users\\boris\\.julia\\artifacts\\18a38c3afb63ea0df1832d3f946b21b58ba401ea\\bin;C:\\Users\\boris\\.julia\\artifacts\\5d74056071181523cbb62efeddd86132d4db12de\\bin;C:\\Users\\boris\\.julia\\artifacts\\db3d60d39813bf806d1b4a8032e3a2f892c75738\\bin;C:\\Users\\boris\\.julia\\artifacts\\b62e4d49f0fa9e8d9daccda7af8ed7315580d6dc\\bin;C:\\Users\\boris\\.julia\\artifacts\\9c9d5d21081d79fb66bb9aaeea87960a22c5599b\\bin;C:\\Users\\boris\\.julia\\artifacts\\3628616b5a5a8fd8628079ae4fb0d8948fdb37bc\\bin;C:\\Users\\boris\\.julia\\artifacts\\a214c5171a32bff0583c70a08027d44e8ef22cc2\\bin;C:\\Users\\boris\\.julia\\artifacts\\dcd0721a2a49cfb8916395419701035173b57955\\bin;C:\\Users\\boris\\.julia\\artifacts\\83d62d00715073512a2e76bbb95e027bf701a4e3\\bin;C:\\Users\\boris\\.julia\\artifacts\\b7dd1809d0626eac3bf6f97ba8ccfbb6cc63c509\\bin;C:\\Users\\boris\\.julia\\artifacts\\5e6cdd29fca31b08026313af46c1db0c724ebbb2\\bin;C:\\Users\\boris\\.julia\\artifacts\\f572524225d8a9266c31227f4e28a0a1091bd882\\bin;C:\\Users\\boris\\.julia\\artifacts\\a97c68e84f8d803b96763aa872ed49381fb9cb7e\\bin;C:\\Users\\boris\\.julia\\artifacts\\5af028f5ee6fc5d6923b4769eca4bc199914e6bd\\bin;C:\\Users\\boris\\.julia\\artifacts\\38d4f07d7bebeb1d35473280d1ef56f7d8e6991d\\bin;C:\\Users\\boris\\.julia\\artifacts\\e5872943f1078a6397a4f4d0f760db959ec50630\\bin;C:\\Users\\boris\\.julia\\artifacts\\f9f799e6ae3fe7f244bb55052db7ff1cea876c27\\bin;C:\\Users\\boris\\.julia\\artifacts\\4f67e82ee394e74ca63b7dc589dcc9218c12b20a\\bin;C:\\Users\\boris\\.julia\\artifacts\\6a5456c156251deff41ae40ac8dfbb3d8c808ffe\\bin;C:\\Users\\boris\\.julia\\artifacts\\c916a560fa109a4db3d6938a8ab1a6a8b4a09731\\bin;C:\\Users\\boris\\.julia\\artifacts\\3aff78028a0bb2ad40ebc3ae7e465a52dcbe2730\\bin;C:\\Users\\boris\\.julia\\artifacts\\ee9d999d28081333cf8ec10830fd77b75359226b\\bin;C:\\Users\\boris\\.julia\\artifacts\\2ae55b726ee1f4063216b19a1ffb59d3d5b0a6eb\\bin;C:\\Users\\boris\\.julia\\artifacts\\017f69ccc5ddfc0405e5bf8326a572300498e06c\\bin;C:\\Users\\boris\\.julia\\artifacts\\1b573c0aaf4a3a73f296a61c2cacbee17ac0d4e3\\bin;C:\\Users\\boris\\.julia\\artifacts\\7d57f5f863fd09f31bbff6d08d53bfc07f49c3e6\\bin;C:\\Users\\boris\\.julia\\artifacts\\6f5e7c4e3249d2073c64e6beaa3abcd7ae4fa9bc\\bin;C:\\Users\\boris\\.julia\\artifacts\\0c393ed4faa8cdf04f992ff605abaed1c1172f69\\bin;C:\\Users\\boris\\.julia\\artifacts\\160ce27d1af09edffbb7d9f1833f6d3bdefad491\\bin;C:\\Users\\boris\\.julia\\artifacts\\c6e48d3563556547ad4cc0021c2cffa066a079df\\bin;C:\\Users\\boris\\.julia\\artifacts\\da2e0ac84abbe712490f8ee73aca3d512f1eaaab\\bin;C:\\Users\\boris\\.julia\\artifacts\\5b9eb98a95e718b9ddc16d01fd70b19736b13e0b\\bin;C:\\Users\\boris\\.julia\\artifacts\\e2bc28a38e71d07d41a27f8d4b3132c5d1b5a22d\\bin;C:\\Users\\boris\\.julia\\artifacts\\cca2b0693e0ae8df2daca37d0779bb7209f32c5f\\bin;C:\\Users\\boris\\.julia\\artifacts\\b00f4bcc0d779dfd70def088d9a629941a6e23b1\\bin;C:\\Users\\boris\\AppData\\Local\\Programs\\Julia-1.8.3\\bin\\..\\lib\\julia;C:\\Users\\boris\\AppData\\Local\\Programs\\Julia-1.8.3\\bin\\..\\lib;C:\\Users\\boris\\AppData\\Local\\Programs\\Julia-1.8.3\\bin;C:\\WINDOWS\\system32;C:\\WINDOWS;C:\\WINDOWS\\System32\\Wbem;C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0\\;C:\\WINDOWS\\System32\\OpenSSH\\;C:\\Program Files\\dotnet\\;C:\\Program Files\\Git\\cmd;C:\\Users\\boris\\AppData\\Local\\Microsoft\\WindowsApps;C:\\Users\\boris\\AppData\\Local\\Programs\\Microsoft VS Code\\bin;C:\\Users\\boris\\AppData\\Local\\Programs\\Julia-1.8.3\\bin;C:\\ProgramData\\boris\\GitHubDesktop\\bin", "USERDOMAIN_ROAMINGPROFILE=DESKTOP-488U587", "HOMEPATH=\\Users\\boris", "VSCODE_INJECTION=1", "PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL", "SESSIONNAME=Console", "SYSTEMROOT=C:\\WINDOWS", "APPDATA=C:\\Users\\boris\\AppData\\Roaming", "PSMODULEPATH=C:\\Users\\boris\\Documents\\WindowsPowerShell\\Modules;C:\\Program Files\\WindowsPowerShell\\Modules;C:\\WINDOWS\\system32\\WindowsPowerShell\\v1.0\\Modules", "COMMONPROGRAMW6432=C:\\Program Files\\Common Files"  …  "PROGRAMFILES(X86)=C:\\Program Files (x86)", "VSCODE_GIT_ASKPASS_MAIN=c:\\Users\\boris\\AppData\\Local\\Programs\\Microsoft VS Code\\resources\\app\\extensions\\git\\dist\\askpass-main.js", "PROGRAMFILES=C:\\Program Files", "CHROME_CRASHPAD_PIPE_NAME=\\\\.\\pipe\\LOCAL\\crashpad_30280_YPJHJPUITZAYFZRU", "LOGONSERVER=\\\\DESKTOP-488U587", "DRIVERDATA=C:\\Windows\\System32\\Drivers\\DriverData", "ORIGINAL_XDG_CURRENT_DESKTOP=undefined", "FONTCONFIG_PATH=C:\\Users\\boris\\.julia\\artifacts\\c916a560fa109a4db3d6938a8ab1a6a8b4a09731\\etc\\fonts", "SYSTEMDRIVE=C:", "PROCESSOR_ARCHITECTURE=AMD64"]), ProcessExited(0))
```

So I think this can be merged.

boriskaus:
On my machine the above command works fine:
```julia
julia> using LibCURL_jll, NetCDF_jll, GDAL_jll, FFTW_jll, PCRE_jll, LAPACK32_jll, OpenBLAS32_jll, FFMPEG_jll, Glib_jll,  Ghostscript_jll, CompilerSupportLibraries_jll, LLVMOpenMP_jll, GMT_jll
julia> 
```
perhaps there is something else on your machine that interferes?

joa-quim:
And was it with the ``libblastrampoline-5-0-2.dll`` downloaded from the site I linked above or without it?

Anyway, I just noticed that we need to add the ``PROJ_jll`` dependency too (I access to its shared library directly).
And do you know how to run the gmt executable with options? If it worked would be  ``run(GMT_jll.gmt("--user-dir"))``. I need to capture the output of a couple of those options.

boriskaus:
No I am not using any additionally installed packages, but rather the ones that were used to compile `GMT_jll`, to ensure that all the packages are compatible. 

I´m using windows 11 for this along with the `GMT_jll` build that was helpfully provided by @giordano 
Here the steps to reproduce what I did:

1. Download a development version of GMT
```julia
 julia> ]dev https://github.com/giordano/GMT_jll.jl
   Cloning git-repo `https://github.com/giordano/GMT_jll.jl`
   Resolving package versions...
  No Changes to `C:\Users\boris\.julia\environments\v1.8\Project.toml`
  No Changes to `C:\Users\boris\.julia\environments\v1.8\Manifest.toml`
```
2. Change to the directory
```julia
 julia> ; 
shell> cd .julia\\dev\\GMT_jll\\
C:\Users\boris\.julia\dev\GMT_jll
```
3. Activate project 
```julia
 julia> ]
(@v1.8) pkg> activate .
  Activating project at `C:\Users\boris\.julia\dev\GMT_jll`
```
4. Precompile
```julia
(GMT_jll) pkg> precompile
    Updating registry at `C:\Users\boris\.julia\registries\General.toml`
    Updating `C:\Users\boris\.julia\dev\GMT_jll\Project.toml`
  [692b3bcd] + JLLWrappers v1.4.1
  [b22a6f82] + FFMPEG_jll v4.4.2+2
  [f5851436] + FFTW_jll v3.3.10+0
  [a7073274] + GDAL_jll v301.600.200+0
  [61579ee1] + Ghostscript_jll v9.55.0+4
  [7746bdde] + Glib_jll v2.74.0+2
  [17f450c3] + LAPACK32_jll v3.10.0+0
  [1d63c593] + LLVMOpenMP_jll v15.0.4+0
  [7243133f] + NetCDF_jll v400.902.5+1
⌅ [656ef2d0] + OpenBLAS32_jll v0.3.17+0
  [2f80f16e] + PCRE_jll v8.44.0+0
  [e66e0078] ~ CompilerSupportLibraries_jll ⇒ v0.5.2+0
  [deac9b47] ~ LibCURL_jll ⇒ v7.84.0+0
    Updating `C:\Users\boris\.julia\dev\GMT_jll\Manifest.toml`
  [692b3bcd] + JLLWrappers v1.4.1
  [21216c6a] + Preferences v1.3.0
  [8ce61222] + Arrow_jll v10.0.0+1
  [6e34b625] + Bzip2_jll v1.0.8+0
  [83423d85] + Cairo_jll v1.16.1+1
  [2e619515] + Expat_jll v2.4.8+0
  [b22a6f82] + FFMPEG_jll v4.4.2+2
  [f5851436] + FFTW_jll v3.3.10+0
  [a3f928ae] + Fontconfig_jll v2.13.93+0
  [d7e528f0] + FreeType2_jll v2.10.4+0
  [559328eb] + FriBidi_jll v1.0.10+0
  [a7073274] + GDAL_jll v301.600.200+0
  [d604d12d] + GEOS_jll v3.11.0+0
  [78b55507] + Gettext_jll v0.21.0+0
  [61579ee1] + Ghostscript_jll v9.55.0+4
  [7746bdde] + Glib_jll v2.74.0+2
  [3b182d85] + Graphite2_jll v1.3.14+0
  [0234f1f7] + HDF5_jll v1.12.2+2
  [2e76f6c2] + HarfBuzz_jll v2.8.1+1
  [aacddb02] + JpegTurbo_jll v2.1.91+0
  [b39eb1a6] + Kerberos_krb5_jll v1.19.3+0
  [c1c5ebd0] + LAME_jll v3.100.1+0
  [17f450c3] + LAPACK32_jll v3.10.0+0
  [88015f11] + LERC_jll v3.0.0+1
  [1d63c593] + LLVMOpenMP_jll v15.0.4+0
  [dd4b983a] + LZO_jll v2.10.1+0
  [08be9ffa] + LibPQ_jll v14.3.0+1
⌅ [e9f186c6] + Libffi_jll v3.2.2+1
  [d4300ac3] + Libgcrypt_jll v1.8.7+0
  [7add5ba3] + Libgpg_error_jll v1.42.0+0
  [94ce4f54] + Libiconv_jll v1.16.1+2
  [4b2f31a3] + Libmount_jll v2.35.0+0
  [89763e89] + Libtiff_jll v4.4.0+0
  [38a345b3] + Libuuid_jll v2.36.0+0
  [d3a379c0] + LittleCMS_jll v2.12.0+0
  [5ced341a] + Lz4_jll v1.9.3+0
  [7243133f] + NetCDF_jll v400.902.5+1
  [e7412a2a] + Ogg_jll v1.3.5+1
⌅ [656ef2d0] + OpenBLAS32_jll v0.3.17+0
  [643b3616] + OpenJpeg_jll v2.4.0+0
  [458c3c95] + OpenSSL_jll v1.1.20+0
  [91d4177d] + Opus_jll v1.3.2+0
  [2f80f16e] + PCRE_jll v8.44.0+0
  [58948b4f] + PROJ_jll v900.100.0+0
  [30392449] + Pixman_jll v0.40.1+0
  [76ed43ae] + SQLite_jll v3.41.0+0
  [e0b8ae26] + Thrift_jll v0.16.0+0
  [02c8fc9c] + XML2_jll v2.10.3+0
  [aed1982a] + XSLT_jll v1.1.34+0
  [4f6342f7] + Xorg_libX11_jll v1.6.9+4
  [0c0b7dd1] + Xorg_libXau_jll v1.0.9+4
  [a3789734] + Xorg_libXdmcp_jll v1.1.3+4
  [1082639a] + Xorg_libXext_jll v1.3.4+4
  [9a3f8284] + Random
  [ea8e919c] + SHA v0.7.0
  [9e88b42a] + Serialization
  [6462fe0b] + Sockets
  [fa267f1f] + TOML v1.0.0
  [a4e569a6] + Tar v1.10.1
  [cf7118a7] + UUIDs
  [4ec0a83e] + Unicode
  [e66e0078] + CompilerSupportLibraries_jll v0.5.2+0
  [deac9b47] + LibCURL_jll v7.84.0+0
  [29816b5a] + LibSSH2_jll v1.10.2+0
  [c8ffd9c3] + MbedTLS_jll v2.28.0+0
  [14a3606d] + MozillaCACerts_jll v2022.2.1
  [4536629a] + OpenBLAS_jll v0.3.20+0
  [efcefdf7] + PCRE2_jll v10.40.0+0
  [83775a58] + Zlib_jll v1.2.12+3
  [8e850b90] + libblastrampoline_jll v5.1.1+0
  [8e850ede] + nghttp2_jll v1.48.0+0
  [3f19e933] + p7zip_jll v17.4.0+0
        Info Packages marked with ⌅ have new versions available but compatibility constraints restrict them from upgrading. To see why use `status --outdated -m`
Precompiling project...
  1 dependency successfully precompiled in 2 seconds. 75 already precompiled.
```

5. check that the libraries can be loaded
```julia
julia> using LibCURL_jll, NetCDF_jll, GDAL_jll, FFTW_jll, PCRE_jll, LAPACK32_jll, OpenBLAS32_jll, FFMPEG_jll, Glib_jll,  Ghostscript_jll, CompilerSupportLibraries_jll, LLVMOpenMP_jll, GMT_jll
```

6. Run GMT:
```julia
julia> run(`$(GMT_jll.gmt())`)

        GMT - The Generic Mapping Tools, Version 6.5.0_19413d4_2023.03.03 [64-bit] [MP] [8 cores]
        (c) 1991-2023 The GMT Team (https://www.generic-mapping-tools.org/team.html).

        Supported in part by the US National Science Foundation (http://www.nsf.gov/)
        and volunteers from around the world.

        GMT is distributed under the GNU LGPL License (http://www.gnu.org/licenses/lgpl.html).
        Dependencies: netCDF, GDAL, PCRE, FFTW, LAPACK, ZLIB, Ghostscript, GraphicsMagick, FFmpeg.

usage: gmt [options]
       gmt <module name> [<module-options>]

options:
  --help              List descriptions of available GMT modules.
  --new-script[=L]    Write GMT modern mode script template to standard output.
                      Optionally specify bash|csh|batch [Default is current shell].
  --new-glue=name     Write C code for external supplements to glue them to GMT.
  --show-bindir       Show directory with GMT executables.
  --show-citation     Show the most recent citation for GMT.
  --show-classic      Show all classic module names.
  --show-classic-core Show all classic module names (core only).
  --show-cores        Show number of available cores.
  --show-datadir      Show directory/ies with user data.
  --show-dataserver   Show URL of the remote GMT data server.
  --show-dcw          Show the DCW data version used.
  --show-doi          Show the DOI for the current release.
  --show-gshhg        Show the GSHHG data version used.
  --show-library      Show path of the shared GMT library.-show-citation     Show the most recent citation for GMT.
  --show-classic      Show all classic module names.
  --show-classic-core Show all classic module names (core only).
  --show-cores        Show number of available cores.
  --show-datadir      Show directory/ies with user data.
  --show-dataserver   Show URL of the remote GMT data server.
  --show-dcw          Show the DCW data version used.
  --show-doi          Show the DOI for the current release.
  --show-gshhg        Show the GSHHG data version used.
  --show-library      Show path of the shared GMT library.
  --show-modules      Show all modern module names.
  --show-modules-core Show all modern module names (core only).
  --show-plugindir    Show directory for plug-ins.
  --show-sharedir     Show directory for shared GMT resources.
  --show-userdir      Show full path of user's ~/.gmt dir
  --version           Print GMT version number.

if <module-options> is '=' we call exit (0) if module exist and non-zero otherwise.
```
  --show-modules      Show all modern module names.
  --show-modules-core Show all modern module names (core only).
  --show-plugindir    Show directory for plug-ins.
  --show-sharedir     Show directory for shared GMT resources.
  --show-userdir      Show full path of user's ~/.gmt dir
  --version           Print GMT version number.

if <module-options> is '=' we call exit (0) if module exist and non-zero otherwise.
```

6. Run GMT with command line options:
```julia
julia> args="--version"
"--version"
julia> run(`$(GMT_jll.gmt()) $(args)`);
6.5.0_19413d4_2023.03.03
```

boriskaus:
> And do you know how to run the gmt executable with options? If it worked would be `run(GMT_jll.gmt("--user-dir"))`. I need to capture the output of a couple of those options.

The option `--user-dir` does not do anything and is also not listed in the help of `gmt`. There is, however, the following option:
```julia
julia> args="--show-userdir"
"--show-userdir"

julia> run(`$(GMT_jll.gmt()) $(args)`);
C:/Users/boris/.gmt
```

joa-quim:
Good, thanks. So that must mean that the ``libblastrampoline-5-0-2.dll`` problem was solved upstream.

If you can load the ``GMT_jll`` branch, you should be able to actually run ``GMT.jl``. It works but I already found one test that fails with the MinGW binaries but works with the MSVC ones.
```
julia> plot(rand(8,2), bg="-circ", Vd=2)
makecpt [ERROR]: Option -G: Unable to parse 0.25
```
vs
```
julia> plot(rand(8,2), bg="-circ", Vd=2)
2-element Vector{String}:
 "grdimage -D  -JX14c/9.5c -P -K > C:\\TEMP\\GMTjl_tmp.ps"
 "psxy  -JX14c/9.5c -Baf -BWSen -R0/0.88/0.06/0.86 -K -O >> C:\\TEMP\\GMTjl_tmp.ps"
```

boriskaus:
I suppose I need to indicate in `GMT.jl` that I want to use the `GMT_jll` version instead of the current way of doing that. 

If you can tell me exactly how to set that up and reproduce the plot command above I’m happy to try.

joa-quim:
I think I saw that you can do ``] add GMT.jl@GMT_jll`` or ``] add GMT.jl#GMT_jll``

Then it would only be to run the command I posted. But there are more and this doesn't look good. The errors come from the binaries themselves and to investigate them I need to debug bur for one side I would need to build them myself locally and for another I certainly do not want to use ``dbg``. For me debugging is only with Visual Studio debugger.

```
julia> gmt("grdview -R-15/15/-15/15/-1/1 -JX6i/0 -Baf -Bza -n+a -p120/30 -Qs -JZ5 -Ctopo -I45 -P -K > lixo.ps", G)
grdview [ERROR]: Option -I: Must specify intensity file, value, or modifiers
ERROR: Something went wrong when calling the module. GMT error number = 72
```

boriskaus:
yes, I can confirm that the tests above don´t work. It is encouraging, though, that quite a few of the tests do work.
Nevertheless I think it is easier to debug this once `GMT_jll` is registered at least once, as you can locally override the libraries with a local BB build.

joa-quim:
The big shit is that, when running from CLI my MinGW build passes all tests of its equivalent MSVC build. But that proves little because the code paths followed by the CLI and wrapped versions (Julia, Matlab, Python) are not the same. It should be relatively easy if I could debug the ``gcc`` (MinGW) build, but again no way that I'll submit to the ``gdb`` path pain.

But I agree that this could be merged though not yet accepted as a Conda binaries replacement. Unless there is no problems with the Linux and MacOS binaries. Yet, before that the ``PROJ_jll`` dependency must be added.

joa-quim:
One thing if you know about that. I see that the dependencies have
```
    Dependency("NetCDF_jll", compat="400.902.5")
    Dependency(PackageSpec(name="GDAL_jll", uuid="a7073274-a066-55f0-b90d-d619367d196c"); compat="301.600.200")
```
Not exactly sure how to convert those *compat* numbers to programs versions. I take it that it means
- netcdf 4.9.x
- gdal 3.6.x (what is that 301?)

but there is an elephant in the room which is the HDF5 versionupon which they linked againt. HDF5 is **very**, **very** picky with their versions. One cannot have a ``netCDF`` linked against 1.12.1 that is used to link in ``GDAL`` that itself was linked against 1.12.2. How are these things handled by BinaryBuilder?

joa-quim:
There are good news though. My Linux (Ubuntu 20.04) build in Codespaces does not fail in those cases (though my WSL with Ubuntu 22.04 still errors with libcurl). But it errors 
```
        GMTLOGO
psconvert [ERROR]: Cannot execute Ghostscript (gs).
GMT: Error During Test at /home/codespace/.julia/packages/GMT/p9huK/test/runtests.jl:5
  Got exception outside of a @test
```
clearly the ``gs`` executable is not being found, but ``Ghostscript_jll`` is a dependency.

boriskaus:
> One thing if you know about that. I see that the dependencies have
> 
> ```
>     Dependency("NetCDF_jll", compat="400.902.5")
>     Dependency(PackageSpec(name="GDAL_jll", uuid="a7073274-a066-55f0-b90d-d619367d196c"); compat="301.600.200")
> ```
> 
> Not exactly sure how to convert those _compat_ numbers to programs versions. I take it that it means
> 
> * netcdf 4.9.x
> * gdal 3.6.x (what is that 301?)
> 
> but there is an elephant in the room which is the HDF5 versionupon which they linked againt. HDF5 is **very**, **very** picky with their versions. One cannot have a `netCDF` linked against 1.12.1 that is used to link in `GDAL` that itself was linked against 1.12.2. How are these things handled by BinaryBuilder?

You can see that in the build recipes for [NetCDF](https://github.com/JuliaPackaging/Yggdrasil/blob/master/N/NetCDF/build_tarballs.jl):
```
upstream_version = v"4.9.0"

# Offset to add to the version number.  Remember to always bump this.
version_offset = v"0.2.5"

```
I believe this is done because the BB version always need to increase by one if you add new options or dependencies. Therefore it may quickly become out of whack with the real version number, so it is multiplied by 100 & incremented accordingly.

boriskaus:
> clearly the `gs` executable is not being found, but `Ghostscript_jll` is a dependency.

according to the [build recipe](https://github.com/JuliaPackaging/Yggdrasil/blob/master/G/Ghostscript/build_tarballs.jl) of  `Ghostscript_jll`, it should create a binary called `gs`:
```Julia
  ExecutableProduct("gs", :gs),
```

Check whether that is available as `Ghostscript_jll.gs()`

joa-quim:
OK, I have this working now on Linux but with a few strange errors on the Codespaces (where some test) fail but all pass in my WSL installation. Clearly we need more testing bur for that we also need this merged.

Note, on Windows the JLL is not used because of other strange errors that I'll need to find a way to debug them.

Boris, could you please rebuild the JLL using a more recent GMT commit? Several issues in GMT were fixed/improved since the last build.

boriskaus:
>Boris, could you please rebuild the JLL using a more recent GMT commit? Several issues in GMT were fixed/improved since the last build.

done (using the latest build of GMT that passed the CI tests).

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,93 @@
...
+# Bash recipe for building across all platforms
+script = """
+GSSHG_VERSION="$(GSHHG_VERSION)"
+GSSHG="gshhg-gmt-$(GSHHG_VERSION)"
+DCW="dcw-gmt-$(DCW_VERSION)"
+""" * raw"""
+
+
+cd $WORKSPACE/srcdir
+cd gmt
```
giordano, line 31:
```suggestion
cd ${WORKSPACE}/srcdir/gmt
```

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,93 @@
...
+    -DGSHHG_VERSION=${GSHHG_VERSION_numeric} \
+    -DDCW_PATH=/workspace/srcdir/${DCW} \
+    .. 
+make -j${nproc} 
+make install 
+
+# copy license
+mkdir ${prefix}/share/licenses/GMT
+cp ../LICENSE.TXT ${prefix}/share/licenses/GMT/LICENSE.TXT
+
```
giordano, line 53:
```suggestion
install_license ../LICENSE.TXT
```

G/GMT/build_tarballs.jl
```diff
@@ -0,0 +1,93 @@
...
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DHAVE_QSORT_R_GLIBC=False \
+    -DHAVE___BUILTIN_BSWAP16=False \
+    -DHAVE___BUILTIN_BSWAP32=False \
+    -DHAVE___BUILTIN_BSWAP64=False \
+    -DGMT_ENABLE_OPENMP=True \
+    -DGSHHG_PATH=/workspace/srcdir/${GSSHG} \
+    -DGSHHG_VERSION=${GSHHG_VERSION_numeric} \
+    -DDCW_PATH=/workspace/srcdir/${DCW} \
```
giordano, line 45:
```suggestion
    -DGSHHG_PATH=${WORKSPACE}/srcdir/${GSSHG} \
    -DGSHHG_VERSION=${GSHHG_VERSION_numeric} \
    -DDCW_PATH=${WORKSPACE}/srcdir/${DCW} \
```

PR #5206, opened 2022-07-22
[GDAL] Bump GEOS Dependency to 3.11

nraynaud:


giordano:
You have to change the version number, for example by incrementing the offset to `0.0.1`

PR #5213, opened 2022-07-24
gRPC linux build

rory-linehan:


G/gRPC/build_tarballs.jl
```diff
@@ -0,0 +1,72 @@
...
+popd
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform(
+      "x86_64", "linux";
+      libc="glibc",
+      libgfortran_version=v"5.0.0",
```
rory-linehan, line 47: I'm not sure if these options are too restrictive, but this is what I needed for what I'm working on. Additionally, there were many compiler errors for other platforms, so maybe this is fine.

G/gRPC/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+version = v"1.48.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/grpc/grpc.git", "d2054ec6c6e8abcecf0e24b0b4ee75035d80c3cc")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/grpc && \
```
giordano, line 15: No need to use `&& \` everywhere, this only makes the code harder to read

G/gRPC/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+git submodule update --init --recursive && \
+mkdir -p cmake/build && \
+pushd "cmake/build" && \
+cmake \
+    -DgRPC_INSTALL=ON \
+    -DgRPC_BUILD_TESTS=OFF \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DgRPC_SSL_PROVIDER=package \
+    ../.. && \
```
giordano, line 24: This is not passing the cmake toolchain file?

G/gRPC/build_tarballs.jl
giordano: Yes, this looks very restrictive, especially setting `libstdcxx_version`, we never use that.

G/gRPC/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+platforms = [
+    Platform(
+      "x86_64", "linux";
+      libc="glibc",
+      libgfortran_version=v"5.0.0",
+      libstdcxx_version = v"3.4.29",
+      cxxstring_abi = "cxx11"
+    )
+]
+platforms = expand_gfortran_versions(platforms)
```
giordano, line 52: This isn't doing much, since you're fixing the libgfortran version

G/gRPC/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+    ExecutableProduct("grpc_php_plugin", :grpc_php_plugin)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("OpenSSL_jll")
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; lock_microarchitecture=false)
```
giordano, line 71: Why `lock_microarchitecture=false`? :weary: Also, going forward we want to require julia v1.6, with `julia_compat="1.6"`.

rory-linehan: I'm not sure why that's there, probably a copypaste leftover. I'm removing it and testing locally right now.

G/gRPC/build_tarballs.jl
rory-linehan: I'm doing further testing locally with this restriction removed

G/gRPC/build_tarballs.jl
rory-linehan: This configuration was done because of this: https://github.com/grpc/grpc/blob/master/BUILDING.md#cross-compiling 

and was lifted from here: https://github.com/grpc/grpc/blob/master/test/distrib/cpp/run_distrib_test_cmake_aarch64_cross.sh

giordano: But we have [protoc](https://github.com/JuliaBinaryWrappers/protoc_jll.jl) already.  Also, the tarballs contains zlib, which it shouldn't, because again we have it already.

rory-linehan: Ahh, Zlib and protoc are fair, I've added those in. However, the build breaks when I add the toolchain flag to this bootstrap section.

G/gRPC/build_tarballs.jl
```diff
@@ -0,0 +1,72 @@
...
+mkdir -p cmake/build
+pushd "cmake/build"
+cmake \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DgRPC_BUILD_TESTS=OFF \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DgRPC_PROTOBUF_PROVIDER=package \
+    -DgRPC_ZLIB_PROVIDER=package \
+    -DABSL_PROPAGATE_CXX_STD=ON \
+    ../..
```
giordano, line 39: You're passing the toolchain file: https://docs.binarybuilder.org/stable/build_tips/#CMake-builds

G/gRPC/build_tarballs.jl
```diff
@@ -0,0 +1,72 @@
...
+    ExecutableProduct("grpc_python_plugin", :grpc_python_plugin),
+    ExecutableProduct("grpc_ruby_plugin", :grpc_ruby_plugin),
+    ExecutableProduct("grpc_node_plugin", :grpc_node_plugin),
+    ExecutableProduct("grpc_php_plugin", :grpc_php_plugin)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency("Zlib_jll"),
+    BuildDependency("protoc_jll"),
```
giordano, line 67: Are you sure `protoc` is a runtime dependency?  I think we usually use it as _**host**_ _build_ dependency: https://github.com/JuliaPackaging/Yggdrasil/blob/cc4e6f619655102d406e5bcc80323303e0b1ab98/O/ONNXRuntime/ONNXRuntime/build_tarballs.jl#L67.  Please read the documentation about the difference: https://docs.binarybuilder.org/stable/build_tips/#Dependencies-for-the-target-system-vs-host-system

giordano (commented):
Can we avoid building, or at least bundling, all the static libraries?

ViralBShah:
Do we want to merge this? Or close?

giordano:
Certainly not merge if comments above aren't addressed

rory-linehan:
This can be closed

PR #5234, opened 2022-07-26
WIP: [pthread_win32] init

sjkelly:
This adds [pthread-win32](https://github.com/GerHobbelt/pthread-win32). The install step is currently broken, and will need to be fixed...


maleadt:
Rebased and fixed; builds locally.

maleadt:
Should probably rename this to `pthreads_win32`, note the trailing `s`. An alternative name used in the upstream repo is `pthreads4w`. Or, maybe we should just name this `pthreads`, in the case we'd also want to provide `libpthreads.so` for other platforms?

maleadt:
The DLL generated here also doesn't contain a single `pthread_` API call, so something is wrong here.

vchuravy:
Don't we build pthreads for out mingw cross compiler?

maleadt:
It should be included in a mingw distribution, yes, but which JLL contains that?

giordano:
> It should be included in a mingw distribution, yes, but which JLL contains that?

None, it's in the compiler shard for windows.

fingolfin:
Mingw contains winpthreads and this is indeed installed in our compiler shards.

The last comment here was three years ago, and the need of this is unclear - close it?

PR #5238, opened 2022-07-27
Wizard recipe: nng-v1.0.0

simacleod:

* Package name: nng
* Version: v1.0.0


N/nng/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libnng", :libnng, "usr/local/lib")
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    Dependency(PackageSpec(name="Ninja_jll", uuid="76642167-d241-5cee-8c94-7a494e8cb7b7"))
```
giordano, line 54: Why Ninja is a runtime dependency at all?
```suggestion
```

N/nng/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+cd $WORKSPACE/srcdir
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+cd nng
+mkdir build
+cd build
+cmake -G Ninja ..
+ninja
+export DESTDIR=$prefix && ninja install
```
giordano, line 25: Please read the documentation: https://docs.binarybuilder.org/stable/build_tips/#CMake-builds.  You must pass the toolchain file.  Also, no need to patch the CMake file, you can set variables from the command line
```suggestion
cd $WORKSPACE/srcdir/nng
mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release ..
make -j${nproc}
make install
```

N/nng/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "freebsd"; )
+]
```
giordano, line 43:
```suggestion
platforms = filter!(p -> Sys.islinux(p) || Sys.isfreebsd(p), supported_platforms())
```
What's the problem with the other platforms?

N/nng/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libnng", :libnng, "usr/local/lib")
```
giordano, line 48: This is was likely being installed in the wrong directory because you weren't passing the proper arguments to CMake as described in the documentation
```suggestion
    LibraryProduct("libnng", :libnng)
```

N/nng/bundled/patches/enable_shared_libs.patch
```diff
@@ -0,0 +1,13 @@
+diff --git before/nng/CMakeLists.txt after/nng/CMakeLists.txt
```
giordano, line 1: This patch is not needed

N/nng/build_tarballs.jl
simacleod: Ah yes, these seem like the correct rocks to smash together :) Thanks for the help!

N/nng/build_tarballs.jl
simacleod: Not sure, this was the wizard output --  Let's find out!

giordano: I know that's the output of the wizard, in the general case it can't figure out the most efficient predicate to filter the platforms appropriately.

N/nng/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=1 ..
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = filter!(p -> Sys.islinux(p) || Sys.isfreebsd(p), supported_platforms())
```
giordano, line 25: Maybe we can try to build for all platforms and try to understand what's the problem on the other platforms?
```suggestion
platforms = supported_platforms()
```

simacleod: All platforms passing, my issue was probably related to not having the cmake toolchain option set.

PR #5240, opened 2022-07-27
Add support for MParT v0.2

dannys4:


dannys4:
Unsure what the `Linked library libjulia.so.1 (resolved path /usr/local/lib/libjulia.so.1) is not within the given prefix` warning is about, but I don't think that's an issue.

M/MParT/build_tarballs.jl
```diff
@@ -2,16 +2,27 @@
...
 sources = [
-    GitSource("https://github.com/MeasureTransport/MParT.git", "64e4b8ee7637382ab7f9aa091ace41ed6a5cbcc6")
+    GitSource("https://github.com/MeasureTransport/MParT.git", "7e138bc9981f4181db0388a2a481155a45f0eafe")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
+# Override compiler ID to silence the horrible "No features found" cmake error
+if [[ $target == *"apple-darwin"* ]]; then
+  macos_extra_flags="-DCMAKE_CXX_COMPILER_ID=AppleClang -DCMAKE_CXX_COMPILER_VERSION=10.0.0 -DCMAKE_CXX_STANDARD_COMPUTED_DEFAULT=11"
```
giordano, line 23: I don't know what horrible error you're talking about, but this sounds like a lie since we do _not_ use AppleClang

M/MParT/build_tarballs.jl
```diff
@@ -21,28 +32,38 @@ fi
 
 cmake -DCMAKE_INSTALL_PREFIX=$prefix \
   -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+  -DCMAKE_CXX_FLAGS="-I${prefix}/include/eigen3 -fopenmp" \
```
giordano, line 35:
```suggestion
  -DCMAKE_CXX_FLAGS="-I${includedir}/eigen3 -fopenmp" \
```

dannys4: I'll make this change, but if this is how we should be doing it, should the [this documentation](https://docs.binarybuilder.org/stable/troubleshooting/#Header-files-of-the-dependencies-can't-be-found) be updated accordingly? I'd be happy to put up a quick PR, I just wasn't aware of this variable (and I'm not sure why I would be).

M/MParT/build_tarballs.jl
dannys4: Sorry if this is wrong, but there are a few examples using cxxwrap on Yggdrasil that use this exact thing (jlqml, sdpa, spot_julia), and I wasn't sure if I needed to include it so i erred on the side of caution. I'll double check if it's needed ASAP.

M/MParT/build_tarballs.jl
giordano: Yes please, that was probably before the `${includedir}` variable was introduced.

M/MParT/build_tarballs.jl
dannys4: looks like it isn't needed, at least according to Azure.

PR #5249, opened 2022-07-28
[HDF5] add LibCURL_jll compat

visr:
Should fix

```
dlopen(/Users/runner/.julia/artifacts/4e3d68673330116b7ef55a34b6d6d405e005fe6d/lib/libhdf5.200.dylib, 1): Library not loaded: @rpath/libcurl.4.dylib
  Referenced from: /Users/runner/.julia/artifacts/4e3d68673330116b7ef55a34b6d6d405e005fe6d/lib/libhdf5.200.dylib
  Reason: Incompatible library version: libhdf5.200.dylib requires version 13.0.0 or later, but libcurl.4.dylib provides version 12.0.0
```

From https://github.com/JuliaIO/HDF5.jl/pull/996

Version 7.73 is the version that julia 1.6 uses:
https://github.com/JuliaLang/julia/blob/v1.6.6/stdlib/LibCURL_jll/Project.toml#L3

Just like #4981

### DIFF
--- a/H/HDF5/build_tarballs.jl
+++ b/H/HDF5/build_tarballs.jl
@@ -76,7 +76,7 @@ products = [
 dependencies = [
     Dependency("Zlib_jll"),
     Dependency("OpenSSL_jll"; compat="1.1.10"),
-    Dependency("LibCURL_jll"),
+    Dependency("LibCURL_jll"; compat="7.73"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.


giordano:
We don't build HDF5, so I don't think this will solve anything.  They're just using a too new libcurl to build hdf5,  sigh

visr:
Oof yes that's unfortunate. Not sure how to proceed. Perhaps better to yank HDF5_jll 1.12.2 quickly?

Would be nice if we could somehow obtain conda builds like hdf5-feedstock but with libcurl set to 7.73.
https://github.com/conda-forge/hdf5-feedstock/blob/main/.ci_support/osx_64_mpinompiopenssl1.1.1.yaml#L19-L20

Not sure how to get them though. I don't think we can ask the conda-forge folks to fix libcurl just for this.

giordano:
> Perhaps better to yank HDF5_jll 1.12.2 quickly?

Yeah

visr:
Closing this, I summarized in https://github.com/JuliaPackaging/Yggdrasil/issues/4511#issuecomment-1198134988.

PR #5253, opened 2022-07-30
Build GLFW with Wayland support

VarLad:
This is a try at building GLFW (latest release version) with Wayland support. 
Here's the Fedora spec for interested people https://src.fedoraproject.org/rpms/glfw/blob/rawhide/f/glfw.spec


VarLad:
This is the repo where I'll try building for git version of GLFW. 
This repo is not for merging.

PR #5257, opened 2022-07-30
Update xkbcommon to 1.4.1

VarLad:


X/xkbcommon/build_tarballs.jl
```diff
@@ -3,36 +3,28 @@
...
-    -Dnative-wayland-scanner="/usr/bin/wayland-scanner"
+    -Denable-docs=false
 ninja -j${nproc}
 ninja install
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = [p for p in supported_platforms() if Sys.islinux(p)]
+platforms = filter!(p -> arch(p) != "armv6l" && Sys.islinux(p), supported_platforms())
```
staticfloat, line 27: What's the problem with `armv6l` here?

VarLad: From what I remember, we don't have the dependencies on those either.

Issue #4969, opened 2022-06-01
The SLICOT_jll-v5.8 fails to link with Julia 1.8.0


andreasvarga:
Recently the SLICOT library has been updated to v5.8 and the corresponding SLICOT_jll has been generated by courtesy of Ralph Smith. I uploaded the package and it works fine locally with Julia 1.7.2. However, running some CI tests, I got complaints of version incompatibility (see below) (e.g., on Julia ^1.8.0-0 - ubuntu-latest - x64 )

```
 
Run julia --color=yes --project=@. -e 'using Pkg; if VERSION >= v"1.1.0-rc1"; Pkg.build(verbose=true); else Pkg.build(); end'
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
ERROR: Unsatisfiable requirements detected for package libblastrampoline_jll [8e850b90]:
 libblastrampoline_jll [8e850b90] log:
 ├─possible versions are: 5.1.0 or uninstalled
 └─found to have no compatible versions left with SLICOT_jll [545525a2]
```
I am not sure what this means, but looking to compatibilities of SLICOT_jll, there is
`libblastrampoline_jll = "3.0.4"`
I wonder if this covers all later versions (e.g., 5.1.0 mentioned above). 

Note that with with Julia 1.7.3 the tests run correctly!

giordano:
Someone should do something like #4770 (for lack of better options at the moment)

RalphAS:
Since this wraps a Fortran library we also have the proliferation of gfortran versions.  Is there any way to avoid combinatorial explosion?

giordano:
Not really, sadly.

andreasvarga:
A question: As I remember, with SLICOT_jll for v5.7, I executed all tests with both Julia 1.7 and 1.8 without problems. What is essentially different now with the version for v5.8?

giordano:
> As I remember, with SLICOT_jll for v5.7, I executed all tests with both Julia 1.7 and 1.8 without problems.

I doubt it, as `SLICOT_jll` `v5.7` was already linking to libblastrampoline.

andreasvarga:
I posted 4 (four) releases of [PeriodicSystems](https://github.com/andreasvarga/PeriodicSystems.jl). 
[...]

PR #5259, opened 2022-08-01
SLICOT 5.8.0, Julia 1.8 compat

mkitti:
> Someone should do something like #4770 (for lack of better options at the moment)

_Originally posted by @giordano in https://github.com/JuliaPackaging/Yggdrasil/issues/4969#issuecomment-1146836975_

I'm "someone".

https://discourse.julialang.org/t/too-many-julia-versions/84538/15?u=mkitti

1. Do we need a version bump to SLICOT 5.8.1 to do this?
2. @andreasvarga can we coordinate a SLICOT 5.8.1 release?

cc: @RalphAS

### DIFF
--- a/S/SLICOT/build_tarballs.jl
+++ b/S/SLICOT/build_tarballs.jl
@@ -4,7 +4,7 @@ using BinaryBuilder, Pkg
 using BinaryBuilderBase: get_addable_spec
 
 name = "SLICOT"
-version = v"5.8.1"
+version = v"5.8.0"
 
 # Collection of sources required to complete build
 # Note to maintainers: extracts from LAPACK are deprecated routines, so probably don't want
--- a/S/SLICOT/build_tarballs.jl
+++ b/S/SLICOT/build_tarballs.jl
@@ -100,7 +100,9 @@ platforms = expand_gfortran_versions(supported_platforms())
 # https://github.com/JuliaPackaging/Yggdrasil/pull/4770
 # Since we need to link to libblastrampoline which has seen multiple
 # ABI-incompatible versions, we need to expand the julia versions we target
-julia_versions = [v"1.7.0", v"1.8.0", v"1.9.0"]
+#julia_versions = [v"1.7.0", v"1.8.0", v"1.9.0"]
+# Focus on Julia v"1.8.0" during debug
+julia_versions = [v"1.8.0"]
 function set_julia_version(platforms::Vector{Platform}, julia_version::VersionNumber)
     _platforms = deepcopy(platforms)
     for p in _platforms
@@ -119,8 +121,9 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(get_addable_spec("libblastrampoline_jll", v"3.0.4+0"); platforms=filter(p -> VersionNumber(p["julia_version"]) == v"1.7.0", platforms)),
-    Dependency(get_addable_spec("libblastrampoline_jll", v"5.1.0+0"); platforms=filter(p -> VersionNumber(p["julia_version"]) >= v"1.8.0", platforms)),
+    # Focus on 1.8 during debug
+    #Dependency(get_addable_spec("libblastrampoline_jll", v"3.0.4+0"); platforms=filter(p -> VersionNumber(p["julia_version"]) == v"1.7.0", platforms)),
+    Dependency(get_addable_spec("libblastrampoline_jll", v"5.1.1+1"); platforms=filter(p -> VersionNumber(p["julia_version"]) >= v"1.8.0", platforms)),
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
 ]
 
--- a/S/SLICOT/build_tarballs.jl
+++ b/S/SLICOT/build_tarballs.jl
@@ -1,9 +1,10 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
+using BinaryBuilderBase: get_addable_spec
 
 name = "SLICOT"
-version = v"5.8.0"
+version = v"5.8.1"
 
 # Collection of sources required to complete build
 # Note to maintainers: extracts from LAPACK are deprecated routines, so probably don't want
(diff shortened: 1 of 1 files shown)


andreasvarga:
OK. So once again my reply tp your proposal.

The best for SLICOT integration would be of course to have compatibility starting with the LTS version, as well as with the current and future versions. I appreciate very much your help in this respect. If necessary, I could produce a version bump to 5.8.1 for SLICOT itself, just to keep the right correspondence between the versions. Is this OK for you?

mkitti:
I'm not sure if compatibility with Julia 1.6 is possible at this point since I think the version mechanism depends on new functionality, but I would be happy to corrected.

andreasvarga:
This was only a wishful thinking of me (not important for further developments).

RalphAS:
Thanks for picking this up.  You might have a new record for the number of tarballs.

Julia v1.6 would need a different approach because libblastrampoline was only partly integrated.  It could be difficult to keep the versioning in order with the two builds in parallel.

RalphAS:
Now I've remembered why I didn't do this earlier: the builds should really be sorted by version of libblastrampoline rather than version of Julia, and I didn't figure out a simple way to do that.

andreasvarga:
Sorry for my ignorance, but what follows now? In the meantime, I have troubles even registering under Julia v1.7 (see [#68456](https://github.com/JuliaRegistries/General/issues/68456)).

giordano:
This is failing to build.  If someone could debug this, that'd be much appreciated.

RalphAS:
The linker is failing to find libblastrampoline.  It seems to be looking in /workspace/destdir/lib which it gets as the shell variable `${libdir}`.  Has there been a relevant recent change in BinaryBuilder?

giordano:
No

mkitti:
It is only failing to find libblastrampoline on certain platform triples.

mkitti:
The Julia 1.7 builds succeed. Thr Julia 1.8 and Julia 1.9 builds fail.

andreasvarga:
For Julia 1.7 v3.0.4 of blastrampoline is used, which is OK. For Julia 1.8 and 1.9, version v5.1.0 is used and not the last available version v5.1.1. I wonder if this is a problem. Moreover, for 1.8 and 1.9, version v3.0.4 is also mentioned in the script. I can not figure out what effect this may have.

mkitti:
OK, I will try that.

Also do you have any plans to release a SLICOT 5.8.1. This will be released as 5.8.1 and it would be nice to keep the versions synchronized.

andreasvarga:
I observed that the last call to generate the binaries contains a dependency to julia 1.7. Is this a lower limit on the version or it must be also adapted to various julia versions ?

andreasvarga:
I propose to keep the version of SLICOT as it is now.

mkitti:
At the moment, I believe we still need to bump the version number here in this packaging for Julia 1.7 compat at least.

mkitti:
> I observed that the last call to generate the binaries contains a dependency to julia 1.7. Is this a lower limit on the version or it must be also adapted to various julia versions ?

`julia_compat=v"1.7"` indicates compatibility with Julia versions `[1.7, 2.0)` per https://pkgdocs.julialang.org/v1/compatibility/#Version-specifier-format

giordano:
I'm not entirely sure we need to bump the version number: that's necessary when changing the generated project file (new dependencies, or julia compat), which isn't the case here?  There was a libblastrampoline compat, but that was kinda redundant since libblastrampoline is a stdlib, bound to julia anyway.

mkitti:
[b9a67f0](https://github.com/JuliaPackaging/Yggdrasil/pull/5259/commits/b9a67f032dbdee5abd57a5fcbbab0e865b80cb8f) looks good for 1.8 compat.
* Expanding again to 1.7 and 1.9
* Shoved this package version back to 5.8.0

Let's see if this still works.

mkitti:
CI is green. Version is 5.8.0. Let's go?

S/SLICOT/build_tarballs.jl
```diff
@@ -103,7 +119,8 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"); compat="3.0.4"),
+    Dependency(get_addable_spec("libblastrampoline_jll", v"3.0.4+0"); platforms=filter(p -> VersionNumber(p["julia_version"]) == v"1.7.0", platforms)),
```
mkitti, line 122: Hmm, this line doesn't seem right.

mkitti: Maybe it's fine? What values does `p["julia_version"]` cover?

giordano: https://github.com/JuliaPackaging/Yggdrasil/blob/5a363ee835394b70dcc268fe958f5c8e5cee912b/S/SLICOT/build_tarballs.jl#L103

andreasvarga:
Congratulations and many thanks!

RalphAS:
Don't celebrate too soon; the package doesn't seem to be consistent yet: https://github.com/JuliaRegistries/General/pull/68512
says

> ERROR: Unsatisfiable requirements detected for package libblastrampoline_jll [8e850b90]:
 libblastrampoline_jll [8e850b90] log:
 ├─possible versions are: 5.1.1 or uninstalled
 └─found to have no compatible versions left with SLICOT_jll [545525a2]
   └─SLICOT_jll [545525a2] log:
     ├─possible versions are: 5.7.0-5.8.0 or uninstalled
     └─restricted to versions 5.8.0 by an explicit requirement, leaving only versions 5.8.0

mkitti:
Actually, I'm pretty sure the current issue is due to the previous compat bound for libblastrampoline_jll.

https://github.com/JuliaRegistries/General/pull/68512/files does not attempt to change Compat.toml. So the existing Compat.toml still applies:
https://github.com/JuliaRegistries/General/blob/master/S/SLICOT_jll/Compat.toml

andreasvarga:
This is a similar problem for which I opened an [issue](https://github.com/JuliaRegistries/General/issues/68456) when trying to register a new version of [PeriodicSystems](https://github.com/andreasvarga/PeriodicSystems.jl). This issue has been closed, with the conclusion that 

> AutoMerge always runs with the latest stable Julia, which is currently Julia 1.8. If your package does not support Julia 1.8, it will fail AutoMerge, and you will need to request a manual merge.

However, in our case we are at a lower level (registering the library itself), so I assume if we manage to solve this issue we will solve also [#68456](https://github.com/JuliaRegistries/General/issues/68456) . 

_Just a remark:_ The PR example used as tempate for producing SLICOT_jll was for a package/library `Pfapack`, which apparently was never registered in Julia (only a Python version exists)!

mkitti:
https://github.com/JuliaRegistries/General/pull/68512 has been merged.

Now we need to turn our attention to [PeriodicSystems](https://github.com/andreasvarga/PeriodicSystems.jl) we have compat issues over there now.

mkitti:
https://juliahub.com/ui/Packages/Pfapack_jll/HOmkl/0.1.0+1?page=1 Pfapack_jll is registered by the way.

andreasvarga:
I formulated an [issue](https://github.com/andreasvarga/PeriodicSystems.jl/issues/13) regarding the failure of [PeriodicSystems](https://github.com/andreasvarga/PeriodicSystems.jl) release 0.4.2 on Windows with Julia 1.8 (with Julia 1.7 it works, as well as on Linux). I can easily reproduce the failure locally by using any call involving the wrappers I implemented for several SLICOT routines. Each call crashes Julia (so I can not see the actual error), but from the [CI.log](https://github.com/andreasvarga/PeriodicSystems.jl/actions/runs/3088971504/jobs/4996050657), the error is basically 

```
Exception: EXCEPTION_ACCESS_VIOLATION at 0x0 -- unknown function (ip: 0000000000000000)
in expression starting at D:\a\PeriodicSystems.jl\PeriodicSystems.jl\test\test_pschur.jl:15
unknown function (ip: 0000000000000000)
Allocations: 107208134 (Pool: 107190352; Big: 17782); GC: 43

```

I guess, something is wrong with the generated SLICOT_jll for Windows under Julia 1.8. 

Very gratefull for any idea how to fix this issue.

PR #5264, opened 2022-08-02
[SoapyPlutoSDR] bump build

sjkelly:
This is the latest git build, using libiio 0.24. 

This line breaks the Musl builds, but that should be fine for now: https://github.com/pothosware/SoapyPlutoSDR/blob/a07c37230369653818b3a5c448c00cee1ac9f8e5/CMakeLists.txt#L77

We drop libad9361_iio for now, since it is only used for filters and unsupported on latest libiio.

### DIFF
--- a/S/SoapyPlutoSDR/build_tarballs.jl
+++ b/S/SoapyPlutoSDR/build_tarballs.jl
@@ -33,7 +33,7 @@ fi
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = filter!(p -> libc(p) != "musl", supported_platforms())
+platforms = supported_platforms()
 platforms = expand_cxxstring_abis(platforms) # requested by auditor
 
 # The products that we will ensure are always built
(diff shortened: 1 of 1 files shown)


S/SoapyPlutoSDR/build_tarballs.jl
```diff
@@ -7,46 +7,33 @@ version = v"0.2.1" # not yet tagged
...
-    make -j${nproc}
-    make install
     mv ${libdir}/SoapySDR/modules0.8/libPlutoSDRSupport.so  ${libdir}/SoapySDR/modules0.8/libPlutoSDRSupport.dylib
 fi
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(;experimental=true)
+platforms = supported_platforms()
```
staticfloat, line 36: Why did you take away `experimental=true`?

sjkelly: It was required for aarch64-apple-darwin, but it is now included in the supported_platforms. I guess keeping it is okay? It currently does not add any platforms at the moment.

PR #5266, opened 2022-08-03
Update libjlnode for node v16

sunoru:


L/libjlnode/build_tarballs.jl
```diff
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "libjlnode"
-version = v"16.14.0"
+version = v"16.0.0"
```
staticfloat, line 6: I think this is going to build a new JLL that has an _older_ version number than the last one.  Is that what you want?

sunoru: The last registered one is 14.17.3 so 16.0.0 is not older :) It is want I intended to do

staticfloat: Ah, I was confused because I was looking at only one commit, not the whole thing. :)

sunoru:
Thanks!

PR #5269, opened 2022-08-03
[quarto] Add quarto

jeremiahpslewis:
Currently blocked (I believe) by the following:

1. Deno needs to support musl builds: https://github.com/denoland/deno/issues/3711
2. We need to rebuild Deno_jll with `x86_64-linux-musl-cxx11` platform
3. Add Deno as `HostBuildDependency`
4. Figure out how to cross-compile quarto to different platform targets (no idea if this is actually possible at the moment)


Q/quarto/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+    GitSource("https://github.com/quarto-dev/quarto-cli.git", "6a42361a388433ad8b52ab71759a50e778b48ba0"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/quarto-cli
+source configuration
+export QUARTO_BIN_DIR=/workspace/destdir/bin
+export QUARTO_PACKAGE_DIR=/workspace/destdir/bin
+/workspace/destdir/bin/deno run --unstable --allow-env --allow-read --allow-write --allow-run --allow-net --allow-ffi --importmap=/workspace/srcdir/quarto-cli/src/dev_import_map.json /workspace/srcdir/quarto-cli/package/src/bld.ts $@
```
giordano, line 19: I'm pretty sure I said this several times already: use environment variables where appropriate, e.g. `${bindir}.  But also, if you have to run something in `${bindir}` that's likely wrong.

Q/quarto/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("quarto", :quarto)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("Deno_jll")
```
giordano, line 39: https://docs.binarybuilder.org/stable/build_tips/#Dependencies-for-the-target-system-vs-host-system

aviks:
For something like this, I'm wondering if it would be better to use the pre-compiled binaries from upstream. That's the route I took for *pandoc*, for example: https://github.com/JuliaPackaging/Yggdrasil/blob/master/P/pandoc/build_tarballs.jl

jeremiahpslewis:
@aviks Good idea, do you know whether pandoc will be able to support M1 / Apple Silicon this way? It looks like homebrew uses Haskell to generate an M1 binary, don't see one being provided directly from pandoc. https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/pandoc.rb

aviks:
Yeah, pandoc does not provide an arm/mac binary download. Presumably they don't have the hardware to build that. You could try asking them?

giordano:
Superseded by #5522

PR #5280, opened 2022-08-06
[CUTEst] Add M1 platform

amontoison:
[26/09/2022]
Mosè, you can merge the PR.

### DIFF
--- a/C/CUTEst/build_tarballs.jl
+++ b/C/CUTEst/build_tarballs.jl
@@ -136,4 +136,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"8.1.0", julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
--- a/C/CUTEst/build_tarballs.jl
+++ b/C/CUTEst/build_tarballs.jl
@@ -96,9 +96,9 @@ if [[ "${target}" == *-apple-* ]]; then
   extra="-Wl,-undefined -Wl,dynamic_lookup -headerpad_max_install_names"
 fi
 cd $CUTEST/objects/$MYARCH/single
-gfortran -fPIC -shared ${extra} -Wl,${all_load} libcutest.a -Wl,${noall_load} -o ${libdir}/libcutest_single.${dlext}
+gfortran -fPIC -shared ${extra} $(flagon -Wl,--whole-archive) libcutest.a $(flagon -Wl,--no-whole-archive) -o ${libdir}/libcutest_single.${dlext}
 cd $CUTEST/objects/$MYARCH/double
-gfortran -fPIC -shared ${extra} -Wl,${all_load} libcutest.a -Wl,${noall_load} -o ${libdir}/libcutest_double.${dlext}
+gfortran -fPIC -shared ${extra} $(flagon -Wl,--whole-archive) libcutest.a $(flagon -Wl,--no-whole-archive) -o ${libdir}/libcutest_double.${dlext}
 
 ln -s $ARCHDEFS/bin/helper_functions ${bindir}/
 ln -s $SIFDECODE/bin/sifdecoder ${bindir}/
--- a/C/CUTEst/build_tarballs.jl
+++ b/C/CUTEst/build_tarballs.jl
@@ -48,7 +48,7 @@ cd $SIFDECODE
 if [[ "${target}" == *-linux* || "${target}" == *-freebsd* ]]; then
   echo "6" > sifdecode.opts   # PC64
   echo "2" >> sifdecode.opts  # Linux
-  echo "6" >> sifdecode.opts  # gfortran
+  echo "8" >> sifdecode.opts  # gfortran
 elif [[ "${target}" == *-apple* ]]; then
   echo "13" > sifdecode.opts  # macOS
   echo "2" >> sifdecode.opts  # gfortran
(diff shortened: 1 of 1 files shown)


C/CUTEst/build_tarballs.jl
```diff
@@ -104,14 +119,21 @@ platforms = expand_gfortran_versions(filter!(!Sys.iswindows, supported_platforms
 
 # The products that we will ensure are always built
 products = [
-    ExecutableProduct("sifdecoder", :sifdecoder),
+    FileProduct("lib/libsifdecode.a", :libsifdecode_a),
+    FileProduct("lib/libcutest_single.a", :libcutest_single_a),
+    FileProduct("lib/libcutest_double.a", :libcutest_double_a),
```
dpo, line 124: I don't think we need to export those.

amontoison: I would like to compile GALAHAD for @tmigot and it could be easier to link with static libraries before we use the "trick" to generate dynamic libraries.

C/CUTEst/build_tarballs.jl
```diff
@@ -81,14 +91,14 @@ all_load="--whole-archive"
...
+  noall_load="-noall_load"
+  extra="-Wl,-undefined -Wl,dynamic_lookup -headerpad_max_install_names"
 fi
-cd $CUTEST/objects/$MYARCH/double
-gfortran -fPIC -shared ${extra} -Wl,${all_load} libcutest.a -Wl,${noall_load} -o libcutest_double.${dlext}
 cd $CUTEST/objects/$MYARCH/single
-gfortran -fPIC -shared ${extra} -Wl,${all_load} libcutest.a -Wl,${noall_load} -o libcutest_single.${dlext}
+gfortran -fPIC -shared ${extra} -Wl,${all_load} libcutest.a -Wl,${noall_load} -o ${libdir}/libcutest_single.${dlext}
+cd $CUTEST/objects/$MYARCH/double
+gfortran -fPIC -shared ${extra} -Wl,${all_load} libcutest.a -Wl,${noall_load} -o ${libdir}/libcutest_double.${dlext}
```
giordano, line 101:
```suggestion
gfortran -fPIC -shared ${extra} $(flagon -Wl,--whole-archive) libcutest.a $(flagon -Wl,--no-whole-archive) -o ${libdir}/libcutest_single.${dlext}
cd $CUTEST/objects/$MYARCH/double
gfortran -fPIC -shared ${extra} $(flagon -Wl,--whole-archive) libcutest.a $(flagon -Wl,--no-whole-archive) -o ${libdir}/libcutest_double.${dlext}
```
and remove all the settings of `all_load` above.

C/CUTEst/build_tarballs.jl
```diff
@@ -1,38 +1,50 @@
...
+# SIFDecode always looks for `ar`, `ranlib`, `sed` and `grep` in `/usr/bin/`
+ln -sf $(which ar) /usr/bin/ar
+ln -sf $(which ranlib) /usr/bin/ranlib
+ln -sf $(which sed) /usr/bin/sed
+ln -sf $(which grep) /usr/bin/grep
+
+if [[ "${target}" == *-apple* ]]; then
+  mkdir -p /opt/homebrew/bin
+  ln -sf $(which gfortran) /opt/homebrew/bin/gfortran
+fi
```
giordano, line 44: Waat

C/CUTEst/build_tarballs.jl
```diff
@@ -1,38 +1,50 @@
...
-cp -r ARCHDefs-2.0.4x ${prefix}/libexec/
-cp -r SIFDecode-2.0.3 ${prefix}/libexec/
-cp -r CUTEst-2.0.3 ${prefix}/libexec/
-export ARCHDEFS=${prefix}/libexec/ARCHDefs-2.0.4x
-export SIFDECODE=${prefix}/libexec/SIFDecode-2.0.3
-export CUTEST=${prefix}/libexec/CUTEst-2.0.3
+
+cp -r ARCHDefs ${prefix}
+cp -r SIFDecode ${prefix}
+cp -r CUTEst ${prefix}
```
giordano, line 25: These directories end up in the final tarball (including their `.git/` subdirectories, which looks rather weird).  Is this really what you want

amontoison: Yes, it's how @dpo uses this artifact for CUTEst.jl.
https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/main/src/CUTEst.jl#L77-L79

C/CUTEst/build_tarballs.jl
amontoison: CUTEst doesn't have a `Makefile`. We can only install it with a shell script and they suppose that gfortran is only provided by homebrew.

dpo: None of this was necessary and is in fact harmful. I created release 2.0.4x of ARCHDefs specifically to cross compile. It shouldn't be updated.

C/CUTEst/build_tarballs.jl
```diff
@@ -101,17 +107,20 @@ install_license $CUTEST/lgpl-3.0.txt
...
 platforms = expand_gfortran_versions(filter!(!Sys.iswindows, supported_platforms()))
+platforms = filter!(p -> !(os(p) == "freebsd" && libgfortran_version(p) == v"3"), platforms)
 
 # The products that we will ensure are always built
 products = [
     ExecutableProduct("sifdecoder", :sifdecoder),
     ExecutableProduct("slct", :slct),
     ExecutableProduct("clsf", :clsf),
+    # LibraryProduct("libcutest_single", :libcutest_single),
+    # LibraryProduct("libcutest_double", :libcutest_double),
```
dpo, line 118: We don't want to export those. They are incomplete (by "design").

PR #5282, opened 2022-08-06
[Spot_julia] add more julia versions

MaximeBouton:
I would like to update Spot_julia so that it supports later julia version, I am not sure if this change is enough.


staticfloat:
You need to embed the libjulia versions into the `platforms` list, follow [libcxxwrap_julia](https://github.com/JuliaPackaging/Yggdrasil/blob/967a35fdfd5d9ede276fbe7408f7d8a9902cdcab/L/libcxxwrap_julia/build_tarballs.jl#L40-L44) for an example

MaximeBouton:
hmmm some of the builds are failing but I don't understand why, it throws this error: 

```
[ Info: Checking spot_julia.git-a3324d6370f948fc4a6d7d009b4ca967a5d86a679fd3c86604f4e2be32a363da out to spot_julia...
    Updating registry at `/depot/registries/General.toml`
   Resolving package versions...
ERROR: LoadError: KeyError: key UUID("a83860b7-747b-57cf-bf1f-3e79990d037f") not found
```

It seems to be when checking out the GitSource I provided, but it is inconsistent across builds. It seems to only affects builds with julia 1.6.3, which was the only one supported until now.

https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=22007&view=logs&j=5850b9b5-9a51-573f-7d03-f1fedbd8a7fb&t=34cffb34-e22b-5097-6aba-496cb946d943&l=1711

@staticfloat do you have any idea of why?

PR #5294, opened 2022-08-09
Update to WhiteboxTools 2.1.0

acgold:
WhiteboxTools stopped doing GitHub releases, but there was a versioning issue in the latest release v2.0.0 that kept everything at v1.5. I want to fix this for my new [package](https://github.com/acgold/Whitebox.jl) that uses WhiteboxTools_jll.jl.

This PR:
- Bumps WhiteboxTools to v2.1.0
- Gets the latest version of the repo with `GitSource()` instead of `ArchiveSource()`
- Changes the working directory within the bash recipe (original code errored when I briefly tested locally)

### DIFF
--- a/W/WhiteboxTools/build_tarballs.jl
+++ b/W/WhiteboxTools/build_tarballs.jl
@@ -7,7 +7,7 @@ version = v"2.1.0"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/jblindsay/whitebox-tools.git", "46229fd27a841e000c51b3b982b93e101a597ebe")
+    GitSource("https://github.com/jblindsay/whitebox-tools.git", "d4f252c84b37a6b70331c59fd930ffa9a574c5e1")
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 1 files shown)


vchuravy:
Is the issue in the tar.gz and thus using the git for 2.0.0 would be fine? There is currently no 2.1.0 release that I can see.

acgold:
The issue wasn't just the tar.gz, so using git for v2.0.0 would work if we pick a commit after they fixed the versioning issue. 

v2.1.0 would be preferred, though. I found the [commit for v2.1.0](https://github.com/jblindsay/whitebox-tools/commit/d4f252c84b37a6b70331c59fd930ffa9a574c5e1) where they bump the version of whitebox_tools from v2.0.1 to v2.1.0 in the Cargo.lock file + Release notes. They stopped doing official GitHub releases, and v2.1.0 is now what you download from their [website](https://www.whiteboxgeo.com/download-whiteboxtools/)

Maybe just changing the PR from using the latest commit to the initial v2.1.0 commit ("d4f252c84b37a6b70331c59fd930ffa9a574c5e1") is the way to go?

PR #5298, opened 2022-08-10
Bump go compiler version to 1.19

quinnj:


staticfloat:
This looks about right to me!  I'm assuming you want to use this in BB though, which means you're going to have to replicate this work here: https://github.com/JuliaPackaging/Yggdrasil/blob/master/0_RootFS/Go/build_tarballs.jl with a `dev`'ed version of BinaryBuilderBase.  The `build_tarballs.jl` script will modify your BB installation to have the new `Go` shards recorded within the `Artifacts.toml`, and you'll need to open a PR for BinaryBuilderBase.

PR #5304, opened 2022-08-12
[GR] bump GR version

jheinen:


jheinen:
Could someone please merge #5304. It contains an important bugfix for both Plots.jl and GR.jl.

Thanks in advance.

PR #5306, opened 2022-08-12
[MUMPS_seq] Version 5.5.1

amontoison:


amontoison:
@ViralBShah Is it possible to compile with `libblastrampoline_jll` for Julia < 1.8 ?
It seems to be not supported by Julia 1.6 but I don't know the issue with Julia 1.7.

ViralBShah:
Yes, in theory it should work for 1.7.

amontoison:
Thank @ViralBShah!
I changed the `julia_compat` for Julia 1.7 but I don't know if if we need something similar to https://github.com/JuliaPackaging/Yggdrasil/pull/4770.

ViralBShah:
Ah, I was wondering about that - and yes, it's best to follow the recipe there.

amontoison:
The release of Julia 1.8 is quite soon, let's just set the Julia compat to `v"1.8"`.

ViralBShah:
@giordano Is it ok to merge this now - or should we wait for 1.8 to release first?

amontoison:
Julia 1.8 is released, we can merge this PR and https://github.com/JuliaPackaging/Yggdrasil/pull/5309.

PR #5309, opened 2022-08-13
[MUMPS] Version 5.5.1

amontoison:


amontoison:
@eschnett 
I remarked that you recompiled all dependencies of MUMPS_jll with `MPITrampoline`.
Do I also need to compile MUMPS with `MPITrampoline` ?
I think that I just need to modify `INCPAR` and `LIBPAR` to support it.

amontoison:
`mpicc` is not working on some platforms.
@staticfloat Do you have an idea why it's not working?

PR #5311, opened 2022-08-14
New Recipe: FlangClassic

Keno:
This adds a new recipe for "flang classic" (not to be confused
with LLVM's new flang which cannot yet generate code) from
https://github.com/flang-compiler/flang. Currently only
x86_64-linux-musl is supported because the build does complicated
bootstrap things to build its runtime library. The intent here
is to try to this in sanitizer builds of fortran-containing jlls,
so the musl version is sufficient (though we'll need to figure out
how to build the rtlib for each target).

### DIFF
--- a/F/FlangClassic/build_tarballs.jl
+++ b/F/FlangClassic/build_tarballs.jl
@@ -2,7 +2,7 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
-name = "ClassicFlang"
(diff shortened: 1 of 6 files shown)


vchuravy:
Huh new flange should be able to generate code these days? At least 14/15

Keno:
> Huh new flange should be able to generate code these days? At least 14/15

My understanding is that while the plumbing exists now, it is still way too incomplete to compile anything real.

PR #5318, opened 2022-08-16
add micromamba

cjdoris:
This "builds" micromamba by grabbing the tarballs of the conda package for all supported platforms and extracting the executable.

### DIFF
--- a/M/micromamba/build_tarballs.jl
+++ b/M/micromamba/build_tarballs.jl
@@ -1,0 +1,75 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "micromamba"
+version = v"0.25.1"
+
+# Collection of sources required to build micromamba
+# These are actually just the conda packages for each platform
+sources = [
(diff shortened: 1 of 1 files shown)


cjdoris:
Looks like one of the builds failed while uploading the artifact. Can you run it again please?

staticfloat:
Did you try actually compiling from source? https://github.com/mamba-org/mamba/tree/master/micromamba

cjdoris:
No I didn't. The reason is that [the first step](https://mamba.readthedocs.io/en/latest/developer_zone/build_locally.html#install-dev-requirements) to building micromamba is to install conda and about 20 other conda packages. At which point I may as well just download the one package already containing the binary I want.

cjdoris:
By the way I'm adding this for Micromamba.jl, which currently grabs the conda package directly as an artifact. This approach works fine, it just seemed more in-keeping with the artifact system to have it be built by BinaryBuilder. Any thoughts? I'm happy to stick with the old approach and close this PR.

staticfloat:
Nope, no worries.  If it's that in-depth, I'm okay with this.  Just in general, we prefer to build from source, as that gives us a lot more freedom when it comes to architecture support.

PR #5325, opened 2022-08-17
[Rscape] new package

marcom:
R-scape is a package for predicting RNA secondary structure using evolutionary covariation observed from multiple sequence alignments.

http://rivaslab.org/
http://eddylab.org/R-scape/


marcom:
One file that was patched used CRLF line endings, unfortunately those got stripped from the patch file when i commited it with git and now the atomic patch fails. I'll try readding the patch with CRLF line endings intact.

marcom:
Closing, will try again with a new pull request and a slightly different approach.

PR #5331, opened 2022-08-18
Add a standalone LibStdCxx jll

Keno:
Rather than trying to shoehorn the libstdcxx build into bootstrap as in #5308,
just add a separate jll for it and in a follow-up PR, I'll adjust CSL to pull
libstdcxx from here instead. In the medium term future, we may want to change
this again, but for now, I think this is the path of least resistance.


Keno:
Ok, this seems to be working for both gcc v11 and v12. The PR as is has v11. I'm thinking merge this and then immediately do a PR to bump to v12 so we have both, depending on the timing of the base CSL bump. Thoughts @staticfloat @giordano ?

PR #5334, opened 2022-08-19
Compile Ipopt with LBT

amontoison:
superseded https://github.com/JuliaPackaging/Yggdrasil/pull/4126


C/Coin-OR/Ipopt/build_tarballs.jl
```diff
@@ -58,7 +58,7 @@ products = [
 dependencies = [
     Dependency("ASL_jll", ASL_version),
     Dependency("MUMPS_seq_jll", compat="=$(MUMPS_seq_version)"),
-    Dependency("OpenBLAS32_jll", OpenBLAS32_version),
+    Dependency("libblastrampoline_jll"),
```
giordano, line 61: Looks overall good to me, but to be clear, you'll have to make sure to have a backing 32-bit BLAS at runtime.  Is that OK?

amontoison: Yes, I just need to add this [line](https://github.com/JuliaSmoothOptimizers/HSL.jl/blob/main/src/HSL.jl#L17) in the `__init__` function of Ipopt.jl and add `OpenBLAS32_jll` as a dependency, right?

giordano: That should do it, yes

PR #5336, opened 2022-08-19
[CSL] Use libstdcxx from jll

Keno:
Needs to be updated with the BBB commit from https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/257 once that's merged, but should work otherwise.

### DIFF
--- a/C/CompilerSupportLibraries/build_tarballs.jl
+++ b/C/CompilerSupportLibraries/build_tarballs.jl
@@ -1,17 +1,36 @@
-using BinaryBuilder, SHA
+using BinaryBuilder, BinaryBuilderBase, SHA
 
 include("../../fancy_toys.jl")
 
 name = "CompilerSupportLibraries"
-version = v"0.6.1"
+version = v"0.5.2"
+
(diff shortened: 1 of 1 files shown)


C/CompilerSupportLibraries/build_tarballs.jl
```diff
@@ -1,17 +1,36 @@
-using BinaryBuilder, SHA
+using BinaryBuilder, BinaryBuilderBase, SHA
 
 include("../../fancy_toys.jl")
 
 name = "CompilerSupportLibraries"
-version = v"0.6.1"
+version = v"0.5.2"
```
giordano, line 6: Why downgrading the version number?

Keno: Sorry, that was just for testing stince the original libstdxx jll build was for that version. I'll undo this.

ViralBShah:
Still relevant or can we close?

fingolfin:
@Keno is this still needed or should it be closed?

PR #5338, opened 2022-08-20
Added libblastrampoline dependency

JuhaHeiskala:
Did not initially realize that liblastrampoline is a required dependency,


JuhaHeiskala:
Maybe this was by mistake, need to learn how libblastrampoline works.

PR #5339, opened 2022-08-20
Update chemfiles to release 0.10.3

Luthaf:
This also requires julia>=1.6 to enable Apple ARM64 CPU (https://github.com/chemfiles/Chemfiles.jl/issues/56)


giordano:
Seems to have problems on ARM architectures?

Luthaf:
Yes, endianness detection is broken on aarch64/musl, I'll add a patch!

Luthaf:
Took a couple of force push, but looks like CI is happy now!

PR #5342, opened 2022-08-22
Added VLFeat

stemann:


V/VLFeat/build_tarballs.jl
```diff
@@ -0,0 +1,81 @@
...
+
+atomic_patch -p1 ../patches/vlfeat-march.patch
+atomic_patch -p1 ../patches/vlfeat-glibc-limitation.patch
+atomic_patch -p1 ../patches/vlfeat-mingw-lowercase-windows-include.patch
+atomic_patch -p1 ../patches/vlfeat-windows.patch
+
+make_args=""
+
+if [[ $target != x86_64-* ]]; then
+    make_args+=" \
```
giordano, line 27: I believe string concatenation with `+=` is a bashism, which is OK because we use bash anyway, but a nicer way to append flags is to have use an array (we have plenty of uses here), which is also a bashism, but again that's ok

V/VLFeat/build_tarballs.jl
```diff
@@ -0,0 +1,81 @@
...
+        DISABLE_AVX=1 \
+        DISABLE_SSE2=1"
+fi
+
+if [[ $target == aarch64-apple-darwin*
```
giordano, line 32: Simply use `*-apple-darwin*` to match both systems?

V/VLFeat/build_tarballs.jl
```diff
@@ -0,0 +1,81 @@
...
+    || $target == x86_64-apple-darwin* ]]; then
+    ARCH=maci64
+elif [[ $target == i686-w64-mingw32 ]]; then
+    ARCH=win32
+elif [[ $target == x86_64-w64-mingw32 ]]; then
+    ARCH=win64
+elif [[ $target == i686-linux-*
```
giordano, line 39: Can also check the `nbits` [environment variable](https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables)

V/VLFeat/bundled/patches/vlfeat-mingw-lowercase-windows-include.patch
```diff
@@ -0,0 +1,39 @@
+diff --git a/src/test_threads.c b/src/test_threads.c
```
giordano, line 1: It'd be good to upstream this patch.

V/VLFeat/build_tarballs.jl
stemann: If I do
```bash
make_args=()
if [[ $target != x86_64-* ]]; then
    make_args+="DISABLE_AVX=1 "
    make_args+="DISABLE_SSE2=1 "
fi

# ...

make $make_args[@]
```
I only get a trailing `[@]` in the targets to `make`

```bash
make "$make_args[@]"
```
only yields the first argument of the array.

https://stackoverflow.com/questions/55316852/append-elements-to-an-array-in-bash

stemann: Never mind :-) `${make_args[@]}`

V/VLFeat/build_tarballs.jl
stemann: Yeah, I was trying to be explicit about the arch's as it was in order to set VLFeat's `ARCH` property - and to avoid future accidental true if clauses (when Apple switch to their next arch. :-) ), but... yeah.

Modified as you suggested.

V/VLFeat/build_tarballs.jl
stemann: Modified as you suggested.

V/VLFeat/bundled/patches/vlfeat-mingw-lowercase-windows-include.patch
stemann: Indeed.

PR #5345, opened 2022-08-23
[LASzip] Bump to also generate Mac M1 builds

evetion:


giordano:
https://github.com/JuliaPackaging/Yggdrasil/issues/2763

> There is a large demand to build packages for the "experimental" platforms `aarch64-apple-darwin` and `armv6l-linux-*`, but there are two _**NON-NEGOTIABLE**_ constraints
> 
> 1. the JLL package _**must require at least Julia v1.6**_, because only Julia v1.6+ is able to deal with the presence of these platforms in the `Artifacts.toml` file
> 2. most existing packages are marked as compatible with Julia v1.0, but as a consequence of the point 1 above if we want to rebuild an existing package for the new platform the generated JLL package _**must have a new version number**_.  No exceptions whatsoever to this rule, otherwise we'll break the ecosystem for Julia v1.5-

In particular, see point 2

L/LASzip/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "LASzip"
-version = v"3.4.3"
+version = v"3.4.3-1.6"
```
giordano, line 6: Sigh, no, you have to stick to `X.Y.Z` only.  I guess I'm going to prohibit anything else completely, since this is just the second attempt in one day: https://github.com/JuliaPackaging/Yggdrasil/pull/5344#discussion_r953604740

evetion: Should we have a PkgVersion (`pv"1.2.3"`) or similar to enforce such things?

PR #5346, opened 2022-08-23
new package R2R-1.0.6

marcom:
R2R is a package for drawing RNA secondary structures.

https://sourceforge.net/projects/weinberg-r2r/
https://pubmed.ncbi.nlm.nih.gov/21205310/


R/R2R/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+for f in src/SymbolicMath.cpp \
+         src/SymbolicMath.h \
+         src/GSCConsensus.cpp \
+         src/ParseOneStockholm.cpp \
+         src/PositionBackbone_MultiStemCircularSolver.cpp \
+         src/RnaDrawer.cpp;
+do
+    cp "$f" "$f".bak
+    tr -d '\015' < "$f".bak > "$f"
+done
```
giordano, line 29: https://github.com/JuliaPackaging/Yggdrasil/pull/5133#issuecomment-1180886129.  I just realised you did the same for Rscape

R/R2R/bundled/patches/fix-format-strings.patch
```diff
@@ -0,0 +1,68 @@
+diff --git a/src/ParseOneStockholm.cpp b/src/ParseOneStockholm.cpp
```
giordano, line 1: Any chance to upstream these patches?  As far as I can tell they can be useful to everyone else?

R/R2R/build_tarballs.jl
marcom: Ok, will rework the build to keep the line endings in the patches by setting .gitattributes appropriately here and in Rscape

marcom: Thanks for the gitattributes tip!

R/R2R/bundled/patches/fix-format-strings.patch
marcom: I will send the original author an email with the patches.

PR #5349, opened 2022-08-24
Add postgres support for GDAL

maxfreu:
This adds postgresql support to GDAL. This is my first PR to Yggdrasil, so I hope it works :D
cc @visr @evetion

### DIFF
--- a/G/GDAL/build_tarballs.jl
+++ b/G/GDAL/build_tarballs.jl
@@ -54,16 +54,17 @@ fi
 # Clear out `.la` files since they're often wrong and screw us up
 rm -f ${prefix}/lib/*.la
 
+# export environment variables for PostgrSQL
+export PQ_CFLAGS="-I${prefix}/include"
+export PQ_LIBS="-L${prefix}/lib -lpq"
+
 # Read the options in the log file
 ./configure --help
 
-./configure \
-    PQ_CFLAGS="-I${prefix}/include" \
-    PQ_LIBS="-L${prefix}/lib -lpq" \
-    --prefix=${prefix} --build=${MACHTYPE} --host=${target} \
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} \
     --with-geos=${bindir}/geos-config \
     --with-proj=$prefix \
-    --with-tiff=$prefix \
+    --with-libtiff=$prefix \
     --with-geotiff=$prefix \
     --with-libz=$prefix \
     --with-expat=$prefix \
--- a/G/GDAL/build_tarballs.jl
+++ b/G/GDAL/build_tarballs.jl
@@ -66,7 +66,7 @@ rm -f ${prefix}/lib/*.la
     --with-expat=$prefix \
     --with-zstd=$prefix \
     --with-sqlite3=$prefix \
-    --with-pg=$prefix \
+    --with-pg=yes \
     --with-curl=${bindir}/curl-config \
     --with-openjpeg \
     --with-python=no \
--- a/G/GDAL/build_tarballs.jl
+++ b/G/GDAL/build_tarballs.jl
@@ -66,6 +66,7 @@ rm -f ${prefix}/lib/*.la
     --with-expat=$prefix \
     --with-zstd=$prefix \
     --with-sqlite3=$prefix \
+    --with-pg=$prefix \
     --with-curl=${bindir}/curl-config \
     --with-openjpeg \
     --with-python=no \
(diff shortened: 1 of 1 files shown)


G/GDAL/build_tarballs.jl
```diff
@@ -66,6 +66,7 @@ rm -f ${prefix}/lib/*.la
     --with-expat=$prefix \
     --with-zstd=$prefix \
     --with-sqlite3=$prefix \
+    --with-pg=$prefix \
```
visr, line 69: The builds are failing with `configure: error: Only --with-pg=yes/no supported`.
Could you try that?

maxfreu:
still fails, looks like it can't find some headers:
```
[10:48:21] In file included from postgisraster.h:41:0,
[10:48:21]                  from postgisrasterdataset.cpp:38:
[10:48:21] /opt/x86_64-linux-gnu/bin/../x86_64-linux-gnu/sys-root/usr/local/include/libpq-fe.h:29:10: fatal error: postgres_ext.h: No such file or directory
[10:48:21]  #include "postgres_ext.h"
[10:48:21]           ^~~~~~~~~~~~~~~~
[10:48:21] compilation terminated.
```
Unfortunately `--with-pg` only allows yes/no and not to specify some include path. Compiling C is always a nice reminder of why to use julia...

visr:
It looks like you might need to set the `PG_LIB` and `PG_INC` environment variables to help it find it.

https://github.com/OSGeo/gdal/issues/1657#issuecomment-503474887
https://lists.osgeo.org/pipermail/gdal-dev/2003-November/001433.html

maxfreu:
>It looks like you might need to set the PG_LIB and PG_INC environment variables to help it find it.

Aaah, good old mailing lists :D

I'm a bit surprised: I cloned the original GDAL directory including the patches, but when I execute the script `build_tarballs.jl` for linux x64 I get `configure: error: PROJ 6 symbols not found` o.O In the meantime I'll run one more test via github build.

visr:
I think you have to set the environment variables before calling `configure`, just like the others with `export VAR="VAL"`.
Some relevant lines from the log:

```
PKG_CONFIG  path to pkg-config utility
PKG_CONFIG_PATH
            directories to add to pkg-config's search path
PKG_CONFIG_LIBDIR
            path overriding pkg-config's built-in search path
PQ_CFLAGS   C compiler flags for PQ, overriding pkg-config
PQ_LIBS     linker flags for PQ, overriding pkg-config

checking for PostgreSQL... checking for arm-linux-gnueabihf-pkg-config... no
checking for pkg-config... /usr/bin/pkg-config
configure: WARNING: using cross tools not prefixed with host triplet
checking pkg-config is at least version 0.21... yes
checking for PQ... yes
checking for PQconnectdb in -lpq... yes
```

maxfreu:
Yes, that's the cleaner solution. I just copied from the issue comment. I still get two errors:
[configure: WARNING: unrecognized options: --with-tiff](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=21633&view=logs&j=8e73093c-0234-55e0-c81b-59fd6c79b504&t=03700b4d-6f44-571c-9577-961261a4a1c7&l=4780) and [configure: error: PROJ 6 symbols not found](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=21633&view=logs&j=8e73093c-0234-55e0-c81b-59fd6c79b504&t=03700b4d-6f44-571c-9577-961261a4a1c7&l=4990). I also get the PROJ6 error locally, looks like some version incompatibility? However, I havent changed anything wrt PROJ_jll and LibPQ_jll does not have PROJ as dependency.

maxfreu:
I also changed the `--with-tiff` to `--with-libtiff` and now it complains `libtiff >= 4.0 is required`. I think I should revert that; one problem is enough.

visr:
Hmm yeah I wonder. It seems the build is indeed broken regardless of Postgres.

Soon we will need to switch this build to the CMake build system, since the current one is deprecated and will be removed in GDAL 3.6: https://github.com/OSGeo/gdal/blob/v3.5.1/NEWS.md#in-a-nutshell

Not sure how much work it would be, but perhaps if we make that switch, adding new dependencies will be easier.

maxfreu:
Yeah, I converted this to a draft then, so that we can first convert the build to cmake and add postgres later. Or if everything works in one go thats also fine.

maxfreu:
@visr the issue with postgres was that it did not install all needed header files. https://github.com/JuliaPackaging/Yggdrasil/pull/5700 should fix this; then we can add support.

PR #5353, opened 2022-08-24
Updated SLEEF to v3.5.2 (2020-12-20)

stemann:


S/SLEEF/build_tarballs.jl
```diff
@@ -3,18 +3,17 @@
 using BinaryBuilder, Pkg
 
 name = "SLEEF"
-version = v"3.4.1"
+version = v"0.0.20201220" # This is an unreleased version later than v3.5.1 (the currently latest released version).
```
ViralBShah, line 6: This will never get used right, because it is older than the latest released version? Maybe just call it 3.5.1?

giordano: Also, why not using 3.5.1 instead?

stemann: I have a commit for Torch waiting where I am very specific about Torch dependencies, e.g.
```
    Dependency("SLEEF_jll", v"0.0.20201220"),
```
in order to match the JLL dependency versions exactly with the git submodules used by PyTorch.

Won't it work then?

Reason for not using v3.5.1 is that PyTorch uses this exact commit - and this is a commit that is not released and includes changes not in v3.5.1.

ViralBShah: Right, but this version number is smaller than 3.4.1. So you need something between 3.4.1 and 3.5.2, I believe. Maybe just make this 3.5.1, and eventually there will be 3.5.2, which we can move to, I imagine.

stemann: OK - since v3.5.1 is already released, I suggest we call this v3.5.2 - we can build a v3.5.2+1 if an official/upstream v3.5.2 is released.

PR #5377, opened 2022-08-25
[pandoc] Version bump to 2.19.2

jeremiahpslewis:


P/pandoc/build_tarballs.jl
```diff
@@ -2,15 +2,15 @@ using BinaryBuilder
 
 # Collection of pre-build pandoc binaries
 name = "pandoc"
-version = v"2.14"
-pandoc_ver = "2.14"
+version = v"2.19.2"
+pandoc_ver = version.patch == 0 ? "$(version.major).$(version.minor)" : string(version)
```
giordano, line 6: I usually do the other way round which is way simpler
```suggestion
pandoc_ver = "2.19.2"
version = VersionNumber(pandoc_ver)
```

PR #2028, opened 2020-11-01
Update glmnet to use latest source from R package

JackDunnNZ:
Ref https://github.com/JuliaStats/GLMNet.jl/issues/46

The current glmnet build uses an [older version of glmnet](https://github.com/JuliaStats/GLMNet.jl/blob/8d8d22d573f7f608515fcf5f5192cf28cced8f1b/deps/glmnet5.f90) from 2014 that was copied into the GLMNet repo a long time ago, and this changes to build from the current [R package source](https://github.com/cran/glmnet) instead.

There are a couple of issues where I need some help:
- the original glmnet build was given version 5, because the file was named `glmnet5.f90` with no version info in the the file other than a datestamp. I figured it would make sense to sync the version to the tagged version of the R package, but the problem is that the most recent tag is currently v4.0.2, so there is the chance for a v5 conflict in future. I'd appreciate input on how to best solve this.
- there is a warning about no license file being present. The license is GPLv2, but many R packages only specify the choice of license in their [DESCRIPTION file](https://github.com/cran/glmnet/blob/b1a4b50de01e0cd24343959d7cf86452bac17b26/DESCRIPTION#L17) rather than including a separate and complete license file. What is the best way to resolve this?

PR #5378, opened 2022-08-25
[glmnet] New version 2.0.13

JackDunnNZ:
This is the most recent version of the glmnet library (tagged as [`2.0-13`](https://github.com/cran/glmnet/tree/2.0-13/src)) that works with the current (0.7) release of the GLMNet.jl package.

Trying to use the following [`2.0-16`](https://github.com/cran/glmnet/tree/2.0-16/src) release or any later version with the current GLMNet.jl results in a segfault. I am trying to figure this out, but have not yet been successful (see [here](https://discourse.julialang.org/t/help-with-updating-fortran-wrapper/86221) for more info). This problem occurs with v4.0.2 build that was added in #2028, this version has never been used by the GLMNet.jl package.

This version of the source is more recent than the JLL that is currently used by GLMNet.jl (which was added in #540 as v5.0.0, because confusingly glmnet changed their versioning scheme at some point, see [prior discussion](https://github.com/JuliaPackaging/Yggdrasil/pull/2028#issuecomment-720110906))

The main reason for this update is to rebuild a working version of the JLL to get support for the new platforms, specifically macaarch64, so that the GLMNet.jl package can work here now that Julia 1.8 is released and supported. Currently there is no artifact available here, and I would like to fix that ASAP while continuing to figure out the problems with later glmnet versions

### DIFF
--- a/G/glmnet/build_tarballs.jl
+++ b/G/glmnet/build_tarballs.jl
@@ -3,24 +3,16 @@
 using BinaryBuilder, Pkg
 
 name = "glmnet"
-version = v"4.0.2"
+version = v"2.0.13"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/cran/glmnet.git", "b1a4b50de01e0cd24343959d7cf86452bac17b26")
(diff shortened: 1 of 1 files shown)


JackDunnNZ:
Thanks @giordano!

PR #5383, opened 2022-08-26
[MDAL] Add tools, drivers

evetion:
Windows builds have less drivers (due to NetCDF missing) than other platforms: https://github.com/evetion/MDAL.jl/actions/runs/2934632776

### DIFF
--- a/M/MDAL/build_tarballs.jl
+++ b/M/MDAL/build_tarballs.jl
@@ -29,11 +29,12 @@ CMAKE_FLAGS=(-DCMAKE_INSTALL_PREFIX=$prefix
 -DWITH_SQLITE3=ON
 -DBUILD_STATIC=OFF
 -DBUILD_SHARED=ON
--DBUILD_TOOLS=OFF
+-DBUILD_TOOLS=ON
 -DBUILD_EXTERNAL_DRIVERS=OFF)
 
-#NetCDF is the most restrictive dependency as far as platform availability, so we'll use it where applicable but disable it otherwise
+# NetCDF is the most restrictive dependency as far as platform availability, so we'll use it where applicable but disable it otherwise
 if ! find ${libdir} -name "libnetcdf*.${dlext}" -exec false '{}' +; then
+    echo "Disabling NetCDF support"
     CMAKE_FLAGS+=(-DWITH_NETCDF=ON)
 else
     CMAKE_FLAGS+=(-DWITH_NETCDF=OFF)
@@ -64,5 +65,5 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-#GDAL uses a preferred of 6 so match that
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"6")
+# GDAL uses a preferred of 6 so match that
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"6")
--- a/M/MDAL/build_tarballs.jl
+++ b/M/MDAL/build_tarballs.jl
@@ -36,7 +36,6 @@ if ! find ${libdir} -name "libnetcdf*.${dlext}" -exec false '{}' +; then
     CMAKE_FLAGS+=(-DWITH_NETCDF=ON)
 else
     echo "Disabling NetCDF support"
-    ls ${libdir}
     CMAKE_FLAGS+=(-DWITH_NETCDF=OFF)
 fi
 
@@ -75,7 +74,7 @@ dependencies = [
     Dependency(PackageSpec(name="HDF5_jll", uuid="0234f1f7-429e-5d53-9886-15a909be8d59"))
     Dependency(PackageSpec(name="XML2_jll", uuid="02c8fc9c-b97f-50b9-bbe4-9be30ff0a78a"))
     Dependency(PackageSpec(name="SQLite_jll", uuid="76ed43ae-9a5d-5a62-8c75-30186b810ce8"))
-    Dependency(PackageSpec(name="NetCDF_jll", uuid="7243133f-43d8-5620-bbf4-c2c921802cf3"); compat="400.802.102 - 400.902")
+    Dependency(PackageSpec(name="NetCDF_jll", uuid="7243133f-43d8-5620-bbf4-c2c921802cf3"); compat="400.802.102 - 400.902.5")
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/M/MDAL/build_tarballs.jl
+++ b/M/MDAL/build_tarballs.jl
@@ -34,9 +34,10 @@ CMAKE_FLAGS=(-DCMAKE_INSTALL_PREFIX=$prefix
 
 # NetCDF is the most restrictive dependency as far as platform availability, so we'll use it where applicable but disable it otherwise
 if ! find ${libdir} -name "libnetcdf*.${dlext}" -exec false '{}' +; then
-    echo "Disabling NetCDF support"
     CMAKE_FLAGS+=(-DWITH_NETCDF=ON)
 else
+    echo "Disabling NetCDF support"
+    ls ${libdir}
     CMAKE_FLAGS+=(-DWITH_NETCDF=OFF)
 fi
 
--- a/M/MDAL/build_tarballs.jl
+++ b/M/MDAL/build_tarballs.jl
@@ -66,5 +66,5 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-# GDAL uses a preferred of 6 so match that
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"6")
+# GDAL uses a preferred of 7 so match that
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"7")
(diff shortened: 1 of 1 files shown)


evetion:
It seems enabling the tools (mdalinfo executable) triggers linking errors on Linux (only): 

```
[20:38:22] /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root//usr/local/lib/libz.so.1: undefined reference to `__msan_init'
[20:38:22] /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root//usr/local/lib/libz.so.1: undefined reference to `__msan_retval_tls'
[20:38:22] /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root//usr/local/lib/libz.so.1: undefined reference to `__msan_memmove'
[20:38:22] /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root//usr/local/lib/libz.so.1: undefined reference to `__msan_va_arg_tls'
[20:38:22] /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root//usr/local/lib/libz.so.1: undefined reference to `__msan_param_tls'
[20:38:22] /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root//usr/local/lib/libz.so.1: undefined reference to `__msan_warning_with_origin_noreturn'
[20:38:22] /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root//usr/local/lib/libz.so.1: undefined reference to `__msan_va_arg_overflow_size_tls'
[20:38:22] /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root//usr/local/lib/libz.so.1: undefined reference to `__msan_memset'
[20:38:22] /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root//usr/local/lib/libz.so.1: undefined reference to `__msan_memcpy'
```

giordano:
Things are a bit in flux here trying to get memory sanitizer working in CI for Julia: https://github.com/JuliaLang/julia/issues/46349

evetion:
I'm not into sanitizers, is there a way to disable it for building the executables?

giordano:
If there was a way to avoid mismatches we'd have put an automated fix in place already. The fact that there is no solution at the moment was the whole point of previous message. We're figuring out what to do as we go.

Keno:
Just to close the circle, the version of BBB on Yggdrasil master was update to avoid the msan issue, so if this is rebased, it might be able to move forward.

giordano:
This has been rerun already after #5391, and `x86_64-linux-gnu` is now passing.  Now there are failures unrelated to msan in other platforms (well, they were there also before, since msan affected only `x86_64-linux-gnu` anyway)

giordano:
The remaining error
```
[12:15:49] /opt/bin/i686-linux-gnu-libgfortran3-cxx03/i686-linux-gnu-g++ --sysroot=/opt/i686-linux-gnu/i686-linux-gnu/sys-root/  -Wall -Wno-long-long -pedantic -O3 -DNDEBUG CMakeFiles/mdalinfo.dir/mdalinfo.cpp.o -o mdalinfo  -Wl,-rpath,/workspace/srcdir/MDAL-release-0.9.4/mdal:/usr/local/lib: ../mdal/libmdal.so /opt/i686-linux-gnu/i686-linux-gnu/sys-root/usr/local/lib/libgdal.so /opt/i686-linux-gnu/i686-linux-gnu/sys-root/usr/local/lib/libxml2.so /opt/i686-linux-gnu/i686-linux-gnu/sys-root/usr/local/lib/libsqlite3.so 
[12:15:49] /opt/i686-linux-gnu/i686-linux-gnu/sys-root/usr/local/lib/libgdal.so: undefined reference to `__divmoddi4@GCC_7.0.0'
[12:15:49] collect2: error: ld returned 1 exit status
```
looks very much like https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/163, which is weird because it should have been fixed already.

giordano:
#4979 increased the version of GCC used to build gdal from 6 to 7, so it may be a good idea to try and use gcc 7 here as well, but then I don't know how #5382 worked, if that's the issue

evetion:
This adds building some Executables, those errors happen on linking those.

If I want to build this _with_ NetCDF on Windows, should I also build NetCDF with `expand_cxxstring_abis`? Now I either get `Dependency NetCDF_jll does not have a mapping for artifact NetCDF for platform x86_64-w64-mingw32-libgfortran3-cxx03`, or without the `expand_cxxstring_abis` in this build, the Windows one does build with NetCDF, but complains about ```/agent/_work/1/s/M/MDAL/build/x86_64-w64-mingw32/Mx3j4jNY/x86_64-w64-mingw32-libgfortran3-cxx11/destdir/lib/libmdal.dll contains std::string values!  This causes incompatibilities across the GCC 4/5 version boundary.  To remedy this, you must build a tarball for both GCC 4 and GCC 5.```

evetion:
Odd, on x64-linux-musl, curl suddenly misbehaves with GCC7.

11:59:56] /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libgdal.so: undefined reference to `curl_multi_init@CURL_4'
[11:59:56] /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libgdal.so: undefined reference to `curl_multi_wait@CURL_4'
[11:59:56] /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libgdal.so: undefined reference to `curl_multi_poll@CURL_4'
[11:59:56] /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libgdal.so: undefined reference to `curl_slist_free_all@CURL_4'
[11:59:56] /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libgdal.so: undefined reference to `curl_version_info@CURL_4'
[11:59:56] /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libgdal.so: undefined reference to `curl_mime_data_cb@CURL_4'
[11:59:56] /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libgdal.so: undefined reference to `curl_multi_perform@CURL_4'
[11:59:56] /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libgdal.so: undefined reference to `curl_mime_free@CURL_4'

evetion:
And when cheating, and just checking whether a hdf5 library is available, we now also fail on all Windows platforms:

```
Could NOT find HDF5 (missing: HDF5_LIBRARIES) (found version "1.12.2")
```

Note sure if that's a bug, or there are explicit version checks or so.

giordano:
> now misbehaves with libexpat...
> /usr/lib/libexpat.so.1: undefined reference to getrandom'

The problem is that something is pushing `/usr/lib` into the search path: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=21839&view=logs&j=37ef6d6b-0cd0-533b-6dc1-e159d89496a0&t=138bfd31-39fd-5067-452b-ad7bf0cf1a7a&l=1269
```
[12:47:18] /opt/bin/x86_64-linux-musl-libgfortran4-cxx03/x86_64-linux-musl-g++ --sysroot=/opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/  -Wall -Wno-long-long -pedantic -O3 -DNDEBUG -lcurl CMakeFiles/mdal_translate.dir/mdal_translate.cpp.o -o mdal_translate  -Wl,-rpath,/workspace/srcdir/MDAL-release-0.9.4/mdal:/usr/local/lib: ../mdal/libmdal.so /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libgdal.so /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libxml2.so /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/local/lib/libsqlite3.so 
```
Needless to say that's not a great idea.


> Note sure if that's a bug, or there are explicit version checks or so.

Quoting from [the documentation](https://docs.binarybuilder.org/stable/troubleshooting/#General-comments):

evetion:
🎉

I'm still missing NetCDF on Windows though, which does work if we don't use `expand_cxxstring_abis` it seems.

M/MDAL/build_tarballs.jl
```diff
@@ -23,46 +23,61 @@ CMAKE_FLAGS=(-DCMAKE_INSTALL_PREFIX=$prefix
...
+    ExecutableProduct("mdalinfo", :mdalinfo_path)
 ]
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency(PackageSpec(name="GDAL_jll", uuid="a7073274-a066-55f0-b90d-d619367d196c"))
+    Dependency(PackageSpec(name="HDF5_jll", uuid="0234f1f7-429e-5d53-9886-15a909be8d59"))
     Dependency(PackageSpec(name="XML2_jll", uuid="02c8fc9c-b97f-50b9-bbe4-9be30ff0a78a"))
     Dependency(PackageSpec(name="SQLite_jll", uuid="76ed43ae-9a5d-5a62-8c75-30186b810ce8"))
     Dependency(PackageSpec(name="NetCDF_jll", uuid="7243133f-43d8-5620-bbf4-c2c921802cf3"); compat="400.802.102 - 400.902")
```
giordano, line 78: As explained in #5429, if you want to link against netcdf for you you may want to pick up a version of NetCDF_jll which actually has a Windows build

M/MDAL/build_tarballs.jl
```diff
@@ -23,46 +23,61 @@ CMAKE_FLAGS=(-DCMAKE_INSTALL_PREFIX=$prefix
...
+-DBUILD_TOOLS=ON
 -DBUILD_EXTERNAL_DRIVERS=OFF)
 
-#NetCDF is the most restrictive dependency as far as platform availability, so we'll use it where applicable but disable it otherwise
+# NetCDF is the most restrictive dependency as far as platform availability, so we'll use it where applicable but disable it otherwise
 if ! find ${libdir} -name "libnetcdf*.${dlext}" -exec false '{}' +; then
     CMAKE_FLAGS+=(-DWITH_NETCDF=ON)
 else
+    echo "Disabling NetCDF support"
+    ls ${libdir}
```
giordano, line 39:
```suggestion
```

M/MDAL/build_tarballs.jl
evetion: D'oh! Sorry the noise and thanks for pointing it out. I didn't notice the low version number for the package. Also somewhat assumed that it would go for the latest, and that got me confused.

evetion:
This now adds NetCDF, HDF5 to x64 Windows and Linux, and to macOS, which roughly enables all drivers for the most popular platforms. I'm happy to merge this.

PR #5384, opened 2022-08-27
[Rscape] use .gitattributes to deal with CRLF line endings in patches

marcom:
This updates the build to use .gitattributes to deal with CRLF line endings in patches


marcom:
Will rebase and resubmit the .gitattributes changes

PR #5392, opened 2022-08-29
[kubectl] Update to v1.25.0 and add binary for aarch64-apple-darwin

giordano:


kleinschmidt:
Thanks for the fix!!  Sorry I dropped the ball on it, was traveling.

PR #5395, opened 2022-08-30
Add blasfeo

bjack205:
Adds the [blasfeo](https://github.com/giaf/blasfeo) package. Note that this package supports specialization to the specific target processor, but this PR builds only for a generic processor, which will incur a performance penalty.

In the future I hope to dig more into figuring out a good way to may the BinaryBuilder targets to the targets supported by blasfeo (some of which are missing from their cmake interface so I'd have to switch this to work directly the the makefile interface, which would mean generating a local makefile specific to each target, which sounds gross). 

I believe one of the critical things blasfeo specializes on is SIMD instructions, but I didn't see any way in BinaryBuilder to specify / specialize on this? Is it best to build for a generic processor, or could we duplicate the artifacts to specialize on the SIMD width? I'm pretty new to BinaryBuilder, so best practices here aren't clear to me.


giordano:
BinaryBuilder does have the ability to target different microarchitectures, but artifact selection on the Pkg side is a bit broken: when you go and install the package, the artifact for the wrong microarchitecture (maybe a lower one than yours, or, worse, one incompatible with yours) may be installed, which isn't a good outcome.  We also haven't settled on the exact set on microarchitectures to support.

However, this is made largely redundant by the fact most high-profile libraries (in the case of BLAS, OpenBLAS and BLIS) use CPUID to do runtime detection of the CPU, so that it isn't necessary to make any decision at compile-time.  I take blasfeo doesn't do that?

giordano:
Maybe @imciner2 knows more?

bjack205:
AFAIK (I'm not an author of the package, just a user), the specialization happens at compile-time. Digging around in the source it doesn't seem to have any support for this type of run-time detection.

imciner2:
I had originally thought of packaging BLASFEO several years ago, but I was actually waiting for the microarchitecture support to mature a bit in BinaryBuilder before I did anything with it (I actually have a blank wrapper repo already here https://github.com/JuliaEmbedded/BLASFEO.jl). The reason for this is the architectures are all decided at compile time in BLASFEO, and using the wrong library on a computer will cause an illegal instruction error and terminate the application. I did build in a system that allows you to query what the library supports, and then also query what the processor supports in the hope that doing that query can allow for catching these inconsistencies, but that will require a careful loading of the library and then unloading if it is the incorrect one for the architecture.

This also isn't really the best way to package BLASFEO in BinaryBuilder, and I don't even know if this library would be useful in this form. It is currently using the `GENERIC` target, which is basically just C-level BLAS functions - not any of the optimized ones that BLASFEO is really known for.

bjack205:
Thanks for your input! I was hoping it'd be simpler to extend the `build_tarballs.jl` in the future to map from the BinaryBuilder targets to those of blasfeo, but it looks like that's probably a lot harder than expected. It still might be useful to have this around, at least as a convenient fallback. 

I've also already wrapped HPIPM (my blasfeo wrapper is a WIP), and plan on letting the user use this generic version by default, or point to a local build of blasfeo via an environment variable.

PR #5396, opened 2022-08-31
Update TempestRemap deps

charleskawczynski:
This PR updates the TempestRemap deps to be compatible with the latest . Refs:
 - https://github.com/Alexander-Barth/NCDatasets.jl/issues/182#issuecomment-1219900206
 - https://github.com/Alexander-Barth/NCDatasets.jl/issues/164
 - https://github.com/JuliaPackaging/Yggdrasil/pull/5319

We need this to update [ClimaCoreTempestRemap](https://github.com/CliMA/ClimaCore.jl/tree/main/lib/ClimaCoreTempestRemap) to julia 1.8. I think this is the only outstanding blocker for updating. cc @simonbyrne

I've never done this before, and I'm almost certainly doing this wrong, but I'll update as I figure out how to do this correctly.

### DIFF
--- a/T/TempestRemap/build_tarballs.jl
+++ b/T/TempestRemap/build_tarballs.jl
@@ -3,9 +3,10 @@
 using BinaryBuilder, Pkg
 
 name = "TempestRemap"
+upstream_version = v"2.1.3"
 version = v"2.1.4"
 sources = [
-    ArchiveSource("https://github.com/ClimateGlobalChange/tempestremap/archive/refs/tags/v$(version).tar.gz",
+    ArchiveSource("https://github.com/ClimateGlobalChange/tempestremap/archive/refs/tags/v$(upstream_version).tar.gz",
                   "f3925871b9bc19c39002665970283a6f70ec5e159f0c87c74d1ef4f7efa4c26a"),
 ]
 
--- a/T/TempestRemap/build_tarballs.jl
+++ b/T/TempestRemap/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "TempestRemap"
-version = v"2.1.3"
+version = v"2.1.4"
 sources = [
     ArchiveSource("https://github.com/ClimateGlobalChange/tempestremap/archive/refs/tags/v$(version).tar.gz",
                   "f3925871b9bc19c39002665970283a6f70ec5e159f0c87c74d1ef4f7efa4c26a"),
@@ -40,7 +40,7 @@ platforms = [
     Platform("x86_64", "linux"),
     Platform("aarch64", "linux"; libc="glibc"),
     Platform("x86_64", "macos"),
-] 
+]
 platforms = expand_cxxstring_abis(platforms)
 
 products = [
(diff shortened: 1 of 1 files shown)


T/TempestRemap/build_tarballs.jl
```diff
@@ -72,11 +72,11 @@ products = [
...
     Dependency("OpenBLAS32_jll"),
-    Dependency("NetCDF_jll", compat="400.702.402 - 400.799"),
+    Dependency("NetCDF_jll", compat="400.702.402 - 400.902.5"),
     # The following is adapted from NetCDF_jll
     BuildDependency(PackageSpec(; name="MbedTLS_jll", version=v"2.24.0")),
 ]
 
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-    julia_compat="1.7",
+    julia_compat="1.8",
```
giordano, line 81: Is this needed?

charleskawczynski: 🤷🏻‍♂️, I've never done this before, so probably not?

T/TempestRemap/build_tarballs.jl
```diff
@@ -72,11 +72,11 @@ products = [
 
 dependencies = [
     Dependency("OpenBLAS32_jll"),
-    Dependency("NetCDF_jll", compat="400.702.402 - 400.799"),
+    Dependency("NetCDF_jll", compat="400.702.402 - 400.902.5"),
```
giordano, line 75: So, this is saying that this package should be compatible with netcdf from v 4.7 to v4.9, but if that's the case, I don't see the need to rebuild this package at all: if netcdf 4.7 - 4.9 are mutually ABI-compatible, there is no need to change anything here.  Instead, if they are incompatible, also the minimum bound should be changed.

Skimming through the history of https://github.com/JuliaBinaryWrappers/NetCDF_jll.jl, I think you may want to raise the minimum bound to at least `400.802.101` (because that's the first version which has support for the same platforms as the latest one, and the library has the same soname -> same ABI).
```suggestion
    Dependency("NetCDF_jll", compat="400.802.101 - 400.902.5"),
```

T/TempestRemap/build_tarballs.jl
giordano: Yeah, I don't think this is necessary at all

T/TempestRemap/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "TempestRemap"
-version = v"2.1.3"
+version = v"2.1.4"
 sources = [
     ArchiveSource("https://github.com/ClimateGlobalChange/tempestremap/archive/refs/tags/v$(version).tar.gz",
```
giordano, line 8: I presume you changed the version number in order to have a new version, different from the previous one, which is the right thing to do, but unfortunately this means we have to diverge from the upstream version:
```suggestion
upstream_version = v"2.1.3"
version = v"2.1.4"
sources = [
    ArchiveSource("https://github.com/ClimateGlobalChange/tempestremap/archive/refs/tags/v$(upstream_version).tar.gz",
```

charleskawczynski: Ah, ok, thanks!

PR #5399, opened 2022-08-31
Recompile IPOPT with MUMPS 5.4.1 and LBT

amontoison:
@odow @ViralBShah

### DIFF
--- a/C/Coin-OR/Ipopt/build_tarballs.jl
+++ b/C/Coin-OR/Ipopt/build_tarballs.jl
@@ -4,7 +4,7 @@ name = "Ipopt"
 version = Ipopt_version
 
 sources = [
-    GitSource("https://github.com/coin-or/Ipopt.git", Ipopt_gitsha),
+    GitSource("https://github.com/coin-or/Ipopt.git", Ipopt_gitsha)
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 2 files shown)


ViralBShah:
@giordano Is this the right way to pick LBT?

amontoison:
Maybe we should also add a compat entry for LBT to be sure to install the version `v5.1.1`.

ViralBShah:
But isn't LBT provided by Julia - and hence you don't get to pick which version it is?

giordano:
they can have different ABIs (like `libblastrampoline.so.5` vs `libblastrampoline.so.6` in the future) so you have to pick one and link against that one.  In julia v1.7+ we always use v5, but apparently the windows library is subtly broken.

ViralBShah:
Should we merge?

ViralBShah:
@giordano Do we need something like https://github.com/JuliaPackaging/Yggdrasil/pull/4770/files#diff-4bc380ae1048b4b2419d0800b074f33fa94b5101e3c56d2bf85f4bc3c6894416R63 for LBT?

giordano:
This package requires Julia v1.8 anyway, so at the moment we don't need to do much

odow:
I understood from https://github.com/jump-dev/Ipopt.jl/pull/327#issuecomment-1233540747 that this PR won't fix the issue?

PR #3915, opened 2021-11-20
[GTK4] Add builder

giordano:
Very WIP, I just copied the GTK3 recipe, no idea whether it'll work at all.

Fix #3898.

PR #5404, opened 2022-09-01
[gdk_pixbuf] Update to 2.42.8, enable introspection for x86_64-linux-gnu

jwahlstrand:
Bump gdkpixbuf to 2.42.8 and include gobject-introspection data for x86_64-linux-gnu.

On Apple (both ARM and x86_64) I ran into a linker bug and copied a workaround from GTK4's build_tarballs.jl. (#3915)

My real motive here is to get more introspection data to use in Gtk4.jl. If this looks OK, I'd like to repeat this for libgtk's other GLib-based dependencies.

### DIFF
--- a/G/gdk_pixbuf/build_tarballs.jl
+++ b/G/gdk_pixbuf/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "gdk_pixbuf"
-version = v"2.42.6"
+version = v"2.42.8"
 
 # Collection of sources required to build gdk-pixbuf
 sources = [
     ArchiveSource("https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/archive/$(version)/gdk-pixbuf-$(version).tar.bz2",
-                  "8a76cffe6a85f2602cf246c1c974eb475aea41c363a932f0c34695fa968f01fd"),
+                  "d122cb5d0ef32349c52c37f1dfd936c734886643b49a37706acccfe3af6aba77"),
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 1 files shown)


G/gdk_pixbuf/build_tarballs.jl
```diff
@@ -49,19 +66,23 @@ products = [
...
+introspect_platform = filter(p -> triplet(p) == "x86_64-linux-gnu", platforms)
+
 # Dependencies that must be installed before this package can be built
 dependencies = [
     # Need a host gettext for msgfmt
     HostBuildDependency("Gettext_jll"),
     # Need a host glib for glib-compile-resources
     HostBuildDependency("Glib_jll"),
-    Dependency("Glib_jll"; compat="2.68.1"),
+    Dependency("Glib_jll"; compat="2.68.3"),
```
giordano, line 78: Is this change necessary?

G/gdk_pixbuf/build_tarballs.jl
```diff
@@ -49,19 +66,23 @@ products = [
 # Some dependencies are needed only on Linux and FreeBSD
 linux_freebsd = filter(p->Sys.islinux(p)||Sys.isfreebsd(p), platforms)
 
+# gobject_introspection is needed only on x86_64-linux-gnu
+introspect_platform = filter(p -> triplet(p) == "x86_64-linux-gnu", platforms)
```
giordano, line 70: I prefer more declarative checks, checking the triplet is problematic in the case in which we need to expand the list of platforms above with more tags
```suggestion
introspect_platform = filter(p -> Sys.islinux(p) && libc(p) == "glibc" && arch(p) == "x86_64", platforms)
```

G/gdk_pixbuf/build_tarballs.jl
jwahlstrand: That GLib version was set in the build for gobject_introspection (by me, oops) and I'm seeing:
```
ERROR: LoadError: Unsatisfiable requirements detected for package Glib_jll [7746bdde]:
 Glib_jll [7746bdde] log:
 ├─possible versions are: 2.59.0-2.68.3 or uninstalled
 ├─restricted to versions 2.68.1 by an explicit requirement, leaving only versions 2.68.1
 └─restricted by compatibility requirements with gobject_introspection_jll [6306888c] to versions: 2.68.3 — no versions left
   └─gobject_introspection_jll [6306888c] log:
     ├─possible versions are: 1.72.0 or uninstalled
     └─restricted to versions * by an explicit requirement, leaving only versions 1.72.0
```
There's no reason to require 2.68.3 in gobject_introspection_jll -- could we create a new version of that with 2.68.1?

giordano: Uhm, I see.  No, ok, that should be fine, I just wanted to understand what's going on.

PR #5408, opened 2022-09-02
[libgeotiff] allow PROJ 9.1 compat

visr:
The current `Dependency("PROJ_jll"; compat="~900.0")`, which is also used in GDAL_jll, blocks PROJ_jll 900.100 that was just released.

The SOVERSION of PROJ 9.0 and 9.1 are the same, so I went for `Dependency("PROJ_jll"; compat="900")` to allow both. Or is it better to just allow compat of 9.1 and not higher?

Version and instructions around PROJ_SOVERSION:
https://github.com/OSGeo/PROJ/blob/9.1.0/CMakeLists.txt#L119
https://github.com/OSGeo/PROJ/blob/9.1.0/HOWTO-RELEASE#L36-L39

### DIFF
--- a/L/libgeotiff/build_tarballs.jl
+++ b/L/libgeotiff/build_tarballs.jl
@@ -1,11 +1,15 @@
 using BinaryBuilder, Pkg
 
 name = "libgeotiff"
-version = v"1.7.1"
+upstream_version = v"1.7.1"
+version_offset = v"0.0.0"
+version = VersionNumber(upstream_version.major * 100 + version_offset.major,
+                        upstream_version.minor * 100 + version_offset.minor,
+                        upstream_version.patch * 100 + version_offset.patch)
 
 # Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://github.com/OSGeo/libgeotiff/releases/download/$version/libgeotiff-$version.tar.gz",
+    ArchiveSource("https://github.com/OSGeo/libgeotiff/releases/download/$upstream_version/libgeotiff-$upstream_version.tar.gz",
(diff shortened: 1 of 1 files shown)


giordano:
The soversion of libproj didn't change between 9.0 and 9.1: https://github.com/JuliaBinaryWrappers/PROJ_jll.jl/commit/177640a37a66e7e573b6acf95af9df8ee7f4c0b0#diff-0ed6f04e79778dda22f5565f89d1f6e95329a8d15c57c4efa586fbe7b7374bcd.  I'd say this is fine as is now, but I think it's better to merge this PR without actually building anything (nothing changed in the build, really) and only fix the compat in the registry.  Does it sound good?

giordano:
Nevermind, the soname on Windows changed: https://github.com/JuliaBinaryWrappers/PROJ_jll.jl/commit/177640a37a66e7e573b6acf95af9df8ee7f4c0b0#diff-03bfa771dcf80ac712ba145fabccdb5413165fc04ae62c0fc327684c58bf979d.  Sigh.  This needs to require 9.1 only.

visr:
Ah right, it's because the DLL contains the minor version number right? `libproj_9_1.dll`
I now remember that's the issue with GEOS as well.

I just pushed a commit to set compat to PROJ 9.1.

giordano:
> Ah right, it's because the DLL contains the minor version number right? `libproj_9_1.dll`

Yes, instead on Unix systems they use a proper soversion (25, whatever number that mean) which didn't change between version 9.0 and 9.1, but changing the name of the file breaks the ABI...  Why they aren't consistent across different systems is everybody's guess.

PR #5414, opened 2022-09-02
[OpenBLAS] Update to v0.3.21

giordano:


giordano:
This worked 3 days ago in #5322 and we haven't updated the Manifest after that.  What the hell.

PR #5415, opened 2022-09-03
Sundials: Build for msan

Keno:
There have been some mysterious Sundials memory errors lately,
so there's been a request for running sundials with msan. It'll
also be useful to have a non-Base jll for testing that we can
have the artifacts system properly load the msan variants of
everything, so this will hopefully be useful for that.

### DIFF
--- a/S/Sundials/Sundials@5/build_tarballs.jl
+++ b/S/Sundials/Sundials@5/build_tarballs.jl
@@ -6,13 +6,32 @@ version = v"5.2.1" # <-- There is no version 5.2.1, but we need to change versio
 # Collection of sources required to build Sundials
 sources = [
     GitSource("https://github.com/LLNL/sundials.git",
-              "b16d3d3995668c9a13c9f4bee8b0113ff6a9cf6d"),
+              "8264ba5614fd9578786e5e8a3ba9f703ff795361"),
     DirectorySource("../bundled@5"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/sundials*
+
+    if [[ ${bb_full_target} == *-sanitize+memory* ]]; then
+        # For msan, we need to use flang to compile.
+        ## Create flang compiler wrapper
+        cat /opt/bin/${bb_full_target}/${target}-clang | sed 's/clang/flang/g' > /opt/bin/${bb_full_target}/${target}-flang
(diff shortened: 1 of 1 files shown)


S/Sundials/Sundials@5/build_tarballs.jl
```diff
@@ -6,14 +6,33 @@ version = v"5.2.1" # <-- There is no version 5.2.1, but we need to change versio
 # Collection of sources required to build Sundials
 sources = [
     GitSource("https://github.com/LLNL/sundials.git",
-              "b16d3d3995668c9a13c9f4bee8b0113ff6a9cf6d"),
+              "8264ba5614fd9578786e5e8a3ba9f703ff795361"),
```
Keno, line 9: This is because the commit sha that was here before is not in the upstream repo anymore - not sure what they did with it, but it used to point to the 5.2.0 commit and they do have a tag for that, so use that instead.

S/Sundials/Sundials@5/build_tarballs.jl
```diff
@@ -66,6 +85,7 @@ fi
 # We attempt to build for all defined platforms
 platforms = filter!(p -> arch(p) != "powerpc64le", supported_platforms(; experimental=true))
 platforms = expand_gfortran_versions(platforms)
+push!(platforms, Platform("x86_64", "linux"; sanitize="memory"))
```
giordano, line 88: Don't you need to do this before the libgfortran expansion?

Keno: Since we're building with flang, it's using the flang fortran runtime anyway. We may want to reflect that in the fortran tag somehow.

giordano: I see, right. I haven't checked this build yet (away from computer), but last time I had a look at flang my understanding was that the runtime is statically linked, so it shouldn't really need a tag of its own, right? Or you're worried about somehow mixing up code built with libgfortran and flang?

Keno: It has a dynamically linked runtime and math library. For now, I think we can just leave it tagless though. I'm not sure classic flang will be our long-term solution here.

giordano: Ok, the flang I looked at was probably a different one: https://twitter.com/owainkenway/status/1505326986243235841

ViralBShah:
I think we don't need this for now.

PR #5417, opened 2022-09-03
Wizard recipe: kdb_c_api-v2022.9.3

p-casgrain:

* Package name: kdb_c_api
* Version: v2022.9.3

### DIFF
--- a/K/kdb_c_api/build_tarballs.jl
+++ b/K/kdb_c_api/build_tarballs.jl
@@ -7,7 +7,7 @@ version = v"2022.09.03"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/KxSystems/kdb.git", "01fb1243c6b32079938ef88a4434e46d2f3c00be")
+    GitSource("https://github.com/KxSystems/kdb.git", "f1b8e5e1dfaf86b1ccd3a2634d10e07718b71ece")
 ]
 
 # Bash recipe for building across all platforms
@@ -18,7 +18,7 @@ mkdir ${libdir}
 if [[ ${target} == aarch64-linux-* ]]; then opath="kdb/l64arm/c.o"; fi
 if [[ ${target} == x86_64-linux-* ]]; then opath="kdb/l64/c.o"; fi
 if [[ ${target} == i686-linux-* ]]; then opath="kdb/l32/c.o"; fi
-if [[ ${target} == aarch64-apple-* ]]; then opath="kdb/m64/c.o"; fi
+if [[ ${target} == aarch64-apple-* ]]; then opath="kdb/m64arm/c.o"; fi
 if [[ ${target} == x86_64-apple-* ]]; then opath="kdb/m64/c.o"; CC="gcc"; fi
 if [[ ${target} == i686-apple-* ]]; then opath="kdb/m32/c.o"; fi
 if [[ ${target} == x86_64-w64-* ]]; then opath="kdb/w64/c.dll"; fi
--- a/K/kdb_c_api/build_tarballs.jl
+++ b/K/kdb_c_api/build_tarballs.jl
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "kdb_c_api"
-version = v"2022.9.3"
+version = v"2022.09.03"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/KxSystems/kdb.git", "f1b8e5e1dfaf86b1ccd3a2634d10e07718b71ece")
+    GitSource("https://github.com/KxSystems/kdb.git", "01fb1243c6b32079938ef88a4434e46d2f3c00be")
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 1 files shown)


K/kdb_c_api/build_tarballs.jl
```diff
@@ -3,39 +3,39 @@
...
-if [[ ${target} == aarch64-apple-* ]]; then opath="kdb/m64/c.o"; fi
-if [[ ${target} == x86_64-apple-* ]]; then opath="kdb/m64/c.o"; fi
-if [[ ${target} == i686-apple-* ]]; then opath="kdb/m32/c.o"; fi
-if [[ ${target} == x86_64-w64-* ]]; then opath="kdb/w64/c.dll"; fi
-if [[ ${target} == i686-w64-* ]]; then opath="kdb/w32/c.dll"; fi
+if [[ ${target} == aarch64-linux-* ]];  then opath="kdb/l64arm/c.o"; fi
+if [[ ${target} == x86_64-linux-* ]];   then opath="kdb/l64/c.o"; fi
+if [[ ${target} == i686-linux-* ]];     then opath="kdb/l32/c.o"; fi
+if [[ ${target} == aarch64-apple-* ]];  then opath="kdb/m64arm/c.o"; fi
+if [[ ${target} == x86_64-apple-* ]];   then opath="kdb/m64/c.o"; CC="gcc"; fi
```
giordano, line 22: Why using gcc?

K/kdb_c_api/build_tarballs.jl
```diff
@@ -3,39 +3,39 @@
...
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = [
     Platform("aarch64", "macos"; ),
+    Platform("x86_64", "windows"; ),
     Platform("i686", "windows"; ),
-    Platform("i686", "linux"; libc = "musl"),
     Platform("x86_64", "linux"; libc = "glibc"),
-    Platform("x86_64", "windows"; ),
+    Platform("i686", "linux"; libc = "musl"),
```
giordano, line 36: If I'm not mistaken this is only adding x86_64-apple-darwin.  Can you please revert the other useless changes?

giordano (commented):
When you want to update an existing builder, it's easier to edit it rather than running the wizard again.  This introduced unrelated changes which only add noise to the diff

K/kdb_c_api/build_tarballs.jl
p-casgrain: For whatever reason, `x86_64-apple-*` build would fail with clang default.

K/kdb_c_api/build_tarballs.jl
p-casgrain: Will do.

K/kdb_c_api/build_tarballs.jl
giordano: Well, I'd like to know the reason

K/kdb_c_api/build_tarballs.jl
```diff
@@ -44,7 +44,7 @@ platforms = [
 
 # The products that we will ensure are always built
 products = [
-    LibraryProduct("c", :kdb_c_so)
+    LibraryProduct("c", :kdb_c_dl)
```
giordano, line 47: Also, note that changing the name of the variable referencing the shared library is a breaking change, I'm not sure why you'd want to do that

K/kdb_c_api/build_tarballs.jl
p-casgrain: I get the following error without `CC="gcc"`:
```Undefined symbols for architecture x86_64:
  "___emutls_get_address", referenced from:
      l004 in c.o
      l005 in c.o
      l006 in c.o
      _krr in c.o
      l038 in c.o
      l039 in c.o
      l035 in c.o
      ...
ld: symbol(s) not found for architecture x86_64
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
[✗]
```

K/kdb_c_api/build_tarballs.jl
p-casgrain: reverted to previous. right now my WIP package Q2.jl is the only one to use it, so I don't think there was anything to break to begin with

giordano:
Wait, you're linking binary blobs?

p-casgrain:
> Wait, you're linking binary blobs?

what do you mean by this?
I take pre-compiled platform-specific binaries from [KxSystems/kdb](https://github.com/KxSystems/kdb) and generate shared library to be used with Julia

giordano:
> > Wait, you're linking binary blobs?
> 
> what do you mean by this?
> I take pre-compiled platform-specific binaries from [KxSystems/kdb](https://github.com/KxSystems/kdb) and generate shared library to be used with Julia

I mean exactly that.  I didn't understand why linking would fail with an undefined symbol if we were compiling everything from source, instead you're linking an existing object file which was clearly compiled with GCC (apparently `___emutls_get_address` is a symbol in GCC for macOS) even though there is no mention of that in the README, which is rather unusual since GCC isn't the default compiler on macOS.

p-casgrain:
> 



> > > Wait, you're linking binary blobs?
> > 
> > 
> > what do you mean by this?
> > I take pre-compiled platform-specific binaries from [KxSystems/kdb](https://github.com/KxSystems/kdb) and generate shared library to be used with Julia
> 
> I mean exactly that. I didn't understand why linking would fail with an undefined symbol if we were compiling everything from source, instead you're linking an existing object file which was clearly compiled with GCC (apparently `___emutls_get_address` is a symbol in GCC for macOS) even though there is no mention of that in the README, which is rather unusual since GCC isn't the default compiler on macOS.

🤷

PR #5422, opened 2022-09-03
Add rocminfo 4.2.0

pxl-th:


giordano:
Files must be called `build_tarballs.jl`

giordano:
Can you please split this into two PRs?  This is going again to create merge conflicts in the registry, I don't have much fun, nor interest, fixing that.  For example keep here only 4.2.0 and, ***only after that is merged in General***, add 4.5.2

PR #5423, opened 2022-09-04
[GDAL] switch build to CMake

visr:
As mentioned in #4979, GDAL deprecated the autoconf build system, so we needed to make the switch. It was probably a good idea to wait until GDAL 3.5.1, since there were a bunch of [CMake build fixes](https://github.com/OSGeo/gdal/blob/v3.5.1/NEWS.md) there. Actually it seems that it's quite a painless transition. All patches are removed. I did have to update to GCC 8 since I hit an internal compiler error on GCC 7 (details in b84cce5e134e8c07bd45a0a9b8eb438d4e60e5e2).

I tried to directly include Postgres support from #5349 but it seems to still pick up the wrong one so I reverted that for now in ddcfbab1ca429222d2573407693b373ddff40bfd.


G/GDAL/build_tarballs.jl
```diff
@@ -13,74 +13,42 @@ version = VersionNumber(upstream_version.major * 100 + version_offset.major,
...
-        export EXTRA_SQLITE3_LIBS="-lm"
-        # libpthread and libldl are needed for libgdal, so let's always use them
-        export LDFLAGS="$LDFLAGS -lpthread -ldl"
-    fi
-    autoreconf -vi
+if [[ "${target}" == *-freebsd* ]]; then
+    # Our FreeBSD libc has `environ` as undefined symbol, so the linker will
+    # complain if this symbol is used in the built library, even if this won't
+    # be a problem at runtime. This flag allows having undefined symbols.
+    export LDFLAGS="-undefined"
```
giordano, line 28: You're shifting a linking time error to a runtime error, it doesn't look like a great tradeoff

giordano (changes requested):

G/GDAL/build_tarballs.jl
giordano: Ugh, I see, it's https://docs.binarybuilder.org/stable/troubleshooting/#undefined-reference-to-environ'-1.  Sigh

visr: Yeah it's used in a `posix_spawnp` call. Interestingly that file got this FreeBSD patch recently: https://github.com/OSGeo/gdal/commit/4ff30d57c654f34d04eb1cc5b2b76e4d52c78b35

giordano: Then it'd probably be better to apply that patch than to change link flags.

visr: Sorry, that patch is already included. With or without it I get the same issue, I tried playing around with it a bit. In https://bugs.webkit.org/show_bug.cgi?id=138420 they resolved it by avoiding using `environ` altogether. Don't know if that is feasible. It is only used once, in this function:

https://github.com/OSGeo/gdal/blob/v3.5.1/port/cpl_spawn.cpp#L578-L612

On this line:

https://github.com/OSGeo/gdal/blob/v3.5.1/port/cpl_spawn.cpp#L708

visr:
I think I'm starting to understand why the PostgreSQL driver is not working. The latest error was

> postgres_ext.h: No such file or directory

When looking in the include directory of https://github.com/JuliaBinaryWrappers/LibPQ_jll.jl there is no such header file.

I thought we needed libpq but we probably need to build PostgreSQL including libpq.

I guess the best way forward would be to create a separate PostgreSQL_jll that builds the whole thing. Would be a bit odd to have that separate from LibPQ_jll, though I'm assuming the authors went for a libpq-only build on purpose there, to keep things light. cc @maxfreu because of #5349.

https://github.com/JuliaPackaging/Yggdrasil/blob/master/L/LibPQ/build_tarballs.jl
https://stackoverflow.com/questions/29803847/how-to-download-compile-install-only-the-libpq-source-on-a-server-that-does-n

Based on that I'd say let's not include PostgreSQL in this PR, but leave it for a follow up.

giordano (approved):

PR #5426, opened 2022-09-04
[HDF5] Also include dll.a in the output on Windows

evetion:
As it might make #5383 build on Windows

### DIFF
--- a/H/HDF5/build_tarballs.jl
+++ b/H/HDF5/build_tarballs.jl
@@ -2,7 +2,7 @@ using BinaryBuilder
 
 # Collection of sources required to build HDF5
 name = "HDF5"
-version = v"1.12.2"
+version = v"1.12.2000"
 
 sources = [
     # 32-bit Windows from https://packages.msys2.org/package/mingw-w64-i686-hdf5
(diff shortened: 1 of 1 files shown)


evetion:
The output on Windows now includes
```
libhdf5-0.dll
libhdf5.dll.a  # new
libhdf5_hl-0.dll
libhdf5_hl.dll.a  # new
libhdf5_tools-0.dll
libhdf5_tools.dll.a  # new
```

H/HDF5/build_tarballs.jl
```diff
@@ -2,7 +2,7 @@ using BinaryBuilder
 
 # Collection of sources required to build HDF5
 name = "HDF5"
-version = v"1.12.2"
+version = v"1.12.2000"
```
giordano, line 5: No need to change the version number

H/HDF5/build_tarballs.jl
```diff
@@ -37,7 +37,10 @@ mkdir -p ${libdir} ${includedir}
 if [[ ${target} == *mingw* ]]; then
     cd ${target}/mingw${nbits}
 
+    rm -f lib/{*_cpp*,*fortran*,*f90*} # we do not need these
     rm -f bin/{*_cpp*,*fortran*,*f90*} # we do not need these
+    
+    mv -v lib/libhdf5*.dll.a ${libdir}
```
giordano, line 43: I think import libraries are typically found in prefix/lib, not libdir

PR #5428, opened 2022-09-04
Use ROCmLLVM for ROCmCompilerSupport

pxl-th:


pxl-th:
@jpsamaroo, I haven't used patches that were used before. Do you think we still need them?

pxl-th:
It worked for me without them. I think we can try and bring them back if needed.

PR #5429, opened 2022-09-04
[NetCDF] Expand platforms for NetCDF.

evetion:
So #5383 can use NetCDF on Windows. Might be a long shot, or a misunderstanding of mine.


evetion:
That was easier than I thought 😅

giordano:
I'm missing the point of this PR. Also, please do not add commits (like merges) which don't change the content of the directory: they're useless and only make it harder for me to find the log

evetion:
Over at #5383, many of the builds https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=21939&view=results warn about ```
##[warning]Dependency NetCDF_jll does not have a mapping for artifact NetCDF for platform x86_64-w64-mingw32-libgfortran4-cxx11
```

I assume this PR could fix that.

giordano:
I don't think so

giordano:
The log shows that [`v400.802.102+0`](https://github.com/JuliaBinaryWrappers/NetCDF_jll.jl/releases/tag/NetCDF-v400.802.102+0) was used, which simply does _not_ have Windows build.  C++ string ABI has nothing to do with this.

PR #5434, opened 2022-09-05
Use ROCmLLVM for ROCmOpenCLRuntime

pxl-th:


pxl-th:
If there are no objections, it'd be good to merge it...

PR #5436, opened 2022-09-06
Add ROCmOpenCLRuntime 4.5.2 & use ROCmLLVM for HIP

pxl-th:
To speed up things a bit, I made HIP use ROCmLLVM as well in this PR.


pxl-th:
As for the patches for HIP, I added only [no-init-abort](https://github.com/JuliaPackaging/Yggdrasil/pull/5133) patch. Other patches are not needed.

pxl-th:
Would be good to have it merged if there are no objections... :)

PR #5437, opened 2022-09-07
PaPILO requires OpenBLAS

matbesancon:
Resolving
https://github.com/JuliaRegistries/General/pull/67137#issuecomment-1233456645
I'm surprised this didn't break earlier


amontoison:
@matbesancon 
The issue is probably with Ipopt_jll. We recompile Ipopt and MUMPS_jll with LBT_jll and you need to provide a 32 bits BLAS backend directly in your package.

matbesancon:
in that case, there seems to be a Blas dependency in PaPILO itself, hence my surprise that things didn't break before

matbesancon:
can we merge this? cc @giordano

giordano:
I'm not sure this is the right fix.  How did libscip end up linking to libopenblas, if that wasn't in the same environment in the first place?

matbesancon:
>  How did libscip end up linking to libopenblas

through papilo, I have to check exactly where it is needed

matbesancon:
@giordano it seems papilo does require it:
https://github.com/scipopt/papilo/blob/master/CMakeLists.txt#L177

giordano:
What is not clear to me is: how this ended up linking to openblas if openblas wasn't a (direct at least) dependency?  What has changed since last build?  Without clear answers to these questions I can't tell what's going on here.

imciner2:
> What is not clear to me is: how this ended up linking to openblas if openblas wasn't a (direct at least) dependency? What has changed since last build? Without clear answers to these questions I can't tell what's going on here.

CMake includes the detection logic for BLAS libraries in its default install, so as long as the library and include headers exist in the library/include folder, CMake will happily find it even if we don't have a listed direct dependency on it. Ipopt (or another package) probably previously relied on OpenBLAS when this was last built, so it was transitively added into the right directories in the build image, and therefore the CMake logic automatically picked it up and linked it without throwing an error at configure/link time. Now that the transitive dependency is broken, it broke at runtime.

giordano:
Ok, now I see.  Thanks for the recap!

giordano:
For the record: #5525

PR #5425, opened 2022-09-04
[Arrow] Another try at building Arrow

evetion:
Previous attempts at #918 and #1645. This is a long shot, but I really like to have #5423 working with GeoParquet.

PR #5440, opened 2022-09-07
[RE2] Add a RE2 build

evetion:
To support the long-shot in #5425.

### DIFF
--- a/R/RE2/build_tarballs.jl
+++ b/R/RE2/build_tarballs.jl
@@ -23,9 +23,9 @@ CMAKE_FLAGS=(-DCMAKE_INSTALL_PREFIX=$prefix
 -DCMAKE_BUILD_TYPE=Release
 -DBUILD_SHARED_LIBS=ON)
 
-# To actually use shared_mutex
+# Until https://github.com/google/re2/issues/390 is fixed
 if [[ ${target} == *mingw* ]]; then
-    CMAKE_FLAGS+=(-DCMAKE_CXX_FLAGS="-std=c++17")
+    sed -i 's/cxx_std_11/cxx_std_17/g' ../CMakeLists.txt
 fi
 
 cmake .. "${CMAKE_FLAGS[@]}"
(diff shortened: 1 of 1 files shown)


evetion:
Failing on [mingw builds](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=21988&view=logs&j=92fa2b0a-4446-5dfc-3cc6-99b6a1eee333&t=5e7bdf7f-262d-5c19-93f9-febcf0e419d6&l=310):
```
[18:54:26] In file included from /workspace/srcdir/re2-2022-06-01/re2/dfa.cc:41:0:
[18:54:26] /workspace/srcdir/re2-2022-06-01/util/mutex.h:36:24: fatal error: shared_mutex: No such file or directory
[18:54:26]  #include <shared_mutex>
```

R/RE2/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+CMAKE_FLAGS=(-DCMAKE_INSTALL_PREFIX=$prefix
+-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}
+-DCMAKE_BUILD_TYPE=Release
+-DBUILD_SHARED_LIBS=ON)
+
+# To actually use shared_mutex
+if [[ ${target} == *mingw* ]]; then
+    CMAKE_FLAGS+=(-DCMAKE_CXX_FLAGS="-std=c++17")
+fi
```
giordano, line 29: According to https://en.cppreference.com/w/cpp/thread/shared_mutex, `shared_mutex` was introduced in C++17.  If this function is used everywhere and not just for Windows, then this isn't Windows-specific.  Also, we should never set the standard in this way, the CMake file should set [`CMAKE_CXX_STANDARD`](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html).  In other words, this is an upstream bug.

giordano (commented):
I thought I submitted the review two days ago, but I didn't click the finish review button

R/RE2/build_tarballs.jl
evetion: I've reported the upstream bug and in the meantime here, I replace all occcurences of `cxx_std_11` in the CMakeLists.txt, which are used in multiple `target_compile_features` calls (`CMAKE_CXX_STANDARD` wasn't used).

PR #5442, opened 2022-09-07
Update GAP and GAP_lib to 4.12.0

fingolfin:


fingolfin:
macOS error *sigh*. Will think about it tomorrow
```
[22:46:57] /usr/bin/install -c -m 0755 -s gap /workspace/destdir/lib/gap
[22:46:57] error: /opt/x86_64-apple-darwin14/bin/x86_64-apple-darwin14-strip: unrecognized option: -p
[22:46:57] Usage: /opt/x86_64-apple-darwin14/bin/x86_64-apple-darwin14-strip [-AnuSXx] [-] [-d filename] [-s filename] [-R filename] [-o output] file [...] 
[22:46:57] install: strip: No such file or directory
```

fingolfin:
@giordano have you seen this kind of error before? The invocation is `/usr/bin/install -c -m 0755 -s gap /workspace/destdir/lib/gap` which I think is busybox; it ends up invoking `/opt/x86_64-apple-darwin14/bin/x86_64-apple-darwin14-strip` with option `-p` which it does not support; but note that `-p` does not come from the initial invocation. So something (busybox?) adds that, but why? can we avoid it?

giordano:
Sorry, I don't know where that comes from exactly (but I feel like I've seen something like that before).  Short-term solution might be not to strip symbols?

fingolfin:
(I'll patch out the `-s` here so I can move on, but it's annoying :-( )

G/GAP/bundled/patches/gcc12.patch
```diff
@@ -0,0 +1,23 @@
+commit 6f851174fe92c077e316005612e7246819434e0e
+Author: Max Horn <max@quendi.de>
+Date:   2022-09-14 00:12:02 +0200
+
+    kernel: workaround GCC 12 warning
```
giordano, line 5: Wait, are we building with GCC 12?

fingolfin: Not saying that, but the only place I've seen this warning before and can reproduce it is with GCC 12, not with GCC 11, so I cherry-picked that patch for use here.

It appeared only in the builds for macos-aarch64, see e.g. [this log](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=22159&view=logs&j=5d923a24-afbc-5e65-3a9d-e5b7b46b446b&t=7772c3a4-c19d-58f1-8481-c85c5a6fe3fe) and there only when invoking `/opt/bin/x86_64-linux-musl-cxx11/x86_64-linux-musl-gcc` (i.e. to compile code to be run during build time)

fingolfin: This is the native / build compiler (targeting linux), isn't that always gcc? Anyway I am not aware of anything in that recipe changing the compiler

PR #5443, opened 2022-09-08
MAGEMin  1.2.4: bugfix

boriskaus:
This fixes a bug that caused the wrong binary to be generated


M/MAGEMin/build_tarballs.jl
```diff
@@ -34,11 +34,16 @@ INC="-I${includedir}"
...
 
-# compile binary
+# Compile binary
 make -j${nproc} CC="${CC}" CCFLAGS="${CCFLAGS}" LIBS="${LIBS}" INC="${INC}" all
 
+# Windows sometimes adds an extension to the binary and sometimes not...
+if test -f "MAGEMin.exe"; then
+    mv MAGEMin.exe MAGEMin
+fi
+
```
giordano, line 44: Isn't there a way to set the name of the executable on the command line, instead of doing this?

boriskaus: Done (and passes the tests now)

PR #5447, opened 2022-09-09
Update rocRAND recipe to use common recipe structure

pxl-th:
Update for 4.5.2 will come after this PR.


R/rocRAND/common.jl
```diff
@@ -0,0 +1,101 @@
...
+mkdir ${prefix}/lib/include
+ln -s ${prefix}/hip/include/* ${prefix}/lib/include
+
+# NOTE
+# This is needed to avoid errors with zipping files older than 1980.
+# See: https://github.com/pypa/wheel/issues/418
+unset SOURCE_DATE_EPOCH
+pip install -U pip wheel setuptools
+
+amdgpu_targets="gfx900"
```
jpsamaroo, line 67: Can we expand this to include all GPUs supported by rocRAND?

jpsamaroo (approved):

R/rocRAND/common.jl
pxl-th: Done!

PR #5456, opened 2022-09-09
Use common builscript structure for rocPRIM

pxl-th:
PR for v4.5.2 will follow next.


giordano:
To be clear, does this actually change anything (apart from the reorganisation)?

pxl-th:
> To be clear, does this actually change anything (apart from the reorganisation)?

Except adding the license, no. Only reorganising for now

PR #5466, opened 2022-09-11
[GDAL] Use NetCDF and HDF5 when available.

### LINKED ISSUE #4950: Why GDAL is not build with NetCDF+HDF dependencies?
Any particular reason, or just because no one had interest on it so far?

evetion:
@visr 

Fixes #4950 (at least partially).

Uses the same approach as #5383, by only using NetCDF and HDF5 on the platforms where they are available. While this introduces a disparity between platforms, it helps the most common platforms (Windows, macOS, Linux gnu).

Platforms without HDF5 and NetCDF
- FreeBSD
- Linux
	- musl
	- i686
	- arm
	- powerpc

### DIFF
--- a/G/GDAL/build_tarballs.jl
+++ b/G/GDAL/build_tarballs.jl
@@ -6,8 +6,8 @@ name = "GDAL"
 upstream_version = v"3.5.1"
 version_offset = v"0.0.1"
 version = VersionNumber(upstream_version.major * 100 + version_offset.major,
-                        upstream_version.minor * 100 + version_offset.minor,
-                        upstream_version.patch * 100 + version_offset.patch)
+    upstream_version.minor * 100 + version_offset.minor,
+    upstream_version.patch * 100 + version_offset.patch)
 
 # Collection of sources required to build GDAL
 sources = [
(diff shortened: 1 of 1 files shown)


G/GDAL/build_tarballs.jl
```diff
@@ -6,8 +6,8 @@ name = "GDAL"
 upstream_version = v"3.5.1"
 version_offset = v"0.0.1"
 version = VersionNumber(upstream_version.major * 100 + version_offset.major,
-                        upstream_version.minor * 100 + version_offset.minor,
-                        upstream_version.patch * 100 + version_offset.patch)
+    upstream_version.minor * 100 + version_offset.minor,
+    upstream_version.patch * 100 + version_offset.patch)
```
giordano, line 10: This seems to be a whitespace-only change which has the only effect of making git-blaming harder?

G/GDAL/build_tarballs.jl
```diff
@@ -91,8 +107,10 @@ dependencies = [
...
     Dependency("libgeotiff_jll"; compat="100.700.100"),
     Dependency("LibCURL_jll"; compat="7.73"),
+    Dependency("NetCDF_jll"; compat="400.902.5"),
+    Dependency("HDF5_jll")
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat="1.6", preferred_gcc_version=v"8")
+    julia_compat="1.6", preferred_gcc_version=v"8")
```
giordano, line 116: And same here

G/GDAL/build_tarballs.jl
```diff
@@ -91,8 +107,10 @@ dependencies = [
     Dependency("Libtiff_jll"; compat="4.3"),
     Dependency("libgeotiff_jll"; compat="100.700.100"),
     Dependency("LibCURL_jll"; compat="7.73"),
+    Dependency("NetCDF_jll"; compat="400.902.5"),
+    Dependency("HDF5_jll")
```
giordano, line 111:
```suggestion
    Dependency("HDF5_jll"),
```

G/GDAL/build_tarballs.jl
```diff
@@ -91,6 +107,8 @@ dependencies = [
     Dependency("Libtiff_jll"; compat="4.3"),
     Dependency("libgeotiff_jll"; compat="100.700.100"),
     Dependency("LibCURL_jll"; compat="7.73"),
+    Dependency("NetCDF_jll"; compat="400.902.5"),
+    Dependency("HDF5_jll"),
```
giordano, line 111: While you're changing the dependencies, it'd be better to change the version number, for example by bumping the version offset.

visr (approved):

G/GDAL/build_tarballs.jl
giordano, line 111: Since these packages aren't available on all platforms, I'd prefer to use them only when available, which would also clear the dozens of warnings about these packages not being available: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=22093&view=results.  Filtering them properly though is a bit tricky, but I don't have the time to do it now

PR #5478, opened 2022-09-12
First version of TestU01 for Yggdrasil

andreasnoack:
This is an attempt to migrate https://github.com/andreasnoack/TestU01Builder to Yggrasil. It doesn't work yet.

### DIFF
--- a/T/TestU01/build_tarballs.jl
+++ b/T/TestU01/build_tarballs.jl
@@ -20,24 +20,36 @@ script = raw"""
 cd ${WORKSPACE}/srcdir/TestU01*
 update_configure_scripts
 
-# Always export this since the only executable we build is a host-tool
-export EXEEXT=""
+MFLAGS=(
+    -j${nproc}
+
+    # Always set this since the only executable we build is a host-tool
+    # and we don't want to accidentally try to rebuild it.
+    EXEEXT=""
+)
+
 if [[ "${target}" = *-mingw* ]]; then
   export lt_cv_deplibs_check_method=pass_all
   export LDFLAGS="-lws2_32"
 fi
+
 ./bootstrap
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
 
 # Make `tcode` to run on the host, as it's used to bootstrap
-make -j${nproc} CC="${HOSTCC}" LD="${HOSTLD}" tcode
+make ${MFLAGS[@]} CC="${HOSTCC}" LD="${HOSTLD}" LDFLAGS="" tcode
+
+# libtool is hopeless, I give up
+if [[ "${target}" = *-mingw* ]]; then
+    cp .libs/tcode tcode
+fi
 
 # Make the library we actually want to ship
-make -j${nproc}
-make -j${nproc} install
+make ${MFLAGS[@]}
+make ${MFLAGS[@]} install
 
 # Compile TestU01extractors shim
-make -j${nproc} -C ../src/TestU01extractors/ install
+make ${MFLAGS[@]} -C ../src/TestU01extractors/ install
 
 install_license COPYING
 """
--- a/T/TestU01/build_tarballs.jl
+++ b/T/TestU01/build_tarballs.jl
@@ -51,6 +51,9 @@ make ${MFLAGS[@]} install
 # Compile TestU01extractors shim
 make ${MFLAGS[@]} -C ../src/TestU01extractors/ install
 
+# Delete improperly-installed `tcode`
+rm ${bindir}/tcode
+
 install_license COPYING
 """
 
--- a/T/TestU01/build_tarballs.jl
+++ b/T/TestU01/build_tarballs.jl
@@ -1,0 +1,60 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "TestU01"
+version = v"1.2.3"
+
+# Use unofficial git mirror instead of official URL, as it is unversioned and may be overwritten in the future.
+# Manually verified to be bit-for-bit identical (except for a missing empty `bin/` directory) to official v1.2.3 release.
+sources = [
+    GitSource(
+        "https://github.com/blep/TestU01.git",
+        "8e39bd74544f2b6d857a59e2787a5f1c87cc4313"
+    ),
+    DirectorySource("./bundled"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/TestU01*
+update_configure_scripts
+
+# Always export this since the only executable we build is a host-tool
+export EXEEXT=""
+if [[ "${target}" = *-mingw* ]]; then
+  export lt_cv_deplibs_check_method=pass_all
+  export LDFLAGS="-lws2_32"
+fi
+./bootstrap
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+
+# Make `tcode` to run on the host, as it's used to bootstrap
+make -j${nproc} CC="${HOSTCC}" LD="${HOSTLD}" tcode
+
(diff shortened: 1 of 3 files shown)


T/TestU01/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "TestU01"
+version = v"1.2.3-1"
```
giordano, line 6: Please stick to `X.Y.Z`

T/TestU01/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+# Collection of sources required to build TestU01
+sources = [
+    ArchiveSource(
+        "http://simul.iro.umontreal.ca/testu01/TestU01.zip",
+        "bc1d1dd2aea7ed3b3d28eaad2c8ee55913f11ce67aec8fe4f643c1c0d2ed1cac"
```
giordano, line 12: This doesn't look like a good source: it isn't versioned, if the author will publish a new version the sha256 hash will become invalid.  Is this the only source we have?

T/TestU01/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+    )
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd TestU01-1.2.3/
```
giordano, line 19:
```suggestion
cd $WORKSPACE/srcdir/TestU01*
```

T/TestU01/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+update_configure_scripts
+if [ $target = "x86_64-w64-mingw32" ] || [ $target = "i686-w64-mingw32" ]; then
+  lt_cv_deplibs_check_method=pass_all ./configure --prefix=$prefix --host=$target LDFLAGS="-L/opt/$target/$target/lib/ -lws2_32"
+else
+  ./configure --prefix=$prefix --host=$target
+fi
```
giordano, line 25:
```suggestion
if [[ "${target}" = *-mingw* ]]; then
  export lt_cv_deplibs_check_method=pass_all
  export LDFLAGS="-lws2_32"
fi
./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
```

T/TestU01/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+make -j${nproc} EXEEXT=""
+make install EXEEXT=""
+# Compile TestU01extractors shim
+cd $WORKSPACE/srcdir/TestU01extractors/
+make install SHLIB_EXT=$dlext
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [AnyPlatform()]
```
giordano, line 37: Why [`AnyPlatform`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.AnyPlatform)?  That's meant for platform-independent tarballs, shared libraries are very much platform-dependent.

T/TestU01/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libtestu01", :libtestu01),
+    LibraryProduct("libprobdist", :libprobdist),
+    LibraryProduct("libtestu01extractors", :libtestu01extractors),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
```
giordano, line 47:
```suggestion
dependencies = Dependency[
```

T/TestU01/src/TestU01extractors/extractors.c
```diff
@@ -0,0 +1,143 @@
...
+   }
+}
+
+void getPValSspectral (sspectral_Res *res, double pvals[])
+{
+   for (int i = 0; i < 11; ++i)
+   {
+      pvals[i] = res->Bas->pVal2[i];
+   }
+}
```
giordano, line 143:
```suggestion
}

```

T/TestU01/src/TestU01extractors/Makefile
```diff
@@ -0,0 +1,13 @@
+ifeq ($(SHLIB_EXT), dll)
+	libdir=$(prefix)/bin
+else
+	libdir=$(prefix)/lib
+endif
+
+install:
```
giordano, line 7:
```suggestion
install:
```

T/TestU01/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+name = "TestU01"
+version = v"1.2.3-1"
+
+# Collection of sources required to build TestU01
+sources = [
+    ArchiveSource(
+        "http://simul.iro.umontreal.ca/testu01/TestU01.zip",
+        "bc1d1dd2aea7ed3b3d28eaad2c8ee55913f11ce67aec8fe4f643c1c0d2ed1cac"
+    )
```
giordano, line 13: I think you wanted to add the `src/` directory as additional source
```suggestion
    ),
    DirectorySource("./src"),
```

T/TestU01/build_tarballs.jl
staticfloat: Andreas probably confused this with `supported_platforms()` :)

T/TestU01/build_tarballs.jl
giordano: @staticfloat related: https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1223 :slightly_smiling_face:

T/TestU01/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+    LibraryProduct("libprobdist", :libprobdist),
+    LibraryProduct("libtestu01extractors", :libtestu01extractors),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_version=v"1.6")
```
giordano, line 52:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat=v"1.6")
```

T/TestU01/build_tarballs.jl
andreasnoack: I was following the version of the source code (which is 1.2.3). Wasn't that policy at some point? Did people get frustrated by all the uncontrolled JLL updates? Since we might make several versions of the wrappers here, we shouldn't use `1.2.3` if we can't append a postfix. Let's then just make it `1.0.0`.

T/TestU01/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+using BinaryBuilder
+
+name = "TestU01"
+version = v"1.2.3"
+
+# Use unofficial git mirror instead of official URL, as it is unversioned and may be overwritten in the future.
+# Manually verified to be bit-for-bit identical (except for a missing empty `bin/` directory) to official v1.2.3 release.
+sources = [
+    GitSource(
+        "https://github.com/blep/TestU01.git",
```
andreasnoack, line 12: The official git repo is https://github.com/umontreal-simul/TestU01-2009 but there are no releases or tags.

andreasnoack: Doh. I should just read the comment. Never mind.

giordano:
The license file is missing

giordano:
`tcode` should probably not installed?

staticfloat:
Yep, agreed.  I guess I'll have to delete it from the output.

andreasnoack:
Thanks for the help here.

PR #5479, opened 2022-09-12
Recompile PROPACK with LBT and new platforms

amontoison:
@dpo


amontoison:
@giordano 
`LBT` is not found on Windows.

I suppose that it's also the case when we compile `MUMPS`.

P/PROPACK/build_tarballs.jl
```diff
@@ -12,10 +13,9 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/PROPACK-*/
 
-OPENBLAS="-lopenblas"
+LBT="-lblastrampoline"
```
amontoison, line 16:
```suggestion
LBT="-L${libdir} -lblastrampoline"
```

ViralBShah (approved):

Issue #5468, opened 2022-09-12
support CoolProp on Apple M1


longemen3000:
seems that the result binary does not include apple M1 in its dependencies. ref: https://github.com/CoolProp/CoolProp.jl/issues/20

longemen3000:
seems that it only needs a rebuild? the version is the same, and the last commit was two weeks before the launch of apple M1

giordano:
Did you see the pinned issue  #2763?

longemen3000:
the mention of apple was cut short hahaha 
![image](https://user-images.githubusercontent.com/38795484/189779722-10561742-c19d-48a4-8de4-efa2056dbf4f.png)
in particular, i don't see any problems with the constraints imposed by the issue, but i will tag @barche

longemen3000:
(i was thinking on using the standalone binary meanwhile, is that possible)

barche:
I made a PR with 6.4.2, we'll see how it goes :)

PR #5482, opened 2022-09-13
[CoolProp] Add Apple M1 support

barche:
Include M1 support (Issue #5468)


C/CoolProp/build_tarballs.jl
```diff
@@ -29,7 +32,7 @@ install_license $WORKSPACE/srcdir/CoolProp*/LICENSE
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = supported_platforms(; experimental=true)
```
giordano, line 35:
```suggestion
platforms = supported_platforms()
```

barche: Was wondering about that, looks like we both made the change :)

PR #5490, opened 2022-09-14
[Doxygen] upgrade to version 1.9.5

marcom:


marcom:
See https://github.com/doxygen/doxygen/issues/9557

giordano:
Out of curiosity, what do you need this for?

marcom:
I'm making a build for OpenMM (https://openmm.org/), which refuses to build without doxygen.

Then i saw that doxygen had a new version out, so i wanted to bump the doxygen build (and it all seemed to work on x86_64-linux).

OpenMM actually builds just fine with the old doxygen version, so this isn't really necessary.

PR #5495, opened 2022-09-15
Add ReLAPACK in OpenBLAS

amontoison:

* Package name: ReLAPACK


R/ReLAPACK/bundled/patches/relapack.patch
```diff
@@ -0,0 +1,107 @@
...
+diff --git before/ReLAPACK/src/relapack.h after/ReLAPACK/src/relapack.h
+index 2cb061c..b33e1fb 100644
+--- before/ReLAPACK/src/relapack.h
++++ after/ReLAPACK/src/relapack.h
+@@ -7,14 +7,14 @@
+ 
+ // add an underscore to BLAS routines (or not)
+ #if BLAS_UNDERSCORE
+-#define BLAS(routine) routine ## _
++#define BLAS(routine) routine ## _64
```
amontoison, line 96:
```suggestion
+#define BLAS(routine) routine ## _
```

R/ReLAPACK/bundled/patches/relapack.patch
```diff
@@ -0,0 +1,107 @@
...
+ #else
+ #define BLAS(routine) routine
+ #endif
+ 
+ // add an underscore to LAPACK routines (or not)
+ #if LAPACK_UNDERSCORE
+-#define LAPACK(routine) routine ## _
++#define LAPACK(routine) routine ## _64
```
amontoison, line 104:
```suggestion
+#define LAPACK(routine) routine ## _
```

R/ReLAPACK/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("librelapack", :librelapack)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"))
+]
```
amontoison, line 41:
```suggestion
dependencies = [
]
```

R/ReLAPACK/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+cd ReLAPACK
+make
+cd src/
+${CC} -shared -o librelapack.${dlext} *.o -lblastrampoline -lm
```
amontoison, line 23:
```suggestion
${CC} -shared -o librelapack.${dlext} *.o
```

R/ReLAPACK/bundled/patches/relapack.patch
```diff
@@ -0,0 +1,107 @@
...
+ #define BLAS(routine) routine
+ #endif
+ 
+ // add an underscore to LAPACK routines (or not)
+ #if LAPACK_UNDERSCORE
+-#define LAPACK(routine) routine ## _
++#define LAPACK(routine) routine ## _
+ #else
+ #define LAPACK(routine) routine
+ #endif
```
amontoison, line 107:
```suggestion
```

R/ReLAPACK/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+cd $WORKSPACE/srcdir
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+cd ReLAPACK
+make
+cd src/
+${CC} -shared -o librelapack.${dlext} *.o
+mv librelapack.${dlext} ${libdir}
+mv ../inc/relapack.h ${includedir}
```
amontoison, line 25:
```suggestion
mv librelapack.${dlext} ${libdir}/ibrelapack.${dlext}
cp ../inc/relapack.h ${includedir}
```

R/ReLAPACK/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+cd $WORKSPACE/srcdir
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+cd ReLAPACK
+make
+cd src/
+${CC} -shared -o librelapack.${dlext} *.o
+mv librelapack.${dlext} ${libdir}/ibrelapack.${dlext}
+cp ../inc/relapack.h ${includedir}
```
amontoison, line 25:
```suggestion
mkdir -p ${libdir}
mv librelapack.${dlext} ${libdir}/ibrelapack.${dlext}
mkdir -p ${includedir}
cp ../inc/relapack.h ${includedir}
```

R/ReLAPACK/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+cd $WORKSPACE/srcdir
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+cd ReLAPACK
+make
+cd src/
+${CC} -shared -o librelapack.${dlext} *.o
+mkdir -p ${libdir}
+mv librelapack.${dlext} ${libdir}/ibrelapack.${dlext}
```
amontoison, line 25:
```suggestion
mv librelapack.${dlext} ${libdir}/librelapack.${dlext}
```

amontoison:
@giordano @staticfloat 
I would like to add `ReLAPACK` in the artifact of OpenBLAS.
One routine is very useful, it's `gemmt` and it is also provided by `MKL`.
It will update one triangle of `C` such that `C <- alpha * A * B + beta * C`.
It's relevant when `C` is hermitian and `A` = `B'` for instance. 
It also adds some variants of the LAPACK routines.

With this PR, the symbol `gemmt` will be added and the symbols of the other routines will have a prefix `RELAPACK`.
I did this modification because the other routines are variants of the LAPACK routines with the same name.
We will be also able to switch between the `gemmt` of `OpenBLAS` or `MKL` with `LBT`.

The current issue with the compilation is that ReLAPACK compiles `.o` files with the same names than LAPACK:
https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=22259&view=logs&j=ba03ae70-8d82-5361-1127-49c7c21d0cb8&t=d960f7f1-61ec-593e-1474-ee3577b9ed1d&l=22560
What is the simplest modification to not overwrite the `.o` of LAPACK?
I suppose that we can just add the prefix `RELAPACK` for all `.o` files generated by ReLAPACK.

giordano:
OpenBLAS is a dependency of Julia and I have zero idea about the consequences of this change.  I'm a bit worried about all those patches, they seem to change lots of code, and every time we want to upgrade OpenBLAS we'd need to refresh them, that's not exactly maintainable.

ViralBShah:
In the past there has been huge resistance to adding more libraries to the Julia distribution (like for example when I tried to enabled the use of METIS for SuiteSparse). I think the best would be to compile relapack as a standalone library, which will be much easier to manage.

amontoison:
Mosè, this PR just adds 10 new routines from ReLAPACK.
If they merge my PR: https://github.com/xianyi/OpenBLAS/pull/3770, you will not need to update the patches related to ReLAPACK in the future.

@ViralBShah
`ReLAPACK` was added in `OpenBLAS` and they maintain it.

The main goal is to add the `gemmt` routine, which is also provided by `MKL`.
I plan to add it in Julia and be able to switch the vendor with LBT.
It's for this reason that I didn't compiled relapack as a standalone library.

ViralBShah:
Ok, I'll review this closely over the weekend. But if we do want to merge this, we also will have to do the appropriate build changes in julia's source build of openblas too.

amontoison:
I started to interface `gemmt` in Julia: https://github.com/JuliaLang/julia/commit/bfc85e30341045e73613e633f22cbb7abe8f4e3c
They added it with the release 0.3.21: https://github.com/xianyi/OpenBLAS/issues/861

ViralBShah:
I assume we still want to get this merged, right? Sorry, it just fell off the radar for me, but hope to come back to it shortly.

amontoison:
Yes, it will be great!
In a nutshell, we compile OpenBLAS with ReLAPACK to have the gemmt routine:
```julia
if [ ${version_patch} -ge 21 ]; then
  flags+=(BUILD_RELAPACK=1)
fi
```
but we don't want that the other routines of `ReLAPACK` replace the routine of `LAPACK`:
`#define INCLUDE_ALL 0` should be used instead of `#define INCLUDE_ALL 1` in `relapack/config.h`.

The goal of the other patch is to avoid that the `.o` files of LAPACK are replaced by those of ReLAPACK because they have the same names.
It will probably be fixed upstream: https://github.com/xianyi/OpenBLAS/pull/3770.

giordano:
I'd prefer to get this sorted out correctly upstream first.

amontoison:
They added the `gemmt` routine in OpenBLAS v"3.22" (https://github.com/xianyi/OpenBLAS/pull/3796). We don't need to compile ReLAPACK!
Mosè already opened a PR to compile the new version of OpenBLAS (https://github.com/JuliaPackaging/Yggdrasil/pull/6457).

PR #5501, opened 2022-09-15
[libxcrypt] add new recipes

benlorenz:
`libxcrypt_legacy_jll` can be used as `RuntimeDependency` for packages that link to `libcrypt.so.1` from the old glibc that is used for building in BinaryBuilder (this is only built for glibc based platforms).
I tested this once by deploying this locally (x86_64-linux-gnu without libcrypt.so.1), loaded the `libxcrypt_legacy_jll` and then successfully loaded the (now obsolete) `Python_jll-3.8.8+2` with the libcrypt problem.

`libxcrypt_jll` which contains `libcrypt.so.2` can be used as a normal dependency for programs that need the crypt api, but this is untested. I am not sure how to make sure that this library will be used instead of the code from the old glibc, but maybe it just works.

This wont compile for mingw platforms, apart from a few minor errors that I was able to work around, the portability notes state:

> libxcrypt should be buildable with any ISO C1999-compliant C compiler, with one critical exception: the symbol versioning macros in crypt-port.h only work with compilers that implement certain GCC and GNU Binutils extensions (__attribute__((alias)), GCC-style asm, and .symver).

And GNU binutils ld supports symbol versions only for ELF platforms.

### DIFF
--- a/L/libxcrypt/common.jl
+++ b/L/libxcrypt/common.jl
@@ -1,0 +1,60 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+function build_libxcrypt(ARGS; legacy::Bool)
+   version = v"4.4.28"
+   name = "libxcrypt"
+
+   # Collection of sources required to build libxcrypt
+   sources = [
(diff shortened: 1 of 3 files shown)


giordano:
This is brilliant, thanks a lot!  Presumably `libxcrypt_legacy` could also be a regular dependency, right?

PR #4177, opened 2022-01-05
[hsa_rocr] Do not install legacy symlinks

giordano:
The symlinks are created for legacy reasons we don't care about, for us they
only cause confusions with RUNPATHs:
https://github.com/JuliaPackaging/Yggdrasil/pull/4077#issuecomment-1000500072

CC: @jpsamaroo

PR #4077, opened 2021-12-17
[ROCmOpenCLRuntime] Update to 4.2.0

jpsamaroo:

PR #5512, opened 2022-09-16
Update HIP buildscript & disable TLS on musl

pxl-th:
- For 4.5.2+ ROCclr is no longer should be built as a standalone project.
We just need to pull the repo and point HIP to it.
- Add ROCclr patch for musl: https://github.com/ROCm-Developer-Tools/ROCclr/blob/442ede037c871420f3c810cb4228f5ebc2a133bb/thread/thread.cpp#L124


jpsamaroo:
@mose the dlopen failure is not reproducible locally if we set `dont_dlopen=true`; any ideas?

pxl-th:
If importing ROCmCompiler support before dlopening, there is no issues:

```julia
pxl-th@Yotun:~/code/Yggdrasil/H/HIP/HIP@5.2.3/build/x86_64-linux-gnu-cxx11/wzvykAp4/destdir/hip/lib$ jlt
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.0-DEV.1383 (2022-09-16)
 _/ |\__'_|_|_|\__'_|  |  Commit a7bef773282 (0 days old master)
|__/                   |

julia> using Libdl

julia> using ROCmCompilerSupport_jll

julia> Libdl.dlopen("./libamdhip64.so")
Ptr{Nothing} @0x0000000002c9a9f0
```

If not importing, we get error:

```julia
pxl-th@Yotun:~/code/Yggdrasil/H/HIP/HIP@5.2.3/build/x86_64-linux-gnu-cxx11/WAQ1RO1d/destdir/hip/lib$ jlt
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.0-DEV.1383 (2022-09-16)
 _/ |\__'_|_|_|\__'_|  |  Commit a7bef773282 (0 days old master)
|__/                   |

julia> using Libdl

julia> Libdl.dlopen("./libamdhip64.so")
ERROR: could not load library "./libamdhip64.so"
libamd_comgr.so.2: cannot open shared object file: No such file or directory
Stacktrace:
 [2] dlopen (repeats 2 times)
   @ ./libdl.jl:116 [inlined]
 [3] top-level scope
   @ REPL[2]:1
```

giordano:
I suggested it in Slack, but I'm copying the suggestion also here: to debug this, you have to try and dlopen the library directly in the build environment, after audit fails.  For example, there may be some files in the build environment which confuses the loader.  See for example https://github.com/JuliaPackaging/Yggdrasil/pull/5028#issuecomment-1165945565 and following messages, there I shared some commands I use to peak into the auditor.

pxl-th:
I'll remove 5.2.3, but wanted to show that it is fixed.
In 5.2.3 they [hardcode](https://github.com/ROCm-Developer-Tools/hipamd/blob/02187ecfc8a1a6ed2bbca29000a83dec5a61cb7f/CMakeLists.txt#L42) RPATH using build-time variables.

Instead I've replaced RPATH to use `$ORIGIN` with relative paths.
Not sure if this is the cleanest way.
@jpsamaroo maybe doing this for 4.5.2 will solve you issue on musl as well?
I've noticed that 4.5.2 has empty RPATH. We can try adding the same thing.

giordano:
Absolute rpaths are definitely bad, but the auditor should already fix them up: https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/c3ab9105f9f4787eda1deee5586a58738d4c9f40/src/auditor/dynamic_linkage.jl#L419-L519.  If that's not the case, perhaps there's something we need to fix in the audit pass?

For the record, we had already a good deal of troubles with rpaths in related packages in the past, see #4077, #4109, #4177.

pxl-th:
> Absolute rpaths are definitely bad, but the auditor should already fix them up: https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/c3ab9105f9f4787eda1deee5586a58738d4c9f40/src/auditor/dynamic_linkage.jl#L419-L519. If that's not the case, perhaps there's something we need to fix in the audit pass?
> 
> For the record, we had already a good deal of troubles with rpaths in related packages in the past, see #4077, #4109, #4177.

I've looked at the auditor code and `update_linkage` function is never called on `libamdhip64.so`.

It is filtered out in [`collapse_symlinks`](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/c3ab9105f9f4787eda1deee5586a58738d4c9f40/src/Auditor.jl#L76). [`bin_files`](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/c3ab9105f9f4787eda1deee5586a58738d4c9f40/src/Auditor.jl#L75) contains only `.../libamdhip64.so` & `.../libamdhip64.so.5` which are symlinks to `.../libamdhip64.so.5.2.3-0` (which is not present in `bin_files`).

pxl-th:
@jpsamaroo I added a patch for RPATH in 4.5.2, can you test it to see if it fixes your issues on musl?

jpsamaroo:
It doesn't fix my audit issues, but I also don't think we need to worry about it right now. I'm :+1: to merging and figuring it out later, since this unblocks me from being able to test with HIP locally.

pxl-th:
> It doesn't fix my audit issues, but I also don't think we need to worry about it right now. I'm +1 to merging and figuring it out later, since this unblocks me from being able to test with HIP locally.

OK, I've removed 5.2.3 for now and add it in the next PR.
This one fixes other issues. I'll also open a PR for 4.2.0 that will disable TLS on musl.

jpsamaroo (approved):

giordano:
I'm going to merge this, but is this perhaps similar to #4177 where useless symlinks where only causing confusion, and the cleanest solution was to simply not install them?

pxl-th:
> I'm going to merge this, but is this perhaps similar to #4177 where useless symlinks where only causing confusion, and the cleanest solution was to simply not install them?

I don't know the entire logic of the auditor, but shouldn't we follow symlink until we hit non symlink file and try to fix rpaths for that?

Here's the place where they create those symlinks btw: https://github.com/ROCm-Developer-Tools/hipamd/blob/3e1ca054b0239170dd763059955e447802e9e5c9/hip-backward-compat.cmake#L145.
Looks more like they just create versioned `.so` objects.

PR #428, opened 2020-01-24
Wizard recipe: SDL2_image-v2.0.5

aviks:
This pull request contains a new build recipe I built using the BinaryBuilder.jl wizard:

* Package name: SDL2_image
* Version: v2.0.5

@staticfloat please review and merge.

PR #5514, opened 2022-09-16
Update SDL2_image, and build for arm mac

### LINKED ISSUE #5484: SDL2_image_jll for aarch64-apple-darwin
It would be great if an Apple Silicon build of SDL2_image_jll could be provided.

#5481

aviks:
Will fix #5484

### DIFF
--- a/S/SDL2_image/build_tarballs.jl
+++ b/S/SDL2_image/build_tarballs.jl
@@ -14,15 +14,11 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/SDL2_image-*/
-./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} --with-pic
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DSDL2_IMAGE_WEBP=ON -DSDL2_IMAGE_TIF=ON SDL2IMAGE_BACKEND_IMAGEIO=OFF ..
 make -j${nproc}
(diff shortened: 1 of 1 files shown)


aviks:
OK, I don't understand why mac (x64 and arm) builds are failing. It seems to be erroring in compiling something from the system framework. The previous version of this package seems to have compiled fine in #428 though that was a while ago.

aviks:
So this seems to be failing to compile the OSX system framework.  Which seems strange. Is it using the wrong compiler somehow? 

```
/opt/aarch64-apple-darwin20/bin/../aarch64-apple-darwin20/sys-root/System/Library/Frameworks/Foundation.framework/Headers/NSObject.h:19:21: error: unknown type name ‘nullable’
[18:20:18]    19 | - (id)copyWithZone:(nullable NSZone *)zone;
[18:20:18]       |                     ^~~~~~~~
[18:20:18] /opt/aarch64-apple-darwin20/bin/../aarch64-apple-darwin20/sys-root/System/Library/Frameworks/Foundation.framework/Headers/NSObject.h:19:29: error: expected ‘)’ before ‘NSZone’
[18:20:18]    19 | - (id)copyWithZone:(nullable NSZone *)zone;
[18:20:18]       |                             ^~~~~~~
[18:20:18]       |                             )

```

aviks:
So using Cmake rather than autoconf makes the build work for macs. There are a few warning from the auditor on apple, but I'm not sure if I should be worried about those? 

Also, I'm wondering if all the SDL2_* builds should now be moved to cmake?

giordano:
> There are a few warning from the auditor on apple, but I'm not sure if I should be worried about those?

I don't see any warnings (`@warn` log messages) from the auditor?  All other messages are fine

> Also, I'm wondering if all the SDL2_* builds should now be moved to cmake?

Maybe next time we update them, but I wouldn't touch existing builds.

PR #5522, opened 2022-09-17
New recipe: quarto

aviks:
Repackages the entire upstream distribution from https://quarto.org/docs/download/. That distribution bundles `pandoc` and `deno`, among other things. 

Supports limited platforms, only those for which upstream binaries exist:
*  Platform("x86_64", "linux")
*  Platform("x86_64", "macos")
* Platform("x86_64", "windows")

Only lightly tested. Can successfully generate plain html and docx. PDF requires installing xelatex. Quarto attempts to install it. Code execution requires a jupyter installation plus a kernel. 

#### Example usage: 

```
using quarto_jll

quarto() do quarto
       run(`$quarto render test.qmd --to html`)
end

quarto() do quarto
       run(`$quarto render test.qmd --to docx`)
end
```

Replaces #5269

Might be good to integrate with https://github.com/quarto-dev/quarto-julia at some point.

### DIFF
--- a/Q/quarto/build_tarballs.jl
+++ b/Q/quarto/build_tarballs.jl
@@ -1,0 +1,47 @@
+using BinaryBuilder, Pkg
+
+# Collection of pre-build quarto binaries
+name = "quarto"
+quarto_ver = "1.1.251"
+version = VersionNumber(quarto_ver)
+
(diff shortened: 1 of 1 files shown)


giordano:
About running executables, we prefer the syntax
```julia
using quarto_jll

run(`$(quarto()) render test.qmd --to html`)

run(`$(quarto()) render test.qmd --to docx`)
```
because that's thread-safe: https://docs.binarybuilder.org/stable/jll/#ExecutableProduct

PR #5524, opened 2022-09-18
New Recipe: LibTeXPrintf v1.11.0

Suavesito-Olimpiada:
I want to make a wrapper library of [libtexprintf](https://github.com/bartp5/libtexprintf/) for the use of Latexify.jl in the terminal.


giordano:
Uhm, the first build had some platforms failing with errors like
```
[09:23:47] mv: can't rename '.deps/drawbox.Tpo': No such file or directory
[...]
[09:23:47] mv: can't rename '.deps/parser.Tpo': No such file or directory
```
I simply restarted the failing jobs and now they're successful.  This is often an indication that the makefile is broken, may not be tracking some dependencies, and so when building with high parallelism as we do (`make -j16`, it isn't even _that_ high) a file may be requested before it's actually generated.  Sigh.

PR #5527, opened 2022-09-18
[Rust] Bump to Rust v1.64.0

jeremiahpslewis:
- [x] Build script tested locally
- [ ] After this PR, follow up with BinaryBuilder `Artifacts.toml` update


.devcontainer/Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM julia:1.7
 
-RUN apt-get update && apt-get install -y xz-utils bzip2 sudo git unzip
```
giordano, line 3: This change seems unrelated

jeremiahpslewis: On second thought, think this solves it a bit more elegantly: https://github.com/JuliaPackaging/Yggdrasil/pull/5541

giordano:
For the record, it appears that it's Rustup v1.25.1 which completely screws up our toolchain.  Downgrading to Rustup v1.24.3 I get again a working compiler.

PR #5533, opened 2022-09-19
add `texfree` function for freeing memory

Suavesito-Olimpiada:
This adds a `texfree` function, so that it can be called from Julia to free memory.

@giordano

### DIFF
--- a/L/LibTeXPrintf/build_tarballs.jl
+++ b/L/LibTeXPrintf/build_tarballs.jl
@@ -15,6 +15,7 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/libtexprintf/
 atomic_patch -p1 ${WORKSPACE}/srcdir/patches/win-fix.diff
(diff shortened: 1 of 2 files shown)


L/LibTeXPrintf/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "LibTeXPrintf"
-version = v"1.11.0"
+version = v"1.14.0"
```
giordano, line 6: Source didn't change?

Suavesito-Olimpiada: No, but they tagged a [new release](https://github.com/bartp5/libtexprintf/releases/tag/v1.14) to the current main. The answer to "why do you not set the source to the release?" is that in the release file, the `autogen.sh` file is not present, and the patch for `configure.ac` does not apply because they have already run it.

PR #5536, opened 2022-09-20
[Boost] upgrade to 1.80.0

lucifer1004:


lucifer1004:
Don't know why it failed on Linux GNU @giordano

giordano:
Do you need boost 1.80 specifically for something?  Otherwise, I'd suggest ***not*** upgrading it: boost breaks ABI in every single version (#3024), so the more versions we have around, the more fractured the ecosystem is.

lucifer1004:
OK, I do not need the version specifically. I just wanted to try upgrading some old packages.

PR #5540, opened 2022-09-20
[HIP 5.2.3] Disable legacy folder structure

pxl-th:
Disable legacy folder structure using `-DFILE_REORG_BACKWARD_COMPATIBILITY=OFF`.
Otherwise it creates a duplicate library directory which confuses rocBLAS and other libraries.
This is only in 5.2.3.


pxl-th:
Would be good to have it merged as it will unblock PRs for other 5.2.3 packages.

PR #5544, opened 2022-09-20
Add rocBLAS 5.2.3

pxl-th:


pxl-th:
Not sure why `musl` OOM's, `glibc` builds normally with any amount of threads.
I tried reducing number of targets to compile agains down to `1` and it compiles on `musl` successfully.

pxl-th:
Limited number of targets to build against on `musl` for now.
Disabled Navi 1-3 architectures.

pxl-th:
We decided to leave out musl for now. So this should be good to go.

PR #5549, opened 2022-09-21
Fix rocBLAS library install path

pxl-th:


giordano:
If you want to rebuild a package, you have to edit its `build_tarballs.jl` file.

giordano:
I think the job is stalling?

pxl-th:
> I think the job is stalling?

Doesn't look like it is stalled. New logs are appearing. It takes on average ~3 hours to finish.

giordano:
Uh, when I opened the log it was stuck at a couple of hours ago, but now I opened again and it is indeed moving.  Ok, good to know it does take a while.

giordano:
To be clear, this is actually only changing the generated JLL package, but not the actual content of the tarball, right?  But I'm a bit confused how this worked before if the path was wrong.

pxl-th:
> To be clear, this is actually only changing the generated JLL package, but not the actual content of the tarball, right? But I'm a bit confused how this worked before if the path was wrong.

Yes. It looks like rocBLAS is installed originally in `/lib` and then in `/rocblas/lib` it symlinks to `/lib` for some reason...
But we need to use ones in the `/lib`, because they have correct RPATHS.
I think, in 5.2.3 they added some dependency files that they look for during runtime (some `Tensile.yaml` files) and specifically assume that `librocblas.so` is in `/lib` and not in `/robclas/lib`.

pxl-th:
Or maybe they do not assume it explicitly...
I've noticed that before auditor `librocblas.so` has `["/workspace/destdir/hip/lib", "/workspace/destdir/lib"]` rpaths.
But after auditor fixes rpaths, `librocblas.so` only has `[$ORIGIN:$ORIGIN/../hip/lib]` rpath, which is correct if the library is in `/lib`, but not in `/rocblas/lib`.

PR #5552, opened 2022-09-22
MAGEMin 1.2.7

NicolasRiel:
MAGEMin 1.2.7
- added fO2 calculation
- update of MAGEMin_EquilibriumPath.mlx
- set mol fraction to 1e-4 when conversion from wt to mol fraction results in mol fraction < 1e-4 for oxides =/= H2O. This avoids issues when reducing the number of system components.


M/MAGEMin/build_tarballs.jl
```diff
@@ -10,7 +10,7 @@ version = v"1.2.4"
 
 # Collection of sources required to complete build
 sources = [GitSource("https://github.com/ComputationalThermodynamics/MAGEMin", 
-                    "2bbe31fbc50827e4442ec971598ea627f3773e7b")
+                    "604ff09eb08ee2b1ac4eaa177b9c30639afb042b")
```
giordano, line 13: Need to change also the version number, right?

NicolasRiel: Yes, thank you, I keep on forgetting this part. My bad!

giordano: This is a reason why I prefer the archive source: it's easier to ensure version number and actual source are in sync!

PR #5565, opened 2022-09-23
adding build for fmt

cmhamel:


F/Fmt/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+script = raw"""
+INSTALL_PATH=${prefix}
+cd $WORKSPACE/srcdir/fmt
+mkdir build
+cd build
+cmake \
+-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+-DBUILD_SHARED_LIBS=True \
+-DCMAKE_INSTALL_PREFIX=${INSTALL_PATH} \
+..
```
giordano, line 23: Why defining a variable, used only once, equal to another variable?  Why not directly using that variable, like in all other recipes here?  Also, you keep not following the [documentation](https://docs.binarybuilder.org/stable/build_tips/#CMake-builds), use also `-DCMAKE_BUILD_TYPE=Release`: 
```suggestion
cd $WORKSPACE/srcdir/fmt
mkdir build
cd build
cmake \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=True \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    ..
```

F/Fmt/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+make
+make install
+
+# the license file isn't found automatically for some reason
+# perhaps it's the .rst extension?
+#
+mkdir ${prefix}/share
+mkdir ${prefix}/share/licenses
+mkdir ${prefix}/share/licenses/Fmt
+cp $WORKSPACE/srcdir/fmt/LICENSE.rst ${prefix}/share/licenses/Fmt/LICENSE
```
giordano, line 33: You might really want to [read the documentation](https://docs.binarybuilder.org/stable/build_tips/#Installing-the-license-file)
```suggestion
install_license ../LICENSE.rst
```

F/Fmt/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libfmt", :libfmt),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
```
giordano, line 53:
```suggestion
dependencies = Dependency[
```

F/Fmt/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
+
```
giordano, line 58:
```suggestion
```

F/Fmt/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+
+cd $WORKSPACE/srcdir/fmt
+mkdir build
+cd build
+
+cmake \
+-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake \
```
giordano, line 21: Why using GCC everywhere?

cmhamel: I run into clang errors on mac without this. Linux and windows build fine either way but Mac is the problem child in this case.

giordano: Can you elaborate on _what_ errors you got?

cmhamel: ld: library not found for -lc++fs
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)

Above is the error encountered on mac (both x86_64 and aarch64).

F/Fmt/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+-DCMAKE_BUILD_TYPE=Release \
+-DBUILD_SHARED_LIBS=True \
+-DCMAKE_INSTALL_PREFIX=${prefix} \
+..
```
giordano, line 25: Don't build the tests
```suggestion
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=True \
    -DFMT_TEST=OFF \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    ..
```

Also, I'm moderately sure I [already suggested](https://github.com/JuliaPackaging/Yggdrasil/pull/5565#discussion_r979091238) using some indentation before.

PR #5567, opened 2022-09-24
[Pango] Upgrade Pango to 1.50.9

aviks:
It apparently has a fix I want, but I'm not sure how it interacts with the rest of the ecosystem. Does Cairo in particular need upgrading?

### DIFF
--- a/P/Pango/build_tarballs.jl
+++ b/P/Pango/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "Pango"
-version = v"1.50.3"
+version = v"1.50.9"
 
 # Collection of sources required to build Pango
 sources = [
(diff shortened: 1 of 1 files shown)


giordano:
As far as I can see the libraries maintained the same sonames (compare [old](https://github.com/JuliaBinaryWrappers/Pango_jll.jl/blob/c862f18d7d9c68fe0a041cfad5f15cc46f948f45/src/wrappers/x86_64-linux-gnu.jl#L11-L13) vs [new](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=22543&view=logs&j=fa2348e3-4ed1-5e7b-66bb-2ec4b077b679&t=602eecc5-1086-5811-004b-b22c5e4590dd&l=4116)), so the new version should be compatible with the old one.

PR #5578, opened 2022-09-26
mlpack: enable experimental platforms for Apple Silicon support

msagarpatel:


msagarpatel:
Hmm, I don't think these changes are actually needed since M1 is no longer an experimental architecture in BinaryBuilderBase.jl. There is still a problem with this package on M1 – I'll gather some more info before opening an issue.

PR #5583, opened 2022-09-27
[Torch] Disable building with MKL

stemann:


stemann:
> What needs to be done here to get this merged?

Fighting a bit with the build process to have it find LAPACK - and BLAS.

... and considering `libblastrampoline` though I guess it would constrain compatibility to Julia v1.7+ - would prefer something compatible with v1.6.

ViralBShah:
I suggest not worrying about 1.6 or even 1.7. These capabilities are fast moving, and we'll have a new LTS next year (I hope - not official). That will make it much easier to work with LBT.

stemann:
I have confirmed that the generated Torch_jll can be imported on `x86_64-linux-gnu-cxx11`.

ViralBShah:
This may need a @giordano review.

giordano:
Thanks a lot @stemann!

T/Torch/build_tarballs.jl
```diff
@@ -232,37 +226,39 @@ for p in cuda_platforms
...
+    # BuildDependency("MKL_Headers_jll"; platforms = mkl_platforms), # MKL is avoided for all platforms
+    Dependency("OpenBLAS32_jll"; platforms = openblas_platforms),
     Dependency("PThreadPool_jll"; compat = "0.0.20210414"),
     Dependency("SLEEF_jll", v"3.5.2"; compat = "3"),
-#    Dependency("TensorRT_jll"; platforms = cuda_platforms), # Building with TensorRT is not supported: https://github.com/pytorch/pytorch/issues/60228
+    # Dependency("TensorRT_jll"; platforms = cuda_platforms), # Building with TensorRT is not supported: https://github.com/pytorch/pytorch/issues/60228
     Dependency("XNNPACK_jll"; compat = "0.0.20210622"),
     Dependency(PackageSpec("protoc_jll", Base.UUID("c7845625-083e-5bbe-8504-b32d602b7110")); compat="~3.13.0"),
     HostBuildDependency(PackageSpec("protoc_jll", Base.UUID("c7845625-083e-5bbe-8504-b32d602b7110"), v"3.13.0")),
+    RuntimeDependency("CUDA_Runtime_jll"),
```
giordano, line 257: Was this supposed to be a dependency for all platforms?  Or only the CUDA ones?

stemann: Its required, I believe, by the CUDA augment block that I added - I guess that is necessary in order to be able to get hold of the CUDA artifacts.

PR #5585, opened 2022-09-27
[CJieba] Add new recipe 

zxygentoo:
Hello,

This is a recipe for [cjieba](https://github.com/yanyiwu/cjieba/).

Some background:
- [jieba](https://github.com/fxsjy/jieba) is a popular Chinese word tokenizor/POS tagger etc.
- Originally implemented in Python.
- The C library this recipe wrapped is the low-level C api for the C++ implementation [cppjieba](https://github.com/yanyiwu/cppjieba), which I found quite adequate for simple tasks.

_I couldn't get all the way through in wizard, keep failing at download/install amd64 apple darwin20 toolchain. But it looks pretty OK._

Cheers,
Alex

### DIFF
--- a/C/CJieba/build_tarballs.jl
+++ b/C/CJieba/build_tarballs.jl
@@ -1,0 +1,41 @@
+using BinaryBuilder, Pkg
+
+name = "CJieba"
+version = v"0.3.0"
+
+sources = [
+    ArchiveSource(
+        "https://github.com/yanyiwu/cjieba/archive/refs/tags/v$(version).tar.gz",
+        "1f5ca82c5d19c38485c8fd50db8711602d339b1e85f15fb864fc116c7f770f12"
+    )
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir
+mkdir $libdir $prefix/share
+curl https://raw.githubusercontent.com/yanyiwu/cppjieba/master/LICENSE > LICENSE
+install_license LICENSE 
+cd cjieba-0.3.0/
+g++ lib/jieba.cpp -shared -fPIC -I./deps/ -std=c++11 -o libcjieba.so
+cp libcjieba.so $libdir
+cp -a dict $prefix/share/
+"""
+
+platforms = supported_platforms()
+
+products = [
+    LibraryProduct("libcjieba", :libcjieba),
+    FileProduct("share/dict/hmm_model.utf8", :hmm_model),
+    FileProduct("share/dict/idf.utf8", :idf),
+    FileProduct("share/dict/jieba.dict.utf8", :jiebadict),
+    FileProduct("share/dict/stop_words.utf8", :stopwords),
+    FileProduct("share/dict/user.dict.utf8", :userdict)
+]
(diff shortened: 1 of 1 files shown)


C/CJieba/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+    ArchiveSource(
+        "https://github.com/yanyiwu/cjieba/archive/refs/tags/v$(version).tar.gz",
+        "1f5ca82c5d19c38485c8fd50db8711602d339b1e85f15fb864fc116c7f770f12"
+    )
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir
+mkdir $libdir $prefix/share
+curl https://raw.githubusercontent.com/yanyiwu/cppjieba/master/LICENSE > LICENSE
```
giordano, line 16: I'm confused: this is a license for a different project?

C/CJieba/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+install_license LICENSE 
+cd cjieba-0.3.0/
```
giordano, line 18:
```suggestion
cd cjieba-*
```

C/CJieba/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+g++ -o "$libdir/libcjieba.$dlext" lib/jieba.cpp -shared -fPIC -I./deps/ -std=c++11
```
giordano, line 19:
```suggestion
c++ -o "$libdir/libcjieba.$dlext" lib/jieba.cpp -shared -fPIC -I./deps/ -std=c++11
```

C/CJieba/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+cp -a dict $prefix/share/
```
giordano, line 20:
```suggestion
cp -va dict $prefix/share/
```

giordano (commented):
Thanks for your contribution!  I have some questions though

C/CJieba/build_tarballs.jl
zxygentoo: This repo is the low level C api for cppjieba, the author didn't add a license for this one, but cppjieba use MIT, so it seems appropriate here.

giordano: I don't want to pesky here, but we shouldn't make assumption about what developers think: they should explicitly state the license they want to use, otherwise we don't have the rights to distribute their software.

zxygentoo: Good point. I could try to contract the author see if he's will to add a explicit license for cjieba.

giordano: That'd be perfect, thanks a lot!

zxygentoo:
@giordano Updated using author's new commit with license along with mentioned changes.

giordano:
Nice, thanks for your contribution!

PR #3662, opened 2021-09-30
qhull semver for v2020.2 is actually v8.0.2

stevengj:
As [explained in #3657](https://github.com/JuliaPackaging/Yggdrasil/pull/3657#issuecomment-930666102), we should really rename this to the [semver version](http://www.qhull.org/news/qhull-news.html).  (Or semver-like, at least; I don't know if they are strictly following semantic versioning.)

Then [dependents of `qhull_jll`](https://juliahub.com/ui/Packages/Qhull_jll/nBD85/2020.2.0+0?t=2) can update their compatibility section to list `qhull_jll = "8"` and it should do the right thing, since Pkg assumes that differing major versions are incompatible.

Once 8.0.2 is released, we can then merge #3657 to release 8.1-alpha1.

PR #5588, opened 2022-09-28
add qhull wrapper jll

SimonDanisch:
Tested it locally and seems to work well :) 
I have a question though, due to https://github.com/JuliaPackaging/Yggdrasil/pull/3662, it seems like I need to specify a compat for the `Dependency`, since it considers the old version to be newer than 8?
CC: @stevengj


giordano:
There is no license

SimonDanisch:
CC: @santiagobadia @victorsndvg @fverdugo, added this as a license https://github.com/JuliaPackaging/Yggdrasil/pull/5588/files#diff-ac28c35c7c79ee8d0984f37972032915c680af76d62ecb3a50bbe270a66bb566 taken from MiniQhull.jl

SimonDanisch:
> Dependency Qhull_jll does not have a mapping for artifact Qhull for platform armv6l-linux-musleabihf-libgfortran3-cxx03

Hm, I guess new platforms were added since Qhull_jll has been build the last time?

fverdugo:
@SimonDanisch Thanks for packaging this!

Q/QhullMiniWrapper/build_tarballs.jl
```diff
@@ -0,0 +1,32 @@
...
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("qhullwrapper", :qhullwrapper)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+    Dependency(PackageSpec(name="Qhull_jll"), compat="8")
```
giordano, line 28: You have to use a version which actually exists: https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.Dependency

> The optional keyword argument `compat` can be used to specify a string for use in the `Project.toml `of the generated Julia package. If `compat` is non-empty and `build_version` is not passed, the latter defaults to the minimum version compatible with the `compat` specifier.

```suggestion
dependencies = [
    Dependency(PackageSpec(name="Qhull_jll"), compat="8.0.1000")
```

SimonDanisch:
Cool cool cool :) Seems like that fixed all problems ;)

PR #5589, opened 2022-09-28
Update Qhull

SimonDanisch:
Not sure what the preferred way is to rebuild a package with the new platforms, but I figured the `experimental=true` needs to go anyways ;)


giordano:
https://github.com/JuliaPackaging/Yggdrasil/pull/5588/files#r982230229

PR #5598, opened 2022-09-28
[Qt5Base] Build x86_64-apple-darwin library with 10.13 SDK

giordano:
According to https://doc.qt.io/qt-5/macos.html Qt5 should require 10.13, not 10.14, I don't know why this was built with 10.14, let's give it a try?  CC @barche


barche:
It fails with

```
Undefined symbols for architecture x86_64:
[23:12:18]   "___isPlatformVersionAtLeast", referenced from:
[23:12:18]       __ZNK16QMacStylePrivate12cocoaControlENS_12CocoaControlE in qmacstyle_mac.o
```

This is because there is a dynamic SDK check in the cocoa code, which doesn't work when cross-compiling unless you compile with the newest SDK it wants. So Qt compiled on Mac works with 10.13, cross compiled it needs 10.14 so the SDK check passes statically (as far as I understand).

giordano:
Uhm,if ` ___isPlatformVersionAtLeast` is needed only for compilation, I wonder if we can build against the 10.14 SDK, but set `QMAKE_MACOSX_DEPLOYMENT_TARGET=10.13`

giordano:
That doesn't work either, still choking on `___isPlatformVersionAtLeast` undefined symbol.  Sigh

Q/Qt5Base/build_tarballs.jl
```diff
@@ -170,6 +174,9 @@ products_macos = [
     FrameworkProduct("QtXml", :libqt5xml),
 ]
 
+# We must use the same version of LLVM for the build toolchain and LLVMCompilerRT_jll
+llvm_version = v"13.0.1"
+
 # Dependencies that must be installed before this package can be built
 dependencies = [
     BuildDependency("Xorg_libX11_jll"),
```
giordano, line 182: Reminder to self to filter the dependencies by platform

giordano:
@barche do you happen to know what's the spell to pass linker flags?  `LDFLAGS` and `LIBS` environment variables are clearly not working, they aren't used at all.

barche:
The current iteration adds

```
-L${libdir}/darwin \
QMAKE_LFLAGS+=-lclang_rt.osx \
```

to the configure arguments. I think this goes through, but now the M1 build fails with:

```
ERROR: Cannot compile a minimal program. The toolchain or QMakeSpec is broken.
```

We'd have to check the config.log to find out why, I'll need to do a local build for that I think.

giordano:
> but now the M1 build fails with:

Doesn't surprise me since libclang_rt.osx isn't available there.  I tried to add those flags only for x86_64

barche (approved):
Looks OK!

giordano:
Awesome, thank you! :rocket:

PR #5607, opened 2022-09-30
Create recipe for OpenCL library

juliohm:
This recipe supersedes the current OpenCL_Headers recipe by adding the header and the library products into a single JLL.

It would be awesome if someone with more experience on Windows could fix the build errors over there 🙏🏽

### DIFF
--- a/O/OpenCL/build_tarballs.jl
+++ b/O/OpenCL/build_tarballs.jl
@@ -55,4 +55,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"9.1.0")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"6.1.0")
--- a/O/OpenCL/build_tarballs.jl
+++ b/O/OpenCL/build_tarballs.jl
@@ -1,0 +1,58 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "OpenCL"
+version = v"2022.9.23"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/KhronosGroup/OpenCL-Headers.git", "4c50fabe3774bad4bdda9c1ca92c82574109a74a"),
+    GitSource("https://github.com/KhronosGroup/OpenCL-ICD-Loader.git", "3dae4803532b11d74e4dc216ee72570c1a4bff24")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+
+install_license ./OpenCL-Headers/LICENSE
+
+wget https://patch-diff.githubusercontent.com/raw/KhronosGroup/OpenCL-Headers/pull/209.patch
+patch ./OpenCL-Headers/tests/test_headers.c 209.patch
+
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -S ./OpenCL-Headers -B ./OpenCL-Headers/build
+cmake --build ./OpenCL-Headers/build --target install -j${nproc}
+
+cmake -DCMAKE_PREFIX_PATH=${prefix} -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -S ./OpenCL-ICD-Loader -B ./OpenCL-ICD-Loader/build
+cmake --build ./OpenCL-ICD-Loader/build --target install -j${nproc}
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
+
+
+# The products that we will ensure are always built
+products = [
+    FileProduct("include/CL/cl.h", :cl_h),
+    LibraryProduct("libOpenCL", :libcl)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"9.1.0")
--- a/O/OpenCL/build_tarballs.jl
+++ b/O/OpenCL/build_tarballs.jl
@@ -22,7 +22,7 @@ cmake --build ./OpenCL-ICD-Loader/build --target install -j${nproc}
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = supported_platforms(; exclude=os -> Sys.iswindows(os) || Sys.isfreebsd(os))
 
 # The products that we will ensure are always built
 products = [
--- a/O/OpenCL/build_tarballs.jl
+++ b/O/OpenCL/build_tarballs.jl
@@ -44,7 +44,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("OpenCL_Headers_jll", v"2022.09.23")
+    BuildDependency(PackageSpec(; name="OpenCL_Headers_jll", version=v"2022.09.23")),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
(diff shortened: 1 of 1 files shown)


O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "OpenCL"
+version = v"2022.9.23"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/KhronosGroup/OpenCL-Headers.git", "4c50fabe3774bad4bdda9c1ca92c82574109a74a"),
```
giordano, line 10: This should depend on `OpenCL_Headers` as a [`BuildDependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.BuildDependency) if they're needed for building this package.  Headers are useless at runtime and this would create a useless clash with the other package.
```suggestion
```

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+
+install_license ./OpenCL-Headers/LICENSE
+
+wget https://patch-diff.githubusercontent.com/raw/KhronosGroup/OpenCL-Headers/pull/209.patch
+patch ./OpenCL-Headers/tests/test_headers.c 209.patch
```
giordano, line 21: Either add this URL as a new [`FileSource`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.FileSource) or copy the patch in a `bundled/patches` subdirectory and add `./bundled` as [`DirectorySource`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.DirectorySource).  There are hundreds of similar examples in Yggdrasil.

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -S ./OpenCL-Headers -B ./OpenCL-Headers/build
+cmake --build ./OpenCL-Headers/build --target install -j${nproc}
+
```
giordano, line 25:
```suggestion
```

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
+
```
giordano, line 45: At the moment this is equivalent to
```suggestion
platforms = supported_platforms(; exclude=!Sys.islinux)
```
If you want to look at the other platforms you have to do
```suggestion
platforms = supported_platforms()
```
so that we can see what's the problem in the other platforms.

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+
+# The products that we will ensure are always built
+products = [
+    FileProduct("include/CL/cl.h", :cl_h),
+    LibraryProduct("libOpenCL", :libcl)
```
giordano, line 50:
```suggestion
    LibraryProduct("libOpenCL", :libcl),
```

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"9.1.0")
```
giordano, line 58: Any specific reasons why this using GCC 9?  For maximum compatibility we want to use _oldest_ version of GCC which can be used (rule which applies only to GCC, not LLVM).

O/OpenCL/build_tarballs.jl
juliohm: But OpenCL_Headers was built with a different version. Wouldn't it be a good idea to bundle headers + library into a single place to avoid API issues in the future?

giordano: You can update `OpenCL_Headers_jll` then.  Headers aren't needed at runtime, not sure what API issues you're referring to.  If you need a particular version for building this package, you can specify the compat with `OpenCL_Headers_jll`.

O/OpenCL/build_tarballs.jl
juliohm: I will retry with older versions of GCC. The default version which is 4.x was causing build errors.

juliohm: The oldest version for which the build is successful is GCC 6.1.0. I will update the PR accordingly.

juliohm, line 58:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"6.1.0")
```

O/OpenCL/build_tarballs.jl
giordano: Why was this marked as resolved?  The comment above still stand and hasn't been addressed

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,51 @@
...
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libOpenCL", :libcl)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("OpenCL_Headers_jll", v"2022.09.23")
```
giordano, line 47: I [asked](https://github.com/JuliaPackaging/Yggdrasil/pull/5607#discussion_r984803681) for this to be a [`BuildDependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.BuildDependency)
```suggestion
    BuildDependency(PackageSpec(; name="OpenCL_Headers_jll", version=v"2022.09.23")),
```

O/OpenCL/build_tarballs.jl
juliohm: I will change it to supported_platforms() to see if the errors are reproduced on CI

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
```
juliohm, line 3:
```suggestion
using BinaryBuilder, Pkg
```

juliohm:
As you can see, the build fails on non-Linux platforms. Any easy fix?

giordano:
No, no easy fix, but it's working on macOS too

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libOpenCL", :libcl)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency(PackageSpec(; name="OpenCL_Headers_jll", version=v"2022.09.23")),
```
juliohm, line 34:
```suggestion
    BuildDependency(PackageSpec(; name="OpenCL_Headers_jll", version=v"2022.09.23"))
```

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+
+install_license ./OpenCL-ICD-Loader/LICENSE
+
+cmake -DCMAKE_PREFIX_PATH=${prefix} -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -S ./OpenCL-ICD-Loader -B ./OpenCL-ICD-Loader/build
+cmake --build ./OpenCL-ICD-Loader/build --target install -j${nproc}
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
```
juliohm, line 25:
```suggestion
platforms = supported_platforms(; exclude=os -> Sys.iswindows(os) || Sys.isfreebsd(os))
```

juliohm:
@giordano I've updated the PR to exclude the platforms that failed. Appreciate if you can review and merge. I will continue the conversation with the OpenCL maintainers to try to add support for Windows and FreeBSD as well.

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "OpenCL"
```
giordano, line 5: Shouldn't this be called 
```suggestion
name = "OpenCL_ICD_Loader"
```
to match the project name?

O/OpenCL/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+cmake --build ./OpenCL-ICD-Loader/build --target install -j${nproc}
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; exclude=os -> Sys.iswindows(os) || Sys.isfreebsd(os))
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libOpenCL", :libcl)
```
giordano, line 29: Product name has typically the same name as the library (modulo case)
```suggestion
    LibraryProduct("libOpenCL", :libopencl)
```

O/OpenCL/build_tarballs.jl
juliohm: I thought about it. Do you think end-users will get lost with this weird name they chosen for the main project? For example, it took a while for me to figure out that this is the actual core OpenCL library. If users find OpenCL_ICD_Loader_jll in the registry they won't be sure if this is what they need. That is why I thought the simpler OpenCL_jll would be best for users.

juliohm: Let me know if it is ok to go ahead with OpenCL_jll only. I think the combo OpenCL_Headers_jll and OpenCL_jll reads better, and will be simpler to understand for end users.

giordano:
Ok, let's go.  Thanks for your contribution!

PR #5609, opened 2022-09-30
Update OpenCL_Headers recipe

juliohm:
Appreciate if you can review and merge. After this is merged, I will come back to #5607

### DIFF
--- a/O/OpenCL_Headers/build_tarballs.jl
+++ b/O/OpenCL_Headers/build_tarballs.jl
@@ -1,6 +1,6 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
-using BinaryBuilder, Pkg
+using BinaryBuilder
 
 name = "OpenCL_Headers"
 version = v"2022.9.23"
--- a/O/OpenCL_Headers/build_tarballs.jl
+++ b/O/OpenCL_Headers/build_tarballs.jl
@@ -35,8 +35,7 @@ products = [
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[
-]
+dependencies = Dependency[]
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"6.1.0")
--- a/O/OpenCL_Headers/build_tarballs.jl
+++ b/O/OpenCL_Headers/build_tarballs.jl
@@ -51,4 +51,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"6.1.0")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"6.1.0")
--- a/O/OpenCL_Headers/build_tarballs.jl
+++ b/O/OpenCL_Headers/build_tarballs.jl
@@ -26,19 +26,7 @@ cmake --build ./OpenCL-Headers/build --target install -j${nproc}
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = [
-    Platform("i686", "linux"; libc = "glibc"),
-    Platform("x86_64", "linux"; libc = "glibc"),
-    Platform("aarch64", "linux"; libc = "glibc"),
-    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
-    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
-    Platform("powerpc64le", "linux"; libc = "glibc"),
-    Platform("i686", "linux"; libc = "musl"),
-    Platform("x86_64", "linux"; libc = "musl"),
-    Platform("aarch64", "linux"; libc = "musl"),
-    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
-    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
-]
+platforms = [AnyPlatform()]
 
 
 # The products that we will ensure are always built
--- a/O/OpenCL_Headers/build_tarballs.jl
+++ b/O/OpenCL_Headers/build_tarballs.jl
@@ -18,8 +18,6 @@ cd $WORKSPACE/srcdir
 
 install_license ./OpenCL-Headers/LICENSE
 
-patch ./OpenCL-Headers/tests/test_headers.c 209.patch
-
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -S ./OpenCL-Headers -B ./OpenCL-Headers/build
 cmake --build ./OpenCL-Headers/build --target install -j${nproc}
 """
--- a/O/OpenCL_Headers/build_tarballs.jl
+++ b/O/OpenCL_Headers/build_tarballs.jl
@@ -8,8 +8,6 @@ version = v"2022.09.23"
 # Collection of sources required to complete build
 sources = [
     GitSource("https://github.com/KhronosGroup/OpenCL-Headers.git", "4c50fabe3774bad4bdda9c1ca92c82574109a74a"),
-    FileSource("https://patch-diff.githubusercontent.com/raw/KhronosGroup/OpenCL-Headers/pull/209.patch",
-               "c3afd4ad0a37f0b61c0b8656ca4914002ba7994bba05aa2c47fde59b652289c9")
 ]
 
 # Bash recipe for building across all platforms
--- a/O/OpenCL_Headers/build_tarballs.jl
+++ b/O/OpenCL_Headers/build_tarballs.jl
@@ -28,7 +28,6 @@ cmake --build ./OpenCL-Headers/build --target install -j${nproc}
 # platforms are passed in on the command line
 platforms = [AnyPlatform()]
 
-
 # The products that we will ensure are always built
 products = [
     FileProduct("include/CL/cl.h", :cl_h)
--- a/O/OpenCL_Headers/build_tarballs.jl
+++ b/O/OpenCL_Headers/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder
 
 name = "OpenCL_Headers"
-version = v"2022.9.23"
+version = v"2022.09.23"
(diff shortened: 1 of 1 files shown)


O/OpenCL_Headers/build_tarballs.jl
```diff
@@ -1,35 +1,54 @@
...
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
```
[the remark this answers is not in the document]
juliohm: The previous recipe was just downloading a generic tar ball, the new one actually builds the source for the limited number of platforms. The build fails in other platforms. Should I list all of them anyways?

juliohm:
The removal of the julia v1.6 compat entry has affected the build somehow. It is now failing, but I don't know why.

O/OpenCL_Headers/build_tarballs.jl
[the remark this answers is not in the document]
giordano: This is not building anything, this is just moving header files to the prefix.  All tarballs you built in https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=22722&view=logs&j=983b6c68-8f90-5831-4b74-0e190ce0b0f9 for 11 different platforms are bit-by-bit identical (you can see that by looking at the git tree hash and the sha256sum of the main tarball in the logs), because that's what it's supposed to be, and that's the entire reason why [`AnyPlatform`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.AnyPlatform) exists and it was used here in the first place.

O/OpenCL_Headers/build_tarballs.jl
```diff
@@ -3,33 +3,52 @@
...
+version = v"2022.9.23"
 
-# Collection of sources required to build this package
+# Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://github.com/KhronosGroup/OpenCL-Headers/archive/v2020.03.13.tar.gz",
-                  "664bbe587e5a0a00aac267f645b7c413586e7bc56dca9ff3b00037050d06f476"),
+    GitSource("https://github.com/KhronosGroup/OpenCL-Headers.git", "4c50fabe3774bad4bdda9c1ca92c82574109a74a"),
+    FileSource("https://patch-diff.githubusercontent.com/raw/KhronosGroup/OpenCL-Headers/pull/209.patch",
+               "c3afd4ad0a37f0b61c0b8656ca4914002ba7994bba05aa2c47fde59b652289c9")
```
giordano, line 12: This patch is only for tests, why do we need that?

giordano:
> The removal of the julia v1.6 compat entry has affected the build somehow. It is now failing, but I don't know why.

Because you aren't doing what I'm telling you.

O/OpenCL_Headers/build_tarballs.jl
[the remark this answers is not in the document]
juliohm: I am following the instructions by the OpenCL maintainers assuming that this is the best set of instructions one could possibly use to build the OpenCL-Headers project: https://github.com/KhronosGroup/OpenCL-Headers#example-build

If you prefer to just copy the raw files from a given commit manually, it may work this time, but not necessarily in the future if they start modifying the source on a per-platform basis. Should I copy the files or preserve their instructions?

O/OpenCL_Headers/build_tarballs.jl
juliohm: If you try to build the headers following their instructions you will get an error without the patch. I know that the patch is changing a test file, but for some reason they included that in the install target. Should I remove the patch? I am trying to be the most conservative here, but can change it to something else if you prefer.

juliohm:
> Because you aren't doing what I'm telling you.

Well, I am trying 😮‍💨 It is my first time using BinaryBuilder.jl and many changes you requested are basically the result of the wizard. I just copied/pasted the result of the wizard here, and it added Pkg and compat=juliav1.6.

O/OpenCL_Headers/build_tarballs.jl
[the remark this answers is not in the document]
giordano: I'm only telling you that you have to restore `[AnyPlatform()]`, what you're saying here looks orthogonal to that.

giordano:
Don't use the wizard: https://docs.binarybuilder.org/stable/#Manually-create-or-edit-build_tarballs.jl

> However, it lacks the full control of all options you can use in a build_tarballs.jl script. To generate this file (explained in greater detail in [Building Packages](https://docs.binarybuilder.org/stable/building/)), one can clone Yggdrasil, copy an existing build recipe, modify it, and submit a new pull request. Manually editing the build_tarballs.jl script is also the recommended way when you want to update an existing recipe, rather then starting from scratch with the wizard.

O/OpenCL_Headers/build_tarballs.jl
```diff
@@ -3,33 +3,40 @@
 using BinaryBuilder
 
 name = "OpenCL_Headers"
-version = v"2020.03.13"
+version = v"2022.9.23"
```
juliohm, line 6:
```suggestion
version = v"2022.09.23"
```

O/OpenCL_Headers/build_tarballs.jl
```diff
@@ -3,33 +3,40 @@
...
+
 # The products that we will ensure are always built
 products = [
     FileProduct("include/CL/cl.h", :cl_h)
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[]
+dependencies = Dependency[
+]
```
juliohm, line 39:
```suggestion
dependencies = Dependency[]
```

juliohm:
I don't understand why the build is working in any platform now. If I run the same script locally I get build errors. Can you clarify the difference?

giordano:
I can't comment if I don't know what you're running exactly.  The latest version has no reason for not working, and it works locally for me.

O/OpenCL_Headers/build_tarballs.jl
giordano:
```suggestion
```

O/OpenCL_Headers/build_tarballs.jl
```diff
@@ -3,27 +3,32 @@
...
-install_license LICENSE
+cd $WORKSPACE/srcdir
 
-mkdir ${prefix}/include
-mv CL ${prefix}/include/
+install_license ./OpenCL-Headers/LICENSE
+
+patch ./OpenCL-Headers/tests/test_headers.c 209.patch
+
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -S ./OpenCL-Headers -B ./OpenCL-Headers/build
```
giordano, line 23:
```suggestion
cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DOPENCL_HEADERS_BUILD_CXX_TESTS=OFF -S ./OpenCL-Headers -B ./OpenCL-Headers/build
```

O/OpenCL_Headers/build_tarballs.jl
```diff
@@ -3,27 +3,32 @@
...
+cmake --build ./OpenCL-Headers/build --target install -j${nproc}
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = [AnyPlatform()]
 
+
```
giordano, line 31:
```suggestion
```

juliohm:
> I can't comment if I don't know what you're running exactly. The latest version has no reason for not working, and it works locally for me.

Is [AnyPlatform()] equivalent to supported_platforms()?

giordano:
> Is [AnyPlatform()] equivalent to supported_platforms()?

No: https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.AnyPlatform (I think I [linked it already before](https://github.com/JuliaPackaging/Yggdrasil/pull/5609#discussion_r984951066))

juliohm:
So what is the effect of choosing it? When I tried with supported_platforms() the build failed locally. I am wondering what magic is performed in the AnyPlatform() scenario to make it work?

giordano:
For the record, you can verify yourself that the [latest build](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=22729&view=logs&j=ea662ff7-bbf1-57a7-cfb8-437f162d4e92&t=7e18c8a8-e55a-5e35-eae5-d9e51bef8844&l=3379) produced a tarball
```
[ Info: Tree hash of contents of OpenCL_Headers.v2022.9.23.any.tar.gz: 7e498283fb547517c42d347fa7a42aa725b7058b
[ Info: SHA256 of OpenCL_Headers.v2022.9.23.any.tar.gz: 85e659d76a09dd2799212c9cf088099cd37a8971674f176ac10968bf0221c200
```
which is ***bit-by-bit identical*** to ***all*** the 11 tarballs you built in https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=22722&view=logs&j=983b6c68-8f90-5831-4b74-0e190ce0b0f9, as I said in https://github.com/JuliaPackaging/Yggdrasil/pull/5609#discussion_r984951066

giordano:
> So what is the effect of choosing it? When I tried with supported_platforms() the build failed locally. I am wondering what magic is performed in the AnyPlatform() scenario to make it work?

`supported_platform()` gives you a vector of ***different*** platforms, `AnyPlatform()` is a single special platform which is compatible with ***any*** platform: it's for platform-***independent*** products, as explained in the [docstring](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.AnyPlatform) (which I believe I linked multiple times at this point).  Header files are platform-independent.

juliohm:
So this instructions in CMakeLists.jl which are platform-specific don't make difference? https://github.com/KhronosGroup/OpenCL-Headers/blob/5e2f8c66476bcf60eb8c2921807e34efb30294ee/CMakeLists.txt#L23

giordano:
People like to complicate their lives :shrug:

juliohm:
Or they do but they are not "build instructions" and that is why it is fine to use a generic AnyPlatform?

juliohm:
Ok, you know better. I will proceed and update the other PR in #5607 tomorrow morning.

PR #5623, opened 2022-10-03
Do not build tests in OpenCL_Headers recipe

juliohm:
Removes unnecessary build instructions.

### DIFF
--- a/O/OpenCL_Headers/build_tarballs.jl
+++ b/O/OpenCL_Headers/build_tarballs.jl
@@ -16,7 +16,7 @@ cd $WORKSPACE/srcdir
 
 install_license ./OpenCL-Headers/LICENSE
 
-cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -S ./OpenCL-Headers -B ./OpenCL-Headers/build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=OFF -S ./OpenCL-Headers -B ./OpenCL-Headers/build
 cmake --build ./OpenCL-Headers/build --target install -j${nproc}
 """
 


giordano:
https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=22787&view=logs&j=ea662ff7-bbf1-57a7-cfb8-437f162d4e92&t=7e18c8a8-e55a-5e35-eae5-d9e51bef8844&l=154
```
[ Info: Tree hash of contents of OpenCL_Headers.v2022.9.23.any.tar.gz: 7e498283fb547517c42d347fa7a42aa725b7058b
[ Info: SHA256 of OpenCL_Headers.v2022.9.23.any.tar.gz: 85e659d76a09dd2799212c9cf088099cd37a8971674f176ac10968bf0221c200
```
This is bit-by-bit identical to the current build: https://github.com/JuliaBinaryWrappers/OpenCL_Headers_jll.jl/blob/ccec27e1dad64bce2b8b571e4594902b6be9d537/Artifacts.toml.  In https://github.com/JuliaPackaging/Yggdrasil/pull/5609#discussion_r984967069 I originally suggested not to build tests (you can look at the history of the my message), before realising that wouldn't make any difference at all.

juliohm:
Thank you @giordano , so it is not worth it to add this option thinking about future updates to this recipe? I am trying to write a recipe that others can update in the future. If the OpenCL maintainers decide to do something different with the tests the option would inhibit surprises on our side? Let me know if you prefer to simply close the PR, it is fine.

giordano:
Header files are unlikely to be affected by this, since we don't really build anything that goes into the final tarball containing only the header files (which is a message I've been trying to convey for some time)

juliohm:
I understand the theory that header files are not compiled, I think you are not getting my point though, or maybe you are? I am saying that OpenCL maintainers can do things in practice that don't match the theory so there is nothing that inhibits them from building things in their CMakeLists.txt? Does it make sense what I am trying to say? It seems that you are assuming a perfect world where the CMakeLists.txt will never build things? Or is it a feature of the AnyPlatform() magic that inhibits the natural execution of CMakeLists.txt that I am missing?

juliohm:
In any case it seems that the policy is to not merge changes unless they make a difference **today**. Closing the PR.

giordano:
I mean, you modified this in #5609 to go from doing a simple `cp` to using CMake with all the added complexity that you're now complaining abou.  If we had to do anything, I'd simply go back to what we were doing before and that's it.

juliohm:
As I explained, I tend to prefer the option that is more general and recommended by the maintainers of the project I am trying to build. I am currently trying to fix the build on FreeBSD and on Windows together with them, and all the proposed fixes will naturally live in CMakeLists.jl. Any approach that assumes a simple `cp foo bar` is less general and will fail with more probability in future releases.

PR #5624, opened 2022-10-03
Do not build tests in OpenCL recipes

juliohm:
Removes unnecessary build instructions.


juliohm:
Is this PR similar to #5623 ? Feel free to close it if it doesn't change the binaries on a bit level.

PR #5630, opened 2022-10-04
Add MIOpen 5.2.3

pxl-th:


M/MIOpen/common.jl
```diff
@@ -0,0 +1,75 @@
...
+        DirectorySource("../scripts"),
+    ]
+    dependencies = [
+        BuildDependency(PackageSpec(; name="ROCmLLVM_jll", version)),
+        BuildDependency(PackageSpec(; name="rocm_cmake_jll", version)),
+        Dependency("HIP_jll", version),
+        Dependency("rocBLAS_jll", version),
+        Dependency("Zlib_jll"),
+        Dependency("SQLite_jll"),
+        Dependency("boost_jll", v"1.79.0"),
```
giordano, line 72: #3024.

Besides the specific case of boost, as a general remark, if you need a specific version of a dependency for building a package, that's a good indication that you should reflect that in the compat bound. I see very little need to specify the build version without the compat

giordano (changes requested):

giordano (approved):

PR #5634, opened 2022-10-05
Update tectonic to 0.11.0

MichaelHatherly:
There's also a 0.11.0 that was released shortly after the 0.10.0. Do we care about having all the intermediate versions available, or should I just switch this PR to updating directly to 0.11.0?


giordano:
I'd say build whatever you actually need, we don't care about covering versions no one would use in practice

MichaelHatherly:
Ok, updating directly to 0.11.0 then.

PR #5637, opened 2022-10-06
[SCIP_PaPILO] up bugfix release

matbesancon:


S/SCIP_PaPILO/build_tarballs.jl
```diff
@@ -23,6 +23,7 @@ cmake -DCMAKE_INSTALL_PREFIX=$prefix\
   -DAMPL=0\
   -DGCG=0\
   -DSYM=bliss\
+  -TPI=tny\
```
matbesancon, line 26:
```suggestion
  -DTPI=tny\
```

S/SCIP_PaPILO/build_tarballs.jl
```diff
@@ -58,7 +59,7 @@ dependencies = [
     Dependency(PackageSpec(name="Bzip2_jll", uuid="6e34b625-4abd-537c-b88f-471c36dfa7a0"); compat="1.0.8"),
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
     Dependency(PackageSpec(name="GMP_jll", uuid="781609d7-10c4-51f6-84f2-b8444358ff6d"), v"6.2.0"),
-    Dependency(PackageSpec(name="Ipopt_jll", uuid="9cc047cb-c261-5740-88fc-0cf96f7bdcc7"), v"300.1400.400"),
+    Dependency(PackageSpec(name="Ipopt_jll", uuid="9cc047cb-c261-5740-88fc-0cf96f7bdcc7"), "300.1400.403"),
```
giordano, line 62: The `v` should have stayed there, but I think these should all use appropriate compat instead: https://github.com/JuliaPackaging/Yggdrasil/pull/5630#discussion_r987419600

S/SCIP_PaPILO/build_tarballs.jl
```diff
@@ -53,12 +54,12 @@ products = [
 ]
 
 dependencies = [
-    Dependency(PackageSpec(name="bliss_jll", uuid="508c9074-7a14-5c94-9582-3d4bc1871065"), v"0.77.0"),
-    Dependency(PackageSpec(name="boost_jll", uuid="28df3c45-c428-5900-9ff8-a3135698ca75"); compat="=1.76.0"),
+    Dependency(PackageSpec(name="bliss_jll", uuid="508c9074-7a14-5c94-9582-3d4bc1871065"), compat="=0.77.0"),
```
giordano, line 57: To be clear, the message I linked suggested to use `=` compat _only_ for boost, because they break ABI in every single version, but that's not the case for BLIS and most of the other packages below, as far as I know. But below you do have boost, for that one you do need `=` 🙂

matbesancon: bliss has had breaking changes all over the place and we don't need the bleeding edge in there so 077 forced seems wise

S/SCIP_PaPILO/build_tarballs.jl
```diff
@@ -53,12 +54,12 @@ products = [
...
+    Dependency(PackageSpec(name="boost_jll", uuid="28df3c45-c428-5900-9ff8-a3135698ca75"); compat="1.76.0"),
```
giordano, line 58: but still need to fix compat for boost
```suggestion
    Dependency(PackageSpec(name="boost_jll", uuid="28df3c45-c428-5900-9ff8-a3135698ca75"); compat="=1.76.0"),
```

matbesancon:
thanks for the help!

PR #5639, opened 2022-10-08
[SCIP] fix Ipopt version to last 1.6 compatible

matbesancon:
This is a bit too constraining to have Ipopt requiring julia 1.8

PR #5640, opened 2022-10-08
[SCIP_PaPILO] fix ipopt version

matbesancon:


matbesancon:
similar to https://github.com/JuliaPackaging/Yggdrasil/pull/5639

matbesancon:
I tried avoiding the ipopt issue which still remained on Julia 1.6

Ipopt 400 is supposed to support Julia 1.6
https://github.com/JuliaBinaryWrappers/Ipopt_jll.jl/blob/Ipopt-v300.1400.400%2B0/Project.toml

but there is still an incompatibility somewhre
```
(jl_0j9wM3) pkg> add SCIP_PaPILO_jll@v800
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package Ipopt_jll [9cc047cb]:
 Ipopt_jll [9cc047cb] log:
 ├─possible versions are: 3.13.1-300.1400.403 or uninstalled
 ├─restricted by compatibility requirements with SCIP_PaPILO_jll [fc9abe76] to versions: 300.1400.403
 │ └─SCIP_PaPILO_jll [fc9abe76] log:
 │   ├─possible versions are: 0.1.0-800.0.200 or uninstalled
 │   └─restricted to versions 800 by an explicit requirement, leaving only versions 800.0.200
 └─restricted by julia compatibility requirements to versions: 3.13.1-300.1400.400 or uninstalled — no versions left
```

giordano:
https://github.com/JuliaRegistries/General/pull/69803/files didn't update compats, because Pkg doesn't like chaning compat bounds within the same patch version.  The solution is to either submit a PR to General to fix that or make here another PR to bump the offset (I'd slightly prefer the former)

PR #5642, opened 2022-10-08
[Triangle] fix crash on windows + new clean upstream source

j-fu:
* Fix issue https://github.com/JuliaGeometry/Triangulate.jl/issues/15 by replacing `unsigned long` with `unsigned long long` on x86_64_w64_mingw32

* Upstream source with all patches is now https://github.com/JuliaGeometry/Triangle

* Synchronized version number with upstream source

### DIFF
--- a/T/Triangle/build_tarballs.jl
+++ b/T/Triangle/build_tarballs.jl
@@ -16,10 +16,10 @@ mkdir -p "${libdir}"
 
 GENERICFLAGS=(-DREAL=double  -DTRILIBRARY  -DNDEBUG -DNO_TIMER -DEXTERNAL_TEST)
 
-ARCHFLAGS=(-DULONG="unsigned long" -fPIC)
-
 if [[ "${target}" == x86_64-w64-mingw32 ]]; then
     ARCHFLAGS=(-DULONG="unsigned long long")
+else
+    ARCHFLAGS=(-DULONG="unsigned long" -fPIC)
 fi
 
 $CC "${ARCHFLAGS[@]}"  "${GENERICFLAGS[@]}" $LDFLAGS -O3  --shared -o "${libdir}/libtriangle.${dlext}" triangle.c triwrapjulia.c
--- a/T/Triangle/build_tarballs.jl
+++ b/T/Triangle/build_tarballs.jl
@@ -1,46 +1,36 @@
 using BinaryBuilder
 
 name = "Triangle"
-version = v"1.6.1"
 
-# Collection of sources required to build Triangle by J. Shewchuk
-#
-# More info:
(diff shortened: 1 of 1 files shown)


j-fu:
See [this discussion](https://github.com/JuliaGeometry/Triangle/issues/1) on tracking upstream source.

T/Triangle/build_tarballs.jl
```diff
@@ -1,46 +1,36 @@
...
-$CC -Itriangle  -Wno-int-to-pointer-cast  -Wno-pointer-to-int-cast -DREAL=double -DTRILIBRARY -O3 -fPIC -DNDEBUG -DNO_TIMER -DEXTERNAL_TEST $LDFLAGS --shared -o "${libdir}/libtriangle.${dlext}" triangle_patched.c triwrap.c
-
-install_license triangle/README
+GENERICFLAGS=(-DREAL=double  -DTRILIBRARY  -DNDEBUG -DNO_TIMER -DEXTERNAL_TEST)
+
+ARCHFLAGS=(-DULONG="unsigned long" -fPIC)
+
+if [[ "${target}" == x86_64-w64-mingw32 ]]; then
+    ARCHFLAGS=(-DULONG="unsigned long long")
+fi
```
giordano, line 23: I think
```suggestion
if [[ "${target}" == x86_64-w64-mingw32 ]]; then
    ARCHFLAGS=(-DULONG="unsigned long long")
else
    ARCHFLAGS=(-DULONG="unsigned long" -fPIC)
fi
```
read slightly clearer (at the beginning I thought the `if` was adding new elements to the original `ARCHFLAGS`, instead they're just two different settings).

T/Triangle/build_tarballs.jl
```diff
@@ -1,46 +1,36 @@
...
+
+$CC "${ARCHFLAGS[@]}"  "${GENERICFLAGS[@]}" $LDFLAGS -O3  --shared -o "${libdir}/libtriangle.${dlext}" triangle.c triwrapjulia.c
```
giordano, line 25: There is no `LDFLAGS`
```suggestion
$CC "${ARCHFLAGS[@]}"  "${GENERICFLAGS[@]}" -O3  --shared -o "${libdir}/libtriangle.${dlext}" triangle.c triwrapjulia.c
```

T/Triangle/build_tarballs.jl
```diff
@@ -1,46 +1,36 @@
...
+
+install_license README
 """
 
 platforms = supported_platforms(; experimental=true)
```
giordano, line 30: This PR didn't change this line, but while we're at it we can remove some craft (`experimental=true` has no effect nowadays):
```suggestion
platforms = supported_platforms()
```

PR #5644, opened 2022-10-09
[protozero] Add header only lib

jeremiahpslewis:


giordano:
> Can't figure out why the build is failing...

For the record, this is failing with
```
[19:06:35] /opt/x86_64-linux-musl/x86_64-linux-musl/sys-root/usr/include/stdlib.h:99:5: error: from previous declaration ‘int posix_memalign(void**, size_t, size_t)’
[19:06:35]  int posix_memalign (void **, size_t, size_t);
[19:06:35]      ^
```

When else fails, [search in the repository](https://github.com/JuliaPackaging/Yggdrasil/search?q=posix_memalign).

> run against the `.devcontainer` with this command: `julia --project=../../.ci build_tarballs.jl --debug x86_64-linux-gnu`, it goes through fine :shrug:

You're building for a different platform :shrug:

jeremiahpslewis:
The line `/opt/x86_64-linux-musl/lib/gcc/x86_64-linux-musl/5.2.0/` in the logs seems to suggest that GCC 5.2.0 is being used even though I haven’t specified a gcc floor. Is there a reason why `AnyPlatform` doesn’t use 4.8?

giordano:
>  Is there a reason why `AnyPlatform` doesn’t use 4.8?

* https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/dd8ee21b49f57bc288a48160ea3dbddb4bf284e3/src/Runner.jl#L8-L16
* https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/dd8ee21b49f57bc288a48160ea3dbddb4bf284e3/src/Rootfs.jl#L679-L680
* https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/dd8ee21b49f57bc288a48160ea3dbddb4bf284e3/src/Rootfs.jl#L448-L454

jeremiahpslewis:
Thanks, that was an exciting story in three parts (hadn’t seen the third snippet, explains it all).

giordano:
In case the solution wasn't clear yet: https://docs.binarybuilder.org/dev/troubleshooting/#Error-in-definition-of-posix_memalign

jeremiahpslewis:
@giordano Thanks again, it builds! :)

P/protozero/build_tarballs.jl
```diff
@@ -0,0 +1,40 @@
...
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [AnyPlatform()]
+
+# The products that we will ensure are always built
+# protozero is a header only library, so no binary products
+products = Product[]
```
giordano, line 34: No products?

jeremiahpslewis: I was following the template for other header-only projects, like https://github.com/JuliaPackaging/Yggdrasil/blob/c063b0692d36ef5c41589a92738a50bfc94ef9ff/T/Taskflow/build_tarballs.jl#L33. Is it better to include file products, and if so, do I need to have a file product for every `*.hpp` generated?

PR #5652, opened 2022-10-11
add Graphviz as dependency to PProf

vchuravy:
Alternative to https://github.com/JuliaPerf/PProf.jl/pull/71

### DIFF
--- a/P/pprof/build_tarballs.jl
+++ b/P/pprof/build_tarballs.jl
@@ -32,7 +32,7 @@ products = [
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[
+dependencies = [
     Dependency("Graphviz_jll"),
 ]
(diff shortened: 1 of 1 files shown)


P/pprof/build_tarballs.jl
```diff
@@ -33,6 +33,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[
```
giordano, line 35:
```suggestion
dependencies = [
```

giordano:
In the build log I don't see any references to graphviz, is it actually used by pprof directly?

vchuravy:
Only when you execute the pprof binary, it calls out to the dot binary. IIUC adding it as a dependency here will set up path/ld_library_path such that the pprof binary should be able to find it.

PR #5661, opened 2022-10-13
Updating Exodus_jll for windows build

cmhamel:
…succesfully build. This came down to a nc-config issue with windows that threw an error by default in the seacas build.

### DIFF
--- a/E/Exodus/build_tarballs.jl
+++ b/E/Exodus/build_tarballs.jl
@@ -3,26 +3,56 @@
 using BinaryBuilder, Pkg
 
 name = "Exodus"
-version = v"0.1.1"
+version = v"0.1.2"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


E/Exodus/build_tarballs.jl
```diff
@@ -3,26 +3,55 @@
...
 
 name = "Exodus"
-version = v"0.1.1"
+version = v"0.1.2"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/gsjaardema/seacas.git", "a1da779b061fbdc750f18bcae29295dc5064cb74")
-    # GitSource("https://github.com/gsjaardema/seacas.git", "1452c325ab5d507d000397c713a5b0c4dc57bf81")
+    GitSource("https://github.com/cmhamel/seacas.git", "177d194ad2a104a21704091e2646af291f5c0a47")
```
giordano, line 10: Why not using the upstream project instead?

PR #5662, opened 2022-10-13
Wizard recipe: GstPluginsBase-v1.20.3

lucifer1004:
This is an attempt to build plugins for GStreamer so that it can be more useful.

### DIFF
--- a/G/GstPluginsBase/build_tarballs.jl
+++ b/G/GstPluginsBase/build_tarballs.jl
@@ -14,12 +14,6 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir
 meson --cross-file=${MESON_TARGET_TOOLCHAIN} --buildtype=release gst-plugins-base-*
-
-# Meson beautifully forces thin archives, without checking whether the dynamic linker
-# actually supports them: <https://github.com/mesonbuild/meson/issues/10823>.  Let's remove
-# the (deprecated...) `T` option to `ar`, until they fix it in Meson.
-# sed -i.bak 's/csrDT/csrD/' build.ninja
-
 ninja -j${nproc}
 ninja install
 install_license gst-plugins-base-*/COPYING
@@ -65,7 +59,6 @@ dependencies = [
     HostBuildDependency("Gettext_jll")
     Dependency(PackageSpec(name="GStreamer_jll", uuid="aaaaf01e-2457-52c6-9fe8-886f7267d736"); compat="1.20.3")
     Dependency(PackageSpec(name="Opus_jll", uuid="91d4177d-7536-5919-b921-800302f37372"))
-    # Dependency(PackageSpec(name="ORC_jll", uuid="fb41591b-4dee-5dae-bf56-d83afd04fbc0"))
     Dependency(PackageSpec(name="Pango_jll", uuid="36c8627f-9965-5494-a995-c6b170f724f3"); compat="1.47.0")
 ]
 
--- a/G/GstPluginsBase/build_tarballs.jl
+++ b/G/GstPluginsBase/build_tarballs.jl
@@ -65,7 +65,7 @@ dependencies = [
     HostBuildDependency("Gettext_jll")
     Dependency(PackageSpec(name="GStreamer_jll", uuid="aaaaf01e-2457-52c6-9fe8-886f7267d736"); compat="1.20.3")
     Dependency(PackageSpec(name="Opus_jll", uuid="91d4177d-7536-5919-b921-800302f37372"))
-    Dependency(PackageSpec(name="ORC_jll", uuid="fb41591b-4dee-5dae-bf56-d83afd04fbc0"))
+    # Dependency(PackageSpec(name="ORC_jll", uuid="fb41591b-4dee-5dae-bf56-d83afd04fbc0"))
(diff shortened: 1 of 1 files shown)


lucifer1004:
@giordano The errors occurred when meson was handling localization files, very strange.

G/GstPluginsBase/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://gstreamer.freedesktop.org/src/gst-plugins-base/gst-plugins-base-$(version).tar.xz", "7e30b3dd81a70380ff7554f998471d6996ff76bbe6fc5447096f851e24473c9f")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+meson --cross-file=${MESON_TARGET_TOOLCHAIN} --buildtype=release gst-plugins-base-*
+sed -i.bak 's/csrDT/csrD/' build.ninja
```
giordano, line 17: Please add a comment to explain what this is doing

G/GstPluginsBase/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+    ExecutableProduct("gst-play-1.0", :gstplay),
+    ExecutableProduct("gst-discoverer-1.0", :gstdiscoverer)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="GStreamer_jll", uuid="aaaaf01e-2457-52c6-9fe8-886f7267d736"))
+    Dependency(PackageSpec(name="Opus_jll", uuid="91d4177d-7536-5919-b921-800302f37372"))
+    Dependency(PackageSpec(name="ORC_jll", uuid="fb41591b-4dee-5dae-bf56-d83afd04fbc0"))
+    Dependency(PackageSpec(name="Pango_jll", uuid="36c8627f-9965-5494-a995-c6b170f724f3"))
```
giordano, line 62: #3024

giordano (commented):
> @giordano The errors occurred when meson was handling localization files, very strange.

From your latest PR: https://github.com/JuliaPackaging/Yggdrasil/blob/79a855fd8926282e5aedfe2d2ef23efa995eaffd/G/GStreamer/build_tarballs.jl#L58-L59

G/GstPluginsBase/build_tarballs.jl
lucifer1004: I will add compat information after all builds turn green.

lucifer1004:
Seems that `gettext` is not enough for some platforms.

giordano:
> Seems that `gettext` is not enough for some platforms.

When else fails, read the documentation: https://docs.binarybuilder.org/stable/troubleshooting/#Running-foreign-executables

G/GstPluginsBase/build_tarballs.jl
```diff
@@ -0,0 +1,73 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    # Need a host gettext for msgfmt
+    HostBuildDependency("Gettext_jll")
+    Dependency(PackageSpec(name="GStreamer_jll", uuid="aaaaf01e-2457-52c6-9fe8-886f7267d736"); compat="1.20.3")
+    Dependency(PackageSpec(name="Opus_jll", uuid="91d4177d-7536-5919-b921-800302f37372"))
+    Dependency(PackageSpec(name="ORC_jll", uuid="fb41591b-4dee-5dae-bf56-d83afd04fbc0"))
+    Dependency(PackageSpec(name="Pango_jll", uuid="36c8627f-9965-5494-a995-c6b170f724f3"); compat="~1.47.0"))
```
giordano: Why `~`?  That's overly restrictive and ***not*** what's suggested in #3024.  Also, there is an extra `)`

lucifer1004: Fixed.

G/GstPluginsBase/build_tarballs.jl
```diff
@@ -0,0 +1,73 @@
...
+    ExecutableProduct("gst-discoverer-1.0", :gstdiscoverer)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    # Need a host gettext for msgfmt
+    HostBuildDependency("Gettext_jll")
+    Dependency(PackageSpec(name="GStreamer_jll", uuid="aaaaf01e-2457-52c6-9fe8-886f7267d736"); compat="1.20.3")
+    Dependency(PackageSpec(name="Opus_jll", uuid="91d4177d-7536-5919-b921-800302f37372"))
+    # Dependency(PackageSpec(name="ORC_jll", uuid="fb41591b-4dee-5dae-bf56-d83afd04fbc0"))
```
giordano, line 68: Since this was failing to run orcc, maybe this is a [***build*** dependency](https://docs.binarybuilder.org/stable/build_tips/#Dependencies-for-the-target-system-vs-host-system)?

lucifer1004:
How do I fix the following for Windows?

```
##[warning]lib/gstreamer-1.0/libgstadder.dll should be in `bin`!
```

Do I need to manually move them?

lucifer1004:
I need to play with the jll package to have a clearer idea. Also in the future, we need to add X11 support for this package.

PR #5669, opened 2022-10-13
Add neper recipe.

fredrikekre:


N/neper/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir/neper-*
+mkdir build && cd build
+if [[ ${target} == *-apple-* ]]; then
+    HAVE_OPENMP=0
+else
+    HAVE_OPENMP=1
+fi
```
giordano, line 17: https://github.com/JuliaPackaging/Yggdrasil/blob/69653bccf74a289e8fa97a548e4dbc294e7b3e30/B/blis/build_tarballs.jl#L135-L138

N/neper/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+
+dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
+    Dependency(PackageSpec(name="GSL_jll", uuid="1b77fbbe-d8ee-58f0-85f9-836ddc23a7a4")),
+    Dependency(PackageSpec(name="NLopt_jll", uuid="079eb43e-fd8e-5478-9966-2cf3e3edb778")),
+    # /opt/aarch64-linux-gnu/bin/../lib/gcc/aarch64-linux-gnu/12.1.0/../../../../aarch64-linux-gnu/bin/ld: /opt/aarch64-linux-gnu/aarch64-linux-gnu/sys-root/usr/local/lib/libscotch.so: undefined reference to `gzwrite'
+    # Dependency(PackageSpec(name="SCOTCH_jll", uuid="a8d0f55d-b80e-548d-aff6-1a04c175f0f9")),
+]
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"12.1.0")
```
giordano, line 55: GCC 12 is not going to work on any released version of Julia :slightly_smiling_face:   For maximum compatibility we want to use the _oldest_ version of GCC, not the latest one.

N/neper/build_tarballs.jl
fredrikekre: Thanks

N/neper/build_tarballs.jl
fredrikekre: Right, I got some errors with gcc 4 and just switch to the latest to get things to run. O am guessing it needs to be 9 (https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SCOTCH/build_tarballs.jl#L52)

giordano: You don't _necessarily_ need the same version as the one used for the dependencies, although that's a good initial guess.  It's really a trial and error process.

fredrikekre: Okay, but in this case neper have a vendored libscotch. I couldn't get it to work when using the JLL dependency as indicated below. Maybe they have some changes to it, or I did something wrong.

giordano: The error is likely a missing `-lz` in the compiler invocation (I don't see it in the error you shared).  In https://github.com/JuliaPackaging/Yggdrasil/blob/69653bccf74a289e8fa97a548e4dbc294e7b3e30/S/SCOTCH/build_tarballs.jl#L15-L31 it looks like we're using a custom makefile which likely doesn't produce a pkgconfig file, so this build system can't know that libscotch needs linking to zlib as well

fredrikekre: Just to clarify, you mean that that information is missing from the `SCOTCH_jll` but that I can fix it by adding `-lz` to `neper`s compiler invocation in this PR?

N/neper/build_tarballs.jl
```diff
@@ -0,0 +1,59 @@
...
+        #  install FILES given no DESTINATION!
+        return true
+    elseif get(p.tags, "os", nothing) == "windows" && get(p.tags, "arch", nothing) == "x86_64"
+        # CMake Error at CMakeLists.txt:262 (install):
+        #  install FILES given no DESTINATION!
+        return true
+    elseif get(p.tags, "os", nothing) == "freebsd"
+        # CMake Error at CMakeLists.txt:262 (install):
+        #  install FILES given no DESTINATION!
+        return true
```
fredrikekre, line 40: This looks to be https://github.com/neperfepx/neper/blob/886d666e038507af5777dc12e7192b4f4d90ac7e/src/CMakeLists.txt#L256-L264

N/neper/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+make install
+"""
+
+function exclude(p)
+    if get(p.tags, "libc", nothing) == "musl" || get(p.tags, "os", nothing) == "freebsd"
+        # src/contrib/ut/src/ut_print/ut_print.c:1685:38: error: parameter 1 (‘beg_time’) has incomplete type
+        #  1685 | ut_print_elapsedtime (struct timeval beg_time, struct timeval end_time)
+        #       |                       ~~~~~~~~~~~~~~~^~~~~~~~
+        return true
+    elseif get(p.tags, "os", nothing) == "windows"
```
fredrikekre, line 30: https://github.com/JuliaPackaging/Yggdrasil/blob/65965a4631866816e6062dfdb592d580b895b56f/S/SCOTCH/build_tarballs.jl#L35

N/neper/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+make -j${nproc}
+make install
+"""
+
+function exclude(p)
+    if get(p.tags, "libc", nothing) == "musl" || get(p.tags, "os", nothing) == "freebsd"
+        # src/contrib/ut/src/ut_print/ut_print.c:1685:38: error: parameter 1 (‘beg_time’) has incomplete type
+        #  1685 | ut_print_elapsedtime (struct timeval beg_time, struct timeval end_time)
+        #       |                       ~~~~~~~~~~~~~~~^~~~~~~~
+        return true
```
fredrikekre, line 29: This can probably be fixed with a patch, but I am not experienced enough with C to do that I'm afraid.

fredrikekre:
There is a runtime dependency on the `gmsh` binary too, should `gmsh_jll` be added here then or not? I am guessing that binary is inluded there, even though not explicitly included in https://github.com/JuliaPackaging/Yggdrasil/blob/65965a4631866816e6062dfdb592d580b895b56f/G/gmsh/build_tarballs.jl#L45-L48

N/neper/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+      -DCMAKE_BUILD_TYPE=Release \
+      -DCMAKE_CXX_FLAGS="${CXXFLAGS}" \
+      -DCMAKE_INSTALL_PREFIX_COMPLETION_FULL=$prefix/completion \
+      ../src
+make -j${nproc}
+make install
+"""
+
+function exclude(p)
+    if get(p.tags, "libc", nothing) == "musl" || get(p.tags, "os", nothing) == "freebsd"
```
giordano, line 25:
```suggestion
    if libc(p) == "musl" || Sys.isfreebsd(p)
```

N/neper/build_tarballs.jl
giordano, line 30:
```suggestion
    Sys.iswindows(p)
```

N/neper/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+platforms = expand_cxxstring_abis(platforms)
+
+products = [
+    ExecutableProduct("neper", :neper)
+]
+
+dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
+    Dependency(PackageSpec(name="GSL_jll", uuid="1b77fbbe-d8ee-58f0-85f9-836ddc23a7a4")),
```
giordano, line 47: See #3024
```suggestion
    Dependency(PackageSpec(name="GSL_jll", uuid="1b77fbbe-d8ee-58f0-85f9-836ddc23a7a4"); compat="~2.7.2"),
```

N/neper/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+script = raw"""
+cd $WORKSPACE/srcdir/neper-*
+mkdir build && cd build
+CXXFLAGS="-lz"
+cmake -DCMAKE_INSTALL_PREFIX=$prefix \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DCMAKE_CXX_FLAGS="${CXXFLAGS}" \
+      -DCMAKE_INSTALL_PREFIX_COMPLETION_FULL=$prefix/completion \
```
giordano, line 18: I'd put this is in `share/` just not to have anything "non-standard" in the prefix, although this isn't much relevant for us
```suggestion
      -DCMAKE_INSTALL_PREFIX_COMPLETION_FULL=$prefix/share/completion \
```

giordano (commented):
> There is a runtime dependency on the gmsh binary too, should gmsh_jll be added here then or not?

Probably yes, maybe as a [`RuntimeDependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.RuntimeDependency) if it isn't needed at all during the build.

> I am guessing that binary is inluded there, even though not explicitly included in https://github.com/JuliaPackaging/Yggdrasil/blob/65965a4631866816e6062dfdb592d580b895b56f/G/gmsh/build_tarballs.jl#L45-L48

Yes, but I don't remember if the directory where `gmsh` is in will be added to the `PATH`, probably not, so neper may struggle finding `gmsh`

N/neper/build_tarballs.jl
fredrikekre: Makes sense. Turns out this isn't even needed for any of the working platforms since I got other errors down the line for other platforms.

fredrikekre:
> Yes, but I don't remember if the directory where gmsh is in will be added to the PATH, probably not, so neper may struggle finding gmsh

Yea, I just noticed this when trying to add it to `ygg`: #5674.

N/neper/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+make install
+"""
+
+function exclude(p)
+    if libc(p) == "musl" || Sys.isfreebsd(p)
+        # src/contrib/ut/src/ut_print/ut_print.c:1685:38: error: parameter 1 (‘beg_time’) has incomplete type
+        #  1685 | ut_print_elapsedtime (struct timeval beg_time, struct timeval end_time)
+        #       |                       ~~~~~~~~~~~~~~~^~~~~~~~
+        return true
+    Sys.iswindows(p)
```
fredrikekre, line 30:
```suggestion
    if Sys.iswindows(p)
```

N/neper/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+make install
+"""
+
+function exclude(p)
+    if libc(p) == "musl" || Sys.isfreebsd(p)
+        # src/contrib/ut/src/ut_print/ut_print.c:1685:38: error: parameter 1 (‘beg_time’) has incomplete type
+        #  1685 | ut_print_elapsedtime (struct timeval beg_time, struct timeval end_time)
+        #       |                       ~~~~~~~~~~~~~~~^~~~~~~~
+        return true
+    if Sys.iswindows(p)
```
fredrikekre, line 30:
```suggestion
    elseif Sys.iswindows(p)
```

PR #5676, opened 2022-10-14
[dev] Ensure environment/dependencies pre-initialized

jeremiahpslewis:


giordano:
I think this should have removed https://github.com/JuliaPackaging/Yggdrasil/blob/2d8fe85767241e27c686af2e46fd4b698a961c75/.devcontainer/Dockerfile#L5 now BinaryBuilder is installed twice with different versions.

PR #5678, opened 2022-10-14
Add `flashrom`

staticfloat:
This is technically not `v1.2`, but instead some random `master` version, because the board I actually need to flash is too new to appear in `v1.2` proper.  So we denote this by calling it `v1.2.900`, in hopes that there will not be a large number of upstream releases on the `v1.2` branch.

### DIFF
--- a/F/flashrom/build_tarballs.jl
+++ b/F/flashrom/build_tarballs.jl
@@ -1,0 +1,38 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Base.BinaryPlatforms
+
+name = "flashrom"
+version = v"1.2.900"
+
+# Collection of sources required to complete build
+sources = [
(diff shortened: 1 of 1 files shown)


F/flashrom/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+make PREFIX=${prefix} -j${nproc}
+make PREFIX=${prefix} install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = filter(Sys.islinux, supported_platforms())
+
+# ppc64le kernel headers might be broken, disable for now
+platforms = filter(p -> arch(p) != "powerpc64le", supported_platforms())
```
giordano, line 26: Was this meant to be 
```suggestion
platforms = filter(p -> arch(p) != "powerpc64le", platforms)
```
or more simply
```suggestion
filter!(p -> arch(p) != "powerpc64le", platforms)
```
?

staticfloat: 😳  yes.....

PR #5680, opened 2022-10-14
[LibCURL] disable `brotli` linkage

t-bltg:
Per https://github.com/JuliaLang/julia/pull/47165#issuecomment-1279459371.


L/LibCURL/common.jl
```diff
@@ -30,9 +30,10 @@ function build_libcurl(ARGS, name::String)
         --without-ssl --without-gnutls
         --without-libidn2 --without-librtmp
         --without-nss --without-libpsl
-	--disable-ares --disable-manual
+        --disable-ares --disable-manual
```
t-bltg, line 33: Evil `tab`.

Issue #5673, opened 2022-10-14
Build glib-networking


lucifer1004:
After #5672 is merged, glib-networking can be built, which should then fix #5656.

t-bltg:
Xref https://github.com/JuliaPackaging/Yggdrasil/pull/5689.

ViralBShah:
Done in #6530

PR #5683, opened 2022-10-15
[GnuTLS] Enable p11 kit

lucifer1004:


lucifer1004:
This will enable #5673

PR #5684, opened 2022-10-15
Wizard recipe: P11Kit-v0.24.1

lucifer1004:

* Package name: P11Kit
* Version: v0.24.1

### DIFF
--- a/P/P11Kit/build_tarballs.jl
+++ b/P/P11Kit/build_tarballs.jl
@@ -14,6 +14,12 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir
 meson --cross-file=${MESON_TARGET_TOOLCHAIN} --buildtype=release p11-kit-*
+
+# Meson beautifully forces thin archives, without checking whether the dynamic linker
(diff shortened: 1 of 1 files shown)


lucifer1004:
This will enable #5683

lucifer1004:
@giordano Seems that on Windows, some built products are installed to different locations. How should I debug this locally?

jeremiahpslewis:
@lucifer1004

1. From the `P11Kit` directory, run `julia —project=../../.ci build_tarballs.jl --debug —verbose x86_64-w64-mingw32` to debug the x86 windows platform, you'll be dropped into a debug shell upon error
2. You probably want to copy the build artifacts into the appropriate directory `${libdir}` or `${bindir}` as part of the build script, this way you don't have to specify the path in the `products` section of the code and can special case windows, like here: https://github.com/JuliaPackaging/Yggdrasil/blob/0d10b8e80ded9cbdab0e028054b26a0a2c9e1ed5/C/cilantro/build_tarballs.jl#L31

lucifer1004:
I checked and found that a library was not built on Windows. How should we handle this situation?

lucifer1004:
BTW, I cannot simply move the products to libdir and bindir because they depend on the relative path.

PR #5688, opened 2022-10-15
[Glib] Add networking support

lucifer1004:
`glib-networking` itself does not produce standalone products.

Related information:

```
The build has produced several libraries and executables.
Please select which of these you want to consider `products`.
These are generally those artifacts you will load or use from julia.

[press: d=done, a=all, n=none]
   [ ] lib/gio/modules/libgioenvironmentproxy.so
 > [ ] lib/gio/modules/libgiognutls.so
```

### DIFF
--- a/G/Glib/build_tarballs.jl
+++ b/G/Glib/build_tarballs.jl
@@ -86,7 +86,7 @@ dependencies = [
     Dependency("Gettext_jll", v"0.21.0"; compat="=0.21.0"),
     Dependency("PCRE2_jll"; compat="10.35"),
     Dependency("Zlib_jll"),
-    Dependency("GnuTLS_jll"; compat="3.7.8+1"),
+    Dependency("GnuTLS_jll"; compat="3.7.8"),
     Dependency("Libmount_jll"; platforms=filter(Sys.islinux, platforms)),
 ]
 
--- a/G/Glib/build_tarballs.jl
+++ b/G/Glib/build_tarballs.jl
@@ -7,12 +7,14 @@ version = v"2.74.0"
 sources = [
     ArchiveSource("https://ftp.gnome.org/pub/gnome/sources/glib/$(version.major).$(version.minor)/glib-$(version).tar.xz",
                   "3652c7f072d7b031a6b5edd623f77ebc5dcd2ae698598abcc89ff39ca75add30"),
+    ArchiveSource("https://ftp.gnome.org/pub/gnome/sources/glib/$(version.major).$(version.minor)/glib-networking-$(version).tar.xz",
+                  "1f185aaef094123f8e25d8fa55661b3fd71020163a0174adb35a37685cda613b"),
     DirectorySource("./bundled"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/glib-*/
+cd $WORKSPACE/srcdir/glib-2*/
 install_license COPYING
 
 if [[ "${target}" == *-freebsd* ]]; then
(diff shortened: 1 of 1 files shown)


G/Glib/build_tarballs.jl
```diff
@@ -7,12 +7,14 @@ version = v"2.74.0"
 sources = [
     ArchiveSource("https://ftp.gnome.org/pub/gnome/sources/glib/$(version.major).$(version.minor)/glib-$(version).tar.xz",
                   "3652c7f072d7b031a6b5edd623f77ebc5dcd2ae698598abcc89ff39ca75add30"),
+    ArchiveSource("https://ftp.gnome.org/pub/gnome/sources/glib-networking/$(version.major).$(version.minor)/glib-networking-$(version).tar.xz",
+                  "1f185aaef094123f8e25d8fa55661b3fd71020163a0174adb35a37685cda613b"),
```
giordano, line 11: Why this isn't a different package?

lucifer1004: It does not produce standalone products.

giordano: The wizard is dumb, it can't possibly deal with all situations out there.  It only searches for libraries and binary executables in fixed directories, if a package does something else the wizard will not know.  Surely actually building glib-networking will result in _something_.  What I don't know, but I guess it's your interest figuring this out.

lucifer1004:
@giordano `glib-networking` products need to be put in `lib/gio/modules` of `glib` so it cannot be a standalone package.

giordano:
> so it cannot be a standalone package.

I'm missing why.

lucifer1004:
OK, let me make another PR and we need to test whether that will work.

giordano:
> OK, let me make another PR and we need to test whether that will work.

I think it'd be better for you to test locally: I'm not thrilled by the idea of publishing a package which may just never use any longer.

lucifer1004:
OK I will try. I need to figure out how to install a jll package locally.

giordano:
>  I need to figure out how to install a jll package locally.

As usual, reading the documentation may actually help: https://docs.binarybuilder.org/stable/FAQ/#Can-I-publish-a-JLL-package-locally-without-going-through-Yggdrasil?-1

lucifer1004:
I bundled a local jll package but when I run the following `build_tarballs.jl`

```julia
using BinaryBuilder
using Pkg: PackageSpec

name = "Soup"
version = v"3.2.1"

# Collection of sources required to build Soup
sources = [
    ArchiveSource("https://download.gnome.org/sources/libsoup/$(version.major).$(version.minor)/libsoup-$(version).tar.xz",
                  "b1eb3d2c3be49fbbd051a71f6532c9626bcecea69783190690cd7e4dfdf28f29"),
]

script = raw"""
cd $WORKSPACE/srcdir/libsoup-*/

mkdir build_glib && cd build_glib
meson --cross-file="${MESON_TARGET_TOOLCHAIN}" \
    --buildtype=release \
    ..
ninja -j${nproc}
ninja install
"""

products = [
    LibraryProduct(["libsuop"], :libsoup),
]

dependencies = [
    Dependency("Glib_jll"; compat="2.74.0"),
    Dependency("SQLite_jll"),
    Dependency("nghttp2_jll"),
    Dependency("brotli_jll"),
    Dependency("libpsl_jll"),
    Dependency("LibUnwind_jll"),
    Dependency(PackageSpec(; name = "GlibNetworking_jll",  uuid = "99fd4003-298c-58dc-a8c7-c8e9475755a1", url = "https://github.com/lucifer1004/GlibNetworking_jll.jl.git")),
]

```

Julia complained that `ERROR: LoadError: expected package `GlibNetworking_jll [99fd4003]` to be registered`

giordano:
Did you try with the path, as suggested in the documentation?

lucifer1004:
Close this in favor of #5689

PR #5689, opened 2022-10-15
Wizard recipe: GlibNetworking-v2.74.0

lucifer1004:

* Package name: GlibNetworking
* Version: v2.74.0


lucifer1004:
@giordano I have confirmed locally that this worked (for building soup).

giordano:
While you're at it, did you also try using the libsoup library with the generated JLL package?

lucifer1004:
Not yet, since soup did not compile successfully (it was stuck at `meson` step without glib-networking, and now there was some other issue)

lucifer1004:
I encountered `x86_64-linux-gnu-gcc: error: unrecognized command line option ‘-R’` during linking.

lucifer1004:
Need to solve OSX issues. I kept both `.dylib` and `.so` but when I tried to build libsoup, glib-networking was still not recognized.

lucifer1004:
Updated soup build recipe:

```julia
using BinaryBuilder
using Pkg: PackageSpec

name = "Soup"
version = v"3.2.1"

# Collection of sources required to build Soup
sources = [
    ArchiveSource("https://download.gnome.org/sources/libsoup/$(version.major).$(version.minor)/libsoup-$(version).tar.xz",
                  "b1eb3d2c3be49fbbd051a71f6532c9626bcecea69783190690cd7e4dfdf28f29"),
]

script = raw"""
cd $WORKSPACE/srcdir/libsoup-*/
install_license COPYING

mkdir build_glib && cd build_glib
meson --cross-file="${MESON_TARGET_TOOLCHAIN}" \
    --buildtype=release \
    ..

# https://github.com/rbgirshick/py-faster-rcnn/issues/706
sed -i "s/-R/-Wl,-rpath=/" build.ninja

# https://stackoverflow.com/questions/2418157/c-error-undefined-reference-to-clock-gettime-and-clock-settime
sed -i "s/\$ARGS -o \$out \$in \$LINK_ARGS/\$ARGS -o \$out \$in \$LINK_ARGS -lrt/" build.ninja

ninja -j${nproc}
ninja install
"""

products = [
    LibraryProduct(["libsoup", "libsoup-3.0"], :libsoup),
]

dependencies = [
    Dependency("Glib_jll"; compat="2.74.0"),
    Dependency("SQLite_jll"),
    Dependency("nghttp2_jll"),
    Dependency("brotli_jll"),
    Dependency("libpsl_jll"),
    Dependency("LibUnwind_jll"),
    Dependency(PackageSpec(; name = "GlibNetworking_jll",  uuid = "99fd4003-298c-58dc-a8c7-c8e9475755a1",path="/home/ubuntu/.julia/dev/GlibNetworking_jll")),
]

```

@giordano This recipe worked for Linux, but not MacOS.

G/GlibNetworking/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+# actually supports them: <https://github.com/mesonbuild/meson/issues/10823>.  Let's remove
+# the (deprecated...) `T` option to `ar`
+sed -i.bak 's/csrDT/csrD/' build.ninja
+
+ninja -j${nproc}
+
+# Darwin products are also .so instead of .dylib so we need to rename
+mkdir -p ${libdir}/gio/modules
+cp proxy/environment/libgioenvironmentproxy.so ${libdir}/gio/modules/libgioenvironmentproxy.${dlext}
+cp tls/gnutls/libgiognutls.so ${libdir}/gio/modules/libgiognutls.${dlext}
```
giordano, line 31: Are you sure this is correct?  These are usually plugins which ***need*** to be called `*.so`.

lucifer1004: No. I could not make OSX work. As I said above, neither `.so` nor `.dylib` worked. Maybe I should just exclude OSX for now.

giordano: Can you elaborate on what "work" means here?

lucifer1004: Soup could not build on OSX even if `libgiognutls` and `libgioenvironmentproxy` were both in `lib/gio/modules` (and both `.so` and `.dylib` were there). The error was the same as when there was no `glib-networking`.

giordano: https://docs.binarybuilder.org/stable/troubleshooting/#General-comments
> The best recommendation is to access the build environment and carefully read the log files generated by the build systems: it is not uncommon that build systems would only print to screen misleading error messages, and the actual problem is completely different (e.g. "library XYZ can't be found", when the problem instead is that the command they run to look for library XYZ fails for unrelated reasons, for example for a wrong compiler flag used in the check). Having an understanding of what the build system is doing would also be extremely useful.

lucifer1004:
@giordano 

- In the latest commit, I just copied `.so` files, and then `BinaryBuilder` failed to find the library products because they were `.so` instead of the expected `.dylib`. This is why I used renaming. I also tested making two copies, one `.so` and one `.dylib`. This could make both `BinaryBuilder` and `gio` happy, but duplication is not elegant. Actually, the two `.so` libraries should never be used directly, instead, they should be loaded as modules by other libraries. So I think it might be better not to make `glib-networking` a standalone package, as I did in #5688.
- I just tested locally and found I could bypass the TLS check with the meson option `-Dtls_check=false`, but the build still failed due to  `/workspace/destdir/bin/glib-compile-resources: line 1: syntax error: unexpected "("` issue, which was strange because I have added `Gettext` as a dependency for Apple and that should solve this.

lucifer1004:
OSX build of `soup` failed at the last step:

```
[09:51:13] Installing libsoup/libsoup-3.0.0.dylib to /workspace/destdir/lib
...
[ Info: Modifying dylib id from "/workspace/destdir/lib/libsoup-3.0.0.dylib" to "@rpath/libsoup-3.0.0.dylib"
...
[ Info: Could not locate libsoup, libsoup-3.0, libsoup-3.0.0 inside ["/home/ubuntu/com.github/lucifer1004/Yggdrasil/S/Soup/build/x86_64-apple-darwin/qF6n2ATl/x86_64-apple-darwin14-libgfortran3-cxx03/destdir/lib64", "/home/ubuntu/com.github/lucifer1004/Yggdrasil/S/Soup/build/x86_64-apple-darwin/qF6n2ATl/x86_64-apple-darwin14-libgfortran3-cxx03/destdir/lib"]
```

giordano:
https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct

> If you are unsure what value to use for `libname`, you can use `Base.BinaryPlatforms.parse_dl_name_version`:

```julia
julia> using Base.BinaryPlatforms

julia> parse_dl_name_version("libsoup-3.0.0.dylib", "macos")[1]
"libsoup-3"
```

lukebemish:
Is there specific work that needs to be done with this PR? Is the current limitation just that it isn't working on macos? I can work on this a bit more if someone needs to.

Additionally, is there a reason that current buildscript does not build for windows? I may be misunderstanding some limitations of glib-networking here, but as far as I'm aware the modules should still be loaded there.

lukebemish:
Ah, windows is probably disabled because of GnuTLS not being there; the OpenSSL module could still be built, though.

G/GlibNetworking/build_tarballs.jl
```diff
@@ -0,0 +1,54 @@
...
+cp tls/gnutls/libgiognutls.so ${libdir}/gio/modules/
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; exclude=Sys.iswindows)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libgioenvironmentproxy", :libgioenvironmentproxy, "lib/gio/modules"),
```
giordano, line 41: [Plugins](https://eli.thegreenplace.net/2012/08/24/plugins-in-c) aren't `LibraryProduct`s.  At most, you can have it as a `FileProduct`.

lukebemish:
If it is helpful at all, I have the following build script working locally on every platform (and allowing me to build libsoup everywhere too): [build_tarballs.jl](https://gist.github.com/lukebemish/2feff1f9617382d69d218d879def9dee)

t-bltg:
@lucifer1004, do you plan on updating this PR ?
I not, then maybe @lukebemish can open a new PR.

lucifer1004:
@t-bltg @lukebemish Feel free to open a new PR. I am not going to work on this for now.

ViralBShah:
Implemented in #6530

PR #5691, opened 2022-10-16
[KaHyPar] update build script for M1

jalving:
KaHyPar v1.3.0 supports building for the M1 architecture. This should be a simple update but please review if other changes are needed.

### DIFF
--- a/K/KaHyPar/build_tarballs.jl
+++ b/K/KaHyPar/build_tarballs.jl
@@ -13,14 +13,16 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir
-for f in ${WORKSPACE}/srcdir/patches/*.patch; do
-    atomic_patch -p1 ${f}
-done
-cd kahypar/
+cd $WORKSPACE/srcdir/kahypar/
+# Never build with `-march=native`
+atomic_patch -p1 ../patches/no_native.patch
 git submodule update --init --recursive --depth=1
 mkdir build && cd build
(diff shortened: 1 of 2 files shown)


giordano:
The [patch](https://github.com/JuliaPackaging/Yggdrasil/blob/7bc6fa71b4d569fd706a896f2d95b2069a405835/K/KaHyPar/bundled/patches/no_native.patch) has to be refreshed.

jalving:
@giordano thanks, I did not realize we were using the patching workflow. I haven't done this before, but I assume I need to use the binary builder wizard and build a new patch like the [docs](https://docs.binarybuilder.org/stable/build_tips/#Editing-files-in-the-wizard) say?

Alternatively, would you be able to help out here again @Roger-luo ?

giordano:
I refreshed the patch for you.  The [documentation also says](https://docs.binarybuilder.org/stable/#Manually-create-or-edit-build_tarballs.jl) that if you're updating an existing recipe, it's ***way*** easier to launch the build script, instead of starting with the wizard from scratch.  Which is what I did: I ran `julia build_tarballs.jl --debug --verbose host`, got into the build environment when I got the error with applying the patch, I fixed the cmake file to never run `--march=native --mtune=native` (we want to ship the library for different systems, we don't want to make it unusable on machines with less capable ISA than the build machine, and that won't work anyway when cross-compiling for different architectures), ran `git diff` and saved the result in the patch file.

Anyway, thanks for your contribution!

jalving:
Thanks @giordano! your instructions were very educational for me. I'll manage the next time this comes up :)

PR #5697, opened 2022-10-17
bump `Python` to `3.10.8` - enable-optimizations

t-bltg:
- bump python to `3.10.8`;
- add `--enable-optimizations` configure flag;
- ~~add `version.major` & `version.minor` to `LibraryProduct`.~~

### DIFF
--- a/P/Python/build_tarballs.jl
+++ b/P/Python/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "Python"
-version = v"3.10.7"
+version = v"3.10.8"
 
 # Collection of sources required to build Python
 sources = [
     ArchiveSource("https://www.python.org/ftp/python/$(version)/$(name)-$(version).tar.xz",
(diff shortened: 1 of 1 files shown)


P/Python/build_tarballs.jl
```diff
@@ -91,7 +96,7 @@ filter!(!Sys.iswindows, platforms)
 # The products that we will ensure are always built
 products = Product[
     ExecutableProduct(["python", "python3"], :python),
-    LibraryProduct(["libpython3", "libpython3.8"], :libpython),
+    LibraryProduct(["libpython3", "libpython$(version.major).$(version.minor)"], :libpython),
```
giordano, line 99: Is the second option actually necessary since it was wrong?

t-bltg: It seems no, `libpython3` matches `libpython3.X`, so I guess we could also just ditch `, "libpython3.8"`.

giordano: Yes, I'd prefer to simplify to what's actually needed, so
```julia
    ExecutableProduct("python3", :python),
    LibraryProduct("libpython3", :libpython),
```

PR #5700, opened 2022-10-18
LibPQ: expose headers to client applications

maxfreu:
Hi @giordano! This PR exposes the postgres header files to client applications, that can build on top of them. This is for example needed in gdal. Maybe the contents of `$prefix/include/postgresql/server` are unused and can be deleted to save space?


giordano:
Thanks!

PR #5703, opened 2022-10-18
[GDAL] add PostgreSQL support

maxfreu:
@visr This is the follow-up to #5700 and adds PostgreSQL support to gdal. Should the version offset be changed? If yes, how?


visr:
Yes you can bump the patch version of the version_offset to 3.

I see there are still errors on Windows:

> Could NOT find PostgreSQL (missing: PostgreSQL_LIBRARY) (found version "14.3")

Could you set that? Similar to 30b894ed8dfeb161bb79bc2c8aed59b2ce4cd3ca.

Hope this works! Will be nice to have.

maxfreu:
Tadaa, it worked! :)

maxfreu:
@giordano thank you for taking care of the repo and the quick merges!

giordano:
Thanks again for your contribution!

PR #5710, opened 2022-10-19
new recipe: extrae

exaexa:
Extrae is a specific profiling/instrumentation library developed for analyzing the performance of highly parallel programs at BSC.es. It works nicely with MPI and other HPC technologies. Our aim is to enable export of (some) extrae events from Julia to enable integration with larger workflows that use the tools from the ecosystem to measure/debug performance.

See https://tools.bsc.es/extrae and https://tools.bsc.es/paraver for illustrations.

~~Regarding this PR, I had a some trouble with the Extrae configure script being (completely un)able to find the BinUtils_jll contents, so I decided to just ignore it for now. Managing that would open some more instrumentation options, but both of them are unlikely to be relevant for Julia stuff.~~ I had to disable most of the platforms/packages because extrae fails for various reasons, most suspiciously by 1) being unable to use unwind 2) configure script failing to produce executable binaries. Opinions on that very welcome, esp. if I can help it with more cross-compiling options.

UPDATE: I patched the extrae autoconf machinery to be able to cross-compile and it seems to work. Opinions still welcome. I didn't succeed on making it work with MPI though.

(Cc @clasqui @ArnauMontagud @laurentheirendt)

### DIFF
--- a/E/extrae/build_tarballs.jl
+++ b/E/extrae/build_tarballs.jl
@@ -31,10 +31,10 @@ make -j${nproc}
 make install
 """
 
-platforms = expand_cxxstring_abis([
+platforms = [
     Platform("i686", "Linux"),
     Platform("x86_64", "Linux"),
-])
+]
 
 products = [
     LibraryProduct("libseqtrace", :libseqtrace),
--- a/E/extrae/build_tarballs.jl
+++ b/E/extrae/build_tarballs.jl
@@ -7,13 +7,20 @@ sources = [
     ArchiveSource(
         "https://github.com/bsc-performance-tools/extrae/archive/refs/tags/$(version).tar.gz",
         "e6765eb087be3f3c162e08d65f425de5f26912811392527d56ecd75d4fb6b99d"),
+        DirectorySource("./bundled"),
 ]
 
 script = raw"""
 cd ${WORKSPACE}/srcdir/extrae-*
+
+atomic_patch -p1 ../patches/0001-autoconf-replace-pointer-size-check-by-AC_CHECK_SIZE.patch
+atomic_patch -p1 ../patches/0002-autoconf-use-simpler-endianiness-check.patch
+
 autoreconf -fvi
 ./configure \
-    --prefix=$prefix \
+    --prefix=${prefix} \
+    --build=${MACHTYPE} \
+    --host=${target} \
     --without-binutils \
     --without-dyninst \
     --without-mpi \
--- a/E/extrae/build_tarballs.jl
+++ b/E/extrae/build_tarballs.jl
@@ -21,6 +21,7 @@ autoreconf -fvi
     --prefix=${prefix} \
     --build=${MACHTYPE} \
     --host=${target} \
+    --disable-openmp \
     --without-binutils \
     --without-dyninst \
     --without-mpi \
@@ -38,7 +39,6 @@ platforms = [
 
 products = [
     LibraryProduct("libseqtrace", :libseqtrace),
-    LibraryProduct("libomptrace", :libomptrace),
     LibraryProduct("libnanostrace", :libnanostrace),
 ]
 
--- a/E/extrae/build_tarballs.jl
+++ b/E/extrae/build_tarballs.jl
@@ -1,0 +1,42 @@
+#!/usr/bin/env julia
+using BinaryBuilder, Pkg
+
+name = "extrae"
+version = v"4.0.1"
+sources = [
+    ArchiveSource(
+        "https://github.com/bsc-performance-tools/extrae/archive/refs/tags/$(version).tar.gz",
+        "e6765eb087be3f3c162e08d65f425de5f26912811392527d56ecd75d4fb6b99d"),
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/extrae-*
+autoreconf -fvi
+./configure \
+    --prefix=$prefix \
+    --without-binutils \
+    --without-dyninst \
+    --without-mpi \
+    --without-papi \
+    --without-unwind
+
+make -j${nproc}
+make install
+"""
+
+platforms = expand_cxxstring_abis([
+    Platform("i686", "Linux"),
+    Platform("x86_64", "Linux"),
+])
+
+products = [
+    LibraryProduct("libseqtrace", :libseqtrace),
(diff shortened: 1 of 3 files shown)


E/extrae/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+])
+
+products = [
+    LibraryProduct("libseqtrace", :libseqtrace),
+    LibraryProduct("libomptrace", :libomptrace),
+    LibraryProduct("libnanostrace", :libnanostrace),
+]
+
+dependencies = [
+    Dependency("OpenMPI_jll"),
```
giordano, line 39: I think it'd be nicer to target more MPI implementations, you can look at https://github.com/JuliaPackaging/Yggdrasil/blob/6a875bb6bdfbd8c299118fec79f1ce0b85cc6b63/A/ADIOS2/build_tarballs.jl#L81-L97 for inspiration

E/extrae/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+sources = [
+    ArchiveSource(
+        "https://github.com/bsc-performance-tools/extrae/archive/refs/tags/$(version).tar.gz",
+        "e6765eb087be3f3c162e08d65f425de5f26912811392527d56ecd75d4fb6b99d"),
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/extrae-*
+autoreconf -fvi
+./configure \
```
giordano, line 15: https://docs.binarybuilder.org/stable/build_tips/#Autoconfigure-builds
```suggestion
./configure \
    --build=${MACHTYPE} \
    --host=${target} \
```

exaexa: Ah nice thanks. I'll retry later today

E/extrae/build_tarballs.jl
exaexa: Thanks!

E/extrae/build_tarballs.jl
giordano: This is still necessary

exaexa: ah sorry no idea why I dropped it. :confounded:

exaexa: ah yes, I dropped it because the configure script can't manage that:
```
[10:01:33] configure: error: in `/workspace/srcdir/extrae-4.0.1':
[10:01:33] configure: error: cannot run test program while cross compiling
[10:01:33] See `config.log' for more details
```
...even without actually cross-compiling.

What is the best solution? Not sure if patching it is viable here; I'll leave a note to the issue in extrae there and update the PR once they fix it?

exaexa:
OK nvm, this won't be mergeable until the issue is fixed upstream. Sorry for noise. :]

exaexa:
Reopening since we got this working with some patchage

E/extrae/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
...
+    --without-unwind
+
+make -j${nproc}
+make install
+"""
+
+platforms = expand_cxxstring_abis([
+    Platform("i686", "Linux"),
+    Platform("x86_64", "Linux"),
+])
```
giordano, line 37: As far as I can tell you don't need to expand the C++ string ABI:
```suggestion
platforms = [
    Platform("i686", "Linux"),
    Platform("x86_64", "Linux"),
]
```

giordano:
Good to go?

exaexa:
Thanks for asking, actually let me test this thoroughly with all the settings disabled. It might have got broken.

mofeing:
@exaexa I've requested a PR to update the Binutils recipe #5957 so by adding `Binutils_jll` and `LibUnwind_jll` as dependencies, Extrae would support source code information.

I would like to suggest adding `PAPI_jll` dependency for hardware counters support and maybe `OpenMPI_jll` for MPI support. Also, it would be nice to reenable the OpenMP support as even if Julia does not use OpenMP, underlying libraries might. One example is OpenBLAS. In general, it would be better to enable all features here since you can enable or disable them at runtime with the configuration file in XML.

As a sidenote, I recommend changing the name to **E**xtrae, with the first letter in capital case, since that is the way it is written several times in the official web page.

E/extrae/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env julia
+using BinaryBuilder, Pkg
```
mofeing, line 2: `Pkg` library not needed
```suggestion
using BinaryBuilder
```

giordano:
> so by adding Binutils_jll and LibUnwind_jll as dependencies, Extrae would support source code information.

Our toolchain already comes with binutils, that `Binutils_jll` is not used for anything, reason why the recipe hadn't been touched in ages.

mofeing:
> Our toolchain already comes with binutils, that `Binutils_jll` is not used for anything, reason why the recipe hadn't been touched in ages.

Okay, yes and no. Extrae needs `libbfd` and `libiberty`. I have just checked out that the pipeline incorporates `libbfd` and `libiberty` as static libraries for the BUILD environment.

Example, inside the pipeline image targetting `powerpc64le-linux-gnu`:
```bash
sandbox:${WORKSPACE}/srcdir # find / -name "libbfd*"
/opt/x86_64-linux-musl/lib/libbfd.a
/opt/x86_64-linux-musl/lib/libbfd.la
sandbox:${WORKSPACE}/srcdir # find / -name "libiberty*"
/opt/x86_64-linux-musl/lib/gcc/x86_64-linux-musl/4.8.5/plugin/include/libiberty.h
/opt/x86_64-linux-musl/lib/libiberty.a
/opt/powerpc64le-linux-gnu/lib/gcc/powerpc64le-linux-gnu/4.8.5/plugin/include/libiberty.h
/opt/powerpc64le-linux-gnu/lib/libiberty.a
```

So when building, the linker would find `libiberty` for `powerpc64le-linux-gnu` but not `libbfd`.

mofeing:
Also @exaexa the `libnanosstrace` is really needed? I would say it doesn't apply to Julia. It is thought to be trace the Nanos runtime developed at BSC, which is the parallel runtime of the OMPSs programming model (not OpenMP).

giordano:
Closing in favour of #5991

PR #5713, opened 2022-10-20
[HiGHS] update to v1.3.0

odow:
<s>Please hold off merging just yet. I'm hopeful that HiGHS will make an official v1.3.0 release in the next few days, but I want check that things still compile, and this v101.200.200 is a good fallback if v1.3.0 isn't released.</s> Updated to v1.3.0

cc @ccoffrin

### DIFF
--- a/H/HiGHS/HiGHS/build_tarballs.jl
+++ b/H/HiGHS/HiGHS/build_tarballs.jl
@@ -1,6 +1,6 @@
 # To ensure a build, it isn't sufficient to modify highs_common.jl.
 # You also need to update a line in this file:
-#     Last updated: 2022-10-20
+#     Last updated: 2022-10-24
 
 include("../highs_common.jl")
 
--- a/H/HiGHS/HiGHS/build_tarballs.jl
+++ b/H/HiGHS/HiGHS/build_tarballs.jl
@@ -1,6 +1,6 @@
 # To ensure a build, it isn't sufficient to modify highs_common.jl.
 # You also need to update a line in this file:
-#     Last updated: 2022-04-22
+#     Last updated: 2022-10-20
 
 include("../highs_common.jl")
 
--- a/H/HiGHS/bundled/patches/v1.3.0.patch
+++ b/H/HiGHS/bundled/patches/v1.3.0.patch
@@ -1,27 +1,68 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
-index e5d3a14b8..8bd2dac90 100644
+index e5d3a14b8..2112a7e54 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -125,8 +125,8 @@ enable_cxx_compiler_flag_if_supported("-Wno-format-truncation")
+@@ -125,9 +125,16 @@ enable_cxx_compiler_flag_if_supported("-Wno-format-truncation")
  enable_cxx_compiler_flag_if_supported("-pedantic")
  endif()
  
 -check_cxx_compiler_flag("-mpopcnt" FLAG_MPOPCNT_SUPPORTED)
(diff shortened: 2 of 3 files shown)


H/HiGHS/bundled/patches/v1.3.0.patch
```diff
@@ -0,0 +1,27 @@
...
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -125,8 +125,8 @@ enable_cxx_compiler_flag_if_supported("-Wno-format-truncation")
+ enable_cxx_compiler_flag_if_supported("-pedantic")
+ endif()
+ 
+-check_cxx_compiler_flag("-mpopcnt" FLAG_MPOPCNT_SUPPORTED)
+-set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -mpopcnt")
++# check_cxx_compiler_flag("-mpopcnt" FLAG_MPOPCNT_SUPPORTED)
++# set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -mpopcnt")
```
odow, line 12: @lgottwald why is `mpopcnt` needed here? It isn't portable across platforms.

giordano: If anything, you have to do something like https://github.com/rdkit/rdkit/pull/5182/files (check [`CMAKE_SYSTEM_PROCESSOR`](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html))

H/HiGHS/bundled/patches/v1.3.0.patch
```diff
@@ -0,0 +1,68 @@
...
+@@ -125,9 +125,16 @@ enable_cxx_compiler_flag_if_supported("-Wno-format-truncation")
+ enable_cxx_compiler_flag_if_supported("-pedantic")
+ endif()
+ 
+-check_cxx_compiler_flag("-mpopcnt" FLAG_MPOPCNT_SUPPORTED)
+-set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -mpopcnt")
+-
++if(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
++  message("FLAG_MPOPCNT_SUPPORTED is not available on aarch64")
++  set(FLAG_MPOPCNT_SUPPORTED OFF)
```
giordano, line 14: It may actually be better to match x86_64/i686 and powerpc, and do nothing for all the rest, whatever it is (instead of matching all other possible architectures, which are unbound)

giordano: Also, it'd be great to upstream this patch :slightly_smiling_face:

odow: Yeah once we get something working, I'll send it to @jajhall and he can release v1.3.1

H/HiGHS/HiGHS/build_tarballs.jl
```diff
@@ -1,6 +1,6 @@
 # To ensure a build, it isn't sufficient to modify highs_common.jl.
 # You also need to update a line in this file:
-#     Last updated: 2022-04-22
+#     Last updated: 2022-10-20
```
giordano, line 3: Let's force a rebuild for all platforms
```suggestion
#     Last updated: 2022-10-24
```

giordano (commented):
Is it good to go?

odow:
> Is it good to go?

If it compiles, I assume so 😆

jajhall:
OK, I'll create v1.3.1 from today's minor changes

tengels:
Note sure this is the correct place; but would it be possible to bump [HiGHSstatic](https://github.com/JuliaBinaryWrappers/HiGHSstatic_jll.jl) to 1.3 as well?

Thanks in advance!

PR #5714, opened 2022-10-20
Add casacorecxx build script

torrance:
This is a CxxWrapper around Casacore, which is already packaged as part of Yggdrasil.

It is intended to be used primarily by https://github.com/torrance/Casacore.jl

### DIFF
--- a/C/casacorecxx/build_tarballs.jl
+++ b/C/casacorecxx/build_tarballs.jl
@@ -1,0 +1,42 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "casacorecxx"
+version = v"0.1.0"
+
+# Collection of sources required to complete build
+sources = [GitSource("https://github.com/torrance/casacorecxx.git", "1b979164db214721c1c3cf52388b0ef9ed0c5d9b")]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/casacorecxx
+mkdir build && cd build
+cmake \
+    -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_FIND_ROOT_PATH=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DJulia_PREFIX=${prefix}\
+    ..
+VERBOSE=ON cmake --build . --config Release --target install -- -j${nproc}
+exit
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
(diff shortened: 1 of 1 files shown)


C/casacorecxx/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+
+# Dependencies that must be installed before this package can be built
+dependencies = [Dependency("libcxxwrap_julia_jll"),
+                Dependency("casacore_jll"),
+                BuildDependency("libjulia_jll")]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="1.8",
+               preferred_gcc_version=v"12") # We need C++17 for CxxWrap
```
giordano, line 42: For maximum compatibility we want to use the ***oldest*** version of GCC possible, not the newest one.  C++17 is [already available in GCC5](https://gcc.gnu.org/projects/cxx-status.html#cxx17).

C/casacorecxx/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DJulia_PREFIX=${prefix}\
+    ..
+VERBOSE=ON cmake --build . --config Release --target install -- -j${nproc}
+exit
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [Platform("x86_64", "linux"; libc="glibc", julia_version="1.8")]
```
giordano, line 28: Please see https://github.com/JuliaPackaging/Yggdrasil/blob/52ff75da5b7eb3bc55b0fe632deaa85a8f937828/L/libsingular_julia/build_tarballs.jl#L15-L16 https://github.com/JuliaPackaging/Yggdrasil/blob/52ff75da5b7eb3bc55b0fe632deaa85a8f937828/L/libsingular_julia/build_tarballs.jl#L40-L45

giordano (commented):
Thanks for your contribution!  CC @kiranshila

torrance:
@giordano Thanks for your comments.

In the latest commit I've increased the supported platforms to the union of those supported by libjulia and casacore, for Julia version 1.6-1.9.

I've also decreased the preferred GCC version to 7, which is the minimum version I am able to successfully build with.

As an aside: is it possible to set up build_tarball.jl to build from a Git subdirectory? Both Casacore.jl and casacorecxx are developed in lockstep. Currently casacorecxx is a subdirectory in https://github.com/torrance/Casacore.jl, and it's not ideal to have to spin out casacorecxx into its own Git repo for the sake of packaging.

giordano:
> is it possible to set up build_tarball.jl to build from a Git subdirectory?

Besides cloning the repository and entering whatever directory you want?

torrance:
@giordano Right. How embarrassing. :)

I've set the GitSource to the active repo. This should now be ready to merge.

PR #5726, opened 2022-10-21
ARPACK 3.7

ViralBShah:
@giordano Updating Arpack 3.7 in here with a modern build. I think this is ok, since it has been yanked for a long time, and nobody is using it.


giordano:
Yes, it should be fine after https://github.com/JuliaRegistries/General/pull/70767 is merged.

ViralBShah:
Do these PRs need to be merged in any specific order?

PR #5730, opened 2022-10-23
[CompilerSupportLibraries] Add `libmsvcrt` to the Windows package

giordano:
@vchuravy requested to get libmsvcrt in for linking purposes in https://github.com/JuliaLang/julia/pull/47184.  Note: since this is using libraries in GCC 12 it requires https://github.com/JuliaLang/julia/pull/45582.  I'm not sure what we want to do here: downgrade GCC to v11?  Getting also https://github.com/JuliaLang/julia/pull/45582 solved would be lovely....


vchuravy:
Since I need it for 1.9 getting it from GCC11 would be lovely.

Generally the strategy we have for other libraries to keep the build script around for anything used by the LTS and currently supported releases of Julia.

vchuravy:
I would also need libgcc.a and it would be nice to have `libssp.a` instead of `libssp-0.dll` but that last one is Ästhetik.

giordano:
@vchuravy hows does it look now?  See for example https://julia-bb-buildcache.s3.amazonaws.com/065707d8382f3ea180b2dfe63948cb608aef6980c2baea8192a31216f4b109b0/5d482154118c398facdd7d9df7a2c2c6e3218d7dfde74506b29befaf3b53eff8/x86_64-w64-mingw32-libgfortran5.tar.gz

vchuravy:
I think this looks great! Just notices that `libgcc_s.a` would also be nice ;) 


There is a lingering question in  https://github.com/JuliaLang/julia/pull/47184 if I need to link in:
`dllcrt2.o`, `crtbegin.o` and `crtend.o`, but those also necessitate some of the mingw libraries...
So far I haven't needed them and it might simply mean that we won't support initializers, but if I link those in I wouldn't need to do:

https://github.com/JuliaLang/julia/pull/47184/files#diff-dcb7645cc96941c621f6aeaaebbdcdd54e83fe925eed0135e4fd840806fb4d91R661-R673

vchuravy (approved):

PR #5734, opened 2022-10-24
Add cereal_jll package (dependency of mlpack).

rcurtin:
Just earlier today, mlpack 4.0.0 was released, so I have begun the process of updating the `mlpack_jll` package.  However, mlpack 4 has a new dependency on cereal, a header-only serialization library (this then let mlpack remove its Boost dependency).  So, we just need to add a `cereal_jll` package to make those headers available for mlpack.

I ran and tested this locally---seems to be fine.  We'll see if CI agrees. :)

### DIFF
--- a/C/cereal/build_tarballs.jl
+++ b/C/cereal/build_tarballs.jl
@@ -14,7 +14,8 @@ sources = [
 
 script = raw"""
 cd ${WORKSPACE}/srcdir/cereal-*/
-cp -vr include/* ${prefix}/include/
+mkdir "${includedir}"
+cp -vr include/* "${includedir}/."
 """
 
 # These are the platforms we will build for by default, unless further
--- a/C/cereal/build_tarballs.jl
+++ b/C/cereal/build_tarballs.jl
@@ -14,7 +14,7 @@ sources = [
 
 script = raw"""
 cd ${WORKSPACE}/srcdir/cereal-*/
-mkdir "${includedir}"
+mkdir -p "${includedir}"
 cp -vr include/* "${includedir}/."
 """
 
--- a/C/cereal/build_tarballs.jl
+++ b/C/cereal/build_tarballs.jl
@@ -1,0 +1,30 @@
+# Build script for cereal: C++ header-only library for serialization.
+# This is basically just a dummy package that installs header files, and makes
+# them available for other packages.
+
+using BinaryBuilder
+
+
+name = "cereal"
+version = v"1.3.2"
+sources = [
+    ArchiveSource("https://github.com/USCILab/cereal/archive/refs/tags/v$(version).tar.gz",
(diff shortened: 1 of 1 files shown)


C/cereal/build_tarballs.jl
```diff
@@ -0,0 +1,30 @@
...
+name = "cereal"
+version = v"1.3.2"
+sources = [
+    ArchiveSource("https://github.com/USCILab/cereal/archive/refs/tags/v$(version).tar.gz",
+                  "16a7ad9b31ba5880dac55d62b5d6f243c3ebc8d46a3514149e56b5e7ea81f85f")
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/cereal-*/
+cp -vr include/* ${prefix}/include/
```
giordano, line 17: I think you may have to create the directory first:
```suggestion
mkdir "${includedir}"
cp -vr include/* "${includedir}/."
```

giordano: Oh, it actually worked already, but I'm confused by how: I don't think `${includedir}` already exists :thinking:

rcurtin: Hmm, do you think we should change it to use `${prefix}/include` then?  Happy to push that if you think it's the right thing to do.

rcurtin:
Thanks @giordano!  Your incredibly quick reply time always stuns me. :)

C/cereal/build_tarballs.jl
giordano: Oh, I see it now: there is a single directory in [`include/`](https://github.com/USCiLab/cereal/tree/ddd467244713ea4fe63733628992efcdd6a9187d/include), and so
```
cp -vr include/* ${prefix}/include/
```
is expanded to
```
cp -vr include/cereal ${prefix}/include/
```
which would work in the sense that the shell won't complain (I'm pretty sure it would if there were more than two arguments) but it was dropping the `cereal` directory inside `${includedir}` as you can see in https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=23186&view=logs&j=21cafe84-7420-5c7b-853f-f088eadd7b23&t=549b827d-5559-5bb6-76a3-17f515936fbd&l=348.

Instead, with my suggested change (and yes, I forgot the `-p` flag to `mkdir`, thanks) we get the correct mapping to `${includedir}/cereal`: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=23188&view=logs&j=21cafe84-7420-5c7b-853f-f088eadd7b23&t=549b827d-5559-5bb6-76a3-17f515936fbd&l=348

giordano: >  (I had thought you meant that the variable `${includedir}` was unset.)

No no, sorry, I was referring to the directory on disk pointed to by `${includedir}`/`${prefix}/include` (both work and are always defined, but I prefer the former as it's more specific)

PR #5735, opened 2022-10-24
Update mlpack to version 4.0.0

rcurtin:
Here's the follow-up to #5734; it updates mlpack to the latest version.  A couple changes here are relevant:

 * No more dependency on Boost (hooray!).  That should help reduce build time.
 * New dependency on `cereal_jll` (see #5734).  I suspect that since #5734 was just merged like two minutes ago that this build may take until tomorrow to succeed?  Not 100% sure.
 * Bundled patch no longer necessary because it's been merged into upstream.

CC: @quinnj @kpamnany --- this should hopefully fix the mlpack issues that you had been encountering in the previous months.  This PR, and then a release of the `mlpack.jl` package (which will come just after this) should hopefully make an upgrade to mlpack easy (and if not just let me know).


quinnj:
❤️

M/mlpack/build_tarballs.jl
```diff
@@ -6,22 +6,21 @@ using BinaryBuilder
 
 # Set sources and other environment variables.
 name = "mlpack"
-version = v"3.4.3"
-source_version = v"3.4.2"
+version = v"4.0.0"
+source_version = v"4.0.0"
```
giordano, line 10: I'd do
```suggestion
source_version = v"4.0.0"
version = source_version
```
so that in the future we can update a single line, if we don't need to diverge from the upstream version number.

giordano:
```
[21:29:31] -- cotire 1.7.10 loaded.
[21:29:31] CMake Error at board/flags-config.cmake:19 (string):
[21:29:31]   string no output variable specified
[21:29:31] Call Stack (most recent call first):
[21:29:31]   CMake/ConfigureCrossCompile.cmake:11 (include)
[21:29:31]   CMakeLists.txt:7 (include)
[21:29:31] 
[21:29:31] 
[21:29:31] CMake Error at CMake/ConfigureCrossCompile.cmake:17 (message):
[21:29:31]   Cannot configure: TOOLCHAIN_PREFIX must be set when performing
[21:29:31]   cross-compiling!
[21:29:31] Call Stack (most recent call first):
[21:29:31]   CMakeLists.txt:7 (include)
[21:29:31] 
[21:29:31] 
[21:29:31] -- Configuring incomplete, errors occurred!
```
well, this is a new one, I've never seen this error message :flushed:

rcurtin:
> ```
> [21:29:31] -- cotire 1.7.10 loaded.
> [21:29:31] CMake Error at board/flags-config.cmake:19 (string):
> [21:29:31]   string no output variable specified
> [21:29:31] Call Stack (most recent call first):
> [21:29:31]   CMake/ConfigureCrossCompile.cmake:11 (include)
> [21:29:31]   CMakeLists.txt:7 (include)
> [21:29:31] 
> [21:29:31] 
> [21:29:31] CMake Error at CMake/ConfigureCrossCompile.cmake:17 (message):
> [21:29:31]   Cannot configure: TOOLCHAIN_PREFIX must be set when performing
> [21:29:31]   cross-compiling!
> [21:29:31] Call Stack (most recent call first):
> [21:29:31]   CMakeLists.txt:7 (include)
> [21:29:31] 
> [21:29:31] 
> [21:29:31] -- Configuring incomplete, errors occurred!
> ```
> 
> well, this is a new one, I've never seen this error message flushed

Yeah, it is specific to mlpack, fortunately.  mlpack now has support for cross-compilation, and it enables this when `CMAKE_CROSSCOMPILING` is set.  I think that just setting `CMAKE_TOOLCHAIN_FILE` is sufficient to set the variable `CMAKE_CROSSCOMPILING`, so I just need to work around that.  It may take a couple attempts, but I'll have it fixed shortly.

M/mlpack/build_tarballs.jl
```diff
@@ -185,9 +178,9 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("boost_jll"; compat="=1.76.0"),
     Dependency("armadillo_jll"),
-    Dependency("OpenBLAS_jll", v"0.3.13")
+    Dependency("OpenBLAS_jll", v"0.3.13"),
+    Dependency("cereal_jll")
```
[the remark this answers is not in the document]
rcurtin: Thanks, handled in d23e30f. :+1:

fingolfin:
It seems some of these jobs run get stuck and keep running? A ton of them are now at 30 minutes and growing, and they all seem to be stuck on the same line.

rcurtin:
mlpack is pretty computationally intensive to build.  Just glancing at the logs, it looks like it's building the individual Julia bindings just fine (just, it takes a while).  (Hopefully that's not an issue.)

rcurtin:
Yeah, it seems like some of the build setups are either way more powerful than others, or use way more cores than others.  But as far as I can tell, even the short (~2m) jobs did build everything successfully, which is quite impressive!

giordano:
It's all running on a single machine, but maybe those platforms had previous successful runs and so now they have hot ccache?

rcurtin:
Okay, as far as I can tell, everything is good to go here, and all the libraries are being built successfully. :crossed_fingers:

M/mlpack/build_tarballs.jl
```diff
@@ -6,15 +6,24 @@ using BinaryBuilder
...
+source_version = v"4.0.0"
+version = source_version
 sources = [
     ArchiveSource("https://www.mlpack.org/files/mlpack-$(source_version).tar.gz",
-                  "9e5c4af5c276c86a0dcc553289f6fe7b1b340d61c1e59844b53da0debedbb171"),
-    DirectorySource("./bundled"),
+                  "041d9eee96445667d2f7b970d2a799592027f1f8818cd96a65dcce1ac0745773"),
+    # STB headers for image support.
+    ArchiveSource("https://www.mlpack.org/files/stb.tar.gz",
+                  "5c01115d6909621af63a502437334f2014065a5ce6bb7ac05acbf15f24acce15"),
```
giordano, line 16: Can this be a separate header-only package like cereal?  Also, is this tarball unversioned?  If it'll be replaced in the future this would make this recipe unreproducible.

M/mlpack/build_tarballs.jl
```diff
@@ -185,9 +188,10 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("boost_jll"; compat="=1.76.0"),
     Dependency("armadillo_jll"),
-    Dependency("OpenBLAS_jll", v"0.3.13")
+    Dependency("OpenBLAS_jll", v"0.3.13"),
```
giordano, line 192: From the audit https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=23225&view=logs&j=f5128b8a-f209-5fa4-85df-c01db7529e06&t=d51dec25-fb1c-5234-c33e-37032ff96d86&l=15729:
```
┌ Warning: To ensure that the correct version of libgomp is found at runtime, add the following entry to the list of dependencies of this builder
│ 
└ @ BinaryBuilder.Auditor /depot/packages/BinaryBuilder/xmfIp/src/auditor/compiler_abi.jl:4
```
```suggestion
    Dependency("OpenBLAS_jll", v"0.3.13"),
    Dependency("CompilerSupportLibraries_jll"),
```

giordano (commented):
Also [from the audit](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=23225&view=logs&j=f5128b8a-f209-5fa4-85df-c01db7529e06&t=d51dec25-fb1c-5234-c33e-37032ff96d86&l=16483), the license is no longer automatically installed, so you need to do it manually with
```
install_license /path/to/license/file
```
See https://docs.binarybuilder.org/stable/build_tips/#Installing-the-license-file

M/mlpack/build_tarballs.jl
rcurtin: Yeah, good point.  STB is only two header files so I thought I could keep it simple, but let me just open another header-only package for it---it'll be quite trivial.

rcurtin:
I didn't know about the audit output, thanks for pointing that out!  Are the audit issues collected somewhere, or would I have to parse each build log separately to find these things?

giordano:
In the main view of the log of the build: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=23225&view=results

M/mlpack/build_tarballs.jl
rcurtin: Now handled via #5741.

rcurtin:
Looks like a random segfault on the musl-cxx11 build; I wonder if a restart would fix it.  (It had previously built just fine.)

giordano:
This is weird, `aarch64-linux-musl-cxx11` is now crashing with an ICE:
```
[21:56:48] In file included from /workspace/destdir/include/armadillo:584:0,
[21:56:48]                  from /workspace/srcdir/mlpack-4.0.0/src/mlpack/core/arma_extend/arma_extend.hpp:21,
[21:56:48]                  from /workspace/srcdir/mlpack-4.0.0/src/mlpack/base.hpp:87,
[21:56:48]                  from /workspace/srcdir/mlpack-4.0.0/src/mlpack/core/util/prefixedoutstream.hpp:16,
[21:56:48]                  from /workspace/srcdir/mlpack-4.0.0/src/mlpack/core/util/log.hpp:17,
[21:56:48]                  from /workspace/srcdir/mlpack-4.0.0/build/src/mlpack/bindings/julia/mlpack/build/generate_jl_lmnn.cpp:13:
[21:56:48] /workspace/destdir/include/armadillo_bits/mul_gemv.hpp: In static member function ‘static void arma::gemv_emul_tinysq<do_trans_A, use_alpha, use_beta>::apply(eT*, const TA&, const eT*, eT, eT) [with eT = double; TA = arma::Col<double>; bool do_trans_A = false; bool use_alpha = true; bool use_beta = true]’:
[21:56:48] /workspace/destdir/include/armadillo_bits/mul_gemv.hpp:137:5: internal compiler error: Segmentation fault
[21:56:48]      }
[21:56:48]      ^
[21:56:48] 0xaa37e6 crash_signal
[21:56:48] 	/workspace/srcdir/gcc-6.1.0/gcc/toplev.c:333
[21:56:48] 0x7f8d6d27f1de ???
[21:56:48] 	/workspace/srcdir/musl-1.2.2/src/signal/x86_64/restore.s:1
[21:56:48] Please submit a full bug report,
[21:56:48] with preprocessed source if appropriate.
[21:56:48] Please include the complete backtrace with any bug report.
[21:56:48] See <[http://gcc.gnu.org/bugs.html>](http://gcc.gnu.org/bugs.html%3E) for instructions.
[21:56:48] make[2]: *** [src/mlpack/methods/CMakeFiles/generate_jl_lmnn.dir/build.make:81: src/mlpack/methods/CMakeFiles/generate_jl_lmnn.dir/__/bindings/julia/mlpack/build/generate_jl_lmnn.cpp.o] Error 1
```
The weird thing is that it worked before, but with a different version of sbt, I presume the new one is triggering the compiler error?  There is usually little we can do in these cases apart from using a different compiler.

rcurtin:
Yeah, let's see what happens.  In the worst case I can go change around the version of stb to what was previously being used.  mlpack (and all the C++ template metaprogramming madness dependencies it has) tends to work compilers pretty hard.  As far as I understanding building mlpack is actually part of MSVC's test suite... anyway, it's known to be really mean to compilers.

M/mlpack/build_tarballs.jl
```diff
@@ -31,17 +32,15 @@ mkdir build && cd build
 # version.  So we'll just create a crappy little script, since Julia may not
 # be available in the build environment.
 echo "#!/bin/bash" > julia
-echo "echo \"Fake Julia version 1.3.0\"" >> julia
+echo "echo \"Fake Julia version 1.8.2\"" >> julia
 chmod +x julia
 
 FLAGS=(-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}
+       -DCMAKE_CROSSCOMPILING=OFF
```
giordano, line 39: Uhm, why this is needed?  We actually do cross-compilation (that's the entire magic of BinaryBuilder :sweat_smile:), so it's a bit weird that incorrectly setting `CMAKE_CROSSCOMPILING=OFF` makes it work (since this is false) :thinking:

rcurtin: mlpack added some amount of support specific to cross-compiling that's built on top of CMake's cross-compiling support.  I didn't expect how that interacted with BinaryBuilder here (that gave [this error](https://github.com/JuliaPackaging/Yggdrasil/pull/5735#issuecomment-1289645105) that you pointed out), so I tried setting `CMAKE_CROSSCOMPILING` to `OFF` so that mlpack's cross-compiling CMake infrastructure didn't kick in.  It appears to work fine.

If you want to be a little more paranoid about it, I can instead use a patch on the `CMakeLists.txt` that forces the mlpack cross-compilation infrastructure to never be used, just let me know.

giordano: That'd be nicer, but ok, let's go on with this one now since this is working.  Thanks!

rcurtin: Awesome---thanks for the merge!  Now I'll update the mlpack.jl package (once `mlpack_jll` is pushed), and then hopefully (:crossed_fingers:) everything will work, otherwise I will be back with some other minor fix PR :smile:

quinnj: Thanks for keeping us updated @rcurtin! This is awesome! Will you ping me (here or elsewhere) when mlpack.jl is updated? We'd love to try things out at RAI with the updates. On a perhaps related note: a few people have upgraded to MacOS Ventura and are reporting issues with armadillo_jll not being able to load correctly. Not sure if the changes here will improve that, but I'm also not quite sure where to file an issue.

rcurtin: @quinnj it is released now, so hopefully the update should work just fine.

I'm happy to take a quick look to try to help with any armadillo_jll issues.  However I agree, it's not clear precisely where to file an issue.  You could use https://github.com/mlpack/mlpack.jl, or, you could just send me an email at ryan@ratml.org.  The choice is yours. :)

quinnj: Great! I'll give it a whirl today.

For the armadillo_jll issue, I believe the BB wizards are discussing what is to be done in [this public Julia slack thread](https://julialang.slack.com/archives/C674ELDNX/p1667222083888139) if you're interested in following along; it sounds like Apple changed the linker behavior in Ventura.

rcurtin: Thanks, following along there.  I'll jump into the conversation if I can add anything useful, but it looks like a lot of the hard digging has already been done.

rcurtin: Don't hesitate to tag me if needed.

quinnj: 😞 
```julia
  From worker 4:    signal (11): Segmentation fault
      From worker 4:    in expression starting at none:0
      From worker 4:    _ZN6mlpack13HoeffdingTreeINS_12GiniImpurityENS_24BinaryDoubleNumericSplitENS_25HoeffdingCategoricalSplitEE5TrainIN4arma11subview_colIdEEEEvRKT_m at /tmp/nix-build-raicode-tests.unit2.drv-0/rai-server-source/.julia/artifacts/5061b33ce5baf49f2784e1ab327aa4d27534afe0/lib/libmlpack_julia_hoeffding_tree.so (unknown line)
      From worker 4:    unknown function (ip: (nil))
      From worker 4:    Allocations: 4686818551 (Pool: 4684775049; Big: 2043502); GC: 294
Worker 4 terminated.
```

rcurtin: I am unable to debug that without a way to reproduce.

quinnj: Yeah, I'll work on getting a smaller reproduction.

PR #5741, opened 2022-10-25
Add stb header-only package

rcurtin:
This is another dependency of mlpack (see #5735, [this discussion](https://github.com/JuliaPackaging/Yggdrasil/pull/5735#discussion_r1004863704)).  It is just some header-only files, from https://github.com/nothings/stb/.

### DIFF
--- a/S/stb/build_tarballs.jl
+++ b/S/stb/build_tarballs.jl
@@ -10,7 +10,7 @@ name = "stb"
 # revision...
 version = v"0.0.20221025"
 sources = [
-    ArchiveSource("https://github.com/nothings/stb/archive/8b5f1f37b5b75829fc72d38e7b5d4bcbf8a26d55.zip"
+    ArchiveSource("https://github.com/nothings/stb/archive/8b5f1f37b5b75829fc72d38e7b5d4bcbf8a26d55.zip",
                   "93a16ee3e866e719feec459f6f132cce932c5ec751eb38e3ec1975f911353d2e")
 ]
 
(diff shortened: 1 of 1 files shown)


S/stb/build_tarballs.jl
```diff
@@ -0,0 +1,33 @@
...
+# stb doesn't ever do releases, so the best we can do is stick to a specific git
+# revision...
+version = v"0.0.20221025"
+sources = [
+    ArchiveSource("https://github.com/nothings/stb/archive/8b5f1f37b5b75829fc72d38e7b5d4bcbf8a26d55.zip",
+                  "93a16ee3e866e719feec459f6f132cce932c5ec751eb38e3ec1975f911353d2e")
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/stb/
```
giordano, line 18:
```suggestion
cd ${WORKSPACE}/srcdir/stb*/
```

rcurtin: Thanks, you beat me to it. :)

S/stb/build_tarballs.jl
```diff
@@ -0,0 +1,33 @@
...
+version = v"0.0.20221025"
+sources = [
+    ArchiveSource("https://github.com/nothings/stb/archive/8b5f1f37b5b75829fc72d38e7b5d4bcbf8a26d55.zip",
+                  "93a16ee3e866e719feec459f6f132cce932c5ec751eb38e3ec1975f911353d2e")
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/stb*/
+mkdir -p "${includedir}"
+cp -vr include/* "${includedir}/."
```
giordano, line 20: I think it's
```suggestion
cp -vr *.h "${includedir}/."
```
?

giordano: Yes, it looks like the tarball https://www.mlpack.org/files/stb.tar.gz has a different tree structure, so here we need to do things differently.  Here there are way more files.

rcurtin: Thanks, you beat me to it again! :)

PR #5742, opened 2022-10-25
binutils: upgrade and set enable-new-dtags to use RUNPATH

vtjnash:
This should prevent issues with RPATH getting into some of our dependencies (openblas, libgit2, and suitesparse, in particular). Refs https://github.com/JuliaLang/julia/pull/46464

### DIFF
--- a/B/Binutils/build_tarballs.jl
+++ b/B/Binutils/build_tarballs.jl
@@ -58,7 +58,8 @@ else
         --host=\${MACHTYPE} \\
         --with-sysroot="\${prefix}/$(compiler_target)/sys-root" \\
         --enable-multilib \\
-        --disable-werror
+        --disable-werror \\
+	--enable-new-dtags
 
     make -j\${nproc}
     make install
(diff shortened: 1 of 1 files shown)


vtjnash:
staticfloat tells me this is actually maintained in a separate branch here, so this won't actually work to update the rootfs image with this: https://github.com/JuliaPackaging/Yggdrasil/blob/sf/gcc/B/Binutils/common.jl

giordano:
Did you mean to do it at https://github.com/JuliaPackaging/Yggdrasil/blob/c9f2c4021660210c5450824f22dd85130ebd4571/0_RootFS/gcc_common.jl#L225-L232? (I'm not looking forward to rebuilding 9 versions of GCC :see_no_evil:)

giordano:
Judging by https://github.com/JuliaBinaryWrappers/Binutils_jll.jl/releases/tag/Binutils-v2.38.0%2B4 I presume this was needed for building julia in CI rather than for BinaryBuilder?

staticfloat:
We're doing it for base Julia first.

staticfloat:
Closing this as it's not going to work immediately

PR #5743, opened 2022-10-25
Purge RPATH from libs installed with julia

apaz-cli:


B/Binutils/build_tarballs.jl
```diff
@@ -58,7 +58,8 @@ else
         --host=\${MACHTYPE} \\
         --with-sysroot="\${prefix}/$(compiler_target)/sys-root" \\
         --enable-multilib \\
-        --disable-werror
+        --disable-werror \\
+	--enable-new-dtags
```
vtjnash, line 62: this is not supported by this version of binutils

C/CompilerSupportLibraries/common.jl
```diff
@@ -113,6 +113,10 @@ if [[ ${target} == *apple* ]]; then
     done
 fi
 
+for lib in libgfortran libasan libtsan libubsan liblsan; do
+  [ ! -f "${libdir}/${lib}.so" ] || patchelf --set-rpath "\$ORIGIN" "${libdir}/${lib}.so"
+done
+
```
giordano, line 119: See https://github.com/JuliaPackaging/Yggdrasil/issues/3703#issuecomment-1291225562.  I believe this can be fixed in the audit pass, instead of having to manually fix it here.

L/LLVM/common.jl
```diff
@@ -273,6 +273,14 @@ ninja -j${nproc} -vv
 # Install!
 ninja install
 
+shopt -s nullglob
+for x in ${prefix}/lib/*.so; do
+    # All things that have RPATH or doesn't have $ORIGIN,
+    # but not those without $RUNPATH and $ORIGIN
+    [ -n "$(objdump -x "$x" | grep 'RPATH\|RUNPATH')" ] && \
+      patchelf --set-rpath "\$ORIGIN" "$x";
+done
```
giordano, line 282: Also here, the fact that [we have](https://github.com/JuliaPackaging/Yggdrasil/issues/3703#issuecomment-1290329832) `$ORIGIN/../lib` as RPATH makes me think this should have always been `$ORIGIN` (library is in `${prefix}/lib`, so `$ORIGIN/../lib` is just `$ORIGIN`).  Maybe when we [normalise the rpaths](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/23847acbc1ecf0c1a7cdc09ad657036776d749f7/src/auditor/dynamic_linkage.jl#L454-L485) we should flatten relative paths, so that these identities becomes clear and we can simplify the paths we put there.

vtjnash: the library may later get moved to `lib/julia`, resulting in breakage for people linking against it

that was fixed in Julia v1.7 (https://github.com/JuliaLang/julia/pull/42788), but strictly only on that release branch (@vchuravy)

C/CompilerSupportLibraries/common.jl
vtjnash: that would be somewhat simpler, but @staticfloat had thought it might fail in the verifier, so he suggested we do this manually

giordano: I replied in https://github.com/JuliaPackaging/Yggdrasil/issues/3703#issuecomment-1291264529, I think the assessment was ill-informed

PR #5758, opened 2022-10-28
casacorecxx: Bump to version 0.1.2

torrance:


giordano:
Note that it's generally a bad practice to open a PR from a branch with the same name as the target branch.  You should always create a ***new*** branch out of the target branch.  As you can see here, the history contains tons of unrelated commits.

torrance:
> Note that it's generally a bad practice to open a PR from a branch with the same name as the target branch. You should always create a _**new**_ branch out of the target branch. As you can see here, the history contains tons of unrelated commits.

Got it, sorry about that!

PR #5766, opened 2022-10-30
[RDKit] Correctly export symbols when targeting Windows

giordano:


giordano:
I'll get back to this when there is a resolution for https://github.com/rdkit/rdkit/issues/5709

PR #5780, opened 2022-10-31
[SCS_MKL] don't build binaries for i686

kalmarek:
There are some failures/segfaults on i686 which I can't reproduce/track locally
https://github.com/jump-dev/SCS.jl/pull/255#issuecomment-1107815005

So to move this forward it's best to disable the build until somebody asks for it ;)

Questions:
 1. Is it possible to add compat as `=3.2.1+1` (the version of binary built here?)
 2. does this merit/require changing to `3.2.2`? 

the point is that I want just to remove i686 binaries from the registry ;)

### DIFF
--- a/S/SCS/build_tarballs.jl
+++ b/S/SCS/build_tarballs.jl
@@ -1,11 +1,11 @@
 using BinaryBuilder
 
 name = "SCS"
-version = v"3.2.0"
+version = v"3.2.1"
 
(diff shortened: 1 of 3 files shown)


giordano:
> Is it possible to add compat as =3.2.1+1 (the version of binary built here?)

No because Pkg can't handle that.  If it's important to have a strict compat I recommend bumping the version number.

PR #3614, opened 2021-09-12
Wizard recipe: zlibng-v2.0.5

Moelf:
This pull request contains a new build recipe I built using the BinaryBuilder.jl wizard:

* Package name: zlibng
* Version: v2.0.5

@staticfloat please review and merge.

PR #5782, opened 2022-11-01
Add zlib-ng

Drvi:
I don't have much experience building C libraries and this is my first direct interaction with BinaryBuilder.jl.
I'm trying to build one of the more performant zlib forks, the `ZLIB_COMPAT` flag builds a version with an API compatible with the original zlib. Since the only purpose of this recipe is to unlock more performance, I tried setting some of the optimizations on but I am not sure if it is safe to do so. Any guidance would be greatly appreciated!

On my two laptops (M1 mac and intel linux), when trying to compress and decompress 1MB of lorem ipsum text, I see ~2x performance increase when I plug the resulting binary into CodecZlib.jl.

cc: @quinnj 

EDIT: No longer using the `ZLIB_COMPAT` due to soname clash with zlib.

### DIFF
--- a/Z/ZlibNG/build_tarballs.jl
+++ b/Z/ZlibNG/build_tarballs.jl
@@ -1,0 +1,62 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "ZlibNG"
+version = v"1.2.12"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/zlib-ng/zlib-ng.git", "0c118a14dd56d0484a8ca432cf801b9a6e6d024a")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/zlib-ng
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_FLAGS="-O3" \
+    -WITH_NATIVE_INSTRUCTIONS=true \
+    -DZLIB_COMPAT=true \
+    ..
+make -j${nproc}
+make install
+
+if [[ "${target}" == *-mingw* ]]; then
+    cp "${WORKSPACE}/srcdir/zlib-ng/build/zlib1.dll" "${libdir}/libz.dll"
+fi
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("i686", "windows"; ),
+    Platform("x86_64", "windows"; )
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libz", :libz)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
--- a/Z/ZlibNG/build_tarballs.jl
+++ b/Z/ZlibNG/build_tarballs.jl
@@ -22,10 +22,6 @@ cmake -DCMAKE_INSTALL_PREFIX=$prefix \
     ..
 make -j${nproc}
 make install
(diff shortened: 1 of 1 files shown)


Z/ZlibNG/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("i686", "windows"; ),
+    Platform("x86_64", "windows"; )
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libz", :libz)
```
giordano, line 54: This can't have the same soname as zlib https://github.com/JuliaPackaging/Yggdrasil/blob/1e926895c26b578d145db8d483654a91289bac4c/Z/Zlib/build_tarballs.jl#L40 otherwise this will never be loaded.

giordano (changes requested):

Z/ZlibNG/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_FLAGS="-O3" \
+    -DZLIB_COMPAT=ON \
+    ..
+make -j${nproc}
+make install
+
+if [[ "${target}" == *-mingw* ]]; then
+    cp "${WORKSPACE}/srcdir/zlib-ng/build/libzlib1.dll" "${libdir}/libz.dll"
+fi
```
giordano, line 28: https://github.com/JuliaPackaging/Yggdrasil/pull/3614#issuecomment-944856951

Z/ZlibNG/build_tarballs.jl
Drvi: Is `LibraryProduct("libz", :libzng)` better?

giordano: Not really, it's the fact that the build produces a file with [soname](https://en.wikipedia.org/wiki/Soname) the problem.  The second argument to [`LibraryProduct`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct) is just the name of the Julia variable used in the generated JLL package, it isn't very important

Drvi: Gotcha. Yeah so I think we won't go the `ZLIB_COMPAT` route, then there won't be a clash in sonames.

Z/ZlibNG/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("i686", "windows"; ),
+    Platform("x86_64", "windows"; )
+]
+
```
giordano, line 46:
```suggestion
platforms = supported_platforms()
```

Z/ZlibNG/build_tarballs.jl
```diff
@@ -0,0 +1,40 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/zlib-ng
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_FLAGS="-O3" \
```
giordano, line 20: I'm quite confident the release build type uses `-O3` already and the log shows duplicate `-O3` flags.
```suggestion
```

Z/ZlibNG/build_tarballs.jl
```diff
@@ -0,0 +1,40 @@
...
+    ..
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(exclude=[Platform("x86_64", "freebsd")])
```
giordano, line 28: What's the problem with FreeBSD?

giordano (commented):
This is starting to look better

Z/ZlibNG/build_tarballs.jl
Drvi: `error: use of undeclared identifier 'isascii'` at various places

[freebsd_error.txt](https://github.com/JuliaPackaging/Yggdrasil/files/9938063/freebsd_error.txt)

giordano: https://www.freebsd.org/cgi/man.cgi?query=isascii&sektion=3&apropos=0&manpath=freebsd

Drvi: The error comes from a test only dependency `gtest.h` so I think disabling tests would get rid of the problem... Probably the reason why https://github.com/JuliaPackaging/Yggdrasil/pull/3614 disabled them as well. Or is that a nono?:)

giordano: We don't run tests, so removing them is them.

Z/ZlibNG/build_tarballs.jl
```diff
@@ -0,0 +1,40 @@
...
+    GitSource("https://github.com/zlib-ng/zlib-ng.git", "b3dcf11b4204a16fde71fa8224d7b7054e225b93")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/zlib-ng
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
```
giordano, line 19: We're building and shipping static archives, looks like defining `BUILD_SHARED_LIBS` should skip them:
```suggestion
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=ON \
```

giordano (approved):
Thank you for your contribution!  CC: @Moelf

Drvi:
Thank you, @giordano, you are a saint:)

quinnj:
Woohoo; thanks @Drvi and @giordano!

PR #4135, opened 2021-12-30
add DynarePreprocessor

MichelJuillard:

PR #5795, opened 2022-11-03
Dynare v6.0.2: sparse storage for derivatives

MichelJuillard:


D/DynarePreprocessor/build_tarballs.jl
```diff
@@ -21,9 +21,14 @@ rm -f /usr/bin/flex
 
 # Help FreeBSD find header files.  See
 # https://github.com/JuliaPackaging/Yggdrasil/issues/3949
+if [[ "${target}" == *-freebsd* ]] || [[ "${target}" == *-apple-* ]]; then
+    CC=gcc
+    CXX=g++
+    LDFLAGS=-static-libgcc
+fi
```
giordano, line 28: We generally don't want to use GCC on BSD systems.  Why would be that necessary?

D/DynarePreprocessor/build_tarballs.jl
```diff
@@ -21,9 +21,14 @@ rm -f /usr/bin/flex
...
 if [[ "${target}" == *-freebsd* ]]; then
     export CPPFLAGS="-I${includedir}"
-elif [[ "${target}" == x86_64-apple-darwin* ]]; then
+    #elif [[ "${target}" == x86_64-apple-darwin* ]]; then
```
giordano, line 31: I'd prefer to remove dead lines of code, if that's really not needed anymore.

D/DynarePreprocessor/build_tarballs.jl
```diff
@@ -34,14 +39,14 @@ elif [[ "${target}" == x86_64-apple-darwin* ]]; then
     #           __ZNSt3__14__fs10filesystem4pathdVERKS2_ in dynare_preprocessor-DynamicModel.o
     #           __ZNSt3__14__fs10filesystem4pathdVERKS2_ in dynare_preprocessor-ModFile.o
     #           _main in dynare_preprocessor-DynareMain.o
-    pushd $WORKSPACE/srcdir/MacOSX10.*.sdk
-    rm -rf /opt/${target}/${target}/sys-root/System
+    pushd $WORKSPACE/srcdir/MacOSX11.*.sdk
+    #    rm -rf /opt/${target}/${target}/sys-root/System
```
giordano, line 43: Not removing the directory would result in a mix of different SDKs

D/DynarePreprocessor/build_tarballs.jl
```diff
@@ -34,14 +39,14 @@ elif [[ "${target}" == x86_64-apple-darwin* ]]; then
...
     cp -ra usr/* "/opt/${target}/${target}/sys-root/usr/."
     cp -ra System "/opt/${target}/${target}/sys-root/."
     popd
 fi
 
-atomic_patch -p1 "../patches/patches.patch"
+#atomic_patch -p1 "../patches/patches.patch"
```
giordano, line 49: Same as above: I'd remove lines which don't do anything.

D/DynarePreprocessor/build_tarballs.jl
```diff
@@ -67,4 +73,4 @@ dependencies = [
     HostBuildDependency("flex_jll"),
 ]
 
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"9")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"13")
```
giordano, line 76: Apart from the fact there is no version 13 of GCC, for maximum compatibility we want to use the _oldest_ version possible, not the latest one which would even result in binaries not compatible with any version of Julia.

MichelJuillard: OK, I will search for the first gcc version to support  std::filesystem::path

D/DynarePreprocessor/build_tarballs.jl
giordano, line 28:
```suggestion
```

D/DynarePreprocessor/build_tarballs.jl
giordano:
```suggestion
    elif [[ "${target}" == x86_64-apple-darwin* ]]; then
```

D/DynarePreprocessor/build_tarballs.jl
```diff
@@ -2,11 +2,11 @@ using BinaryBuilder
...
 name = "DynarePreprocessor"
-version = v"6.0.0"
+version = v"6.0.2"
 sources = [
-    GitSource("https://git.dynare.org/Dynare/preprocessor.git", "285a4389a3a368774124c9ae66ed7fc367592946"),
-    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
-                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
+    GitSource("https://git.dynare.org/Dynare/preprocessor.git", "00fd9dadb6dde5929c224c74e0a43715554a2c05"),
+    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/11.3/MacOSX11.3.sdk.tar.xz",
+                  "cd4f08a75577145b8f05245a2975f7c81401d75e9535dcffbb879ee1deefcbf4"),
```
giordano, line 9:
```suggestion
    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
```

D/DynarePreprocessor/build_tarballs.jl
giordano, line 43:
```suggestion
    pushd $WORKSPACE/srcdir/MacOSX10.*.sdk
    rm -rf /opt/${target}/${target}/sys-root/System
```

D/DynarePreprocessor/build_tarballs.jl
```diff
@@ -34,14 +39,14 @@ elif [[ "${target}" == x86_64-apple-darwin* ]]; then
     #           __ZNSt3__14__fs10filesystem4pathdVERKS2_ in dynare_preprocessor-DynamicModel.o
     #           __ZNSt3__14__fs10filesystem4pathdVERKS2_ in dynare_preprocessor-ModFile.o
     #           _main in dynare_preprocessor-DynareMain.o
-    pushd $WORKSPACE/srcdir/MacOSX10.*.sdk
-    rm -rf /opt/${target}/${target}/sys-root/System
+    pushd $WORKSPACE/srcdir/MacOSX11.*.sdk
+    #    rm -rf /opt/${target}/${target}/sys-root/System
     cp -ra usr/* "/opt/${target}/${target}/sys-root/usr/."
     cp -ra System "/opt/${target}/${target}/sys-root/."
     popd
```
giordano, line 46:
```suggestion
    export MACOSX_DEPLOYMENT_TARGET=10.15
    popd
```

D/DynarePreprocessor/build_tarballs.jl
```diff
@@ -56,6 +61,7 @@ install_license COPYING
 """
 
 platforms = expand_cxxstring_abis(supported_platforms(; experimental=true))
+platforms = expand_cxxstring_abis(platforms)
```
giordano, line 64:
```suggestion
platforms = expand_cxxstring_abis(supported_platforms())
```

D/DynarePreprocessor/build_tarballs.jl
MichelJuillard: SDK 10.15 doesn't support syntax like
```
Environment.cc:31:17: error: no member named 'contains' in `'std::map<std::basic_string<char>,std::tuple<std::shared_ptr<macro::Function>,std::shared_ptr<macro::Expression>>>'
if (functions.contains(name))
Environment.cc:40:17: error: no member named 'contains' in 'std::map<std::basic_string<char>,std::shared_ptr<macro::Expression>>'
if (variables.contains(name))
SDK 11.3 doesn't complain about this syntax, but raises on error on exception constructs elsewhere in the original source code. 
I will search what is the best combination of minimal SDK version and patches.

giordano:
Why dropping BSD systems?  I'm not a C++ expert, but I suspect this is a situation similar to #4135 where compilation with GCC was successful just because of the use of non-standard features and @fingolfin pointed out a few actual bugs in the source code.

MichelJuillard:
It's just too much work for me at the moment to try to make the original source code compatible with clang. I need to advance on the Julia code for Dynare and for that I mainly need the binary preprocessor for Linux and Windows.

PR #5798, opened 2022-11-03
Add gfm recipe

pfitzseb:


G/gfm/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd $WORKSPACE/srcdir/cmark-gfm
+mkdir build
+cd build
+cmake ../ -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+make
+make install
+exit
```
giordano, line 22: Some simplification
```suggestion
cd $WORKSPACE/srcdir/cmark-gfm
mkdir build
cd build
cmake ../ -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
make -j${nproc}
make install
```

G/gfm/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "gfm"
+version = v"29.0.6"
```
giordano, line 6: I liked 0.29.6 better, it was somewhat closer to the upstream version

pfitzseb: But what do I do when up-upstream releases `0.29.1`, which is then built as `0.29.1-gfm.0` by upstream? Seems much less likely that anyone releases a 1.x :)

giordano: People often do stuff like `100 * patch + else`, so that'd be `0.29.100`, and `0.29.2-gfm.5` would be `0.29.205`.

giordano: For example https://github.com/JuliaPackaging/Yggdrasil/blob/23c4fc91c6c00e1dc91f3423e7ba80829cd4ce1d/S/Singular/build_tarballs.jl#L30-L34

pfitzseb: Ok, I'm convinced.

mortenpi:
cmark-gfm already exists in Yggdrasil though: https://github.com/JuliaPackaging/Yggdrasil/blob/master/C/cmark_gfm/build_tarballs.jl

giordano:
Aaargh

PR #5803, opened 2022-11-04
Add ducc0 package

mreineck:
This adds a rudimentary JLL interface for the ducc0 package (https://gitlab.mpcdf.mpg.de/mtr/ducc). This is most likely not generally useful yet, but I don't know how to test the binary building without opening a PR and letting the CI build and test the package.

### DIFF
--- a/D/ducc0/build_tarballs.jl
+++ b/D/ducc0/build_tarballs.jl
@@ -11,8 +11,8 @@ version = v"0.27.0"
 # Collection of sources required to complete build
 sources = [
 GitSource("https://gitlab.mpcdf.mpg.de/mtr/ducc.git", "84967dd5d3e3062874a03c99a6d51ab375d3fb9d"),
-#ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
-#                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
+ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
+                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
 ]
 
 # Bash recipe for building across all platforms
@@ -20,12 +20,12 @@ script = raw"""
 
 #XFLAGS=
 if [[ "${target}" == x86_64-apple-darwin* ]]; then
-#    # Install a newer SDK to work around compilation failures
-#    pushd $WORKSPACE/srcdir/MacOSX10.*.sdk
-#    rm -rf /opt/${target}/${target}/sys-root/System
-#    cp -ra usr/* "/opt/${target}/${target}/sys-root/usr/."
-#    cp -ra System "/opt/${target}/${target}/sys-root/."
-#    popd
+    # Install a newer SDK to work around compilation failures
+    pushd $WORKSPACE/srcdir/MacOSX10.*.sdk
+    rm -rf /opt/${target}/${target}/sys-root/System
+    cp -ra usr/* "/opt/${target}/${target}/sys-root/usr/."
+    cp -ra System "/opt/${target}/${target}/sys-root/."
+    popd
     MACFLAGS=-mmacosx-version-min=10.14
 fi
 
--- a/D/ducc0/build_tarballs.jl
+++ b/D/ducc0/build_tarballs.jl
@@ -16,8 +16,8 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/ducc-*/julia
-${CXX} ${CFLAGS} -O3 -ffast-math -I ../src/ ducc_julia.cc -Wfatal-errors -pthread -std=c++17 -fPIC -c
-${CXX} ${CFLAGS} -O3 -march=native -o libducc_julia.${dlext} ducc_julia.o -Wfatal-errors -pthread -std=c++17 -shared -fPIC
+${CXX} ${CFLAGS} -O3 -I ../src/ ducc_julia.cc -Wfatal-errors -pthread -std=c++17 -fPIC -c
+${CXX} ${CFLAGS} -O3 -o libducc_julia.${dlext} ducc_julia.o -Wfatal-errors -pthread -std=c++17 -shared -fPIC
 install -Dvm 0755 "libducc_julia.${dlext}" "${libdir}/libducc_julia.${dlext}"
 """
 
--- a/D/ducc0/build_tarballs.jl
+++ b/D/ducc0/build_tarballs.jl
@@ -10,7 +10,7 @@ version = v"0.27.0"
 
 # Collection of sources required to complete build
 sources = [
-GitSource("https://gitlab.mpcdf.mpg.de/mtr/ducc.git", "0fde4979684d28966157356211ad278ec3278cf9")
+GitSource("https://gitlab.mpcdf.mpg.de/mtr/ducc.git", "84967dd5d3e3062874a03c99a6d51ab375d3fb9d")
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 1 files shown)


mreineck:
The package currently fails to build on `x86_64-apple-darwin` targets, because apparently the OS version is too old (older than `macOS 10.14`). Is there a way to disable building for such targets?

giordano:
Instead of skipping macOS, you can build for it.  It's a bit of a hack, but it works:
https://github.com/JuliaPackaging/Yggdrasil/blob/84f736f26a915dc39a07bfb721c54f4e194f4ffc/M/minio/build_tarballs.jl#L9-L10
https://github.com/JuliaPackaging/Yggdrasil/blob/84f736f26a915dc39a07bfb721c54f4e194f4ffc/M/minio/build_tarballs.jl#L18-L25
You'd use SDK 10.14 instead of 10.15 as in that example.

mreineck:
Thanks a lot for the hint! First attempt doesn't seem successful, but I'll investigate further tomorrow!

mreineck:
OK, I think I got it working!

Is there any chance I can try the generated JLLs before proceeding with the merge?

mreineck:
Found a way of obtaining the artifacts and can do local tests!

mreineck:
From my point of view this is OK to merge. Since I'm a total noob at Julia (and creating Julia packages) I expect that there will be updates soon, and that the interface will be unstable for some time. If this is nevertheless OK with the maintainers, it would be very convenient for me to have this prototype package in Yggdrasil.

D/ducc0/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+    cp -ra System "/opt/${target}/${target}/sys-root/."
+    popd
+    MACFLAGS=-mmacosx-version-min=10.14
+fi
+
+cd $WORKSPACE/srcdir/ducc*/julia
+install_license ../LICENSE
+${CXX} ${CFLAGS} ${MACFLAGS} -O3 -I ../src/ ducc_julia.cc -Wfatal-errors -pthread -std=c++17 -fPIC -fno-math-errno -fassociative-math -freciprocal-math -fno-signed-zeros -fno-trapping-math -ffp-contract=fast -ffinite-math-only -fno-rounding-math -fno-signaling-nans -fexcess-precision=fast -c
+# -fcx-limited-range is not supported by clang
+${CXX} ${CFLAGS} ${MACFLAGS} -O3 -o libducc_julia.${dlext} ducc_julia.o -Wfatal-errors -pthread -std=c++17 -shared -fPIC
```
giordano, line 35:
```suggestion
    CXXFLAGS="-mmacosx-version-min=10.14"
fi

cd $WORKSPACE/srcdir/ducc*/julia
install_license ../LICENSE
${CXX} ${CXXFLAGS} -O3 -I ../src/ ducc_julia.cc -Wfatal-errors -pthread -std=c++17 -fPIC -fno-math-errno -fassociative-math -freciprocal-math -fno-signed-zeros -fno-trapping-math -ffp-contract=fast -ffinite-math-only -fno-rounding-math -fno-signaling-nans -fexcess-precision=fast -c
# -fcx-limited-range is not supported by clang
${CXX} ${CXXFLAGS} -O3 -o libducc_julia.${dlext} ducc_julia.o -Wfatal-errors -pthread -std=c++17 -shared -fPIC
```

D/ducc0/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+    end
+    """
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libducc_julia", :libducc_julia)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = []
```
giordano, line 61:
```suggestion
dependencies = [
    Dependency("CompilerSupportLibraries_jll"; platforms=filter(p -> Sys.islinux(p) || Sys.isfreebsd(p), platforms))
]
```

D/ducc0/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+include(joinpath(@__DIR__, "..", "..", "platforms", "microarchitectures.jl"))
+
+name = "ducc0"
+version = v"0.27.0"
+
+# Collection of sources required to complete build
+sources = [
+GitSource("https://gitlab.mpcdf.mpg.de/mtr/ducc.git", "84967dd5d3e3062874a03c99a6d51ab375d3fb9d"),
+ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
+                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
```
giordano, line 15:
```suggestion
    GitSource("https://gitlab.mpcdf.mpg.de/mtr/ducc.git", "84967dd5d3e3062874a03c99a6d51ab375d3fb9d"),
    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
```

D/ducc0/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+${CXX} ${CFLAGS} ${MACFLAGS} -O3 -I ../src/ ducc_julia.cc -Wfatal-errors -pthread -std=c++17 -fPIC -fno-math-errno -fassociative-math -freciprocal-math -fno-signed-zeros -fno-trapping-math -ffp-contract=fast -ffinite-math-only -fno-rounding-math -fno-signaling-nans -fexcess-precision=fast -c
+# -fcx-limited-range is not supported by clang
+${CXX} ${CFLAGS} ${MACFLAGS} -O3 -o libducc_julia.${dlext} ducc_julia.o -Wfatal-errors -pthread -std=c++17 -shared -fPIC
+install -Dvm 0755 "libducc_julia.${dlext}" "${libdir}/libducc_julia.${dlext}"
+"""
+
+# Expand for microarchitectures on x86_64 (library doesn't have CPU dispatching)
+# Tests on Linux/x86_64 yielded a slow binary with avx512 for some reason, so disable that
+platforms = expand_cxxstring_abis(expand_microarchitectures(supported_platforms(), ["x86_64", "avx", "avx2"]); skip=!Sys.iswindows)
+
```
giordano, line 42:
```
┌ Warning: /agent/_work/1/s/D/ducc0/build/x86_64-linux-gnu-march+avx2/97WhZ0rQ/x86_64-linux-gnu-libgfortran5-cxx11-march+avx2/destdir/lib/libducc_julia.so contains std::string values!  This causes incompatibilities across the GCC 4/5 version boundary.  To remedy this, you must build a tarball for both GCC 4 and GCC 5.  To do this, immediately after your `platforms` definition in your `build_tarballs.jl` file, add the line:
│ 
└ @ BinaryBuilder.Auditor /depot/packages/BinaryBuilder/7iMm2/src/auditor/compiler_abi.jl:247
```
```suggestion
platforms = expand_cxxstring_abis(platforms)

```

mreineck: Will this still produce separate packages for AVX and AVX2 capable targets? This is fairly important for the package.

giordano: Yes.

Side note, support for microarchitectures is ***extremely*** experimental at the moment and liable to be changed in breaking way later.  This isn't even documented (or just barely) precisely for this reason.  More specifically, building the libraries works well, the problem is handling the installation of the artifacts, we haven't had the time to improve that.

mreineck: Thanks for pointing this out. I copied the block from https://github.com/JuliaPackaging/Yggdrasil/tree/master/F/finufft, since the requirements are pretty similar.

I'm happy to apply changes to this as and when they become necessary. If I may make a suggestion, please consider adding extra architectures for FMA3/4 support as well. Some numerical codes can get around 50% speedup when they are allowed to use these instructions.

giordano:
Thanks for your contribution!

PR #5568, opened 2022-09-24
[SDL2] Add a dependency on alsa_plugins

aviks:
In partial mitigation of https://github.com/JuliaPackaging/Yggdrasil/issues/1432#issuecomment-851065063

PR #5812, opened 2022-11-06
Update SDL2 to 2.24

Gnimuc:
https://github.com/JuliaPackaging/Yggdrasil/pull/5568

### DIFF
--- a/S/SDL2/build_tarballs.jl
+++ b/S/SDL2/build_tarballs.jl
@@ -3,17 +3,17 @@
 using BinaryBuilder
 
 name = "SDL2"
-version = v"2.0.20"
+version = v"2.24.0"
 
 # Collection of sources required to build SDL2
 sources = [
-    ArchiveSource("https://libsdl.org/release/SDL2-$(version).tar.gz",
-                  "c56aba1d7b5b0e7e999e4a7698c70b63a3394ff9704b5f6e1c57e0c16f04dd06"),
+    GitSource("https://github.com/libsdl-org/SDL.git",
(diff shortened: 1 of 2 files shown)


Gnimuc:
cc @mfiano I don't have time to debug the apple aarch64 build, so I disabled that platform in this PR.

mfiano:
> I disabled that platform in this PR

That's quite alright. Thank you again!

Gnimuc:
This upstream issue might be related: https://github.com/libsdl-org/SDL/issues/4958

giordano:
> That's quite alright.

I doubt users of that platform, which isn't a small fraction of Julia users, would agree.

giordano:
Sounds like this is trying to build the joystick components for iOS when targeting aarch64-apple-darwin, which looks like wrong.

giordano (changes requested):
In case it wasn't clear, someone (!= me) should investigate why the aarch64-apple-darwin build is failing (building for iOS looks suspiciously wrong to me.  ***Edit***: looking at the [build log of 2.0.20](https://github.com/JuliaBinaryWrappers/SDL2_jll.jl/releases/download/SDL2-v2.0.20%2B1/SDL2-logs.v2.0.20.aarch64-apple-darwin.tar.gz) I noticed that also back then we built the joystick component for iOS, but at least it wasn't failing, so something happened), it isn't very nice to drop support for widespread platforms, which is the reason why https://github.com/JuliaPackaging/Yggdrasil/pull/5568 wasn't merged including the upgrade to v2.24

giordano (approved):

PR #5814, opened 2022-11-06
Add msgpack

pxl-th:
This PR adds [msgpack](https://msgpack.org/index.html) which is needed to switch rocBLAS internal serialization format from yaml to msgpack.

### DIFF
--- a/M/msgpack/build_tarballs.jl
+++ b/M/msgpack/build_tarballs.jl
@@ -31,7 +31,7 @@ function configure_build(version)
     """
 
     sources = [GitSource(GIT, GIT_TAGS[tag])]
-    dependencies = []
+    dependencies = Dependency[]
     NAME, version, sources, buildscript, PLATFORMS, PRODUCTS, dependencies
 end
 
--- a/M/msgpack/build_tarballs.jl
+++ b/M/msgpack/build_tarballs.jl
@@ -7,10 +7,7 @@ const GIT_TAGS = Dict(
     "cpp-3.0.1" => "2c4f2b890ef1546fc022d270d11e657f6fc8022f",
 )
 
-const PLATFORMS = [
-    Platform("x86_64", "linux"; libc="glibc", cxxstring_abi="cxx11"),
-    Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx11"),
-]
+const PLATFORMS = supported_platforms()
 const PRODUCTS = [LibraryProduct(["libmsgpackc"], :libmsgpackc, ["lib"])]
 
 function configure_build(version)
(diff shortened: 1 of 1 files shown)


M/msgpack/build_tarballs.jl
```diff
@@ -0,0 +1,40 @@
...
+        -DMSGPACK_BUILD_TESTS=OFF \
+        -DMSGPACK_BUILD_EXAMPLES=OFF \
+        -DCMAKE_BUILD_TYPE=Release
+    make -j${nproc} -C build install
+
+    install_license ${WORKSPACE}/srcdir/msgpack*/LICENSE_1_0.txt
+    """
+
+    sources = [GitSource(GIT, GIT_TAGS[tag])]
+    dependencies = []
```
giordano, line 34:
```suggestion
    dependencies = Dependency[]
```

M/msgpack/build_tarballs.jl
```diff
@@ -0,0 +1,40 @@
...
+const NAME = "msgpack"
+const GIT = "https://github.com/msgpack/msgpack-c.git"
+const GIT_TAGS = Dict(
+    "cpp-3.0.1" => "2c4f2b890ef1546fc022d270d11e657f6fc8022f",
+)
+
+const PLATFORMS = [
+    Platform("x86_64", "linux"; libc="glibc", cxxstring_abi="cxx11"),
+    Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx11"),
+]
```
giordano, line 13: Why only these two platforms?

pxl-th: These are the platforms that we target for ROCm, but we can build for all supported platforms.

giordano: If this is potentially useful beyond ROCm it may be worth to try and build for more platforms.

M/msgpack/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    sources = [GitSource(GIT, GIT_TAGS[tag])]
+    dependencies = Dependency[]
+    NAME, version, sources, buildscript, PLATFORMS, PRODUCTS, dependencies
+end
+
+build_tarballs(
+    ARGS, configure_build(v"3.0.1")...;
+    preferred_gcc_version=v"7", preferred_llvm_version=v"9", julia_compat="1.6")
```
giordano, line 37: Is `preferred_llvm_version=v"9"` necessary? :flushed:

pxl-th: Apparently no, even fixed macos build.

PR #5817, opened 2022-11-06
[SDL2] Go back to autoconf

giordano:
When switching to CMake the soname of macOS binaries went from `@rpath/libSDL2-2.0.0.dylib` to `@rpath/libSDL2-2.0.dylib`, breaking the ABI.

CC: @Gnimuc


Gnimuc:
Thanks for looking into this!

PR #5821, opened 2022-11-07
[rocBLAS] Add compat and switch from yaml to msgpack serialization

pxl-th:
~~Should hopefully work again on musl as well...~~ Still gives LLVM OOM
However, if building for one target, it works.


pxl-th:
We can merge it. Future rocBLAS versions should reduce memory consumption, that might fix musl build for `all` targets.

PR #5824, opened 2022-11-07
[Ipopt] Bump version

jeremiahpslewis:


giordano:
CC @odow @amontoison

amontoison:
We will have different versions of Ipopt on Julia 1.6 and Julia >= 1.8 if we merge this PR.
Even if we compiled different artifacts Ipopt_jll with or without LBT for Julia v1.6 / v1.7 / v1.8, it was always for Ipopt 3.14.4.
It could more difficult to maintain for @odow.

jeremiahpslewis:
Here's the commit diff, doesn't look like it's necessary / worth it if merging this creates overhead for a maintainer. https://github.com/coin-or/Ipopt/compare/releases/3.14.4...releases/3.14.10

odow:
I think it's probably okay to merge this. 

We're strict on which versions Ipopt.jl accepts: https://github.com/jump-dev/Ipopt.jl/blob/292640b936e76bcacddbe7841c99526054f9d40e/Project.toml#L13

So we can test before bumping Ipopt.jl. It'd only be a problem if there was a breaking API change.

ViralBShah:
Merge?

ViralBShah:
Should be rebased on the latest master, in order to pull in the recent Coin-OR build system updates.

fingolfin:
I've rebased it to trigger a fresh rebuild. @odow agreed this can be merged. No objection from @amontoison . So OK to merge if the CI passes?

PR #5827, opened 2022-11-08
[PARMETIS] Add Int32/Int64 and Real32/Real64 versions of PARMETIS

Wimmerer:
Matching PR of METIS_jll. To allow for doing similar to SuperLU_DIST_jll.

### DIFF
--- a/P/PARMETIS/build_tarballs.jl
+++ b/P/PARMETIS/build_tarballs.jl
@@ -56,8 +56,8 @@ build_parmetis()
     -DCMAKE_C_FLAGS="-DIDXTYPEWIDTH=${1} -DREALTYPEWIDTH=${2}" \
     -DBINARY_NAME="${PARMETIS_NAME}" \
     -DMETIS_LIBRARY="${METIS_NAME}" \
-    -DCMAKE_SHARED_LINKER_FLAGS="-L${METIS_PATH}/lib" \
-    -DCMAKE_EXE_LINKER_FLAGS="-L${METIS_PATH}/lib"
+    -DCMAKE_SHARED_LINKER_FLAGS="-L${METIS_PATH}/lib -L${libdir}" \
+    -DCMAKE_EXE_LINKER_FLAGS="-L${METIS_PATH}/lib -L${libdir}"
 
     make -j${nproc}
     make install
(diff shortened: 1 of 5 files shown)


Wimmerer:
Hmm, it's looking for the mpiwrapper in the subdirectory, not in the `$libdir`.

amontoison:
@Wimmerer 
Do you think that we can add `windows` platforms now?
We added PARMETIS for MUMPS a long time ago and `MPITrampoline` was not available.

Wimmerer:
I can look into it.

@giordano do I need to `dont_dlopen` for libraries that depend on libraries in subfolders? That will be resolved by dlopening them at runtime right?

amontoison:
@Wimmerer 
You could use this [trick](https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SCALAPACK32/build_tarballs.jl#L70-L90) if you want to have all versions of the shared libraries of METIS and PARMETIS in the same folder.
You will just link with `-lmetis_Int32_Real64` instead of `-lmetis`.

Wimmerer:
@giordano this is ready now. Just flattened the directories, the include files are identical.

P/PARMETIS/build_tarballs.jl
```diff
@@ -63,12 +85,15 @@ platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && Sys.isfreebsd(p)), plat
...
+    LibraryProduct("libparmetis", :libparmetis),
+    LibraryProduct("libparmetis_Int32_Real64", :libparmetis_Int32_Real64),
+    LibraryProduct("libparmetis_Int64_Real32", :libparmetis_Int64_Real32),
+    LibraryProduct("libparmetis_Int64_Real64", :libparmetis_Int64_Real64)
 ]
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("METIS_jll"),
+    Dependency("METIS_jll", v"5.1.2"),
```
giordano, line 96: If a specific version is important for building, probably it's also important to explicitly set compat:
```suggestion
    Dependency("METIS_jll"; compat="5.1.2"),
```

PR #5829, opened 2022-11-08
Add a shared library for METIS4

amontoison:
Mosè, do we need to modify the soname with `patchelf` when we generate a shared library from a static library with a different name?
I don't want to create conflicts with `metis.$dlext` from METIS_jll.
I'm also not sure that we can have two shared libraries that coexist with the same symbols even if they have different name and soname.

### DIFF
--- a/M/METIS/METIS@4/build_tarballs.jl
+++ b/M/METIS/METIS@4/build_tarballs.jl
@@ -15,7 +15,7 @@ using BinaryBuilder, Pkg
 
 name = "METIS4"
 upstream_version = v"4.0.3"
-version_offset = v"0.0.0" # reset to 0.0.0 once the upstream version changes
+version_offset = v"0.0.1" # reset to 0.0.0 once the upstream version changes
 version = VersionNumber(upstream_version.major * 100 + version_offset.major,
                         upstream_version.minor * 100 + version_offset.minor,
                         upstream_version.patch * 100 + version_offset.patch)
@@ -48,6 +48,8 @@ mkdir -p ${prefix}/lib
 mv libmetis.a ${prefix}/lib
 mkdir -p ${prefix}/include
 cp Lib/metis.h ${prefix}/include
+cd ${prefix}/lib
+$CC -shared $(flagon -Wl,--whole-archive) libmetis.a $(flagon -Wl,--no-whole-archive) -o libmetis4.${dlext}
 """
 
 # These are the platforms we will build for by default, unless further
@@ -56,6 +58,7 @@ platforms = supported_platforms()
 
 # The products that we will ensure are always built
 products = [
+    LibraryProduct("libmetis4", :libmetis4),
     FileProduct("lib/libmetis.a", :libmetis_a)
 ]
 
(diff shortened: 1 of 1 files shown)


giordano:
You mean https://github.com/JuliaPackaging/Yggdrasil/blob/59b210b344372b22fdbe50b693875e115848bb5b/M/METIS/METIS%405/build_tarballs.jl#L56 ? That seems to be called `libmetis`, not `libmetis4`, right?

amontoison:
Yes, but we have two different artifacts, `METIS4_jll` and `METIS_jll`. They can be installed in the same environment.

ViralBShah:
This is why I deliberately did not make this a shared library, and thought linking it statically would avoid problems with METIS 5. I think it should work almost everywhere.

But if we want it to be a shared library, we should certainly change the name of the file - but is that sufficient?

ViralBShah:
@giordano Is changing the filename sufficient to have a `libmetis4`, or do we need to run any other tools on the resulting binary? If just the rename is sufficient, we should merge.

amontoison:
If it's not sufficient, I will link with the static library.

imciner2:
Since you are calling the compiler to form the shared library directly, I don't think you are even setting a soname for the library right now. To properly set the soname you will need to add `-Wl,-soname,libmetis4.${dlext}` to the compiler command (or at least for gcc-based compilers).

ViralBShah:
We can avoid building the old static library going forward. It'll probably break a couple of Yggdrasil package builds, but we should probably find those and fix once we merge this. Should be straightforward.

amontoison:
I think that `METIS4_jll` is only used by `HSL.jl`. I can easily update it.
I just don't understand why `-soname` is working on Mac, I had an issue with it for MUMPS (https://github.com/JuliaPackaging/Yggdrasil/blob/master/M/MUMPS/MUMPS%405/build_tarballs.jl#L32-L36).

imciner2:
> I just don't understand why -soname is working on Mac, I had an issue with it for MUMPS ([master/M/MUMPS/MUMPS%405/build_tarballs.jl#L32-L36](https://github.com/JuliaPackaging/Yggdrasil/blob/master/M/MUMPS/MUMPS%405/build_tarballs.jl?rgh-link-date=2022-11-18T22%3A35%3A12Z#L32-L36)).

I don't know exactly, but when I downloaded the artifact from the build where it used `-soname`, it gives the proper name:

```
          cmd LC_ID_DYLIB
      cmdsize 48
         name @rpath/libosqp.dylib (offset 24)
   time stamp 1 Thu Jan  1 01:00:01 1970
      current version 0.0.0
compatibility version 0.0.0
```

Maybe it ignored that string and just used the name of the output file instead?

amontoison:
Yes, this is highly likely that it's the default behavior.
I  updated the PR to use `install_name` on Mac just in case.

ViralBShah:
METIS4 is used in a couple of packages:

```
➜  Yggdrasil git:(master) grep METIS4 * -r
M/MUMPS/MUMPS_seq@4/build_tarballs.jl:    BuildDependency(PackageSpec(; name = METIS4_jll",
S/SDPA/build_tarballs.jl:METIS_packagespec = PackageSpec(; name = "METIS4_jll",
```

amontoison:
`METIS4_jll` is used as a `BuildDependency` for MUMPS4 and SDPA and the version `v"400.000.300"` is already specified. It's great because we don't need to update these artifacts.

ViralBShah:
Right, but new builds of those libraries will fail because no more static library, and also change of name. So, it would be best to update those build recipes now - but old artifacts will keep working.

Issue #5787, opened 2022-11-02
libjulia_jll: adding a `debug` variant


fingolfin:
A bunch of our JLLs link against `libjulia`, e.g. `libcxxwrap_julia_jll` or `GAP_jll` and several more.

When debugging things, it's useful to run a debug build of Julia -- but those are linked against `libjulia-debug`, not `libjulia`. This then leads to errors when trying to load the JLLs which want to use `libjulia`, resulting in this error (see https://github.com/oscar-system/GAP.jl/issues/834):
```
julia> using GAP_jll
ERROR: InitError: could not load library "/home/tim/.julia/artifacts/e9b6449c9e90ab3718db7e31df350994ebe44003/lib/libgap.so"
libjulia.so.1: cannot open shared object file: No such file or directory
Stacktrace:
  [1] dlopen(s::String, flags::UInt32; throw_error::Bool)
    @ Base.Libc.Libdl ./libdl.jl:117
...
```

In a Slack discussion on this,  @staticfloat suggested the following:
> Easy `.pkg` platform augmentation hook; add a `debug` tag based on whether the current Julia process is debug or not, then build both types of libraries

> Debatable PR to base: include the `debug` tag by default in `HostPlatform()` so that you don’t have to do that for future Julia builds

I'd like to work towards this, but would need some help. Hopefully with help by e.g. @giordano @staticfloat @benlorenz and perhaps others we can get there... I assume @barche might also be interested in this feature?

Here is what my (possibly incorrect) understanding is of what we need to do it, and how to get there, plus questions.

1. Modify `libjulia_jll` to also build `libjulia-debug` shared libraries
   - I hope that for this it'll be sufficient to add `julia-src-debug julia-cli-debug` to the make targets?
   - (and in theory similar for older Julia versions, but I wouldn't bother with those)
   - I don't think we'd need `debug` variants of `libjulia_jll`, though -- we can just ship both shared libraries side by side? At least I see no immediate advantage for such a split (remember, this JLL is only used for building other JLLs, users are never supposed to install or load it)
2. tweak JLLs using `libjulia_jll` (such as `GAP_jll`, `libcxxwrap_julia_jll` etc.) to offer `debug` variants (as separate artifacts) which will be built exactly like the regular ones, only they use `-ljulia-debug` instead of `-ljulia`
[...]

PR #5838, opened 2022-11-09
libjulia: add 1.10, update 1.9, add debug shared libs

fingolfin:
Also update Julia 1.9-DEV to latest git version, just in case.

First step towards #5787. It does *not* add `debug` to the platform as Julia happily allows `libjulia` and `libjulia-debug` to co-exist...

... though perhaps we *should* augment the platform like that, and only build either `libjulia` or `libjulia-debug`? One argument for doing that is that it should make it impossible to accidentally link against `libjulia` when one should be linking against `libjulia-debug`, and vice-versa.

As an additional data point, I looked into adding a `platforms/julia-debug.jl`, and realized that all packages that would include it already including `L/libjulia/common.jl`. So now I am wondering if either the `augment_platform` helpers should just go in there -- or dually, that it should all go into a `platforms/julia.jl` file (which then would also be included by `L/libjulia/common.jl`).

Thoughts, opinions, preferences?

### DIFF
--- a/L/libjulia/build_tarballs.jl
+++ b/L/libjulia/build_tarballs.jl
@@ -1,7 +1,7 @@
 include("common.jl")
-jllversion=v"1.8.0"
+jllversion=v"1.10.0"
 build_julia(ARGS, v"1.6.3"; jllversion)
 build_julia(ARGS, v"1.7.0"; jllversion)
 build_julia(ARGS, v"1.8.2"; jllversion)
 build_julia(ARGS, v"1.9.0-DEV"; jllversion)
-
+build_julia(ARGS, v"1.10.0-DEV"; jllversion)
--- a/L/libjulia/bundled/patches/1.10.0-DEV/cycleclock.patch
+++ b/L/libjulia/bundled/patches/1.10.0-DEV/cycleclock.patch
@@ -1,0 +1,54 @@
+From cc7f4029d82343f1c109d925fa4beb1c87a62387 Mon Sep 17 00:00:00 2001
+From: Max Horn <max@quendi.de>
+Date: Thu, 27 Oct 2022 17:06:17 +0200
+Subject: [PATCH] Provider cycleclock() for 32bit ARM targets
+
+Taken from https://github.com/google/benchmark/blob/main/src/cycleclock.h
+---
+ src/julia_internal.h | 23 +++++++++++++++++++++++
+ 1 file changed, 23 insertions(+)
+
+diff --git a/src/julia_internal.h b/src/julia_internal.h
+index 7eb34239e7..9eec7f5b69 100644
+--- a/src/julia_internal.h
++++ b/src/julia_internal.h
+@@ -20,6 +20,9 @@
+ #else
+ #define sleep(x) Sleep(1000*x)
+ #endif
++#if defined(_CPU_ARM_)
++#include <sys/time.h>
++#endif
+ 
+ #ifdef __cplusplus
+ extern "C" {
+@@ -216,6 +219,26 @@ static inline uint64_t cycleclock(void) JL_NOTSAFEPOINT
+     int64_t virtual_timer_value;
+     __asm__ volatile("mrs %0, cntvct_el0" : "=r"(virtual_timer_value));
+     return virtual_timer_value;
++#elif defined(_CPU_ARM_)
++  // V6 is the earliest arch that has a standard cyclecount
++#if (__ARM_ARCH >= 6)
++  uint32_t pmccntr;
++  uint32_t pmuseren;
++  uint32_t pmcntenset;
++  // Read the user mode perf monitor counter access permissions.
++  asm volatile("mrc p15, 0, %0, c9, c14, 0" : "=r"(pmuseren));
++  if (pmuseren & 1) {  // Allows reading perfmon counters for user mode code.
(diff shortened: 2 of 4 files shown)


giordano:
As a datapoint, doubling the number of platforms would likely go over the hard limit of [256 jobs for a stage](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/stages?view=azure-devops&tabs=yaml) Azure Pipelines, so avoiding that would have some benefits in terms of infrastructure.

fingolfin:
@giordano that's good.

I see that the `release-1.9` branch was opened, so we *also* need to add 1.10-DEV to libjulia... and then rebuild all JLLs using libjulia for that. perhaps I can combine the work.

fingolfin:
I've now added 1.10-DEV and updated 1.9-DEV to match the current `release-1.9` branch. And of course it adds `libjulia-debug` -- I figure this won't hurt even if perhaps we don't need it.

So assuming the tests pass, it should be OK to merge (of course resolving those ARM 32bit issues would be great, but that can't be helped for now :-/)

fingolfin:
Hmm, `LoadError: Unsupported Julia version`

giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/a0b7b6b816aa029cac6d074d832963a684057676/L/libjulia/common.jl#L360

fingolfin:
Argh :facepalm: Yeah, it helps if one adds all changes before committing&pushing *sigh*

fingolfin:
Huh, I don't get why `mingw` fails for 1.10 but not for 1.9... the two given commits only differ in three files: `VERSION`, `VERSION`, `NEWS.md`.

fingolfin:
Thanks @benlorenz . I didn't realize we had a patch there.

fingolfin:
OK, now it passes, phew :-)

fingolfin:
This already has 118 platforms; once/if ARM32 is fixed for Julia 1.9 and 1.10, that'll be another 8, for a total of 126... So yeah, that is already quite tight! If a JLL needs libjulia and e.g. has to expand fortran platforms, there's already a  problem now, I guess?

I wonder if we could not do away with the `expand_cxxstring_abis` after all. Well, we wisely documented why it is used for libjulia_jll:
```
    # While the "official" Julia kernel ABI itself does not involve any C++
    # symbols on the linker level, `libjulia` still exports "unofficial" symbols
    # dependent on the C++ strings ABI (coming from LLVM related code). This
    # doesn't matter if the client code is pure C, but as soon as there are
    # other (actual) C++ dependencies, we must make sure to use the matching C++
    # strings ABI. Hence we must use `expand_cxxstring_abis` below.

```
But I had a look at `nm libjulia.so` and `nm libjulia.dylib` for various Linux and macOS versions of `libjulia`, and in none was there a single C++ symbol visible. Did I look in the wrong place? (of course on macOS, we don't expand on the C++ ABI anyway, but that's orthogonal to the point)

fingolfin:
@benlorenz thanks, that makes sense. I've now pushed a commit which disables the C++ ABI expansion for Julia >= 1.6. Let's see if @giordano (or @staticfloat ?) have concerns.

staticfloat:
That's actually a great find, thank you @benlorenz!  Yes, since we've separated out `libjulia` from `libjulia-internal`, and `libjulia-internal` is the only one that does C++ stuff, that actually does give us greater freedom to link against the C-only `libjulia` without needing to worry about cxxstring ABI expansion.

giordano:
I usually look at the auditor output: if it says "locks us to..." for one of the libraries we care about then we need to expand the ABIs, otherwise we can't. This would be possible to automate if auditor was more structured and we could collect the status of each pass individually (at the moment we only have a global true/false). If auditor is happy with not splitting, then I'm happy as well.

fingolfin:
There are 490 warnings on https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=23639&view=results , including:
- complaints about the C++ API/ABI for `libjulia-internal` -- but we know that, and are OK with it, I think?
  - would it make sense / be helpful to remove `libjulia-internal` from the JLL, thus removing the warning and also making it impossible to "accidentally" link to it?
  - this would also silence further warnings about use of cpuid in there
- `Warning: Dependency LibGit2_jll does not have an (Stdlib)Artifacts.toml` (for older version used in Julia 1.6)
- `Warning: Dependency LibGit2_jll does not have an (Stdlib)Artifacts.toml` (for older version used in Julia 1.6)
- many variants of `Skipping abspath scanning of /agent/_work/1/s/L/libjulia/build/powerpc64le-linux-gnu-julia_version+1.7.0/bynylOc5/powerpc64le-linux-gnu-libgfortran4-cxx11-julia_version+1.7.0/destdir/tools/wasm-ld, as we can't open it` -- huh?

(Boy, it is super annoying that I have to click dozens of times to expand the list of warning, and then I can't even right-click on entries to open them in a new tab, because those aren't real hyperlinks; and opening a link and then going back resets the list of warnings to the initial state with just 10 shown. Grrr)

fingolfin:
I've now added another commit to remove `libjulia-internal`.

fingolfin:
Drawback of removing `libjulia-internal` is that the auditor then complains that it can't find `libjulia-internal`... so for now I added it back.

From my POV this is ready for merging.

PR #5839, opened 2022-11-10
[GR] bump version

jheinen:
bump GR version to 0.70.0

### DIFF
--- a/G/GR/build_tarballs.jl
+++ b/G/GR/build_tarballs.jl
@@ -3,13 +3,13 @@
 using BinaryBuilder
 
 name = "GR"
-version = v"0.69.1"
+version = v"0.70.0"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/sciapp/gr.git", "5d7f4fd40a3c36d839d08fdc45864a8382a1facd"),
+    GitSource("https://github.com/sciapp/gr.git", "25cf336838e06b5df0a7d74df2095a5ac7ce9e56"),
     FileSource("https://github.com/sciapp/gr/releases/download/v$version/gr-$version.js",
-               "b5a36d7bf6cdc36eb190eb230feed08ed3c10527155d9c52dc6c15b2e062ebca", "gr.js")
+               "9a086916c3a6f331db5b7a5344989f825fd7db1c0b10c236cbb9b9cccf3c006c", "gr.js")
 ]
 
 # Bash recipe for building across all platforms


mkitti:
We're going to need a retroactive pinning of older GR minor versions to their corresponding GR_jll minor versions so that we do not get a GR v0.69 and GR_jll v0.70 installed in the same environment due to `dont_dlopen`.

https://github.com/JuliaRegistries/General/pull/71976

giordano:
> We're going to need a retroactive pinning of older GR minor versions to their corresponding GR_jll minor versions so that we do not get a GR v0.69 and GR_jll v0.70 installed in the same environment due to dont_dlopen.

Isn't it easier to remove `dont_dlopen` for the time being until that's sorted out?

jheinen:
> Isn't it easier to remove `dont_dlopen` for the time being until that's sorted out?

Sounds like a chicken or egg problem ...

giordano:
> Sounds like a chicken or egg problem ...

Is it?  @mkitti proposal is to publish versions which are not going to used anywhere because they have known problems, my suggestion is to remove the known problem in the short term while a long-term solution is worked out.

jheinen:
@mkitti : I have no idea how to proceed now. Should we remove `dont_dlopen`?

mkitti:
The issues with `dont_dlopen` have already been resolved in the GR.jl. To be clear the issue was that the path to the libraries is not added to `LIBPATH_list` when `dont_dlopen` is true. See https://github.com/jheinen/GR.jl/issues/486#issuecomment-1302587583 for details.

1. https://github.com/jheinen/GR.jl/pull/483 resolved the issue by making sure that `LIBPATH_list` contains the directory holding the shared libraries with `dont_dlopen`. 
https://github.com/mkitti/GR.jl/blob/005b721753beb8e2af7dc6d2eb459128ad3fe466/src/preferences.jl#L59-L65
Eventually GR.jl uses this to set the PATH among other items. 
2. https://github.com/jheinen/GR.jl/pull/492 ensures that the environment variable `GRDIR` is set. This makes sure that GR can find its plugins: https://github.com/sciapp/gr/blob/9aa4f55f75101b76ba8e4406b78601e410642cc6/lib/gks/plugin.c#L50-L68 .

The summary is that we've already addressed the underlying issues with `dont_dlopen` on the master branch of GR.jl last week with some additional enhancements. We know that `dont_dlopen` will be problematic for GR.jl before v0.70.0. Versions of GR.jl before v0.70.0 did not declare compatibility with GR_jll, so that needs to be retroactively imposed.

giordano:
So you're saying that we can go on with this PR but, _in addition to it_, you want to set compat bounds?

mkitti:
I want to set
1. Set compat bounds first so that people who have a GR.jl released before v0.70 do not use this version of GR_jll. I'm just not sure if it should look like
a. https://github.com/JuliaRegistries/General/pull/71976 or
b.  https://github.com/JuliaRegistries/General/pull/71979
2. Release this GR_jll.jl as v0.70.0
3. Do additional testing on GR.jl
4. Release GR.jl v0.70.0 with a compat pin to GR_jll v0.70.0: https://github.com/jheinen/GR.jl/blob/028095d866878f5967810132bd8d6c82351f447b/Project.toml#L31

PR #5844, opened 2022-11-10
[Coin-OR] OpenBLAS version bump

ViralBShah:
@giordano We use gcc v6. Is that good for M1? Or should we be on something newer?

`MUMPS_seq` has a `julia_compat` of 1.8. I wonder if that will create a problem.

### DIFF
--- a/C/Coin-OR/coin-or-common.jl
+++ b/C/Coin-OR/coin-or-common.jl
@@ -76,7 +76,7 @@ SHOT_version = offset_version(v"1.1.0", v"0.0.0")
 Julia_compat_version = "1.6"
 ASL_version = v"0.1.3"
 METIS_version = v"5.1.2"
-MUMPS_seq_version = v"500.500.101"
+MUMPS_seq_version = v"5.4.1"
 OpenBLAS32_version = v"0.3.21"
 
 # These are the platforms we will build for by default, unless further
--- a/C/Coin-OR/coin-or-common.jl
+++ b/C/Coin-OR/coin-or-common.jl
@@ -18,7 +18,7 @@ function offset_version(upstream, offset)
 end
 
 # GCC version for building the whole system
-gcc_version = v"12"
+gcc_version = v"8"
 
 # Versions of various COIN-OR libraries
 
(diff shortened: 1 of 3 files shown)


amontoison:
It will probably be an issue because LBT is not working with Julia < 1.8 and Clp is still linked with `OpenBLAS32` and not LBT.
You can use `MUMPS_seq` v"5.4.1" to avoid issues.

C/Coin-OR/coin-or-common.jl
```diff
@@ -18,7 +18,7 @@ function offset_version(upstream, offset)
 end
 
 # GCC version for building the whole system
-gcc_version = v"6"
+gcc_version = v"12"
```
giordano, line 21: This won't work in any released version of Julia

ViralBShah: What's the highest I can go?

C/Coin-OR/coin-or-common.jl
```diff
@@ -76,7 +76,7 @@ SHOT_version = offset_version(v"1.1.0", v"0.0.0")
 Julia_compat_version = "1.6"
 ASL_version = v"0.1.3"
 METIS_version = v"5.1.2"
-MUMPS_seq_version = v"500.500.101"
```
amontoison, line 78: Can you add a `MUMPS_seq_version_LBT` such we keep track of this version?

ViralBShah: Not sure what you mean - like just another version that is called `MUMPS_seq_version_LBT` that is 500.500.101, but is not used anywhere?

amontoison: Yes, it's what I mean.
It is not used anywhere but we should use it in the `build_tarballs.jl` of `Ipopt_jll`.

ViralBShah:
@giordano Is it possible for someone to download the new libClp from this (before merging) and see if it still crashes? If so, no point merging, I suppose.

amontoison:
You can directly download the artifact after a build for information: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=23592&view=logs&j=5039a21e-38a3-5533-11b9-bc8aaa6b2712&t=bc9e26fe-39ad-5ebd-d02a-5d50b03cfddd&l=1156

ViralBShah:
I'm merging this. If the gcc version turns out to be too aggressive, we can make another quick release to bump it downwards.

PR #5846, opened 2022-11-10
Soapy bindings for BladeRF

zsoerenm:
This builds the SoapySDR bindings for BladeRF SDR.

Windows is excluded because it depends on `BladeRFHardwareDriver_jll`.

### DIFF
--- a/S/SoapyBladeRF/build_tarballs.jl
+++ b/S/SoapyBladeRF/build_tarballs.jl
@@ -30,6 +30,7 @@ make install
 if [[ "${target}" == *-apple-* ]]; then
     mv ${libdir}/SoapySDR/modules0.8/libbladeRFSupport.so ${libdir}/SoapySDR/modules0.8/libbladeRFSupport.dylib
 fi
+install_license $WORKSPACE/srcdir/SoapyBladeRF/LICENSE.LGPLv2.1
 """
 
 # These are the platforms we will build for by default, unless further
--- a/S/SoapyBladeRF/build_tarballs.jl
+++ b/S/SoapyBladeRF/build_tarballs.jl
@@ -1,0 +1,60 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg, Base.BinaryPlatforms
+
+name = "SoapyBladeRF"
+version = v"2.4.1"
+
+# Collection of sources required to complete build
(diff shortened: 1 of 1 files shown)


S/SoapyBladeRF/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+      -DCMAKE_BUILD_TYPE=Release \
+      ..
+
+make -j${nproc}
+make install
+if [[ "${target}" == *-apple-* ]]; then
+    mv ${libdir}/SoapySDR/modules0.8/libbladeRFSupport.so ${libdir}/SoapySDR/modules0.8/libbladeRFSupport.dylib
```
giordano, line 31: 1. I recommend using `install -Dvm 0755`
2. Just renaming the file is not enough, the soname is likely wrong

giordano (commented):
The licence file is missing

S/SoapyBladeRF/build_tarballs.jl
zsoerenm: 1. I'm sorry, I'm not a compile export. Does the first mean I shall replace `make install` with `install -Dvm 0755`?
What is the advantage?
2. To be honest I copied this line from the other SoapySDR bindings:
https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SoapyUHD/build_tarballs.jl#L32 &
https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SoapyRTLSDR/build_tarballs.jl#L31 etc

zsoerenm:
I added the license file.

S/SoapyBladeRF/build_tarballs.jl
zsoerenm: Alright I replaced
```
make install
    mv ${libdir}/SoapySDR/modules0.8/libbladeRFSupport.so ${libdir}/SoapySDR/modules0.8/libbladeRFSupport.dylib
fi
```
with
```
install -Dvm 755 "libbladeRFSupport.so" "${libdir}/SoapySDR/modules0.8/libbladeRFSupport.${dlext}"
```
Is that correct?

zsoerenm:
I think this ready to be merged.

PR #5862, opened 2022-11-15
[DynarePreprocessor] Update to v6.1.0

MichelJuillard:
New version of Dynare preprocessor source  and patches for clang


MichelJuillard:
For some weird reason in aarch64-apple-darwin SDK 11.3 doesn't know map::contains but other apple-darwin architectures do as well as FreeBSD. I will deal with it tomorrow

giordano:
Maybe because https://github.com/JuliaPackaging/Yggdrasil/blob/781206cac2788ebcdc43df01904fa4f6b256cb97/D/DynarePreprocessor/build_tarballs.jl#L31 says 11.1? That setting is important for declaring compatibility.

MichelJuillard:
Thanks @giordano !

ViralBShah:
As someone who studied macroecon at one point, I'm excited to see Dynare becoming first class in Julia! Thanks @MichelJuillard.

MichelJuillard:
Thanks @ViralBShah ! We have still a way to go :)

PR #5872, opened 2022-11-17
Wizard recipe: PHYLIP-v3.697.0

cossio:

* Package name: PHYLIP
* Version: v3.697.0


fingolfin:
There is a good chance that the build instructions for Linux will also work on FreeBSD and macOS -- those .dmg files for macOS are about precompiled binaries, so not relevant here.

What I usually do in this situation is to just turn on the additional platforms, and then see what happens. If it works, great; if not, just turn them back off again.

P/PHYLIP/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd phylip-*/src/
+make -f Makefile.unx install
+mkdir -p $bindir $libdir
+mv ../exe/libdrawtree.so $libdir/
+mv ../exe/libdrawgram.so $libdir/
```
staticfloat, line 20: Are these actually called `.so` on other platforms (like macOS)?  If not, you can use `${dlext}` as a convenient bash variable to get `.so`, `.dylib` and `.dll` as appropriate.

fingolfin: No, better to look at the actual error I'd say. Eg for macos:
`[18:26:03] bash: line 1: /Developer/Tools/Rez: No such file or directory`

PR #5893, opened 2022-11-17
Add ittapi

vchuravy:
Following https://github.com/mchristianl/IntelITT.jl

### DIFF
--- a/I/ittapi/build_tarballs.jl
+++ b/I/ittapi/build_tarballs.jl
@@ -16,9 +16,8 @@ cd $WORKSPACE/srcdir/ittapi/
 cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DITT_API_IPT_SUPPORT=1
 make
 ar -x bin/libittnotify.a
-cc -shared ittnotify_static*.o ittptmark*.o -o libittnotify.so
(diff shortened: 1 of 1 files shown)


I/ittapi/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/ittapi/                                                                                                                                                  
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DITT_API_IPT_SUPPORT=1                                      
+make                                                                                                                                                                          
+ar -x bin/libittnotify.a                                                                                                                                                      
+cc -shared ittnotify_static*.o ittptmark*.o -o libittnotify.so                                                                                                                
+mkdir -p $prefix/lib                                                                                                                                                          
+cp libittnotify.so $prefix/lib/
```
fingolfin, line 21: To fix macOS, perhaps this will work?
```suggestion
cc -shared ittnotify_static*.o ittptmark*.o -o libittnotify.${dlext}                                                                                                                
mkdir -p $prefix/lib                                                                                                                                                          
cp libittnotify.${dlext} $prefix/lib/                                                                                                                                               
```

giordano: I prefer using install -Dvm 0755 for moving the file to the prefix, which enforces the rights and saves one command

PR #5897, opened 2022-11-19
Fix the version of METIS4_jll in MUMPS4

amontoison:


ViralBShah:
This needs more work to move MUMPS 4 to build a shared library (and also change the filenames). Perhaps best to leave it as is?

amontoison:
I forgot that METIS4_jll was `BuildDependency` and not a `Dependency` of MUMPS4 and SDPA when I opened this PR and https://github.com/JuliaPackaging/Yggdrasil/pull/5898, which means that we can still use the old version of `METIS4_jll` for new builds of these artifacts.

It depends if we want shared libraries of MUMPS4 or not now.

PR #5909, opened 2022-11-23
[SLICOT] Another attempt to work around mingw/BB/LBT problems

RalphAS:
Following in the footsteps of Ipopt, we give up on LBT under Windows for the time being.
Since LBT on Julia v1.7 is a minefield (no versioning) we require v1.8+.
We'll hope that the BB/Pkg guys do some magic before LBT jumps a major version. But if we're in the same boat with a more prominent package, we may have a better chance at rescue otherwise.

This PR is intended to address #4969


amontoison:
> Following in the footsteps of Ipopt, we give up on LBT under Windows for the time being. Since LBT on Julia v1.7 is a minefield (no versioning) we require v1.8+. We'll hope that the BB/Pkg guys do some magic before LBT jumps a major version. But if we're in the same boat with a more prominent package, we may have a better chance at rescue otherwise.

You need to link with OpenBLAS_jll and not OpenBLAS32_jll on Windows here.
SLICOT is compiled with an ILP64 BLAS/LAPACK.
I have an example here: https://github.com/JuliaPackaging/Yggdrasil/blob/master/P/PROPACK/build_tarballs.jl#L19

S/SLICOT/build_tarballs.jl
```diff
@@ -96,21 +105,6 @@ install_license ../LICENSE
...
-function set_julia_version(platforms::Vector{Platform}, julia_version::VersionNumber)
-    _platforms = deepcopy(platforms)
-    for p in _platforms
-        p["julia_version"] = string(julia_version)
-    end
-    return _platforms
-end
-expand_julia_versions(platforms::Vector{Platform}, julia_versions::Vector{VersionNumber}) =
-    vcat(set_julia_version.(Ref(platforms), julia_versions)...)
-platforms = expand_julia_versions(platforms, julia_versions)
```
giordano, line 113: Why was this removed?

giordano: Oh, you're changing the required version of Julia.  But you can't do that without changing the version number of the package, otherwise you're breaking the previous build

giordano (changes requested):

giordano:
@andreasvarga would be ok with requiring Julia v1.8 for this?  It'd simplify a few things.

andreasvarga:
You mean Julia 1.8 and higher? Then I see no problem.

giordano (approved):

andreasvarga:
The MWE of [#47638](https://github.com/JuliaLang/julia/issues/47638) works now with Julia 1.8.1 and SLICOT_jll v5.8.1. Thanks everybody for your efforts.

andreasvarga:
I successfuly performed the complete tests with the [PeriodicSystems](https://github.com/andreasvarga/PeriodicSystems.jl) (see[ here](https://github.com/andreasvarga/PeriodicSystems.jl/actions/runs/3542049747/jobs/5947073410)). With Julia 1.7 under both Linux and Windows the v5.8.0+2 of SLICOT_jll was loaded, while for Julia 1.8 under  both Linux and Windows the v5.8.1+0 of SLICOT_jll was loaded. Perfect!

PR #5912, opened 2022-11-23
Add stork recipe

Seelengrab:
The wizard crashed ala https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1248, so I'm trying my luck in CI instead.

### DIFF
--- a/S/stork/build_tarballs.jl
+++ b/S/stork/build_tarballs.jl
@@ -1,0 +1,39 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, BinaryBuilderBase
+
+name = "stork"
+version = v"1.5.0"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/jameslittle230/stork/archive/refs/tags/v$(version).tar.gz",
+                  "4f9cf8dcbf3ebec83d64f34d7b1d9b39ce753e70d872b46d9a4a314bad10bf01"),
+]
+
+# Adapted from the justfile of the repo
+script = raw"""
+cd $WORKSPACE/srcdir/stork-*
+cargo build --release --all-features
+install -Dvm 0755 "target/${rust_target}/release/stork${exeext}" "${bindir}/stork${exeext}"
+install_license license.txt
+"""
+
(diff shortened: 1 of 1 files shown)


S/stork/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/jameslittle230/stork/archive/refs/tags/v1.5.0.tar.gz",
+                  "4f9cf8dcbf3ebec83d64f34d7b1d9b39ce753e70d872b46d9a4a314bad10bf01"),
+]
+
+# Adapted from the justfile of the repo
+script = raw"""
+cd $WORKSPACE/srcdir/stork-*
+cargo build --release --all-features
```
giordano, line 17: `cargo build` doesn't install anything.  You probably need to do something like
```suggestion
cargo build --release --all-features
install -Dvm 0755 "target/${rust_target}/release/stork${exeext}" "${bindir}/stork${exeext}"
```

S/stork/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, BinaryBuilderBase
+
+name = "stork"
+version = v"1.5.0"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/jameslittle230/stork/archive/refs/tags/v1.5.0.tar.gz",
```
giordano, line 10:
```suggestion
    ArchiveSource("https://github.com/jameslittle230/stork/archive/refs/tags/v$(version).tar.gz",
```

giordano:

giordano:
@Seelengrab let me know when you're happy :slightly_smiling_face:

Seelengrab:
Alright, this ought to work now! The cli binary itself seems to work. The only thing that may change is whether this recipe can also provide the corresponding JS/WASM library (which can be built as well) for direct serving of web pages. The README of stork mentions that they'd be under `dist`, with `stork.js` and `stork.wasm` respectively. I don't know how to make that available for e.g. a statically generated site that isn't served via julia, but I guess I could add them as artifacts as well?

PR #5928, opened 2022-11-28
Wizard recipe: TMscore-v1.0.0

cossio:

* Package name: TMscore
* Version: v1.0.0

### DIFF
--- a/T/TMscore/build_tarballs.jl
+++ b/T/TMscore/build_tarballs.jl
@@ -22,6 +22,7 @@ install_license LICENSE_1_0.txt
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
 
 # The products that we will ensure are always built
 products = [
--- a/T/TMscore/build_tarballs.jl
+++ b/T/TMscore/build_tarballs.jl
@@ -7,13 +7,12 @@ version = v"1.0.0"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


T/TMscore/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "TMscore"
+version = v"1.0.0"
+
+# Collection of sources required to complete build
+sources = [
+    FileSource("https://seq2fun.dcmb.med.umich.edu//TM-score/TMscore.cpp", "30274251f4123601af102cf6d4f1a9cc496878c1ae776702f554e2fc25658d7f"),
```
giordano, line 10: I presume that when there is a new version they replace this file in-place, making this recipe non-reproducible, right?

T/TMscore/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("TMscore", :TMscore)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
```
giordano, line 33:
```suggestion
dependencies = [
    Dependency("CompilerSupportLibraries_jll"; platforms=filter(p -> Sys.islinux(p) || Sys.isfreebsd(p), platforms)),
```

T/TMscore/build_tarballs.jl
giordano: Sometimes people put the code in a repository and update the source in the repository when needed, at least we get a more reproducible source (as long as the repository isn't deleted)

PR #5930, opened 2022-11-29
[CMake] Update to v3.23.3

tylerjthomas9:
`cmake` v3.23.1+ is required to build cuDF (https://github.com/rapidsai/cudf/blob/branch-22.10/CONTRIBUTING.md#setting-up-your-build-environment).

There are newer versions of CMake (v3.24, v3.25), but I just bumped the version to the minimum required version.


giordano:
Note that this isn't used by default in the build environment (although it may work as host build tool)

PR #5933, opened 2022-11-29
[SuperLU_DIST] Correct ParMETIS dep

Wimmerer:
And use LBT.

### DIFF
--- a/S/SuperLU_DIST/build_tarballs.jl
+++ b/S/SuperLU_DIST/build_tarballs.jl
@@ -24,23 +24,27 @@ if [[ "${target}" == *-mingw* ]]; then
     # This is required to ensure that MSMPI can be found by cmake
     export LDFLAGS="-L${libdir} -lmsmpi"
     PLATFLAGS="-DTPL_ENABLE_PARMETISLIB:BOOL=FALSE -DMPI_C_ADDITIONAL_INCLUDE_DIRS=${includedir}"
-else
-    PLATFLAGS="-DTPL_PARMETIS_INCLUDE_DIRS=${includedir} -DTPL_PARMETIS_LIBRARIES=${libdir}/libparmetis.${dlext};${libdir}/libmetis.${dlext}"
 fi
 
 build_superlu_dist()
 {
     if [[ "${1}" == "Int64" ]]; then
         INT=64
+        METIS_PATH="${libdir}/metis/metis_Int64_Real32/lib/libmetis_Int64_Real32.${dlext}"
+        PARMETIS_PATH="${libdir}/libparmetis_Int64_Real32.${dlext}"
     else
         INT=32
+        METIS_PATH="${libdir}/libmetis.${dlext}"
+        PARMETIS_PATH="${libdir}/libparmetis.${dlext}"
     fi
-    SUPERLU_PREFIX=${libdir}/superlu_dist/Int${INT}
-    mkdir -p ${SUPERLU_PREFIX}
+    if [[ "${taret}" != *-mingw* ]]; then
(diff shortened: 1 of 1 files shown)


S/SuperLU_DIST/build_tarballs.jl
```diff
@@ -95,8 +99,8 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="OpenBLAS32_jll", uuid="656ef2d0-ae68-5445-9ca0-591084a874a2")),
-    Dependency(PackageSpec(name="PARMETIS_jll", uuid="b247a4be-ddc1-5759-8008-7e02fe3dbdaa"); platforms=filter(!Sys.iswindows, platforms)),
+    Dependency("libblastrampoline_jll"),
```
giordano, line 102: Consumers of this package will have to ensure that an LP64 Blas is available.  Is that the case?

Wimmerer: I just want to build against whatever Julia is using at that moment. I thought LBT was the way to do this?

Wimmerer: Should I force it to build against the 32 bit symbols of LBT somehow?

giordano: > I just want to build against whatever Julia is using at that moment.

That's not how it works.  Your library makes call for a specific interface, either LP64 or ILP64, you must make sure that at runtime there is a backing BLAS library to receive those calls, otherwise libblastrampoline will forward calls into the void (well, it'll throw an error about no BLAS library available)

Wimmerer: LBT *provides* both symbols though, correct? So it's a SuperLU build flag I just need to switch to 32-bit?

Wimmerer: It's detected automatically for known implementations. It doesn't know LBT so I'll pass a flag.

Wimmerer: I'm checking with Sherry but it should always be LP64 I believe, I was thinking of SuiteSparse that could be both. So this should be fine with LBT right? Is there a problem with PETSc_jll building against OpenBLAS32 and this building against LP64 LBT?

giordano: > So this should be fine with LBT right? 

LBT isn't the problem, it provides interface for both LP64 and ILP64 :slightly_smiling_face:  the problem is that with this PR you're removing the BLAS library providing the implementation for the needed function, which would create a runtime disaster.

Wimmerer: I can't read 💀 . Julia provides ***I***LP64, so we still need OpenBLAS32 💀 now I get it.

Wimmerer: I see a lot of swapping back to OpenBLAS for mingw. Is that still necessary?

giordano: Yes, it's a mess.

giordano: If you restored OpenBLAS32 here then you don't need?

giordano: There is no default OpenBLAS32.  Consumers will need to load it explicitly, and not JLL packages, otherwise you're forcing to load OpenBLAS32_jll when consumers may want to use something else.  This is the whole point of [my question](https://github.com/JuliaPackaging/Yggdrasil/pull/5933#discussion_r1035436129) :slightly_smiling_face:

Wimmerer: PETsc_jll is already taken care of in that regard I think. So at the Julia package (SuperLU_DIST) level I can take care of this then? Load LBT, if it doesn't have LP64 config I can `using OpenBLAS32_jll`?

PR #5934, opened 2022-11-30
Update libjulia for latest master

fingolfin:
[Julia PR #44527](https://github.com/JuliaLang/julia/pull/44527) changes how foreign types work. To accommodate this, we need a new libjulia_jll

I think this PR is meant to be backported to the release-1.9 branch. At that point, another update of libjulia_jll will be needed. But I am not sure when this backport will happen, and so I'd like to just move ahead with this now, so that I can un-break our CI tests against Julia nightly.

BTW, I've updated the version for libjulia_jll here. In the past we didn't always do that. I am not really sure if I am "supposed" to increment the version, or supposed to not do that and just add a whitespace to force a rebuild... I don't mind either way... *shrug*


vchuravy:
I would strongly prefer the jll version to match the actual Julia version... otherwise things get confusing.

vchuravy:
Oh right... Morning brain

fingolfin:
Perhaps it should be  the maximum of all versions, so here still 1.10 ? Whatever we agree on (or agreeD ON -- I probably just forgot the outcome) I suggest we add it as a comment to avoid repeating the discussion needlessly:-)

fingolfin:
Hrm, 1.9 fails in mingw, and 1.10 fails everywhere:
```
[11:59:22] make[2]: *** No rule to make target '/workspace/srcdir/julia/src/support/hashing.o', needed by '/workspace/srcdir/julia/src/support/libsupport.a'.  Stop.
[11:59:22] make[2]: Leaving directory '/workspace/srcdir/julia/src/support'
[11:59:22] make[1]: *** [Makefile:329: support/libsupport.a] Error 2
[11:59:22] make[1]: Leaving directory '/workspace/srcdir/julia/src'
[11:59:22] make: *** [Makefile:62: julia_flisp.boot.inc.phony] Error 2
[11:59:22] make: *** Waiting for unfinished jobs....
```

fingolfin:
OK, I went one commit back, leaving out Julia commit https://github.com/JuliaLang/julia/commit/41b73e2c71ca844a372af2b75325940cfcb681b6 ("Bump libuv"). That resolves the 1.10 builds for now.

That leaves the failing mingw builds.

fingolfin:
> Once again it is that msys patch that is not applied since the julia version changed from 1.9.0-DEV to 1.9.0-alpha1

Urgh, one would think I'd have learned that by now :-(. Thank you, trying that now.

Assuming this works, I'd like to merge and then try the "remove fake uname" trick in a separate PR (it would be nice to not require those patches)

PR #5935, opened 2022-11-30
libjulia: experiment with uname hack

fingolfin:
... as suggested by @benlorenz


L/libjulia/common.jl
```diff
@@ -82,6 +82,12 @@ function build_julia(ARGS, version::VersionNumber; jllversion=version)
     script = raw"""
     apk add coreutils libuv-dev utf8proc
 
+    # WORKAROUND for mingw: remove the fake `uname` binary, it throws off the
+    # Julia buildsystem
```
fingolfin, line 86: @benlorenz if you can suggest a better description, that'd be great (I must admit I do not fully understand the problem, and honestly am happy with that right now ;-))

PR #5939, opened 2022-12-01
Bump Xyce to 7.6

Keno:


giordano:
ICE on `aarch64-linux-musl-cxx03`, we've seen it quite often (and had to bump GCC version because of that...).

Keno:
> ICE on `aarch64-linux-musl-cxx03`, we've seen it quite often (and had to bump GCC version because of that...).

Is there recommended gcc version to use?

giordano:
I usually bump to the next one.

PR #5945, opened 2022-12-04
Add libadwaita recipe

jwahlstrand:
Most of the recipe is copied from GTK4's.

### DIFF
--- a/L/libadwaita/build_tarballs.jl
+++ b/L/libadwaita/build_tarballs.jl
@@ -6,7 +6,7 @@ name = "libadwaita"
 version = v"1.2.0"
 
 sources = [
-    ArchiveSource("https://download.gnome.org/sources/libadwaita/1.2/libadwaita-1.2.0.tar.xz",
+    ArchiveSource("https://download.gnome.org/sources/libadwaita/$(version.major).$(version.minor)/libadwaita-$(version).tar.xz",
                   "322f3e1be39ba67981d9fe7228a85818eccaa2ed0aa42bcafe263af881c6460c"),
 ]
 
--- a/L/libadwaita/build_tarballs.jl
+++ b/L/libadwaita/build_tarballs.jl
@@ -46,7 +46,7 @@ glib-compile-schemas ${prefix}/share/glib-2.0/schemas
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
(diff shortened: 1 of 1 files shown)


L/libadwaita/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+ninja -j${nproc}
+ninja install
+
+# post-install script is disabled when cross-compiling
+glib-compile-schemas ${prefix}/share/glib-2.0/schemas
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = filter!(p -> arch(p) != "armv6l", supported_platforms(; experimental=true))
```
giordano, line 49:
```suggestion
platforms = filter!(p -> arch(p) != "armv6l", supported_platforms())
```

L/libadwaita/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+rm ${prefix}/lib/pkgconfig/gio-2.0.pc
+
+# oldest version that works, due to use of "@available" macro
+export MACOSX_DEPLOYMENT_TARGET=10.14
+
+cd libadwaita*
+
+mkdir build-libadwaita && cd build-libadwaita
+meson .. \
+    --buildtype=plain \
```
giordano, line 33: Why plain and not release?

jwahlstrand: I can't remember how I decided to try `buildtype=plain`, since I started this months ago and picked it up again today. But on i686-w64-mingw32, when I use `buildtype=release`, I'm getting:
```
[123/123] Linking target src/libadwaita-1-0.dll
[19:44:21] ninja: job failed: ...
[19:44:21] ninja: subcommand failed
[19:44:21] src/libadwaita-1-0.dll.p/adw-leaflet.c.obj:adw-leaflet.c:(.text+0xca5): undefined reference to `__stack_chk_guard'
```

giordano: Sigh.  Ok, I don't know what to do with that (I presume we need to link some mingw libraries, but don't know which), so let's stick to plain for the time being.

jwahlstrand: Thanks

PR #5946, opened 2022-12-05
Enable builds for MacOS by not explicitly requiring glibc

torrance:
We want to require only glibc build, as Casacore itself uses glibc specific calls. However, for some reason MacOS platforms have no libc information at all and by requiring `libc(p) == "glibc"` we exclude all MacOS builds.

In this patch, we will require `libc(p) != "musl"` (which is consistent with `casacore_jll`).

### DIFF
--- a/C/casacorecxx/build_tarballs.jl
+++ b/C/casacorecxx/build_tarballs.jl
@@ -34,7 +34,7 @@ platforms = expand_cxxstring_abis(platforms)
 
 # Filter this list based on the same filter criteria used in the casacore build script
 filter!(platforms) do p
-    !Sys.iswindows(p) && !Sys.isfreebsd(p) && libc(p) == "glibc"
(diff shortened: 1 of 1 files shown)


torrance:
Also: I haven't bumped the version number as this only increases the platforms we're building for, not the underlying library. Please let me know if there's a proper way for handling a change like that.

giordano:
> However, for some reason MacOS platforms have no libc information at all and by requiring libc(p) == "glibc" we exclude all MacOS builds.

Well, even if libc(p) wasn't nothing for macOS it wouldn't certainly be glibc, since that's ***not*** the libc used by macOS, so the previous filter was wrong anyway 🙂

torrance:
> Well, even if libc(p) wasn't nothing for macOS it wouldn't certainly be glibc, since that's not the libc used by macOS, so the previous filter was wrong anyway

I'm going to proceed with this PR anyway, since it will be consistent with `casacore_jll` and I am aware of Casacore successfully building on macOS.

PR #5959, opened 2022-12-08
[RDKit] 2022.09.3 update

eloyfelix:


eloyfelix:
@giordano did it really failed or one of the jobs just got kind of stuck? Shall we try to re-run the tests?

PR #4205, opened 2022-01-08
Update HELICS recipe to v3.1.2

kdheepak:

PR #3644, opened 2021-09-18
[HELICS] update to v3.0.0

nightlark:
Adds a HELICS package for v3.0.0

If needed this one can be skipped and just v3.0.1 could be added -- though there are new maintenance releases being made in the 2.8.x series for bug fixes that will still need updating somehow. Is there a way to specify multiple versions or just open a new PR to Yggdrasil for each individual version if the result is not a monotonically increasing package version in the Yggdrasil commit history?

PR #5967, opened 2022-12-09
[HELICS] update to v3.3.2

nightlark:
This PR supercedes #3644, #3645, and #4205.

Changes:
- Get HELICS package working again for the latest HELICS 3.x release (3.3.2)
- Use gcc v9 to address bug with mingw std::filesystem library on older versions of gcc
- Use MacOSX 10.15 SDK on Intel mac builds to address issues with std::filesystem library on older versions of macOS
- Enable FreeBSD builds (CI system for HELICS has tested FreeBSD for some time now)
- Remove build support for HELICS 2.x, which is no longer supported/receiving updates

Squash/merge commit message should include the line (at the end of the commit message):
`Co-authored-by: Dheepak Krishnamurthy <me@kdheepak.com>`

### DIFF
--- a/H/HELICS/build_tarballs.jl
+++ b/H/HELICS/build_tarballs.jl
@@ -47,7 +47,7 @@ products = [
 ]
 
 
-platforms = expand_cxxstring_abis(supported_platforms(exclude=Sys.isfreebsd))
+platforms = expand_cxxstring_abis(supported_platforms())
 
 dependencies = [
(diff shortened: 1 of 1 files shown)


nightlark:
@giordano I think I got working packages building on all supported platforms now. Any changes or better ways of doing things that I should make?

giordano:
Thanks a lot, it looks all good!

PR #5969, opened 2022-12-09
[xz] Bump version to 5.2.9

jeremiahpslewis:


giordano:
https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=24074&view=logs&j=202d3545-3ddc-5ea1-ddc7-275fc05b0971&t=aed1f72d-f13c-5267-49c9-e7e20670b5d8&l=804
```
[22:01:33] configure: error: 
[22:01:33]     On GNU/Linux, building both shared and static library at the same time
[22:01:33]     is not supported if --with-pic or --without-pic is used.
[22:01:33]     Use either --disable-shared or --disable-static to build one type
[22:01:33]     of library at a time. If both types are needed, build one at a time,
[22:01:33]     possibly picking only src/liblzma/.libs/liblzma.a from the static build.
```
well, that's stupid, and we ***do*** need both of them because libunwind does static linking of liblzma: https://github.com/JuliaPackaging/Yggdrasil/blob/cd1eba6a18d92609022dfa22714dc3bb40731e3f/L/LibUnwind/LibUnwind%401.5.0/build_tarballs.jl#L45-L50

jeremiahpslewis:
@giordano Thanks for fishing out the error, I ran this last night as 'fire and forget'. ;)

jeremiahpslewis:
Maybe next year we should consider an 'advent of Yggdrasil' for bumping versions of well maintained binaries! 🎄

PR #5973, opened 2022-12-11
[FLINT] Update to 2.9.0 + something

thofma:
@fingolfin @benlorenz

There have been some important bugfixes and we finally can land https://github.com/Nemocas/Nemo.jl/pull/1342.

I have to do some testing first, but maybe you could already confirm that the version is correct.

@tthsqe12 The TLS patch is not necessary anymore, right?


F/FLINT/build_tarballs.jl
```diff
@@ -22,19 +22,18 @@ using BinaryBuilder, Pkg
 #
 
 # WARNING WARNING WARNING: any change to the the version of this JLL should be carefully
-# coordinated with corresponding changes to Singular_jll.jl, LoadFlint.jl, Nemo.jl,
+# coordinated with corresponding changes to Singular_jll.jl, Nemo.jl and polymake_jll
 # and possibly other packages.
 name = "FLINT"
 upstream_version = v"2.9.0"
-version_offset = v"0.0.1"
+version_offset = v"0.0.2"
```
fingolfin, line 29: So is the new JLL version binary / ABI compatible with the previous version?

thofma: Yes, I think so. Can you confirm @tthsqe12?

tthsqe12: 1. The main thing is
https://github.com/flintlib/flint2/commit/4efd13eb180c4ae796ff766a321494c2d8b9dc56
which *adds* a bunch of new symbols.

2. We also have
https://github.com/flintlib/flint2/commit/71b744ace263b7a8253b0903c3569d4cb8083d29
which also only add symbols

3. There is
https://github.com/flintlib/flint2/commit/6aa8103e508c6b2dad072c762431c12a36557745
which means the TLS patch is not necessary anymore

PR #5996, opened 2022-12-13
[Xgboost] fix gpu build target logic

tylerjthomas9:
This fixes the GPU build that was merged earlier today (https://github.com/JuliaPackaging/Yggdrasil/pull/5892). The logic for determining the build always built the CPU version of the library. I have updated it to use `$bb_full_target` instead of `$target`, so we can correctly build the CUDA version. 

The only thing this should change is that the x64_linux + CUDA build should now have GPU support. 

I have also removed NCCL (for the time being) because it will require extra work. The no issues that I experienced were due to the incorrect target logic, not the CUDA+NCCL build working.

@ExpandingMan

### DIFF
--- a/X/XGBoost/build_tarballs.jl
+++ b/X/XGBoost/build_tarballs.jl
@@ -22,7 +22,7 @@ git submodule update --init
 (cd dmlc-core; atomic_patch -p1 "../../patches/dmlc_windows.patch")
 
 mkdir build && cd build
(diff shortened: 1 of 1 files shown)


tylerjthomas9:
This build script should be good to go. I apologize for the errors in the last one that caused every build to just be a CPU build. I took some stuff from the Torch_jll build to simplify this build script.

PR #6003, opened 2022-12-15
New recipe: sdl12-compat

aviks:
https://github.com/libsdl-org/sdl12-compat/

### DIFF
--- a/S/SDL12_compat/build_tarballs.jl
+++ b/S/SDL12_compat/build_tarballs.jl
@@ -1,0 +1,43 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "SDL12_compat"
+version = v"1.2.60"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/libsdl-org/sdl12-compat.git", "43887001c05c64eb23eb5312c0f62e9841618700")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd sdl12-compat/
+mkdir build
+cd build/
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DSDL12TESTS=OFF -DSDL12DEVEL=ON ..
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms() 
+
+x11_platforms = filter(p -> Sys.islinux(p) || Sys.isfreebsd(p), platforms)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct(["SDL", "libSDL"], :libSDL)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="SDL2_jll", uuid="ab825dc5-c88e-5901-9575-1e5e20358fcf"))
+    BuildDependency(PackageSpec(name="Xorg_xorgproto_jll", uuid="c4d99508-4286-5418-9131-c86396af500b");
+		     platforms=x11_platforms)
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
--- a/S/SDL12_compat/build_tarballs.jl
+++ b/S/SDL12_compat/build_tarballs.jl
@@ -40,4 +40,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", dont_dlopen=true)


giordano:
https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=24176&view=logs&j=245b2b2f-7f8c-5a58-5514-778d8a7014da&t=adea98ce-e913-5d07-f635-1543f91218a8&l=514
```
[ Info: /agent/_work/1/s/S/SDL12_compat/build/x86_64-linux-gnu/fMfxV97i/x86_64-linux-gnu-libgfortran3-cxx03/destdir/lib/libSDL.so matches our search criteria of libSDL
[ Info: /agent/_work/1/s/S/SDL12_compat/build/x86_64-linux-gnu/fMfxV97i/x86_64-linux-gnu-libgfortran3-cxx03/destdir/lib/libSDL.so cannot be dlopen'ed
```

giordano:
So, it's failing around https://github.com/libsdl-org/sdl12-compat/blob/e8de4b55eff752cd5fa80b37b1d1e4261bf11932/src/SDL12_compat.c#L1396-L1398, `LoadSDL20Library` on Linux is https://github.com/libsdl-org/sdl12-compat/blob/e8de4b55eff752cd5fa80b37b1d1e4261bf11932/src/SDL12_compat.c#L1154.  The worst part is that it works locally for me (but it may be picking up system libSDL2, if that's any better) :frowning_face:

giordano:
Also, why doing dynamic linking when you can dlopen the library at `dllinit` time?  :roll_eyes:

aviks:
Could it be because it wants the library name to be fully version
qualified: `libSDL2-2.0.so.0` ... Wondering if our sdl2 build does that.

aviks:
Looks like our SDL has the so's in the format this requires: 
image

aviks:
So doing this works 

```
julia> using SDL2_jll

julia> Libc.Libdl.dlopen("libSDL2-2.0.so.0")
Ptr{Nothing} @0x0000000000ca3a80
```

I guess JLLWrappers sets some library search paths to enable that code to work? Presumably that has not been done when the problematic C code is being run? Or should it be -- `SDL2_jll` is defined as  a dependency of this library.

aviks:
And yeah, apparently SDL2 is widely installed, and so it's easy to pick up a local installation: (the following is in a new Julia session)

```
julia> h = Libc.Libdl.dlopen("libSDL2-2.0.so.0")
Ptr{Nothing} @0x0000000000a82bf0

julia> Libc.Libdl.dlpath(h)
"/usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0"
```

aviks:
So I tried this on a machine without SDL2 installed, and it seems to have worked: 

```
julia> using SDL12_compat_jll
[ Info: Precompiling SDL12_compat_jll [7322c9eb-3a94-563c-b3a7-b438160070c5]

julia> h = Libc.Libdl.dlopen("libSDL2-2.0.so.0")
Ptr{Nothing} @0x00000000018f7750

julia> Libc.Libdl.dlpath(h)
"/home/aviks/.julia/artifacts/8f20658b625a577cdc6b81980bbae941cb291139/lib/libSDL2.so"
```

Even on a machine *with* SDL2 natively installed, `using SDL12_compat` causes subsequent dlopens to correctly use the version in the artifact, as you'd expect. 

So not really sure why this is failing on the build machine!

giordano:
[`LoadSDL20Library()`](https://github.com/libsdl-org/sdl12-compat/blob/e8de4b55eff752cd5fa80b37b1d1e4261bf11932/src/SDL12_compat.c#L1154) dlopen's by soname, not full absolute path as you're doing, so it needs to be able to find libsdl2 in the default search paths, which don't include the current directory.  Also,
```console
% readelf -d libSDL-1.2.so 

Dynamic section at offset 0x2ede0 contains 27 entries:
 0x000000000000000e (SONAME)             Library soname: [libSDL-1.2.so.0]
 0x000000000000000c (INIT)               0x5000
 0x000000000000000d (FINI)               0x24564
 0x0000000000000019 (INIT_ARRAY)         0x2fc70
 0x000000000000001b (INIT_ARRAYSZ)       8 (bytes)
 0x000000000000001a (FINI_ARRAY)         0x2fc78
 0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
 0x0000000000000004 (HASH)               0x200
 0x000000006ffffef5 (GNU_HASH)           0x910
 0x0000000000000005 (STRTAB)             0x2890
 0x0000000000000006 (SYMTAB)             0x10f0
 0x000000000000000a (STRSZ)              3772 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000003 (PLTGOT)             0x30000
 0x0000000000000002 (PLTRELSZ)           984 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x3e38
 0x0000000000000007 (RELA)               0x3988
 0x0000000000000008 (RELASZ)             1200 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffe (VERNEED)            0x3948
 0x000000006fffffff (VERNEEDNUM)         2
 0x000000006ffffff0 (VERSYM)             0x374c
 0x000000006ffffff9 (RELACOUNT)          44
 0x0000000000000000 (NULL)               0x0
```
it doesn't even set the rpath, which might help with finding the library in the same directory.

aviks:
My point was to show that `using SDL12_compat_jll` on a machine without a native install of SDL2 does work. Yes, I do a dlopen later to show that Julia knows about the correct path for the artifact. But the main point is that the `using` works.

aviks:
Also, I'm dlopen-ing with just the soname `Libc.Libdl.dlopen("libSDL2-2.0.so.0")` which is what the C code does. The `dlpath` function *returns* the aboslute path from the already open handle. Unless I'm misunderstanding you.

giordano:
Yeah, I was trying to explain why I believe the audit doesn't work.  The system likely doesn't have libsdl2, and the audit directly does `dlopen("libSDL-1.2.so")`, which would internally try to do `dlopen("libSDL2-2.0.so.0")`, failing to find libsdl2 because it's in the same directory, which isn't a valid search path.  That's at least my hunch.

S/SDL12_compat/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="SDL2_jll", uuid="ab825dc5-c88e-5901-9575-1e5e20358fcf"))
+    BuildDependency(PackageSpec(name="Xorg_xorgproto_jll", uuid="c4d99508-4286-5418-9131-c86396af500b");
+		     platforms=x11_platforms)
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
giordano, line 43: Can try skipping dlopen check during audit with
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", dont_dlopen=true)
```
and then hope it'll work correctly at runtime as you suggested :slightly_smiling_face:

aviks: Well, the thing seems to be working, so lets go!

```
julia> ccall( (:SDL_HasSSE, SDL12_compat_jll.libSDL), Cchar, ())
1

julia> ccall( (:SDL_Linked_Version, SDL12_compat_jll.libSDL), Ptr{Cvoid}, ())
Ptr{Nothing} @0x00007f5231b3ba0b
```

giordano:
Looks like it worked: https://github.com/JuliaRegistries/General/pull/74369 :partying_face:

PR #6004, opened 2022-12-16
Added TorchCAPI recipe

stemann:


Wimmerer:
I really think we should use: https://github.com/LaurentMazare/tch-rs/tree/main/torch-sys/libtch which is the upstream here.

stemann:
> I really think we should use: https://github.com/LaurentMazare/tch-rs/tree/main/torch-sys/libtch which is the upstream for that (kind of, the ocaml binding generator was used, the Rust version is the more recent version of _that_ generator).

I agree with the aim of switching to the libtch-rs wrapper, but due to the issue below I think it would be best to first get the wrapper updated (keeping the ocaml-torch-based wrapper), e.g. to get Torch.jl going on CPU, and CUDA 11.

There's a couple of issues to be aware of:
* The C wrapper used by Torch.jl was changed such that all functions return integer status codes, cf. https://github.com/FluxML/Torch.jl/pull/4
* It seems there is at least an aim/attempt to keep ocaml-torch and libtch-rs C wrappers in sync - but there are also quite a few differences between the two.

It's a bit of a mess right now: I actually used the tch-rs binding generator, but there were a bit too many changes from the handwritten ocaml-torch wrapper to the handwritten libtch-rs wrapper, that I felt comfortable switching to libtch-rs, so the handwritten part of the wrapper is based on the ocaml-torch one.

Wimmerer:
Bump?

stemann:
> Bump?

Thanks - trying to revive this...

stemann:
Please note the existence of [L/LibTorchCAPI](https://github.com/JuliaPackaging/Yggdrasil/tree/master/L/LibTorchCAPI) - though that seems like a more specific build/use - for [ThArrays.jl](https://github.com/data-efficient-ml/ThArrays.jl).

... and ThArrays has dropped use of LibTorchCAPI: https://github.com/compintell/THArrays.jl/pull/22

Attempt is ongoing to join forces: https://github.com/compintell/THArrays.jl/issues/11#issuecomment-2276543096

T/TorchCAPI/build_tarballs.jl
```diff
@@ -0,0 +1,114 @@
...
+configure() {
+    cmake \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_INSTALL_PREFIX=$prefix \
+        -DCMAKE_TOOLCHAIN_FILE=$CMAKE_TARGET_TOOLCHAIN \
+        $cmake_extra_args \
+        -S . \
+        -B build
+}
+configure || configure || configure
```
giordano, line 48: Sigh.

T/TorchCAPI/build_tarballs.jl
```diff
@@ -0,0 +1,114 @@
...
+
+if [[ $bb_full_target == aarch64-linux-gnu-*-cxx03 ]]; then # Somehow CMake ends up adding '-D_GLIBCXX_USE_CXX11_ABI=' to CXX_FLAGS
+    sed -i 's/ -D_GLIBCXX_USE_CXX11_ABI= / -D_GLIBCXX_USE_CXX11_ABI=0 /' build/CMakeFiles/torch_c_api.dir/flags.make
+fi
+
+cmake --build build -- -j $nproc
+
+install -Dvm 755 build/libtorch_c_api.$dlext $libdir/libtorch_c_api.$dlext
+
+for header_path in $(ls -1 *.h | grep -v cpp); do
```
giordano, line 58: https://mywiki.wooledge.org/ParsingLs

T/TorchCAPI/build_tarballs.jl
stemann: Yeah - it’s just the same silly hack for getting a CUDA 10.2 / 11.3 cmake configure through :-)

Will get rid of that eventually.

T/TorchCAPI/build_tarballs.jl
stemann: Just a hack to get around a “.cpp.h”-file which should not be copied - it’s implementation.

giordano: I was suggesting not to use `ls`.

stemann: Yeah, I know - but not sure what would be better here? Just get rid of the loop and call install twice?

There are just these specific .h-files present:
```
torch_api.h
torch_api_generated.h
```
and then the silly `torch_api_generated.cpp.h`; Which I will fix in the next iteration in [deps/c_wrapper_generator/bin/main.ml#L282](https://github.com/FluxML/Torch.jl/blob/master/deps/c_wrapper_generator/bin/main.ml#L282)

stemann: Can we go with the lazy/easy solution here (no change - cf. edit 2)?

PR #6005, opened 2022-12-16
Add recipe for libboxer

VarLad:


VarLad:
@giordano Any idea how to fix the Windows issue here? ^^;
Maybe Rust changed folder naming internally and its not yet updated in BinaryBuilder? 
Also, sorry for the messy commit history (but I hope squash merge can save me there) ^^;

L/libboxer/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+]
+
+# Adapted from the justfile of the repo
+script = raw"""
+cd $WORKSPACE/srcdir/libboxer*
+cargo build --release
+install -Dvm 0755 "target/${rust_target}/release/libBoxer.${dlext}" "${libdir}/libBoxer.${dlext}"
+install_license LICENSE
+""" * """
+curl "https://github.com/feenkcom/libboxer/releases/download/v$(version)/boxer.h" --output """ * raw""" "${includedir}/boxer.h" --create-dirs"""
```
giordano, line 21: Add this as additional source, don't download files

L/libboxer/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+sources = [
+    ArchiveSource("https://github.com/feenkcom/libboxer/archive/refs/tags/v$(version).tar.gz",
+                  "57d25e5339b0c7b1d8c8242484ce2c4baca05fbea035c8e2d8aab05573eed0c9"),
+]
+
+# Adapted from the justfile of the repo
+script = raw"""
+cd $WORKSPACE/srcdir/libboxer*
+cargo build --release
+install -Dvm 0755 "target/${rust_target}/release/libBoxer.${dlext}" "${libdir}/libBoxer.${dlext}"
```
giordano, line 18: I can't possibly know what's the content of the directory, I'm away from computer and will be so for a while, but why don't you simply add some ls commands to see what's inside the directory? Also, I already suggested you on slack to use [github codespaces](https://docs.binarybuilder.org/stable/#Github-Codespaces) which is now free for everyone, with with which you can interactively debug the builds yourself instead of several small commits to get this working.

VarLad: Ah, good idea! Let me try!

VarLad:
Ready to merge! (if nothing seems wrong!) :)

L/libboxer/build_tarballs.jl
VarLad: Rust didn't work out for me on codespaces either (for a different reason though) :sweat_smile:

VarLad:
@giordano Is there an important reason why Windows binaries are named Boxer.dll instead of libBoxer.dll? (is it some convention?) 
I renamed it to libBoxer.dll when copying it to $libdir, hoping that won't give an issue

giordano:
> @giordano Is there an important reason why Windows shared libraries are named Boxer.dll instead of libBoxer.dll? (is it some convention?)

No, names are chose by build systems, we have no influence over that, so you'll have to look at the build system.

L/libboxer/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+platforms = supported_platforms(; exclude=p -> (Sys.iswindows(p) && arch(p) == "i686") || (Sys.islinux(p) && libc(p) == "musl") || Sys.isbsd(p))
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libBoxer", :libBoxer),
+    FileProduct("include/boxer.h", :boxer_h),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
```
giordano, line 41:
```suggestion
dependencies = Dependency[
```

L/libboxer/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+
+name = "libboxer"
+version = v"0.2.0"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/feenkcom/libboxer/archive/refs/tags/v$(version).tar.gz",
+                  "57d25e5339b0c7b1d8c8242484ce2c4baca05fbea035c8e2d8aab05573eed0c9"),
+    FileSource("https://github.com/feenkcom/libboxer/releases/download/v$(version)/boxer.h", 
+                  "f7e679d2faddca7a99399a03a1fa21c49c0780cf2b73be2c4dd2f70d7a963637"),
```
giordano, line 14:
```suggestion
               "f7e679d2faddca7a99399a03a1fa21c49c0780cf2b73be2c4dd2f70d7a963637"),
```

PR #6009, opened 2022-12-19
[finufft] Support for AVX512 platforms

jkrimmer:
Currently, FINUFFT.jl does not work on the new Zen 4 Ryzen series since there is no binary for AVX512 platforms available in finufft_jll. Hence, disable AVX512 via CFLAGS on x86 to provide binaries for AVX512 platforms such as Zen 4 without introducing the performance issues mentioned in the comment of line 38.

### DIFF
--- a/F/finufft/build_tarballs.jl
+++ b/F/finufft/build_tarballs.jl
@@ -21,6 +21,9 @@ CFLAGS="-fopenmp -fPIC -O3 -funroll-loops -Iinclude"
 if [[ "${target}" != *-freebsd* ]] && [[ "${target}" != *-apple-* ]]; then
     CFLAGS="${CFLAGS} -fcx-limited-range"
 fi
+if [[ "${proc_family}" == *intel* ]]; then
+    CFLAGS="${CFLAGS} -mno-avx512f"
+fi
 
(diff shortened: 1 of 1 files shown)


F/finufft/build_tarballs.jl
```diff
@@ -36,7 +39,7 @@ install -Dvm 0755 "lib/libfinufft.${dlext}" "${libdir}/libfinufft.${dlext}"
 
 # Expand for microarchitectures on x86_64 (library doesn't have CPU dispatching)
 # Tests on Linux/x86_64 yielded a slow binary with avx512 for some reason, so disable that
-platforms = expand_cxxstring_abis(expand_microarchitectures(supported_platforms(), ["x86_64", "avx", "avx2"]); skip=!Sys.iswindows)
+platforms = expand_cxxstring_abis(expand_microarchitectures(supported_platforms(), ["x86_64", "avx", "avx2", "avx512"]); skip=!Sys.iswindows)
```
giordano, line 42: This won't work because of https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1194 and https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/233

giordano: Oh, well, it works precisely because you disabled avx512 instructions.

PR #4653, opened 2022-03-20
[blis] [tblis] Use LLVM OpenMP for BSD systems

giordano:
CC @xrq-phys

PR #6012, opened 2022-12-20
[tblis] Make `CompilerSupportLibraries_jll` available on all platforms

giordano:
libtci needs libatomic also on macOS.  Ref https://github.com/JuliaPackaging/Yggdrasil/pull/4653#issuecomment-1358816331.  CC: @xrq-phys.


xrq-phys:
Thanks so much!

PR #6016, opened 2022-12-21
Update to CoolProp 6.4.3

longemen3000:
@barche is this alright? i think we have no option in julia but to skip 6.4.2, but thankfully there is now 6.4.3


longemen3000:
first build attempt: fails with `[16:04:06] sed: CoolProp.sources/dev/Tickets/60.cpp: No such file or directory`

longemen3000:
seems that the folder structure changed between releases. now, the build fails with `[16:26:51] CMake Error: The source directory "/workspace/srcdir/CoolProp*" does not exist.`, but is created before?

giordano:
I guess you'll have to change the name of the directory everywhere it's used

PR #6020, opened 2022-12-22
[CryptoMiniSat] Update to 5.11.4

laurentbartholdi:
Updated to CMS 5.11.4; removed the anchored dependency to boost (CMS builds well with boost 1.8.0, at least).


C/CryptoMiniSat/build_tarballs.jl
```diff
@@ -54,7 +53,7 @@ products = Product[
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("boost_jll"; compat="=1.76.0"),
+    Dependency("boost_jll"),
```
giordano, line 56: Why? Boost ***must*** be pinned: #3024

C/CryptoMiniSat/build_tarballs.jl
```diff
@@ -6,20 +6,19 @@ const YGGDRASIL_DIR = "../.."
 include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
 
 name = "CryptoMiniSat"
-version = v"5.8.2"
-cryptominisat_version = v"5.8.0"
+version = v"5.11.4"
+cryptominisat_version = v"5.11.4"
```
giordano, line 10: It doesn't look like this is used at all, or am I missing something?

laurentbartholdi: Indeed, I don't think it is. I just left it from the previous file.

C/CryptoMiniSat/build_tarballs.jl
laurentbartholdi: Sorry, I didn't know that (and the link you give doesn't enlighten me further). Are you saying all boosts are binary incompatible? That would give the impression that each package should compile its own boost, otherwise how could two packages with incompatible boosts coexist? Please advise.

giordano: > Are you saying all boosts are binary incompatible?

Yes, all boost versions are incompatible with each other because they embed the full version number (major.minor.patch) in the soversion.

> That would give the impression that each package should compile its own boost

What does that mean?

> otherwise how could two packages with incompatible boosts coexist? 

Ask the boost developers who the brilliant idea to make each single version ABI-breaking (perhaps they had their own reasons, but that's still a terrible idea in terms of compatibility).

laurentbartholdi: OK, got it.
I was suggesting that CryptoMiniSat could download and compile its own boost, and include it. It seems to me that in the current setup lots of package updates will be held back till when each of them gets updated to use the new boost.

C/CryptoMiniSat/bundled/Picosat.patch
```diff
@@ -0,0 +1,17 @@
+--- a/src/picosat/picosat.c
++++ b/src/picosat/picosat.c
+@@ -8144,10 +8144,13 @@ picosat_stats (PS * ps)
+            ps->prefix, picosat_max_bytes_allocated (ps) / (double) (1 << 20));
+ }
+
++#if defined(WIN32) || defined(_WIN32) || defined(__WIN32) || defined(WIN32_) || defined(WIN32__) || defined(__WIN32__) /* mingw does not include sys/resource.h */
```
giordano, line 7: Why not just `_WIN32` which is the [only documented macro](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170#microsoft-specific-predefined-macros) to identify a Windows target?  Anything else is wrong (C standard mandates that compiler should ***not*** define macros without a leading underscore) or custom, no need to cater for those, we should expect any well behaved Windows compiler defines `_WIN32`.

Also, this patch should probably be upstreamed.

laurentbartholdi: OK, done.

PR #6026, opened 2022-12-25
Adding ParallelMaximumCliques.jl dependency LibPMC

dev10110:


P/PMC/build_tarballs.jl
```diff
@@ -0,0 +1,33 @@
+using BinaryBuilder
+
+name = "LibPMC"
```
giordano, line 3: This name must match exactly the name of the diretory

dev10110: of the directory in `Yggdrasil`?

giordano: Yes, the directory is called `PMC, the package is called `LibPMC`, they should match up to the case.  Also, please read the documentation: https://docs.binarybuilder.org/stable/building/#Name

L/LibPMC/build_tarballs.jl
```diff
@@ -0,0 +1,33 @@
+using BinaryBuilder
+
+name = "LibPMC"
```
giordano, line 3: Quoting from the [documentation section](https://github.com/JuliaPackaging/Yggdrasil/pull/6026#discussion_r1056998837) I [suggested reading](https://github.com/JuliaPackaging/Yggdrasil/pull/6026#discussion_r1056998837):

>  It should be the name of the upstream package, not for example that of a specific library or executable provided by it

Looks like a more appropriate name would be PMC?

dev10110: ok sure, I've changed it.

dev10110:
@giordano looks like all the checks have passed - do you know what happens next to integrate this package into Julia's package registry?

P/PMC/build_tarballs.jl
```diff
@@ -0,0 +1,33 @@
...
+rm -rf build
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release ..
+make -j${nproc}
+make install
+"""
+
+platforms = supported_platforms()
+# platforms = [Platform("x86_64", "macOS")]
```
giordano, line 21:
```suggestion
```

P/PMC/build_tarballs.jl
```diff
@@ -0,0 +1,33 @@
...
+platforms = expand_cxxstring_abis(platforms)
+
+products = [
+    LibraryProduct("libpmc", :libpmc),
+]
+
+dependencies = [
```
giordano, line 28:
```suggestion
dependencies = Dependency[
```

P/PMC/build_tarballs.jl
dev10110: updated.

P/PMC/build_tarballs.jl
dev10110: updated.

dev10110:
Thanks @giordano!

PR #6029, opened 2022-12-27
[libsharp2] Enable OpenMP on macOS

giordano:
Ref https://github.com/ziotom78/Healpix.jl/issues/101.  CC: @ziotom78


L/libsharp2/build_tarballs.jl
```diff
@@ -23,17 +23,17 @@ if [[ "${target}" == x86_64-* ]] && [[ "${target}" != *-apple-* ]] && [[ "${targ
     # Enable runtime multiarch, but not for macOS nor Windows. For Mac OS X, see
     # https://gitlab.mpcdf.mpg.de/mtr/libsharp/-/blob/6374a3a1ffb935443c56f09b371b11cf982a7e28/COMPILE
```
giordano, line 24: @ziotom78 I'm missing what's the problem with macOS, I can't tell it from the referenced page.

ziotom78: Mmm, I completely forgot what was the issue.

PR #6030, opened 2022-12-28
[Wayland] update to 1.21

jwahlstrand:
Alpine 3.15 RootFS has version 1.19, so use package from Alpine 3.17 for `wayland-scanner` (thanks @giordano).


giordano:
Nice to see it worked out smoothly!

PR #6039, opened 2022-12-29
[Qt6Base] Update to version 6.4.1 (target builds)

barche:


giordano:
CI machine is currently down.

barche:
@giordano I think this is good to go, thanks to your work on Qt5Base the macOS requirement is now also lowered to 10.14.

PR #6040, opened 2022-12-29
[LibCURL] Update to 7.87.0

ararslan:


giordano:
You have also to edit https://github.com/JuliaPackaging/Yggdrasil/blob/3573c32ce13a40902477be7993fa0bf8fe377f35/L/LibCURL/LibCURL/build_tarballs.jl#L5

giordano:
`--without-ssl` seems to be wrong

PR #6053, opened 2023-01-04
[GR] bump version

jheinen:
bump GR version to 0.71.3


jheinen:
- added IPv6 support
- improve time-to-first-plot

Issue #5651, opened 2022-10-10
[SCIP]: bliss, boost


benlorenz:
We would like to enable the SCIP interface (and maybe soplex as well) for our `polymake_jll` but I have a few questions:

- The latest SCIP_jll contains `bin/bliss`, `include/bliss/`, and `lib/libbliss.a` which might cause conflicts when used together with `bliss_jll` ?
  This is probably because of a [change in SCIP 8.0.1](https://scipopt.org/doc-8.0.2/html/RN80.php):
  > BLISS is now built from the vendored version from https://github.com/ds4dm/Bliss
- Is there a reason not to enable boost? The [`Dependency`](https://github.com/JuliaPackaging/Yggdrasil/blob/51a1e23b487b58633deeac9ecc36b36377b0063d/S/SCIP/build_tarballs.jl#L62) is still listed anyway. We would need the boost interface to use rational arithmetic in soplex.
- <del>Under which license is SCIP distributed here?
  The [license](https://github.com/scipopt/scip/blob/master/COPYING) in the GitHub repository mandates user tracking and only applies to members academic institutions which is probably not OSI-compatible:
  ```
  1. This license applies to you only if you are a member of a noncommercial 
     and academic institution, e.g., a university. The license expires as
     soon as you are no longer a member of this institution. 
  [...]
     c. You must keep track of access to the Program (e.g., similar to the 
        registration procedure at ZIB). 
  ```
  </del>
  Resolved: SCIP is now distributed under the Apache 2.0 license

lkastner:
Pinging @matbesancon as he seems to be the original author and might know more.

odow:
> Under which license is SCIP distributed here? The [license](https://github.com/scipopt/scip/blob/master/COPYING) in the GitHub repository mandates user tracking and only applies to members academic institutions which is probably not OSI-compatible

That link is the correct license. SCIP is source-available. It is not open-source.

matbesancon:
yes that's correct (for now)

matbesancon:
> Is there a reason not to enable boost? The [Dependency](https://github.com/JuliaPackaging/Yggdrasil/blob/51a1e23b487b58633deeac9ecc36b36377b0063d/S/SCIP/build_tarballs.jl#L62) is still listed anyway. We would need the boost interface to use rational arithmetic in soplex.

When I tried building with boost it errored, maybe the issues are now fixed and we could build with it now, PR welcome to try

odow:
[...]

PR #6054, opened 2023-01-04
[SCIP, SCIP_PaPILO] use bliss_jll and enable boost

benlorenz:
I bumped the patch-level from 300 to 301 to be able to add a compat entry for this version in polymake_jll.
This uses `FindBLISS.cmake` from the scip repo which does not do any version checks but this should not be necessary here.
I ran the testsuite for SCIP.jl on linux with locally built binaries and checked that I can now compile polymake_jll with the scip and soplex interfaces active.

cc: @matbesancon 
fixes: #5651

### DIFF
--- a/S/SCIP/build_tarballs.jl
+++ b/S/SCIP/build_tarballs.jl
@@ -3,11 +3,12 @@
 using BinaryBuilder, Pkg
 
 name = "SCIP"
-version = v"800.0.300"
+version = v"800.0.301"
 
 # Collection of sources required to complete build
 sources = [
     ArchiveSource("https://scipopt.org/download/release/scipoptsuite-8.0.3.tgz", "5ad50eb42254c825d96f5747d8f3568dcbff0284dfbd1a727910c5a7c2899091"),
+    DirectorySource("./bundled"),
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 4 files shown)


S/SCIP/build_tarballs.jl
```diff
@@ -30,10 +37,12 @@ cmake -DCMAKE_INSTALL_PREFIX=$prefix\
   -DGCG=0\
   -DUG=0\
   -DAMPL=0\
-  -DBOOST=off\
+  -DBOOST=ON\
   -DSYM=bliss\
   -DTPI=tny\
-  -DIPOPT_DIR=${prefix} -DIPOPT_LIBRARIES=${libdir} ..
+  -DIPOPT_DIR=${prefix} -DIPOPT_LIBRARIES=${libdir} \
+  -DBLISS_INCLUDE_DIR=${includedir} -DBLISS_LIBRARY=bliss \
```
fingolfin, line 44: A matter of taste perhaps, but I would suggest to list each argument on its own line; that makes it easier to see what is there (I tend to overlook the ones coming later in the line *sigh*).
```suggestion
  -DIPOPT_DIR=${prefix} \
  -DIPOPT_LIBRARIES=${libdir} \
  -DBLISS_INCLUDE_DIR=${includedir} \
  -DBLISS_LIBRARY=bliss \
```

matbesancon:
apart from the minor comment from @fingolfin looks good to me, thanks!

S/SCIP/build_tarballs.jl
```diff
@@ -17,9 +18,15 @@ script = raw"""
 # remove when CMake accounts for this
 if [[ "${target}" == *86*-linux-gnu ]]; then
    export LDFLAGS="-lrt"
+elif [[ "${target}" == *-mingw* ]]; then
```
matbesancon, line 21: was this tested on windows? So far I've never managed to have the windows binaries work

matbesancon (approved):

matbesancon:
@benlorenz it seems it didn't suffice to get the windows build working, there is still a segfault somewhere :/
https://github.com/scipopt/SCIP.jl/pull/254

matbesancon:
ping @leoneifler in case if you've an idea about the thread local option?

PR #6067, opened 2023-01-05
[Qt6ShaderTools] Update to v6.4.1 (targets build)

barche:


barche:
It seems that somehow the musl cxx03 binaries for Qt6Base got the C++11 string abi:

```
% strings libQt6Core.so | grep QByteArray | grep toStdString | c++filt -t 
QByteArray::toStdString[abi:cxx11]() const
QByteArray::toStdString[abi:cxx11]() const
```

Also 6.4.2 got released today: https://download.qt.io/official_releases/qt/6.4/

So I propose to start over with Qt6Base and try to figure out there what went wrong.

giordano:
The auditor also noticed it: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=24377&view=logs&j=39daa719-2c25-5f24-c641-c59889458f76&t=b2632530-9cab-5bbf-924f-8049026d48b2&l=86328. But this was lost in the sea of warnings we get....

PR #6076, opened 2023-01-08
[Qt6Base] v6.4.2 host build

barche:


barche:
This fails to unpack a dependency because bzip2 isn't installed, is this a known issue?

giordano:
https://github.com/JuliaCI/rootfs-images/pull/242#issuecomment-1386246749

PR #6081, opened 2023-01-09
[libucl] Builder for Universal Configuration Language library

ararslan:


L/LibUCL/build_tarballs.jl
```diff
@@ -0,0 +1,34 @@
...
+    --host="${target}"
+make -j${nproc}
+make install
+"""
+
+platforms = supported_platforms()
+
+products = [LibraryProduct("libucl", :libucl)]
+
+dependencies = [Dependency("LibCURL_jll")]
```
giordano, line 31:
```suggestion
dependencies = [
    Dependency("LibCURL_jll"),
    Dependency("PCRE_jll"; platforms=filter(Sys.iswindows, platforms)),
]
```

L/LibUCL/build_tarballs.jl
```diff
@@ -0,0 +1,34 @@
...
+./configure \
+    --enable-urls \
+    --enable-shared \
+    --disable-static \
+    --prefix="${prefix}" \
+    --build="${MACHTYPE}" \
+    --host="${target}"
+make -j${nproc}
+make install
+"""
```
giordano, line 25:
```suggestion

if [[ "${target}" == *-mingw* ]]; then
    # Cover up the traces of the hack
    rm "${includedir}/regex.h"
fi
"""
```

L/LibUCL/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+version = v"0.8.2"
+
+sources = [GitSource("https://github.com/vstakhov/libucl.git",
+                     "a615938cec3ae35c70aa2fb9845c9c12e5c7326f")]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/libucl*
+if [[ "${target}" == *-mingw* ]]; then
+    cp "${includedir}/pcreposix.h" "${includedir}/regex.h"
+    sed -i 's/-lregex/-lpcreposix-0/' configure
```
ararslan, line 13: Actually this probably needs to be `configure.ac`, or maybe just `./autogen.sh` should come first

giordano: Ugh, you're right

ararslan: Actually if we're using PCRE, this probably needs to be
```suggestion
    sed -i 's/-lregex/-lpcre/' configure
```
right? Or does the PCRE builder include a `libpcreposix-0.dll` and it just isn't listed as a library product in the build recipe?

giordano: I'm away from computer now, can't check.

giordano: > Or does the PCRE builder include a libpcreposix-0.dll and it just isn't listed as a library product in the build recipe?

That's indeed the case.

ararslan:
Thanks for your help getting this finished up, @giordano!

PR #6086, opened 2023-01-13
add `glib-networking` recipe

t-bltg:
Fix https://github.com/JuliaPackaging/Yggdrasil/issues/5673.

For the naming, I thought it would be consistent with `Glib` to use `CamelCase`.
IIUC, we cannot use a dash, so instead use an underscore.

Comments welcome.


t-bltg:
Crap, I missed https://github.com/JuliaPackaging/Yggdrasil/pull/5689.

PR #6088, opened 2023-01-14
Build MLIR-C on LLVM >= 14

ToucheSir:
Originally added in https://reviews.llvm.org/D113731. See also https://github.com/google/mlir-hs for an example of non-Julia language bindings.


vchuravy:
Superseded by #6210

PR #6102, opened 2023-01-18
CI: Use the `yggdrasil.x86_64` rootfs image

DilumAluthge:


vchuravy:
We don't need it in the pipeline, but the build step defined in utils.jl will open a PR with that.

Thanks!

.buildkite/utils.jl
```diff
@@ -81,14 +83,20 @@ end
...
-    apt install -y xz-utils
     # Don't share secrets with build_tarballs.jl
     BUILDKITE_PLUGIN_CRYPTIC_BASE64_SIGNED_JOB_ID_SECRET="" AWS_SECRET_ACCESS_KEY="" .buildkite/build.sh
     """
 
     build_plugins = plugins()
     push!(build_plugins,
+        "staticfloat/sandbox#v1" => Dict(
+            "rootfs_url" => "https://github.com/JuliaCI/rootfs-images/releases/download/v5.52/yggdrasil.x86_64.tar.gz",
+            "rootfs_treehash" => "0000000000000000000000000000000000000000", # TODO: put in the real hash
```
DilumAluthge, line 94: You'll need to put the actual treehash in.

vchuravy: Where do I find that?

DilumAluthge: Finally got around to doing this.

PR #6120, opened 2023-01-20
[MAGMA] fix compat for LBT

Wimmerer:
I want this to work on 1.8 as well.

### DIFF
--- a/M/MAGMA/build_tarballs.jl
+++ b/M/MAGMA/build_tarballs.jl
@@ -48,7 +48,8 @@ dependencies = [
     # one cuda version, you have to include them as Archive Sources. (see Torch_jll)
     RuntimeDependency(PackageSpec(name="CUDA_Runtime_jll")),
     BuildDependency(PackageSpec(name="CUDA_full_jll", version=v"11.0.3")),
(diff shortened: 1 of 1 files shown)


M/MAGMA/build_tarballs.jl
```diff
@@ -48,7 +48,8 @@ dependencies = [
     # one cuda version, you have to include them as Archive Sources. (see Torch_jll)
     RuntimeDependency(PackageSpec(name="CUDA_Runtime_jll")),
     BuildDependency(PackageSpec(name="CUDA_full_jll", version=v"11.0.3")),
-    Dependency("libblastrampoline_jll"; compat="5.2.0"),
+    Dependency(get_addable_spec("libblastrampoline_jll", v"5.1.1+0"); platforms=filter(p -> VersionNumber(p["julia_version"]) == v"1.8.0", platforms)),
+    Dependency(get_addable_spec("libblastrampoline_jll", v"5.2.0+0"); platforms=filter(p -> VersionNumber(p["julia_version"]) >= v"1.9.0", platforms)),
```
giordano, line 52: Isn't it enough to require 5.1.1?  I'm not sure filtering on `"julia_version"` is doing anything at all here

PR #6125, opened 2023-01-21
[Qt6Base] Version 6.4.2 (target builds)

barche:
I need to check if the problem with wrong C++ ABI on MUSL still exists with this version  before this can be merged.


barche:
The problem was that all musl builds were using `CMAKE_HOST_TOOLCHAIN`. Filing this under the "how did it ever work?" category. Should be OK now.

PR #6127, opened 2023-01-22
New binary: Valhalla

jeremiahpslewis:


V/Valhalla/build_tarballs.jl
```diff
@@ -0,0 +1,69 @@
...
+    -DProtobuf_INCLUDE_DIR=${includedir} \
+    -DPROTOBUF_LIBRARY=${libdir}/libprotobuf.${dlext} \
+    -DENABLE_SERVICES=OFF \
+    -DENABLE_TOOLS=OFF \
+    -DENABLE_CCACHE=OFF \
+    -DENABLE_BENCHMARKS=OFF \
+    -DLOGGING_LEVEL=DEBUG
+    
+make -j$(nproc)
+make -j$(nproc) install
```
giordano, line 38: `$(nproc)` is wrong
```suggestion
make -j${nproc}
make -j${nproc} install
```

V/Valhalla/build_tarballs.jl
```diff
@@ -0,0 +1,69 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="protoc_jll", uuid="c7845625-083e-5bbe-8504-b32d602b7110"))
+    Dependency(PackageSpec(name="LibCURL_jll", uuid="deac9b47-8bc7-5906-a0fe-35ac56dc84c0"))
+    Dependency(PackageSpec(name="jq_jll", uuid="f8f80db2-c0ba-59e9-a5c3-38d72e3c5ac2"))
+    Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a"))
+    Dependency(PackageSpec(name="Lz4_jll", uuid="5ced341a-0733-55b8-9ab6-a4889d929147"))
+    Dependency("boost_jll")
```
giordano, line 60: How many more times will I need to link #3024? :slightly_smiling_face:

jeremiahpslewis: 😅

giordano:
> Try gcc 9 (which is protoc's lower bound)

Why?

V/Valhalla/build_tarballs.jl
```diff
@@ -0,0 +1,98 @@
...
+    GitSource("https://github.com/valhalla/valhalla.git", "ea7d44af37c47fcf0cb186e7ba0f9f77e96f202a"),
+    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz", "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
+    DirectorySource("./bundled"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/valhalla/
+
+# Handle Mac SDK <10.12 errors
```
giordano, line 19: Why do you need to download the 10.15 sdk if we have already the 10.12?

jeremiahpslewis: Yup, great point. Just need an env var here.

V/Valhalla/bundled/patches/pkg-config-mingw.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
+index 4632fdbcf..c76b29884 100644
+--- a/src/CMakeLists.txt
++++ b/src/CMakeLists.txt
+@@ -291,7 +291,7 @@ if(PKG_CONFIG_FOUND)
+   if(NOT \${_res} EQUAL 0)
+     message(FATAL_ERROR \"Configuring libvalhalla.pc failed: \${_err}\")
+   endif()")
+-  install(FILES ${VALHALLA_BINARY_DIR}/libvalhalla.pc
++  install(FILES ${VALHALLA_BINARY_DIR}/libvalhalla.pc.exe
```
giordano, line 10: :raised_eyebrow:

jeremiahpslewis: The file that is created is `pc.exe`...and its windows so it seemed plausible to me that we should accommodate it. Or is the error actually that a `.pc.exe` file is generated? In which case is something going wrong here? https://github.com/valhalla/valhalla/blob/f7c31e5c232dd20156235ef1ba2184dc5a27309a/src/CMakeLists.txt#L290

giordano: I don't see why a pkg config file should have any extensions but `.pc`, on ***any*** operating system.

jeremiahpslewis: Ok, so I think I'm going to accept that the Valhalla pkg-config setup for windows is broken and beyond my ability to debug. I've raised the issue upstream: https://github.com/valhalla/valhalla/issues/3931

V/Valhalla/build_tarballs.jl
```diff
@@ -0,0 +1,83 @@
...
+# The products that we will ensure are always built
+products = Product[
+    LibraryProduct("libvalhalla", :libvalhalla),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("boost_jll"; compat="=1.76.0")
+    Dependency("GEOS_jll")
+    Dependency("jq_jll")
```
giordano, line 74: jq is an executable, is it really a runtime dependency?  Looks unlikely.

giordano: Looking at https://aur.archlinux.org/packages/valhalla, jq is listed a build dependency.

jeremiahpslewis: Looks like jq is only used in tests during the build, but tests are off so let's try without it entirely

PR #6075, opened 2023-01-07
Create file `CONTRIBUTING.md`

giordano:

PR #6136, opened 2023-01-24
Bump to Libxc 6.1.0

mfherbst:
One patch seems to have been integrated, the other not. I'll remove the patch files if CI goes through.


mfherbst:
Interesting ... how can I make Buildkite build the CPU version (which is also changed since I changed the common.jl) ?

giordano:
You always have to modify `build_tarballs.jl`, or a file in a subdirectory: https://github.com/JuliaPackaging/Yggdrasil/pull/6075/files.  This is how we know what to rebuild.

mfherbst:
Cool. Then putting the version into the common.jl is not the most sensible thing to do.

giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/1f69c249af143bf44762f0ea81a28b0681fd35b1/L/LibCURL/LibCURL/build_tarballs.jl#L5

mfherbst:
Yeah, but that's a bit cheating and might get forgotten. This way if you forget to increment the version the hash does not match. Ok ... not actually too sure that's better. Happy to go with your proposal if that's what people do in this case.

PR #6138, opened 2023-01-24
update ducc0 to version 0.28.0

mreineck:
Update ducc0 to version 0.28

(I'm exprimenting with "-fvisibility=hidden" to reduce library size and improve load times; let's hope this works on Windows platforms.)


giordano:
Good to go?

mreineck:
If the Winows builds produce something useful, yes. Unfortunately I have no way to verify this...

mreineck:
Still, if something is broken, we can fix it next time. This is not an essential package yet :-)

PR #6140, opened 2023-01-24
[MAGMA] Missing underscore in ABI

Wimmerer:


giordano:
As I mentioned already on Slack, it would be most useful for me if you let me know that [previous registration wasn't successful](https://github.com/JuliaRegistries/General/pull/76026#issuecomment-1402463762), because that'd break many things during the registration process here, causing further delays.  I can't possibly keep an eye on everything.

Wimmerer:
Yes sorry, I was about to message. I missed a commit my apologies.

giordano:
I don't think the [dyanmically launch pipelines](https://buildkite.com/julialang/yggdrasil/builds/405#0185e541-94c0-46c8-8283-4ab2d0d514af) step is working.  Try touching the build_tarballs file, even though that shouldn't ideally be necessary.

PR #6115, opened 2023-01-19
[oneTBB] Version bump

jeremiahpslewis:
- Expand build to all platforms

PR #6144, opened 2023-01-25
[mingw] Build v10.0.0

jeremiahpslewis:
Seems like a mingw version bump may 'heal' compile errors we are getting.

- https://github.com/oneapi-src/oneTBB/issues/863

### DIFF
--- a/M/Mingw/build_tarballs.jl
+++ b/M/Mingw/build_tarballs.jl
@@ -1,12 +1,12 @@
 using BinaryBuilder
 
 name = "Mingw32"
-version = v"5.0.4"
+version = v"10.0.0"
 
 # sources to build, such as mingw32, our patches, etc....
 sources = [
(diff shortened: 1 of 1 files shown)


giordano:
1. this is not used in the BinaryBuilder toolchain, so touching this isn't going to do anything that you think it is
2. as already said in https://github.com/JuliaPackaging/Yggdrasil/pull/6115#issuecomment-1399079196, `_MCW_DN` ***is*** defined already, but only under the condition that `__STRICT_ANSI__` is undefined.

jeremiahpslewis:
> * this is not used in the BinaryBuilder toolchain, so touching this isn't going to do anything that you think it is
> * as already said in [[oneTBB] Version bump #6115 (comment)](https://github.com/JuliaPackaging/Yggdrasil/pull/6115#issuecomment-1399079196), `_MCW_DN` _**is**_ defined already, but only under the condition that `__STRICT_ANSI__` is undefined.
This change to MINGW seems to fix the ANSI issue: https://github.com/mingw-w64/mingw-w64/commit/c6e13e0c105eab7797c2373819b49fff6b05566c

PR #6149, opened 2023-01-26
[gitoxide] Update to v0.20.0

giordano:


giordano:
It requires glibc 2.26 on i686, I'll come back to this when we can easily swap glibc: https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/863

PR #6150, opened 2023-01-26
[bottom] Upgrade to v0.8.0

giordano:


giordano:
Also waiting on https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/863 (I suspect newer rustc needs glibc 2.16 on some platforms).

PR #6152, opened 2023-01-26
[RichDEM] Differentiate on Julia versions

barche:
This should fix issue https://github.com/JuliaInterop/libcxxwrap-julia/issues/115 and also attempts to add a few more platforms.


barche:
Most failures are due of the style:

```
/workspace/destdir/lib/libgdal.so: undefined reference to `curl_easy_getinfo@CURL_4'
```

Any idea how to fix this, @giordano ? It seems in some cases there is a version mismatch with `libcurl`, which comes in through the `GDAL_jll` dependency, but it seems to correctly specify libcurl compat?

giordano:
Sounds like you need to link to libcurl as well (but I don't why it wasn't a problem before).

sunayana:
Earlier I was not building for all platforms supported by Julia, but then the generated binaries were not working as reported in the issue linked by @barche

barche:
The apple failures are because OpenMP is not found, I think that can be solved by using `LLVMOpenMP_jll` on macOS as is done e.g. here:
https://github.com/barche/Yggdrasil/blob/3b5c42fac291891dba6636fc60d2cd157791a046/D/Darknet/common.jl#L68

Not sure about the others, but they are more obscure platforms, so they could be disabled if there is no easy fix.

sunayana:
In the meanwhile I have also tested against the tarballs in artifacts generated by buildkite and can confirm that this PR solves my issues reported in https://github.com/JuliaInterop/libcxxwrap-julia/issues/115 so could I request this PR to be merged.

R/RichDEM/build_tarballs.jl
```diff
@@ -17,24 +22,46 @@ sources = [
...
-    Platform("powerpc64le", "linux"; libc = "glibc"),
-    Platform("aarch64", "linux"; libc = "musl"),
-    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
-    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
-]
-
+julia_versions = [v"1.6.3", v"1.8"]
+include("../../L/libjulia/common.jl")
+platforms = vcat(libjulia_platforms.(julia_versions)...)
+platformfilter(p) = (p.tags["arch"] != "armv6l" && p.tags["os"] ∉ ["macos", "freebsd"])
```
giordano, line 62: Avoid accessing tags, we have functions
```suggestion
platformfilter(p) = (arch(p) != "armv6l" && Sys.isbsd(p))
```

barche:
Thanks for the update, I think this is OK. I'll click the "update branch" button and then it should be good.

giordano:
> I'll click the "update branch" button and then it should be good.

Please don't, it's almost always totally useless.  I'm considering removing it again.

sunayana:
Thank you very much @barche and @giordano

PR #6169, opened 2023-01-30
[polymake] Dummy edit to trigger build as a test

giordano:


giordano:
Sigh: https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/ (sighing even if they're reverting the change for now)

PR #6175, opened 2023-02-01
Wizard recipe: bdwgc-v8.2.2

lucifer1004:

* Package name: bdwgc
* Version: v8.2.2

### DIFF
--- a/B/bdwgc/build_tarballs.jl
+++ b/B/bdwgc/build_tarballs.jl
@@ -1,0 +1,39 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "bdwgc"
+version = v"8.2.2"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/ivmai/bdwgc/releases/download/v$(version)/gc-$(version).tar.gz", "f30107bcb062e0920a790ffffa56d9512348546859364c23a14be264b38836a0"),
+    ArchiveSource("https://github.com/ivmai/libatomic_ops/releases/download/v7.6.14/libatomic_ops-7.6.14.tar.gz", "390f244d424714735b7050d056567615b3b8f29008a663c262fb548f1802d292")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/gc-8.2.2/
+mv ../libatomic_ops-7.6.14/ libatomic_ops
+./autogen.sh 
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j${nproc}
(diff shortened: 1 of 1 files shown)


B/bdwgc/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "bdwgc"
+version = v"8.2.2"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/ivmai/bdwgc/releases/download/v$(version)/gc-$(version).tar.gz", "f30107bcb062e0920a790ffffa56d9512348546859364c23a14be264b38836a0"),
+    ArchiveSource("https://github.com/ivmai/libatomic_ops/releases/download/v7.6.14/libatomic_ops-7.6.14.tar.gz", "390f244d424714735b7050d056567615b3b8f29008a663c262fb548f1802d292")
```
giordano, line 11: Shouldn't this be a separate package?

giordano (commented):
License file is missing (which is ***alwaays*** the case when there are multiple sources, since we can't guess from where to get the license)

B/bdwgc/build_tarballs.jl
lucifer1004: `bdwgc` requires `libatomic_ops` to be built, see https://github.com/ivmai/bdwgc

image

The documentation does say that we can switch to a compiler that natively supports atomic intrinsics, but the default GCC (4.8.5) cannot meet this requirement. I have not tested other GCC versions, though.

lucifer1004:
For license, the original Git repos do have license files, but the released archives do not have the license files with them. Maybe I can switch to Git source and install the license files?

giordano:
> but the released archives do not have the license files with them. 

Wut.  Looks like a mistake on their side.

> Maybe I can switch to Git source and install the license files?

That'd be better, yes.

B/bdwgc/build_tarballs.jl
giordano: Better to use a compiler which supports atomics (perhaps gcc 5 is enough?)

lucifer1004:
The master branch contains licenses, however, the latest released tag does not contain licenses :cry: .

fingolfin:
You could also specify the license file as a separate source. Or if it is a standard license, copy it from elsewhere... But then make sure that the release version really has the same license as the master branch?

fingolfin:
On a meta-level, I wonder how well bdwgc will work in practice inside a process that also uses the Julia GC?

fingolfin:
@lucifer1004 are you still interested in this (despite 1.5 years having passed) ? If so, the license issue still needs to be addressed. If not, let's close this.

lucifer1004:
I think I will pick it up.

PR #6182, opened 2023-02-02
[ImageMagick] Add support for JPEG2000 decoding

jochalek:


giordano:
In case you haven't seen [it](https://buildkite.com/julialang/yggdrasil/builds/553#0186101f-23bc-4c1f-903d-7711cadbdbce/558-565):
```
ArgumentError: The archive automatically generated by GitHub
  https://github.com/ImageMagick/ImageMagick6/archive/6.9.12-62.tar.gz
may not have a stable checksum in the future, thus cannot be used as a reliable source, see
<https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/>.
Use a different source, for example a `GitSource`, or an official release artifact uploaded
by the maintainers of the package (*not* the automatic archive produced by GitHub).
```

PR #6185, opened 2023-02-02
Increase pipeline depot cache size to 250GB

staticfloat:
Let us increase our depot cache size to 400GB; this should greatly increase our cache hit rate


giordano:
Should this be reduced  now that we have 12 agents?

staticfloat:
Yeah, I'll drop it down to 250GB to be conservative.

giordano:
Need to update the signature

PR #6208, opened 2023-02-05
[PETSc] Version 3.18.3

amontoison:
I'm not an expert of PETSc so I hope that it will work without modifications of the `build_tarballs.jl` :crossed_fingers:


amontoison:
The patch `mpi-constants.patch` is not working with the new release.
I don't know enough PETCs to update it.
I close the PR.

PR #6211, opened 2023-02-05
Aeron IPC / UDP multicast library

sefffal:
I'd like to create a Julia wrapper around the Aeron c library. It provides low latency multicast streaming between processes.

This is my first attempt at using BinaryBuilder and I've hit a snag. 
The project builds locally and in an Ubuntu container but hits compiler errors when running the wizard.

I'll continue to poke at this, but any help would be appreciated.


A/Aeron/build_tarballs.jl
```diff
@@ -0,0 +1,54 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+apk update
+apk add openjdk11 hdrhistogram-c-dev
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DAERON_ENABLE_NONSTANDARD_OPTIMIZATIONS=OFF AERON_TESTS=OFF aeron
+make
```
giordano, line 23: This will make the build less horribly slow
```suggestion
make -j${nproc}
```

But `make install` disappeared.

A/Aeron/build_tarballs.jl
```diff
@@ -0,0 +1,54 @@
...
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; )
+]
+
+
+# The products that we will ensure are always built
+products = Product[
+]
```
giordano, line 47: Reminder that we'll need to add products here.

A/Aeron/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+
+# Bash recipe for building across all platforms
+script = raw"""
+apk update
+apk add openjdk11 hdrhistogram-c-dev libbsd-dev util-linux-dev
+cd $WORKSPACE/srcdir/aeron
+mkdir build && cd build
+CMAKE_FLAGS=(-DCMAKE_INSTALL_PREFIX=$prefix
+-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}
+-DCMAKE_BUILD_TYPE=Release
```
giordano, line 21: I don't know what they had in mind in https://github.com/real-logic/aeron/blob/354608c4c3b12ecc363508da06fce2b52869db6d/CMakeLists.txt#L95-L97, but they very much use GNU extensions, so we very much need to enable extensions
```suggestion
-DCMAKE_BUILD_TYPE=Release
-DCMAKE_C_EXTENSIONS=ON
```

A/Aeron/bundled/patches/remove_std11_flags.patch
```diff
@@ -0,0 +1,12 @@
+diff --git before/aeron/CMakeLists.txt after/aeron/CMakeLists.txt
```
giordano, line 1: This patch is unneeded.

A/Aeron/build_tarballs.jl
sefffal: Thanks! Great spot.

sefffal:
This seems to now be building on most supported platforms expect x86-64-apple-darwin (but aarch64 works?). 
I also ran into issues when cross compiling it to windows (says EPOLL etc aren't available) but I'd like to punt on solving these for now.

sefffal:
Would it be okay to merge this only linux support to start?

A/Aeron/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+]
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = Product[
+    LibraryProduct(["libaeron"], :libaeron)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
```
giordano, line 61: From the log:
```suggestion
dependencies = [
    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
```

giordano:
Ok, let's start with a small set of platforms.  Thanks for your contribution!

Issue #6214, opened 2023-02-07
[OpenBLAS] CONSISTENT_FPCSR=1 for certified computation


orkolorko:
It would be interesting to have a build of OpenBLAS with the flag 

`CONSISTENT_FPCSR = 1`

enabled, to allow use of multithreaded BLAS operations in
[IntervalLinearAlgebra](https://github.com/JuliaIntervals/IntervalLinearAlgebra.jl)

At the moment, each thread in OpenBLAS may have a different rounding mode see Issue
https://github.com/JuliaIntervals/IntervalLinearAlgebra.jl/issues/131
and this breaks the algorithms needed for the certified linear algebra computations

I open this issue mainly to understand what would be considered the best course of action and would like opinions on the following idea:

- create a new jll package, called OpenBLASConsistentFPCSR_jll
- use libblastrampoline to switch from OpenBLAS to OpenBLASConsistentFPCSR

Thank you in advance 

P.S. @lucaferranti, do you have any opinion on this?

giordano:
Having a separate package sounds like a good idea. I'm not familiar with this option, I presume enabling it comes with a performance penalty?

orkolorko:
By following discussions on the OpenBLAS github, the issue whether this generates performance penalty is unclear,

https://github.com/xianyi/OpenBLAS/pull/3772

As the thread pool is started, the rounding mode is set on all the threads:

https://github.com/xianyi/OpenBLAS/blob/8e851160d758d9b2bdfd61a1b19663dd4418b351/driver/others/blas_server_omp.c#L388

This should cost 1flop per thread, but probably things are more complicated; it seems even an open discussion in the OpenBLAS community whether to enable or not this option by default.

On one side, it may have some effect on performance (if it is 1flop per thread in a matrix multiplication or BLAS operation, this should be really marginal) on the other side, if for some reason the users are expecting the BLAS computation to respect some rounding mode, this can lead to non-predictable results.

lucaferranti:
> P.S. @lucaferranti, do you have any opinion on this?

I dont have strong opinions on the two options, I presume the separate package could be easier as option? So maybe we can go with that

orkolorko:
So, to go down the separate option package, to build OpenBLASConsistentFPCSR_jll, I think 
the process should be to fork Yggdrasil, and essentially I would copy the setup of OpenBLAS_jll substituting

[...]

PR #6215, opened 2023-02-07
Added OpenBLASConsistentFPCSR

orkolorko:
We add a package OpenBLASConsistentFPCSR, which is nothing else than OpenBLAS
compiled with the `CONSISTENT_FPCSR=1` flag enabled, following issues 

https://github.com/JuliaPackaging/Yggdrasil/issues/6214
and
https://github.com/JuliaIntervals/IntervalLinearAlgebra.jl/issues/131

This flag controls the fact that, as the thread pool is started, all the threads have the same rounding mode; this flag is not implemented by default since in principle this may affect performance and there are no benchmarks proving the opposite.

The packages is essentially the same as OpenBLAS with two differences:

- we removed the build recipes for 32 bit BLAS (maybe it is not a good idea, but it should be not so much work to add them again)
- we added the flag `CONSISTENT_FPCSR=1` at
https://github.com/orkolorko/Yggdrasil/blob/697e6e7314378e42617c223bbc284780f00270ee/O/OpenBLASConsistentFPCSR/common.jl#L86
- we changed the names of the built packages from "OpenBLAS*" to "OpenBLASConsistentFPCSR*"


giordano:
I'd only use the latest version, there's little point in building all previous ones.

orkolorko:
Ok, even if it seems that this flag is supported on aarch64
https://github.com/xianyi/OpenBLAS/commit/84453b924fe7695029cad974dfe0cf7bf6ffe0f6
the build is failing.

I will work to exclude unsupported platforms.

orkolorko:
@giordano the compilation went through without  any problems, but I needed to restrict to x86_64

giordano:
Thanks for your contribution!

orkolorko:
Thank you @giordano!

PR #6220, opened 2023-02-07
[Pandoc] Add platform for aarch64 macos in pandoc recipe

kdheepak:


P/pandoc/build_tarballs.jl
```diff
@@ -48,4 +49,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies, julia_compat="1.6")
```
giordano, line 52: Can't do this without changing the version number

kdheepak: But the version number is tied to the version of pandoc, right? What version number can I bump it to?

kdheepak: @mose I'm not understanding what I can change the version number to. 

The docs say this: https://docs.binarybuilder.org/stable/building/#Version-number

> This is the version number used in tarballs and should coincide with the version of the upstream package. However, note that this should only contain major, minor and patch numbers, so
>
> ```
> julia> v"1.2.3"
> v"1.2.3"
> ```
>
> is acceptable, but
>
> ```
> julia> v"1.2.3-alpha"
> v"1.2.3-alpha"
> 
> julia> v"1.2.3+3"
> v"1.2.3+3"
> ```
>
> or a version including more than three levels (e.g., 1.2.3.4) are not [acceptable]. Truncate the version to the patch number if necessary.
> The generated JLL package will automatically add a build number, increasing it for each rebuild of the same package version.

The latest pandoc 2.x release is `v2.19.2`: https://github.com/jgm/pandoc/releases/tag/2.19.2

and this file is already set to `v2.19.2`: https://github.com/kdheepak/Yggdrasil/blob/3efba3bd2c9903f4178bca7267cab45627705e4c/P/pandoc/build_tarballs.jl#L5

What can I change the version to? 

Right now the version number of `pandoc_jll` is `pandoc-v2.19.2+1`. Do you want me to manually set it here to `v2.19.2+2`?

There is a new version of pandoc out that's `v3.0`. I'm assuming that's not what you are asking me to change here.

giordano: Call it version `2.19.3`

kdheepak: So is the idea that binarybuilder artifacts are no longer tied to the version of the actual binaries? What if pandoc release a `2.19.3` at some point in the future?

giordano: > So is the idea that binarybuilder artifacts are no longer tied to the version of the upstream package?

We have dozens of packages here with fake versions numbers because of point 2 of #2763 (which is a pinned issue BTW).  There's no exception to that, the version number ***MUST*** change.

> What if pandoc release a 2.19.3 at some point in the future?

We'll live with that, as we're living with all other packages with fake versions numbers for the same reason.  Also, I'm not into pandoc development but that might be unlikely since they moved to version 3

P/pandoc/build_tarballs.jl
```diff
@@ -2,7 +2,7 @@ using BinaryBuilder
 
 # Collection of pre-build pandoc binaries
 name = "pandoc"
-pandoc_ver = "2.19.2"
+pandoc_ver = "2.19.3"
 version = VersionNumber(pandoc_ver)
```
giordano:
```suggestion
pandoc_ver = "2.19.2"
version = v"2.9.13" # <-- This version number is a lie to allow using aarch64-apple-darwin
```

kdheepak: Caught that when I was making a similar change to `pandoc_citeproc`. Fixed in [5c2209a](https://github.com/JuliaPackaging/Yggdrasil/pull/6220/commits/5c2209a1e04bef7493197e5465b20b0192b2269d).

P/pandoc/build_tarballs.jl
kdheepak: Thanks for a link to the pinned issue! Great. I've updated it as suggested.

PR #6221, opened 2023-02-07
[pandoc_crossref] Add macos aarch64 platform support based on x86_64

kdheepak:
Similar changes to #6218 and #6220


P/pandoc_crossref/build_tarballs.jl
```diff
@@ -46,4 +48,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies, julia_compat="1.6")
```
giordano, line 51: Also here, can't change Julia compat without changing the 6

kdheepak: pandoc-crossref appears to be using 4 parts to their version number:

```
crossref_ver = "0.3.11.0"
```

I don't know exactly what that is about, but in Julia BinaryBuilder only supports 3 parts to it right?

So I bumped the version to `v"0.3.12"` here.

PR #6228, opened 2023-02-08
build Python with sqlite

### LINKED ISSUE #6189: Python_jll missing _sqlite.so?
```bash
(test) pkg> st
Status `~/work/test/Project.toml`
  [93d3a430] Python_jll v3.10.8+0

julia> using Python_jll

julia> run(Python_jll.python())
Python 3.10.8 (main, Oct 19 2022, 08:14:58) [GCC 4.8.5] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sqlite3
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/marius/.julia/artifacts/b5193ebf33adf32ff59321e19167fc91556f8d94/lib/python3.10/sqlite3/__init__.py", line 57, in <module>
    from sqlite3.dbapi2 import *
  File "/home/marius/.julia/artifacts/b5193ebf33adf32ff59321e19167fc91556f8d94/lib/python3.10/sqlite3/dbapi2.py", line 27, in <module>
    from _sqlite3 import *
ModuleNotFoundError: No module named '_sqlite3'
>>> 
```

Based on a normal Python installation it appears the file should have been in `<artifact_dir>/lib/pythonX.Y/lib-dynload/`. Looks like several extensions make it there correctly but not sqlite?

marius311:
Fixes #6189. The presence of sqlite3 lib causes the Python build to include it, no extra configure flags needed. I also needed the update otherwise was getting package not found errors. 

I personally also need this for Python 3.8 (since I depend on boostpython_jll which is capped at 3.8). Is there a way to also add this as a "backport" to 3.8? (this is my first time doing anything Yggdrasil)

### DIFF
--- a/P/Python/build_tarballs.jl
+++ b/P/Python/build_tarballs.jl
@@ -21,6 +21,7 @@ apk add autoconf-archive
 rm -f $(which python3)
 
 # We need these for the host python build
+apk update
 apk add zlib-dev libffi-dev
 
 # Create fake `arch` command:
@@ -104,6 +105,7 @@ dependencies = [
     Dependency("Expat_jll"; compat="2.2.10"),
     Dependency("Bzip2_jll"; compat="1.0.8"),
     Dependency("Libffi_jll"; compat="~3.2.2"),
+    Dependency("SQLite_jll"),
     Dependency("LibMPDec_jll"),
     Dependency("Zlib_jll"),
     Dependency("XZ_jll"),


giordano:
@Keno have you ever used boostpython_jll in practice?   Statistics on JuliaHub show [some downloads](https://juliahub.com/ui/Packages/boostpython_jll/3F9zb/1.76.0+1?page=2) last month. I'm wondering if we can bend the rules for once and rebuild boostpython with python 3.9 (and change the compat accordingly).

Keno:
I haven't tried xdm with 3.9, but I'm also not actively using it, so if boostpython needs to be messed with, I have no objection.

marius311:
Sorry if I'm missing something but this PR fixes Python 3.10, so we'd need boostpython with 3.10?

giordano:
Yeah, sorry, I thought we were on python 3.9

marius311:
So should I upgrade boostpython to 3.10 then? I'm fine with that solution barring unexpected things happening.

giordano:
Maybe an even better solution is to create have a version 1.76.1 of boostpython_jll which links to boost 1.76.0 and this new python 3.10, so we don't bend rules too much (just a slightly off version number for boostpython_jll)

marius311:
Tbh thinking about it further,, I think bending the rules by having Julia 1.76.1 be boostpython 1.76.0 is potentially more confusing than just having 1.76.0+1 need Python 3.8 but 1.76.0+2 need Python 3.10. Assuming the resolver is fine with the latter, that'd be my preferred choice. Any thoughts?

giordano:
> just having 1.76.0+1 need Python 3.8 but 1.76.0+2 need Python 3.10.

That's not confusing, that's not possible at all.  As far as Pkg is concerned, there's only one "1.76.0" release, the build number is discarded whatever we write.  And regarding versions numbers of packages diverging from upstream packages, I already posted about that today: https://github.com/JuliaPackaging/Yggdrasil/pull/6220#discussion_r1100467250

marius311:
I see, thanks. 1.76.1 it is then.

marius311:
Actually, why don't I just upgrade boostpython to 1.79.0 which is the latest boost_jll, this avoids this altogether, and I think is fine with my stuff.

giordano:
Because of

> but boost is terrible package to have as dependency because each single version is ABI breaking.

I think most packages are depending on 1.76.0 at the moment.

PR #6236, opened 2023-02-09
add madgraph

Moelf:


Moelf:
`MadGraph` is basically a Python package that at run time, produce Fortran source code and call `gcc-fortran` to compile them, thus it's impossible to ship as a binary `_jll`.

PR #6248, opened 2023-02-11
Update ANTLR4 from 4.9.3 to 4.11.1

gkastrinis:


A/Antlr4CppRuntime/build_tarballs.jl
```diff
@@ -3,26 +3,24 @@
...
-atomic_patch -p1 ../../../patches/forward-cmake-toolchain.patch
 mkdir build && cd build
-cmake .. -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+cmake .. -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DANTLR_BUILD_CPP_TESTS=
 make -j${nproc}
 make install
 
 # Remove large static library
-rm "${prefix}/lib/libantlr4-runtime.a"
+rm -f "${prefix}/lib/libantlr4-runtime.a"
```
giordano, line 23:
```suggestion
rm "${prefix}/lib/libantlr4-runtime*.a"
```

A/Antlr4CppRuntime/build_tarballs.jl
```diff
@@ -3,26 +3,24 @@
...
-atomic_patch -p1 ../../../patches/forward-cmake-toolchain.patch
 mkdir build && cd build
-cmake .. -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+cmake .. -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DANTLR_BUILD_CPP_TESTS=
 make -j${nproc}
 make install
 
 # Remove large static library
-rm "${prefix}/lib/libantlr4-runtime.a"
+rm "${prefix}/lib/libantlr4-runtime*.a"
```
giordano, line 23:
```suggestion
rm ${prefix}/lib/libantlr4-runtime*.a
```

A/Antlr4CppRuntime/build_tarballs.jl
```diff
@@ -41,4 +39,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"7")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"11")
```
giordano, line 42: GCC 10 worked before, no?
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"10")
```

gkastrinis: apple-darwin seemed to fail with v10
https://buildkite.com/julialang/yggdrasil/builds/967#01865a14-2d2b-4a05-8696-75d6bbbbcd5f

giordano: 1. we build with Clang, not GCC, on BSD platforms (macOS and FreeBSD)
2. even if we used GCC, `aarch64-apple-darwin` has only one GCC toolchain, which is version 12, so that using asking for version 10 or 11 is totally irrelevant, and this is already made irrelevant by the previous point

gkastrinis: Thanks for all the help! Although I have no idea how to proceed.

PR #6273, opened 2023-02-15
[OpenBLAS] Build for PowerPC with dynamic arch

giordano:
I'm not sure whether to use dynamic arch for powerpc through a config option to `openblas_script`, happy to do it if that's preferrable.

### DIFF
--- a/O/OpenBLAS/common.jl
+++ b/O/OpenBLAS/common.jl
@@ -147,7 +147,7 @@ function openblas_script(;num_64bit_threads::Integer=32, openblas32::Bool=false,
     elif [[ ${target} == arm-* ]]; then
         flags+=(TARGET=ARMV7)
     elif [[ ${target} == powerpc64le-* ]]; then
-        flags+=(TARGET=POWER8)
+        flags+=(TARGET=POWER8 DYNAMIC_ARCH=1)
     fi
 
     # If we're building for x86_64 Windows gcc7+, we need to disable usage of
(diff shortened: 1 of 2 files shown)


O/OpenBLAS/common.jl
```diff
@@ -147,7 +147,7 @@ function openblas_script(;num_64bit_threads::Integer=32, openblas32::Bool=false,
     elif [[ ${target} == arm-* ]]; then
         flags+=(TARGET=ARMV7)
     elif [[ ${target} == powerpc64le-* ]]; then
-        flags+=(TARGET=POWER8)
+        flags+=(TARGET=POWER8 DYNAMIC_ARCH=1)
```
vchuravy, line 150:
```suggestion
        flags+=(TARGET=GENERIC DYNAMIC_ARCH=1)
```

Like on x86?

O/OpenBLAS/common.jl
```diff
@@ -147,7 +147,7 @@ function openblas_script(;num_64bit_threads::Integer=32, openblas32::Bool=false,
     elif [[ ${target} == arm-* ]]; then
         flags+=(TARGET=ARMV7)
     elif [[ ${target} == powerpc64le-* ]]; then
-        flags+=(TARGET=POWER8)
+        flags+=(TARGET=GENERIC DYNAMIC_ARCH=1)
```
vchuravy, line 150: Turns out I was wrong. We want `TARGET=POWER8` here. But let me check maybe `generic` is enough

PR #6282, opened 2023-02-17
[DuckDB] julia client support set timezone

xcaptain:
currently using the cli we can change timezone, it's good if we can also do in julia


giordano:
Man, getting all  the macros right was a pain.  Thanks for your contribution!

PR #6288, opened 2023-02-19
Add libaec for szip support for HDF5

mkitti:
xref:
https://github.com/conda-forge/libaec-feedstock/blob/main/recipe/build.sh
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-libaec/PKGBUILD

It's unclear if we need any of these patches:
https://github.com/conda-forge/libaec-feedstock/blob/main/recipe/disable_static_cmake.patch
https://github.com/conda-forge/libaec-feedstock/blob/main/recipe/windows_restrict_keyword.diff
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-libaec/0005-cmake-fix-cmake-install.patch

### DIFF
--- a/L/libaec/build_tarballs.jl
+++ b/L/libaec/build_tarballs.jl
@@ -1,0 +1,57 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "libaec"
+version = v"1.0.6"
+
+# Collection of sources required to complete build
+sources = [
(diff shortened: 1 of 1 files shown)


L/libaec/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
+
```
giordano, line 43:
```suggestion
platforms = supported_platforms(; exclude=!Sys.islinux)
```

PR #6292, opened 2023-02-19
Exodus_jll updates

cmhamel:
Adding a patch to the seacas git repository that fixes 4 minors issues on Windows builds. Three of them are the name of header files are capitalized vs. not and the fourth is related to the seacas cmake build automatically killing windows builds, that was simply commented out in the git patch.

Differences to the build_tarballs.jl include
1. Changed version number to actually match the exodusII library version number in seacas.
2. Added atomic_patch of git patch described above.
3. Added exodiff as executable product. This is used to test diffs between exodusII databases.
4. nem_slice and nem_spread libraries as products. These are used to perform mesh decomposition for parallel calculations.
5. epu as executable product. This is used to 'stitch' decomposed meshes back together after parallel computations.
6. Added windows platforms
7. Added preferred gcc version as 5 since one of the new products requires this to compile.

### DIFF
--- a/E/Exodus/build_tarballs.jl
+++ b/E/Exodus/build_tarballs.jl
@@ -32,7 +32,7 @@ cmake \
     -D CMAKE_Fortran_FLAGS="" \
     -D Seacas_ENABLE_STRONG_C_COMPILE_WARNINGS="" \
     -D Seacas_ENABLE_STRONG_CXX_COMPILE_WARNINGS="" \
-    -D CMAKE_INSTALL_RPATH:PATH=${prefix}/lib \
+    -D CMAKE_INSTALL_RPATH:PATH="${libdir}" \
     -D BUILD_SHARED_LIBS:BOOL=YES \
     -D Seacas_ENABLE_SEACASExodus=YES \
     -D Seacas_ENABLE_SEACASExodus_for=NO \
--- a/E/Exodus/build_tarballs.jl
+++ b/E/Exodus/build_tarballs.jl
@@ -3,26 +3,67 @@
 using BinaryBuilder, Pkg
 
 name = "Exodus"
-version = v"0.1.1"
+version = v"8.19.0"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/gsjaardema/seacas.git", "a1da779b061fbdc750f18bcae29295dc5064cb74")
-    # GitSource("https://github.com/gsjaardema/seacas.git", "1452c325ab5d507d000397c713a5b0c4dc57bf81")
+    GitSource("https://github.com/gsjaardema/seacas.git", "cfc1edd1e1602fd1edc8da90053b66e92499c8e9"),
+    DirectorySource("bundled")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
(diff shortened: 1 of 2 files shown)


E/Exodus/build_tarballs.jl
```diff
@@ -3,26 +3,67 @@
...
-../cmake-exodus
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    \
+    -D CMAKE_CXX_COMPILER:FILEPATH=${CXX} \
+    -D CMAKE_C_COMPILER:FILEPATH=${CC} \
+    -D CMAKE_Fortran_COMPILER:FILEPATH=${FC} \
```
giordano, line 32: This is useless, this is already done in the toolchain file
```suggestion
```

E/Exodus/bundled/patches/0001-Patch-for-changes-to-some-library-names-and-NetCDF-c.patch
```diff
@@ -0,0 +1,73 @@
...
+ 		           NO_DEFAULT_PATH
+diff --git a/packages/seacas/applications/epu/EP_Internals.C b/packages/seacas/applications/epu/EP_Internals.C
+index 39700e0dcd..0e86f3d7f2 100644
+--- a/packages/seacas/applications/epu/EP_Internals.C
++++ b/packages/seacas/applications/epu/EP_Internals.C
+@@ -23,7 +23,7 @@
+ 
+ #if defined(WIN32) || defined(__WIN32__) || defined(_WIN32) || defined(_MSC_VER) ||                \
+     defined(__MINGW32__) || defined(_WIN64) || defined(__MINGW64__)
+-#include <Shlwapi.h>
```
giordano, line 40: Please upstream the changes in the name of the header files.  This is ***not*** specific to BinaryBuilder, but it's helpful for anyone building on case-sensitive file systems.

E/Exodus/build_tarballs.jl
cmhamel: Thanks, now I know!

E/Exodus/bundled/patches/0001-Patch-for-changes-to-some-library-names-and-NetCDF-c.patch
cmhamel: Ok. I'll make the header file name changes upstream.

Are you ok with me leaving the following part of the patch? Probably not a good idea to make this change upstream.

-        if (WIN32)
-            message(FATAL_ERROR "nc-config can't be used on Windows, please use CMake to install NetCDF")
-        endif()
+        #if (WIN32)
+        #    message(FATAL_ERROR "nc-config can't be used on Windows, please use CMake to install NetCDF")
+        #endif()

Thanks!

giordano: Yes, although it'd be nice not to have to do that at all eventually.

PR #6294, opened 2023-02-20
[libexif] Upgrade to 0.6.24 and build for new platforms

ashwani-rathee:
- To add support for libexif in arm based macos

### DIFF
--- a/L/libexif/build_tarballs.jl
+++ b/L/libexif/build_tarballs.jl
@@ -13,7 +13,7 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/libexif-*/
+cd $WORKSPACE/srcdir/libexif/
 if [[ "${target}" == powerpc64le-* ]] || [[ "${target}" == *-freebsd* ]]; then
     # Install `autopoint` and other tools needed by `autoreconf`
     apk add gettext-dev
--- a/L/libexif/build_tarballs.jl
+++ b/L/libexif/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "libexif"
-version = v"0.6.21"
+version = v"0.6.24"
 
 # Collection of sources required to complete build
 sources = [
-    "https://jaist.dl.sourceforge.net/project/libexif/libexif/0.6.21/libexif-0.6.21.tar.gz" =>
-    "edb7eb13664cf950a6edd132b75e99afe61c5effe2f16494e6d27bc404b287bf",
+    GitSource("https://github.com/libexif/libexif.git", 
+    "a7121eb4075df87c95004ca0c9d5385f23e6d777"),
 ]
 
 # Bash recipe for building across all platforms
@@ -24,11 +24,12 @@ update_configure_scripts
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
 make -j${nproc}
 make install
+install_license COPYING
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = supported_platforms(; experimental=true)
 
 
 # The products that we will ensure are always built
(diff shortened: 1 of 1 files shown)


L/libexif/build_tarballs.jl
```diff
@@ -24,11 +24,12 @@ update_configure_scripts
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
 make -j${nproc}
 make install
+install_license COPYING
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = supported_platforms(; experimental=true)
```
giordano, line 32:
```suggestion
platforms = supported_platforms()
```

L/libexif/build_tarballs.jl
```diff
@@ -3,17 +3,17 @@
...
 name = "libexif"
-version = v"0.6.21"
+version = v"0.6.24"
 
 # Collection of sources required to complete build
 sources = [
-    "https://jaist.dl.sourceforge.net/project/libexif/libexif/0.6.21/libexif-0.6.21.tar.gz" =>
-    "edb7eb13664cf950a6edd132b75e99afe61c5effe2f16494e6d27bc404b287bf",
+    GitSource("https://github.com/libexif/libexif.git", 
+    "a7121eb4075df87c95004ca0c9d5385f23e6d777"),
```
giordano, line 11: If you use https://github.com/libexif/libexif/releases/download/v0.6.24/libexif-0.6.24.tar.bz2 as source, you don't have to go through bootstrap

ashwani-rathee: It won't give that unstable link error? I tried the archive source but it threw error as I think github changes links and aren't stable

giordano: I guess you used the automatically generated archive https://github.com/libexif/libexif/archive/refs/tags/v0.6.24.tar.gz.

ashwani-rathee: Yeah, it errored out that's why changed to git source. Also my linux(old hdd problems) crashed while I tried building it to test before pushing it here straight twice, can you get please this going cuz I am unable to test on my end locally. I would need to go w whole setup in windows to progress which might be a while :)

giordano: > Yeah, it errored out that's why changed to git source.

But we don't accept only automatically generated archives, like https://github.com/libexif/libexif/archive/refs/tags/v0.6.24.tar.gz.  Release artifacts like https://github.com/libexif/libexif/releases/download/v0.6.24/libexif-0.6.24.tar.bz2 are fine, which is the reason why I suggested it in the first place.

> Also my linux(old hdd problems) crashed while I tried building it to test before pushing it here straight twice, can you please get this going cuz I am unable to test on my end locally.

You can use GitHub Codespaces, it's free (up to 60 hours per month, which should be plenty for simple testing): https://docs.binarybuilder.org/stable/#GitHub-Codespaces.

> I would need to go w whole setup in windows to progress which might be a while, ofcourse you are already helping out enough despite busy I understand. Really need that binary builder's debug tool rt now which helps with the script after it errors out

I already told you what to do: use https://github.com/libexif/libexif/releases/download/v0.6.24/libexif-0.6.24.tar.bz2 as source.

ashwani-rathee: Codespaces is amazing and thanks for setting it up and making things so easy, Love it !! One pointer tho, documentation doesn't seem to mention ArchiveUtils for hash, I did it the unintuitive way by changing it after build fail.

giordano: > One pointer tho, documentation doesn't seem to mention ArchiveUtils for hash, I did it the unintuitive way by changing it after build fail.

What do you mean exactly?

ashwani-rathee: I had to mention a hash in the ArchiveSource, I didn't know how to calculate that so used a online tool, but that failed for some reason when I ran build_tarballs.jl. So I ended up using the checksum mentioned in the build fail log, so would be nice to have mention of ArchiveUtils for hash calculation a little more I think.

giordano: You mean https://docs.binarybuilder.org/stable/FAQ/#What-are-those-numbers-in-the-list-of-sources?-How-do-I-get-them?-1?

ashwani-rathee: Exactly😶

ashwani-rathee: In my defense, I did search sha256 and hash in documentation search and FAQ page is mentioned but not this part specifically

PR #6297, opened 2023-02-22
[LibGit2] Update to v1.6.0

fxcoudert:
Latest version


fxcoudert:
I have upstreamed the patches and they were accepted and included in 1.6.0
Will be easier that way…

giordano:
Thanks a lot for doing that! :pray:

PR #6307, opened 2023-02-24
[LibTracyClient] Revise selected feature set.

topolarity:
This is a relatively opinionated set of (non-)features that we'd like to support initially for profiling base Julia.

Eventually, we'll probably want to provide other variants with:
  - `TRACY_ON_DEMAND` disabled
  - `TRACY_DELAYED_INIT` enabled (and possibly `TRACY_MANUAL_LIFETIME` enabled, too)

### DIFF
--- a/L/LibTracyClient/LibTracyClient@0.9.0/bundled/patches/libTracyClient-no-crash-handler.patch
+++ b/L/LibTracyClient/LibTracyClient@0.9.0/bundled/patches/libTracyClient-no-crash-handler.patch
@@ -1,0 +1,22 @@
+diff --git a/public/client/TracyProfiler.cpp b/public/client/TracyProfiler.cpp
+index ea168e4f..9287d433 100644
+--- a/public/client/TracyProfiler.cpp
++++ b/public/client/TracyProfiler.cpp
+@@ -1454,7 +1454,7 @@ Profiler::~Profiler()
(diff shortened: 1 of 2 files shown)


L/LibTracyClient/LibTracyClient@0.9.0/bundled/patches/libTracyClient-no-crash-handler.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/public/client/TracyProfiler.cpp b/public/client/TracyProfiler.cpp
```
giordano, line 1: Should this patch be upstreamed?

giordano (commented):
Looks good to me, but I'm not sure why @vchuravy stopped automerge

L/LibTracyClient/LibTracyClient@0.9.0/bundled/patches/libTracyClient-no-crash-handler.patch
topolarity: It's already available on master upstream, but unfortunately didn't make it into the v0.9 release

giordano: Ok, perfect, since this isn't a git patch I couldn't guess where it was coming from 🙂

PR #6311, opened 2023-02-25
Recipe for jlrs-ledger

Taaitaaiger:


J/JlrsLedger/build_tarballs.jl
```diff
@@ -0,0 +1,48 @@
...
+    install -Dvm 0755 "target/${rust_target}/release/libjlrs_ledger.${dlext}" "${libdir}/libjlrs_ledger.${dlext}"
+fi
+"""
+
+platforms = supported_platforms()
+
+is_supported(p) = arch(p) ∈ ("i686", "x86_64") &&
+                  (Sys.islinux(p) || nbits(p) == 64 && Sys.iswindows(p))
+
+filter!(is_supported, platforms)
```
giordano, line 37: Can slightly simplify to
```suggestion
is_supported(p) = arch(p) ∈ ("i686", "x86_64") &&
                  (Sys.islinux(p) || nbits(p) == 64 && Sys.iswindows(p))
platforms = supported_platforms(; exclude=!is_supported)
```
I'm a bit curious about only these few platforms

giordano (commented):
Thanks for your contribution!

J/JlrsLedger/build_tarballs.jl
```diff
@@ -0,0 +1,48 @@
...
+fi
+
+RUSTFLAGS=$rust_flags cargo build --release --verbose
+
+install_license LICENSE
+if [[ "${target}" == *-mingw* ]]; then
+    install -Dvm 0755 "target/${rust_target}/release/jlrs_ledger.${dlext}" "${libdir}/libjlrs_ledger.${dlext}"
+else
+    install -Dvm 0755 "target/${rust_target}/release/libjlrs_ledger.${dlext}" "${libdir}/libjlrs_ledger.${dlext}"
+fi
```
giordano, line 29: This can also be reduced to a single line with
```suggestion
install -Dvm 0755 "target/${rust_target}/release/"*jlrs_ledger".${dlext}" "${libdir}/libjlrs_ledger.${dlext}"
```

J/JlrsLedger/build_tarballs.jl
Taaitaaiger: Thanks for the advice! 

The main reason is that this library is only intended to be used in combination with jlrs, and these are the only platforms jlrs currently supports.

PR #6316, opened 2023-02-26
[FAMSA] new build FAMSA v2.2.2

marcom:
FAMSA is a program for large-scale multiple sequence alignments

### DIFF
--- a/F/FAMSA/build_tarballs.jl
+++ b/F/FAMSA/build_tarballs.jl
@@ -1,0 +1,85 @@
+using BinaryBuilder, Pkg
+
+name = "FAMSA"
+version = v"2.2.2"
+
+# url = "https://github.com/refresh-bio/FAMSA"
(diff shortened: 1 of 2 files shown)


F/FAMSA/build_tarballs.jl
```diff
@@ -0,0 +1,85 @@
...
+
+products = [
+    ExecutableProduct("famsa", :famsa)
+]
+
+dependencies = Dependency[
+]
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="1.6", preferred_gcc_version = v"11")
```
giordano, line 85: I cancelled registration because GCC 11 is too new.

marcom: Would version 10 be better?

giordano: Yes, I think that should work for Julia v1.6

PR #6318, opened 2023-02-26
[LibCURL] update to v7.88.1

fxcoudert:
https://github.com/curl/curl/releases/tag/curl-7_88_1


fxcoudert:
Just so I understand: why doesn't `CURL/build_tarballs.jl` also need a build trigger?

giordano:
There's a bootstrap problem: that one requires a build of libcurl to exist already.

fxcoudert:
@giordano I think this is ready now

PR #6319, opened 2023-02-26
Cleanup Libxc buildscript

mfherbst:


L/Libxc/Libxc_GPU/build_tarballs.jl
```diff
@@ -55,25 +55,25 @@ dependencies = [
...
+for cuda_version in [v"10.2", v"11.0", ], platform in platforms
     augmented_platform = Platform(arch(platform), os(platform); cuda=CUDA.platform(cuda_version))
     should_build_platform(triplet(augmented_platform)) || continue
 
     cuda_deps = [
         BuildDependency(PackageSpec(name="CUDA_full_jll",
                                     version=cuda_full_versions[cuda_version])),
-        RuntimeDependency(PackageSpec(name="CUDA_Runtime_jll",
-                                      version=v"0.2"), compat="0.2"),  # avoid pulling in CUDA 12 for now.
+        RuntimeDependency(PackageSpec(name="CUDA_Runtime_jll")),
```
giordano, line 76: Without changing version number, this won't change the compat.  Is the compat not ok anyway?

mfherbst: Well. This might be me getting it wrong, but as of now the pin to version 0.2 of the runtime prevents me from using CUDA.jl beyond 4.0.1, because from 4.1.0 onwards it needs CUDA_Runtime_jll > 0.4. I was hoping by removing this compat to prevent this issue.

mfherbst: @giordano Ok, I found out what's causing my observed behaviour, which is indeed not related to this package, so this PR becomes obsolete. Thanks!

PR #6321, opened 2023-02-26
[Geant4] Upgrade to version 11.1.1

peremato:
Upgrade to latest version and enable GDML option (dependency with Xerces)


G/Geant4/build_tarballs.jl
```diff
@@ -65,15 +84,19 @@ products = [
...
+    Dependency("Expat_jll"; compat="2.4.8"),
     Dependency("Xorg_libXmu_jll"),
     Dependency("Libglvnd_jll"),
+    Dependency("Xerces_jll"),
     BuildDependency("Xorg_xorgproto_jll"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"7", julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"11", julia_compat="1.6")
```
giordano, line 102: Using a too new version of GCC is a compatibility problem, we won't be able to actually use this binary on Julia v1.6.

G/Geant4/build_tarballs.jl
```diff
@@ -3,35 +3,55 @@
...
+
+install_license ../LICENSE
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(supported_platforms(; experimental=true))
-filter!(p -> !Sys.iswindows(p) && arch(p) != "armv6l", platforms)
+platforms = expand_cxxstring_abis(supported_platforms())
+platforms = filter(p -> libc(p) != "musl" && os(p) != "windows" && os(p) != "freebsd" && arch(p) != "armv6l" && cxxstring_abi(p) != "cxx03", platforms)
```
giordano, line 53: The [problem with FreeBSD](https://buildkite.com/julialang/yggdrasil/builds/1158#01868f36-d669-46f7-8c22-c355914746dc/6-7884) is
```
[19:36:59] /workspace/srcdir/geant4-v11.1.1/source/global/management/include/tls.hh:72:8: error: "No Thread Local Storage (TLS) technology supported for this platform. Use sequential build !"
```
The error message suggests we should be able to use sequential build (whatever that is) rather thank skipping this platform altogether.

Also, why skipping musl and all C++03 string ABIs?

G/Geant4/build_tarballs.jl
peremato: OK. I'll use gcc8, with is the minimal that supports C++17.

G/Geant4/build_tarballs.jl
peremato: - Yes, the problem with FeeBSD is exactly this, which would mean that I have to configure the build with different options depending on the OS. So, to avoid the extra complexity I decided to remove it from the list of platforms. This is also a platform for which the package Geant4 have never been tested.
- For musl, I was getting errors like:
  ```
/workspace/srcdir/geant4-v11.1.1/source/global/management/include/G4Backtrace.hh:90:12: fatal error: execinfo.h: No such file or directory
   90 | #  include <execinfo.h>
      |            ^~~~~~~~~~~~
compilation terminated.
  ```
  so, since Geant4 was never tested/validated I thought that it was not necessary.
- C++17 is required to build, so I was thinking that was not necessary to support C++03 strings.

giordano: https://docs.binarybuilder.org/stable/building/#Platforms

> This ABI does *not* have to do with the C++ standard used by the source code, in fact you can build a C++03 library with the C++11 `std::string` ABI and a C++11 library with the C++03 `std::string` ABI. This is achieved by appropriately setting the `_GLIBCXX_USE_CXX11_ABI` macro.

peremato: Thanks, this is what I was understanding. What I wanted to say is that any 'client' of this library/toolkit needs to be compiled with  -std=c++17, so only the C++11 string will be used in practice.

giordano: That's ***not*** what I said: the C++ string ABI has ***nothing*** to do with the C++ standard.  You can have a binary with C++03 string ABI built from a C++17 source code.

peremato: Of course, it does not harm to have them also built. I'll remove it from the filter.

peremato:
Dear @giordano, is anything else missing that prevents the merge? I would like to develop against this version as soon as possible. Thanks.

peremato:
Thanks very much!!

PR #6322, opened 2023-02-27
[FAMSA] prefer gcc version 10

marcom:


F/FAMSA/build_tarballs.jl
```diff
@@ -61,17 +68,14 @@ if [[ ${target} == *-apple-darwin* ]]; then
...
-
-make -j${nproc} CC=${CC} CXX=${CXX} PLATFORM=${PLATFORM} UNAME_S=${UNAME_S} ${DEFINE_FLAGS}
+make -j${nproc} CC=${CC} CXX=${CXX} PLATFORM=${PLATFORM} UNAME_S=${UNAME_S}
 install -Dvm 755 "famsa${exeext}" "${bindir}/famsa${exeext}"
 install_license LICENSE
 """
 
-platforms = supported_platforms(; exclude = p -> Sys.iswindows(p) || Sys.isfreebsd(p))
+platforms = supported_platforms(;
+    exclude = p -> Sys.iswindows(p) || Sys.isfreebsd(p) || (arch(p) ∉ ["x86_64", "aarch64", "powerpc64le"])
```
giordano, line 77:
```suggestion
    exclude = p -> Sys.iswindows(p) || Sys.isfreebsd(p) || nbits(p) == 32)
```
is a bit simpler than listing all architectures.

If this package ***requires*** GCC 11 to build correctly than it's OK to use it, I guess, but be aware this won't work at all with Julia v1.6, which comes with GCC 10 libraries, if I remember correctly.  Up to you.

marcom: I think that julia-1.6 support is probably more important than the i686, armv6l, and armv7l architectures for this kind of program. If someone really wants to run this program on i686 or an old arm arch we can try and fix it then. So i guess we leave it like this for now.

PR #6324, opened 2023-02-27
[Baobzi] Add Baobzi_jll package (v0.9.6)

blackwer:
Adds build of https://github.com/flatironinstitute/baobzi shared lib.

### DIFF
--- a/B/Baobzi/build_tarballs.jl
+++ b/B/Baobzi/build_tarballs.jl
@@ -7,7 +7,7 @@ version = v"0.9.6"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/flatironinstitute/baobzi.git", "7b602bed96e504c927fa3adc619a1d748a07dad8")
+    GitSource("https://github.com/flatironinstitute/baobzi.git", "dab45ba8bc0f08fc9cb364f3d1c776550940ac14")
 ]
 
 # Bash recipe for building across all platforms
--- a/B/Baobzi/build_tarballs.jl
+++ b/B/Baobzi/build_tarballs.jl
@@ -39,4 +39,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"12.1.0")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"7.1.0")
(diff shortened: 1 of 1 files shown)


B/Baobzi/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "freebsd"; ),
+    # Platform("i686", "windows"; ),
+    # Platform("x86_64", "windows"; )
+]
```
giordano, line 45:
```suggestion
platforms = supported_platforms(; exclude=Sys.iswindows)
```

B/Baobzi/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir
+cmake -DCMAKE_INSTALL_PREFIX=$prefix \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DBAOBZI_BUILD_TESTS=OFF \
+      -DBAOBZI_BUILD_EXAMPLES=OFF \
+      -DBAOBZI_SET_ARCH=OFF \
+      baobzi
+make -j4
```
giordano, line 23:
```suggestion
make -j${nproc}
```

B/Baobzi/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+products = [
+    LibraryProduct("libbaobzi", :libbaobzi)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"12.1.0")
```
giordano, line 58: For maximum compatibility, we want to use the _oldest_ version of GCC that allows us to build the code.  Too recent versions are problematic: https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html.

blackwer: I've had performance/c++ compatibility issues with older gcc versions in the past. From my limited testing, I think this is mostly a non-issue now. I've dropped the version as suggested.

giordano:
Thanks for your contribution!

PR #6342, opened 2023-03-05
[Qt65Compat] New package (version 6.4.2)

barche:
Needed for GUIs that want to include legacy Qt5 elements in Qt6.


barche:
Should be OK!

PR #6349, opened 2023-03-06
MAGEMin v1.3.0

NicolasRiel:
- Added metapelite database with MnO (White et al., 2014), use the flag --db=ig or --db=mp to change database from igneous to metapelite
- Improved GUI (PlotPseudosection.mlapp): added option to change database; added panel to plot isocontour (mode and end-member fractions); added option to save pseudosection with isocontour in vector format (eps)
- Updated python scripts to translate database from TC to MAGEMin (check ./python/SS_Analytical_multi_DB.ipynb)
- Documentation udpate


NicolasRiel:
Hi, 
It seems like the registry cannot be downloaded.  How can I fix this issue? @giordano

PR #6352, opened 2023-03-06
libzip

GunnarFarneback:
Build recipe for libzip.

### DIFF
--- a/L/libzip/build_tarballs.jl
+++ b/L/libzip/build_tarballs.jl
@@ -1,0 +1,44 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "libzip"
+version = v"1.9.2"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://libzip.org/download/libzip-$(version).tar.gz", "fd6a7f745de3d69cf5603edc9cb33d2890f0198e415255d0987a0cf10d824c6f")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd libzip-1.9.2/
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+make
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libzip", :libzip),
+    ExecutableProduct("zipmerge", :zipmerge),
+    ExecutableProduct("ziptool", :ziptool),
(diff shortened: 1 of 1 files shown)


L/libzip/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+    ExecutableProduct("zipcmp", :zipcmp)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a"))
+    Dependency(PackageSpec(name="Bzip2_jll", uuid="6e34b625-4abd-537c-b88f-471c36dfa7a0"))
+    Dependency(PackageSpec(name="XZ_jll", uuid="ffd25f8a-64ca-5728-b0f7-c24cf3aae800"))
+    Dependency(PackageSpec(name="Zstd_jll", uuid="3161d3a3-bdf6-5164-811a-617609db77b4"))
+    Dependency(PackageSpec(name="MbedTLS_jll", uuid="c8ffd9c3-330d-5841-b78e-0817d7145fa1"))
```
giordano, line 40: Do we really need mbedtls? Sigh. In any case you have to add a compat bound restricted to a minor version and be aware that this will work only with a limited set of Julia versions

GunnarFarneback: There are alternatives. The libzip INSTALL.md has the following to say about crypto dependencies:
```
For AES (encryption) support, you need one of these cryptographic libraries,
listed in order of preference:

- Apple's CommonCrypto (available on macOS and iOS)
- [GnuTLS](https://www.gnutls.org/) and [Nettle](https://www.lysator.liu.se/~nisse/nettle/) (at least nettle 3.0)
- [mbed TLS](https://tls.mbed.org/)
- [OpenSSL](https://www.openssl.org/) >= 1.0.
- Microsoft Windows Cryptography Framework
```
Would another of those be nicer?

giordano: I think gnutls might be the least bad in terms of compatibility (at least, it isn't widely used here, so less likely to conflict with something else, and they seem to favour it too). Also, I think you may want to restrict it to freebsd and Linux with the `platforms` keyword argument

GunnarFarneback: What is the problem with the other platforms? What I really need is a library that can read AES encrypted, password protected zip files on Linux and Windows.

giordano: Well, on macOS and Windows you should be able to use native solutions, no?

GunnarFarneback: I wouldn't know. Do I need to do anything for those builds to pick up native solutions?

giordano: Don't think so, for example I presume CommonCrypto is a framework which comes with the macOS SDK.

GunnarFarneback: Ok, I'll try to switch to GnuTLS.

GunnarFarneback: MacOS found CommonCrypto in the configure step. Windows didn't search for it's library but it did build `zip_crypto_win.c`, so I assume that's a good sign.

PR #6356, opened 2023-03-08
[Cares] Update to version 1.19.0

marcom:
From the changelog:

```
Security:

    Low. Stack overflow in ares_set_sortlist() which is used during
    c-ares initialization and typically provided by an administrator and
    not an end user.
```

See https://c-ares.org/changelog.html


giordano:
You have to trigger the build again because registration failed

PR #6368, opened 2023-03-09
[SuiteSparse@7] Try harder to disable CUDA

fxcoudert:
Trying to use `SUITESPARSE_CUDA` to disable the building of `_cuda` libraries, as suggested in https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/308#issuecomment-1462203259

May not work, I think building these stub libraries is actually by design :(


fxcoudert:
Nope, still builds `_cuda` variants of libcholmod and libspqr. It appears they cannot be disabled.

PR #6370, opened 2023-03-09
bump Pythia8 version

Moelf:


giordano:
Wow, this recipe was very old

PR #6373, opened 2023-03-09
[NVTX] Build CUDA stream naming functions

simonbyrne:
I'm not sure if this is the right way to do this. May need to add a build time dependency on CUDA?

cf https://github.com/JuliaGPU/NVTX.jl/issues/27

cc @maleadt


simonbyrne:
Okay, it looks like that at the very least I need CUDA as a build time dependency. How do i do that?

maleadt (approved):
LGTM. I guess this doesn't put a runtime dependency on libcuda, seeing how Yggdrasil is still able to dlopen libnvtx.

simonbyrne:
No, I think `cuda.h` is included just for the types.

PR #6385, opened 2023-03-11
[LaMEM] version 1.2.3

boriskaus:
updated version

### DIFF
--- a/L/LaMEM/build_tarballs.jl
+++ b/L/LaMEM/build_tarballs.jl
@@ -73,6 +73,7 @@ platforms = filter(p -> !(p["mpi"] == "openmpi" && arch(p) == "armv6l" && libc(p
 # MPItrampoline
 platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && libc(p) == "musl"), platforms)
 platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && Sys.isfreebsd(p)), platforms)
+platforms = filter(p -> !(p["mpi"] == "mpich"), platforms)
 
 # The products that we will ensure are always built
 products = [
--- a/L/LaMEM/build_tarballs.jl
+++ b/L/LaMEM/build_tarballs.jl
@@ -83,7 +83,7 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("PETSc_jll"; compat=PETSc_COMPAT_VERSION),
-    Dependency("CompilerSupportLibraries_jll", compat="1.0.2")
+    Dependency("CompilerSupportLibraries_jll")
 ]
 append!(dependencies, platform_dependencies)
 
--- a/L/LaMEM/build_tarballs.jl
+++ b/L/LaMEM/build_tarballs.jl
@@ -6,14 +6,14 @@ const YGGDRASIL_DIR = "../.."
 include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
 
 name = "LaMEM"
-version = v"1.2.2"
+version = v"1.2.3"
 
 PETSc_COMPAT_VERSION = "3.16.8" # Note: this is the version of the PETSc_jll package, which is sometimes larger than the PETSc version  
 
 # Collection of sources required to complete build
 sources = [
     GitSource("https://bitbucket.org/bkaus/lamem.git", 
-    "9bfd15fa11c3fd70b2d6e31812a8bd6c08299026")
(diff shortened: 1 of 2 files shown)


boriskaus:
@vchuravy, @giordano: this was working just fine 2 weeks ago in [this commit](https://github.com/JuliaPackaging/Yggdrasil/pull/6314), so I guess something must have changed upstream (particular package versions?). 

If I look at the different builds, here some of the changes:
- MPICH_jll v4.0.2+5 -> MPICH_jll v4.1.0+1
- CompilerSupportLibraries_jll v1.0.2+1 -> CompilerSupportLibraries_jll v1.0.4
- OpenMPI_jll v4.1.3+3 -> OpenMPI_jll v4.1.5+0

Is there a way to force the build to use these previous version of the packages? Since the compilation of PETSc is rather complex, it is unclear to me which of the many upstream packages have the issues.

If I do this:
```Julia
dependencies = [
    Dependency("PETSc_jll"; compat=PETSc_COMPAT_VERSION),
    Dependency("CompilerSupportLibraries_jll", compat="1.0.2"),
    Dependency("MPICH_jll", compat="4.0.2"),
    Dependency("OpenMPI_jll", compat="4.1.3")
]
```
it fails:
```Julia
MacBook-Pro-2:LaMEM kausb$ julia build_tarballs.jl --verbose --debug x86_64-apple-darwin
┌ Warning: Running Registrator on a Julia version that does not support weak dependencies. Weak dependencies will not be registered.
└ @ RegistryTools ~/.julia/packages/RegistryTools/tHkoa/src/RegistryTools.jl:18
[ Info: Building for x86_64-apple-darwin
    Updating registry at `~/.julia/registries/General.toml`
┌ Info: Using cached git repository
│   url = "https://bitbucket.org/bkaus/lamem.git"
└   repo_path = "/Users/kausb/.julia/packages/BinaryBuilderBase/Toc7t/deps/downloads/clones/lamem.git-4160cef5befec547689fc5cc4eca46f89bde80c30297409063782ebe2ad29b90"
[ Info: Checking lamem.git-4160cef5befec547689fc5cc4eca46f89bde80c30297409063782ebe2ad29b90 out to lamem...
ERROR: LoadError: it is invalid to specify multiple packages with the same name: `MPICH_jll [7cb0a576]`
```

boriskaus:
ok that did the trick

L/LaMEM/build_tarballs.jl
```diff
@@ -73,6 +73,7 @@ platforms = filter(p -> !(p["mpi"] == "openmpi" && arch(p) == "armv6l" && libc(p
 # MPItrampoline
 platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && libc(p) == "musl"), platforms)
 platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && Sys.isfreebsd(p)), platforms)
+platforms = filter(p -> !(p["mpi"] == "mpich"), platforms)
```
giordano, line 76: Removing MPICH doesn't sound like a solution, just ignoring a problem

giordano: A better ***solution***, perhaps, is to come up with an idea of how to pass from here a compat value for MPICH in https://github.com/JuliaPackaging/Yggdrasil/blob/95e2c9f55ef6c82211d039392d7c7febc436c3ea/platforms/mpi.jl#L47 in a backward compatible way so that it doesn't break other existing recipes.

boriskaus: Ok that did the job - have a look to see whether you are happy with the implementation.

platforms/mpi.jl
```diff
@@ -50,16 +50,33 @@ mpi_abis = (
...
     all_platforms = AbstractPlatform[]
     dependencies = []
     for (abi, pkg, compat, f) in mpi_abis
+        compat_version = compat;
+        
+        # set specific versions of MPI packages
+        if (abi=="OpenMPI" && !isnothing(OpenMPI_compat)) compat_version = OpenMPI_compat; end
+        if (abi=="MPICH" && !isnothing(MPICH_compat)) compat_version = MPICH_compat; end
+        if (abi=="MicrosoftMPI" && !isnothing(MicrosoftMPI_compat)) compat_version = MicrosoftMPI_compat; end
+        if (abi=="MPItrampoline" && !isnothing(MPItrampoline_compat)) compat_version = MPItrampoline_compat; end
```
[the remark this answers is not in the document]
boriskaus: No why? If no keyword is specified the default `compat` will be used as far as I can tell.

PR #6388, opened 2023-03-12
[OpenEXR] version 3.1.4

Astro-mh:
Updates OpenEXR to version 3.1.6 and switches to GitSource method: https://github.com/AcademySoftwareFoundation/openexr/releases/tag/v3.1.6

### DIFF
--- a/O/OpenEXR/build_tarballs.jl
+++ b/O/OpenEXR/build_tarballs.jl
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "OpenEXR"
-version = v"3.1.1"
+version = v"3.1.6"
 
 # Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://github.com/AcademySoftwareFoundation/openexr/archive/refs/tags/v3.1.1.tar.gz", "045254e201c0f87d1d1a4b2b5815c4ae54845af2e6ec0ab88e979b5fdb30a86e")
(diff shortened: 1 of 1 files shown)


Astro-mh:
Are the details of the platforms used by buildkite available somewhere? The musl errors here are a bit odd as its failing on code that was specifically added upstream to fix musl builds.

Astro-mh:
> What do you want to know specifically?

I've found it now, via the Artifacts.toml in BinaryBuilderBase. I'll try testing locally to fix this later today.

> I have the feeling it was broken by [AcademySoftwareFoundation/openexr@7e4630f](https://github.com/AcademySoftwareFoundation/openexr/commit/7e4630f9773810297831b9a31c6fecab51623e12), which addressed glibc systems, not musl, and matches the fact that it worked in 3.1.1 but it's broken in 3.1.6.

Could well be, the weird thing is that the `fpregs->cwd` that's causing the error looks like it's correct for musl, but I'll find out when I'm testing

Thanks

Astro-mh:
So the issue with the aarch64 builds was that AcademySoftwareFoundation/openexr@436fcd2 assumes that all NEON vector modes are defined. For gcc this is only true since gcc-mirror/gcc@66f206b which was first released in gcc 12.1.0.

I fixed this by requiring gcc v12 for all platforms, is that OK or is there a better per-platform method that could be used?

giordano:
C++ code compiled with GCC 12 won't work with any released version of Julia, I think....

Astro-mh:
By downgrading to 3.1.4 we now have a successful build. This fixes the 2 issues that I had with the previous version: we now have an aarch64-apple-darwin build and the fix for [CVE-2021-45942](https://nvd.nist.gov/vuln/detail/CVE-2021-45942) is included.

PR #6389, opened 2023-03-14
Wizard recipe: t8code-v1.1.2

jmark:

* Package name: t8code
* Version: v1.1.2


T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,51 @@
...
+sources = [
+    ArchiveSource("https://github.com/DLR-AMR/t8code/releases/download/v$(version)/t8code_v$(version).tar.gz", "01ceae6dd0ed596cf1abb5bc78caa77b5da94ca60d080aae3ccf0ba3c6464deb")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd t8code/
+./configure --enable-mpi CFLAGS=-O3 CXXFLAGS=-O3 CC=mpicc CXX=mpicxx --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j
```
sloede, line 18:
```suggestion
make -j${nproc}
```
Based on https://docs.binarybuilder.org/stable/build_tips/#Autoconfigure-builds

T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,51 @@
...
+products = [
+    LibraryProduct("libt8", :libt8),
+    LibraryProduct("libsc", :libsc),
+    LibraryProduct("libp4est", :libp4est)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="MPICH_jll", uuid="7cb0a576-ebde-5e09-9194-50597f1243b4"))
+    Dependency(PackageSpec(name="MPItrampoline_jll", uuid="f1f71cc9-e9ae-5b93-9b94-4fe0e1ad3748"))
```
vchuravy, line 44: This is slighlty over-optimistic. See L/LAMMPS for an example of how to use the MPI platform

sloede:
@jmark With https://github.com/DLR-AMR/t8code/pull/473, your build script also works for me with MUSL (at least on some systems).

With the addition of something like
https://github.com/JuliaPackaging/Yggdrasil/blob/83900358eee9ba66384031faecdd8edde2d27029/P/P4est/build_tarballs.jl#L24-L25
it also works on FreeBSD. In general, I think if you borrow some more ideas from the p4est `build_tarballs.jl`, you might get it to work also on, e.g., Windows.

sloede:
@jmark With the fixes in , I can build t8code using BB for the following platforms:
```julia
platforms = [
    Platform("i686", "linux"; libc = "glibc"),
    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
    Platform("powerpc64le", "linux"; libc = "glibc"),
    Platform("i686", "linux"; libc = "musl"),
    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
    Platform("x86_64", "macos"; ),
    Platform("aarch64", "macos"; ),
    Platform("x86_64", "freebsd"; )
]
```
That is, everything but Windows. On Windows, there seems to be a general issue with building MPI programs, since `cc ... -lmsmpi` yields an error about incompatible MPI licenses already during the configure step, which I do not understand (or know how to fix) yet...

T/t8code/build_tarballs.jl
giordano: > What do you mean by "over-optimistic"?

Smashing all MPI implementations together is bound to end up very bad: you're just linking to a single implementation, the only one surviving this: https://buildkite.com/julialang/yggdrasil/builds/1409#0186e07f-9078-4309-aa3a-1c1b7fce553c/637-1751.  Which one is it it's everybody's guess.  What Valentin suggested is a more straightforward way to deal with the different implementations.

jmark: I left out `MPItrampoline_jll` for now.

jmark:
Thank you very much for your advice, @sloede ! Revised the build recipe. Besides Windows I could build t8code for all default targets.

sloede:
> Thank you very much for your advice, @sloede ! Revised the build recipe. Besides Windows I could build t8code for all default targets.

No worries.

Should we wait for https://github.com/DLR-AMR/t8code/pull/476 to also build t8code for Windows? Alternatively, if the review process for t8code is too lengthy (or the proposed change too controversial), we can always just provide a local patch for Windows here (I did this for p4est too once) and then wait for it to be merged upstream.

jmark:
> Should we wait for [DLR-AMR/t8code#476](https://github.com/DLR-AMR/t8code/pull/476) to also build t8code for Windows?
We can wait and see what happens. Thanks for your effort and investment in this. :)

sloede:
With https://github.com/JuliaPackaging/Yggdrasil/pull/6396 being merged, we could also try to use the MPI aware setup that @vchuravy mentioned, while we're waiting on the t8code PR...

T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,68 @@
...
+export LDFLAGS="-L${libdir}"
+# Use MPI including MPI I/O on all other platforms
+export CC="mpicc"
+export CXX="mpicxx"
+mpiopts="--enable-mpi"
+# Configure, build, install
+# Note: BLAS is disabled since it is only needed for SC if it is used outside of p4est
+./configure --disable-static --without-blas ${mpiopts} CFLAGS='-O3 -std=c99' CXXFLAGS='-O3 -std=c++11' --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j${nproc}
+make install
```
sloede, line 31:
```suggestion
cd t8code/

# Set default preprocessor and linker flags
# Note: This is *crucial* for Windows builds as otherwise the wrong libraries are picked up!
export CPPFLAGS="-I${includedir}"
export LDFLAGS="-L${libdir}"
export CFLAGS="-O3 -std=c99"
export CXXFLAGS="-O3 -std=c++11"

# Set necessary flags for FreeBSD
if [[ "${target}" == *-freebsd* ]]; then
  export LIBS="-lm"
fi

# Set necessary flags for Windows and non-Windodws systems
FLAGS=()
if [[ "${target}" == *-mingw* ]]; then
  # Set linker flags only at build time (see https://docs.binarybuilder.org/v0.3/troubleshooting/#Windows)
  FLAGS+=(LDFLAGS="$LDFLAGS -no-undefined")
  # Configure does not find the correct Fortran compiler
  export F77="f77"
  # Link against ws2_32 to use the htonl function from winsock2.h
  export LIBS="-lmsmpi -lws2_32"
  # Disable MPI I/O on Windows since it causes p4est to crash
  mpiopts="--enable-mpi --disable-mpiio"
  # Linker looks for libmsmpi instead of msmpi, copy existing symlink
  cp -d ${libdir}/msmpi.dll ${libdir}/libmsmpi.dll
else
  # Use MPI including MPI I/O on all other platforms
  export CC="mpicc"
  export CXX="mpicxx"
  mpiopts="--enable-mpi"
fi

# Run configure
./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} --disable-static --without-blas ${mpiopts}

# Build & install
make -j${nproc} "${FLAGS[@]}"
make install
```

This is what makes the build script work for me for Windows x86_64. I'll just put it here such that you are not dependent on me to finalize this.

T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,94 @@
...
+  mpiopts="--enable-mpi"
+fi
+
+# Run configure
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} --disable-static --without-blas ${mpiopts}
+
+# Build & install
+make -j${nproc} "${FLAGS[@]}"
+make install
+"""
```
sloede, line 56:
```suggestion

# On Windows: copy DLLs to make them findable by BinaryBuilder
if [[ "${target}" == *-mingw* ]]; then
  cp ${libdir}/libp4est*.dll ${libdir}/libp4est.dll
  cp ${libdir}/libsc*.dll ${libdir}/libsc.dll
  cp ${libdir}/libt8*.dll ${libdir}/libt8.dll
fi
"""
```
@jmark can you please try if this works for you (it seemed to work for me)

sloede:
Nice! It seems to build properly on Windows now. Before merging this, let's try once more to use the new, sophisticated way of building MPI-based packages, what do you think @jmark ?

T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,101 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd t8code/
+
+# Set default preprocessor and linker flags
+# Note: This is *crucial* for Windows builds as otherwise the wrong libraries are picked up!
+export CPPFLAGS="-I${includedir}"
+export LDFLAGS="-L${libdir}"
+export CFLAGS="-O3 -std=c99"
```
giordano, line 22: Sounds like the upstream package should set the standard flags. Users should never have to do that themselves.

jmark: In general I agree. Support for Windows on t8code is still in its infancy and will be improved in the future. For now, we want to focus on shipping a `t8code_jll` for all major platforms.

jmark: @sloede By sophisticated way, you mean having `MPITrampoline` as an dependency?

sloede: @jmark Yes, basically trying to use a setup like it is used for p4est:
https://github.com/JuliaPackaging/Yggdrasil/blob/d55ac06f2ab98afc070f1e0e6fd5a5d300578c85/P/P4est/build_tarballs.jl#L66-L89

@giordano We are using an ancient compiler here that is not officially supported by t8code. I'll try switching to a minimum gcc v8 (like we do for p4est), which might allow us to remove these flags.

giordano: That's good, but if a source code requires a certain language standard, its build system should set it up correctly, whatever is the default standard used by the compilers. Mine was just a general comment, outside of the scope of this PR.

sloede: @giordano We we able to remove the language standard flags after switchting to GCC 8.

T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,112 @@
...
+  # Set linker flags only at build time (see https://docs.binarybuilder.org/v0.3/troubleshooting/#Windows)
+  FLAGS+=(LDFLAGS="$LDFLAGS -no-undefined")
+  # Configure does not find the correct Fortran compiler
+  export F77="f77"
+  # Link against ws2_32 to use the htonl function from winsock2.h
+  export LIBS="-lmsmpi -lws2_32"
+  # Disable MPI I/O on Windows since it causes p4est to crash
+  mpiopts="--enable-mpi --disable-mpiio"
+  # Linker looks for libmsmpi instead of msmpi, copy existing symlink
+  cp -d ${libdir}/msmpi.dll ${libdir}/libmsmpi.dll
```
giordano, line 48: Uhm, the tarball ends up containing msmpi, which looks wrong.

T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,112 @@
...
+  export LIBS="-lm"
+fi
+
+# Set necessary flags for Windows and non-Windodws systems
+FLAGS=()
+if [[ "${target}" == *-mingw* ]]; then
+  # Set linker flags only at build time (see https://docs.binarybuilder.org/v0.3/troubleshooting/#Windows)
+  FLAGS+=(LDFLAGS="$LDFLAGS -no-undefined")
+  # Configure does not find the correct Fortran compiler
+  export F77="f77"
```
giordano, line 42: `F77`, really? :smiley:

T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,112 @@
...
+# Build & install
+make -j${nproc} "${FLAGS[@]}"
+make install
+
+# On Windows: copy DLLs to make them findable by BinaryBuilder
+if [[ "${target}" == *-mingw* ]]; then
+  cp ${libdir}/libp4est*.dll ${libdir}/libp4est.dll
+  cp ${libdir}/libsc*.dll ${libdir}/libsc.dll
+  cp ${libdir}/libt8*.dll ${libdir}/libt8.dll
+fi
```
giordano, line 68: What's the problem?  If libraries have different names on different products, just list them: https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct.  You shouldn't move files around.

T/t8code/build_tarballs.jl
sloede, line 42:
```suggestion
```
This does not seem to be necessary anymore...

T/t8code/build_tarballs.jl
sloede: I know. However, when we don't do this, the build process works without errors but for some reason no dynamic libraries are created. I've scratched my head as well, but I really don't know how to debug this any further, especially since no errors are thrown. We use the same strategy for building p4est with BB successfully, but if you have another idea, I'm all ears!

T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,110 @@
...
```
sloede, line 66:
```suggestion
```

T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,110 @@
...
+platforms = filter(p -> !(p["mpi"] == "openmpi" && arch(p) == "armv6l" && libc(p) == "glibc"), platforms)
+# MPItrampoline
+platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && libc(p) == "musl"), platforms)
+platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && Sys.isfreebsd(p)), platforms)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libt8", :libt8),
+    LibraryProduct("libsc", :libsc),
+    LibraryProduct("libp4est", :libp4est)
```
sloede, line 98:
```suggestion
# The products that we will ensure are always built
# Note: the additional, non-canonical library names are required for the Windows build
products = [
    LibraryProduct(["libt8", "libt8-1-1-0-207-d6a74"], :libt8),
    LibraryProduct(["libsc", "libsc-2-8-1-5-0b70"], :libsc),
    LibraryProduct(["libp4est", "libp4est-2-2-259-ec120"], :libp4est)
```

T/t8code/build_tarballs.jl
sloede: Thank for the link, I've fixed it in 9c4a82d3a5f6c84796a671d391f08676874ef204.

T/t8code/build_tarballs.jl
sloede: Right, this is not required anymore. Fixed in d3c1bad884f89a844f39c323d72fd621d9e5816a.

T/t8code/build_tarballs.jl
sloede: These are the pertinent messages, I believe, I get when build without linking to `libmsmpi.dll` to `msmpi.sll`:
```
[05:58:43] *** Warning: This system cannot link to static lib archive p4est/src/libp4est.la.
[05:58:43] *** I have the capability to make that library automatically link in when
[05:58:43] *** you link to this library.  But I can only do this if you have a
[05:58:43] *** shared version of the library, which you do not appear to have.
[05:58:43]
[05:58:43] *** Warning: This system cannot link to static lib archive sc/src/libsc.la.
[05:58:43] *** I have the capability to make that library automatically link in when
[05:58:43] *** you link to this library.  But I can only do this if you have a
[05:58:43] *** shared version of the library, which you do not appear to have.
[05:58:43]
[05:58:43] *** Warning: linker path does not have real file for library -lmsmpi.
[05:58:43] *** I have the capability to make that library automatically link in when
[05:58:43] *** you link to this library.  But I can only do this if you have a
[05:58:43] *** shared version of the library, which you do not appear to have
[05:58:43] *** because I did check the linker path looking for a file starting
[05:58:43] *** with libmsmpi and none of the candidates passed a file format test
[05:58:43] *** using a file magic. Last file checked: /workspace/destdir/bin/../../artifacts/70c46814df14fc801299ae69f463aff3f03f9034/bin/libz.dll
[05:58:43] *** The inter-library dependencies that have been dropped here will be
[05:58:43] *** automatically added whenever a program is linked with this library
[05:58:43] *** or is declared to -dlopen it.
[05:58:43]
[05:58:43] *** Since this library must not contain undefined symbols,
[05:58:43] *** because either the platform does not support them or
[05:58:43] *** it was explicitly requested with -no-undefined,
[05:58:43] *** libtool will only create a static version of it.
```

sloede: I also tried setting
```shell
export LIBS="/workspace/x86_64-w64-mingw32-libgfortran5-cxx11-mpi+microsoftmpi/artifacts/4dec3ecde33187b4e2fcf816ec20e27d18c2c43e/bin/msmpi.dll -lmsmpi -lws2_32"
export LDFLAGS="-L${libdir} -L$WORKSPACE/x86_64-w64-mingw32-libgfortran5-cxx11-mpi+microsoftmpi/artifacts/4dec3ecde33187b4e2fcf816ec20e27d18c2c43e/bin"
```
but to no avail - it still produces the warning above and does not create the shared library :-/

jmark:

T/t8code/build_tarballs.jl
giordano: > ```sh
> export LIBS="/workspace/x86_64-w64-mingw32-libgfortran5-cxx11-mpi+microsoftmpi/artifacts/4dec3ecde33187b4e2fcf816ec20e27d18c2c43e/bin/msmpi.dll -lmsmpi -lws2_32"
> ```

There are a couple of errors there:

1. `${libdir}/msmpi.dll` is much simpler than hardcoding the full path which points to an artifact directory including its tree hash
2. You either use the absolute path of a shared library ***or*** link it with `-lname`, not both

giordano: Also,
```console
sandbox:${WORKSPACE} # cat hello.c 
#include <mpi.h>
#include <stdio.h>

int main(int argc, char** argv) {
    // Initialize the MPI environment
    MPI_Init(NULL, NULL);

    // Get the number of processes
    int world_size;
    MPI_Comm_size(MPI_COMM_WORLD, &world_size);

    // Get the rank of the process
    int world_rank;
    MPI_Comm_rank(MPI_COMM_WORLD, &world_rank);

    // Get the name of the processor
    char processor_name[MPI_MAX_PROCESSOR_NAME];
    int name_len;
    MPI_Get_processor_name(processor_name, &name_len);

    // Print off a hello world message
    printf("Hello world from processor %s, rank %d out of %d processes\n",
           processor_name, world_rank, world_size);

    // Finalize the MPI environment.
    MPI_Finalize();
}
sandbox:${WORKSPACE} # cc -o hello.exe hello.c -lmsmpi -L${libdir}
sandbox:${WORKSPACE} # ${target}-objdump -x hello.exe | grep DLL
        DLL Name: KERNEL32.dll
        DLL Name: msvcrt.dll
        DLL Name: msmpi.dll
```
The linker here seems to be completely happy with using `-lmsmpi` to link to `msmpi.dll`.

sloede: > There are a couple of errors there:
> 
>     1. `${libdir}/msmpi.dll` is much simpler than hardcoding the full path which points to an artifact directory including its tree hash
> 
>     2. You either use the absolute path of a shared library _**or**_ link it with `-lname`, not both

Thanks for pointing that out. However, in that case there's another problem, where the autotools-generated `libtool` somehow "drops" the absolute path. For some reasons I don't understand, when called from the Makefile as
```
/bin/sh ./libtool  --tag=CC   --mode=link cc  -O3 -release 2.8.1.5-0b70 -L/workspace/destdir/bin -no-undefined -o src/libsc.la -rpath /workspace/destdir/lib   src/libsc_la-sc.lo src/libsc_la-sc_mpi.lo src/libsc_la-sc_containers.lo src/libsc_la-sc_avl.lo src/libsc_la-sc_string.lo src/libsc_la-sc_unique_counter.lo src/libsc_la-sc_options.lo src/libsc_la-sc_functions.lo src/libsc_la-sc_statistics.lo src/libsc_la-sc_ranges.lo src/libsc_la-sc_io.lo src/libsc_la-sc_amr.lo src/libsc_la-sc_search.lo src/libsc_la-sc_sort.lo src/libsc_la-sc_dmatrix.lo src/libsc_la-sc_blas.lo src/libsc_la-sc_lapack.lo src/libsc_la-sc_bspline.lo src/libsc_la-sc_flops.lo src/libsc_la-sc_random.lo src/libsc_la-sc_getopt.lo src/libsc_la-sc_getopt1.lo src/libsc_la-sc_polynom.lo src/libsc_la-sc_keyvalue.lo src/libsc_la-sc_refcount.lo src/libsc_la-sc_warp.lo src/libsc_la-sc_shmem.lo src/libsc_la-sc_allgather.lo src/libsc_la-sc_reduce.lo src/libsc_la-sc_notify.lo iniparser/src_libsc_la-dictionary.lo iniparser/src_libsc_la-iniparser.lo libb64/src_libsc_la-cencode.lo libb64/src_libsc_la-cdecode.lo  -lz /workspace/destdir/bin/msmpi.dll -lws2_32
```
`libtool` internally calls
```
libtool: link: cc -shared  src/.libs/libsc_la-sc.o src/.libs/libsc_la-sc_mpi.o src/.libs/libsc_la-sc_containers.o src/.libs/libsc_la-sc_avl.o src/.libs/libsc_la-sc_string.o src/.libs/libsc_la-sc_unique_counter.o src/.libs/libsc_la-sc_options.o src/.libs/libsc_la-sc_functions.o src/.libs/libsc_la-sc_statistics.o src/.libs/libsc_la-sc_ranges.o src/.libs/libsc_la-sc_io.o src/.libs/libsc_la-sc_amr.o src/.libs/libsc_la-sc_search.o src/.libs/libsc_la-sc_sort.o src/.libs/libsc_la-sc_dmatrix.o src/.libs/libsc_la-sc_blas.o src/.libs/libsc_la-sc_lapack.o src/.libs/libsc_la-sc_bspline.o src/.libs/libsc_la-sc_flops.o src/.libs/libsc_la-sc_random.o src/.libs/libsc_la-sc_getopt.o src/.libs/libsc_la-sc_getopt1.o src/.libs/libsc_la-sc_polynom.o src/.libs/libsc_la-sc_keyvalue.o src/.libs/libsc_la-sc_refcount.o src/.libs/libsc_la-sc_warp.o src/.libs/libsc_la-sc_shmem.o src/.libs/libsc_la-sc_allgather.o src/.libs/libsc_la-sc_reduce.o src/.libs/libsc_la-sc_notify.o iniparser/.libs/src_libsc_la-dictionary.o iniparser/.libs/src_libsc_la-iniparser.o libb64/.libs/src_libsc_la-cencode.o libb64/.libs/src_libsc_la-cdecode.o   -L/workspace/destdir/bin -lz -lws2_32  -O3   -o src/.libs/libsc-2-8-1-5-0b70.dll -Wl,--enable-auto-image-base -Xlinker --out-implib -Xlinker src/.libs/libsc.dll.a
```

As you can see, `-lz -lws2_32` is passed on to the linker, but the `msmpi.dll` part is dropped.

> The linker here seems to be completely happy with using `-lmsmpi` to link to `msmpi.dll`.

Right. If in the same call as above, I just call the `cc -shared ...` command with `-lmsmpi`, it works and the dll is successfully built 🤷‍♂️

I just pushed another attempt that is not pretty but that works without having to copy `msmpi.dll`.

T/t8code/build_tarballs.jl
```diff
@@ -0,0 +1,104 @@
...
+
+# Set necessary flags for FreeBSD
+if [[ "${target}" == *-freebsd* ]]; then
+  export LIBS="-lm"
+fi
+
+# Set necessary flags for Windows and non-Windodws systems
+FLAGS=()
+if [[ "${target}" == *-mingw* ]]; then
+  # Pass -lmsmpi explicitly to linker as the absolute library path gets lost sometimes
```
sloede, line 39:
```suggestion
  # Pass -lmsmpi explicitly to linker as the absolute library path specified in LIBS below is not always propagated properly
```

T/t8code/build_tarballs.jl
sloede: If this doesn't work, I don't think I continue with this, I've already spent an inordinate amount of time debugging the Windows build 😅

sloede:
Yay, all builds pass and no copying of libraries 🎉

giordano:
```console
% readelf -d libt8.so                

Dynamic section at offset 0x76ae0 contains 33 entries:
 0x0000000000000001 (NEEDED)             Shared library: [libp4est-2.2.259-ec120.so]
 0x0000000000000001 (NEEDED)             Shared library: [libsc-2.8.1.5-0b70.so]
 0x0000000000000001 (NEEDED)             Shared library: [libmpicxx.so.12]
 0x0000000000000001 (NEEDED)             Shared library: [libmpi.so.12]
 0x000000000000000e (SONAME)             Library soname: [libt8-1.1.0.207-d6a74.so]
```
The soname being `libt8-1.1.0.207-d6a74.so` is highly problematic for compatibility reasons if we are ever going to build libraries linking to this, because this means you can link to exactly a single version of libt8 at a time, since each version will have a different full version number and hash.  ABI version should ***not*** be exactly the same as version number, nor unnecessarily restrict compatibility.  Of course, same for libp4est and libsc.  Probably not a problem right now, but something that you should probably deal with upstream.

sloede:
> The soname being `libt8-1.1.0.207-d6a74.so` is highly problematic for compatibility reasons if we are ever going to build libraries linking to this, because this means you can link to exactly a single version of libt8 at a time, since each version will have a different full version number and hash. ABI version should _**not**_ be exactly the same as version number, nor unnecessarily restrict compatibility. Of course, same for libp4est and libsc. Probably not a problem right now, but something that you should probably deal with upstream.

Yes, you are absolutely right. For now, we're just happy that we can use this on Windows at all, but I agree: this should be fixed, and I will create an issue upstream. Would you happen to know/have an idea as to why this occurs for this autotools-built library? I have never encountered such a behavior before...

And thanks for taking an interest/helping us out so much 💯🙏

PR #6397, opened 2023-03-16
 [Singular] update to 4.3.2-0

hannes14:


fingolfin (approved):
This looks good to me. Let's see how the test fare.

S/Singular/build_tarballs.jl
```diff
@@ -27,15 +27,15 @@ import Pkg.Types: VersionSpec
 # to all components.
 #
 name = "Singular"
-upstream_version = v"4.3.1-5" # 4.3.1 plus some changes
-version_offset = v"0.1.0"
+upstream_version = v"4.3.2-0" # 4.3.2p1 plus some changes
+version_offset = v"0.0.0"
```
fingolfin, line 31: Indeed! My bad, I suggested to @hannes14 that this should be OK (my excuse is that I slept 2 hours last night and my brain is even more mushy than usual)
```suggestion
version_offset = v"0.1.0"
```

PR #6406, opened 2023-03-17
[CUDA_full] Support aarch64 arch in v11.0, v11.1, v11.2

mofeing:
Installers for the aarch64 architecture are present in CUDA_full >= v11.3 but the installers for CUDA v11.0, v11.1, v11.2 were missing. This PR adds the respective installers to the recipes.


maleadt (approved):
LGTM

PR #6409, opened 2023-03-17
Update SwiftShader (Linux/MacOS only)

serenity4:
It seems like no new binaries have been distributed for a while, so we need to build manually. This should provide Vulkan 1.3 support (the current version of SwiftShader supports Vulkan 1.1 only, which is quite old and lacking a lot of functionality).

Linux and MacOS support was not too hard to get, but is more delicate on Windows.
If that is alright, I would like to exclude Windows from this update, as I personally don't have any use for it (and do not know of any projects that use it, see [this JuliaHub search](https://juliahub.com/ui/Search?q=SwiftShader_jll&type=code) where it seems only Vulkan.jl uses it, and it does not use it with Windows).


S/SwiftShader/build_tarballs.jl
```diff
@@ -2,37 +2,85 @@ using BinaryBuilder
...
+if [[ "${target}" == *-apple-* ]]; then
+    echo "hello"
+    folderName=Darwin
+elif [[ "${target}" == *-linux-* ]]; then
+    folderName=Linux
+elif [[ "${target}" == *-mingw* ]]; then
+    folderName=Windows
+fi
+
+mv build/$folderName/* ${libdir}
```
giordano, line 65: Please, always make these operations verbose so the log has a record of what has been moved where
```suggestion
mv -v build/$folderName/* ${libdir}
```

S/SwiftShader/build_tarballs.jl
```diff
@@ -2,37 +2,85 @@ using BinaryBuilder
...
+
+install_license LICENSE.txt
 """
 
 platforms = [
     Platform("x86_64", "linux"; libc="glibc"),
-    Platform("x86_64", "macos"),
-    Platform("x86_64", "windows"),
+    Platform("x86_64", "macos"; os_version = v"10.15"),
```
giordano, line 72: Why setting the OS version?  That's mostly unused anyway.

S/SwiftShader/build_tarballs.jl
```diff
@@ -2,37 +2,85 @@ using BinaryBuilder
...
-    LibraryProduct(["libvulkan", "vk_swiftshader"], :libvulkan),
+    LibraryProduct("libvulkan", :libvulkan),
+    LibraryProduct("libvk_swiftshader", :libvk_swiftshader),
 ]
 
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[]
 
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies, preferred_gcc_version=v"11")
```
giordano, line 86: Please don't use such a late version of GCC.  Use the ***oldest*** version that can actually compile the source code, and in general avoid going past 9 or 10, especially for C++ code, like this one.

S/SwiftShader/build_tarballs.jl
```diff
@@ -2,37 +2,85 @@ using BinaryBuilder
...
+elif [[ "${target}" == *-mingw* ]]; then
+    folderName=Windows
+fi
+
+mv build/$folderName/* ${libdir}
+
+install_license LICENSE.txt
 """
 
 platforms = [
```
giordano, line 70: Quoting [from the log](https://buildkite.com/julialang/yggdrasil/builds/1535#0186f0e8-d183-4446-b37e-dfc23d77c295/638-2251):
```
┌ Warning: /cache/build/yggy-amdci7-6/julialang/yggdrasil/S/SwiftShader/build/x86_64-linux-gnu/LZPx879y/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libSPIRV-Tools-shared.so contains std::string values! This causes incompatibilities across the GCC 4/5 version boundary. To remedy this, you must build a tarball for both GCC 4 and GCC 5. To do this, immediately after your `platforms` definition in your `build_tarballs.jl` file, add the line:
│
│  platforms = expand_cxxstring_abis(platforms)
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/ejTbj/src/auditor/compiler_abi.jl:247
```

S/SwiftShader/build_tarballs.jl
```diff
@@ -2,37 +2,85 @@ using BinaryBuilder
...
 ]
 
 products = [
-    LibraryProduct(["libvulkan", "vk_swiftshader"], :libvulkan),
+    LibraryProduct("libvulkan", :libvulkan),
+    LibraryProduct("libvk_swiftshader", :libvk_swiftshader),
 ]
 
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[]
```
giordano, line 84:
```suggestion
dependencies = [
    Dependency("CompilerSupportLibraries_jll"),
]
```

S/SwiftShader/build_tarballs.jl
serenity4: I started with GCC 4, then 5, then 10 and it did not work; GCC 11 seemed to be required. Fortunately, this library should be rarely added as a dependency and is more of a "final product".
I am not 100% certain that the errors encountered while experimenting with older GCC versions could not be worked around in some way (e.g. patching more things). I can try again if it really is a major pain to include such a late version of GCC.

S/SwiftShader/build_tarballs.jl
serenity4: Is it necessary with a GCC version > 5?

S/SwiftShader/build_tarballs.jl
serenity4: I ran into issues that led to having to include `export MACOSX_DEPLOYMENT_TARGET=10.12`, but did not try setting a lower OS version. Will see how it goes.

S/SwiftShader/build_tarballs.jl
giordano: Yes.

S/SwiftShader/build_tarballs.jl
giordano: Well, if you really need GCC 11 then be aware that this will likely not work at all with Julia v1.6.

Side note, even if not strictly required, it'd be good to specify the julia compat, like
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"11")
```
or whatever

serenity4: > Well, if you really need GCC 11 then be aware that this will likely not work at all with Julia v1.6.

I will go with GCC 11 for now, if it becomes a pain it can always be attempted to require a lower version as an improvement.

Given the suggestion, did you mean Julia < 1.6 or should the compat be e.g. >= 1.7?

serenity4:
Thanks :)

PR #5088, opened 2022-06-30
Update armadillo to latest version + support apple M1

quinnj:

PR #6410, opened 2023-03-18
[CeresSolver] Bump version to 2.1.0

junyuan-chen:
The purpose of this PR is to generate the latest version of CeresSolver for all platforms, including the Apple Silicon.

Updates:

@giordano Thank you for your comment! There are at least three issues remaining here:

- [x] Setting `preferred_gcc_version=v"5"` fixed the issues with platforms with C++03 ABI. With Eigen_jll.jl v3.4.0, errors are thrown on `powerpc64le`. Setting `preferred_gcc_version=v"8"` fixes this issue.
- [x] SuiteSparse needs to be turned on even though the build will succeed without it. With the latest SuiteSparse_jll.jl v7.0.1, CMake cannot recognoize its version number correctly and SuiteSparse is not used. By manually specifying the older v5.10.1, it works as expected. (SuiteSparse related flags from the old recipe were removed as they are not effective and contains a typo in file path.)
- [x] glog_jll.jl is a dependency that needs to be updated first: https://github.com/JuliaPackaging/Yggdrasil/pull/6412

### DIFF
--- a/C/CeresSolver/build_tarballs.jl
+++ b/C/CeresSolver/build_tarballs.jl
@@ -60,7 +60,7 @@ dependencies = [
     Dependency("glog_jll"),
     Dependency("METIS_jll"),
     Dependency("OpenBLAS_jll"),
-    Dependency("SuiteSparse_jll", build_version=v"5.10.1")
+    Dependency("SuiteSparse_jll", v"5.10.1")
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/C/CeresSolver/build_tarballs.jl
+++ b/C/CeresSolver/build_tarballs.jl
@@ -1,4 +1,4 @@
-using BinaryBuilder, Pkg
+using BinaryBuilder
 
 name = "CeresSolver"
 version = v"2.1.0"
@@ -39,7 +39,7 @@ products = Product[
 dependencies = [
     # CeresSolver is removing OpenMP and dependencies for OpenMP are dropped
     # https://github.com/ceres-solver/ceres-solver/issues/886
-    # Eigen_jll v0.3.4 throws error on powerpc64le with older GCC versions
+    # Eigen_jll v0.3.4 throws errors on powerpc64le with older GCC versions
     BuildDependency("Eigen_jll"),
     Dependency("glog_jll"),
     # Metis replaces SuiteSparse on Windows
--- a/C/CeresSolver/build_tarballs.jl
+++ b/C/CeresSolver/build_tarballs.jl
@@ -36,15 +36,11 @@ products = Product[
     LibraryProduct("libceres", :libceres),
 ]
 
-ispowerpc64le(x) = arch(x) == "powerpc64le"
-
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    # CeresSolver prefers CXX_THREADS over OpenMP and dependencies for OpenMP are removed
-    BuildDependency("Eigen_jll"; platforms=filter(!ispowerpc64le, platforms)),
-    # Eigen v3.4.0 does not work on powerpc64le
-    BuildDependency(PackageSpec(name="Eigen_jll", version=v"3.3.9");
-        platforms=filter(ispowerpc64le, platforms)),
+    # CeresSolver is removing OpenMP and dependencies for OpenMP are dropped
+    # https://github.com/ceres-solver/ceres-solver/issues/886
+    BuildDependency("Eigen_jll"),
     Dependency("glog_jll"),
     # Metis replaces SuiteSparse on Windows
     Dependency("METIS_jll"),
@@ -54,4 +50,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"5", julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"6", julia_compat="1.6")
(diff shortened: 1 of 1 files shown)


C/CeresSolver/build_tarballs.jl
```diff
@@ -26,33 +21,27 @@ else
 fi
 
 CMAKE_FLAGS+=(-DCMAKE_INSTALL_PREFIX=${prefix}
+              -DCMAKE_CXX_FLAGS="-std=c++17"
```
giordano, line 24: You should rather use [`CMAKE_CXX_STANDARD`](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html#variable:CMAKE_CXX_STANDARD).  Also, if this is really needed, this option should be set by upstream CMake file, it doesn't make sense that users need to set this option manually.

C/CeresSolver/build_tarballs.jl
```diff
@@ -1,20 +1,15 @@
...
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/ceres-solver-1.14.0
-
-mkdir cmake_build
-cd cmake_build/
-
 if [[ "$target" == *-freebsd* || "$target" == *-apple-* ]]; then
   # Clang doesn't play nicely with OpenMP and
   # compilation fails with glog due to a c++11 error
```
giordano, line 15: https://github.com/JuliaPackaging/Yggdrasil/blob/48d3d50451f098e243808bb4e3f704947808ff20/B/blis/build_tarballs.jl#L135-L138

C/CeresSolver/build_tarballs.jl
```diff
@@ -26,33 +21,27 @@ else
...
+cmake .. ${CMAKE_FLAGS[@]} -DCMAKE_BUILD_TYPE=Release
 make -j${nproc}
 make install
 
 cd ..
 """
 
 platforms = expand_cxxstring_abis(supported_platforms())
+# Build for all platforms with cxx03 ABI fails
+filter!(p->cxxstring_abi(p) === :cxx11, platforms)
```
giordano, line 44: Uh?  This sounds suspiciously wrong.

giordano:
It predates this PR, but https://github.com/JuliaPackaging/Yggdrasil/blob/48d3d50451f098e243808bb4e3f704947808ff20/C/CeresSolver/build_tarballs.jl#L64 is also wrong: `Eigen_jll` is headers-only, it should be a [`BuildDependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.BuildDependency)

junyuan-chen:
@giordano Thank you for your comment! There are at least three issues remaining here:

- [ ] Should I simply skip the platforms with C++03 ABI? They all fail.
- [ ] The package requires SuiteSparse for certain functionality. I see that for all the build made so far, the support for SuiteSparse is disabled as not all SuiteSparse dependencies are found. I don't know how to fix that. This won't break the building process but we certainly need SuiteSparse. (I removed the SuiteSparse related flags from the old recipe as there was an error saying the directory doesn't exist.)
- [ ] `glog_jll.jl` is a dependency that needs to be updated. The current available version is old and there is no build for Apple Silicon.

giordano:
> Should I simply skip the platforms with C++03 ABI?

I don't think so.  What's the problem?

junyuan-chen:
> I don't think so. What's the problem?

For these platforms, I keep getting a bunch of errors like below:

```
[19:45:33] /workspace/srcdir/ceres-solver/internal/ceres/partitioned_matrix_view_impl.h:269:10: error: ‘make_unique’ is not a member of ‘std’
--
  | [19:45:33]    return std::make_unique<BlockSparseMatrix>(block_diagonal_structure);
```

giordano:
Try requiring GCC 5

junyuan-chen:
@giordano Things look good to me now. This PR is ready for review.

C/CeresSolver/build_tarballs.jl
```diff
@@ -59,15 +36,22 @@ products = Product[
     LibraryProduct("libceres", :libceres),
 ]
 
+ispowerpc64le(x) = arch(x) == "powerpc64le"
+
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Eigen_jll"),
+    # CeresSolver prefers CXX_THREADS over OpenMP and dependencies for OpenMP are removed
+    BuildDependency("Eigen_jll"; platforms=filter(!ispowerpc64le, platforms)),
```
giordano, line 44: I think OpenMP on PowerPC requires GCC 6: https://github.com/JuliaPackaging/Yggdrasil/pull/5088#issuecomment-1174406071

junyuan-chen: I got the same `error: invalid parameter combination for AltiVec intrinsic` on `powerpc64le` with GCC 6 but no more errors after switching to GCC 8.

C/CeresSolver/build_tarballs.jl
```diff
@@ -1,55 +1,32 @@
...
-              -DCOLAMD_LIBRARY="${libdir}/libcolamd.${dlext} ${libdir}/libsuitepsarseconfig.${dlext}"
-              -DSUITESPARSEQR_LIBRARY="${libdir}/libspqr.${dlext} ${libdir}/libsuitesparseconfig.${dlext}"
-              -DSUITESPARSE_CONFIG_LIBRARY="${libdir}/libsuitesparseconfig.${dlext}"
               )
 
-cmake ${CMAKE_FLAGS[@]} ..
-
+cd $WORKSPACE/srcdir/ceres-solver/
+mkdir build && cd build
+export OPENBLAS_NUM_THREADS=1
```
giordano, line 26: Why do you need to set this during the build? Looks weird.

junyuan-chen: I saw that on the BLAS section here: http://ceres-solver.org/installation.html

giordano: I doubt that they're talking about the build. That's probably a comment about runtime.

junyuan-chen: That makes sense. I just removed that line.

PR #6411, opened 2023-03-18
[MPITrampoline] Do not install MPICH man files

giordano:
They cause clashes on case insensitive file systems. Same as #6345. CC @eschnett

### DIFF
--- a/M/MPItrampoline/build_tarballs.jl
+++ b/M/MPItrampoline/build_tarballs.jl
@@ -139,11 +139,8 @@ if [[ "${target}" == aarch64-apple-* ]]; then
 fi
 
 # Do not install doc and man files which contain files which clashing names on
-
 # case-insensitive file systems:
-
 # * https://github.com/JuliaPackaging/Yggdrasil/pull/315
-
 # * https://github.com/JuliaPackaging/Yggdrasil/issues/6344
 ./configure \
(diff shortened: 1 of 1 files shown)


M/MPItrampoline/build_tarballs.jl
```diff
@@ -138,11 +138,19 @@ if [[ "${target}" == aarch64-apple-* ]]; then
     )
 fi
 
+# Do not install doc and man files which contain files which clashing names on
+
```
giordano, line 142:
```suggestion
```

M/MPItrampoline/build_tarballs.jl
```diff
@@ -138,11 +138,19 @@ if [[ "${target}" == aarch64-apple-* ]]; then
...
+# case-insensitive file systems:
+
```
giordano, line 144:
```suggestion
```

M/MPItrampoline/build_tarballs.jl
```diff
@@ -138,11 +138,19 @@ if [[ "${target}" == aarch64-apple-* ]]; then
...
+# * https://github.com/JuliaPackaging/Yggdrasil/pull/315
+
```
giordano, line 146:
```suggestion
```

PR #6413, opened 2023-03-18
[Eigen] Bump version to v3.4.0

junyuan-chen:


E/Eigen/build_tarballs.jl
```diff
@@ -1,22 +1,21 @@
...
-cd $WORKSPACE/srcdir/eigen-*
-
-mkdir build
-cd build/
+cd $WORKSPACE/srcdir/eigen
+mkdir build && cd build
 
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_HOST_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
```
giordano, line 18: This probably doesn't matter much since these are only header files 😉

junyuan-chen: @giordano Thanks!

PR #6414, opened 2023-03-19
[Diamond] Add new package Diamond v2.1.6

marcom:


D/Diamond/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+mkdir build && cd build
+cmake .. \
+    -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=RELEASE
+make -j${nproc}
+make install
+install_license ../LICENSE
+"""
+
+platforms = supported_platforms(exclude = p -> Sys.iswindows(p) || nbits(p) == 32)
+platforms = expand_cxxstring_abis(platforms; skip = p -> Sys.isfreebsd(p) || (Sys.isapple(p) && arch(p) == "aarch64"))
```
giordano, line 34: Sorry?  The [cxx03](https://buildkite.com/julialang/yggdrasil/builds/1583#0186f998-0c02-4309-83f8-31bae9433c5d/637-1952) and [cxx11](https://buildkite.com/julialang/yggdrasil/builds/1583#0186f998-0c05-4212-b8c3-8cbf577075b9/626-1941) builds of x86_64 darwin have ***identical*** git tree hash, and the log doesn't say that the built executables are locked to a specific C++ string ABI, which would be highly suspicious since we should be building with Clang, and if that wasn't the case something would be very wrong.

D/Diamond/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
+using BinaryBuilder, Pkg
+
+name = "Diamond"
+version = v"2.1.6"
+
+# url = "https://github.com/bbuchfink/diamond"
+# description = "Accelerated BLAST-compatible local sequence aligner"
+
+# TODO
+# - WITH_ZSTD (fails to find the library)
```
giordano, line 10: Uh?  https://buildkite.com/julialang/yggdrasil/builds/1583#0186f998-0be8-4174-96f7-f84ab2f847b5/637-796

marcom: The log refers to ZLIB. With `Zstd_jll` as a dependency and `cmake -DWITH_ZSTD=1`, cmake can't find the zstd library.

D/Diamond/build_tarballs.jl
marcom: I'll have to admit that I just copied the `expand_cxxstring_abis` line from the last `build_tarballs.jl` script I had, is this `expand_cxxstring_abis` line wrong in general? Any way to easily check this from my side?

giordano: You need to expand if and only if the audit tells you so.  Otherwise, don't.

D/Diamond/build_tarballs.jl
giordano: > The log refers to zlib. With Zstd_jll as a dependency and cmake -DWITH_ZSTD=1, cmake can't find the zstd library.

Sorry, right, I confused them

> Would be thankful for an idea how to fix this.

I'd need to know what the problem is, to make any guess.

D/Diamond/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+mkdir build && cd build
+cmake .. \
+    -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=RELEASE
+make -j${nproc}
+make install
+install_license ../LICENSE
+"""
+
+platforms = supported_platforms(exclude = p -> Sys.iswindows(p) || nbits(p) == 32)
+platforms = expand_cxxstring_abis(platforms; skip = p -> Sys.isfreebsd(p) || Sys.isapple(p))
```
giordano, line 34:
```suggestion
platforms = expand_cxxstring_abis(platforms)
```

D/Diamond/build_tarballs.jl
giordano: For the record, the problem was that diamond for some reasons expects zstd to be only a static library: https://github.com/bbuchfink/diamond/blob/0540f86d3d5965cf9a6ef8871068999ef61211dc/CMakeLists.txt#L470 :shrug:

marcom: Many thanks for the help! That zstd is expected to be a static library seems to be quite common with cmake-based projects, I think some other packages that I made for Yggdrasil also might be able to use this fix.

PR #6417, opened 2023-03-20
RuntimeDependency doesn't work with top-level

vchuravy:


giordano:
What's the problem exactly?  I think we should fix whatever is the bug with `RuntimeDependency` + `top_level`, instead of using `Dependency` + `top_level`, [which is deprecated](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/f73c892c66be7e1b3ec0f6b56089c065498ec401/src/Dependencies.jl#L131).

PR #6424, opened 2023-03-20
so version cuda libraries in PAPI

vchuravy:
cc: @maleadt

I skipped `libcuda.so` since I am unsure if the SO version will actually match there.


P/PAPI/build_tarballs.jl
```diff
@@ -34,8 +34,14 @@ fi
 
 COMPONENTS=()
 if [[ -d "${prefix}/cuda" ]]; then
-    export PAPI_CUDA_ROOT="${prefix}/cuda"
     COMPONENTS+=(cuda nvml)
+
```
vchuravy, line 38:
```suggestion
    export PAPI_CUDA_ROOT="${prefix}/cuda"
```

PR #6436, opened 2023-03-22
[llama_cpp] Add build script for llama_cpp

marcom:
From the same developer as whisper.cpp (`whisper_cpp` in Yggdrasil)

https://github.com/ggerganov/llama.cpp

A very basic Julia interface is here:

https://github.com/marcom/Llama_cpp.jl


L/llama_cpp/build_tarballs.jl
```diff
@@ -0,0 +1,69 @@
...
+products = [
+    ExecutableProduct("main", :main),
+    ExecutableProduct("quantize", :quantize),
+]
+
+dependencies = Dependency[
+]
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="1.6", preferred_gcc_version = v"11")
```
giordano, line 69: Use the oldest GCC version possible, GCC 11 is quite new, this won't work with Julia 1.6 in practice.

PR #6440, opened 2023-03-22
bump version

dev10110:


dev10110:
@giordano @vchuravy when you get a chance, could you take a look at this. 

For now I've disabled the powerpc64 - I've tried a few things, but I need this library for some following work, so if we could approve this first, and then work on getting powerpc64 to work later that would be good.

D/DecompUtil/build_tarballs.jl
```diff
@@ -21,7 +21,7 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = [p for p in supported_platforms() if triplet(p) != "powerpc64le-linux-gnu"]
```
giordano, line 24: This is a nicer way to exclude the PowerPC architecture:
```suggestion
platforms = supported_platforms(; exclude=p->arch(p)=="powerpc64le")
```

dev10110: cool! updated

PR #6441, opened 2023-03-22
[Ipopt] update to v3.14.11

odow:
Seems like it is now safe to use LBT on Windows: https://github.com/JuliaLinearAlgebra/libblastrampoline/issues/89


giordano:
> Seems like it is now safe to use LBT on Windows:

***...only with LBT 5.4+, which requires Julia v1.9***

odow:
> only with LBT 5.4+, which requires Julia v1.9

Oh :cry:

PR #6446, opened 2023-03-24
[SuiteSparse] build with PARTITION (Metis)

ViralBShah:
It didn't work the last time. So removing the option altogether, putting the build on a default path that should include the internal included Metis.


ViralBShah:
Is there a way for me to download the built artifacts and examine them?

giordano:
See the "artifacts" tab for the job of the platform you want

PR #6450, opened 2023-03-24
[llama_cpp] bump version and install library and header files

marcom:


L/llama_cpp/build_tarballs.jl
```diff
@@ -40,16 +40,37 @@ cmake .. \
...
+
+# executables
 for prg in main quantize; do
     install -Dvm 755 "./bin/${prg}${exeext}" "${bindir}/${prg}${exeext}"
 done
 
+# libs
+for lib in libllama; do
+    if [[ "${target}" == *-w64-mingw32* ]]; then
+        install -Dvm 644 "./bin/${lib}.${dlext}" "${libdir}/${lib}.${dlext}"
```
giordano, line 62: Libraries should be installed with 755 permissions, otherwise you can't even dlopen them (we should have an audit pass to correct this, but doing the right thing in the first place is better)

L/llama_cpp/bundled/patches/windows-examples-fix-missing-ggml-link.patch
giordano, line 1: Should this patch be upstreamed?

marcom: I submitted a patch here: https://github.com/ggerganov/llama.cpp/pull/542

PR #6454, opened 2023-03-25
[PRIMME] add builder

RalphAS:
C language library for eigenvalues of very large Hermitian matrices and singular values of large general matrices.

### DIFF
--- a/P/PRIMME/build_tarballs.jl
+++ b/P/PRIMME/build_tarballs.jl
@@ -1,0 +1,56 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "PRIMME"
+version = v"3.2.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/primme/primme.git",
+              "8d0ca1812436665564e5fba82d2a485e96fd7627"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+if [[ ${nbits} == 64 ]]; then
+CFLAGS="-O -fPIC -DNDEBUG -DPRIMME_BLASINT_SIZE=64 -DPRIMME_BLAS_SUFFIX=_64"
+else
+CFLAGS="-O -fPIC -DNDEBUG -DPRIMME_BLASINT_SIZE=32"
+fi
+
+if [[ "${target}" == *mingw* && ${nbits} == 32 ]]; then
+  LDFLAGS="-L${libdir} -lopenblas"
+elif [[ "${target}" == *mingw* && ${nbits} == 64 ]]; then
+  LDFLAGS="-L${libdir} -lopenblas64_"
+else
+  LDFLAGS="-L${libdir} -lblastrampoline"
+fi
+
+cd primme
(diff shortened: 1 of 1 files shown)


P/PRIMME/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+else
+  LDFLAGS="-L${libdir} -lblastrampoline"
+fi
+
+cd primme
+if [[ "${target}" == *musl* ]]; then
+  sed -i 's/undef I/define I_is_a_stupid_sytem_macro 1/' ./src/include/common.h
+fi
+if [[ "${target}" == *mingw* ]]; then
+  sed -i -e 's/defined (__unix__)/1/' -e '/resource.h/d' ./src/linalg/wtime.c
```
giordano, line 35: This looks...uhm...wrong

giordano: If this was meant to "fix"
```
[20:53:49] linalg/wtime.c:81:21: fatal error: Windows.h: No such file or directory
[20:53:49]  #include <Windows.h>
[20:53:49]                      ^
```
then the right way to do it is to patch the source code and use `windows.h` (lowercase) and submit the patch upstream.  Not to pretend to be a Unix system when targeting Windows which is blatantly false.

RalphAS: The context here is that the file in question only uses the test to determine whether `gettimeofday()` and `struct timeval` are available; they are provided by MinGW, hence the white lie. My proposed fix should make behavior identical on all the JLL targets, if that matters.
I was confused by the fact that Microsoft says the alternative `GetTickCount()` is declared in `sysinfoapi.h` which is **not** provided by MinGW (I have since learned that the MinGW declaration is in `winbase.h` **instead**). 

MS also frequently uses the capitalized header file name in documentation, which **might** be just a trap for cross compilers but I'm not ready to claim to upstream that I know the "right way" for the rest of the Windows world.

Sorry if I'm ranting. I'm not yet inured to the trials of building binaries for systems I don't actually use.  
Anyway, if you prefer just patching the header name that's also ok with me.

giordano: > MS also frequently uses the capitalized header file name in documentation

Windows documentation is very inconsistent about case of header files and libraries, because Windows typically uses case insensitive file systems, with the result that developers using Windows only generally don't care. But cases matter when compiling on case sensitive file system, and mingw consistently uses lowercase names everywhere. We have upstreamed very often patches for wrongly capitalised Windows.h

P/PRIMME/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+  LDFLAGS="-L${libdir} -lopenblas"
+elif [[ "${target}" == *mingw* && ${nbits} == 64 ]]; then
+  LDFLAGS="-L${libdir} -lopenblas64_"
+else
+  LDFLAGS="-L${libdir} -lblastrampoline"
+fi
+
+cd primme
+if [[ "${target}" == *mingw* ]]; then
+  sed -i 's/Windows\.h/windows.h/' ./src/linalg/wtime.c
```
giordano, line 32: I still recommend submitting a PR upstream to fix this once and for all.

RalphAS: I inadvertently hid the comment where I mentioned that.  They've already processed it: https://github.com/primme/primme/pull/67

PR #6459, opened 2023-03-27
Recompile MUMPS with LBT for all platforms

amontoison:


amontoison:
@giordano Do you have an idea why the builds failed after that you merged the PR?
Capture d’écran du 2023-03-27 14-44-25

PR #6461, opened 2023-03-27
Recompile CUTEst

amontoison:


C/CUTEst/build_tarballs.jl
```diff
@@ -67,13 +56,13 @@ if [[ "${target}" == *-linux* || "${target}" == *-freebsd* ]]; then
   echo "2" >> cutest.opts  # Linux
   echo "6" >> cutest.opts  # gfortran
   echo "2" >> cutest.opts  # build all tools except Matlab
-  echo "9" >> cutest.opts  # gcc
+  echo "8" >> cutest.opts  # gcc
```
giordano, line 59: Honestly, having to write down some random numbers to indicate a toolchain is quite awful. Where is this stuff even documented?

amontoison: I know that it's horrible. It's an old Fortran package. I'm working with the developers to modernize it and use Meson.

PR #6462, opened 2023-03-27
[libblastrampoline] Update to 5.5.0

ViralBShah:


L/libblastrampoline/build_tarballs.jl
```diff
@@ -41,5 +41,5 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat="1.9",
+               julia_compat="1.9", preferred_gcc_version = v"6"
```
giordano, line 44:
```suggestion
               julia_compat="1.10"
```

ViralBShah: I was thinking that we may want to backport to 1.9.

ViralBShah: Perhaps that is unlikely?

PR #6474, opened 2023-03-29
Update SPRAL and compile it with LBT for all platforms

amontoison:


S/SPRAL/build_tarballs.jl
```diff
@@ -3,37 +3,44 @@
...
-    GitSource("https://github.com/ralna/spral.git", "cd2d2e817275f16d586cf72767631b3c7472ce02")
+    GitSource("https://github.com/ralna/spral.git", "b1d6a7409d41d575bff7106a4e9d9a60b8935c2e")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd ${WORKSPACE}/srcdir/spral
+
+if [[ "${target}" == *mingw* ]]; then
+  BLAS_LAPACK="-L${libdir} -lblastrampoline-5"
```
giordano, line 18: Aren't you duplicating `-L${libdir}` everywhere?

S/SPRAL/build_tarballs.jl
```diff
@@ -3,37 +3,44 @@
...
     CC=gcc
     CXX=g++
 fi
+
 ./autogen.sh
 mkdir build
 cd build
-CFLAGS=-fPIC CPPFLAGS=-fPIC CXXFLAGS=-fPIC FFLAGS=-fPIC FCFLAGS=-fPIC \
+CFLAGS="-O3 -fPIC" CPPFLAGS="-O3 -fPIC" CXXFLAGS="-O3 -fPIC" FFLAGS="-O3 -fPIC" FCFLAGS="-O3 -fPIC" \
     ../configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} \
```
giordano, line 32: I find
```suggestion
export CFLAGS="-O3 -fPIC"
export CXXFLAGS="-O3 -fPIC"
export FFLAGS="-O3 -fPIC"
export FCLAGS="-O3 -fPIC"
../configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} \
```
easier to read.  Also, `-O3 -fPIC` are pointless flags to pass to the preprocessor.

S/SPRAL/build_tarballs.jl
amontoison: Indeed, thanks Mosè!

S/SPRAL/build_tarballs.jl
giordano: > Also, `-O3 -fPIC` are pointless flags to pass to the preprocessor.

Well, given the errors my gut feeling is that they're either using `CPPFLAGS` where they shouldn't, or they aren't using `CFLAGS`/`CXXFLAGS` where they should :roll_eyes:

amontoison: You wrote `FCLAGS` instead of `FCFLAGS` in your suggestion. :wink:

giordano: Well, that'd also do it!

S/SPRAL/build_tarballs.jl
```diff
@@ -3,37 +3,47 @@
...
+if [[ "${target}" == *mingw* ]]; then
+  LBT="-lblastrampoline-5"
+else
+  LBT="-lblastrampoline"
+fi
+
 if [[ "${target}" == *-freebsd* ]] || [[ "${target}" == *-apple-* ]]; then
     CC=gcc
     CXX=g++
 fi
```
giordano, line 26: Now that I see this: why?

amontoison: Because the tests of SPRAL are failing if it's compiled with `clang`.
I don't know if it's also the case with FreeBSD.
https://github.com/ralna/spral/pull/97

giordano: That....looks messy

S/SPRAL/build_tarballs.jl
```diff
@@ -42,11 +52,11 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="METIS_jll", uuid="d00139f3-1899-568f-a2f0-47f597d42d70"))
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
-    Dependency(PackageSpec(name="OpenBLAS32_jll", uuid="656ef2d0-ae68-5445-9ca0-591084a874a2"))
```
giordano, line 47: Ok, last question: the library uses LP64 interface, you must make sure that your consumer package will have at least one backing LP64 BLAS, otherwise you'll get runtime errors.  Would you do that?

amontoison: Yes, I recompiled SPRAL to use it as a linear solver of Ipopt, like MUMPS. And in Ipopt.jl, we are already taking care of that: https://github.com/jump-dev/Ipopt.jl/blob/master/src/Ipopt.jl#L15-L28

PR #6478, opened 2023-03-30
New Recipe: FireFly v2.0.3

Fenyutanchan:


Fenyutanchan:
Hello, @giordano.
Could you help me solve these strange problems in both x86-64 and ARM MacOS?
I don't understand it well.

giordano:
At a quick glance, I think the linker invocation is missing to link the needed libraries, hence the undefined symbols errors.  _Why_ that's happening, I don't know without having a more thorough look, which now I can't.

F/FireFly/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+mkdir $WORKSPACE/srcdir/FireFly-build
+cd $WORKSPACE/srcdir/FireFly-build/
+
+if [ ${target} == "*darwin*" ]; then
+    extra_cmake_flags="-DCMAKE_OSX_DEPLOYMENT_TARGET=11.0"
+fi
```
giordano, line 20: This is wrong on x86_64 and redundant on aarch64.

Fenyutanchan:
I have disabled the CMake flags `-DCUSTOM` for the darwin building.
However, all other platforms got stuck with the target of "example".
I switched the GCC version to v6.1.0 for solving it.
There is then a strange segmentation fault raised during making the "x86-64-linux-gnu-cxx11", while there is no problem in my local machine.
I have no idea what happened.
@giordano, do you have any comments on it?

F/FireFly/build_tarballs.jl
```diff
@@ -0,0 +1,59 @@
...
+products = [
+    LibraryProduct("libfirefly", :libfirefly)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="FLINT_jll", uuid="e134572f-a0d5-539d-bddf-3cad8db41a82"))
+    Dependency(PackageSpec(name="GMP_jll", uuid="781609d7-10c4-51f6-84f2-b8444358ff6d"); compat="6.2.0")
+    Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a"))
+    Dependency(PackageSpec(name="MPICH_jll", uuid="7cb0a576-ebde-5e09-9194-50597f1243b4"))
```
giordano, line 49: You should really follow what the other MPI packages do.  See https://github.com/search?q=repo%3AJuliaPackaging%2FYggdrasil%20mpi.jl&type=code

Fenyutanchan: Done it.

PR #6480, opened 2023-03-30
New Recipe: FastCDR v1.0.0

twadleigh:
Recipe to build https://github.com/eProsima/Fast-CDR C++ library along with a parallel C interface (https://github.com/twadleigh/cfastcdr)


F/FastCDR/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "FastCDR"
+version = v"1.0.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/eProsima/Fast-CDR.git", "5d782877435b569e0ed38541f362e212c9123dd4"),
+    GitSource("https://github.com/twadleigh/cfastcdr.git", "5baf8c6eb3c5d749caa070fa478b207b39896f47")
```
giordano, line 11: Shouldn't these be two separate packages?

PR #6481, opened 2023-03-30
Add CyrusSASL dependency to librdkafka

andyferris:
This change enables SASL-based authentication when connecting to Kafka. Currently RDKafka.jl can segfaul with certain connection configurations, and this should provide the necessary functionality to support e.g. SASL-SCRAM authentication.

I have verified locally that the configure script finds libsasl2 and that it is linked.

### DIFF
--- a/L/librdkafka/build_tarballs.jl
+++ b/L/librdkafka/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"2.0.2"
 sources = [
     # git rev-list -n 1 v2.0.2
     GitSource("https://github.com/confluentinc/librdkafka.git",
-              "292d2a66b9921b783f08147807992e603c7af059",
+        "292d2a66b9921b783f08147807992e603c7af059",
     ),
 ]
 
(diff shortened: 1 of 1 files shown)


giordano:
Is there a newer version to upgrade to? Changing dependencies within the same patch version is a bit non-kosher

andyferris:
Not yet, it seems 2.1.0 is a work-in-progress.

I assumed we could release 2.0.2+1?

andyferris:
If that's not possible, what are the alternatives?

giordano:
We automatically bump the build version number, but that's the problem: the patch number is the same, and Pkg doesn't really like changing things within the same patch version. If the next version is slated to be 2.1.0, we can lie a little bit and use a fictional 2.0.3 version.

andyferris:
OK, thanks, I bumped the patch number (with a comment).

L/librdkafka/build_tarballs.jl
```diff
@@ -45,12 +45,13 @@ products = [
...
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="Lz4_jll", uuid="5ced341a-0733-55b8-9ab6-a4889d929147"))
-    Dependency(PackageSpec(name="Zstd_jll", uuid="3161d3a3-bdf6-5164-811a-617609db77b4"))
-    Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a"))
-    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"))
+    Dependency(PackageSpec(name="Lz4_jll", uuid="5ced341a-0733-55b8-9ab6-a4889d929147")),
+    Dependency(PackageSpec(name="Zstd_jll", uuid="3161d3a3-bdf6-5164-811a-617609db77b4")),
+    Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a")),
+    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95")),
```
giordano, line 51: Should specify the compat for openssl 
```suggestion
    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="1.1.10"),
```

andyferris: Ah, thanks.

(I don't know what we are going to do in September when OpenSSL 1.1 support ends...)

giordano: https://github.com/JuliaPackaging/Yggdrasil/pull/6289

PR #6483, opened 2023-03-31
[llama_cpp] Bump version 

marcom:
This version enables a new mmap-friendly model weights file
format. Unfortunately w64-mingw32 fails to build after these changes,
so disable it for now (no Windows.h or sys/mman.h).

### DIFF
--- a/L/llama_cpp/build_tarballs.jl
+++ b/L/llama_cpp/build_tarballs.jl
@@ -19,11 +19,11 @@ version = v"0.0.6"  # fake version number
 # 0.0.3           22.03.2023       master-d5850c5    https://github.com/ggerganov/llama.cpp/releases/tag/master-d5850c5
 # 0.0.4           25.03.2023       master-1972616    https://github.com/ggerganov/llama.cpp/releases/tag/master-1972616
 # 0.0.5           30.03.2023       master-3bcc129    https://github.com/ggerganov/llama.cpp/releases/tag/master-3bcc129
-# 0.0.6           31.03.2023       master-02c5b27    https://github.com/ggerganov/llama.cpp/releases/tag/master-02c5b27
+# 0.0.6           31.03.2023       master-3525899    https://github.com/ggerganov/llama.cpp/releases/tag/master-3525899
 
 sources = [
     GitSource("https://github.com/ggerganov/llama.cpp.git",
-              "02c5b27e91a6d18cf1043d3a2d8dbc59610ac257"),
+              "3525899277d2e2bdc8ec3f0e6e40c47251608700"),
     DirectorySource("./bundled"),
 ]
 
--- a/L/llama_cpp/build_tarballs.jl
+++ b/L/llama_cpp/build_tarballs.jl
@@ -32,8 +32,6 @@ cd $WORKSPACE/srcdir/llama.cpp*
 
 # remove -march=native from cmake files
 atomic_patch -p1 ../patches/cmake-remove-mcpu-native.patch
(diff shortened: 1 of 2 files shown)


marcom:
Thanks a lot for the tip with lower-case windows.h! That seems to have gotten things building again on w64-mingw32.

I will add a patch upstream soon

L/llama_cpp/bundled/patches/fix-mingw32-windows-include.patch
```diff
@@ -0,0 +1,19 @@
...
+ 
+ #if defined(_WIN32) && !defined(_POSIX_MAPPED_FILES)
++#if !defined(__MINGW32__)
+ #define WIN32_LEAN_AND_MEAN
+ #include <Windows.h>
+ #else
++// ref: https://github.com/ggerganov/whisper.cpp/issues/168
++#include <windows.h>
++#endif
++#else
```
giordano, line 16: No, this should ***always*** be `windows.h`.

giordano: The difference isn't MinGW vs else, but case-sensitive vs case-insensitive file systems.

marcom: You mean also when compiling with on non-mingw32 (gcc) compilers on windows, i.e. visual studio c++? I don't really have a means of testing this setup myself (honestly I don't use windows, and it always seems to cause the most work :smile: ).

I copied the logic from https://github.com/ggerganov/llama.cpp/blob/53dbba769537e894ead5c6913ab2fd3a4658b738/ggml.c#L31-L36 and assumed there was a reason for the lower-case / upper-case includes.

giordano: > You mean also when compiling with non-mingw32 (gcc) compilers on windows, i.e. visual studio c++?

Yes, because as I said the problem isn't the toolchain used but the file system used.

giordano: > I copied the logic from [ggerganov/llama.cpp@53dbba7/ggml.c#L31-L36](https://github.com/ggerganov/llama.cpp/blob/53dbba769537e894ead5c6913ab2fd3a4658b738/ggml.c#L31-L36) and assumed there was a reason for the lower-case / upper-case includes.

That's a.... stupid logic......

marcom: Thanks again for the explanation! Sorry I didn't see your comment about the windows case-insensitive filesystem until after I had sent my comment off.

`windows.h` it is.

L/llama_cpp/build_tarballs.jl
```diff
@@ -1,14 +1,15 @@
...
+version = v"0.0.6"  # fake version number
 
 # url = "https://github.com/ggerganov/llama.cpp"
 # description = "Port of Facebook's LLaMA model in C/C++"
 
 # TODO
-# - i686, x86_64, aarch64 build
-#   missing architectures: powerpc64le, armv6l, arm7vl
+# - w64-mingw32 fails since master-02c5b27 (fake v0.0.6) because of
+#   missing mmap support
```
giordano, line 11: This comment is now inaccurate, isn't it?

marcom:
Upstream pull request here: https://github.com/ggerganov/llama.cpp/pull/747

PR #6475, opened 2023-03-29
Wizard recipe: GiNaC-v1.8.6

Fenyutanchan:
This pull request contains a new build recipe I built using the BinaryBuilder.jl wizard:

* Package name: GiNaC
* Version: v1.8.6

PR #6485, opened 2023-03-31
New Recipe: Kira v2.3.0

Fenyutanchan:
Kira needs the external program `Fermat` from [page](https://home.bway.net/lewis/).
Though the source code of the old version of `Fermat` was open-sourced under the GNU GPL as [page](https://home.bway.net/lewis/) says, I have no copy for it.
Therefore I cannot build `Fermat` for most platforms.

However, the building of `Kira` is irrelevant to the Fermat binary, so I try this recipe here for the platforms supported by `Fermat`.
This is just a try.
It would be great if `Yggdrasil` accepts it or not.

### DIFF
--- a/K/Kira/build_tarballs.jl
+++ b/K/Kira/build_tarballs.jl
@@ -29,6 +29,7 @@ pip3 install -U meson
 
 mkdir ${WORKSPACE}/srcdir/Kira-build
 cd ${WORKSPACE}/srcdir/Kira-build/
+
 meson setup \
     -Dfirefly=true \
     -Dflint=true \
@@ -39,7 +40,9 @@ meson setup \
     ${WORKSPACE}/srcdir/kira/ \
     ${WORKSPACE}/srcdir/Kira-build/
 sed -i "s/\/workspace\/destdir\/opt/\/opt/g" build.ninja
+
 meson install
+install_license ${WORKSPACE}/srcdir/kira/COPYING
 """
 
 # These are the platforms we will build for by default, unless further
--- a/K/Kira/build_tarballs.jl
+++ b/K/Kira/build_tarballs.jl
@@ -1,0 +1,196 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Kira"
+version = v"2.3.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://gitlab.com/kira-pyred/kira.git", "8001d64f23406d6c4cb06578b89011c1a0c04c65"),
+    ArchiveSource("https://www.ginac.de/ginac-1.8.6.tar.bz2", "00b320b1116cae5b7b43364dbffb7912471d171f484d82764605d715858d975b"),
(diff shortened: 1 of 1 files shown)


K/Kira/build_tarballs.jl
```diff
@@ -0,0 +1,78 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Kira"
+version = v"2.3.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://gitlab.com/kira-pyred/kira.git", "8001d64f23406d6c4cb06578b89011c1a0c04c65"),
```
giordano, line 10: Having multiple sources is usually a good indication that you should have multiple packages.

K/Kira/build_tarballs.jl
```diff
@@ -0,0 +1,78 @@
...
+make -j${nproc}
+make install
+
+mkdir $WORKSPACE/srcdir/FireFly-build
+cd $WORKSPACE/srcdir/FireFly-build/
+cmake -DWITH_FLINT=true -DWITH_JEMALLOC=true -DWITH_MPI=true -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release ${WORKSPACE}/srcdir/firefly
+cmake --build . -j${nproc}
+cmake --build . -t install
+
+pip3 install -U meson
```
giordano, line 28: Uh? BinaryBuilder environment comes with meson already.

K/Kira/build_tarballs.jl
Fenyutanchan: IC, all other packages are prepared and waited for merging.

K/Kira/build_tarballs.jl
Fenyutanchan: This is for upgrading. See #6488 and last failed building logs for reasons.

Fenyutanchan:
Waiting for #6475 and #6478~

K/Kira/build_tarballs.jl
Fenyutanchan: I opened PR #6489 for updating the meson.

K/Kira/build_tarballs.jl
```diff
@@ -0,0 +1,81 @@
...
+cd ${WORKSPACE}/srcdir/Kira-build/
+
+meson setup \
+    -Dfirefly=true \
+    -Djemalloc=true \
+    --cross-file=${MESON_TARGET_TOOLCHAIN} \
+    --buildtype=release \
+    ${WORKSPACE}/srcdir/kira/ \
+    ${WORKSPACE}/srcdir/Kira-build/
+sed -i "s/\/workspace\/destdir\/opt/\/opt/g" build.ninja
```
giordano, line 30: Why do you need to do this?  Also, if you really need to do this
```suggestion
sed -i "s?/workspace/destdir/opt?/opt?g" build.ninja
```
is more readable

Fenyutanchan: See #6488.
I have checked the meson build configurations in this project.
However, I didn’t see the cause of the problem.
Therefore, I use this line for fixing it.

PR #6501, opened 2023-04-03
[dmg2img] Set OpenSSL_jll compat to 1.1.10, add julia_compat 1.6

mkitti:


D/dmg2img/build_tarballs.jl
```diff
@@ -33,8 +33,8 @@ dependencies = [
     # Future versions of bzip2 should allow a more relaxed compat because the
     # soname of the macOS library shouldn't change at every patch release.
     Dependency("Bzip2_jll", v"1.0.6"; compat="=1.0.6"),
```
giordano, line 35:
```suggestion
    Dependency("Bzip2_jll"; compat="1.0.8"),
```

PR #6508, opened 2023-04-04
[PulseAudio] Set compat for OpenSSL

giordano:


giordano:
Alright, pulseaudio is a trashfire I have zero intention to spend one more second with, so I'm just going to fix list of platforms and that's it.

PR #6518, opened 2023-04-04
[entwine] Switch to GitSource, openssl compat to 1.1.10

mkitti:


mkitti:
This needs a patch. Maybe just skip build here?

giordano:
Yeah

PR #6532, opened 2023-04-05
[SCALE_MAMBA] Upgrade to 1.14, Modernize, set OpenSSL compat to 1.1.10

mkitti:


mkitti:
I'm also considering moving to [6449e807c99c68203f6584166a7130055da52adb](https://github.com/KULeuven-COSIC/SCALE-MAMBA/commit/6449e807c99c68203f6584166a7130055da52adb) for 1.14

mkitti:
I can keep poking at this. However, I do wonder if anyone actually needs this? Is there even an existing JLL?

mkitti:
`julia --project=. build_tarballs.jl x86_64-linux-gnu-march+avx` works locally

mkitti:
The current failure on macOS is

```
  | [21:24:58] System/RunTime.cpp:41:17: error: redefinition of 'time_s' with a different type: 'struct timeval' vs 'struct timespec'
  | [21:24:58] struct timespec time_s= {0, sleep_val};
  | [21:24:58]                 ^
  | [21:24:58] /workspace/srcdir/SCALE-MAMBA/src/config.h:177:24: note: previous declaration is here
  | [21:24:58] extern struct timespec time_s;
  | [21:24:58]                        ^
  | [21:24:58] 8 warnings and 1 error generated.
  | [21:24:58] make: *** [Makefile:29: System/RunTime.o] Error 1

```

mkitti:
There is a macOS specific typedef here:
https://github.com/KULeuven-COSIC/SCALE-MAMBA/blob/c111516e3ebc1efd12a2bd47dd2122b160e13d1e/src/Tools/Timer.h#L19-L21

mkitti:
Might need this patch for macOS:

```diff
diff --git a/src/config.h b/src/config.h
index 43d18d4..f082a1f 100644
--- a/src/config.h
+++ b/src/config.h
@@ -173,9 +173,11 @@ All rights reserved
  *    sleep_val = 252000000L;
  * This value seems to give the best performance on our machines
  */
+#define _POSIX_C_SOURCE 199309L
 #include <time.h>
-extern struct timespec time_s;
 #define sleep_val 250000000L
+#define timespec timeval
+struct timespec time_s= {0, sleep_val};
 
 // This defines the memory size as 2^MEMSIZE for the different
 // memory types

```

mkitti:
Also  we have this error:
`Online/Online.cpp:63:11: error: no matching function for call to 'nanosleep'           nanosleep(&time_s, NULL);`

mkitti:
@giordano could we lose the macOS support here?

S/SCALE_MAMBA/build_tarballs.jl
```diff
@@ -35,16 +36,13 @@ fi
 make -j${nproc} CC="${CXX}" -C src
 
 # Install executables into ${prefix}
-mkdir -p ${prefix}/bin
-cp Player.x${exe} ${prefix}/bin/
-cp Setup.x${exe} ${prefix}/bin/
+install -Dvm 755 "Player.x${exe}" "${bindir}/Player.x${exe}"
```
giordano, line 39: Did you mean exeext?

mkitti: I have no idea. This was you:
https://github.com/JuliaPackaging/Yggdrasil/pull/6532/commits/c01efa6b231b0f41f7c36e764fa3614ee65b8366

S/SCALE_MAMBA/build_tarballs.jl
```diff
@@ -35,16 +36,13 @@ fi
...
+install -Dvm 755 "Setup.x${exe}" "${bindir}/Setup.x${exe}"
```
mkitti, line 40:
```suggestion
install -Dvm 755 "Player.x${exeext}" "${bindir}/Player.x${exeext}"
install -Dvm 755 "Setup.x${exeext}" "${bindir}/Setup.x${exeext}"
```

S/SCALE_MAMBA/build_tarballs.jl
giordano: The environment variable I mean, there's no exe, only exeext (and I didn't touch it there): https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables. It doesn't make much difference here since exeext is the empty string for Linux, but using a non-existing variable doesn't do any good

PR #6539, opened 2023-04-06
[LibTracyClient] Upgrade to v0.9.1

topolarity:
This also includes a change to the selected feature flags: automatic sampling is disabled via `TRACY_NO_SAMPLING`


topolarity:
By the way, do we typically only split by version if we expect that patch releases might still occur on older branches?

Tracy development is pretty linear, so I think we might like to merge this into one version variant

vchuravy:
> By the way, do we typically only split by version if we expect that patch releases might still occur on older branches?

Yeah essentially if you expect to be making releases for older versions due to Julia LTS or something like that.

PR #6546, opened 2023-04-07
[LaMEM] v1.2.4

boriskaus:
This bumps LaMEM to version 1.2.4, which fixes a bug that prevented it from running on Apple Silicon machines


boriskaus:
oops, not sure why this is uploading all the PETSc files as well; let me undo that

PR #6549, opened 2023-04-07
[HSL] Compile v1.0.0

amontoison:
@dpo @ViralBShah @staticfloat

With the HSL team, we developed a `JuliaHSL` package that contains all the [HSL packages](https://www.hsl.rl.ac.uk/catalogue/) together.
Because the package `JuliaHSL` requires an academic license, we can't precompile it here with Yggdrasil.
To ease the use of it in the Julia ecosystem, we compile a dummy version here such that we can add it as a dependency of any Julia package.
The dummy version only has one [C routine](https://github.com/ralna/JuliaHSL/blob/main/dummy/juliahsl.c#L3-L5) and the users will be able to call it to check if they have the dummy version or not and potentially call other routines.
```julia
using HSL_jll

# Function also available in the Julia interface HSL.jl
function JULIAHSL_isfunctional()
    @ccall libhsl.JULIAHSL_isfunctional()::Bool
end

if JULIAHSL_isfunctional()
  ...  # use an HSL linear solver
else
  ...  # use a SuiteSparse linear solver
end
```

The users will be able to download the `JuliaHSL` package (source code) with the "real" `HSL_jll.jl` Julia package (for Julia >= 1.9) from https://licences.stfc.ac.uk/.
It requires Julia v1.9 because we want to use LBT on all platforms.

The real `HSL_jll` has the same products than the dummy `HSL_jll.jl` package generated here. The main difference is that we have an `override` folder with the content of the "real" tarballs precompiled with BinaryBuilder.jl at the Rutherford Appleton Laboratory (RAL).

Thus, the RAL will provide these archives:
- JuliaHSL.zip (source code)
- HSL_jll_x86_64-apple-darwin-libgfortran4.zip (Julia package)
- HSL_jll_x86_64-linux-gnu-libgfortran5.zip (Julia package)
- HSL_jll_x86_64-w64-mingw32-libgfortran5.zip (Julia package)
- HSL_jll_aarch64-apple-darwin-libgfortran5 (Julia package)
- ...

The user will choose the archive based on the result of `Base.BinaryPlatforms.host_triplet()` and install it with just a `dev .`.

Comments:
- We also need a dummy version because Julia must downloads an artifact from the `Artifacts.toml` even if we have an `override` folder. Julia will not use it but it returns an error if it can't download an archive;
- The `hsl_ma57` solver is used for sparse `\` in MATLAB.


H/HSL/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libhsl", :libhsl)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="METIS_jll", uuid="d00139f3-1899-568f-a2f0-47f597d42d70"))
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    Dependency(PackageSpec(name="MPICH_jll", uuid="7cb0a576-ebde-5e09-9194-50597f1243b4"); platforms=filter(!Sys.iswindows, platforms)),
```
amontoison, line 35:
```suggestion
    Dependency(PackageSpec(name="MPICH_jll", uuid="7cb0a576-ebde-5e09-9194-50597f1243b4"); platforms=filter(!Sys.iswindows, platforms))
```

H/HSL/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+    Dependency(PackageSpec(name="MicrosoftMPI_jll", uuid="9237b28f-5490-5468-be7b-bb81f5f5e6cf"); platforms=filter(Sys.iswindows, platforms)),
```
amontoison, line 36:
```suggestion
    Dependency(PackageSpec(name="MicrosoftMPI_jll", uuid="9237b28f-5490-5468-be7b-bb81f5f5e6cf"); platforms=filter(Sys.iswindows, platforms))
```

vchuravy:
I am not sure why a `jll` package is necessary? You could just handle this in a normal Julia package and allow the user to set the path to the library they obtained from you?

amontoison:
@vchuravy The shared library that we precompiled requires some dependencies (METIS, BLAS/LAPACK, MPI). By using a `jll` package we are sure that they installed and loaded on all platforms before any call to `libhsl.${dlext}`.

vchuravy:
Your non JLL package can also depend on these dependencies. JLL packages are not special in that way.

amontoison:
Ok so an alternative solution is to add `METIS_jll`, `LBT_jll`, `MicrosoftMPI_jll` and `MPICH_jll` in the Project.toml of HSL.jl and also an option to provide the path to the shared library `libhsl`.

The drawbacks that I have in mind are:
- The Windows users must add an environment variable and some of them don't know how to do that;
- If we compile the tarballs with different compat entries of the dependencies or Julia in the future, a Project.toml for each new `jll` is preferable otherwise we will not be able to support multiple version of JuliaHSL in HSL.jl; 
- We can't use the `include` directory of `HSL_jll` to regenerate or update the Julia wrappers;
- One special application of HSL_jll is with Ipopt.jl that can dynamically load it to use HSL solvers (https://github.com/jump-dev/Ipopt.jl/issues/342) and the `jll` package is probably the most user-friendly form.

@odow Can you give your point of view too?

odow:
I think @vchuravy's point was that we don't need this to be  a jll with a C dependency.

The actual `HSL_jll.jl` that you download can be different structure to the dummy package. 

We register an `HSL_jll.jl` dummy package without needing to compile via Yggdrasil. There's nothing special about the `_jll` suffix in the  package manager.

odow:
> The drawbacks that I have in mind are:

> * The Windows users must add an environment variable and some of them don't know how to do that;

I don't understand this. Can you explain why it's needed?

amontoison:
> > The drawbacks that I have in mind are:
> 
> > * The Windows users must add an environment variable and some of them don't know how to do that;
> 
> I don't understand this. Can you explain why it's needed?

I thought that Valentin wanted that we give a path to the precompiled shared library `libhsl` with an environment variable like:
- https://github.com/JuliaSmoothOptimizers/QRMumps.jl/blob/main/src/QRMumps.jl#L8-L15
- https://github.com/JuliaSmoothOptimizers/MUMPS.jl/blob/main/src/MUMPS.jl#L35-L42

amontoison:
> I think @vchuravy's point was that we don't need this to be a jll with a C dependency.
> 
> The actual `HSL_jll.jl` that you download can be different structure to the dummy package.
> 
> We register an `HSL_jll.jl` dummy package without needing to compile via Yggdrasil. There's nothing special about the `_jll` suffix in the package manager.

Ok, I misunderstood the point of @vchuravy. 
He probably wanted to say that we don't need an `HSL_jll.jl` package with the current form.
I thought that he said that we didn't need a  dummy `HSL_jll.jl` at all.

For the current use of `HSL_jll.jl` in HSL.jl and Ipopt.jl, we could do what you suggest.
My last worry is for the future `GALAHAD_jll`, [GALAHAD](https://github.com/ralna/GALAHAD) is a package developed by the HSL team (like [SPRAL](https://github.com/ralna/spral)) that we plan to precompile with Ygddrasil.
It uses the HSL solvers as optional dependencies and we are working a new release with a Julia interface.

GALAHAD doesn't provide an option to load a `libhsl` library externally like Ipopt.
What they did is different, they developed dummy Fortran files (cc @jfowkes) :
- [hsl_ad02s.f90](https://github.com/ralna/GALAHAD/blob/master/src/dum/hsl_ad02s.f90)
- [hsl_ad02d.f90](https://github.com/ralna/GALAHAD/blob/master/src/dum/hsl_ad02d.f90)
- [hsl_ma57s.f90](https://github.com/ralna/GALAHAD/blob/master/src/dum/hsl_ma57s.f90)
- [hsl_ma57d.f90](https://github.com/ralna/GALAHAD/blob/master/src/dum/hsl_ma57d.f90)
- [hsl_ma97s.f90](https://github.com/ralna/GALAHAD/blob/master/src/dum/hsl_ma97s.f90)
- [hsl_ma97d.f90](https://github.com/ralna/GALAHAD/blob/master/src/dum/hsl_ma97d.f90)
- ...

such that we can compile them here in the shared library of the dummy `HSL_jll` and link the future `GALAHAD_jll` with it.
In this way, if the user has the "real" `HSL_jll.jl`, it will be able to use the HSL solvers without doing anything.

jfowkes:
The HSL team is happy to have this dummy JuliaHSL v1.0.0 package.

This can then be bumped to v2.0.0 for GALAHAD in future once its Julia interfaces are ready.

ViralBShah:
@amontoison Please merge when you feel this is ready.

amontoison:
@ViralBShah Can you merge it?
I'm mot authorized to merge pull requests for this repository.

ViralBShah:
@amontoison I have invited you to have Yggdrasil commit access.

amontoison:
Thanks @ViralBShah and @giordano for the invitations!

I found a way to have a `HSL_jll.jl`, I modified the the files in `HSL_jll.jl/src/wrapper` such that the shared libraries of all platforms coexist in the `override` folder.
```julia
function __init__()
    JLLWrappers.@generate_init_header(CompilerSupportLibraries_jll, MPICH_jll, METIS_jll, libblastrampoline_jll)
    JLLWrappers.@init_library_product(
        libhsl,
        "lib/aarch64-apple-darwin-libgfortran5/libhsl.dylib",
        RTLD_LAZY | RTLD_DEEPBIND,
    )

    JLLWrappers.@generate_init_footer()
end  # __init__()
```
```julia
function __init__()
    JLLWrappers.@generate_init_header(CompilerSupportLibraries_jll, MicrosoftMPI_jll, METIS_jll, libblastrampoline_jll)
    JLLWrappers.@init_library_product(
        libhsl,
        "bin\\x86_64-w64-mingw32-libgfortran5\\libhsl.dll",
        RTLD_LAZY | RTLD_DEEPBIND,
    )

    JLLWrappers.@generate_init_footer()
end  # __init__()
```
```julia
function __init__()
    JLLWrappers.@generate_init_header(CompilerSupportLibraries_jll, MPICH_jll, METIS_jll, libblastrampoline_jll)
    JLLWrappers.@init_library_product(
        libhsl,
        "lib/x86_64-linux-gnu-libgfortran5/libhsl.so",
        RTLD_LAZY | RTLD_DEEPBIND,
    )

    JLLWrappers.@generate_init_footer()
end  # __init__()
```

it works perfectly except for Apple platforms,
```julia
julia> using HSL_jll
ERROR: InitError: could not load library "/Users/alexis/Downloads/HSL_jll/override/lib/x86_64-apple-darwin-libgfortran4/libhsl.dylib"
dlopen(/Users/alexis/Downloads/HSL_jll/override/lib/x86_64-apple-darwin-libgfortran4/libhsl.dylib, 1): no suitable image found.  Did find:
	/Users/alexis/Downloads/HSL_jll/override/lib/x86_64-apple-darwin-libgfortran4/libhsl.dylib: code signature in (/Users/alexis/Downloads/HSL_jll/override/lib/x86_64-apple-darwin-libgfortran4/libhsl.dylib) not valid for use in process using Library Validation: library load disallowed by system policy
```
I generated the tarballs on my Linux (Ubuntu 22.04 -- x86_64-linux-gnu-libgfortran5).
Are you using Apple computers to sign the tarballs with Yggdrasil?

giordano:
> Are you using a Mac to sign the tarballs with Yggdrasil?

No, we sign on Linux, like the rest of the building pipeline. But I'm not sure this answers your question.

amontoison:
It's a good news that you sign the tarballs on Linux.
Is it possibly that the tarballs are not signed if we use the option `--deploy="local"`?

We use the command
```julia
julia --color=yes build_tarballs.jl aarch64-apple-darwin-libgfortran5,aarch64-linux-gnu-libgfortran3,aarch64-linux-gnu-libgfortran4,aarch64-linux-gnu-libgfortran5,aarch64-linux-musl-libgfortran3,aarch64-linux-musl-libgfortran4,aarch64-linux-musl-libgfortran5,powerpc64le-linux-gnu-libgfortran3,powerpc64le-linux-gnu-libgfortran4,powerpc64le-linux-gnu-libgfortran5,x86_64-apple-darwin-libgfortran3,x86_64-apple-darwin-libgfortran4,x86_64-apple-darwin-libgfortran5,x86_64-linux-gnu-libgfortran3,x86_64-linux-gnu-libgfortran4,x86_64-linux-gnu-libgfortran5,x86_64-linux-musl-libgfortran3,x86_64-linux-musl-libgfortran4,x86_64-linux-musl-libgfortran5,x86_64-unknown-freebsd-libgfortran3,x86_64-unknown-freebsd-libgfortran4,x86_64-unknown-freebsd-libgfortran5,x86_64-w64-mingw32-libgfortran3,x86_64-w64-mingw32-libgfortran4,x86_64-w64-mingw32-libgfortran5 --debug --verbose --deploy="local"
```
to generate the tarballs for all 64-bits platforms.

giordano:
Honestly I don't understand the problem.  The library is already signed during audit as far as I can tell: https://buildkite.com/julialang/yggdrasil/builds/2175#01876299-9fbf-439c-8655-972f62f972e1/638-770

PR #6553, opened 2023-04-10
[LAPACK] Update LAPACK to 3.11

ViralBShah:


ViralBShah:
@amontoison What do you think about just having one LAPACK package that has both the ILP64 and LP64 versions? And discontinue LAPACK32?

ViralBShah:
We have `OpenBLAS32` because that is not part of Julia, and then I followed that for a few others. But it occurs to me that it would be simpler overall to just build both versions in the same BB package if the symbol and library names are separate.

PR #6555, opened 2023-04-12
[SCS_GPU] update to latest commit

odow:
I don't have a GPU, so I haven't tested this. 

But monkey see, monkey do: https://github.com/jump-dev/SCS.jl/pull/267. The tests are failing and this warning is printed:

cc @kalmarek 

@bodono: is this correct?

### DIFF
--- a/S/SCS_GPU/build_tarballs.jl
+++ b/S/SCS_GPU/build_tarballs.jl
@@ -12,7 +12,7 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/scs*
-flags="DLONG=0 USE_OPENMP=0 DINDIRECT=1"
+flags="DLONG=0 USE_OPENMP=0 INDIRECT=1"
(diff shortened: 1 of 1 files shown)


kalmarek:
I thought that `libscsgpuindir.so` requires `SCS_INDIR_O` which does set the `INDIRECT` flag correctly:
https://github.com/cvxgrp/scs/blob/master/Makefile#L36-L37

But now I see that `libscsgpuindir.$(SHARED)` asks for `SCS_O` instead (which does not set the flag:
https://github.com/cvxgrp/scs/blob/master/Makefile#L154-L156

@bodono is this a bug?

kalmarek:
Indeed I've just modified locally `Makefile` of scs and now (locally) `SCS.GpuIndirectSolver` passes all tests ;)

odow:
Okay, let's wait for https://github.com/cvxgrp/scs/pull/251

odow:
This uses the latest commit with the GPU fix. Not changing the version number so this can be a `+1` release. It's just the old version plus the fix and a couple of documentation commits.

PR #6568, opened 2023-04-13
[M4] Add M4 v1.4.19

staticfloat:


M/M4/build_tarballs.jl
```diff
@@ -0,0 +1,29 @@
...
+
+script = raw"""
+cd $WORKSPACE/srcdir/m4-*/
+
+./configure --prefix=${prefix} --host=${target}
+make -j${nprocs}
+make install
+"""
+
+# Windows platforms fail with messages about `__memcpy_chk`
```
giordano, line 19: If you really care about Windows, I think you need to link to `libssp` for `__memcpy_chk`

M/M4/build_tarballs.jl
```diff
@@ -0,0 +1,29 @@
...
+platforms = filter(p -> !Sys.iswindows(p), supported_platforms())
```
giordano, line 20:
```suggestion
platforms = filter(!Sys.iswindows, supported_platforms())
```

M/M4/build_tarballs.jl
```diff
@@ -0,0 +1,29 @@
...
+
+products = [
+    ExecutableProduct("m4", :m4),
+]
+
+dependencies = [
```
giordano, line 26:
```suggestion
dependencies = Dependency[
```

M/M4/build_tarballs.jl
staticfloat: I don't care about Windows (this is for host tools in BB2) but looks like `-fstack-protector` is what is required to get this to work; thanks for the tip!

PR #6572, opened 2023-04-13
Regenerate neper_jll with a compat entry for SCOTCH_jll

amontoison:


giordano:
You have to fix this stuff in the registry first

PR #6573, opened 2023-04-13
[file] Add GNU file `v5.44`

staticfloat:
It's incredibly difficult to search for this tool online.

### DIFF
--- a/F/file/build_tarballs.jl
+++ b/F/file/build_tarballs.jl
@@ -33,7 +33,7 @@ platforms = filter(p -> !(libc(p) == "musl" && arch(p) == "i686"), platforms)
 products = [
     ExecutableProduct("file", :file)
 ]
-dependencies = Dependency[
+dependencies = [
     Dependency("Bzip2_jll"),
     Dependency("XZ_jll"),
     Dependency("Zlib_jll"),
--- a/F/file/build_tarballs.jl
+++ b/F/file/build_tarballs.jl
@@ -1,0 +1,41 @@
+using BinaryBuilder
+
+name = "file"
+
+# NOTE: the cross-compilation story of `file` is kind of broken
+# and requires a two-step build; we dodge this by providing `file`
+# in the rootfs right now, but this locks us to exactly this version.
+version = v"5.41"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/file/file.git",
+              "504206e53a89fd6eed71aeaf878aa3512418eab1")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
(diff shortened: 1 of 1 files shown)


F/file/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+
+# Disable windows for now, as that requires `libgnurx`.
+platforms = filter(!Sys.iswindows, supported_platforms())
+# Disable i686-linux-musl because we end up in dynamic linker hell
+platforms = filter(p -> !(libc(p) == "musl" && arch(p) == "i686"), platforms)
+
+products = [
+    ExecutableProduct("file", :file)
+]
+dependencies = Dependency[
```
giordano, line 36:
```suggestion
dependencies = [
```

F/file/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+
+# NOTE: the cross-compilation story of `file` is kind of broken
+# and requires a two-step build; we dodge this by providing `file`
+# in the rootfs right now, but this locks us to exactly this version.
+version = v"5.41"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/file/file.git",
+              "504206e53a89fd6eed71aeaf878aa3512418eab1")
```
fingolfin, line 13: Is there a reason you use a git snapshot instead of a release tarball from http://ftp.astron.com/pub/file/ ?

giordano: There are different schools of thought about this, I believe @staticfloat is squarely in the "always use git repositories whenever possible" camp, I personally prefer release tarballs when available.

fingolfin: the thing is, git snapshots may be different from what's in a tarball (e.g. obviously the `configure` script is missing in the former). I just found that I eliminate a whole class of issues by preferring release tarballs (at least for projects that handle them sensibly; and of course if I need a bleeding edge version, then a git snapshot may be the only alternative...).

Aaaanyway, I did not mean to block anything, I was just wondering

giordano: Yeah, avoiding the autoconf bootstrap stage is one of the main reasons why I prefer a release tarball (in some cases it's non-trivial), in addition to requiring a much smaller download.

staticfloat: I do like git repositories, but honestly it's because the official webpage lists an `ftp://` server as the official download location, and chrome refused to open it for me.  So rather than just try `http://` with the same URL, I googled "gnu file download source code" and found a github mirror.

If I had run into bootstrapping issues, I would have searched harder for a release tarball.  ;)

PR #6575, opened 2023-04-14
Regenerate PTSCOTCH_jll with a compat entry for SCOTCH_jll

amontoison:
I don't understand why `libgfortran5` can't be found for `x86_64-linux-gnu-mpi+mpitrampoline` :thinking: .


giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/b3dbe533b88292cc0ffd246369e672ba5f050288/P/PETSc/build_tarballs.jl#L251

PR #6579, opened 2023-04-14
[gawk] Add v5.2.1

staticfloat:
Windows is currently disabled because of an issue in our mingw setup.

### DIFF
--- a/G/gawk/build_tarballs.jl
+++ b/G/gawk/build_tarballs.jl
@@ -1,0 +1,34 @@
+using BinaryBuilder
+
+name = "gawk"
+version = v"5.2.1"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://ftp.gnu.org/gnu/gawk/gawk-$(version).tar.xz",
+                  "673553b91f9e18cc5792ed51075df8d510c9040f550a6f74e09c9add243a7e4f")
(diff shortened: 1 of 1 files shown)


G/gawk/build_tarballs.jl
```diff
@@ -0,0 +1,34 @@
...
+make install
+install_license COPYING
+"""
+
+# Windows currently fails due to a problem with mingw headers (langinfo.h) not being found
+platforms = filter(!Sys.iswindows, supported_platforms())
+products = [
+    ExecutableProduct("gawk", :gawk)
+]
+dependencies = []
```
giordano, line 33:
```suggestion
dependencies = Dependency[]
```

staticfloat: I did this because you told me to get rid of this in my last PR: 
https://github.com/JuliaPackaging/Yggdrasil/pull/6573#discussion_r1166129576

Which do you prefer?

giordano: Oh, it was useless there, and forcing the type to Dependency of a non-empty vector risks converting the elements, if they have different types 🙂 I eventually to force `dependencies` to be `AbstractArray{<:AbstractDependency}`, it just shouldn't be `Vector{Any}`

PR #6581, opened 2023-04-14
[Libtool] Clean up, add more executable products

staticfloat:


giordano:
Windows :smiling_face_with_tear:

giordano:
> libtool and libtoolize are shell scripts 
>
> So they wouldn't have a `.exe` extension.  Heh.

Ah, that's an excellent point. The audit should really check that an `ExecutableProduct` is a binary executable, because that's what it's supposed to be

PR #6591, opened 2023-04-16
New recipe: Soup3

lukebemish:
Here's my attempt to get libsoup3 bundled, assuming I didn't mess anything up locally. I based this off some of the scripts in https://github.com/JuliaPackaging/Yggdrasil/pull/5689.

### DIFF
--- a/S/Soup3/build_tarballs.jl
+++ b/S/Soup3/build_tarballs.jl
@@ -28,9 +28,9 @@ if [[ "${target}" != *darwin* ]]; then
     sed -i "s/-R/-Wl,-rpath=/" build.ninja
 fi
 
-if [[ "${target}" != *darwin* && "${target}" != *w64* ]]; then
+if [[ "${target}" == *86*-linux-gnu* ]]; then
     # https://stackoverflow.com/questions/2418157/c-error-undefined-reference-to-clock-gettime-and-clock-settime
-    sed -i "s/\$ARGS -o \$out \$in \$LINK_ARGS/\$ARGS -o \$out \$in \$LINK_ARGS -lrt/" build.ninja
+    sed -i 's/$ARGS -o $out $in $LINK_ARGS/$ARGS -o $out $in $LINK_ARGS -lrt/' build.ninja
 fi
 
 ninja -j${nproc}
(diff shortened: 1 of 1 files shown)


S/Soup3/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+    -Ddocs=disabled \
+    -Dsysprof=disabled \
+    ..
+
+if [[ "${target}" != *darwin* ]]; then
+    # https://github.com/rbgirshick/py-faster-rcnn/issues/706
+    sed -i "s/-R/-Wl,-rpath=/" build.ninja
+fi
+
+if [[ "${target}" != *darwin* && "${target}" != *w64* ]]; then
```
giordano, line 31: I don't think this is very accurate:
```suggestion
if [[ "${target}" == *86*-linux-gnu* ]]; then
```

S/Soup3/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+    # https://stackoverflow.com/questions/2418157/c-error-undefined-reference-to-clock-gettime-and-clock-settime
+    sed -i "s/\$ARGS -o \$out \$in \$LINK_ARGS/\$ARGS -o \$out \$in \$LINK_ARGS -lrt/" build.ninja
```
giordano, line 33: If you use single quotes you don't have to escape every single dollar sign, and this becomes much more readable:
```suggestion
    sed -i 's/$ARGS -o $out $in $LINK_ARGS/$ARGS -o $out $in $LINK_ARGS -lrt/' build.ninja
```

S/Soup3/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+products = [
+    LibraryProduct(["libsoup", "libsoup-3", "libsoup-3.0"], :libsoup),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    # Host gettext needed for "msgfmt"
+    HostBuildDependency("Gettext_jll"),
+    # Gettext is only needed on macOS
+    Dependency("Gettext_jll", v"0.21.0"; compat="=0.21.0", platforms=filter(Sys.isapple, platforms)),
```
giordano, line 54:
```suggestion
    Dependency("Gettext_jll"; compat="=0.21.0", platforms=filter(Sys.isapple, platforms)),
```
Also, are you sure we cannot avoid it at all?  Gettext is a terrible dependency to have

S/Soup3/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+    Dependency("Glib_jll"; compat="2.74.0"),
+    Dependency("SQLite_jll"),
+    Dependency("nghttp2_jll"),
+    Dependency("brotli_jll"),
+    Dependency("libpsl_jll"),
+    Dependency("LibUnwind_jll"),
```
giordano, line 60: Are you sure this is really needed?  Libunwind isn't a common dependency.

S/Soup3/build_tarballs.jl
lukebemish: I'm pretty sure variable interpolation doesn't work inside single quotes, though it's been a while

giordano: That's the point.

lukebemish: Oh wait, I misread that line. Thanks

PR #6602, opened 2023-04-18
HDF5 cross compile for macOS M1 and Windows

mkitti:
This is a test pull request to check if the new configuration files work.


imciner2:
This has a lot of conflicts with the current master branch, so is this something that will ever be able to be resurrected (or is it even needed anymore)? If not, let's close this.

PR #6609, opened 2023-04-18
[SDPA] add methods to sdpawrap.cpp

odow:
I don't really understand what I'm doing, but this looks correct to me.

x-ref https://github.com/jump-dev/SDPA.jl/issues/49

cc @blegat

### DIFF
--- a/S/SDPA/build_tarballs.jl
+++ b/S/SDPA/build_tarballs.jl
@@ -26,7 +26,7 @@ julia_compat = join("~" .* string.(getfield.(julia_versions, :major)) .* "." .* 
 
 name = "SDPA"
 upstream_version = v"7.3.8"
-version_offset = v"0.0.1" # reset to 0.0.0 once the upstream version changes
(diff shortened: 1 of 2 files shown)


odow:
I don't understand the compilation failures with the `(g|s)etDisplay` methods: https://buildkite.com/julialang/yggdrasil/builds/2415#01879757-da46-48be-95fa-15f9f98e69ec

> error: invalid use of incomplete type ‘struct _IO_FILE’

Do we need to handle `FILE*` arguments differently somewhere? The headers for the methods are:
```cpp
void setDisplay(FILE* Display  = stdout);
FILE* getDisplay();
```

giordano:
I don't have much time to look into this, but https://www.openwall.com/lists/musl/2021/10/06/4 is likely relevant.

odow:
@blegat do you want to take a look at this?

blegat:
Maybe we can merge with `getDisplay`/`setDisplay` for now ?

ViralBShah:
Should we merge?

blegat:
No objection from me

PR #6613, opened 2023-04-20
Address a missing extern "C" statement

MathieuDutSik:
The right solution would be to upgrade the version of nauty, but this would require more work.


fingolfin (approved):
I see no harm in adding this.

Though you don't really *need* this, because you can just do this in your C++ code:
```
extern "C" {
#include "traces.h"
}
```

PR #6627, opened 2023-04-21
New Recipe: RSF v1.0.0

mloubout:


R/RSF/build_tarballs.jl
```diff
@@ -0,0 +1,1888 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir
+
+# install_license
+install_license src/COPYING.txt
+
+# Build
+cd src/
+
+./configure --prefix=${prefix} CC=$(which gcc) CXX=$(which g++) CAIROPATH=${includedir}/cairo FFMPEGPATH=${includedir}/libavcodec/ AR=$(which ar) CPPPATH=${includedir} LIBPATH=${libdir} BLAS=openblas
```
giordano, line 23: Why `CC=$(which gcc)`, `CXX=$(which g++)` and `AR=$(which ar)` (the former two are also wrong)?  The corresponding environment variable are already set to the right values.

R/RSF/build_tarballs.jl
```diff
@@ -0,0 +1,1888 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    Dependency(PackageSpec(name="SuiteSparse_jll", uuid="bea87d4a-7f5b-5778-9afe-8cc45184846c"))
+    Dependency(PackageSpec(name="FFMPEG_jll", uuid="b22a6f82-2f65-5046-a5b2-351ab43fb4e5"))
+    Dependency(PackageSpec(name="Cairo_jll", uuid="83423d85-b0ee-5818-9007-b63ccbeb887a"))
+    Dependency(PackageSpec(name="MPICH_jll", uuid="7cb0a576-ebde-5e09-9194-50597f1243b4"))
+    Dependency(PackageSpec(name="MicrosoftMPI_jll", uuid="9237b28f-5490-5468-be7b-bb81f5f5e6cf"))
```
giordano, line 1881: For MPI please look at https://github.com/JuliaPackaging/Yggdrasil/blob/3f318236dead5863a703287db226946b62b83794/P/P4est/build_tarballs.jl#L66-L70 https://github.com/JuliaPackaging/Yggdrasil/blob/3f318236dead5863a703287db226946b62b83794/P/P4est/build_tarballs.jl#L78

giordano (commented):
As a general rule working with Git and GitHub, opening a PR from a branch with the same name as the target branch is [anti-pattern](https://blog.jasonmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/).  You should always create a ***new branch***.

R/RSF/build_tarballs.jl
mloubout: The configure is using `scons` which doesn't find any of the compiles from the env vars unless specified here so all three `gcc/g++/ar` are needed or it fails.

PR #6631, opened 2023-04-23
[llama_cpp] Bump version

marcom:


L/llama_cpp/build_tarballs.jl
```diff
@@ -85,12 +92,15 @@ products = [
...
+    ExecutableProduct("vdot", :vdot),
     LibraryProduct("libllama", :libllama),
 ]
 
 dependencies = Dependency[
 ]
 
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat="1.6", preferred_gcc_version = v"8.1.0")
+               julia_compat="1.6", preferred_gcc_version = v"12")
```
giordano, line 106: GCC 12 with C++ code won't work with any released version of julia, as simple as that.

L/llama_cpp/build_tarballs.jl
```diff
@@ -1,12 +1,18 @@
...
 # url = "https://github.com/ggerganov/llama.cpp"
 # description = "Port of Facebook's LLaMA model in C/C++"
 
 # NOTES
+# - _mm_loadu_si64 needs at least gcc-11.3.0
+#   https://stackoverflow.com/questions/72837929/mm-loadu-si32-not-recognized-by-gcc-on-ubuntu
+#   gcc-11.1.0 will silently produce wrong code, so we have to use gcc-12
+#   https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99754
+# - windows and mmap working?
+#   warning: You are building for pre-Windows 8; prefetch not supported
```
giordano, line 15: Would defining [`_WIN32_WINNT`/`WINVER`](https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170) appropriately help here since Julia [requires Windows 10](https://julialang.org/downloads/#supported_platforms) anyway?  Maybe we should just set those macros for Windows 10 in BB directly, PRs welcome.

L/llama_cpp/build_tarballs.jl
marcom: Can we bump gcc-11.1.0 to gcc-11.3.0 in 0_RootFS in Yggdrasil? Not sure how painful that is. gcc-11.1.0 will produce wrong code as mentioned in the NOTES for one of the intrinsics used.

giordano: Actually
```console
% nm -D libllama.so | grep GLIBCXX_3.4.2 | cut -d'@' -f2|sort -u
GLIBCXX_3.4.20
GLIBCXX_3.4.21
GLIBCXX_3.4.22
GLIBCXX_3.4.29
```
So there isn't much difference between GCC 12.1 and 11.3 and apparently julia v1.6.7 has libstdc++.so.6.0.29, I haven't checked previous versions

L/llama_cpp/build_tarballs.jl
marcom: I will try a patch here for llama_cpp.

Where would the patch for BB go?

giordano: > I will try a patch here for llama_cpp.

What patch?  It's just a macro to pass to the compiler.

> Where would the patch for BB go?

https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/df99a14849abf2f97aef595de8265acea6877b1e/src/Runner.jl#L461-L482
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/df99a14849abf2f97aef595de8265acea6877b1e/src/Runner.jl#L316-L337

marcom: PR for BB here: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/308

PR #6632, opened 2023-04-23
[louvain_jll] Adding macOS aarch64

KrainskiL:
Looks like louvain_jll long time ago was not built for M-Series Macs and it's causing an issue now.

### DIFF
--- a/L/louvain/build_tarballs.jl
+++ b/L/louvain/build_tarballs.jl
@@ -7,13 +7,13 @@ version = v"0.2.0" # Rebuilding louvain_jll to include macOS aarch64
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/KrainskiL/louvain.git", "9e3e081f054e531e32af38fe71d2991f954ea347")
+    GitSource("https://github.com/KrainskiL/louvain.git", "3e9efdb930efc4a3715d0978ea08ed52172f7231")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/louvain/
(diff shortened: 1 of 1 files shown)


L/louvain/build_tarballs.jl
```diff
@@ -14,9 +14,8 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/louvain/
 make -j${nproc}
```
giordano, line 16: Uhm, this is forcing `g++` as compiler everywhere, which is wrong

giordano:
```suggestion
make -j${nproc} CXX="${CXX}"
```

giordano: Side note: https://github.com/KrainskiL/louvain/blob/9e3e081f054e531e32af38fe71d2991f954ea347/Makefile#L4 `-O5` looks a bit too high

giordano: https://github.com/KrainskiL/louvain/pull/1

L/louvain/build_tarballs.jl
```diff
@@ -13,10 +13,9 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/louvain/
-make -j${nproc}
-mkdir -p ${bindir}
+make -j${nproc} CXX="${CXX}"
```
giordano, line 16: The redundant `CXX="${CXX}"` isn't needed if you update to the latest revision now.

KrainskiL: True, I switched the source version and removed the change with CXX assignment.

PR #6638, opened 2023-04-24
new Madagascar wrapper

mloubout:
Superseed https://github.com/JuliaPackaging/Yggdrasil/pull/6627 with proper branch.


R/RSF/build_tarballs.jl
```diff
@@ -0,0 +1,1887 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir
+
+# install_license
+install_license src/COPYING.txt
+
+# Build
+cd src/
+
+./configure --prefix=${prefix} CC=$(which gcc) CXX=$(which g++) CAIROPATH=${includedir}/cairo FFMPEGPATH=${includedir}/libavcodec/ AR=$(which ar) CPPPATH=${includedir} LIBPATH=${libdir} BLAS=openblas
```
giordano, line 23:
```suggestion
./configure --prefix=${prefix} CC="${CC}" CXX="${CXX}" CAIROPATH=${includedir}/cairo FFMPEGPATH=${includedir}/libavcodec/ AR=$(which ar) CPPPATH=${includedir} LIBPATH=${libdir} BLAS=openblas
```

R/RSF/build_tarballs.jl
```diff
@@ -0,0 +1,1887 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+
+# install_license
+install_license src/COPYING.txt
+
+# Build
+cd src/
+
```
giordano, line 22:
```suggestion
cd $WORKSPACE/srcdir/src

# install_license
install_license COPYING.txt

# Build
```

R/RSF/build_tarballs.jl
```diff
@@ -0,0 +1,1887 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "RSF"
```
giordano, line 5: Why is this called RSF and not Madagascar?

mloubout: The main library is `libdrsf` but can change the name

giordano: https://docs.binarybuilder.org/stable/building/#Name
> It should be the name of the upstream package, not for example that of a specific library or executable provided by it, even though they may coincide.

R/RSF/build_tarballs.jl
mloubout: Those error as well.

PR #6639, opened 2023-04-24
[0_RootFS] Update gcc-11 to gcc-11.3.0

marcom:
This updates gcc-11 to gcc-11.3.0.

### DIFF
--- a/0_RootFS/GCCBootstrap@11/build_tarballs.jl
+++ b/0_RootFS/GCCBootstrap@11/build_tarballs.jl
@@ -1,2 +1,2 @@
 include("../gcc_common.jl")
-build_and_upload_gcc(v"11.1.0")
+build_and_upload_gcc(v"11.3.0")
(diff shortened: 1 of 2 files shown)


imciner2:
Out of curiosity, was there an impetus for bumping to GCC 11.3 from 11.1?

marcom:
I don't remember anymore for which jll package it was, but I had a build failure that looked like it was caused by a bug that was fixed after gcc-11.1, and there was a reason that using gcc-12 was not possible (I think the jll wouldn't have been loadable with julia-1.6? Not quite sure, this is my recollection of talking to Mose around the time I made this PR).

I was later told it's not planned in BinaryBuilder to update a compiler minor version (i.e. going from gcc-11.1 to gcc-11.3) because of binary reproducibility of jll tarballs.  I still think it would be a good idea to be able to update compiler versions to avoid bugs fixed in later minor releases, but I guess for binary reproducibility one would have to record the toolchain used to make the jll somewhere inside the jll tarball.

I think this PR can be closed unless there are plans to allow for compiler version updates.

imciner2:
Yea, we tend to be conservative with compiler updates once it is in Yggdrasil.

PR #6640, opened 2023-04-24
[OpenSpiel] Version Bump, Build for More Platforms

jeremiahpslewis:


O/OpenSpiel/build_tarballs.jl
```diff
@@ -10,58 +10,89 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
...
+cd abseil-cpp
+atomic_patch -p1 ../abseil.patch
+cd ../
+
+macosflags=
+case $target in
+  *apple-darwin*)
+    macosflags="-DCMAKE_CXX_COMPILER_ID=AppleClang"
+    macosflags="$macosflags -DCMAKE_CXX_COMPILER_VERSION=10.0.0"
+    macosflags="$macosflags -DCMAKE_CXX_STANDARD_COMPUTED_DEFAULT=11"
```
giordano, line 48: This looks wrong: the compiler is _not_ Apple Clang, and the variable you should set for the standard is called [`CMAKE_CXX_STANDARD`](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html#variable:CMAKE_CXX_STANDARD).  Also, why are these flags needed at all?  CMake should figure out on its own the compiler name and version, and the standard version should be set by the upstream project, if you have to do that because something it fails then it's an upstream bug.

jeremiahpslewis: I copied this (blindly) from https://github.com/JuliaPackaging/Yggdrasil/blob/c91eb808489a50fa4cc2e3d2dcdadcc3b51c8f9c/L/libcgal_julia/build_tarballs.jl#L36, let me see if it builds without it.

PR #6643, opened 2023-04-25
[libcgal_julia] Try dropping mac flags

jeremiahpslewis:
- Build errors come from MPFR v4.2 not being compatible with current Ygg version of CGAL and v4.1.1 having a build error, need to rebuild once https://github.com/JuliaPackaging/Yggdrasil/pull/6653 is merged


L/libcgal_julia/build_tarballs.jl
```diff
@@ -32,15 +32,6 @@ script = raw"""
 # exit on error
 set -eu
 
-macosflags=
```
giordano, line 35: You can remove also all the above lines at this point.

L/libcgal_julia/build_tarballs.jl
```diff
@@ -25,37 +25,26 @@ sources = [
...
-jlcgaldir = ifelse(isyggdrasil, rname, ".")
 script = raw"""
-## pre-build setup
-# exit on error
-set -eu
+if [[ "${YGGDRASIL}" == "true" ]]; then
+    jlcgaldir="libcgal-julia"
+else
+    jlcgaldir="."
+fi
```
giordano, line 33: There's no predefined `YGGDRASIL` environment variable as far as I know.

PR #6647, opened 2023-04-25
[MPFR] Apply patch

jeremiahpslewis:


jeremiahpslewis:
@giordano Do I need to 'fake' increment the version to ensure the patched version replaces the original one on merge?

M/MPFR/build_tarballs.jl
```diff
@@ -10,6 +10,8 @@ version = v"4.2.0"
 sources = [
     ArchiveSource("https://www.mpfr.org/mpfr-$(version)/mpfr-$(version).tar.xz",
                   "06a378df13501248c1b2db5aa977a2c8126ae849a9d9b7be2546fb4a9c26d993"),
+    FileSource("https://www.mpfr.org/mpfr-current/allpatches",
```
giordano, line 13: Sounds like this isn't versioned, it's likely to break in the future. What's this for?

giordano:
> Do I need to 'fake' increment the version to ensure the patched version replaces the original one on merge?

No, but mpfr is a Julia dependency, you'd need to update it there and wait a few months for next release if you need this patch for something

M/MPFR/build_tarballs.jl
jeremiahpslewis: This is how MPFR ships patches to `4.2`, weird, but implicitly versioned by the checksum...

giordano: > but implicitly versioned by the checksum...

This is backwards. When the file will be replaced in-place the checksum will change and this recipe will be unreproducible. It's a hard no.

jeremiahpslewis: (Fear we might run into licensing issues? Patch is licensed LGPL

PR #6651, opened 2023-04-26
[Foldseek] Fix expand_cxxstring_abis and sync to official version number

marcom:


giordano (commented):
Also the openmp filtering below is wrong

marcom:
Then it's also wrong in this PR for MMseqs2 that I just sent: https://github.com/JuliaPackaging/Yggdrasil/pull/6652

Is this a correct example: https://github.com/JuliaPackaging/Yggdrasil/blob/3cae7c5ad3b0dc7f49d5c30214ff1a2d6c60d11c/A/AMReX/build_tarballs.jl#L75-L82

PR #6653, opened 2023-04-26
[CGAL] Bump version to 5.5.2

jeremiahpslewis:


C/CGAL/build_tarballs.jl
```diff
@@ -44,8 +39,8 @@ products = Product[]
 dependencies = [
     # Essential dependencies
     Dependency("boost_jll"; compat="=1.76.0"),
-    Dependency("GMP_jll"; compat="6.2.0"),
-    Dependency("MPFR_jll"; compat="4.1.1"),
+    Dependency("GMP_jll"; compat="6.2.1"),
+    Dependency("MPFR_jll"; compat="4.2"),
```
giordano, line 43: Julia 1.6 uses 4.1.1: https://github.com/JuliaLang/julia/blob/f9720dc2ebd6cd9e3086365f281e62506444ef37/stdlib/MPFR_jll/Project.toml#L3.  If you require 4.2 (why?) you make the package uninstallable.
```suggestion
    Dependency("MPFR_jll"; compat="4.1.1"),
```

jeremiahpslewis: There is a compile error in 4.1.1 which is fixed in 4.2, without which cgal won't compile (in all cases) https://github.com/CGAL/cgal/issues/7064

giordano: Do I understand correctly that's a problem in 4.1.1 only?  Then using an _older_, rather than _newer_, version would help?

jeremiahpslewis: I believe both options would solve the problem, but don't we have the same issue of Julia using 4.1.1 if we pin to 4.1?

giordano: Uhm, no, how?

PR #6657, opened 2023-04-27
Update mlpack to version 4.1.0.

rcurtin:
With #6656 merged, now I can update mlpack.

(Although: should this wait for a day until the updated `armadillo_jll` package properly hits the repositories?  I'm not sure how long that takes.  I'll check the build logs when they're done to make sure Armadillo 12.2.0 was used...)

### DIFF
--- a/M/mlpack/build_tarballs.jl
+++ b/M/mlpack/build_tarballs.jl
@@ -174,7 +174,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("armadillo_jll"),
+    Dependency("armadillo_jll"; compat="12.2.0"),
     Dependency("OpenBLAS_jll", v"0.3.13"),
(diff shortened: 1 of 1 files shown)


giordano:
I'm on the phone and can't check the logs (they're too long to be displayed in the web interface, you have to download them), was the right version of armadillo used? If not, you have to do a dummy edit anywhere in the recipe (add or remove an empty line) to force a new clean build which will pick up the new version.

giordano:
Actually, judging by https://github.com/JuliaBinaryWrappers/armadillo_jll.jl/commit/a6843d64e9931a965819bb30e06c9832a71163c8, the new version of armadillo changed the soversion, so you'll want to specify the compat bound

rcurtin:
Thanks @giordano, you rock, as always 🎸 
The first build did not use the correct Armadillo version, but I can see from the logs for the updated build that Armadillo 12.2.0 is being used. :+1:

PR #6668, opened 2023-04-30
Add recipe for robin-hood-hashing

serenity4:
Adding this header-only library: https://github.com/martinus/robin-hood-hashing

Added in context of an effort to build https://github.com/KhronosGroup/Vulkan-ValidationLayers

### DIFF
--- a/R/robin-hood-hashing/build_tarballs.jl
+++ b/R/robin-hood-hashing/build_tarballs.jl
@@ -9,9 +9,7 @@ sources = [GitSource("https://github.com/martinus/robin-hood-hashing", "9145f963
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/robin-hood-hashing
-
-mkdir -p "${includedir}"
-mv -v src/include/robin_hood.h "${includedir}/robin_hood.h"
+install -Dv src/include/robin_hood.h "${includedir}/robin_hood.h"
 """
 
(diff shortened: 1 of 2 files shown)


serenity4:
I essentially reused the build script from #5734, but it seems like the pipeline is expecting products to be built. Should I add the source file as `FileProduct`?

R/robin-hood-hashing/build_tarballs.jl
```diff
@@ -0,0 +1,25 @@
...
+version = v"3.11.5"
+
+sources = [GitSource("https://github.com/martinus/robin-hood-hashing", "9145f963d80d6a02f0f96a47758050a89184a3ed")]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/robin-hood-hashing
+
+mkdir -p "${includedir}"
+mv -v src/include/robin_hood.h "${includedir}/robin_hood.h"
```
giordano, line 14:
```suggestion
install -Dv src/include/robin_hood.h "${includedir}/robin_hood.h"
```

giordano (commented):
> but it seems like the pipeline is expecting products to be built

No, the pipeline is expecting the `name` of the recipe to ***exactly*** match the name of the directory, which isn't the case in this PR.  But adding the header file as product is a sensible idea anyway.

serenity4:
Should be all good now, thanks!

PR #6673, opened 2023-05-01
[MUMPS] Version 5.6.0

amontoison:
Let's see if we can compile SCALAPACK32 on Windows before that we merge this PR.


giordano:
#6675 was merged, do you need to do something here?

amontoison:
l also need PARMETIS on Windows but it seems that the registeration failed. I can try to compile MUMPS with Microsoft-MPI after that.
Screenshot_20230503_205307_Samsung Internet

PR #6676, opened 2023-05-01
[Wine] Try version 8.7 from upstream

mkitti:
We switch away from https://github.com/JuliaComputing/wine-staging.git back to https://github.com/wine-mirror/wine . I hope the important patches have been upstreamed.

@Keno , do we still need your patches?

I would like to update this to work with HDF5 cross-compilation on Windows.

### DIFF
--- a/W/Wine/build_tarballs.jl
+++ b/W/Wine/build_tarballs.jl
@@ -13,9 +13,9 @@ platform_spec = [a for a in ARGS if !startswith(a, "-")]
 is_meta = any(a->startswith(a, "--meta-json"), ARGS)
 requested_platforms = map(p->parse(Platform, p; validate_strict=true), platform_spec)
 
-version = v"7.0-rc3"
+version = v"8.7"
 sources = Any[
-    GitSource("https://github.com/JuliaComputing/wine-staging.git", "10ff6d27133070cff6369c48d7ee4c56ee8da7aa"),
+    GitSource("https://github.com/wine-mirror/wine", "c732d2458994d3242741552c47d45d8b04eeb915"),
     DirectorySource("./bundled")
 ]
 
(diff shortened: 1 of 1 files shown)


Keno:
I managed to upstream about half the patches, but I think the patchlist that's still in there represents everything that wasn't upstreamed.

mkitti:
@Keno Werey you actually using this to do something?

W/Wine/build_tarballs.jl
```diff
@@ -125,7 +125,7 @@ end
 
 # Finally, install both:
 script = raw"""
-cd $WORKSPACE/srcdir/wine-staging
+cd $WORKSPACE/srcdir/wine
 atomic_patch -p1 $WORKSPACE/srcdir/patches/hwcap2.patch
 atomic_patch -p1 $WORKSPACE/srcdir/patches/darwin.patch
```
mkitti, line 130:
```suggestion
# atomic_patch -p1 $WORKSPACE/srcdir/patches/hwcap2.patch
# atomic_patch -p1 $WORKSPACE/srcdir/patches/darwin.patch
```

W/Wine/build_tarballs.jl
```diff
@@ -95,18 +95,18 @@ if !is_meta
 
         # Next, build wine32, without wine64, then use those tools to build wine32 WITH wine64
         wine32_script = raw"""
-        cd $WORKSPACE/srcdir/wine-staging
+        cd $WORKSPACE/srcdir/wine
         atomic_patch -p1 $WORKSPACE/srcdir/patches/hwcap2.patch
```
mkitti, line 99:
```suggestion
        # atomic_patch -p1 $WORKSPACE/srcdir/patches/hwcap2.patch
```

mkitti (commented):
Commenting out the patches for now

Keno:
> @Keno Were you actually using this to do something?

Cross compiles for julia itself.

mkitti:
I see. So do we still need the old build then? Wine seems to be barely working for me in binary builder.

Keno:
I mean, the old build will stick around. Feel free to do whatever.

Keno:
Superseded by #13173

PR #6688, opened 2023-05-03
[libsigc++] Bump to v3.4.0 and add v2.12.0

topolarity:
libsigc++-2.0 and libsigc++-3.0 are "parallel-installable" ABI's used with GTK3 and GTK4, respectively.

Rather unintuitively, the entire `*mm` stack is split on this ABI even for libraries (e.g. Pango) whose C bindings have no such breaking ABI change. This means that the minor version of every GTK-upstream library must be constrained to respect the `libsigc++` ABI bifurcation.

### DIFF
--- a/L/libsigcpp/libsigcpp@2.12.0/build_tarballs.jl
+++ b/L/libsigcpp/libsigcpp@2.12.0/build_tarballs.jl
@@ -1,0 +1,38 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "libsigcpp"
+version = v"2.12.0"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://download.gnome.org/sources/libsigc++/2.12/libsigc%2B%2B-$(version).tar.xz",
+                  "1c466d2e64b34f9b118976eb21b138c37ed124d0f61497df2a90ce6c3d9fa3b5")
+]
+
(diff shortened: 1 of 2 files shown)


topolarity:
Looks like I probably can't do two version bumps at once like this.

I guess I need to split the PR?

giordano:
> Looks like I probably can't do two version bumps at once like this.

So, this works ok on the Yggdrasil side, but will result in two necessarily-conflicting PRs in the General registry (because each will be adding a new different version).  I'm going to merge this, but will have to fix one of the PRs to the registry.

L/libsigcpp/libsigcpp@2.12.0/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+mkdir meson
+cd meson
+meson --cross-file=${MESON_TARGET_TOOLCHAIN%.*}_gcc.meson
+ninja -j${nproc}
+ninja install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = expand_cxxstring_abis(supported_platforms(; experimental=true))
```
giordano, line 26: Argh, I just realised that this should have expanded C++ string ABI also for BSD systems.  Why do you need GCC?

topolarity: Build fails with `clang++`, it looks like they've only taken fixes for it on the newer libsigc++-3.0 branches (and the fix does not directly backport)

PR #6691, opened 2023-05-03
[libsigc++] Expand C++ string ABI for all systems

topolarity:
See https://github.com/JuliaPackaging/Yggdrasil/pull/6688#discussion_r1183640855

Unfortunately, it looks like several of the libraries in the GTK4mm dependency chain also do not build with `clang++` so we'll probably have to switch to GCC and fully-expand string ABI's for `libsigc++-3.0` too

### DIFF
--- a/L/libsigcpp/libsigcpp@2.12.0/build_tarballs.jl
+++ b/L/libsigcpp/libsigcpp@2.12.0/build_tarballs.jl
@@ -23,7 +23,7 @@ ninja install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(supported_platforms(; experimental=true); skip=Returns(false))
+platforms = expand_cxxstring_abis(supported_platforms(); skip=Returns(false))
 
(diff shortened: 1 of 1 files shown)


L/libsigcpp/libsigcpp@2.12.0/build_tarballs.jl
```diff
@@ -23,7 +23,7 @@ ninja install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(supported_platforms(; experimental=true))
+platforms = expand_cxxstring_abis(supported_platforms(; experimental=true); skip=Returns(false))
```
giordano, line 26: Min nit-pick: `experimental` does nothing nowadays, it's kept only for compatibility
```suggestion
platforms = expand_cxxstring_abis(supported_platforms(); skip=Returns(false))
```

topolarity: Good to know, thanks :+1:

PR #6693, opened 2023-05-03
[Cairomm] Bump to v1.16.2

topolarity:
This is the latest version with libsigc++-3.0, for use with GTK4.


topolarity:
Remaining failures are all due to not locating the cairo provided in the dependency artifacts.

Investigating now

PR #6694, opened 2023-05-03
[Cairomm] Add v1.14.4

topolarity:
This is the latest version with libsigc++-2.0, for use with GTK3.

### DIFF
--- a/C/Cairomm/Cairomm@1.14.4/build_tarballs.jl
+++ b/C/Cairomm/Cairomm@1.14.4/build_tarballs.jl
@@ -1,0 +1,46 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Cairomm"
+version = v"1.14.4"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://www.cairographics.org/releases/cairomm-$(version).tar.xz",
+                  "4749d25a2b2ef67cc0c014caaf5c87fa46792fc4b3ede186fb0fc932d2055158")
+    DirectorySource("./bundled")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/cairomm*/
+atomic_patch -p1 ../patches/disable-quartz-aarch64-apple-darwin.patch
+mkdir output && cd output/
+meson --cross-file=${MESON_TARGET_TOOLCHAIN%.*}_gcc.meson ..
+ninja -j${nproc}
+ninja install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+
+# armv6l is not supported by Cairo_jll yet
+platforms = filter!(p -> arch(p) != "armv6l", supported_platforms())
+
+platforms = expand_cxxstring_abis(platforms; skip=Returns(false))
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct(["libcairomm-1", "libcairomm-1.0"], :libcairomm)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="Cairo_jll", uuid="83423d85-b0ee-5818-9007-b63ccbeb887a"), compat="1.16.1")
+    Dependency(PackageSpec(name="libsigcpp_jll", uuid="8855df87-3ca1-5a3d-a68b-4f0ddc198ba8"), v"2.12.0"; compat="2")
+    BuildDependency(PackageSpec(name="Xorg_xorgproto_jll", uuid="c4d99508-4286-5418-9131-c86396af500b"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"7.1.0")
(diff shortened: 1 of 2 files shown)


C/Cairomm/Cairomm@1.14.4/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+ninja install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+
+# These platforms are not supported by Cairo_jll yet
+platforms = filter!(p -> arch(p) != "armv6l", supported_platforms())
+platforms = filter!(p -> !(Sys.isapple(p) && arch(p) == "aarch64"), platforms)
+platforms = filter!(p -> libc(p) != "musl", platforms)
```
giordano, line 29: Uhm, only armv6l isn't there (and we mainly just make sure to rebuild all dependencies, nothing fundamental): https://github.com/JuliaPackaging/Yggdrasil/blob/fba5fe9f5c4e78e2dc0c25f479c65ce75111049a/C/Cairo/build_tarballs.jl#L67

giordano: Yeah, finding which dependencies of Cairo are missing armv6l builds.  I think it's mostly the X11 stack, we never touched that

topolarity: Is it alright if I avoid enabling armv6l right now, and just trigger the rebuild for musl/aarch64-darwin?

I'm still trying to make all the way up to `GTKmm3` on the existing platforms, and I've got a ways to go..

giordano: I'm ok with forgetting about armv6l, but Cairo_jll should be already available for all the other platforms: https://github.com/JuliaBinaryWrappers/Cairo_jll.jl/blob/5105339972cc9aad07d9ef2760a24e61ebeb4674/README.md#platforms

C/Cairomm/Cairomm@1.14.4/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+platforms = expand_cxxstring_abis(platforms; skip=Returns(false))
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct(["libcairomm-1", "libcairomm-1.0"], :libcairomm)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="Cairo_jll", uuid="83423d85-b0ee-5818-9007-b63ccbeb887a"), v"1.14.12"; compat="^1.12.0")
```
giordano, line 40: #3024
```suggestion
    Dependency(PackageSpec(name="Cairo_jll", uuid="83423d85-b0ee-5818-9007-b63ccbeb887a"); compat="1.16.1")
```

giordano: Side note, I think it's usually fishy to use both a build version and a compat spec:
```julia
Dependency("Packagename_jll", build_version; compat="<compat_spec>")
```
If you care about compatibility you should usually use for building the oldest version you claim compaitiblity with, and conversely if you use a specific version for building in general you can't expect to be actually compatible with an older version, especially on macOS where the compatibility version can create problems.  All in all I prefer to use the compat spec only, the oldest version corresponding to that range will be then automatically used for building.

topolarity: That makes sense.

I've been specifying compatibility this way to have things in terms of the full upstream version span (in particular, BinaryBuilder will error out if the bottom of your version span hasn't actually been released and registered), but I'm happy to treat this as a compat on the released BinaryBuilder artifacts and pull in the lower bound

topolarity:
I think your suggestion above fixes most of my platform problems, but now I'm facing GCC-related troubles. The build fails upon including some Objective-C header files from the macOS SDK, but succeeds if I use `clang++`.

Which seems to leave me with these options:
1. Try to wrestle with GCC and get it to support the ImageIO header file correctly
2. Wrestle with `libsigc++=2.0` (and a few other dependencies) to get them to support clang++
3. Give up and don't support Cairomm on macOS 😢

giordano:
I think using clang would be a better option if possible, because interaction between libraries linking to libstdc++ and libc++ is...unexplored and potentially very interesting :grimacing:  If you want to stick to GCC, I don't know if this is a problem that could be solved by using a more recent version, but please avoid going past GCC 10 for C++ code.

giordano:
> It looks like the problem is that Apple never merged its blocks patches to open source GCC, and the newer macOS SDK that we use on aarch64-apple-darwin is missing `__BLOCKS__` guards in many places (x86-64 has the appropriate guards and works fine).

That's interesting, I'd expect older SDKs to be problematic, not newer ones.... :grimacing: 

> Sure thing. For my own education, can I ask why?

Because Julia v1.6 comes with libstdc++ from GCC 10, so libraries requiring [`GLIBCXX_3.4.29` or above](https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html) would not work with the current LTS of Julia.

topolarity:
One of the remaining errors seems to be caused by:
```
┌ Warning: Dependency Cairo_jll does not have a mapping for artifact Cairo for platform i686-linux-musl-libgfortran4-cxx11
└ @ BinaryBuilderBase /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilderBase/ilpz3/src/Prefix.jl:752
```

Is this one of those cases where the dependency-matching logic doesn't know that the Cairo artifacts should match to any cxx11/libgfortran ABI, and so the `i686-linux-musl` artifact doesn't match?

staticfloat:
I think it was just a transient download issue; I restarted the build and it worked.

PR #6701, opened 2023-05-03
[Python] Support internal ssl package

omus:
When we build our Python_jll with OpenSSL support we can see the following message while running `make`:

```
*** WARNING: renaming "_ssl" since importing it failed: Error relocating /workspace/srcdir/Python-3.10.8/build_target/build/lib.linux-x86_64-3.10/_ssl.cpython-310-x86_64-linux-gnu.so: COMP_get_type: symbol not found

The necessary bits to build these optional modules were not found:
_curses              _curses_panel        _dbm
_gdbm                _tkinter             _uuid
nis                  readline
To find the necessary bits, look in setup.py in detect_modules() for the module's name.


The following modules found by detect_modules() in setup.py, have been
built by the Makefile instead, as configured by the Setup files:
_abc                 pwd                  time


Following modules built successfully but were removed because they could not be imported:
_ssl


Could not build the ssl module!
Python requires a OpenSSL 1.1.1 or newer
Custom linker flags may require --with-openssl-rpath=auto
```

Removing the `_ssl` module is problematic as it results in this build of Python not being able to use the `ssl` module. I tried `--with-openssl-rpath=auto` as suggested and a few other options which didn't address this problem. In investigating the removal message I found that the `setup.py` has an option to skip loading of extensions during cross compilation builds:

```python
        # Don't try to load extensions for cross builds
        if CROSS_COMPILING:
            return
```

To enable this flag we need to define the [undocumented environmental variable `_PYTHON_HOST_PLATFORM`](https://discuss.python.org/t/pep-425-platform-tag/5157/3). As this environmental variable is also used in `get_platform` within the `setup.py` which typically defaults to using [`sys.platform`](https://docs.python.org/3/library/sys.html#sys.platform) it seems like the appropriate value to set this to is just "linux".

### DIFF
--- a/P/Python/build_tarballs.jl
+++ b/P/Python/build_tarballs.jl
@@ -22,7 +22,7 @@ rm -f $(which python3)
 
 # We need these for the host python build
 apk update
-apk add zlib-dev libffi-dev
+apk add zlib-dev libffi-dev openssl-dev
 
 # Create fake `arch` command:
 echo '#!/bin/bash' >> /usr/bin/arch
@@ -66,6 +66,9 @@ export CPPFLAGS="${CPPFLAGS} -I${prefix}/include"
 export LDFLAGS="${LDFLAGS} -L${prefix}/lib -L${prefix}/lib64"
 export PATH=$(echo ${WORKSPACE}/srcdir/Python-*/build_host):$PATH
 
+# Force Python build to run in cross compilation mode
+export _PYTHON_HOST_PLATFORM="linux"
+
 conf_args=()
 conf_args+=(--enable-shared)
 conf_args+=(--disable-ipv6)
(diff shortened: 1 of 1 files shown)


omus:
I was encountering this particular issue when using the Python_jll as a `HostBuildDependency`. This doesn't seem to be a problem when using the Python_jll in a normal Julia package.

omus:
I managed to fix the `COMP_get_type: symbol not found` warning which allows `import ssl` to work. The hacky fix used `ln -s $prefix/lib/libssl.* $prefix/lib/libcrypto.* /usr/local/lib/` which means we just need to get Python to use our library linking changes.

omus:
I'm not planning on pursuing this so I'll just close this.

PR #6703, opened 2023-05-04
[Pangomm] Add v2.46.3

topolarity:
This is the latest version with libsigc++-2.0, for use with GTK3.


topolarity:
Sigh. Failures are once again due to:
```
┌ Warning: Dependency Cairo_jll does not have a mapping for artifact Cairo for platform x86_64-unknown-freebsd12.2-libgfortran4-cxx11
└ @ BinaryBuilderBase
```
```
[18:56:43] pango| Run-time dependency cairo found: NO (tried pkgconfig and cmake)
```

I think something important might be wrong here.

topolarity:
cc @staticfloat

staticfloat:
It always seems to be `amdci7.10` I wonder if there's something bugged in its cache.  I'll wipe it.

PR #6706, opened 2023-05-05
[Vulkan_Loader] Add v1.3.243

topolarity:


topolarity:
The version bump means that we're now hitting https://github.com/godotengine/godot/issues/44637 which apparently requires MinGW >= 8 to be fixed

The alternative is to downgrade to v1.3.239, which is the most recent version with an official release

topolarity:
Also the Windows builds here are missing the `.lib` files.

I need to use the full Vulkan SDK (rather than just the RT redistributable) to get those, but that's unfortunately only released as a `.exe`. It does look like the installer downloads a `.7z` for its main payload, so maybe I can try to sniff out that URL and use that instead.

PR #6712, opened 2023-05-05
Bump the version number of PARMETIS_jll

amontoison:
Update: Mosè already fixed the problem (https://github.com/JuliaRegistries/General/pull/82872/files)


giordano:
> Recompile PARMETIS without a compat entry for MPITrampoline 

Compat bounds are there for a reason, removing them is rarely is a sensible idea.

PR #6714, opened 2023-05-05
Wizard recipe: LightGBM-v3.3.5

characat0:

* Package name: LightGBM
* Version: v3.3.5

### DIFF
--- a/L/LightGBM/build_tarballs.jl
+++ b/L/LightGBM/build_tarballs.jl
@@ -8,19 +8,15 @@ version = v"3.3.5"
 # Collection of sources required to complete build
 sources = [
     GitSource("https://github.com/microsoft/LightGBM.git", "ca035b2ee0c2be85832435917b1e0c8301d2e0e0"),
-    GitSource("https://github.com/boostorg/compute.git", "36c89134d4013b2e5e45bc55656a18bd6141995a"),
-    GitSource("https://gitlab.com/libeigen/eigen.git", "8ba1b0f41a7950dc3e1d4ed75859e36c73311235"),
-    GitSource("https://github.com/lemire/fast_double_parser.git", "ace60646c02dc54c57f19d644e49a61e7e7758ec"),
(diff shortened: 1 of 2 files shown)


L/LightGBM/bundled/patches/windows/Ws2_32-iphlpapi.patch
giordano, line 1: This patch should be upstreamed

L/LightGBM/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("lib_lightgbm", :lib_lightgbm),
+    ExecutableProduct("lightgbm", :lightgbm)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"))
```
giordano, line 58: https://github.com/JuliaPackaging/Yggdrasil/blob/06a50b638edb4055c569dde2fc425d0c998bacb6/B/blis/build_tarballs.jl#L135-L138

giordano (commented):
Thanks for your contribution!

L/LightGBM/bundled/patches/windows/Ws2_32-iphlpapi.patch
characat0: I read that windows uses uppercase for those libraries, while linux doesn't. So this patch would not apply for the configuration of the main project. 
Anyways, I could be terribly mistaken since this is the first time I use BinaryBuilder and cmake at all. If that's the case, I will try to get the patch into the main project.

giordano: Windows systems use case-insensitive file systems, so they literally don't care in the slightest about how those paths are written out. So this patch is very much well suited for being upstreamed.

PR #6715, opened 2023-05-06
HYPRE: Rebuild with new blastrampoline

fredrikekre:
Try to fix https://github.com/fredrikekre/HYPRE.jl/issues/11


giordano:
The problem is in lapack, not this package

fredrikekre:
Ok, could you trigger the relevant builds to fix the HYPRE problem? That would be nice! 🙂

PR #6716, opened 2023-05-06
[LAPACK] Upgrade to v3.10.1 and require newer libblastrampoline

giordano:
Libblastrampoline v5.4.0 is compatible only with Julia v1.9+, but it has a stable ABI on all platforms, including Windows.

CC: @fredrikekre.

### DIFF
--- a/L/LAPACK/LAPACK/build_tarballs.jl
+++ b/L/LAPACK/LAPACK/build_tarballs.jl
@@ -8,4 +8,4 @@ script = lapack_script(lapack32=false)
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
                julia_compat="1.9", preferred_gcc_version=v"6")
 
-# Build trigger: 1
+# Build trigger: 2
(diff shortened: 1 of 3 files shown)


ViralBShah:
How about the lapack 3.11.0 PR that is open?

giordano:
I wanted to make a smaller jump first.

ViralBShah (approved):

giordano:
Compilation for aarch64-linux-*-libgfortran4 is crashing (looks like an ICE on both glibc and musl), but we don't have many other compilers for those versions of libgfortran (to be clear, libgfortran4 corresponds _only_ to GCC 7), so this is quite problematic.  Sigh.

amontoison:
Don't we need LBT v"5.7.0" for LAPACK_jll?
https://github.com/JuliaLinearAlgebra/libblastrampoline/pull/114

ViralBShah:
I believe 5.4 is used in Julia 1.9, although there is a PR to backport 5.7.

amontoison:
I don't think that 5.7 can be backported because it was compiled with a Julia compat `1.10`.

PR #6719, opened 2023-05-07
LAPACK 3.11.0

ViralBShah:
Continuation from #6553

### DIFF
--- a/L/LAPACK/common.jl
+++ b/L/LAPACK/common.jl
@@ -1,13 +1,13 @@
-using BinaryBuilder
+using BinaryBuilder, Pkg
 
 # LAPACK mirrors the OpenBLAS build, whereas LAPACK32 mirrors the OpenBLAS32 build.
 
-version = v"3.10.1"
+version = v"3.11.0"
 
 # Collection of sources required to build lapack
 sources = [
-    GitSource("https://github.com/Reference-LAPACK/lapack.git",
-              "32b062a33352e05771dcc01b981ebe961bf2e42f"),
+    GitSource("https://github.com/Reference-LAPACK/lapack",
+              "7866626840f5d5e7e27f027a55182da8b3303872"),
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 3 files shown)


ViralBShah:
Can we use Julia 1.9 as the conpat version here and hence LBT 5.4?

ViralBShah:
1.9.1 is going to have LBT 5.8. https://github.com/JuliaLang/julia/pull/49680

amontoison:
> 1.9.1 is going to have LBT 5.8. I suppose the only lever of control we have here is which version of Julia to use. [JuliaLang/julia#49680](https://github.com/JuliaLang/julia/pull/49680)
> 
> I feel that Julia 1.9 is the right compat to define.

Oh, I suppose that you can use LBT 5.7.0 in Julia 1.9 because you compile it from sources when you build Julia.
We are unable to compile an artifact here with LBT 5.7.0 directly because the compat entry of LBT_jll  is 1.10 but that's fine to link with LBT 5.4.0 because it's compatible with versions of LBT in `[5.4.0, 6.0.0[`.

L/LAPACK/common.jl
```diff
@@ -306,32 +304,50 @@ function lapack_script(;lapack32::Bool=false)
...
     """
 end
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = expand_gfortran_versions(supported_platforms())
+
 # Building ILP64 LAPACK on aarch64 linux runs into internal compiler errors with
 # GCC ≤ 7 (=> libgfortran ≤ 4).
 filter!(p -> !(arch(p) == "aarch64" && Sys.islinux(p) && libgfortran_version(p) ≤ v"4"), platforms)
```
amontoison, line 347: When I recompiled SCALAPACK32 I found that GCC 7 has a bug on some platforms if `-cpp` is used without preprocessing macros.
@ViralBShah 
Can you add a condition to use
```FFLAGS=(-cpp -ffixed-line-length-none -DUSE_ISNAN)``` only for `LAPACK_jll` and not for `LAPACK32_jll`?

ViralBShah: I am keeping `-ffixed-line-length-none` for all, but moved `-cpp -DUSE_ISNAN` for LAPACK_jll only.

ViralBShah: Didn't help with building the failing platforms.

PR #6724, opened 2023-05-09
[libjulia] rebuild 1.10 for changes on master, update 1.9 to release

benlorenz:
Since https://github.com/JuliaLang/julia/pull/49556 several packages depending on libjulia_jll failed on nightly (CxxWrap, Polymake, GAP, maybe more?), for the rebuilds we first need a new libjulia_jll.

- drop merged patch for 1.10
- rebuild 1.9 with release version instead of rc3

~~I commented the builds for 1.6-1.8 since I don't think we need to rebuild those. To make sure this doesn't break the old versions I changed the compat to 1.9 and did a version bump. I think it should work this way?~~

~~If we need to rebuild older julia releases we can bump the version again to lower the compat.~~

cc: @fingolfin

PS: I tried this locally and rebuilding the full chain of `libjulia_jll`, `libcxxwrap_julia_jll` and `libpolymake_julia_jll` did fix the Polymake.jl tests.


L/libjulia/build_tarballs.jl
```diff
@@ -31,9 +31,12 @@ function Pkg.Types.is_stdlib(uuid::Base.UUID, julia_version::VersionNumber)
...
 end
 
-jllversion=v"1.10.1"
-build_julia(ARGS, v"1.6.3"; jllversion)
-build_julia(ARGS, v"1.7.0"; jllversion)
-build_julia(ARGS, v"1.8.2"; jllversion)
-build_julia(ARGS, v"1.9.0-rc3"; jllversion)
+# WARNING: the current version was built with julia_compat=1.9
+# increase the version when lowering the compat again to build for older julia versions
+# (julia_compat is set in common.jl)
```
fingolfin, line 36: Will this work, though? Isn't Yggdrasil itself still using Julia 1.7, so will it even be able to use the libjulia_jll updates produced by this PR?

I guess the goal is to avoid rebuilding older versions Julia of the JLL? Of course in general it is good if we don't get new artifacts, though in this particular case no use installs `libjulia_jll`, only Yggdrasil needs it, so perhaps we can get away with just rebuilding it all and that's it...

(I guess being able to reuse artifacts of older builds of a Julia still is in on the wishlist, @giordano ?)

fingolfin (approved):

PR #6725, opened 2023-05-09
[RDKit] 2023.03.1

eloyfelix:


fingolfin:
Oops, this PR is from May 2023 and nobody even commented on it since then !!! Sorry for that, it simply has been overlooked by us collectively!

On the other hand, also nobody complained about it being ignored -- so I am going to close this for now as it is unlikely to work without changes. However, if you still want this, just submit a fresh PR (so that it is at the top of the list), it can be even from the same branch.

PR #6728, opened 2023-05-09
[LightGBM] add support for julia >= 1.0 

characat0:
This PR aims to provide support for Julia 1.0, but I'm not sure if this is the correct approach.

### DIFF
--- a/L/LightGBM/build_tarballs.jl
+++ b/L/LightGBM/build_tarballs.jl
@@ -47,6 +47,9 @@ install_license LICENSE
 # platforms are passed in on the command line
 platforms = expand_cxxstring_abis(supported_platforms())
 
+# Add compatibity with older Julia versions
+filter!(p -> !(arch(p) == "armv6l" || (Sys.isapple(p) && arch(p) == "aarch64")), platforms)
+
 # The products that we will ensure are always built
 products = [
     LibraryProduct("lib_lightgbm", :lib_lightgbm),
@@ -62,4 +65,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"7.1.0")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.0", preferred_gcc_version = v"8")
--- a/L/LightGBM/build_tarballs.jl
+++ b/L/LightGBM/build_tarballs.jl
@@ -48,7 +48,11 @@ install_license LICENSE
 platforms = expand_cxxstring_abis(supported_platforms())
 
 # Add compatibity with older Julia versions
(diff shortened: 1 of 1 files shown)


L/LightGBM/build_tarballs.jl
```diff
@@ -62,4 +69,6 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"7.1.0")
+for (julia_compat, platforms) in platform_compat
+    build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat=julia_compat, preferred_gcc_version = v"8")
```
fingolfin, line 73: Hmm, I am not sure how this could work, because in the end there is a single `Project.toml`, and it can only specify a single Julia compat bound.

Older Julia versions get confused by the "experimental" platforms they don't know. IIRC this was addressed in Julia 1.6 (or thereabout, so that we don't run into this problem again in the future. But for now, older Julia versions don't like "unknown" platforms.

Hmm, also, Artifacts (which are a key aspect of JLLs) did not exist in Julia 1.0. IIRC (and I may not -- it's only been like 2 years, but it feels like eternity, and it seems I forgot quite some details) to use JLL in versions before Julia 1.3 (?) is non-trivial, you need to essentially manually wrangle the artifact data (?).

Overall, I am not sure it's really possible to do what you want to do here...

May I ask what the motivation is? What's stopping you from just using Julia 1.6 or newer?

characat0: Thank you for the detailed response, I made this JLL for the [LighGBM.jl](https://github.com/IQVIA-ML/LightGBM.jl) repo, and they aim to be compatible with Julia 1.0 and newer (based on their test suite).

from the README.md:
### BinaryProvider.jl 
  
 We hope we convinced you about why it is important to switch to JLL packages.  However, if you really need to support Julia v1.2 or previous versions, you should keep using [BinaryProvider.jl](https://github.com/JuliaPackaging/BinaryProvider.jl).  As explained in the previous section, a `build.jl` file is not generated after the build anymore as that is not used for JLL packages, instead you would need to generate it yourself: In the top-directory of Yggdrasil there is a script to do this: [`generate_buildjl.jl`](./generate_buildjl.jl).  It takes three positional arguments: 
  
 * the path to the `build_tarballs.jl` script 
 * the `owner/name` of the repository where the tarballs have been uploaded.  If omitted, this defaults to `JuliaBinaryWrappers/BuilderName_jll.jl` 
 * the tag name where the tarballs have been uploaded.  If omitted, this defaults to the latest version of the JLL package in the [General registry](https://github.com/JuliaRegistries/General).  If there are no versions of the package in the registry, the script will fail. 
---

Maybe I should try this way?

giordano: BinaryProvider is totally unmaintained and it'll just make your package unusable on newer architectures like aarch64-apple-darwin which were invented after BinaryProvider was developed. You really do ***not** want to use it nowadays. I don't see a single reason for you to spend time on supporting a 5-year old version of Julia.

imciner2:

PR #6748, opened 2023-05-16
T8code without depending on p4est_jll.jl

jmark:
I removed the build dependency `p4est_jll.jl` and turned back to the default build process. For some reason, building against external p4est library causes errors on Windows platforms.


sloede:
Have you checked whether it is still possible to use t8code_jll and p4est_jll side by side, i.e., load both libraries and still use both libraries in the same Julia process?

jmark:
@sloede Yes, this works without problems.

jmark:
All tests for `T8code.jl` with my new changes for `t8code_jll.jl` pushed to a private repo work on Windows and Linux on my local machine. I need this package via official channels for future work.

@giordano Please review and merge.

PR #6750, opened 2023-05-16
GLEW: Use correct installation convention for MinGW targets

topolarity:
The existing artifacts have `libglew32.a` in bin and an empty lib directory (except for pkgconfig).

This commit overrides the auto-detection script so that MinGW is correctly configured as the target system, which means that `libglew32.a` and `libglew32.dll.a` are installed into `lib` as expected.

### DIFF
--- a/G/GLEW/build_tarballs.jl
+++ b/G/GLEW/build_tarballs.jl
@@ -18,7 +18,11 @@ EXTRA_VARS=()
 if [[ "${target}" == *-linux-* ]] || [[ "${target}" == *-freebsd* ]]; then
     # On Linux and FreeBSD this variable by default does `-L/usr/lib`
     EXTRA_VARS+=(LDFLAGS.EXTRA="")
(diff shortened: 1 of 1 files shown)


G/GLEW/build_tarballs.jl
```diff
@@ -42,4 +46,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
giordano, line 49: You need to change the version number to do this: #2763. If you can't change the version number, don't build for aarch64-apple-darwin

topolarity: Thanks for the heads up! I've bumped to v2.2.0

PR #6752, opened 2023-05-16
[GR] Add GRQt5 package

barche:
Draft PR to support proposed changes in GR


barche:
I think this is OK now, @jheinen 
@giordano Any objections to splitting the GR JLL into a Qt5 and a Core part?

jheinen:
@barche : Thanks for the MR.

Unfortunately I had to make a bugfix release "in between" (0.72.7) and can probably deal with your MR only next week.

barche:
OK, I updated this one and PR #6751 as well.

PR #6758, opened 2023-05-17
[GR] bump version

jheinen:
bump GR version to 0.72.5

### DIFF
--- a/G/GR/build_tarballs.jl
+++ b/G/GR/build_tarballs.jl
@@ -3,13 +3,13 @@
 using BinaryBuilder
 
 name = "GR"
-version = v"0.72.4"
+version = v"0.72.5"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/sciapp/gr.git", "34385d25b9dca50b1a775d4ac1e669732b6bf5e1"),
+    GitSource("https://github.com/sciapp/gr.git", "0082c330d02fa5597acdf5c8145f1ea1424588f2"),
     FileSource("https://github.com/sciapp/gr/releases/download/v$version/gr-$version.js",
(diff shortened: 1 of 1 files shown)


jheinen:
The `x86-64-apple-darwin` build fails because `macosx-version-min` is set to `10.10` (see compiler flags). How can I set them to `10.14`?

giordano:
Try setting
```sh
export MACOSX_DEPLOYMENT_TARGET=10.14
```
for `x86_64-apple-darwin`, if that's not enough because you actually need a 10.14 sdk then you have to use that as a source
https://github.com/JuliaPackaging/Yggdrasil/blob/806bbbf2cc4882aa3dc3de98fc2924953b1e72f2/O/OpenSpiel/build_tarballs.jl#L23
https://github.com/JuliaPackaging/Yggdrasil/blob/806bbbf2cc4882aa3dc3de98fc2924953b1e72f2/O/OpenSpiel/build_tarballs.jl#L28-L36

jheinen:
@giordano : Thanks for your help!

G/GR/build_tarballs.jl
```diff
@@ -31,7 +33,17 @@ else
...
+    pushd $WORKSPACE/srcdir/MacOSX10.*.sdk
+    rm -rf /opt/${target}/${target}/sys-root/System
+    cp -ra usr/* "/opt/${target}/${target}/sys-root/usr/."
+    cp -ra System "/opt/${target}/${target}/sys-root/."
+    export MACOSX_DEPLOYMENT_TARGET=10.14
+    popd
+fi
+
 if [[ "${target}" == *apple* ]]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.14
```
giordano, line 46: This is done already 5 lines above and is necessary only on x86_64 anyway 
```suggestion
```

PR #6760, opened 2023-05-18
[SDPA] rebuild with no changes for latest libcxxwrap

odow:
x-ref https://github.com/jump-dev/SDPA.jl/issues/56
x-ref https://github.com/jump-dev/SDPA.jl/pull/57

OpenSpiel is similar:

x-ref https://github.com/JuliaPackaging/Yggdrasil/pull/6759

If we're going to rebuild libcxxwrap, https://github.com/JuliaPackaging/Yggdrasil/pull/6726, does it make sense to also rebuild all dependent jll at the same time (or after, once version registered)? I can see this being a re-occuring pain point.

### DIFF
--- a/S/SDPA/build_tarballs.jl
+++ b/S/SDPA/build_tarballs.jl
@@ -1,5 +1,6 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
+
 using BinaryBuilder, Pkg
 
 # See https://github.com/JuliaLang/Pkg.jl/issues/2942


odow:
With this latest version, I'm seeing failures on Windows and Julia v1.9

image

https://github.com/jump-dev/SDPA.jl/actions/runs/5017955895/jobs/8997264217?pr=57
https://github.com/jump-dev/SDPA.jl/pull/57

Things are working on 1.6 and 1.10, so perhaps the uploaded binary just got corrupted? Any ideas? Do we just try rebuilding?

giordano:
We've seen this error randomly popping out, but no one understands it

odow:
:cry: so it sufficient to just rebuild?

giordano:
Unlikely

PR #6766, opened 2023-05-19
make MKL_jll a BuildDependency only for SCS_MKL

kalmarek:
so that MKL_jll is a truly optional dependency for SCS ;)

@giordano does that require version bump, or is +1 fine?

cc: @odow

### DIFF
--- a/S/SCS_MKL/build_tarballs.jl
+++ b/S/SCS_MKL/build_tarballs.jl
@@ -1,3 +1,4 @@
+using Pkg
 using BinaryBuilder
 
 name = "SCS_MKL"
@@ -37,7 +38,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("MKL_jll"; compat="2022.2.0"),
+    BuildDependency(PackageSpec(;name="MKL_jll", version=v"2022.2")),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well
--- a/S/SCS_MKL/build_tarballs.jl
+++ b/S/SCS_MKL/build_tarballs.jl
@@ -38,7 +38,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    BuildDependency(PackageSpec(;name="MKL_jll", version=v"2022.2")),
+    BuildDependency(PackageSpec(;name="MKL_jll", version=v"2023.1")),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well


giordano:
I don't understand this: if this package has a package which requires MKL (I don't know), how's this optional?

kalmarek:
well if I do
```
JULIA_DEPOT_PATH=/tmp/julia_tmp_depot julia --project=.
```
and then `Pkg.instantiate()` in `SCS` folder now I see
```
dust /tmp/julia_tmp_depot 
 33M         ┌── 833b9_8gJVM.so                      │                                                        ▒▒▒▒▒▒██ │   2%
 33M       ┌─┴ Parsers                               │                                                        ▒▒▒▒▒▒██ │   2%
 95M       │ ┌── tyub8_8gJVM.so                      │                                                        ▒▒▒▓████ │   6%
 98M       ├─┴ MathOptInterface                      │                                                        ▒▒▒█████ │   6%
170M     ┌─┴ v1.9                                    │                                                        ████████ │  11%
170M   ┌─┴ compiled                                  │                                                        ████████ │  11%
 30M   │     ┌── julia                               │                                           ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒██ │   2%
 43M   │   ┌─┴ share                                 │                                           ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒██ │   3%
 31M   │   │   ┌── libopenblas64_.0.3.21.so          │                                           ▒▒▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓██ │   2%
 55M   │   │   ├── libjulia-codegen.so.1.9           │                                           ▒▒▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓███ │   4%
 83M   │   │   ├── libLLVM-14jl.so                   │                                           ▒▒▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓████ │   5%
217M   │   │   ├── sys.so                            │                                           ▒▒▓▓▓▓▓▓▓▓▓▓█████████ │  14%
444M   │   │ ┌─┴ julia                               │                                           ▒▒███████████████████ │  29%
444M   │   ├─┴ lib                                   │                                           ▒▒███████████████████ │  29%
497M   │ ┌─┴ julia-1.9.0+0.x64.linux.gnu             │                                           █████████████████████ │  32%
497M   ├─┴ juliaup                                   │                                           █████████████████████ │  32%
 30M   │     ┌── libopenblas.0.3.17.so               │                            ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░██ │   2%
 30M   │   ┌─┴ lib                                   │                            ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░██ │   2%
 31M   │ ┌─┴ 7fca6a59f5903d74ab82fd8c36b7656125da4318│                            ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░██ │   2%
 67M   │ │ ┌── lib                                   │                            ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░███ │   4%
 68M   │ ├─┴ a2e1b408224738d429e9a65b8ad5d94ff58cc1a7│                            ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░███ │   4%
 20M   │ │   ┌── libmkl_intel_lp64.so.2              │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓█ │   1%
 27M   │ │   ├── libmkl_sequential.so.2              │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓██ │   2%
 29M   │ │   ├── libmkl_gnu_thread.so.2              │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓██ │   2%
 36M   │ │   ├── libmkl_pgi_thread.so.2              │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓██ │   2%
 38M   │ │   ├── libmkl_tbb_thread.so.2              │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓██ │   3%
 40M   │ │   ├── libmkl_def.so.2                     │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓██ │   3%
 46M   │ │   ├── libmkl_mc.so.2                      │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓██ │   3%
 47M   │ │   ├── libmkl_avx2.so.2                    │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓██ │   3%
 48M   │ │   ├── libmkl_mc3.so.2                     │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓██ │   3%
 50M   │ │   ├── libmkl_avx.so.2                     │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓███ │   3%
 62M   │ │   ├── libmkl_intel_thread.so.2            │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓███ │   4%
 63M   │ │   ├── libmkl_avx512.so.2                  │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓███ │   4%
 70M   │ │   ├── libmkl_core.so.2                    │                            ░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓███ │   5%
760M   │ │ ┌─┴ lib                                   │                            ░░░░████████████████████████████████ │  49%
760M   │ ├─┴ 347e4bf25d69805922225ce6bf819ef0b8715426│                            ░░░░████████████████████████████████ │  49%
864M   ├─┴ artifacts                                 │                            ████████████████████████████████████ │  56%
1.5G ┌─┴ julia_tmp_depot
```
So it seems that MKL_jll has been eagerly downloaded; How to make it truly lazy then?

Note that we `import SCS_MKL_jll` within `@requrie` block:
```julia
    Requires.@require(
        MKL_jll = "856f044c-d86e-5d09-b602-aeab76dc8ba7",
        begin
            import SCS_MKL_jll
            if SCS_MKL_jll.is_available()
                import SCS_MKL_jll.MKL_jll
                global mkldirect = SCS_MKL_jll.libscsmkl
                push!(available_solvers, MKLDirectSolver)
            end
        end
    )
```

giordano:
> So it seems that MKL_jll has been eagerly downloaded

Because something has done `using MKL_jll` somewhere somewhere.

giordano:
More direct example:
```
% JULIA_PKG_SERVER="pkg.julialang.org" JULIA_DEPOT_PATH=$(mktemp -d) ./julia --startup-file=no --project=/tmp -q
(tmp) pkg> add MKL_jll
  Installing known registries into `/tmp/tmp.ZGDfNuRhok`
    Updating registry at `/tmp/tmp.ZGDfNuRhok/registries/General.toml`
   Resolving package versions...
   Installed IntelOpenMP_jll ─ v2023.1.0+0
   Installed MKL_jll ───────── v2023.1.0+0
   Installed TimerOutputs ──── v0.5.23
   Installed ExprTools ─────── v0.1.9
   Installed Preferences ───── v1.4.0
   Installed JLLWrappers ───── v1.4.1
  Downloaded artifact: IntelOpenMP
    Updating `/tmp/Project.toml`
  [856f044c] + MKL_jll v2023.1.0+0
    Updating `/tmp/Manifest.toml`
  [692b3bcd] + JLLWrappers v1.4.1
  [21216c6a] + Preferences v1.4.0
  [1d5cc7b8] + IntelOpenMP_jll v2023.1.0+0
  [856f044c] + MKL_jll v2023.1.0+0
  [0dad84c5] + ArgTools v1.1.1
  [56f22d72] + Artifacts
  [2a0f44e3] + Base64
  [ade2ca70] + Dates
  [f43a241f] + Downloads v1.6.0
  [7b1f6079] + FileWatching
  [b77e0a4c] + InteractiveUtils
  [4af54fe1] + LazyArtifacts
  [b27032c2] + LibCURL v0.6.4
  [76f85450] + LibGit2
  [8f399da3] + Libdl
  [56ddb016] + Logging
  [d6f4376e] + Markdown
  [ca575930] + NetworkOptions v1.2.0
  [44cfe95a] + Pkg v1.10.0
  [3fa0cd96] + REPL
  [9a3f8284] + Random
  [ea8e919c] + SHA v0.7.0
  [9e88b42a] + Serialization
  [6462fe0b] + Sockets
  [fa267f1f] + TOML v1.0.3
  [a4e569a6] + Tar v1.10.0
  [cf7118a7] + UUIDs
  [deac9b47] + LibCURL_jll v8.0.1+0
  [29816b5a] + LibSSH2_jll v1.10.2+0
  [c8ffd9c3] + MbedTLS_jll v2.28.2+0
  [14a3606d] + MozillaCACerts_jll v2023.1.10
  [83775a58] + Zlib_jll v1.2.13+0
  [8e850ede] + nghttp2_jll v1.52.0+0
  [3f19e933] + p7zip_jll v17.4.0+0
Precompiling project...
  10 dependencies successfully precompiled in 2 seconds

julia> using MKL_jll
  Downloaded artifact: MKL
```

giordano:
But this PR still doesn't make any sense to me.  Generating a broken JLL is a no-go.  I believe you're trying to address the problem at the wrong level.

kalmarek:
SCS_GPU_jll also has CUDA_full_jll only as `BuildDependency`, and you may say that it is broken on face value. Since we're the only consumers of those jlls we know when to enable them so that the users experience non-broken code.

If this is not the right way of achieving this, what are other possibilities?
I believed that hiding `import SCS_MKL_jll` behind `@requires` does the trick, but apparently it doesn't, hence this pull request.

for reference this is our `__init__()`:
https://github.com/jump-dev/SCS.jl/blob/master/src/SCS.jl#L13-L38

odow:
## This PR

So this PR is breaking if people to use SCS_MKL_jll directly, and they haven't also run `using MKL_jll`.

But @kalmarek's point is that no one is using SCS_MKL_jll outside of SCS.jl, and that SCS.jl doesn't load SCS_MKL_jll unless MKL_jll is already loaded.

This is also the case for SCS_GPU_jll, which only loads if CUDA.jl is also loaded: 
https://github.com/JuliaPackaging/Yggdrasil/blob/d55279a3b2f5d17c61a256c3edf3ba9f0d009d71/S/SCS_GPU/build_tarballs.jl#L42

## Alternative

@giordano's suggestion is to remove `SCS_MKL_jll` and `SCS_GPU_jll` from
https://github.com/jump-dev/SCS.jl/blob/ae9186d7f45ef809c3075fa57df3c314d68fe263/Project.toml#L17-L18
and make them the "requires" packages instead of MKL_jll and CUDA_jll. That would break existing users of SCS.jl, but that's okay. And it might even be a good excuse to try out the weak dependency stuff in Julia v1.9. But to do this we'd first have to fix the SCS_GPU_jll build.

## Conclusion

To be honest, I'd prefer this PR for now. But we'll need to bump the MAJOR version number because it's a breaking change. Then once Julia 1.9 or 1.10 is the LTS, we can switch to a weakdep and fix this properly?

kalmarek:
I see, I didn't think of asking users to `using SCS_MKL_jll` directly, but that's also a possible solution.

S/SCS_MKL/build_tarballs.jl
```diff
@@ -37,7 +38,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("MKL_jll"; compat="2022.2.0"),
+    BuildDependency(PackageSpec(;name="MKL_jll", version=v"2022.2")),
```
odow, line 41: @kalmarek latest version is 2023.1. Do we want to update to that? x-ref https://github.com/jump-dev/SCS.jl/issues/274

kalmarek: We do ;) But first we'd need to come to a conclusion what to do with the `BuildDependency` vs `Dependency`.

Updating the version of MKL_jll is a trivial thing ;)

kalmarek:
> To be honest, I'd prefer this PR for now. But we'll need to bump the MAJOR version number because it's a breaking change. Then once Julia 1.9 or 1.10 is the LTS, we can switch to a weakdep and fix this properly?

@odow So this is a breaking change for `SCS_MKL_jll`, but I'd argue this is a bugfix for `SCS.jl` i.e. the intended behaviour was not to download `MKL_jll` just because somebody does `add SCS`. Are we on the same page?

As I said, since we're the only users of `SCS_MKL_jll` I was tempted not to bump the MAJOR, but maybe that's the easiest thing; The other one is (to keep in sync with version of `scs` is to make it into `v301.200.300` as some people here do). but I'm fine with `4.0.0` as well

S/SCS_MKL/build_tarballs.jl
kalmarek:
```suggestion
    BuildDependency(PackageSpec(;name="MKL_jll", version=v"2023.1")),
```

odow:
> The other one is (to keep in sync with version of scs is to make it into v301.200.300 as some people here do)

This is what I had in mind.  

> since we're the only users of SCS_MKL_jll I was tempted not to bump the MAJOR,

But yes, I'd agree that this is a bug fix, so I'm okay if you want to leave as-is.

kalmarek:
@giordano We're happy to merge this, as @odow summarised above;
If you have any further objections please let us know.

giordano:
I can't say _I_ am happy.  This won't even work in the registration in General.

> But @kalmarek's point is that no one is using SCS_MKL_jll outside of SCS.jl

My point is that you can't control that.  You can't forbid users from using this package directly.

I hope it's clear on record that I don't like this and I don't agree with the change.  I'm going to merge this just to move on but I won't deal with any breakage this should cause.

PR #6778, opened 2023-05-21
Update LibTeXPrintf to upstream v1.17

stevengj:


stevengj:
Build error:
```
[20:11:02] mv -f .deps/drawbox.Tpo .deps/drawbox.Po
[20:11:02] mv -f .deps/drawbox.Tpo .deps/drawbox.Plo
[20:11:02] mv: can't rename '.deps/drawbox.Tpo': No such file or directory
```
Looks suspicious that it's trying to  `mv -f deps/drawbox.Tpo` twice to two different destinations?  Could be an upstream bug?

giordano:
Yes, and given the randomness of the failures and the fact that restarting failed jobs makes them succeed, my hunch is that this is due to the build being racy. I guess the developers don't frequently build with `-j16` (and maybe should learn about the new option `--shuffle` in gnu make 4.4)

Suavesito-Olimpiada:
Yes, previous builds were also random for me. I'll report the bug upstream.

P.S. Thanks for the tip about `--shufle`! That seems like a nice addition.

stevengj:
I found a fix for the problem — I'll submit a PR upstream.

PR #6779, opened 2023-05-21
add builder for isa-l v2.30.0

maltesie:


L/libisal/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "musl")
+]
+
```
giordano, line 31: Maybe this can work on other platforms?
```suggestion
platforms = supported_platforms()
```

L/libisal/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libisal", :libisal),
+    ExecutableProduct("igzip", :igzip)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    HostBuildDependency(PackageSpec(name="NASM_jll", uuid="08ca2550-6d73-57c0-8625-9b24120f3eae"))
+    HostBuildDependency(PackageSpec(name="YASM_jll", uuid="997772c2-56d0-5ccd-9329-3f55f14e5768"))
```
giordano, line 42: https://github.com/intel/isa-l/blob/2bbce31943289d5696bcf2a433124c50928226a2/README.md#prerequisites suggests using nasm as assembler on x86_64, with yasm as an _alternative_, so maybe it's sufficient to have only the former?

L/libisal/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "libisal"
```
giordano, line 5: https://docs.binarybuilder.org/stable/building/#Name

> This is the name that will be used in the tarballs and for the JLL package. It should be the name of the upstream package, not for example that of a specific library or executable provided by it, even though they may coincide.

If you're going to rename this variable, remember that the name of the directory must match exactly the `name`.

L/libisal/build_tarballs.jl
maltesie: I tried to build for all platforms. On aarch it complains about unknown mnemonics and on Apple and Windows it complains either not to have found NASM or not in a recent version (>= 2.11.01). I checked, and NASM_jll should be available for all platforms in version 2.15. I dont know how to address this.

L/libisal/build_tarballs.jl
maltesie: I used isa_l now instead of the actual name isa-l.

L/libisal/build_tarballs.jl
maltesie: I also tried with both, NASM_jll and YASM_jll and for those platforms, where NASM isnt detected properly, detection of YASM also fails.

maltesie: For Windows, there is an alternative Makefile and it should be used like this:
`nmake -f Makefile.nmake`
But I also did not succeed to make this work.

giordano: For aarch64 you need gcc 8, because it's trying to compile for SVE.

giordano: On Windows it finds yasm no problem for me, but it fails to compile some avx512 assembly

PR #6790, opened 2023-05-24
[NetCDF] Set compat bounds for HDF5_jll up to patch, ~1.12.2 [skip build]

mkitti:
Set compat bounds for HDF5_jll

NetCDF_jll probably should be ~1.12.2 unless established otherwise

Otherwise, this hould finish out the diff for #6707 except that I skip build.

### DIFF
--- a/N/NetCDF/build_tarballs.jl
+++ b/N/NetCDF/build_tarballs.jl
@@ -98,7 +98,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="HDF5_jll"), compat="~1.12"),
+    Dependency(PackageSpec(name="HDF5_jll"), compat="~1.12.2"),
     Dependency("Zlib_jll"),
     Dependency("XML2_jll"),
     Dependency("LibCURL_jll"; compat = "7.73.0"),


S/SZ/build_tarballs.jl
```diff
@@ -80,7 +80,8 @@ dependencies = [
                platforms=filter(!Sys.isbsd, platforms)),
     Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e");
                platforms=filter(Sys.isbsd, platforms)),
-    Dependency("HDF5_jll"; platforms=hdf5_platforms),
+    # Updating to a newer HDF5 version is likely possible without problems but requires rebuilding this package
```
giordano, line 83: This is already on `master` https://github.com/JuliaPackaging/Yggdrasil/commit/d995ccc02f0de759380aea9d005ca9cf51a146b6 or am I missing something?

N/NetCDF/build_tarballs.jl
```diff
@@ -98,7 +98,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="HDF5_jll"), compat="~1.12"),
+    Dependency(PackageSpec(name="HDF5_jll"), compat="~1.12.2"),
```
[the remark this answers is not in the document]
mkitti: This is is different. 3fbc357 changed the compat to ~1.12. Here we change the compat to ~1.12.2 which I think is needed for NetCDF_jll to function correctly.

S/SZ/build_tarballs.jl
mkitti: I think this part is already merged from #6707 . I did this before the automerge completed.

PR #6792, opened 2023-05-24
Deno: update to 1.33.4 and add aarch64-linux-gnu

fonsp:
This is a regular update to 1.33.4, and also I am adding the ARM linux binary.

Deno currently builds from source on ARM, but they don't publish a binary in their [releases](https://github.com/denoland/deno/releases) ([because they only use GH Actions](https://github.com/denoland/deno/issues/1846)). There is however an ARM release from the community: https://github.com/LukeChannings/deno-arm64/releases

So our options for Deno_jll are building from source, or using the community release. It [sounds like](https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1166#issuecomment-1030819653) building from source on Linux ARM is not supported with BinaryBuilder, so let's go with the community release!

### DIFF
--- a/D/Deno/build_tarballs.jl
+++ b/D/Deno/build_tarballs.jl
@@ -3,16 +3,18 @@
 using BinaryBuilder, Pkg
 
 name = "Deno"
-version = v"1.28.1"
+version = v"1.33.4"
 
 release_url = "https://github.com/denoland/deno/releases/download/v$version"
+release_arm_url = "https://github.com/LukeChannings/deno-arm64/releases/download/v$version"
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


fonsp:
@Pangoraw could you review?

D/Deno/build_tarballs.jl
```diff
@@ -27,6 +29,7 @@ install -m 755 "${target}/deno${exeext}" "${bindir}"
 # platforms are passed in on the command line
 platforms = [
     Platform("x86_64", "linux"; libc="glibc"),
+    Platform("aarch64", "linux"; libc="glibc"),
```
fonsp, line 32: I just copied the `libc` keyword argument from x86 linux, but do we also need it here?

giordano: Yes

giordano:
Thanks!

PR #6801, opened 2023-05-26
[MiniZinc] update to 2.7.4

odow:
cc @chriscoey

Quite a bit changed since the last build: https://github.com/MiniZinc/libminizinc/compare/2.6.2...2.7.4

So not sure if this will succeed. I didn't try locally.

### DIFF
--- a/M/MiniZinc/build_tarballs.jl
+++ b/M/MiniZinc/build_tarballs.jl
@@ -4,12 +4,12 @@ using BinaryBuilder, Pkg
 
 name = "MiniZinc"
 
-version = v"2.6.2"
+version = v"2.7.4"
 
 sources = [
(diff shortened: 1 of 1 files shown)


M/MiniZinc/build_tarballs.jl
```diff
@@ -51,7 +51,7 @@ products = [
 platforms = expand_cxxstring_abis(supported_platforms())
 
 # TODO(odow): fix build issues on Windows
-platforms = filter(!Sys.iswindows, platforms)
+# platforms = filter(!Sys.iswindows, platforms)
```
odow, line 54:
```suggestion
platforms = filter(!Sys.iswindows, platforms)
```

PR #6807, opened 2023-05-26
[MiniZinc] build with HiGHS

odow:
https://github.com/JuliaPackaging/Yggdrasil/pull/6801 rebuild v2.7.4, but it didn't actually build with HiGHS because the `vendor` directory isn't downloaded by default. 

We could potentially use `HiGHS_jll` here, but let's see if libminizinc's vendored version works first.

cc @chriscoey


odow:
Success! I'll chalk this up to another obvious in retrospect but not obvious at the time problems...

PR #6808, opened 2023-05-27
[ITSOL_2] Updated for Julia 1.9 build

JuhaHeiskala:
Version 0.1.1 dit not build under Julia 1.9

Updated to use libblastrampoline v5.4+ instead of BLAS/LAPACK

LLVM11 requirement left as is from v0.1.1

### DIFF
--- a/I/ITSOL_2/build_tarballs.jl
+++ b/I/ITSOL_2/build_tarballs.jl
@@ -25,7 +25,7 @@ install_license ../{COPYRIGHT,LGPL}
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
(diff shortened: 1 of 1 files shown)


I/ITSOL_2/build_tarballs.jl
```diff
@@ -25,7 +25,6 @@ install_license ../{COPYRIGHT,LGPL}
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-# FreeBSD build failed with libgfortran_version=3.0.0
 platforms = supported_platforms(; experimental=true)
```
giordano, line 28:
```suggestion
platforms = supported_platforms()
```

PR #6809, opened 2023-05-27
[ZITSOL_1] Updated for Julia 1.9

JuhaHeiskala:
Version 0.1.1 dit not build under Julia 1.9

Updated to use libblastrampoline v5.4+ instead of BLAS/LAPACK

LLVM left at version 11 as set for v0.1.1

### DIFF
--- a/Z/ZITSOL_1/build_tarballs.jl
+++ b/Z/ZITSOL_1/build_tarballs.jl
@@ -26,7 +26,7 @@ install_license ../{COPYRIGHT,LGPL}
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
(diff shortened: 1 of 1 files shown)


Z/ZITSOL_1/build_tarballs.jl
```diff
@@ -26,7 +26,6 @@ install_license ../{COPYRIGHT,LGPL}
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-# FreeBSD build failed with libgfortran_version=3.0.0
 platforms = supported_platforms(; experimental=true)
```
giordano, line 29:
```suggestion
platforms = supported_platforms()
```

PR #6817, opened 2023-05-29
[MiniZinc] add support for Windows

odow:
This PR is going to be a WIP until I figure out the reasons why I punted on Windows support previously.


odow:
So I have now learnt that `wstring` in MinGW is complicated.

PR #6828, opened 2023-05-31
NetCDF: apply curl CA info patch

Alexander-Barth:
This PR back-ports the following patch to allow connection over https:

https://github.com/Unidata/netcdf-c/pull/2690

Without the patch the CA certificates are not found.

Ref:
https://github.com/Unidata/netcdf-c/issues/2705
https://github.com/Unidata/netcdf4-python/issues/1246#issuecomment-1529268770

CC @visr

### DIFF
--- a/N/NetCDF/build_tarballs.jl
+++ b/N/NetCDF/build_tarballs.jl
@@ -34,9 +34,8 @@ export LDFLAGS="-L${libdir}"
 export LDFLAGS_MAKE="${LDFLAGS}"
 CONFIGURE_OPTIONS=""
 
-for p in ../patches/*.patch; do
-    atomic_patch -p1 "${p}"
-done
+# Apply patch https://github.com/Unidata/netcdf-c/pull/2690
+atomic_patch -p1 ../patches/0001-curl-cainfo.patch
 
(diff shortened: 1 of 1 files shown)


N/NetCDF/build_tarballs.jl
```diff
@@ -33,6 +34,10 @@ export LDFLAGS="-L${libdir}"
 export LDFLAGS_MAKE="${LDFLAGS}"
 CONFIGURE_OPTIONS=""
 
+for p in ../patches/*.patch; do
+    atomic_patch -p1 "${p}"
+done
```
giordano, line 39: It'd be easier to add a reference to where the patch comes from
```suggestion
# Apply patch https://github.com/Unidata/netcdf-c/pull/2690
atomic_patch -p1 ../patches/0001-curl-cainfo.patch
```

Alexander-Barth:
Thanks for the suggestion @giordano !

Alexander-Barth:
For information, the produced NetCDF_jll binary passed all tests on Linux, Mac and Windows:

https://github.com/Alexander-Barth/NCDatasets.jl/actions/runs/5130586314

PR #6832, opened 2023-06-01
[Singular] re-release update with increased minor version

fingolfin:
Turns out this update was ABI breaking...


fingolfin:
See https://github.com/JuliaRegistries/General/pull/84667 (and this PR here should not be merged before that PR there is merged)

CC @hannes14 @thofma

PR #6836, opened 2023-06-02
Compile Ipopt with LBT for all platforms

amontoison:
Please do not merge.
I need to test if we can switch to the SPRAL solver with the Julia interface Ipopt.jl.

### DIFF
--- a/C/Coin-OR/Ipopt/build_tarballs.jl
+++ b/C/Coin-OR/Ipopt/build_tarballs.jl
@@ -26,12 +26,16 @@ if [[ ${target} == *mingw* ]]; then
 fi
 
 if [[ "${target}" == *mingw* ]]; then
-  BLAS_LAPACK="-L${libdir} -lopenblas"
+  LBT="-lblastrampoline-5"
 else
-  BLAS_LAPACK="-L${libdir} -lblastrampoline"
+  LBT="-lblastrampoline"
 fi
 
 ./configure \
+    CFLAGS="-O3 -DNDEBUG" \
+    CXXFLAGS="-O3 -DNDEBUG" \
+    FFLAGS="-O3" \
+    FCFLAGS="-O3" \
     --prefix=${prefix} \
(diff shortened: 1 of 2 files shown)


C/Coin-OR/Ipopt/build_tarballs.jl
```diff
@@ -40,10 +44,12 @@ fi
...
     --disable-dependency-tracking \
     lt_cv_deplibs_check_method=pass_all \
-    --with-lapack-lflags="${BLAS_LAPACK}" \
+    --with-lapack-lflags="-L${libdir} ${LBT}" \
     --with-mumps-cflags="-I${includedir}" \
-    --with-mumps-lflags="-ldmumps -lzmumps -lcmumps -lsmumps -lmumps_common -lmpiseq -lpord -lmetis ${BLAS_LAPACK} -lgfortran -lpthread" \
-    --with-asl-lflags="${LIBASL}"
+    --with-mumps-lflags="-L${libdir} -lsmumps -ldmumps -lcmumps -lzmumps -lmumps_common -lmpiseq -lpord -lmetis ${LBT} -lgfortran -lpthread" \
+    --with-spral-cflags="-I${includedir}" \
+    --with-spral-lflags="-L${libdir} -lspral -lgfortran -lhwloc -lm -lmetis ${LBT} -lstdc++ -lgomp" \
```
giordano, line 51: It looks a bit weird we have to specify all the standard runtime libraries like gfortran and stdc++

amontoison: The build systems of MUMPS and SPRAL only generated static libraries.
They probably do something similar to what we do with `flagon` [here](https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SPRAL/build_tarballs.jl#L39).

Because we generated shared libraries for these packages with Yggdrasil, we just need to specify `-ldmumps` and `-lspral`.
Thanks Mosè :smiley:

PR #6838, opened 2023-06-03
[MUSCLE] add MacOS build

M-PERSIC:
- Replace GCC with Clang (should address gccv11 concerns, see [Building under OSX](https://github.com/rcedgar/muscle/wiki/Building-MUSCLE#building-under-osx))
- `LLVMOpenMP_jll` dependency for non-Windows builds

---------

Co-authored-by: James Foster <James.Foster@csiro.au>

### DIFF
--- a/M/MUSCLE/build_tarballs.jl
+++ b/M/MUSCLE/build_tarballs.jl
@@ -11,14 +11,21 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd ${WORKSPACE}/srcdir/muscle/src/
-make -j${nproc} CXXFLAGS="-O3 -fopenmp"
-install -Dvm 755 "Linux/muscle" "${bindir}/muscle"
+
+if [[ "${target}" == *-apple-* ]]; then
+    make -j${nproc} CXX="clang++" CXXFLAGS="-O3 -fopenmp"
+else
+    make -j${nproc} CXXFLAGS="-O3 -fopenmp"
+fi
+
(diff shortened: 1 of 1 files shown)


M/MUSCLE/build_tarballs.jl
```diff
@@ -11,14 +11,21 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd ${WORKSPACE}/srcdir/muscle/src/
-make -j${nproc} CXXFLAGS="-O3 -fopenmp"
-install -Dvm 755 "Linux/muscle" "${bindir}/muscle"
+
+if [[ "${target}" == *-apple-* ]]; then
+    make -j${nproc} CXX="clang++" CXXFLAGS="-O3 -fopenmp"
```
giordano, line 16: Use CXX=c++ everywhere and you don't need an if

M/MUSCLE/build_tarballs.jl
```diff
@@ -11,14 +11,21 @@ sources = [
...
+else
+    make -j${nproc} CXXFLAGS="-O3 -fopenmp"
+fi
+
+OS_NAME=`uname`
+install -Dvm 755 "${OS_NAME}/muscle" "${bindir}/muscle"
```
giordano, line 22: Can also do
```suggestion
install -Dvm 755 "$(uname)/muscle" "${bindir}/muscle"
```
if you don't need the variable for anything else

PR #6849, opened 2023-06-06
Add CUDA support to a version of SuperLU

Wimmerer:
May try HIP later as well, several labs have AMD gpu clusters instead of CUDA.

I couldn't find examples of CUDA and MPI augmentation, let's see if this works.


Wimmerer:
@liuyangzhuan this is the attempt to add a CUDA version of the SuperLU builder. It's a little involved but the error right now seems to be that `find_package(CUDA)` isn't finding CUDA 11. It seems to find `nvcc` but doesn't like the `CUDART` and `INCLUDE_DIRS` that it finds.

fingolfin:
Hi there, this hasn't been updated in over 2 years, and likely needs changes to be in a mergable state.

Given that, I am going to close this for now. But of course if you are still interested in this, feel free to open a fresh PR.

PR #6850, opened 2023-06-07
New Recipe: USalign v2022.9.24

diegozea:


U/USalign/build_tarballs.jl
```diff
@@ -0,0 +1,89 @@
...
+    '    };\n' \
+    '}\n' \
+    '\n#endif // PSTREAM_H' > pstream.h;
+    # Fix the EXIT_SUCCESS not declared error
+    sed -i '1i#include <cstdlib>' pdbAtomName.cpp
+fi
+make
+PROGRAM=$(grep "PROGRAM=" Makefile | cut -d '=' -f2)
+for prog in $PROGRAM; do
+    cp $prog ${bindir};
```
giordano, line 57: More verbose and ensures that the files have the file permissions (and the directory is created automatically if needed)
```suggestion
    install -Dvm 755 "${prog}" "${bindir}/${prog}"
```

U/USalign/build_tarballs.jl
```diff
@@ -0,0 +1,89 @@
...
+    '        }\n' \
+    '\n        void open(const std::string&) {\n' \
+    '            std::cout << "ipstream::open is not supported on this platform.\\n";\n' \
+    '        }\n' \
+    '        bool good() const { return false; }\n' \
+    '        void close() {}\n' \
+    '        friend ipstream& getline(ipstream& is, std::string& str) { return is; }\n' \
+    '    };\n' \
+    '}\n' \
+    '\n#endif // PSTREAM_H' > pstream.h;
```
giordano, line 49: Do you know [here documents](https://tldp.org/LDP/abs/html/here-docs.html)?  Also, it may be easier to add the the header file as an [extra source](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.DirectorySource) and copy the file where you need it.  Where does it come from BTW?

U/USalign/build_tarballs.jl
diegozea: Thanks! I didn't know that that was safer :)

U/USalign/build_tarballs.jl
diegozea: Thanks! :) I didn't know about the extra source possibility. I will use the _here document_ as the header is on UNIX. This is only a dummy file I created defining the needed methods to avoid the compilation failing on Windows where POSIX and pstream are not supported (as far as I understood).

U/USalign/build_tarballs.jl
```diff
@@ -0,0 +1,97 @@
...
+}
+#endif // PSTREAM_H
+CONTENT
+
+fi
+
+make
+PROGRAM=$(grep "PROGRAM=" Makefile | cut -d '=' -f2)
+for prog in $PROGRAM; do
+    install -Dvm 755 "${prog}" "${bindir}/${prog}"
```
giordano, line 64: Can't check it myself, but for Windows that's probably
```suggestion
    install -Dvm 755 "${prog}${exeext}" "${bindir}/${prog}${exeext}"
```

giordano: OR maybe
```suggestion
    install -Dvm 755 "${prog}" "${bindir}/${prog}${exeext}"
```
I don't know and can't check what's the actual name of the file on disk.

giordano:
Thanks for your contribution!

PR #6863, opened 2023-06-08
T8code with p4est and libsc

jmark:
I added `libp4est` and `libsc` as library products so that Windows will find all three libraries for symbol search.


jmark:
@giordano Please review and merge. Thank you! :)

PR #6866, opened 2023-06-08
[polymake] Don't use the tarball from GitHub releases

ararslan:
`ArchiveSource` complains that the checksum is unreliable. Luckily polymake.org has downloads that we can use instead.

### DIFF
--- a/P/polymake/build_tarballs.jl
+++ b/P/polymake/build_tarballs.jl
@@ -30,8 +30,8 @@ version = VersionNumber(upstream_version.major*100+version_offset.major,
 
 # Collection of sources required to build polymake
 sources = [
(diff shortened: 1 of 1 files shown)


giordano:
I think you want to use [`[skip build]`](https://github.com/JuliaPackaging/Yggdrasil/blob/23df258503e9e956d93fec040febc1dac4d7af4e/CONTRIBUTING.md#special-keywords-in-commit-messages), not `[skip ci]`, the latter shall [only be used on merge commits](https://github.com/JuliaPackaging/Yggdrasil/blob/23df258503e9e956d93fec040febc1dac4d7af4e/CONTRIBUTING.md#information-for-maintainers)

ararslan:
Amended commit message to use that instead 👍

PR #6869, opened 2023-06-11
[ViennaRNA] update to v2.6.0

marcom:


V/ViennaRNA/build_tarballs.jl
```diff
@@ -1,45 +1,59 @@
...
+# otherwise always use g++
 export CC=cc
 export CXX=c++
 
 # where to find dependencies
 export CPPFLAGS="-I${includedir}"
 export LDFLAGS="-L${libdir}"
 
+# fix compile fail on musl
+# ViennaRNA PR: https://github.com/ViennaRNA/ViennaRNA/pull/187
```
giordano, line 37: Thanks!

V/ViennaRNA/build_tarballs.jl
```diff
@@ -1,45 +1,59 @@
...
+    ArchiveSource("https://www.tbi.univie.ac.at/RNA/download/sourcecode/" *
+                  "$(version.major)_$(version.minor)_x/ViennaRNA-$(version).tar.gz",
+                  "c239596ac63ff30d98c7629f3cfeaa6a066dc3b577e1b920eb704381bb6c3f85"),
     DirectorySource("./bundled")
 ]
 
-# Bash recipe for building across all platforms
-# Build issues
+# Build issues and notes
+# - windows: disabled RNAxplorer build, regex.h not found
```
giordano, line 14: If you want to try it, see https://github.com/JuliaPackaging/Yggdrasil/blob/b8e358c6e37ec47c58f177a900aa5a8f78ad6eb7/L/LibUCL/build_tarballs.jl#L17 and the rest of the recipe

marcom: Thanks, that helped. Had to also work around a missing `sysconf(_SC_NPROCESSORS_ONLN)` with `GetActiveProcessorCount(ALL_PROCESSOR_GROUPS)`. Now I'm stuck on missing `getline`, any ideas? Googling around suggests taking one from NetBSD sources or similar.

Isn't there some kind of compatibility shim that supplies all these missing POSIX things for Windows? Then we could just use `MissingPosixForWindows_jll` and be done? Maybe cygwin has these things?

marcom: I used the one from NetBSD

giordano: >  and the rest of the recipe

Was referring also to
https://github.com/JuliaPackaging/Yggdrasil/blob/b8e358c6e37ec47c58f177a900aa5a8f78ad6eb7/L/LibUCL/build_tarballs.jl#L35-L38

PR #6872, opened 2023-06-11
new recipe: protobuf-c

aviks:
[WIP] having trouble with running `protoc` at buildtime.


P/protobuf-c/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/protobuf-c/protobuf-c/releases/download/v$(version)/protobuf-c-$(version).tar.gz", "4cc4facd508172f3e0a4d3a8736225d472418aee35b4ad053384b137b220339f")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd protobuf-c-1.4.1/
```
giordano, line 16:
```suggestion
cd $WORKSPACE/srcdir/protobuf-c*
```

P/protobuf-c/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
```
giordano, line 17: Haven't tried, but this _should_ make it find the right `protoc`
```suggestion
./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} PROTOC="$(which protoc)"
```

P/protobuf-c/build_tarballs.jl
giordano, line 1: The directory should be called `protobuf_c` to match exactly the name of the recipe

PR #6875, opened 2023-06-12
New Recipe: NATS C client

aviks:
https://nats.io/ and https://github.com/nats-io/nats.c 

~Windows can't find OpenSSL apparently.~ ✅

### DIFF
--- a/N/nats_c/build_tarballs.jl
+++ b/N/nats_c/build_tarballs.jl
@@ -30,23 +30,8 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = [
-    Platform("i686", "linux"; libc = "glibc"),
-    Platform("x86_64", "linux"; libc = "glibc"),
-    Platform("aarch64", "linux"; libc = "glibc"),
-    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
-    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
-    Platform("powerpc64le", "linux"; libc = "glibc"),
-    Platform("i686", "linux"; libc = "musl"),
-    Platform("x86_64", "linux"; libc = "musl"),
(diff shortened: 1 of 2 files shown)


aviks:
```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
--
  | [10:08:15]   Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  | [10:08:15]   system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY) (found
  | [10:08:15]   version "3.0.9")
```

aviks:
This should be good to go now.

ararslan (approved):

N/nats_c/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libnats", :libnats)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="protobuf_c_jll", uuid="d730a6b3-54e8-5a61-8821-996059275344"))
+    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"))
```
giordano, line 45: You have to specify compat for openssl 
```suggestion
    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="3.0.8")
```

aviks: Thanks. 3.0.9 seems to be the latest. Any reason not to use that?

giordano: "3.0.8" means both 3.0.8 and 3.0.9, instead of just the latter (and following versions in the 3.x series in both cases). We declare compatibility with the oldest version

PR #6884, opened 2023-06-13
update llama_cpp to latest (13 June)

aviks:
So I'm trying to update llama_cpp to the latest code, the changes are fairly obvious, but this version of the code is failing to compile. This is really strange since I can see upstream CI compiling successfully with gcc 8 and 11. And this build worked till a few weeks ago (cf: #6769 ) Submitting this PR to see if anyone has any ideas. 

cc: @marcom

### DIFF
--- a/L/llama_cpp/build_tarballs.jl
+++ b/L/llama_cpp/build_tarballs.jl
@@ -1,7 +1,7 @@
 using BinaryBuilder, Pkg
 
 name = "llama_cpp"
-version = v"0.0.9"  # fake version number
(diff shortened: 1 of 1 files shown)


aviks:
Errors seem to be in files added since our last Ygg build, but I don't see how it should fail here when upstream is successfully compiling these. Wondering what I'm missing.

aviks:
I'll leave this open and in draft for later investigation, but for the moment, I've created #6885 for the minimum update I needed. This PR should never be merged now (the version number is wrong)

PR #6885, opened 2023-06-14
update llama to commit 2d5db48 (19 May 2023)

aviks:
This updates llama_cpp_jll to support ggml file format v3.


aviks:
@giordano this is ready to merge when you have a moment, thanks!

PR #6887, opened 2023-06-14
[llama_cpp] Bump version and fix compile

marcom:
cc: @aviks 

Supersedes https://github.com/JuliaPackaging/Yggdrasil/pull/6884

Upstream issue: https://github.com/ggerganov/llama.cpp/issues/1788

With the `static_assert` to `_Static_assert` change it compiles on `x86_64-linux-gnu`. Unfortunately, it still fails to compile with a different error on `aarch64-linux-gnu`, now in `k_quants.c`. Maybe it's a bug in gcc-8.1.0 (the version in BinaryBuilder) that got fixed later (current is gcc-8.5.0), or maybe what -std=gnu11 allows changed between versions? Not sure what the C11 standard says wrt freestanding `static_assert`.

I've disabled the k_quants stuff for now. I left the patch in for when we want to fix k_quant compilation on the other arches later. Another option would be to enable it only on the arches where it works. What do you think @aviks?

Note: I'm compiling gcc-8.1.0 and gcc-8.5.0 to see if I can reproduce the build fail/success, but that will take a few hours.


aviks:
> Maybe it's a bug in gcc-8.1.0 

They seem to use the following compilers combinations in their CI. 
```
CC:       gcc-8 (Ubuntu 8.4.0-3ubuntu2) 8.4.0
CXX:      g++ (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
```
and 
```
The C compiler identification is GNU 11.3.0
The CXX compiler identification is GNU 11.3.0
```

I tried with `preferred_gcc_version=v"11"` in BB, but that didn't work either.

aviks:
> I've disabled the k_quants stuff for now. ... What do you think @aviks?

It seems that a lot of the downloadable models on Huggingface are moving to the k_quants stuff. So I'm afraid we'll be left behind on that. But I guess this is OK for now till we figure out what to do.

marcom:
@aviks I started a Julia interface to llama_cpp here https://github.com/marcom/Llama.jl if you are interested. It's a bit buggy at the moment though

aviks:
Yes thanks! I have been using it, using llama_run. generate obviously needs some work. I'll see if I have time to contribute in a month or two.

PR #6896, opened 2023-06-16
Adding double-precision versions to FastTree

### LINKED ISSUE #6895: Request: Add double-precision version of FastTree
(I'm not 100% sure this is the right place for this request - please let me know if not).

I was delighted to see FastTree here: https://github.com/JuliaBinaryWrappers/FastTree_jll.jl

One issue is that, in many applications, we really need the version of FastTree with double-precision branch lengths. This is from their website (http://www.microbesonline.org/fasttree/):
```
Why does FastTree report so many branch lengths of 0.0005 or 0.0001, or even negative branch lengths?
By default, the maximum-likelihood phase of FastTree does not allow branch lengths of less than 0.0005, so unresolvable internal branches often end up with lengths of about 0.0005. (Before FastTree 2.1.7, the limit was 0.0001, so if you are using an older version, you may see many branch lengths of 0.0001 instead.)

If you need to resolve very-short branch lengths accurately, you can use a double-precision version of FastTree. To do this, you need to compile it with -DUSE_DOUBLE and you need to use version 2.1.8 or later, e.g.

gcc -DUSE_DOUBLE -O3 -finline-functions -funroll-loops -Wall -o FastTree FastTree.c -lm
```

Can I request that a version with `-DUSE_DOUBLE` is added?

murrellb:
Fix #6895

Caveat: I have no idea what I'm doing.

### DIFF
--- a/F/FastTree/build_tarballs.jl
+++ b/F/FastTree/build_tarballs.jl
@@ -27,16 +27,16 @@ cc -DOPENMP -fopenmp -O3 -finline-functions -funroll-loops -Wall -o "${bindir}/f
 cc -DNO_SSE -fopenmp -O3 -finline-functions -funroll-loops -Wall -o "${bindir}/fasttreeMP_noSSE${exeext}" FastTree*.c -lm
 
 # install single thread with SSE instructions and double precision
-cc -DUSE_DOUBLE -O3 -finline-functions -funroll-loops -Wall -o "${bindir}/fasttree${exeext}" FastTree*.c -lm
+cc -DUSE_DOUBLE -O3 -finline-functions -funroll-loops -Wall -o "${bindir}/fasttree_double${exeext}" FastTree*.c -lm
 
 # install single thread no SSE and double precision
-cc -DUSE_DOUBLE -DNO_SSE -O3 -finline-functions -funroll-loops -Wall -o "${bindir}/fasttree_noSSE${exeext}" FastTree*.c -lm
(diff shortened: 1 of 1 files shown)


F/FastTree/build_tarballs.jl
```diff
@@ -35,6 +47,10 @@ products = [
     ExecutableProduct("fasttree_noSSE", :fasttree_noSSE),
     ExecutableProduct("fasttreeMP", :fasttreeMP),
     ExecutableProduct("fasttreeMP_noSSE", :fasttreeMP_noSSE),
+    ExecutableProduct("fasttree_Double", :fasttree_Double),
+    ExecutableProduct("fasttree_Double_noSSE", :fasttree_Double_noSSE),
+    ExecutableProduct("fasttreeMP_Double", :fasttreeMP_Double),
+    ExecutableProduct("fasttreeMP_Double_noSSE", :fasttreeMP_Double_noSSE),
```
giordano, line 53: This is not going to work ***if*** downstream packages expect libraries with different names, hence my suggestion in https://github.com/JuliaPackaging/Yggdrasil/issues/6895#issuecomment-1594395750.  Note the "if" before, I have no idea what downstream packages expect.

giordano:
I'm going to merge this PR, I don't know how this is going to work, but I also don't care much.  At least this shouldn't break any existing package, also this isn't a library but just an executable, theoretically slightly easier to deal with, #6895 didn't provide evidence of standard direct usage of this program in downstream packages (this package doesn't even have a pkgconfig or cmake module to find it), so I guess this is sparsely and randomly used anyway.

PR #6906, opened 2023-06-19
[ViennaRNA] update to ViennaRNA-2.6.2, fix shared lib linking on julia-1.6

marcom:
To help with https://github.com/marcom/ViennaRNA.jl/pull/15

- update to ViennaRNA-2.6.2

- only link RNAxplorer to BLAS/LAPACK (and PCRE on windows), not the other programs or shared library

- add patch to fix missing include in headers, causes problems when parsing headers with Clang.jl to build bindings

### DIFF
--- a/V/ViennaRNA/build_tarballs.jl
+++ b/V/ViennaRNA/build_tarballs.jl
@@ -64,10 +64,7 @@ else
 fi
 
 
-# configure script fails (only in sandbox) without
-# setting ac_cv_func_malloc_0_nonnull=yes ac_cv_func_realloc_0_nonnull=yes
-ac_cv_func_malloc_0_nonnull=yes ac_cv_func_realloc_0_nonnull=yes \
-    ./configure \
+./configure \
     --prefix=${prefix} --build=${MACHTYPE} --host=${target} \
     --with-pic --disable-c11 \
     --with-mpfr --with-json --with-svm --with-gsl \
--- a/V/ViennaRNA/build_tarballs.jl
+++ b/V/ViennaRNA/build_tarballs.jl
@@ -25,7 +25,7 @@ sources = [
 #   works fine on julia-1.9
 
 # TODO
-# - build shared library for RNAxplorer?
+# - build shared library for RNAxplorer? (other programs like RNAforester etc?)
 #   configure script seems to indicate there is a python interface
 
 # Notes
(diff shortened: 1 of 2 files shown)


marcom:
The RNAxplorer executable links to `libblastrampoline.so.5`. For julia-1.6 on x86_64-linux this executable can't be run, because libblastrampoline_jll there only supllies `libblastrampoline.so`. @giordano: any idea how to fix this?

giordano:
LBT is only really usable starting from Julia v1.9 (I think LBT v5.4.0?), previous versions have multiple ABI compatibility problems.

marcom:
So after the next Julia LTS release, hopefully everything should be ok for LTS users.

Is there an easy way to build against a BLAS/LAPACK implementation inside BinaryBuilder without involving libblastrampoline? As far as I can see, LAPACK_jll has libblastrampoline_jll as a Dependency so it seems the answer is no.

If there is no easy way, I guess the RNAxplorer executable won't work on julia-1.6 on certain platforms. Would be ok for me too. If this is the most feasible option, the PR is ready to be merged from my side.

V/ViennaRNA/build_tarballs.jl
```diff
@@ -104,6 +113,44 @@ fi
...
+# BLAS/LAPACK, and PCRE on windows
+
+LIBS_RNAxplorer="$LIBS"
+if [[ "${target}" == *-w64-mingw32* ]]; then
+    # work around there not being a regex.h on w64-mingw32
+    cp "${includedir}/pcreposix.h" "${includedir}/regex.h"
+    LIBS_RNAxplorer="$LIBS_RNAxplorer -lpcreposix-0"
+
+    # BLAS/LAPACK support
+    LIBS_RNAxplorer="$LIBS_RNAxplorer -lblastrampoline-5"
```
giordano, line 127: Uhm, I'm a bit confused, I don't see libblastrampoline.  Also, `-lblastrampoline-5` on Windows requires lbt v5.4.0 (I think) which is only compatible with Julia v1.9+.

marcom: > Uhm, I'm a bit confused, I don't see libblastrampoline.

I need `LAPACK_jll` to build the RNAxplorer executable, and `liblapack.so` from `LAPACK_jll v3.11.0+0` links to `libblastrampoline.so.5`.  That's why I need to add `-lblastrampoline` to `$LIBS_RNAxplorer`.

> Also, `-lblastrampoline-5` on Windows requires lbt v5.4.0 (I think) which is only compatible with Julia v1.9+.

So use `julia_compat="1.6"` in the call to `build_tarballs`?

marcom:
In the meantime, ViennaRNA-2.6.2 got released... Should I also bump the version in this release?

Re `julia_compat=1.9`:

The downside to having to increase the `julia_compat=1.9` is that now the newest installable version of `ViennaRNA_jll` is completely broken for julia-1.6, i.e. none of the programs nor the shared library work.

If we leave `julia_compat=1.6` then with the changes in this PR where only the RNAxplorer executable links against libblastrampoline, only the RNAxplorer executable would be broken on julia-1.6, but everything else would still work.

giordano:
> The downside to having to increase the julia_compat=1.9 is that now the newest installable version of ViennaRNA_jll is completely broken for julia-1.6, i.e. none of the programs nor the shared library work.

I mean, if you want to use libblastrampoline you don't have many choices: you either use it only with Julia v1.9, or you don't use it, stop.  If you want to have a version supporting julia v1.6 you don't use libblastrampoline, it's as easy as that.  I'm just stating how things are.

marcom:
Ok, thanks for the clarification.

I didn't realize that OpenBLAS already provides LAPACK functionality, thinking LAPACK_jll was the only way to get it (and LAPACK_jll forced me to use libblastrampoline).

I've now changed the build to work without libblastrampoline until the next julia LTS release.

PR #6907, opened 2023-06-19
[SLATE] Try updating with LBT & CUDA support

Wimmerer:
I fully expect this to die, but it's too large to test locally.

### DIFF
--- a/S/SLATE/build_tarballs.jl
+++ b/S/SLATE/build_tarballs.jl
@@ -2,6 +2,8 @@ using BinaryBuilder
 using Base.BinaryPlatforms
 const YGGDRASIL_DIR = "../.."
 include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
+include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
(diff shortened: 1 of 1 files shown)


S/SLATE/build_tarballs.jl
```diff
@@ -61,12 +94,30 @@ products = [
...
+    if !isnothing(cuda_version)
+        push!(dependencies, BuildDependency(PackageSpec(name="CUDA_full_jll", version=cuda_full_versions[cuda_version])))
+        push!(dependencies, RuntimeDependency(PackageSpec(name="CUDA_Runtime_jll")))
+    end
+    
+    build_tarballs(ARGS, name, version, sources, script, [augmented_platform], products, dependencies; 
+                    preferred_gcc_version=v"8", 
+                    julia_compat="1.6",
+                    augment_platform_block)
+end
```
giordano, line 123:
```suggestion
end

```

Wimmerer:
Hmm. I don't think I want `cuda-plus-none` on all these

fingolfin:
@rayegun hi there, is this still something you are working on and would like to complete, or can it be closed now?

PR #6929, opened 2023-06-22
[geotiff] build with libtiff 4.5

visr:
Looks like the [upgrade](https://github.com/JuliaPackaging/Yggdrasil/pull/6924) from libtiff 4.4 to 4.5 was [breaking](https://github.com/JuliaBinaryWrappers/Libtiff_jll.jl/commit/fddff204e013fe40cff4fac05fa9289adaaa8034#diff-248757425c80d519c91e1be718766832fdd989ae7837c5c4b6e64f95931a4757) at least on Windows.

Reported in https://github.com/JuliaGeo/GDAL.jl/issues/158
Also addresses one case of #6857 in an effort to get nightly support working again (https://github.com/JuliaGeo/GDAL.jl/issues/152).

### DIFF
--- a/L/libgeotiff/build_tarballs.jl
+++ b/L/libgeotiff/build_tarballs.jl
@@ -2,7 +2,7 @@ using BinaryBuilder, Pkg
 
 name = "libgeotiff"
 upstream_version = v"1.7.1"
-version_offset = v"0.0.0"
+version_offset = v"0.0.1"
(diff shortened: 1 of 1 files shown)


visr:
I need to do PROJ_jll first, since that also has a libtiff dependency, as well as a geotiff dependency.
Then, this should be used:
```julia
dependencies = [
    Dependency("PROJ_jll"; compat="~900.200.100"),
    Dependency("LibCURL_jll"),
]
```

PR #6930, opened 2023-06-22
[PROJ] update to 9.2.1, build with libtiff 4.5

visr:
See #6929

### DIFF
--- a/P/PROJ/build_tarballs.jl
+++ b/P/PROJ/build_tarballs.jl
@@ -87,7 +87,7 @@ dependencies = [
     # Host SQLite needed to build proj.db
     HostBuildDependency("SQLite_jll")
     Dependency("SQLite_jll")
-    Dependency("Libtiff_jll"; compat="4.5")
+    Dependency("Libtiff_jll"; compat="4.5.1")
     Dependency("LibCURL_jll"; compat="7.73,8")
 ]
 
--- a/P/PROJ/build_tarballs.jl
+++ b/P/PROJ/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "PROJ"
-upstream_version = v"9.1.0"
+upstream_version = v"9.2.1"
 version_offset = v"0.0.0"
 version = VersionNumber(upstream_version.major * 100 + version_offset.major,
                         upstream_version.minor * 100 + version_offset.minor,
@@ -11,7 +11,8 @@ version = VersionNumber(upstream_version.major * 100 + version_offset.major,
 
 # Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://download.osgeo.org/proj/proj-$upstream_version.tar.gz", "81b2239b94cad0886222cde4f53cb49d34905aad2a1317244a0c30a553db2315")
+    ArchiveSource("https://download.osgeo.org/proj/proj-$upstream_version.tar.gz",
+        "15ebf4afa8744b9e6fccb5d571fc9f338dc3adcf99907d9e62d1af815d4971a1")
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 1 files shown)


P/PROJ/build_tarballs.jl
```diff
@@ -90,8 +87,8 @@ dependencies = [
     # Host SQLite needed to build proj.db
     HostBuildDependency("SQLite_jll")
     Dependency("SQLite_jll")
-    Dependency("Libtiff_jll")
-    Dependency("LibCURL_jll"; compat="7.73")
+    Dependency("Libtiff_jll"; compat="~4.5.1")
```
giordano, line 90: I think this is potentially overly restrictive, they didn't break the ABI from 4.1 to 4.3, they just break it in an unpredictable way.

Also, as usual this stuff has to be fixed first in the registry

P/PROJ/build_tarballs.jl
```diff
@@ -90,8 +87,8 @@ dependencies = [
...
+    Dependency("LibCURL_jll")
```
giordano, line 91: Curl must have colpa bounds: https://github.com/JuliaPackaging/Yggdrasil/issues/3024

P/PROJ/build_tarballs.jl
visr: Ok, is `compat="4.5"` good? Done in cd47fa86b2f47ca0d51472bbd9b9dd2e0770384b

P/PROJ/build_tarballs.jl
visr: Should I leave it as `compat="7.73"`, or is the `compat="7.73,8"` I set in cd47fa86b2f47ca0d51472bbd9b9dd2e0770384b also ok?

giordano: Well, you to allow v8 if you want to make this usable in newer versions

visr: Doesn't `compat="7.73,8"` allow v8? Not sure what to put here.

giordano: Yes, I was answering the previous question.

P/PROJ/build_tarballs.jl
```diff
@@ -90,8 +87,8 @@ dependencies = [
     # Host SQLite needed to build proj.db
     HostBuildDependency("SQLite_jll")
     Dependency("SQLite_jll")
-    Dependency("Libtiff_jll")
-    Dependency("LibCURL_jll"; compat="7.73")
+    Dependency("Libtiff_jll"; compat="4.5")
```
visr, line 90:
```suggestion
    Dependency("Libtiff_jll"; compat="4.5.1")
```
Apparently needed since 4.5.0 doesn't exist.

giordano:
Delete system libcurl on x86_64 musl, we often have to do that

P/PROJ/build_tarballs.jl
```diff
@@ -23,15 +24,18 @@ EXE_SQLITE3=${host_bindir}/sqlite3
...
     SQLITE3_LIBRARY=${libdir}/libsqlite3.${dlext}
     CURL_LIBRARY=${libdir}/libcurl.${dlext}
     TIFF_LIBRARY_RELEASE=${libdir}/libtiff.${dlext}
 fi
 
-if [[ "${target}" == x86_64-linux-musl* ]]; then
-    export LDFLAGS="-lcurl"
+if [[ ${target} == x86_64-linux-musl ]]; then
+    # ${libdir}/libcurl.so needs a libnghttp, and it prefers to load /usr/lib/libnghttp2.so for this.
+    # Unfortunately, that library is missing a symbol. Setting LD_LIBRARY_PATH is not enough to avoid this.
```
giordano, line 36: `LD_LIBRARY_PATH` is only for dynamic loading, if has nothing to do with linking, it isn't surpring it doesn't help anything here.

PR #6931, opened 2023-06-23
[SuiteSparse] Build Test (for branch with 32-bit SPQR)

Wimmerer:
One question is whether I should host this JLL here or on my own acct (or JuliaSparse). I don't think it should be registered? I'm not really sure.

Let's just see if it builds everywhere first.


Wimmerer:
What the heck is that error with cmake?!

S/SuiteSparse/SuiteSparse@Test/build_tarballs.jl
```diff
@@ -0,0 +1,118 @@
...
+    LibraryProduct("libklu",                 :libklu),
+    LibraryProduct("libumfpack",             :libumfpack),
+    LibraryProduct("librbio",                :librbio),
+    LibraryProduct("libspqr",                :libspqr),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("libblastrampoline_jll"; compat="5.4.0"),
+    BuildDependency("LLVMCompilerRT_jll",platforms=[Platform("x86_64", "linux"; sanitize="memory")]),
```
giordano, line 30: As a general remark, ***never*** make comparison with exact platforms, it's finicky and error-prone, check that the sanitizer of the platform is `memory`: https://github.com/JuliaPackaging/Yggdrasil/blob/c291c43414edd807000867a7c38796dc266b3dd9/O/OpenBLAS/common.jl#L245
```suggestion
    BuildDependency("LLVMCompilerRT_jll"; platforms=filter(p -> sanitize(p)=="memory", platforms)),
```

giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/c291c43414edd807000867a7c38796dc266b3dd9/O/OpenBLAS/common.jl#L2 (PR to export `sanitize` from `BinaryBuilder` welcome)

S/SuiteSparse/SuiteSparse@Test/build_tarballs.jl
```diff
@@ -0,0 +1,117 @@
...
+]
+
+
+sources = [GitSource("https://github.com/Wimmerer/SuiteSparse.git",
+"0e906339dc13abe30969a4b341d9208f41026adc")]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/SuiteSparse
+rm -rf /usr/bin/cmake
```
giordano, line 42:
```suggestion
apk rm cmake
```

Wimmerer:
@ViralBShah @giordano do we want to release this as a jll? Maybe it will be useful in the future too? 

Or now that it's building just point Julia's build process to the fork and don't use the jll until upstream is ready?

ViralBShah:
I would release it as a JLL. You can always move the SuiteSparse fork to JuliaSparse, but I don't see any reason to change the process. We can point to your fork, or alternatively, the cleaner option may be to just apply the whole thing as a patch to the existing SuiteSparse build.

ViralBShah:
Coming to think of it, a patch to the existing SuiteSparse release in Ygg seems best to me.

Wimmerer:
Our version numbers will be out of sync, and it's an ~8k line diff. But that's fine if that's what you want I can do that

giordano:
I think I'm missing some context. What's this test?

Wimmerer:
I've fixed SuiteSparse on 32-bit (I *think*). But it's not yet been accepted upstream, and we kind of need the fix asap. I don't want to just straight make it a new version because I might have missed something etc. So I'd like to use it for Julia CI.

giordano:
And patching suitesparse instead of having another package isn't an option?

Wimmerer:
That's fine, I am a little concerned about how to avoid it ever making it into a release if it's just a normal SuiteSparse version but that's probably unfounded.

ViralBShah:
Version number can just be what it is for SuiteSparse now, and with this patch it just bumps up the Ygg patch number for the new binaries. Julia builds can depend on the exact patch release.

ViralBShah:
@Wimmerer I notice that the SPQR 32-bit work is on top of SuiteSparse 7.1 (dev), and hence it may not be a good idea to generate a patch file for 7.0.1. I suppose the best thing would be to wait for SuiteSparse 7.1. Would you agree?

Wimmerer:
It's unclear whether this will make it into a 7.1 or 8.0. Hence my thought that this should be separate.

ViralBShah:
I meant generating a patch file against SuiteSparse 7.1. I know it won't make it - but we still need to generate a patch file that applies to a release version of SuiteSparse (so that we are not shipping SuiteSparse dev).

Wimmerer:
@ViralBShah the version I developed was against the **dev2** branch of SuiteSparse at Tim's request. I don't think the patch will necessarily map to 7.1 when it's released. I can give it a try, he might not have touched SPQR in dev2 before my PR. I will test the diffs tonight.

ViralBShah:
It doesn't appear he has touched SPQR from a quick look.

Wimmerer:
@ViralBShah @giordano Can I get a review and then merge on this? Then we can hopefully move forward and 🤞 backport

ViralBShah:
The problem with using 7.2.0 as the version number is that it will conflict with Tim's upstream version when he releases 7.2. Should we use 7.1.1 or something that is unlikely to conflict?

ViralBShah:
Looks good to me. Let's merge.

PR #6933, opened 2023-06-23
[XGBoost] Add cxxstring_abi back in, prevent augmenting Windows with a `cuda=none` tag

tylerjthomas9:
Ref: https://github.com/JuliaPackaging/Yggdrasil/pull/6746 and https://github.com/JuliaPackaging/Yggdrasil/issues/6911#issuecomment-1599350319

This PR accomplishes 2 things. The first is that it adds the `cxxstring_abi` back into the builds. I forgot to add it to the platform when porting the CUDA build format from AMGX and other libraries with multiple CUDA versions, where the `cxxstring_abi` is not added to the `augmented_platform`. 

The second thing is that I dropped the `cuda=none` tags on all Windows builds.

### DIFF
--- a/X/XGBoost/build_tarballs.jl
+++ b/X/XGBoost/build_tarballs.jl
@@ -89,20 +89,20 @@ platforms = expand_cxxstring_abis(supported_platforms())
 
 for cuda_version in versions_to_build, platform in platforms
 
-    build_cuda = (os(platform) == "linux") && (arch(platform) in ["x86_64"])
-    if !isnothing(cuda_version) && !build_cuda
+    cuda_platform = (os(platform) == "linux") && (arch(platform) in ["x86_64"])
+    if !isnothing(cuda_version) && !cuda_platform
         continue
     end
     
-    # For Windows, we want to avoid adding cuda=none
+    # For platforms we can't create cuda builds on, we want to avoid adding cuda=none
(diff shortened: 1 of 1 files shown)


X/XGBoost/build_tarballs.jl
```diff
@@ -93,10 +93,17 @@ for cuda_version in versions_to_build, platform in platforms
...
     end
-
-    augmented_platform = Platform(arch(platform), os(platform);
-        cuda=isnothing(cuda_version) ? "none" : CUDA.platform(cuda_version)
-    )
+    
+    # For Windows, we want to avoid adding cuda=none
+    # https://github.com/JuliaPackaging/Yggdrasil/issues/6911#issuecomment-1599350319
+    if os(platform) == "windows"
+        augmented_platform = deepcopy(platform)
```
giordano, line 100: I think you want to do this for all platforms for which you don't build CUDA, i.e. anything different from x86_64-linux-gnu?

tylerjthomas9: I am dropping the tags from all the other platforms where we can't build CUDA binaries.

giordano:
Uhm, now I don't see a job for `x86_64-linux-gnu-*-cuda+none`?

PR #6944, opened 2023-06-26
[VARNA] Add 'build' for VARNA-3.93 jar file

marcom:
Used in https://github.com/marcom/PlotRNA.jl

Having this as a jll avoids trying to load this at first package load time (it is cached though).

My main issue is that it is downloaded on every CI run, which might not be nice for the upstream server: https://github.com/marcom/PlotRNA.jl/issues/11.

### DIFF
--- a/V/VARNA/build_tarballs.jl
+++ b/V/VARNA/build_tarballs.jl
@@ -1,0 +1,34 @@
+using BinaryBuilder, Pkg
+
+name = "VARNA"
+version = v"3.93.0"
+
+# url = "https://varna.lisn.upsaclay.fr/"
+# description = "Java component and applet for drawing RNA secondary structure"
+
+sources = [
+    FileSource("https://varna.lisn.upsaclay.fr/bin/VARNAv$(version.major)-$(version.minor).jar",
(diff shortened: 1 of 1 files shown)


V/VARNA/build_tarballs.jl
```diff
@@ -0,0 +1,34 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir/
+
+install -Dvm 644 VARNA*.jar ${prefix}/share/VARNA.jar
+
+# VARNA uses GPL as stated here: https://varna.lisn.upsaclay.fr/
+install_license /usr/share/licenses/GPL-3.0+
+"""
+
+platforms = supported_platforms()
```
giordano, line 23: If anything, this should be `[AnyPlatform()]`, but I'm missing the point of this being a JLL

marcom: Ok I am using an artifact now. Wishing the artifacts system were as easy to use as BinaryBuilder.

marcom:
Ok I am using an artifact now. Wishing the artifacts system were as easy to use as BinaryBuilder.

PR #6948, opened 2023-06-27
[Qt6] wayland targets

barche:


Q/Qt6Wayland/build_tarballs.jl
```diff
@@ -87,4 +73,9 @@ end
 
 include("../../fancy_toys.jl")
 
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"9", julia_compat="1.6")
+init_block = raw"""
+ENV["QT_PLUGIN_PATH"] = qt6plugins_dir
+ENV["XKB_CONFIG_ROOT"] = joinpath(Qt6Base_jll.xkbcommon_jll.Xorg_xkeyboard_config_jll.artifact_dir, "share", "X11", "xkb")
```
barche, line 78: @giordano Is this OK here or should we add it to `Xorg_xkeyboard_config_jll`?

giordano: If that's needed by all packages using `Xorg_xkeyboard_config_jll` then I guess it'd make more sense there.

Sometimes I prefer doing something like
```julia
ENV["ABC"] = get(ENV, "ABC", ...)
```
just in case that's already set

barche: OK, I'll add it there. I don't know if it's *always* needed, but the error I got has manifested before (and you fixed it ;) https://github.com/JuliaGraphics/Gtk.jl/issues/469

PR #6950, opened 2023-06-27
bump Gettext to upstream version 0.22

stevengj:
Also, as discussed in https://github.com/Julia-i18n/Gettext.jl/pull/5, export a `libintl` for the `libintl` library, which is probably what callers want to link to.

(Also removes the mingw patch, which seems to have been incorporated upstream.)


stevengj:
Hmm, it looks like `libintl` doesn't get installed if you are on a system that has `gettext` in `libc` (e.g. Linux systems).   One might be able to bypass this by building via `--with-included-gettext`.

stevengj:
Grrr, mingw build is failing with
```
 .libs/libgettextsrc_la-write-catalog.o:write-catalog.c:(.text+0x7bd): undefined reference to `close_used_without_requesting_gnulib_module_close'
```

giordano:
Uhm, try with a more recent gcc (but [don't go beyond gcc 10](https://github.com/JuliaPackaging/Yggdrasil/blob/6197573cc772a0711cb356ec227480acb09fa77c/CONTRIBUTING.md#compatibility-tips)). Sounds like it may be an old bug in binutils, which we bundle with gcc: https://github.com/JuliaPackaging/Yggdrasil/blob/6197573cc772a0711cb356ec227480acb09fa77c/RootFS.md#compiler-shards

stevengj:
I tried requiring gcc 10, but it still seems to be failing?

fingolfin:
Some additional potentially relevant discussion is on #7267

ViralBShah:
Closing in favor of #8289

PR #6949, opened 2023-06-27
[ViennaRNA] Build script for historic version ViennaRNA-1.8.5

marcom:
This is a build script for ViennaRNA-1.8.5 (from 2011), the last version of ViennaRNA-1.x.y.

I need it for some old projects of mine to make them easily reproducible.

I have placed the script in an extra directory so as not to mess up the build script for the current ViennaRNA-2.x.y.

PR #6951, opened 2023-06-28
[ViennaRNA] Take two for build script for ViennaRNA-1.8.5

marcom:
Supersedes: #6949 

This is a build script for ViennaRNA-1.8.5 (from 2011), the last version of ViennaRNA-1.x.y.

I need it for some old projects of mine to make them easily reproducible.

I have placed the script in an extra directory so as not to mess up the build script for the current ViennaRNA-2.x.y.


marcom:
Ok this time the right version (ViennaRNA-1.8.5) got built. So now I know what the `@` sign in the subdir names is for :)

PR #6952, opened 2023-06-28
[argtable] add recipe

M-PERSIC:
[libargtable2](https://sourceforge.net/projects/argtable/) v2.13 and [libargtable3](https://github.com/argtable/argtable3) v3.2.2.

PR #6953, opened 2023-06-28
[ClustalO] add recipe

M-PERSIC:
[Clustal Omega](http://www.clustal.org/omega/) v1.2.4. Requires libargtable2 (#6952)


M-PERSIC:
Do not merge

PR #6955, opened 2023-06-28
[Xorg_xkeyboard_config] Initialize XKB_CONFIG_ROOT [skip build]

barche:
I also had to add `julia_compat` because of an error about experimental platforms, not sure if the `skip build` is still OK then @giordano ?

### DIFF
--- a/X/Xorg_xkeyboard_config/build_tarballs.jl
+++ b/X/Xorg_xkeyboard_config/build_tarballs.jl
@@ -34,5 +34,9 @@ dependencies = [
     Dependency("Xorg_xkbcomp_jll"),
 ]
 
+init_block = raw"""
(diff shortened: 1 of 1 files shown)


giordano:
> not sure if the skip build is still OK then

Not only `[skip build]` is not ok, as you're building for more platforms than before you ***must*** do a proper new build, but you also ***must*** change the version number: #2763.

barche:
OK, so I think I'm going to bite the bullet and update this one and its dependencies first. Any objections to that? It seems most of the X11-related libs are affected, I have no idea if upgrading these will break binary compatibility, as it seems no compat bounds are set?

The other alternative is that I just filter out the experimental platform here (only armv6l-linux actually) and remove the `julia_compat` again.

PR #6960, opened 2023-06-29
New Recipe: libqasm v0.4.0

mofeing:
I'm having some problems with `flex` and `bison` on other platforms, so I don't think it's ready to merge yet.

I will post some stacktraces when I figure out more what is happening.

### DIFF
--- a/L/libqasm/build_tarballs.jl
+++ b/L/libqasm/build_tarballs.jl
@@ -1,0 +1,51 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "libqasm"
+version = v"0.4.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/QuTech-Delft/libqasm.git", "1f317077e2f2462cafec8385666d6e996340d2e8")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+mkdir -p libqasm/build
+cd libqasm/build/
+rm /usr/share/cmake/Modules/Compiler/._*
+rm /usr/share/cmake/Modules/CompilerId/._*
+rm /usr/share/cmake/Modules/._*
+git submodule init
+git submodule update
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DLIBQASM_COMPAT=ON -DBUILD_SHARED_LIBS=ON -DLIBQASM_BUILD_PYTHON=OFF ..
+export PATH=$PATH:/workspace/srcdir/libqasm/build/src/cqasm/tree-gen/:/workspace/srcdir/libqasm/build/src/cqasm/func-gen/
+make -j
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "musl")
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libcqasm", :libcqasm)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CMake_jll", uuid="3f4e10e2-61f2-5801-8945-23b9d642d0e6"))
+    Dependency(PackageSpec(name="Bison_jll", uuid="0f48145f-aea8-549d-8864-7f251ac1e6d0"))
+    Dependency(PackageSpec(name="flex_jll", uuid="48a596b8-cc7a-5e48-b182-65f75e8595d0"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"5.2.0")


L/libqasm/build_tarballs.jl
```diff
@@ -0,0 +1,51 @@
...
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libcqasm", :libcqasm)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CMake_jll", uuid="3f4e10e2-61f2-5801-8945-23b9d642d0e6"))
+    Dependency(PackageSpec(name="Bison_jll", uuid="0f48145f-aea8-549d-8864-7f251ac1e6d0"))
+    Dependency(PackageSpec(name="flex_jll", uuid="48a596b8-cc7a-5e48-b182-65f75e8595d0"))
```
giordano, line 47: These are all host dependencies: https://docs.binarybuilder.org/stable/build_tips/#Dependencies-for-the-target-system-vs-host-system.  Also, cmake is already there, unless you need a specific version (in which case you should specify it)

L/libqasm/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "musl")
+]
```
giordano, line 35: Now you can try to build for all platforms 🙂

mofeing:
Super! Thank you so much @giordano 

I'm putting the PR as "Ready to review" so in case all test succeed, then merge.

L/libqasm/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DLIBQASM_COMPAT=ON -DBUILD_SHARED_LIBS=ON -DLIBQASM_BUILD_PYTHON=OFF ..
+export PATH=$PATH:/workspace/srcdir/libqasm/build/src/cqasm/tree-gen/:/workspace/srcdir/libqasm/build/src/cqasm/func-gen/
+make -j
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
```
giordano, line 32: From the audit:
```suggestion
platforms = expand_cxxstring_abis(platforms)
```

mofeing:
mmm actually the library requires C++11. Do we have to support `cxx03`?

giordano:
https://docs.binarybuilder.org/stable/building/#Expanding-C-string-ABIs-or-libgfortran-versions

> This ABI does not have to do with the C++ standard used by the source code, in fact you can build a C++03 library with the C++11 std::string ABI and a C++11 library with the C++03 std::string ABI. This is achieved by appropriately setting the _GLIBCXX_USE_CXX11_ABI macro.

mofeing:
> https://docs.binarybuilder.org/stable/building/#Expanding-C-string-ABIs-or-libgfortran-versions
> 
> > This ABI does not have to do with the C++ standard used by the source code, in fact you can build a C++03 library with the C++11 std::string ABI and a C++11 library with the C++03 std::string ABI. This is achieved by appropriately setting the _GLIBCXX_USE_CXX11_ABI macro.

Ah right, now I remember reading this time ago.

giordano:
https://docs.binarybuilder.org/stable/troubleshooting/#Running-foreign-executables

L/libqasm/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+cd $WORKSPACE/srcdir
+mkdir -p libqasm/build
+cd libqasm/build/
+# rm /usr/share/cmake/Modules/Compiler/._*
+# rm /usr/share/cmake/Modules/CompilerId/._*
+# rm /usr/share/cmake/Modules/._*
+git submodule init
+git submodule update
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DLIBQASM_COMPAT=ON -DBUILD_SHARED_LIBS=ON -DLIBQASM_BUILD_PYTHON=OFF ..
+export PATH=$PATH:/workspace/srcdir/libqasm/build/src/cqasm/tree-gen/:/workspace/srcdir/libqasm/build/src/cqasm/func-gen/
```
giordano, line 24: You can't do this, you're trying to run foreign executables.  You have to do a double pass with bootstrap, where you first build the executable for the host, then use _that_ one to compile the target one.

L/libqasm/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+make -j
```
giordano, line 25: Please specify `-j{nproc}`, `${nproc}` is set to a value which doesn't oversubscribe the machine.

L/libqasm/build_tarballs.jl
mofeing: Okay, is there any example of this?

giordano: LLVM 😂 You basically have to build first the executables you need with the host cmake toolchain file, then you install in hostprefix (so that it's automatically in PATH), then you do the build of the package.

Seriously, the problem is that cmake doesn't allow you to specify both host and target toolchains, one has to deal this kind of bootstrap processes manually, which is rather annoying. Any other saner build system (autoconf, meson) allows bootstrapping because it supports double toolchains.

mofeing: Seems like I'm gonna spend some days patching... Every year I ask myself the same question: "Why the hell I was in love with C++ and CMake?" 🤦

I'm gonna be a couple of days out so I will return to this next week.

Thanks @giordano

fingolfin:
@mofeing hi there, is still something you are interested in completing?

mofeing:
No, it can be closed. I'm no longer interested in this recipe.

PR #6962, opened 2023-06-30
[libjulia] bump nightly version

benlorenz:
The soname for libjulia was changed to libjulia.so.MAJOR.MINOR in https://github.com/JuliaLang/julia/pull/49012, now libcxxwrap and everything that depends on it now fails on nightly with: 
```
libjulia.so.1: cannot open shared object file: No such file or directory
```
Lets see if it still builds.

cc: @fingolfin @barche

### DIFF
--- a/L/libjulia/build_tarballs.jl
+++ b/L/libjulia/build_tarballs.jl
@@ -31,7 +31,7 @@ function Pkg.Types.is_stdlib(uuid::Base.UUID, julia_version::VersionNumber)
     return false
 end
 
-jllversion=v"1.10.2"
+jllversion=v"1.10.3"
 build_julia(ARGS, v"1.6.3"; jllversion)
 build_julia(ARGS, v"1.7.0"; jllversion)
 build_julia(ARGS, v"1.8.2"; jllversion)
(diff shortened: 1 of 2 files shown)


giordano:
> The soname for libjulia was changed to libjulia.so.MAJOR.MINOR in JuliaLang/julia#49012

That sounds like a bad idea

giordano:
It'd force us to have a libjulia for each single patch version, it's just unsustainable

giordano:
> I don't really understand, it does not contain the patch version in the soname, just 1.10 and we do builds for every minor version anyway?

Ah, sorry, I read too quickly, but I always thought the soname was libjulia.so.MAJOR.MINOR, wasn't it?

giordano:
Uh, ok, yeah, since ABI _is_ often broken between minor versions then the new soname makes more sense, I just somehow assumed it was that one already, hence my confusion.

barche:
I can confirm that simply rebuilding `libcxxwrap-julia` with this change makes the CxxWrap tests pass. Is there a way to rebuild the binaries without committing some meaningless change (e.g. add a comment) in the `build_tarballs.jl` to trigger the CI??

PR #6971, opened 2023-06-30
[Xorg_xcb_proto] Version 1.15

barche:


giordano:
Proto packages should target `[AnyPlatform()]`, which was probably introduced after this package was first created

barche:
OK, I suppose we can leave this one as-is, and I'll apply this to the other proto packages?

PR #6977, opened 2023-07-01
[libcifpp] Add build script for libcifpp v5.1.0

marcom:
https://github.com/PDB-REDO/libcifpp

C++ library for reading and manipulating mmCIF and PDB files

This is a build requirement for DSSP: https://github.com/PDB-REDO/dssp

Note: the build wants to download components.cif, I have disabled this for the time being as the file is quite large (380 MB decompressed). I don't know what functionality is impacted by this, but the test suite passes at least. We should probably have a better way of sharing components.cif between different packages that need it.

cc: @lmiq

### DIFF
--- a/L/libcifpp/build_tarballs.jl
+++ b/L/libcifpp/build_tarballs.jl
@@ -14,11 +14,6 @@ sources = [
 ]
 
 # TODO
-# - the installed file etc/cron.weekly/update-libcifpp-data contains
-#   hardcoded paths with /workspace/destdir/...
-#   It is used to update components.cif, which should probably be done
-#   in a different way.
-#
 # - the CCD (chemical components dictionary) normally gets downloaded
 #   by default and saved under share/libcifpp/components.cif, but we
 #   don't do this
--- a/L/libcifpp/build_tarballs.jl
+++ b/L/libcifpp/build_tarballs.jl
@@ -117,8 +117,7 @@ install_license ../LICENSE
 """
 
 platforms = supported_platforms()
-# Only FreeBSD uses clang in this build
-platforms = expand_cxxstring_abis(platforms; skip=Sys.isfreebsd)
+platforms = expand_cxxstring_abis(platforms)
 
 products = [
     LibraryProduct("libcifpp", :libcifpp),
--- a/L/libcifpp/build_tarballs.jl
+++ b/L/libcifpp/build_tarballs.jl
@@ -113,4 +113,4 @@ dependencies = [
 ]
 
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat="1.6", preferred_gcc_version = v"10")
+               julia_compat="1.6", preferred_gcc_version = v"9")
--- a/L/libcifpp/build_tarballs.jl
+++ b/L/libcifpp/build_tarballs.jl
@@ -69,6 +69,11 @@ if [[ "${bb_full_target}" == "${MACHTYPE_FULL}" ]]; then
     CFG_TESTING="-DENABLE_TESTING=ON"
 fi
 
+# use gcc/g++ on apple
+if [[ "${target}" == *-apple-darwin* ]]; then
+    CMAKE_TARGET_TOOLCHAIN="$(dirname "${CMAKE_TARGET_TOOLCHAIN}")/target_${target}_gcc.cmake"
+fi
+
 cmake .. \
     -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
@@ -93,7 +98,8 @@ install_license ../LICENSE
 """
 
 platforms = supported_platforms()
-platforms = expand_cxxstring_abis(platforms)
+# Only FreeBSD uses clang in this build
+platforms = expand_cxxstring_abis(platforms; skip=Sys.isfreebsd)
 
 products = [
     LibraryProduct("libcifpp", :libcifpp),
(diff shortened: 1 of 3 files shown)


L/libcifpp/build_tarballs.jl
```diff
@@ -0,0 +1,93 @@
...
+
+platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
+
+products = [
+    LibraryProduct("libcifpp", :libcifpp),
+]
+
+dependencies = [
+    Dependency("Eigen_jll"),
```
giordano, line 86: Eigen is header-only, no?

giordano: The point of the above question was that in that case Eigen should be a [`BuildDependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.BuildDependency), as that's not needed at runtime.

marcom: Yes thanks, that's what I understood, sorry for the miscommunication.

marcom:
Weird that it compiled with clang on freebsd, and failed on apple with the same compiler. I am using g++ now on apple and it seems to compile.

This issue https://github.com/PDB-REDO/libcifpp/issues/39 seems to indicate that maybe the build should use gcc 10 or higher.

Was there a minimum gcc version one has to use on aarch64-apple?

marcom:
No idea why changing `preferred_gcc_version` from 9 to 10 is causing the freebsd build to fail. It's using clang, but suddenly the build has the same compile errors (relating to things new in c++20) that before the clang build on apple had...

L/libcifpp/build_tarballs.jl
```diff
@@ -0,0 +1,116 @@
...
+
+CFG_TESTING="-DENABLE_TESTING=OFF"
+if [[ "${bb_full_target}" == "${MACHTYPE_FULL}" ]]; then
+    # build the tests if we are building for the build host platform
+    CFG_TESTING="-DENABLE_TESTING=ON"
+fi
+
+# use gcc/g++ on apple
+if [[ "${target}" == *-apple-darwin* ]]; then
+    CMAKE_TARGET_TOOLCHAIN="$(dirname "${CMAKE_TARGET_TOOLCHAIN}")/target_${target}_gcc.cmake"
```
giordano, line 65: https://docs.binarybuilder.org/stable/build_tips/#Using-GCC-on-macOS-and-FreeBSD
```suggestion
    CMAKE_TARGET_TOOLCHAIN=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake
```
Side note, it's a bit sad to switch to gcc on BSD systems especially for C++ libraries, as GCC has a different runtime than llvm, and compatibility with other libraries is untested at best (but likely to be highly problematic)

L/libcifpp/bundled/patches/mingw-no-ioctl.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/src/utilities.cpp b/src/utilities.cpp
+index 5238a5b..f5e54a8 100644
+--- a/src/utilities.cpp
++++ b/src/utilities.cpp
+@@ -43,7 +43,7 @@
+ #include <sstream>
+ #include <thread>
+ 
+-#if not defined(_MSC_VER)
++#if not (defined(_MSC_VER) || defined(__MINGW32__))
```
giordano, line 10: More likely this should be `_WIN32`: https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

L/libcifpp/bundled/patches/mingw-no-ioctl.patch
```diff
@@ -0,0 +1,25 @@
...
+ #include <sys/ioctl.h>
+ #include <termios.h>
+ #endif
+@@ -68,9 +68,9 @@ std::string get_version_nr()
+ 
+ // --------------------------------------------------------------------
+ 
+-#ifdef _MSC_VER
++#if defined(_MSC_VER) || defined(__MINGW32__)
```
giordano, line 19: Same.

L/libcifpp/bundled/patches/mingw-no-ioctl.patch
giordano, line 1: This would be good to upstream

L/libcifpp/build_tarballs.jl
marcom: Ok I will try again with clang on apple, those warnings in the build with gcc also didn't look like it would work reliably.

L/libcifpp/bundled/patches/mingw-no-ioctl.patch
marcom: https://github.com/PDB-REDO/libcifpp/pull/45

marcom:
So I have done some investigating into the compile failures on apple with clang.

It builds fine on Ubuntu x86_64 clang++-13 (13.0.1-2ubuntu2.1), but running the failing compile line with `-H` on both Ubuntu and in BinaryBuilder shows that different C++ libraries are included: on ubuntu clang++-13 uses 
the GNU c++ library, and in BinaryBuilder it uses the llvm c++ library.

In BinaryBuilder, the header for `std::set` is `/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root/usr/include/c++/v1/set`, and indeed it does not have the `contains` member function.

L/libcifpp/build_tarballs.jl
```diff
@@ -0,0 +1,135 @@
...
+fi
+
+make install
+
+install_license ../LICENSE
+"""
+
+platforms = supported_platforms()
+# Only FreeBSD uses clang in this build
+platforms = expand_cxxstring_abis(platforms; skip=Sys.isfreebsd)
```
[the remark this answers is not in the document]
marcom: Thanks, I forgot to change that back from when g++ was used on macos

PR #6980, opened 2023-07-01
[Xorg_xkeyboard_config] Initialize XKB_CONFIG_ROOT

barche:


barche:
This should make sure downstream packages such as Gtk.jl don't need to set this. See https://github.com/JuliaGraphics/Gtk.jl/pull/470

X/Xorg_xkeyboard_config/build_tarballs.jl
```diff
@@ -34,5 +35,9 @@ dependencies = [
     Dependency("Xorg_xkbcomp_jll"),
 ]
 
+init_block = raw"""
+ENV["XKB_CONFIG_ROOT"] = get(ENV, "XKB_CONFIG_ROOT", joinpath(artifact_dir, "share", "X11", "xkb"))
```
giordano, line 39: Need `if Sys.islinux() || Sys.isfreebsd()`?  Especially now that this package became `AnyPlatform`

barche: OK, done!

X/Xorg_xkeyboard_config/build_tarballs.jl
```diff
@@ -34,5 +35,11 @@ dependencies = [
...
 
+init_block = raw"""
+if Sys.islinux() || Sys.isfreebsd()
+    ENV["XKB_CONFIG_ROOT"] = get(ENV, "XKB_CONFIG_ROOT", joinpath(artifact_dir, "share", "X11", "xkb"))
+end
+"""
+
 # Build the tarballs.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; init_block)
```
giordano, line 45: AnyPlatform doesn't require changing Julia compat

giordano: Ah, sorry, misread the line!

giordano: Although there's actually a benefit in requiring Julia 1.6: we can safely drop Pkg from the dependencies of the package.

barche: Good to merge I think?

barche: ah, missed that last comment. Want me to add Julia 1.6 compat and remove the Pkg dep?

giordano: Yeah, even if not strictly required, dropping Pkg from dependencies is a good thing

PR #6985, opened 2023-07-02
[ecCodes] Update ecCodes to 2.28.0 and add libaec

benelsen:
ECMWF recently switched to AEC compression in their grib2 files with cycle 48r1. This means the current version provided by eccodes_jll can't read these grib files anymore.

Thankfully we now have [libaec_jll](https://github.com/JuliaBinaryWrappers/libaec_jll.jl) so we can just enable AEC here.

I tried bumping the version as well, but building the tarballs locally the latest version I managed to build on all platforms without further changes was 2.28.0. This might might provide a good stopgap measure for everyone using GRIB.jl etc with ECMWF data before moving further. Version 2.30.2 is available but has dropped 32 bit support requiring dropping some platforms which I didn't want to do without some feedback first.

### DIFF
--- a/E/eccodes/build_tarballs.jl
+++ b/E/eccodes/build_tarballs.jl
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "eccodes"
-version = v"2.25.0"
+version = v"2.28.0"
 
 # Collection of sources required to complete build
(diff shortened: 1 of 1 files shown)


giordano:
Thanks for your contribution!

Issue #455, opened 2020-02-01
Cannot initialize Ncurses in macOS


ronisbr:
Hi!

For some reason, I cannot initialize Ncurses using `Ncurses_jll` in macOS.

```julia
julia> using Ncurses_jll

julia> ccall( (:initscr, libncurses), Cvoid, ())
Error opening terminal: xterm-256color.
```

After that, julia crashes. Passing the env. option `TERM=xterm` did not help also.

I can start Ncurses normally if I use the library that ships with macOS or the library in home-brew.

giordano:
You probably need to set the environment variable `TERMINFINO_DIRS`:
```julia
julia> using Ncurses_jll

julia> withenv("TERMINFO_DIRS" => joinpath(Ncurses_jll.artifact_dir, "share", "terminfo")) do
           ccall( (:initscr, libncurses), Cvoid, ())
       end
```

giordano:
Also `TERMINFO` should work.  In any case, it's better to wrap the calls in a `withenv` to avoid setting this variable for the whole Julia session

ronisbr:
This work! However, I need to set for the whole Julia session, otherwise I will have to do a strange logic because I need to let the user to load its own ncurses version. Is there any downside?

giordano:
> This work!

Great, then I'm going to close the ticket.

> Is there any downside?

The only problem can arise if there are other, different instances of ncurses loaded in the same session, that might be wrecked by this setting.  The probability of this situation is likely very low.  If all Julia packages needing ncurses use `Ncurses_jll` there will be no problem, of course.

You're free to set the variable with `ENV["TERMINFO"]`, just be aware that if someone will complain about `TextUserInterfaces.jl` breaking other packages that use ncurses, that's likely the issue.

ronisbr:
Thanks very much for the quick help!

PR #6994, opened 2023-07-03
[ncurses] set `TERMINFO_DIRS` environment variable

fingolfin:
See https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/687

The recent change by @barche in PR #6980 reminded me about this.

(I should have done it with my recent ncurses 6.4 update *sigh* but since ncurses only seems to use versions of the form MAJOR.MINOR, it should be OK to use this fake version)

### DIFF
--- a/N/Ncurses/build_tarballs.jl
+++ b/N/Ncurses/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder
 
 name = "Ncurses"
-version = v"6.4"
+version = v"6.4.1" # <-- This version is a lie, to add the TERMINFO_DIRS override
 
 # Collection of sources required to build Ncurses
 sources = [
@@ -88,5 +88,17 @@ dependencies = [
     HostBuildDependency("Ncurses_jll"),
 ]
 
+init_block = raw"""
+if Sys.isunix()
+    path = joinpath(artifact_dir, "share", "terminfo")
+    old = get(ENV, "TERMINFO_DIRS", nothing)
+    if old === nothing
+        ENV["TERMINFO_DIRS"] = path
+    else
+        ENV["TERMINFO_DIRS"] = old * ":" * path
+    end
+end
+"""
+
 # Build the tarballs.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", init_block)


N/Ncurses/build_tarballs.jl
```diff
@@ -88,5 +88,11 @@ dependencies = [
     HostBuildDependency("Ncurses_jll"),
 ]
 
+init_block = raw"""
+if Sys.isunix()
```
fingolfin, line 92: I am not even sure we need this... Adapted it from `Xorg_xkeyboard_config`

N/Ncurses/build_tarballs.jl
```diff
@@ -88,5 +88,11 @@ dependencies = [
...
+    ENV["TERMINFO"] = get(ENV, "TERMINFO", joinpath(Ncurses_jll.artifact_dir, "share", "terminfo"))
```
fingolfin, line 93: This I also adapted from `Xorg_xkeyboard_config` -- it only sets `TERMINFO` to point at the artifact if it wasn't already set. So the user can still set their own `TERMINFO`. 

However, looking at `man terminfo`, it seems the precedence is `TERMINFO > $HOME/.terminfo > TERMINFO_DIRS` so perhaps I should set `TERMINFO_DIRS` here instead?

giordano: https://github.com/JuliaPackaging/Yggdrasil/issues/455#issuecomment-581026705

fingolfin: I am aware of that issue, and in fact I am using the `withenv` workaround myself in `GAP.jl`. However, it is limited: I have code paths the user triggers which could also in the end need this. It is not practical to repeat the `withenv` in all code. So I think addressing this once and forall here makes sense.

Of course there is a hypothetical chance we'll break something. But that's true for almost any change we make. We'll deal with it if we get reports about it? In the worst case we can revert this change here -- of course we'll then (re-)break other things, but such is life.... ?

fingolfin: Another option is that I just set the variable globally in `GAP.jl`, then it'll be limited in which code the change affects -- so breakage still is possible, but minimized *shrug*

giordano: Yeah, especially for `ccall`s into library wrapping every single call with withenv is annoying.

giordano (approved):

N/Ncurses/build_tarballs.jl
```diff
@@ -88,5 +88,11 @@ dependencies = [
     HostBuildDependency("Ncurses_jll"),
 ]
 
+init_block = raw"""
+if Sys.isunix()
+    ENV["TERMINFO_DIRS"] = get(ENV, "TERMINFO_DIRS", joinpath(artifact_dir, "share", "terminfo"))
```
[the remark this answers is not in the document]
fingolfin: That sounds like a good idea.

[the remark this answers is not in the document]
fingolfin: done

giordano:
I just realised this PR makes https://github.com/giordano/Htop.jl redundant, as it was just setting TERMINFO_DIRS, I think I'll delete the repo, `Htop_jll` is enough now.  Thanks!

PR #6995, opened 2023-07-03
Update Libxc to 6.2.2

mfherbst:


ViralBShah:
#9910 is a more recent attempt

PR #6998, opened 2023-07-03
[DSSP] Build script for DSSP v4.3.1

marcom:
https://github.com/PDB-REDO/dssp

Application to assign secondary structure to proteins 

cc: @lmiq


marcom:
So something is wrong with the `libcifpp_jll` build on windows, it installs a tiny static library that also lacks most/all code/symbols. Removing that (see #7001) is not enough for DSSP to build, as the cmake files `libcifpp_jll` installs still point to the shared library.

In order to have at least a working DSSP binary `mkdssp`, I have now resorted to building libcifpp statically in this build, and that seems to be working.

The `make install` of the libcifpp build inside this script has installed quite a few files, should those be removed so as to not conflict with the ones installed by `libcifpp_jll`?

marcom:
Update: I'm not so sure anymore there is anything wrong with the `libcifpp_jll` build, the `.dll.a` is something completely normal on windows called an import library which I unfortunately had no idea even existed.

I still would like to just link dssp statically to libcifpp as done in this build for the time being though, as that way we at least have a working mkdssp program to work with.

D/DSSP/build_tarballs.jl
```diff
@@ -0,0 +1,170 @@
...
+# url = "https://github.com/PDB-REDO/dssp"
+# description = "Application to assign secondary structure to proteins"
+
+sources = [
+    GitSource("https://github.com/PDB-REDO/dssp",
+              "b87ef206a071e6f086c8dc01551afd5e9b23eb43"),
+    GitSource("https://github.com/mhekkel/libmcfp",
+              "4aa95505ded43e663fd9dae61c49b08fdc6cce0c"), # v1.2.4
+    GitSource("https://github.com/PDB-REDO/libcifpp",
+              "836aed6ea9a227b37e5b0d9cbcb1253f545d0778"), # v5.1.0 (git-tag v5.1.0.1)
```
giordano, line 16: Why these aren't three distinct packages?

marcom: `libmcfp` is a small header-only package for command-line parsing, but I could make it separate. Should I do this?

`libcifpp_jll` already is a separate package, but there are linking errors on windows when linking dssp to the shared library that `libcifpp_jll` produces. https://buildkite.com/julialang/yggdrasil/builds/3927#01892115-4b4a-4314-87d8-36265c992572/6-8351

The default upstream compilation setup for libcifpp is to produce a static library, and in this case everything works.

From my investigations it seems that something is wrong with the symbol visibility of the `libcifpp_jll` shared library on windows. Under windows there is an [import library](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-creation#creating-an-import-library) that gets created. In `libcifpp_jll`, this is `lib/libcifpp.dll.a`, which only has a few symbols defined. If I try to re-run the failing command manually in the BB shell and link to `bin/libcifpp.dll` I still get linking errors. I currently don't know enough about windows/mingw to fix this and also lack the time to investigate. So building libcifpp here and linking to it statically is a way to get dssp working which is the main goal.

If it's ok to leave it building libcifpp here and linking statically to it, should the files installed by the libcifpp `make install` be removed?

giordano: > libmcfp is a small header-only package for command-line parsing, but I could make it separate. Should I do this?

Might be nicer, yes, especially if there's a chance other packages may need it.  This would be used as a build dependency of course.

marcom: What about this part:

> If it's ok to leave it building libcifpp here and linking statically to it, should the files installed by the libcifpp make install be removed?

Maybe let libcifpp in this build install in a non-destdir directory, then nothing should get packaged up in the final tarball?

marcom: Ok, all ready to merge from my side

PR #7000, opened 2023-07-04
[GAP] rebuild for updated libjulia

fingolfin:
This PR is "too early" as `libjulia_jll` 1.10.4 did not yet complete
its release cycle. But I thought I'd use the opportunity to verify
that the version specification for the build dependency works as
expect (i.e., makes the CI for this PR fail).

### DIFF
--- a/G/GAP/build_tarballs.jl
+++ b/G/GAP/build_tarballs.jl
@@ -29,7 +29,7 @@ name = "GAP"
 upstream_version = v"4.12.2"
 version = v"400.1200.200"
 
-julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10"]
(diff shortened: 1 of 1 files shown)


fingolfin:
CC @benlorenz

fingolfin:
Maybe the `julia_versions` list is also something that should be defined in `libjulia/common.jl` -- most packages just use the full list. Those which don't wish to do that could filter out versions that are not supported -- usually just old ones, I'd guess (and in some cases I suspect there is no reason at all why some Julia versions are excluded, just accidents?)

fingolfin:
Had to separate this from the `GAP_pkg_juliainterface` update because that one first requires a `GAP_jll` with Julia 1.11 support...

PR #7001, opened 2023-07-04
[libcifpp] Remove non-functional static library only built on windows

marcom:
Even though the build specifies to build shared libraries to cmake via `-DBUILD_SHARED_LIBS=ON`, on `*-w64-mingw32` a small non-functional static library gets built and installed.

This can mess up other builds that then try to link against this static library, e.g. `DSSP_jll`, as the static library is non-functional and lacks most symbols the shared library has.

This should hopefully help with https://github.com/JuliaPackaging/Yggdrasil/pull/6998

### DIFF
--- a/L/libcifpp/build_tarballs.jl
+++ b/L/libcifpp/build_tarballs.jl
@@ -113,6 +113,14 @@ fi
 
 make install
 
+if [[ "${target}" == *-w64-mingw* ]]; then
+    # libcifpp v5.1.0
+    # On windows, there is a non-functional static library that also gets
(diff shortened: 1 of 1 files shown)


marcom:
There still seems to be something wrong with the build on windows, I think the installed cmake files still refer to the static library (that is broken).

marcom:
Ok I just found out that a `.dll.a` file (mingw) or `.lib` file (windows) is called an import library and is something completely normal, and is therefore probably not the cause of the linking errors in dssp...

PR #7002, opened 2023-07-04
[Geant4_julia] New version of Geant4 wrappers

peremato:
Added new solids


giordano:
As a general remark working with Git and GitHub, opening a PR from a branch with the same name as the target branch is an [anti-pattern](https://blog.jasonmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/). You should always create a ***new*** branch. This is the reason why the history of this PR has completely unrelated commits, which make review slightly confusing.

peremato:
Thanks. I'll try not to forget.

PR #7003, opened 2023-07-04
[Exodus] update to 8.19.1; update compat for hdf5

cmhamel:
…king changes MBedTLS_jll). Also adding a hack to the build process to fix a tree hash mismatch caused on mac with a comment describing the hack in the script.

### DIFF
--- a/E/Exodus/build_tarballs.jl
+++ b/E/Exodus/build_tarballs.jl
@@ -73,7 +73,7 @@ make install
 #
 # The build process has far too many CMake files to track this down.
 #
-rm -r lib/cmake/Seacas/
+rm -r "${prefix}/lib/cmake/Seacas"
 """
 
 # These are the platforms we will build for by default, unless further
--- a/E/Exodus/build_tarballs.jl
+++ b/E/Exodus/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "Exodus"
-version = v"8.19.0"
+version = v"8.19.1"
 
 # Collection of sources required to complete build
 sources = [
--- a/E/Exodus/build_tarballs.jl
+++ b/E/Exodus/build_tarballs.jl
@@ -73,7 +73,6 @@ make install
 #
 # The build process has far too many CMake files to track this down.
 #
-cd $WORKSPACE/destdir
 rm -r lib/cmake/Seacas/
 """
 
--- a/E/Exodus/build_tarballs.jl
+++ b/E/Exodus/build_tarballs.jl
@@ -64,6 +64,18 @@ cmake \
 
 make -j${nproc}
 make install
+
+# below is an absolute hack to fix a tree hash mismatch on macos
+# this is due to a case insensitivity issue. 
+#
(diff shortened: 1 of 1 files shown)


E/Exodus/build_tarballs.jl
```diff
@@ -92,7 +104,7 @@ products = [
 dependencies = [
     Dependency(PackageSpec(name="Fmt_jll", uuid="5dc1e892-f187-50dd-85f3-7dff85c47fc5"))
     # Updating to a newer HDF5 version is likely possible without problems but requires rebuilding this package
-    Dependency(PackageSpec(name="HDF5_jll", uuid="0234f1f7-429e-5d53-9886-15a909be8d59"); compat="~1.12")
+    Dependency(PackageSpec(name="HDF5_jll", uuid="0234f1f7-429e-5d53-9886-15a909be8d59"); compat="~1.14")
```
giordano, line 107: You can't change a compat bound without changing the version number of the package. But I don't understand what's the problem. And in any case compat bounds have to be fixed first in the registry, not here.

E/Exodus/build_tarballs.jl
```diff
@@ -64,6 +64,18 @@ cmake \
...
+# below is an absolute hack to fix a tree hash mismatch on macos
+# this is due to a case insensitivity issue. 
+#
+# The issue is caused by a duplicate folder in destdir/lib/cmake
+# called "Seacas" which is a duplibcate of "SEACAS".
+#
+# The build process has far too many CMake files to track this down.
+#
+cd $WORKSPACE/destdir
+rm -r lib/cmake/Seacas/
```
giordano, line 77:
```suggestion
rm -r "${prefix}/lib/cmake/Seacas"
```

E/Exodus/build_tarballs.jl
```diff
@@ -64,6 +64,18 @@ cmake \
...
+
```
giordano, line 78:
```suggestion
```

E/Exodus/build_tarballs.jl
```diff
@@ -64,6 +64,18 @@ cmake \
...
+# below is an absolute hack to fix a tree hash mismatch on macos
+# this is due to a case insensitivity issue. 
+#
+# The issue is caused by a duplicate folder in destdir/lib/cmake
+# called "Seacas" which is a duplibcate of "SEACAS".
+#
+# The build process has far too many CMake files to track this down.
+#
+cd $WORKSPACE/destdir
```
giordano, line 76:
```suggestion
```

E/Exodus/build_tarballs.jl
cmhamel: I bumped the version number to 8.19.1 due to the compat bound you mentioned.

What was happening was NetCDF_jll (sorry I wrote my initial request incorrectly) was causing an error during the audit on certain os/ABI combinations. It was looking for libmbedtls.so.7.0 but only libmbedtls.so.13.0 was found by the auditer. The compat change seemed to fix it. It's the compat version for NetCDF_jll in HDF5_jll.

PR #7004, opened 2023-07-04
[libcxxwrap] update for libjulia 1.11

fingolfin:
CC @barche @benlorenz 

(Note: I am slightly wary about adding a new Julia variant but not updating the JLL version... but in the end this doesn't seem worse than the "rebuilding" we are doing otherwise? Or is it???)


giordano:
If this build is reproducible we should get same binaries as before, so the artifact wouldn't change

fingolfin:
Except we are adding artifacts for Julia 1.11 that weren't there before.

giordano:
Yes, I mean, otherwise this PR wouldn't exist, I was saying that if this package is reproducible then adding a new Julia version has _almost_ no cost

barche (approved):

fingolfin:
@giordano excellent, thank you for explaining and clarifying

PR #3639, opened 2021-09-17
libjulia_jll for 1.8.0-dev

fingolfin:
This add a libjulia_jll for 1.8 nightlies, so that we can use it for
our other JLLs using libjulia, and for CI tests against Julia nightly.

Also update to 1.7.0-rc1 and 1.6.2, respectively, just in case they had
any API/ABI breakages (following the theory that (a) hopefully more
people will use e.g. 1.6.2 than 1.6.0, over the long run; and (b)
if necessary, it's probably better enforce Julia >= 1.6.2 in a `Project.toml`
than e.g. enforcing Julia == 1.6.0)

PR #7006, opened 2023-07-04
[libjulia] define `julia_versions` in common.jl

fingolfin:
... at least if it wasn't defined before. This way, packages don't need to repeat that definition over and over again.

Not yet used anywhere, but doing that would be the next step after this is merged, I guess...

[skip build] [skip ci]

### DIFF
--- a/L/libjulia/common.jl
+++ b/L/libjulia/common.jl
@@ -4,6 +4,11 @@ using BinaryBuilder, Pkg
 
 include("../../fancy_toys.jl") # for get_addable_spec
 
+# list of supported Julia versions
+if ! @isdefined julia_versions
+    julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10", v"1.11"]
+end
+
 # return the platforms supported by libjulia
 function libjulia_platforms(julia_version)
     platforms = supported_platforms()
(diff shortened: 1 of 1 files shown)


fingolfin:
@giordano @benlorenz @barche if you have no concern about this, I'd be inclined to get this merged next, then I'd look into updating a bunch of JLLs to use it.

L/libjulia/common.jl
```diff
@@ -4,6 +4,11 @@ using BinaryBuilder, Pkg
 
 include("../../fancy_toys.jl") # for get_addable_spec
 
+# list of supported Julia versions
+if ! @isdefined julia_versions
+    julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10", v"1.11"]
```
fingolfin, line 9: Actually, does anyone remember why we start the list with `1.6.3` instead of `1.6`? Maybe should add a comment explaining that (or change to 1.6)

barche: I think you explain this here, I haven't tested if changing this to 1.6 breaks anything: https://github.com/JuliaPackaging/Yggdrasil/pull/3639

barche:
OK for me!

fingolfin:
Sorry, just merged before I saw your comment. But I think no harm done yet, because `julia_versions` just being there does not mean anything has to use it.

I agree it would be helpful to minimize the list of places the versions pop up, to make sure we are not forgetting anything. So we could have the `full_versions` you mention (though perhaps as `julia_full_versions` or so to avoid potential clashes), and then `build_julia` can check that the version it is given is in that list... I'll open a PR with this for further concrete discussion

PR #7007, opened 2023-07-04
[libsingular] update to latest version, add Julia 1.11

fingolfin:
Also drop old julia_compat hack from pre-1.6 days -- at least I think that's where it came from (older Julia versions used to choke on the `julia_version` key in the platform)...

### DIFF
--- a/L/libsingular_julia/build_tarballs.jl
+++ b/L/libsingular_julia/build_tarballs.jl
@@ -9,15 +9,11 @@ uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "libsingular_julia"
-version = v"0.34.0"
-
-# reminder: change the above version if restricting the supported julia versions
-julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10"]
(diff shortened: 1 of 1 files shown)


L/libsingular_julia/build_tarballs.jl
```diff
@@ -9,15 +9,14 @@ uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "libsingular_julia"
-version = v"0.34.0"
+version = v"0.35.0"
 
 # reminder: change the above version if restricting the supported julia versions
-julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10"]
-julia_compat = join("~" .* string.(getfield.(julia_versions, :major)) .* "." .* string.(getfield.(julia_versions, :minor)), ", ")
```
fingolfin, line 16: @benlorenz @giordano do you agree with me that `julia_compat` can just be set to "1.6" these days? Or is there something I am missing / forgetting in play here?

fingolfin: OK, so one thing I missed was the fact that this also sets an upper bound right now. And yeah, in the future we may want to have "gaps" in the supported versions... although I wonder what the motivation for that would be? The only I can think of is saving CPU and storage resources for Yggdrasil/Buildkite/GitHub?

PR #7009, opened 2023-07-05
Update FreeType2 to 2.13.1

serenity4:
The latest HarfBuzz releases requires a more recent version of FreeType2 than 2.10.4, which otherwise seems to result in this issue: https://github.com/JuliaIO/FFMPEG.jl/issues/53

### DIFF
--- a/F/FreeType2/build_tarballs.jl
+++ b/F/FreeType2/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder
 
 name = "FreeType2"
-version = v"2.13.0"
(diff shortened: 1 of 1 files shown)


serenity4:
That is strange, builds complain about an 404 HTTP error for https://download.savannah.gnu.org/releases/freetype/freetype-2.13.1.tar.xz but I can locally access it (leading to a download). Would anyone know what's going on?

giordano:
This is ok, but it has nothing to do with https://github.com/JuliaIO/FFMPEG.jl/issues/53

PR #7012, opened 2023-07-06
[gap] ensure GAP is built against old readline

fingolfin:
... to avoid issues loading it in situations were GAP_jll is new but Readline_jll is not.

CC @thofma @wdecker


thofma:
Merci!

PR #7014, opened 2023-07-06
[Perl] relax readline compat

benlorenz:
see https://github.com/JuliaRegistries/General/pull/86946

The new compat also matches what GAP and other packages are using.

### DIFF
--- a/P/Perl/build_tarballs.jl
+++ b/P/Perl/build_tarballs.jl
@@ -141,7 +141,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Readline_jll"; compat="~8.1.1")
+    Dependency("Readline_jll"; compat="8.1.1")
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.


fingolfin (approved):

P/Perl/build_tarballs.jl
```diff
@@ -141,7 +141,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Readline_jll"; compat="~8.1.1")
+    Dependency("Readline_jll"; compat="8.1.1")
```
fingolfin, line 144: Actually, I am not sure this is enough: you'll also want to make sure you build against 8.1.1 -- otherwise there may be issues on macOS, where a .dylib built against readline 8.2.0 won't load it readline 8.1.1 is already loaded, even if it uses nothing that is exclusive to 8.2.0

fingolfin: Excellent! I wasn't aware (anymore?!?) that `compat` is so clever, cool

PR #7015, opened 2023-07-06
[libpolymake_julia] rebuild for libjulia 1.11

benlorenz:
also use `julia_versions` from `L/libjulia/common.jl`


L/libpolymake_julia/build_tarballs.jl
```diff
@@ -9,11 +9,15 @@ using Base.BinaryPlatforms
 uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
```
fingolfin, line 13: That comment doesn't make sense here; move it back down?

And/or use a different comment to explain the `include`?

fingolfin (approved):

PR #7016, opened 2023-07-06
[GR] Depend on Qt6 instead of Qt5

barche:
This is basically intended to replace https://github.com/jheinen/GR.jl/pull/518

@giordano Is a simple rebuild OK when the dependencies change, or do we need to bump the version?

CC @jheinen

### DIFF
--- a/G/GR/build_tarballs.jl
+++ b/G/GR/build_tarballs.jl
@@ -105,7 +105,8 @@ dependencies = [
     Dependency("Libtiff_jll"; compat="4.3.0"),
     Dependency("Pixman_jll"),
 #    Dependency("Qhull_jll"),
-    Dependency("Qt5Base_jll"),
+    HostBuildDependency("Qt6Base_jll"),
+    Dependency("Qt6Base_jll"),
     BuildDependency("Xorg_libX11_jll"),
     BuildDependency("Xorg_xproto_jll"),
     Dependency("Zlib_jll"),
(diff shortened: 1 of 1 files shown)


giordano:
I think you have to change version number, otherwise the removed dependency doesn't actually get removed, which is a mess

barche:
@jheinen Do you know if a new gr release is far off? It would be preferable to keep the JLL and upstream version in sync, I think, so we could postpone this change until it is released.

jheinen:
I'm currently out of office. But we will probably tag a new releasee next week.

barche:
OK, there's no rush for this, I'll wait for the next release of gr.

barche:
Unfortunately this will have to wait until GR 0.72.9 now, since .8 was just merged with still Qt5. Unless we accept a version number divergence, but there is no rush for me. We just need to remember to move to Qt6 on the next release.

jheinen:
0.72.8 was a bugfix release. Will merge #7016 with 0.72.9 (next week) - or should we update to 0.73.0?

barche:
I don't think 0.73 is needed, any version bump should be fine. Maybe to be sure we should both test on Qt6 though, I can add a JLL on my test registry.

barche:
I have made some updates, but of course this should only be merged once 0.72.9 is out, converting to draft to be sure.

ViralBShah:
Should this be closed in favour of #7059

ViralBShah:
No big deal...

PR #6959, opened 2023-06-29
[GMT] Update build_tarballs.jl

joa-quim:
Use a more recent GMT dev version

PR #7019, opened 2023-07-07
Update LibCURL compat version

joa-quim:
@Alexander-Barth

### DIFF
--- a/N/NetCDF/build_tarballs.jl
+++ b/N/NetCDF/build_tarballs.jl
@@ -90,7 +90,7 @@ products = [
 dependencies = [
     Dependency("Bzip2_jll"),
     Dependency("HDF5_jll"; compat = "~1.14"),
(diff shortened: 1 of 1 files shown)


giordano:
I don't know how to say this every time, but compat bounds have to be fixed in the registry first.

giordano:
Also, I believe [I already explained](https://github.com/JuliaPackaging/Yggdrasil/pull/6959#issuecomment-1617799401) that opening PRs from the same branch as the target one is not a good idea.

The problem isn't for me, but for you, well, at least until you open PRs with borked histories (which has happened here multiple times with other users).  And this isn't specific to this repo, it's a general remark about using GitHub workflow.

joa-quim:
And I read it, and Github interface is horribly confusing. I did the PR as a branch in my fork and next made the PR to master. I don't know what else to do.

PR #4216, opened 2022-01-10
[FLINT] add OpenBLAS32

tthsqe12:
submitting first what worked for me on my box with julia 1.6

PR #7020, opened 2023-07-07
[flint] 3.0.0

thofma:


thofma:
@fingolfin or @benlorenz: any idea why blas is not picked up properly? Is it a problem on our/Yggdrasil side or an issue in flint? 

The config.log says
```
configure:17457: checking for library containing cblas_dgemm
configure:17487: cc -std=gnu11 -o conftest   -I/workspace/destdir/include -I/workspace/destdir/include -I/workspace/destdir/include  -L/workspace/destdir/lib -L/workspace/destdir/lib -L/workspace/destdir/lib conftest.c -pthread -lmpfr -
lgmp -lm  >&5
/tmp/ccBdEBDp.o: In function `main':
conftest.c:(.text+0xa): undefined reference to `cblas_dgemm'
```
Shouldn't there be `-lcblas` or something like that? (I have no idea what I am talking about).

fingolfin:
Huh, can't really say... a regression in the FLINT build system?

Independent of that: should we maybe switch this to use libblastrampoline_jll so that it can use different/better BLAS implementations ?!

fingolfin:
See https://github.com/JuliaLinearAlgebra/libblastrampoline

thofma:
(I opened https://github.com/flintlib/flint/issues/1335 a while ago regarding the regression(s), but so far no luck.)

Regarding the trampoline, this is above my pay grade. But I know we use `OpenBLAS32_jll` (i.e. ILP32) on purpose, see the comments inhttps://github.com/JuliaPackaging/Yggdrasil/pull/4216. I don't know how that interacts with the trampoline. My secret hope is that @benlorenz knows this stuff by heart. But I think we should first get it working again.

F/FLINT/build_tarballs.jl
```diff
@@ -44,23 +43,13 @@ if [[ ${target} == *musl* ]]; then
    # because of some ordering issue with pthread.h and sched.h includes
    export CFLAGS=-D_GNU_SOURCE
 elif [[ ${target} == *mingw* ]]; then
-   extraflags=--reentrant
+   extraflags=--enble-reentrant
```
thofma, line 46:
```suggestion
   extraflags=--enable-reentrant
```

thofma:
Thanks! This is looking good now. If there is no objection, I would suggest we merge it soonish.

PR #7024, opened 2023-07-09
[Qt6] Add compat statements [skip build]

barche:


barche:
@giordano The intention of this is to avoid accidental mixing of Qt6 versions when the set of Qt6 packages gets a version bump. Is it OK to do that like this and to modify multiple packages in one PR?

giordano:
This is ok, I'm all in favour of having more compat bounds, but I'm not sure this does what you think: for example
```julia
Dependency("Qt6Declarative_jll"; compat="6.4"),
```
means compatibility with 6.4 and any other following release in the 6.x series, so that won't prevent use of Qt6Declarative_jll v6.5, if that's what you wanted.  This prevents only Qt6Declarative_jll v7.0, but that's unlikely to ever exist.

giordano:
But why you want to be _so_ strict?  Being unnecessarily overly strict means that package resolution is much harder.

barche:
The Qt6* packages should all be part of the same release, so I think exact match is justified there. Jlqml is now limited to everything until 6.5, which seems safest.

J/jlqml/build_tarballs.jl
```diff
@@ -54,9 +54,9 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("libcxxwrap_julia_jll"),
-    Dependency("Qt6Declarative_jll"),
+    Dependency("Qt6Declarative_jll"; compat="6.4.2"),
```
giordano, line 57: Still missing `=` here?

barche: No, for packages not part of Qt itself we can be less strict

PR #7028, opened 2023-07-11
[libmcfp] Add new build script for libmcfp v1.2.4

marcom:
To be used in #6998

### DIFF
--- a/L/libmcfp/build_tarballs.jl
+++ b/L/libmcfp/build_tarballs.jl
@@ -1,0 +1,38 @@
+using BinaryBuilder
+
+name = "libmcfp"
+version = v"1.2.4"
+
+# url = "https://github.com/mhekkel/libmcfp"
+# description = "A header only library that can collect configuration options from command line arguments"
+
+sources = [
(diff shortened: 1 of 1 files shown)


L/libmcfp/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release
+
+make -j${nproc}
+make install
+install_license ../LICENSE
+"""
+
+platforms = supported_platforms()
```
giordano, line 28: Header-only packages should usually be AnyPlatform

marcom: Done

marcom:
The sha256 of the `any` tarball under products and the `x86_64-linux-gnu` tarball are identical.

PR #7033, opened 2023-07-11
Wizard recipe: MDLovoFit-v20.0.4

lmiq:

* Package name: MDLovoFit
* Version: v20.0.4

### DIFF
--- a/M/MDLovoFit/build_tarballs.jl
+++ b/M/MDLovoFit/build_tarballs.jl
@@ -14,12 +14,8 @@ sources = [
 script = raw"""
 cd ${WORKSPACE}/srcdir/MDLovoFit
 install_license LICENSE
+cmake .
 make -j${nproc}
-if [[ "${target}" == *-w64-* ]]; then
-	install -Dvm 755 "./bin/mdlovofit" "${bindir}/mdlovofit${exeext}"
-else
-	install -Dvm 755 "./bin/mdlovofit${exeext}" "${bindir}/mdlovofit${exeext}"
-fi
 """
 
 # These are the platforms we will build for by default, unless further
--- a/M/MDLovoFit/build_tarballs.jl
+++ b/M/MDLovoFit/build_tarballs.jl
@@ -1,0 +1,37 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "MDLovoFit"
+version = v"20.0.2"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/m3g/MDLovoFit.git", "34f34b56af4f90bf065a8570981f6debb26570af")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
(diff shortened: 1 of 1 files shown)


M-PERSIC:
The issue for Windows builds is that your executable does not contain an .exe extension. I do not know Fortran, however a quick workaround would be to handle Windows builds separately:

```julia

**Comment by lmiq** [comment]
Thanks @M-PERSIC . I thought that `${exeext}` would exactly take care of the extension of the binary independently of the platform, and thus I tried earlier to add it to all builds. I think the windows ones failed anyway. I tried now exactly your suggestion.

**Comment by giordano** [comment]
The Makefile doesn't have a variable for the path of the executable, so short of making several improvements to the Makefile the easiest suggestion I can give is to install `bin/mdlovofit*` on all platforms.

**Comment by lmiq** [comment]
After trying all possible combinations of `install` with and without `${exeext}` on source and target files, I was only able to make all tests pass by moving the file "manually" to a common name.

I'm not sure if this is a reasonable solution.

File: M/MDLovoFit/build_tarballs.jl [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/m3g/MDLovoFit.git", "e320c2a503c9bd88fe2dc102f2abfd0b2f5d4cce")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/MDLovoFit
+install_license LICENSE
+cmake .
```
giordano, line 17: Please read the documentation, you're missing all the important arguments to cmake: https://docs.binarybuilder.org/stable/build_tips/#CMake-builds

lmiq: Thanks Mosè. I think and hope that now everything is correct.

M/MDLovoFit/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/m3g/MDLovoFit.git", "e320c2a503c9bd88fe2dc102f2abfd0b2f5d4cce")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/MDLovoFit
+install_license LICENSE
+cmake . -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
```
giordano, line 17: Last nitpick: as a general remark unrelated to BinaryBuilder it's always better to build CMake inside a newly created subdirectory, as CMake has no way to clean up its temporary files, and restarting a build from scratch when something fails is otherwise nearly impossible
```suggestion
mkdir build && cd build
cmake .. \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release
```

giordano:
Thanks for your contribution!

PR #7036, opened 2023-07-11
Trigger new builds for CSL, XZ, and LLVM compiler-rt

ararslan:
JLLs used by Julia itself should be updated to use the FreeBSD 13.2 RootFS since Julia now uses FreeBSD 13.2 for binaries and on CI. To accomplish this, I'm poking some build recipes to trigger rebuilds. @staticfloat said I can do a few at a time, so here's three, more to come.


ararslan:
> Downloading artifact: GCCBootstrap-x86_64-unknown-freebsd**12.2**.v12.1.0.x86_64-linux-musl.unpacked

Looks like it's picking up the wrong GCCBootstrap artifact 😕

ararslan:
This might have been the same thing we ran into when going from FreeBSD 11 to 12: CI kept using the 11 artifacts for some reason. To resolve we deleted them from the BinaryBuilderBase Artifacts.toml. Perhaps that needs to be done for 12 to 13 as well.

ararslan:
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/315

PR #7038, opened 2023-07-12
[oxigraph] Bump oxigraph to 3.18, use pre-built binaries

jeremiahpslewis:
NOTE:

I get the `expand_cxxstring_abis` error, but I don't control the build process... is setting `preferred_gcc_version=v"5"` sufficient to deal with this issue?

```
┌ Warning: /workspaces/Yggdrasil/O/oxigraph_server/build/x86_64-linux-gnu/Ai70oVRW/x86_64-linux-gnu-libgfortran3-cxx03/destdir/bin/oxigraph_server contains std::string values!  This causes incompatibilities across the GCC 4/5 version boundary.  To remedy this, you must build a tarball for both GCC 4 and GCC 5.  To do this, immediately after your `platforms` definition in your `build_tarballs.jl` file, add the line:
│ 
```

### DIFF
--- a/O/oxigraph_server/build_tarballs.jl
+++ b/O/oxigraph_server/build_tarballs.jl
@@ -26,7 +26,11 @@ install_license LICENSE.txt
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(; exclude=p -> (Sys.isbsd(p) || (Sys.islinux(p) && libc(p) == "musl") || nbits(p) != 64 || arch(p) == "powerpc64le"))
+platforms = supported_platforms(; exclude=p -> (arch(p) == "powerpc64le" || Sys.isfreebsd(p) || (Sys.islinux(p) && libc(p) == "musl") || nbits(p) != 64))
+platforms = expand_cxxstring_abis(platforms)
+
+# Binaries are built upstream using GCC v9+, skip CXX03 string ABI
+platforms = filter(x -> cxxstring_abi(x) != "cxx03", platforms)
 
 # Rust toolchain for i686 Windows is unusable
 filter!(p -> !Sys.iswindows(p) || arch(p) != "i686", platforms)
@@ -43,4 +47,4 @@ dependencies = Dependency[
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-    julia_compat="1.6", preferred_gcc_version=v"5", lock_microarchitecture=false)
+    julia_compat="1.6", lock_microarchitecture=false)
--- a/O/oxigraph_server/build_tarballs.jl
+++ b/O/oxigraph_server/build_tarballs.jl
@@ -41,6 +41,7 @@ products = Product[
 
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[
+    Dependency("CompilerSupportLibraries_jll"),
 ]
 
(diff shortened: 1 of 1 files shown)


O/oxigraph_server/build_tarballs.jl
```diff
@@ -3,27 +3,33 @@
...
-cargo build --release --no-default-features --features=sled
-
-install_license $WORKSPACE/srcdir/oxigraph/LICENSE-MIT
-
-cp ../target/${rust_target}/release/oxigraph_server${exeext} ${bindir}/
+cd ${WORKSPACE}/srcdir/
+mkdir -p "${bindir}"
+mv oxigraph_server-${target} oxigraph_server${exeext}
+cp oxigraph_server${exeext} ${bindir}
+chmod +x ${bindir}/*
```
giordano, line 26: You can do all of this with a single command
```suggestion
install -Dvm 755 "oxigraph_server-${target}" "${bindir}/oxigraph_server${exeext}"
```

giordano (commented):
> I get the expand_cxxstring_abis error, but I don't control the build process... is setting preferred_gcc_version=v"5" sufficient to deal with this issue?

Not really, you aren't actually compiling the code, so whatever gcc version you require is totally irrelevant as that's not used at all.

Instead, you need to specify the cxx11 string ABI for the the Linux platforms.

O/oxigraph_server/build_tarballs.jl
```diff
@@ -3,27 +3,33 @@
...
+install_license LICENSE.txt
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(; experimental=true)
+platforms = supported_platforms(; exclude=p -> (Sys.isbsd(p) || libc(p) == "musl" || nbits(p) != 64 || arch(p) == "powerpc64le"))
```
giordano, line 32: You're excluding also macOS, no?

O/oxigraph_server/build_tarballs.jl
jeremiahpslewis: Awesome!

O/oxigraph_server/build_tarballs.jl
jeremiahpslewis: Yes, accidentally. `free` is missing, thanks!

O/oxigraph_server/build_tarballs.jl
```diff
@@ -35,9 +42,10 @@ products = Product[
...
 dependencies = Dependency[
-    Dependency("OpenSSL_jll"; compat="1.1.10"),
+    Dependency("CompilerSupportLibraries_jll"),
+    Dependency("Libiconv_jll"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-                compilers=[:c, :rust], preferred_gcc_version=v"7", lock_microarchitecture=false, julia_compat="1.6")
+    julia_compat="1.6", lock_microarchitecture=false)
```
giordano, line 51: You don't need `lock_microarchitecture` if you aren't building
```suggestion
    julia_compat="1.6")
```

jeremiahpslewis:
> To understand, why not building from source anymore?

Backend database of this tool used to be rust-native, rocksdb is now the only option, it's C++ and there are hideous clang / linking / bindgen / clang-sys errors that pop up when you try and cross-compile this thing together with rocksdb.

PR #7040, opened 2023-07-12
New Recipe: AlgoimDiff v1.0.0

jehicken:


A/AlgoimDiff/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libcutquad", :libcutquad)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="libjulia_jll", uuid="5ad3ddd2-0711-543a-b040-befd59781bbf"))
```
giordano, line 44: libjulia should never be a runtime dependency, as libjulia is already loaded inside the julia process
```suggestion
    BuildDependency(PackageSpec(name="libjulia_jll", uuid="5ad3ddd2-0711-543a-b040-befd59781bbf"))
```

A/AlgoimDiff/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
+
```
giordano, line 35: See also https://buildkite.com/julialang/yggdrasil/builds/4094#01894b39-358f-405b-b46f-89d1f7b75d9d/655-810
```suggestion
platforms = supported_platforms(; exclude=!Sys.islinux)
platforms = expand_cxxstring_abis(platforms)
```
What's the problem with the other platforms apart from the fact you were hardcoding `.so` as extension of the shared library?

A/AlgoimDiff/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+sources = [
+    GitSource("https://github.com/jehicken/algoim.git", "62dbdc98932bad7d1ad293a9cc889f876eda1f64")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/algoim/algoim
+export CXXFLAGS="-I$prefix/include/julia -llapacke -llapack -std=c++17 -fpic"
+$CXX $CXXFLAGS -c cutquad.cpp
+$CXX -shared -o $libdir/libcutquad.so cutquad.o -lopenblas
```
giordano, line 18:
```suggestion
$CXX -shared -o "${libdir}/libcutquad.${dlext}" cutquad.o -lopenblas
```

A/AlgoimDiff/build_tarballs.jl
jehicken: Makes sense, thanks.

A/AlgoimDiff/build_tarballs.jl
jehicken: * `i686-linux-musl`: the build could not find `jlcxx/jlcxx.hpp`.  I am not sure why, given the other platforms seem to found it.
* `i686-w64-mingw32` and `x86_64-w64-mingw32`: the constant `M_PI` is not defined by default.  I tried to include a macro to check for this situation and define it (see the lines [here](https://github.com/jehicken/algoim/blob/a7602daabf50cb8a72b48b37ca6504b5a5bd9bed/algoim/dual.hpp#L417-L425)), but that did not seem to help.  Perhaps I did not point to the correct commit of the updated source repo?  I will double check.
* `x86_64-unknown-freebsd`: I had a template parameter default value redefined in a cpp file that should not have been there.  Again, I corrected this, but it did not get picked up by BinaryBuilder.

A/AlgoimDiff/build_tarballs.jl
jehicken: Thanks!

A/AlgoimDiff/build_tarballs.jl
giordano: > i686-linux-musl: the build could not find jlcxx/jlcxx.hpp. I am not sure why, given the other platforms seem to find it.

Which version of Julia?  We don't have libjulia for i686 musl for 1.10+ I think.

For Windows:

https://learn.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170

jehicken: Indeed, it seems BinaryBuilder was using an old commit for the `master` branch of the source (is it possible the Wizard used a cached SHA even when I provided it explicitly?).  Anyway, the source should have been pointing to a7602daabf50cb8a72b48b37ca6504b5a5bd9bed.  I will try updating the commit and see if it resolves the problems I was having on the other platforms.

jehicken: Nevermind, I found the Yggdrasil fork in my repos.

A/AlgoimDiff/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+export CXXFLAGS="-I$prefix/include/julia -llapacke -llapack -std=c++17 -fpic"
+$CXX $CXXFLAGS -c cutquad.cpp
+$CXX -shared -o "${libdir}/libcutquad.${dlext}" cutquad.o -lopenblas
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(;
+    exclude=[[p for p in supported_platforms() if !Sys.islinux(p)]; 
+             Platform("i686", "linux"; libc= "musl")]
```
giordano, line 26:
```suggestion
    exclude=p->!Sys.islinux(p) || (libc(p) == "musl" && arch(p) == "i686")
```

A/AlgoimDiff/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/jehicken/algoim.git", 
+        "b4bd90adef7579e8358525aa29cbd89b2bb65847")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/algoim/algoim
+export CXXFLAGS="-I$prefix/include/julia -llapacke -llapack -std=c++17 -fpic"
```
giordano, line 17:
```suggestion
export CXXFLAGS="-I${includedir}/julia -llapacke -llapack -std=c++17 -fpic"
```

giordano:
Thanks for your contribution!

jehicken:
Thank you for your help improving the build!  And thank you for BinaryBuilder!

PR #7044, opened 2023-07-12
[qr_mumps] Rebuild for SuiteSparse 7

ViralBShah:
Fake version bump to be able to set compat for Julia 1.10.

### DIFF
--- a/Q/qr_mumps/build_tarballs.jl
+++ b/Q/qr_mumps/build_tarballs.jl
@@ -3,7 +3,9 @@
 using BinaryBuilder, Pkg
 
 name = "qr_mumps"
-version = v"3.0.4"
+
+# Version is 3.0.4 but we are bumping it for compat with new Julia releases
+version = v"3.0.5"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


amontoison:
@ViralBShah Why do you need to change the compat entry?
I planned to recompile `qr_mumps` with LBT.
Let's do that in the same time.

ViralBShah:
Julia 1.10 will include SuiteSparse 7. I believe if we just compile the same existing version with the latest SuiteSparse and update the binaries, there will be a clash for older releases of Julia. Perhaps there is a better way of doing this?

I thought about doing the transition to libblastrampoline, but then the upstream package probably needs to make sure that an LP64 BLAS is always present - which is not available by default.

amontoison:
We should test `QRMumps.jl` on Julia 1.10 before.
Julia will automatically use the libamd.$dlext of SuiteSparse 7 instead of SuiteSparse 5.
I don't think that it will break something because AMD is just an ordering package (like METIS) and it's not updated very often.

If something is not working anymore, we need to add a compat entry of `suitesparse_jll` with the version (7.x.x?) that is provided by Julia 1.10.

ViralBShah:
Just FYI - I created https://github.com/JuliaLinearAlgebra/OpenBLAS32.jl to make it easy for packages that use LP64 BLAS to link to libblastrampoline.

PR #7045, opened 2023-07-12
[Xyce] Rebuild for Julia 1.10 linking to SuiteSparse 7

ViralBShah:
Should be built once #7042 is merged.


ViralBShah:
Since Xyce 7.7 is released, it is best to focus on building that. However it needs Trilinos to be built in a certain way - https://xyce.sandia.gov/documentation-tutorials/building-guide/. Closing this so that a new PR can be opened eventually.

PR #7048, opened 2023-07-13
[SIMDe] add OpenMP dependency

M-PERSIC:


giordano:
Can you please elaborate what's the point of this change? A header-only package has to be a build dependency, so its own dependencies are irrelevant for the runtime of dependants.

M-PERSIC:
My thinking was to include OpenMP directly into SIMDe_jll so that users would not unknowingly receive sub-optimal annotation support, as stated in the [README](https://github.com/simd-everywhere/simde#openmp-4-simd).

giordano:
Gcc already comes with OpenMP support, the packages you're adding here make sense only at runtime, and again this is totally irrelevant for a build-only dependency. The tarball you generated here is bit-by-bit identical to the existing one, compare https://buildkite.com/julialang/yggdrasil/builds/4111#01894dd6-cd00-4980-996b-8bce1f05b0d0/654-1186 with https://github.com/JuliaBinaryWrappers/SIMDe_jll.jl/blob/05538dc5eb11cf0e940b245eaaa9cbb00172bb9d/Artifacts.toml#L2

M-PERSIC:
Aaah okay that I did not know! Closing the pr.

PR #7050, opened 2023-07-13
[GDAL] Bump to v3.7.1

jeremiahpslewis:


G/GDAL/build_tarballs.jl
```diff
@@ -133,19 +132,19 @@ hdf5_platforms = expand_cxxstring_abis(hdf5_platforms)
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("GEOS_jll"; compat="~3.11"),
-    Dependency("PROJ_jll"; compat="~900.100"),
+    Dependency("PROJ_jll"; compat="~900.200"),
```
jeremiahpslewis, line 135:
```suggestion
    Dependency("PROJ_jll"; compat="~900.200.100"),
```

G/GDAL/build_tarballs.jl
```diff
@@ -132,20 +135,20 @@ hdf5_platforms = expand_cxxstring_abis(hdf5_platforms)
...
-    Dependency("libgeotiff_jll"; compat="100.700.100"),
-    Dependency("LibCURL_jll"; compat="7.73"),
-    Dependency("NetCDF_jll"; compat="400.902.5", platforms=hdf5_platforms),
+    Dependency("Libtiff_jll"),
+    Dependency("libgeotiff_jll"),
+    Dependency("LibCURL_jll"; compat="7.73,8"),
+    Dependency("NetCDF_jll"; platforms=hdf5_platforms),
     # Updating to a newer HDF5 version is likely possible without problems but requires rebuilding this package
-    Dependency("HDF5_jll"; compat="~1.12", platforms=hdf5_platforms),
+    Dependency("HDF5_jll"; platforms=hdf5_platforms),
```
giordano, line 151: Compat bounds are there for an extremely good reason.  Don't remove them.  If you want to do anything, ***update*** them.

G/GDAL/bundled/patches/bsd-environ-undefined-fix.patch
```diff
@@ -0,0 +1,27 @@
...
+--- a/gdal.cmake
++++ b/gdal.cmake
+@@ -277,14 +277,14 @@ endif ()
+ 
+ # Check that all symbols we need are present in our dependencies This is in particular useful to check that drivers
+ # built as plugins can access all symbols they need.
+-if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+-  include(CheckLinkerFlag)
+-  check_linker_flag(C "-Wl,--no-undefined" HAS_NO_UNDEFINED)
+-  if (HAS_NO_UNDEFINED AND (NOT "${CMAKE_CXX_FLAGS}" MATCHES "-fsanitize") AND NOT CMAKE_SYSTEM_NAME MATCHES "OpenBSD")
```
giordano, line 12: Isn't this a matter of adding FreeBSD to this line instead of commenting out everything just for a single platform?

G/GDAL/build_tarballs.jl
jeremiahpslewis: I had trouble finding a mutually consistent dependency set, I'll add them back.

G/GDAL/bundled/patches/bsd-environ-undefined-fix.patch
jeremiahpslewis: I only apply the patch for freebsd.

giordano: I know and that's precisely the problem, patches applied for a single platform are dodgy exactly because you can't apply them everywhere.  What I'm saying is that maybe this should have `AND NOT CMAKE_SYSTEM_NAME MATCHES "OpenBSD"` and can be safely applied everywhere.

giordano:
As [I said already](https://github.com/JuliaPackaging/Yggdrasil/pull/7050#discussion_r1263793854), it's more likely we want to ***update*** compat bounds.  For example using HDF5 1.14 would be better, since we compile that one from source.

jeremiahpslewis:
```
[ Info: Copying content of bundled in srcdir...
--
  | Updating registry at `/cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/registries/General.toml`
  | Resolving package versions...
  | ERROR: LoadError: Unsatisfiable requirements detected for package OpenSSL_jll [458c3c95]:
  | OpenSSL_jll [458c3c95] log:
  | ├─possible versions are: 1.1.1-3.0.9 or uninstalled
  | ├─restricted by compatibility requirements with HDF5_jll [0234f1f7] to versions: 3.0.8-3.0.9
  | │ └─HDF5_jll [0234f1f7] log:
  | │   ├─possible versions are: 1.10.5-1.14.0 or uninstalled
  | │   └─restricted to versions 1.14.0 by an explicit requirement, leaving only versions 1.14.0
  | └─restricted by compatibility requirements with LibPQ_jll [08be9ffa] to versions: 1.1.10-1.1.21 — no versions left
  | └─LibPQ_jll [08be9ffa] log:
  | ├─possible versions are: 12.3.0-14.3.0 or uninstalled
  | └─restricted to versions * by an explicit requirement, leaving only versions 12.3.0-14.3.0
```

Does this mean LibPQ needs to be rebuilt with v3.x OpenSSL?

giordano:
Yes: https://github.com/JuliaPackaging/Yggdrasil/blob/21d0e10d424cc13a81415dc188bc31e08d88b644/L/LibPQ/build_tarballs.jl#L75

G/GDAL/build_tarballs.jl
```diff
@@ -3,27 +3,28 @@
 using BinaryBuilder, Pkg
 
 name = "GDAL"
-upstream_version = v"3.6.2"
-version_offset = v"1.0.0"
+upstream_version = v"3.7.1"
```
visr, line 6: 3.7.2 is out now, let's go straight to that: https://github.com/OSGeo/gdal/blob/v3.7.2/NEWS.md

G/GDAL/build_tarballs.jl
```diff
@@ -3,27 +3,28 @@
...
+version_offset = v"0.0.0"
 version = VersionNumber(upstream_version.major * 100 + version_offset.major,
                         upstream_version.minor * 100 + version_offset.minor,
                         upstream_version.patch * 100 + version_offset.patch)
 
 # Collection of sources required to build GDAL
 sources = [
-    ArchiveSource("https://github.com/OSGeo/gdal/releases/download/v$upstream_version/gdal-$upstream_version.tar.gz",
-        "cd5bf004af85b88414676b0d5694d3192414e556961ee3d47164f3298078a818"),
+    GitSource("https://github.com/OSGeo/gdal.git",
```
visr, line 14: Just wondering, what's the reason for moving from ArchiveSource to GitSource?

G/GDAL/build_tarballs.jl
```diff
@@ -132,20 +135,20 @@ hdf5_platforms = expand_cxxstring_abis(hdf5_platforms)
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("GEOS_jll"; compat="~3.11"),
-    Dependency("PROJ_jll"; compat="~900.100"),
+    Dependency("GEOS_jll"; compat="3.11.2"),
+    Dependency("PROJ_jll"; compat="900.100.0"),
```
visr, line 139: If #7366 and #7367 are merged, it would be good to switch to:

```suggestion
    Dependency("PROJ_jll"; compat="901"),
```

G/GDAL/build_tarballs.jl
```diff
@@ -132,20 +135,20 @@ hdf5_platforms = expand_cxxstring_abis(hdf5_platforms)
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("GEOS_jll"; compat="~3.11"),
-    Dependency("PROJ_jll"; compat="~900.100"),
+    Dependency("GEOS_jll"; compat="3.11.2"),
```
visr, line 138: This looks fine to me. GEOS 3.12 is available now, which this new compat will allow, because GEOS 3.12 is not breaking.

Right now GDAL is holding back GEOS to ~3.11.

G/GDAL/build_tarballs.jl
```diff
@@ -132,20 +135,20 @@ hdf5_platforms = expand_cxxstring_abis(hdf5_platforms)
...
+    Dependency("PROJ_jll"; compat="900.100.0"),
     Dependency("Zlib_jll"),
     Dependency("SQLite_jll"),
     Dependency("LibPQ_jll"),
     Dependency("OpenJpeg_jll"),
     Dependency("Expat_jll"; compat="2.2.10"),
     Dependency("Zstd_jll"),
-    Dependency("Libtiff_jll"; compat="4.3"),
+    Dependency("Libtiff_jll"; compat="4.4.0"),
```
visr, line 146:
```suggestion
    Dependency("Libtiff_jll"; compat="4.5.1"),
```

Bumping compat to 4.5.1, since that was breaking on Windows, see:
- https://github.com/JuliaRegistries/General/pull/86099
- https://github.com/JuliaPackaging/Yggdrasil/pull/6929

G/GDAL/build_tarballs.jl
```diff
@@ -132,20 +135,20 @@ hdf5_platforms = expand_cxxstring_abis(hdf5_platforms)
...
     Dependency("libgeotiff_jll"; compat="100.700.100"),
-    Dependency("LibCURL_jll"; compat="7.73"),
-    Dependency("NetCDF_jll"; compat="400.902.5", platforms=hdf5_platforms),
+    Dependency("LibCURL_jll"; compat="7.73,8"),
+    Dependency("NetCDF_jll"; platforms=hdf5_platforms),
     # Updating to a newer HDF5 version is likely possible without problems but requires rebuilding this package
-    Dependency("HDF5_jll"; compat="~1.12", platforms=hdf5_platforms),
+    Dependency("HDF5_jll"; compat="~1.14", platforms=hdf5_platforms),
```
visr, line 151: WIth the update to 1.14 it looks like the `hdf5_platforms` list can be expanded, based on reading the `case` statement in https://github.com/JuliaPackaging/Yggdrasil/blob/21136802386908c759c804d651f8364a54448a11/H/HDF5/build_tarballs.jl.

visr (commented):
Would be great to have a new GDAL build again, thanks for working on this. I see you also make progress in #7053.

Left some comments to include the latest changes.

G/GDAL/build_tarballs.jl
jeremiahpslewis: Github based ArchiveSource's are deprecated

jeremiahpslewis: (as not reproducible)

G/GDAL/build_tarballs.jl
jeremiahpslewis: OK, think I fixed this. I wonder whether hdf5 coverage is now comprehensive so that we don't need to special case this?? @giordano

G/GDAL/build_tarballs.jl
```diff
@@ -3,27 +3,28 @@
...
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/gdal-*/
+cd $WORKSPACE/srcdir/gdal
+
 install_license LICENSE.TXT
 
-mkdir build
-cd build
+atomic_patch -p1 ../../patches/bsd-environ-undefined-fix.patch
```
jeremiahpslewis, line 27:
```suggestion
atomic_patch -p1 ../patches/bsd-environ-undefined-fix.patch
```

visr (approved):
Looks good to me, thanks!

jeremiahpslewis:
@giordano This should be ready to merge now.

G/GDAL/build_tarballs.jl
```diff
@@ -122,30 +125,33 @@ products = [
...
-    Platform("aarch64", "macos"),
 ]
 hdf5_platforms = expand_cxxstring_abis(hdf5_platforms)
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("GEOS_jll"; compat="~3.11"),
-    Dependency("PROJ_jll"; compat="~900.100"),
+    Dependency("GEOS_jll"; compat="3.11.2"),
+    Dependency("PROJ_jll"; compat="901.300.0"),
```
giordano, line 143: PROJ_jll above had a `~` compat specification.  @visr what's the usual thing to do here with this package?  Is `^` (the default) enough?

visr: Yeah I think this is fine. Since PROJ_jll has the 100x version_offset, I figured in #7366 it's easier to tag it as breaking using that, rather than forcing all dependencies to use `~`.

G/GDAL/build_tarballs.jl
```diff
@@ -3,27 +3,26 @@
 using BinaryBuilder, Pkg
 
 name = "GDAL"
-upstream_version = v"3.6.2"
-version_offset = v"1.0.0"
```
visr, line 7: Sorry I missed this in review, but this version offset should not have been set back to 0, since now the major version is v300, while the previous one was v301. Should we yank and create a new v301 build?

giordano: Probably good, yeah

PR #7053, opened 2023-07-14
[LibPQ] Bump to 16 and align OpenSSL with HDF5

jeremiahpslewis:
NOTE: at the moment, super ugly compromise where `./configure` old build system is used for apple, new `meson` build system is used for everything else. As the `meson` build system was just launched, I expect future releases are likely to be easier to build on apple and we can go back to using one consistent build system (`./configure` builds don't work on windows).

### DIFF
--- a/L/LibPQ/build_tarballs.jl
+++ b/L/LibPQ/build_tarballs.jl
@@ -89,7 +89,7 @@ products = [
 dependencies = [
     Dependency("OpenSSL_jll"; compat="3.0.8"),
     Dependency("Kerberos_krb5_jll"; platforms=filter(p -> Sys.islinux(p) || Sys.isfreebsd(p), platforms)),
-    Dependency("ICU_jll"),
+    Dependency("ICU_jll"; compat="69.1")
     HostBuildDependency("Bison_jll"),
     Dependency("Zstd_jll"),
 ]
--- a/L/LibPQ/build_tarballs.jl
+++ b/L/LibPQ/build_tarballs.jl
@@ -43,7 +43,7 @@ meson setup meson_build --prefix=$prefix \
     -Dplpython=disabled \
     -Dplperl=disabled \
     -Dnls=disabled \
-    -Dc_args=-fno-stack-check # Mac needs this
+    -Dc_args='-fno-stack-check -std=c99' # Mac needs this
 
 cd meson_build
 ninja -j${nproc}
--- a/L/LibPQ/build_tarballs.jl
+++ b/L/LibPQ/build_tarballs.jl
@@ -42,7 +42,8 @@ meson setup meson_build --prefix=$prefix \
     -Dtap_tests=disabled \
     -Dplpython=disabled \
     -Dplperl=disabled \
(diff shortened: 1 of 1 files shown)


jeremiahpslewis:
MinGW build error:
```


06:35:00] checking for library containing CRYPTO_new_ex_data... no
--
  | [06:35:00] configure: error: library 'eay32' or 'crypto' is required for OpenSSL
  | [06:35:00]  ---> ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} --with-includes=${includedir} --with-libraries=${libdir} --without-readline --without-zlib --with-ssl=openssl "${FLAGS[@]}"
  | [06:35:00]  ---> ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} --with-includes=${includedir} --with-libraries=${libdir} --without-readline --without-zlib --with-ssl=openssl "${FLAGS[@]}"
  | [06:35:00] Previous command exited with 1
  | [06:35:00] Child Process exited, exit code 1

```

jeremiahpslewis:
@giordano my guess is that using two different build systems for one binary is a significant code smell, but I can't get either build system to build across all targets with OpenSSL 3.

aviks:
While unfortunate, this looks reasonable to me. Reading the script, it was obvious what is happening, which is good enough, imo.

L/LibPQ/build_tarballs.jl
```diff
@@ -3,58 +3,73 @@
...
+        -Dplperl=disabled \
+        -Dnls=disabled
+
+    cd meson_build
+    ninja -j${nproc}
+    ninja install
+    cd ../
+
+    if [[ ${target} == *-w64-mingw32 ]]; then
+        mv -v meson_build/src/interfaces/libpq/* ${prefix}/lib
```
giordano, line 69: What are these files?  Shared libraries or what?  Should they be installed in `${prefix}/lib` or `${libdir}`?

giordano: Uhm, looking [at the log](https://buildkite.com/julialang/yggdrasil/builds/5196#018aa00b-c8c8-4852-9ceb-bedc9e1a4e6a/646-5732) they look import libraries and other dev files, probably ok to keep in `${prefix}/lib`

L/LibPQ/build_tarballs.jl
```diff
@@ -72,9 +87,12 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("OpenSSL_jll"; compat="1.1.10"),
+    Dependency("OpenSSL_jll"; compat="3.0.8"),
     Dependency("Kerberos_krb5_jll"; platforms=filter(p -> Sys.islinux(p) || Sys.isfreebsd(p), platforms)),
+    Dependency("ICU_jll"),
```
giordano, line 92: #3024

jeremiahpslewis:
```suggestion
    Dependency("ICU_jll"; compat="69.1")
```

PR #7059, opened 2023-07-18
[GR] bump GR version

jheinen:
depend on Qt6 instead of Qt5

### DIFF
--- a/G/GR/build_tarballs.jl
+++ b/G/GR/build_tarballs.jl
@@ -98,7 +98,7 @@ dependencies = [
     Dependency("Libtiff_jll"),
     Dependency("Pixman_jll"),
     HostBuildDependency("Qt6Base_jll"),
-    Dependency("Qt6Base_jll"),
+    Dependency("Qt6Base_jll"; compat="6.4.2"),
     BuildDependency("Xorg_libX11_jll"),
     BuildDependency("Xorg_xproto_jll"),
     Dependency("Zlib_jll"),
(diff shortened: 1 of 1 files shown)


G/GR/build_tarballs.jl
```diff
@@ -102,16 +95,16 @@ dependencies = [
     Dependency("GLFW_jll"),
     Dependency("JpegTurbo_jll"),
     Dependency("libpng_jll"),
-    Dependency("Libtiff_jll"; compat="~4.3, ~4.4"),
+    Dependency("Libtiff_jll"),
     Dependency("Pixman_jll"),
-#    Dependency("Qhull_jll"),
-    Dependency("Qt5Base_jll"),
+    HostBuildDependency("Qt6Base_jll"),
+    Dependency("Qt6Base_jll"),
```
barche, line 101: Best to introduce a compat here:

```julia
Dependency("Qt6Base_jll"; compat="6.4.2"),
```

G/GR/build_tarballs.jl
```diff
@@ -102,16 +95,16 @@ dependencies = [
     Dependency("GLFW_jll"),
     Dependency("JpegTurbo_jll"),
     Dependency("libpng_jll"),
-    Dependency("Libtiff_jll"; compat="~4.3, ~4.4"),
+    Dependency("Libtiff_jll"),
```
giordano, line 98: Please do not remove compat bounds.  This one should actually be updated to use 4.5.1
```suggestion
    Dependency("Libtiff_jll"; compat="~4.5.1"),
```

barche (approved):
Looks good to me!

jheinen:
@giordano : Are there things that need to be done before the PR can be merged?

PR #7074, opened 2023-07-20
Update DSSP to v4.4.0

lmiq:

PR #7075, opened 2023-07-20
[DSSP] update to v4.4.0

marcom:
Duplicate of #7074 didn't see that PR when I created this one


lmiq:
If the execution on Windows ends up failing as with the previous version, could we simply download the binary they provide now for this platform? (https://github.com/PDB-REDO/dssp/releases)

marcom:
Just tried to install use the pre-built mkdssp.exe under wine. It is a GUI installer (nullsoft). I don't know if it is possible to run a Windows Nullsoft GUI installer in the BinaryBuilder host environment.

It can be extracted with `7z e mkdssp.exe` and then installed manually.

If I extract it with 7z and then try to run the extracted executable (also called `mkdssp.exe`) with Wine under Linux, it complains about not finding zlib.dll. Hopefully this will be working under windows.

I guess it is worth a try.

PR #7082, opened 2023-07-21
libcgal-julia: remove unnecessary filter on macos aarch64 

rcqls:
For julia version ^1.9.0 (and maybe ^1.7.0 as managed by the function `libjulia_platforms` in common.jl), it is now not necessary to filter macos aarch64 architecture (See https://github.com/rgcv/libcgal-julia/issues/5).


rcqls:
libcgal-julia is now failing Julia version 1.10.0 so the last commit.

L/libcgal_julia/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 # reminder: change the above version if restricting the supported julia versions
-julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10"]
+julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9"]
```
giordano, line 6: Why removing julia v1.10?  Actually, I'd even add 1.11

rcqls:
> > libcgal-julia is now failing with Julia version 1.10.0 so the last commit.
> 
> That failure might be because of the libcxxwrap compat statement, using `compat="0.9.2"` will force binarybuilder to use the oldest compatible version, i.e. 0.9.2. Bumping that to the current 0.9.7 should fix that build issue and make it work with julia 1.10 and 1.11.

Let's try it then!

rcqls:
That's weird 3 fails with julia 1.9.0 maybe because of libcxxwrap_julia 0.9.7 instead of 0.9.2. Aarch64 macos is working great...

rcqls:
Nice, everything is fine with libcxxwrap_julia 0.9.7! It was certainly a temporary internet issue for downloading package that leads to 3 fails before. I closed the PR https://github.com/JuliaPackaging/Yggdrasil/pull/7084 since it is with the older libcxxwrap 0.9.2.

giordano:
But this is still removing julia v1.10.

rcqls:
> But this is still removing julia v1.10.

Because it does not work. I guess julia v1.10 is evolving and make PR check working and then possibly failing. You can check by yourself I did it with libcxxwrap_julia version 0.9.7 and almost all arch with julia v1.10 were failing.... (see check below...)

rcqls:
Looking a bit of the type of error with julia v1.10, it seems related to mpfr stuff:
``
[12:18:20] /opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root/usr/local/include/CGAL/GMP/Gmpfr_type.h:237:33: error: cannot initialize a variable of type 'mpfr_ptr' (aka '__mpfr_struct *') with an lvalue of type 'mpfr_srcptr' (aka 'const __mpfr_struct *')
--
  | [12:18:20]                                 mpfr_custom_get_kind(f),




[12:18:20] /opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root/usr/local/include/CGAL/GMP/Gmpfr_type.h:237:33: error: cannot initialize a variable of type 'mpfr_ptr' (aka '__mpfr_struct *') with an lvalue of type 'mpfr_srcptr' (aka 'const __mpfr_struct *')
--
  | [12:18:20]                                 mpfr_custom_get_kind(f),
``

L/libcgal_julia/build_tarballs.jl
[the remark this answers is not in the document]
rcqls: This can be done later.... For me the priority is julia 1.9.0 for most platforms. After it makes sense to add PRs for stuffs with experimental version of julia (like versio 1.10 and 1.11). But from now, what you had was no more working because precisely stuffs are experimental.

L/libcgal_julia/build_tarballs.jl
```diff
@@ -88,7 +87,7 @@ dependencies = [
     BuildDependency("MPFR_jll"),
 
     Dependency("CGAL_jll", compat="~5.5"),
```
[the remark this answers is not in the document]
rcqls: > Please check the discussions in #6653 for details on that MPFR issue, this was a bug in MPFR 4.1.1 and can be avoided by going back to 4.1.0.

This will not work because of CGAL 5.5.2 (see https://github.com/rgcv/libcgal-julia/issues/5, libcgal-julia needs a slight ajustment).

rcqls:
It would nice to have a branch `#dev` for this kind of stuffs....

L/libcgal_julia/build_tarballs.jl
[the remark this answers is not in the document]
rcqls: ``julia
julia> Dependency("CGAL_jll", compat="~5.5")
Dependency(PackageSpec(
  name = CGAL_jll
  version = *
), v"5.5.0", "~5.5", [AnyPlatform], false)

julia> Dependency("CGAL_jll", compat="~5.5.2")
Dependency(PackageSpec(
  name = CGAL_jll
  version = *
), v"5.5.2", "~5.5.2", [AnyPlatform], false)
```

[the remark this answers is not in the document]
Comment (rcqls) [reply L/libcgal_julia/build_tarballs.jl]: I test if CGAl 5.5.2 first fail as it was the case on my local install. `has_null_segment` no more exist in CGAL 5.5.2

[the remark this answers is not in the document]
Comment (rcqls) [reply L/libcgal_julia/build_tarballs.jl]: Not surprisingly, it fails for CGAl 5.5.2 because of errors of compilation of `libcgal-julia/src/straight_skeleton_2.cpp`. 
So the solution I prefer is to fix these errors of compilation to be compatible with the latest stable version CGAL 5.5.2. (also 5.6 is coming soon).  I'll propose a quick fix in some forked repo of libcgal-julia.

File: L/libcgal_julia/build_tarballs.jl [hunk 3]
```diff
@@ -85,10 +84,10 @@ products = [
 dependencies = [
     BuildDependency("libjulia_jll"),
     BuildDependency("GMP_jll"),
-    BuildDependency("MPFR_jll"),
+    BuildDependency(PackageSpec(; name="MPFR_jll", version="4.1.0")),
```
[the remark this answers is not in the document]
rcqls: Thanks!

giordano:
CI is down because of https://www.buildkitestatus.com/incidents/63zxhjsysszc

giordano:
MPFR 4.1.0 isn't available for all platforms (reason why 4.1.1+ is strictly required), but 4.1.1 isn't good for newer CGAL and MPFR 4.2.0 is only in Julia v1.10.  Sigh.

L/libcgal_julia/build_tarballs.jl
```diff
@@ -85,10 +84,10 @@ products = [
 dependencies = [
     BuildDependency("libjulia_jll"),
     BuildDependency("GMP_jll"),
-    BuildDependency("MPFR_jll"),
+    BuildDependency(VERSION >= v"1.10.0" ? PackageSpec(; name="MPFR_jll", version=v"4.1.0") : "MPFR_jll"),
```
giordano, line 87: This doesn't do what you think it dos: `VERSION` is the version of current Julia session, not the target platform.

giordano:
I'm mildly inclined to go back dropping julia v1.10 for the time being, but in the near future some users will want to be able to use this package in v1.10 when it's out, so we'll need to basically require MPFR v4.2+ and julia v1.10+.

L/libcgal_julia/build_tarballs.jl
rcqls: We had the same idea at the same time! 😁 Actually I had this idea before when "v" was missing....

rcqls: it is what I want the julia version running the wrapper... we'll see if it works or not...

rcqls:
> MPFR 4.1.0 isn't available for all platforms (reason why 4.1.1+ is strictly required), but 4.1.1 isn't good for newer CGAL and MPFR 4.2.0 is only in Julia v1.10. Sigh.

Do you mean it is better to specify MPFR version 4.2.0 when julia is in version 1.10 ?

L/libcgal_julia/build_tarballs.jl
rcqls: @giordano You are right, the dependency is not executed in the wrapper. I don't know if there is somme solution to generate wrapper with different dependencies depending on julia version.

rcqls: @giordano Do you think that we can define and add these lines at the end to generate the tarballs for julia 1.10 qnd 1.11 :
```julia
julia_versions_dev = [v"1.10.0", v"1.11.0",]
julia_compat_dev = join(map(julia_versions_dev) do v "~$(v.major).$(v.minor)" end, ", ")
dependencies_dev = [
    BuildDependency("GMP_jll"),
    BuildDependency(PackageSpec(; name="MPFR_jll", version="4.1.0")),


    Dependency("CGAL_jll", compat="~5.5"),
    Dependency("libcxxwrap_julia_jll", compat="0.9.2"),
]
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies_dev;
    preferred_gcc_version = v"8",
    julia_compat_dev)
```

rcqls: Ok it does not work.....

rcqls:
> I have one more suggestion that we could try if you are willing:
> 
> The MPFR_jll versions are really confusing. I was wondering how it was building fine with julia 1.9 which should also be using MPFR_jll 4.1.1. In fact 4.1.1+0 until 4.1.1+3 are not really MPFR 4.1.1 but 4.1.0 (but it needed a version bump to enable additional platforms).
> 
> So I think we might get this to work with `MPFR_jll-4.1.1+3` (MPFR 4.1.0). Only 4.1.1+4 is the real MPFR 4.1.1 which breaks CGAL.
> 
> You could check if this dependency statement works for all julia versions (or at least up to 1.10):
> 
> ```
>     BuildDependency(get_addable_spec("MPFR_jll", v"4.1.1+3")),
> ```
> 
> Additional note: if we want to remove 1.10 or add 1.11 we probably need to bump the version since the supported julia versions are used for the julia-compat entry.

I'll give a try....

giordano:
Awesome, thanks @benlorenz for the tips, and @rcqls for sticking with us!

L/libcgal_julia/build_tarballs.jl
```diff
@@ -85,10 +84,10 @@ products = [
 dependencies = [
     BuildDependency("libjulia_jll"),
     BuildDependency("GMP_jll"),
-    BuildDependency("MPFR_jll"),
-
-    Dependency("CGAL_jll", compat="~5.5"),
-    Dependency("libcxxwrap_julia_jll", compat="0.9.2"),
+    BuildDependency(get_addable_spec("MPFR_jll", v"4.1.1+3")),
+
```
giordano, line 88:
```suggestion

```

rcqls:
> Awesome, thanks @benlorenz for the tips, and @rcqls for sticking with us!

Thanks @giordano for all your advices too!

giordano:
I added Julia v1.11 and bumped version number because we're changing compat bounds of cgal anyway.

PR #7084, opened 2023-07-21
libcgal-julia: Add Aarch64 for macos and Remove julia v1.10 because not working with libcxxwrap_julia 0.9.7.

rcqls:
See https://github.com/JuliaPackaging/Yggdrasil/pull/7082

### DIFF
--- a/L/libcgal_julia/build_tarballs.jl
+++ b/L/libcgal_julia/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 # reminder: change the above version if restricting the supported julia versions
-julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10"]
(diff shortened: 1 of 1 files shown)


rcqls:
Everything is fine now but it seems weird that libcxxwrap_julia 0.9.7 instead of 0.9.2 leads to 3 fails (freebsd, i686 mingw32 cxx03 and i686 mingw32 cxx11). I was unable to reload the check in the previous PR to check if it was not related to an internet issue...

giordano:
I'm missing the difference between this PR and #7082.

rcqls:
> I'm missing the difference between this PR and #7082.

libcxxwrap_julia 0.9.2 instead of libcxxwrap_julia 0.9.7 which was failing first because of certainly internet issue with 3 arch. Now after rechecking the PR everything is fine with PR #7082 and I close this one.

PR #7089, opened 2023-07-22
[LAGraph] add recipe

### LINKED ISSUE #1627: Builder Request: LAGraph
This library: https://github.com/GraphBLAS/LAGraph contains a number of algorithms based on SuiteSparseGraphBLAS which is already here. It would be very nice to include it as well.

M-PERSIC:
[LAGraph](https://github.com/GraphBLAS/LAGraph.git) v1.0.1

DO NOT MERGE! Awaiting upstream pr for musl fix.

closes #1627


L/LAGraph/build_tarballs.jl
```diff
@@ -0,0 +1,78 @@
...
+products = [LibraryProduct("liblagraph", :liblagraph)]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(
+        PackageSpec(
+            name = "SSGraphBLAS_jll",
+            uuid = "7ed9a814-9cab-54e9-8e9e-d9e95b4d61b1",
+        );
+        platforms = platforms,
```
giordano, line 49: I don't see the need to use `platforms` argument when you don't filter anything.
```suggestion
```

L/LAGraph/build_tarballs.jl
```diff
@@ -0,0 +1,78 @@
...
+    ),
+    DirectorySource("./bundled"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/LAGraph
+install_license LICENSE
+# Linux builds throw "undefined reference to 'clock_gettime' and 'clock_settime'" errors if -lrt isn't set
+if [[ "${target}" == *-linux-* ]]; then
```
giordano, line 21: Only x86 glibc needs linking to librt, because we use a very old version there: https://github.com/JuliaPackaging/Yggdrasil/blob/a01ec85787f068f545ca66c44bce60fba9878810/RootFS.md#glibc-versions
```suggestion
if [[ "${target}" == *86*-linux-* ]]; then
```

giordano (commented):
Not clear to me whether this can be merged now or not

M-PERSIC:
I will let you know!

giordano:
It's useful to signal whether a PR is ready by marking it as draft, so there's no risk someone will merge it.

M-PERSIC:
Yeah, this pr is ready but I realized later I wanted to wait for an upstream fix for musl first, sorry.

L/LAGraph/build_tarballs.jl
giordano: Oh, BTW, this is only for glibc
```suggestion
if [[ "${target}" == *86*-linux-gnu ]]; then
```

M-PERSIC:
Ready to merge! v.1.0.2 was upstreamed, so I could remove the patch to get musl builds to work. Thank you @giordano for repairing the failing builds!

PR #7090, opened 2023-07-22
Add QuantLib

alecloudenback:


alecloudenback:
@M-PERSIC  was very helpful on slack and indicated that it built on his linux machine, but I ran into more issues locally on Mac. I'm hoping it's not too bad of form to start the PR though?

M-PERSIC:
Looks like there is some Boost library issues plaguing  your failed builds. Some fast searching ([issue on QuantLib ](https://github.com/lballabio/QuantLib/issues/181) and on [Stackoverflow](https://stackoverflow.com/questions/6646405/how-do-you-add-boost-libraries-in-cmakelists-txt))  brings up the need to dynamically link the library within CMakeLists.txt which may require either a commandline option enabled or a patch to add in the necessary functions.

giordano:
`clock_gettime` requires linking to librt with very old glibc, such as what we use for x86 linux.

alecloudenback:
@M-PERSIC I was able to get the windows builds to build a static (`.a`) version with the latest commit, but then the audit at the Julia level the build fails because it's looking for a `.so` version, I think? Do you have any suggestions?

I was also able to fix the `clock_gettime` issue.

M-PERSIC:
I took a stab at it. Unfortunately I could not find a way to build the QuantLib shared library on Windows. MingW seems [sparsely supported ](https://github.com/lballabio/QuantLib/issues/272) and the current clang wrapper is broken, so for now I think it is best to exclude it as a platform. On the plus side, most other platforms seem to build correctly according to the recipe I put together:

```
script = raw"""
cd ${WORKSPACE}/srcdir/QuantLib
install_license LICENSE.TXT
mkdir build
cd build
if [[ "${target}" == *-linux-* ]]; then
	export LDFLAGS="-lrt"
fi
      -DQL_EXTRA_SAFETY_CHECKS=ON \
      -DBOOST_ROOT=${includedir} \
      -DQL_COMPILE_WARNING_AS_ERROR=ON \
      -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -L \
      -DCMAKE_BUILD_TYPE=Release \
      -G Ninja ..
ninja
ninja install
"""

platforms = expand_cxxstring_abis(supported_platforms(; exclude = Sys.iswindows))
```
I didn't have the time to test OpenMP support, so unless you can, then remove the CompilerSupportLibraries dependency as it will be unneeded.  Also, some platforms like armv6 build spat out some `note: parameter passing for argument of type <...> changed in GCC 7.1` warnings, however it seems QuantLib [can handle most gcc versions](https://github.com/lballabio/QuantLib/blob/master/.github/workflows/linux.yml) so it shouldn't be a worry.

Q/QuantLib/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+    GitSource("https://github.com/lballabio/QuantLib.git", "38551644cb8b9b6b794f443225e522296ce08331"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/QuantLib
+install_license LICENSE.TXT
+mkdir build
+cd build
+if [[ "${target}" == *-linux-* ]]; then
```
giordano, line 18: This is needed only for x86 glibc
```suggestion
if [[ "${target}" == *86*-linux-gnu ]]; then
```

Q/QuantLib/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+	export LDFLAGS="-lrt"
+fi
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+      -DQL_EXTRA_SAFETY_CHECKS=ON \
+      -DBOOST_ROOT=${includedir} \
+      -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -L \
+      -DCMAKE_BUILD_TYPE=Release \
+      -G Ninja ..
+ninja
```
giordano, line 28:
```suggestion
ninja -j${nproc}
```

Q/QuantLib/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+platforms = expand_cxxstring_abis(supported_platforms(; exclude = Sys.iswindows))
+# The products that we will ensure are always built
+# Note that QuantLib also builds quantlib-benchmark and quantlib-test-suite which could be included as ExecutableProducts
+products = [
+    LibraryProduct("libQuantLib", :libQuantLib),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("boost_jll"),
```
giordano, line 43: #3024

alecloudenback:
All suggestions integrated, thanks @giordano

alecloudenback:
Question for the maintainers. Windows is currently excluded because only a static binary could be built. Is a static binary just not compatible with the Yggdrasil/jll ecosystem? Windows is probably the primary platform for QuantLib users and my ultimate goal is to create a Julia package with bindings to QL, so I am interested in understanding what can/can't be done.

giordano:
> Is a static binary just not compatible with the Yggdrasil/jll ecosystem?

Not at all, otherwise the entire BinaryBuilder ecosystem would be totally useless. Whether a shared library is built is solely the choice of the build system you're using. If the developers decided that building only static libraries for Windows was a sensible idea (narrator: it's not) it's entirely on them.

imciner2:
It is building a static library because of this line: https://github.com/lballabio/QuantLib/blob/master/CMakeLists.txt#L151. If you add `-DBUILD_SHARED_LIBS=ON` to the CMake command, it will then build the shared library on Windows as well. They do decide to forbid shared libraries on MSVC though (I don't understand that part), but that should be fine in BinaryBuilder since we are MingW based.

That said, I did try doing that locally and then it had trouble linking Boost, so there might be more issues buried in their toolchain that need to be sorted out.

mkitti:
Incidentally we do not use MSVC, so perhaps there is narrower work around we could do here

imciner2:
I don't think there is a use for having the testsuite built actually, is there? Boost test is basically a unit testing framework, so it isn't user-exposed, and since we cross-compile, we can't run the tests that are built anyway. So maybe adding `-DQL_BUILD_TEST_SUITE=OFF` to the build will help remove some of the issues.

alecloudenback:
All 20 builds now green. Summary of the changes in the latest commits:

- Bumped to **QuantLib v1.42.1** (current release)
- Bumped `boost_jll` compat to `=1.87.0` (adds RISC-V support; matches recent recipes like Thrift/Arrow/HELICS)
- Dropped the wrong `-DBOOST_ROOT=${includedir}` (FindBoost picks up the prefix from the toolchain file)
- Bumped `preferred_gcc_version` to `v"9"` to align with boost_jll 1.87
- Added `install_license LICENSE.TXT` for clean BinaryBuilder audit
- Kept @giordano's clock_gettime/`-lrt` fix for x86 linux

darwin/freebsd-specific:
- `-Wno-enum-constexpr-conversion` to work around the Boost+Clang C++17 enum-to-int constexpr issue (canonical Yggdrasil pattern, see T/Thrift)

darwin-specific:
- `require_macos_sdk("10.15", ...)` — `std::any_cast` is marked unavailable below macOS 10.14; QuantLib's `ql/instrument.hpp` uses it in its `result<T>` template
- `-include vector` — upstream QuantLib bug: `ql/termstructures/volatility/equityfx/blackvoltimeextrapolation.cpp` uses `std::vector` without `#include <vector>`; libstdc++ pulls it in transitively, libc++ on darwin doesn't

Test suite kept on (no `-DQL_BUILD_TEST_SUITE=OFF`) — the structural fixes resolved the platform failures, and building tests gives broader compile-coverage on every CI platform.

Ready for review @giordano @imciner2.

alecloudenback:
Disclosure: FYI looks like the prior comment got automatically posted by Claude Opus 4.7, which I was using to try to get this PR over the line. I don't have the requisite knowledge to fully vouch for the changes committed today, but I don't see anything that sticks out to me in those commits as a bad thing to do.

Q/QuantLib/build_tarballs.jl
```diff
@@ -0,0 +1,101 @@
...
+cmake "${CMAKE_FLAGS[@]}" -G Ninja ..
+ninja -j${nproc}
+ninja install
+
+if [[ ${target} == *mingw* ]]; then
+    # QuantLib's install() for ql_library specifies ARCHIVE+LIBRARY but no
+    # RUNTIME, so the import library (.dll.a) is installed to lib/ but the
+    # actual DLL is left in the build tree. Copy it into bin/ ourselves.
+    mkdir -p ${prefix}/bin
+    cp ql/libQuantLib.dll ${prefix}/bin/
```
giordano, line 82:
```suggestion
    install -Dvm 755 ql/libQuantLib.dll -t "${libdir}"
```

alecloudenback:
Update: added Windows (MinGW) support. All 22 builds now green — including both `i686-w64-mingw32-cxx11` and `x86_64-w64-mingw32-cxx11`.

Changes for Windows:
- `-DBUILD_SHARED_LIBS=ON` — overrides upstream's `set(BUILD_SHARED_LIBS \${UNIX})` (guarded by `if (NOT DEFINED ...)`, so the command-line setting wins) to ship a DLL instead of a static lib on MinGW
- `-DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON` — auto-generates the symbol export list since QuantLib doesn't use `__declspec(dllexport)`
- `-DBOOST_ALL_DYN_LINK` on MinGW only — Boost headers use `dllimport` against shared boost_jll
- `-DQL_BUILD_TEST_SUITE=OFF` on MinGW only — `quantlib-test-suite.exe` and `quantlib-benchmark.exe` fail to link because their CMake config doesn't add `libboost_unit_test_framework` to the link line (canonical issue @imciner2 ran into); the library and Examples link fine. Tests/benchmark are never shipped anyway, so this is per-platform only.
- Post-`ninja install` `cp ql/libQuantLib.dll \${prefix}/bin/` on MinGW — upstream `ql/CMakeLists.txt` line 2452 specifies `ARCHIVE+LIBRARY` destinations but no `RUNTIME`, so the import library lands in `lib/` while the actual DLL is left in the build tree. Copy it into `bin/` manually.

Still ready for review @giordano @imciner2.

ViralBShah:
I pushed a commit to this branch to get CI green.

The last run (buildkite #32561) timed out at 5h on `aarch64-apple-darwin` and `armv6l-linux-gnueabihf-cxx11`. Both had finished compiling in ~18 min like every other platform and then hung inside the BinaryBuilder audit while checking the ~40 example executables and the test suite. None of those are shipped, so as @imciner2 suggested the recipe now passes `-DQL_BUILD_EXAMPLES=OFF -DQL_BUILD_TEST_SUITE=OFF` on all platforms (the benchmark lives under `test-suite/`, so it is covered too). That also removes the MinGW-only block and its `QL_BUILD_BENCHMARK` flag, which CMake warned is not a real option, and should roughly halve build time.

The audit also warned on glibc Linux that `libQuantLib.so` links `libgcc_s.so.1` without it being resolvable, so `CompilerSupportLibraries_jll` is added as a dependency.

Earlier review items (`-lrt` only on x86 glibc, `ninja -j${nproc}`, boost compat pin per #3024, `install -Dvm 755` for the DLL) are all in.

ViralBShah:
Also bumped to QuantLib v1.43 (released 2026-07-14, commit 6b57206e). Nothing build-relevant changed upstream: same CMake options, same license file, and the darwin `<vector>` workaround is still needed.

ViralBShah:
Rebased onto master and CI is green on all 20 platforms with QuantLib 1.43 (buildkite #32811), including `aarch64-apple-darwin` and `armv6l-linux-gnueabihf` that timed out before. Builds now take 8-15 min each and the audit reports no warnings. Ready for merge from my side.

PR #7093, opened 2023-07-22
[Clipper2] Update from v1.0.4 to v1.2.2

ferrolho:


C/Clipper2/build_tarballs.jl
```diff
@@ -41,4 +40,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"11.1.0")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"13.1.0")
```
giordano, line 43: GCC 13 [isn't even available](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/master/src/Rootfs.jl#L392-L406)
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"11.1.0")
```

giordano:
Sounds like the shared library isn't built?

C/Clipper2/build_tarballs.jl
ferrolho: Oops, didn't realise that version is not available... Why do you propose v11.1.0 instead of v12.1.0?

giordano: https://github.com/JuliaPackaging/Yggdrasil/blob/5fb343ac5c89a09866678eedcedc1daccb33bebb/CONTRIBUTING.md#compatibility-tips

ferrolho: > _As a general remark, when using GCC as compiler (which is default when targeting Linux and Windows platforms) try to use the oldest versions that is able to compile your code (...) In any case, avoid using GCC 11+ when building C++, as that would be incompatible with Julia v1.6 (...)_

Thanks!

C/Clipper2/bundled/patches/sharedlib.patch
```diff
@@ -1,61 +1,52 @@
...
-index bb1e924..1b1860b 100644
+index e3afbc8..87ffa25 100644
 --- before/Clipper2/CPP/CMakeLists.txt
 +++ after/Clipper2/CPP/CMakeLists.txt
-@@ -25,7 +25,7 @@ set(CLIPPER2_SRC
- )
- 
- # 2d version of Clipper2
--add_library(Clipper2 STATIC ${CLIPPER2_INC} ${CLIPPER2_SRC})
-+add_library(Clipper2 SHARED ${CLIPPER2_INC} ${CLIPPER2_SRC})
```
giordano, line 10: I think this is the library we're missing?

ferrolho: You mean this should be put inside the currently-empty `dependencies = Dependency[]` of the `build_tarballs.jl` file?

imciner2: He means that this patch for this new version should still include these changes to the CMake file to generate the `Clipper2` library.

ferrolho:
I tried running the [BinaryBuilder.jl](https://github.com/JuliaPackaging/BinaryBuilder.jl) wizard locally, but I am getting the following error at the `cmake` stage:
```
sandbox:${WORKSPACE}/srcdir/Clipper2/CPP/build # cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DCLIPPER2_TESTS=OFF -DCLIPPER2_EXAMPLES=OFF -DCLIPPER2_UTILS=OFF ..
Re-run cmake no build system arguments
CMake Error at /usr/share/cmake/Modules/Compiler/._ADSP-DetermineCompiler.cmake:1:
  Parse error.  Expected a command name, got bad character with text "".
Call Stack (most recent call first):
  /usr/share/cmake/Modules/CMakeCompilerIdDetection.cmake:6 (include)
  /usr/share/cmake/Modules/CMakeCompilerIdDetection.cmake:33 (_readFile)
  /usr/share/cmake/Modules/CMakeDetermineCompilerId.cmake:282 (compiler_id_detection)
  /usr/share/cmake/Modules/CMakeDetermineCompilerId.cmake:301 (CMAKE_DETERMINE_COMPILER_ID_WRITE)
  /usr/share/cmake/Modules/CMakeDetermineCompilerId.cmake:6 (CMAKE_DETERMINE_COMPILER_ID_BUILD)
  /usr/share/cmake/Modules/CMakeDetermineCompilerId.cmake:59 (__determine_compiler_id_test)
  /usr/share/cmake/Modules/CMakeDetermineCCompiler.cmake:123 (CMAKE_DETERMINE_COMPILER_ID)
  CMakeLists.txt:2 (project)


-- Configuring incomplete, errors occurred!
See also "/workspace/srcdir/Clipper2/CPP/build/CMakeFiles/CMakeOutput.log".
```

Also, I manually had to create the patch file locally within the Docker container. I wanted to add the `bundled` folder as a `DirectorySource` during the wizard setup, but I couldn't find a way to do it.

ferrolho:
For completeness, the earlier issues were due to a missing `SHARED` keyword on an `add_library` instruction and also a stupid mistake from my part in which I was creating the patch file from a git diff w.r.t. the master branch of Clipper2 instead of w.r.t. the latest tagged version.

Special thanks to @giordano for the help during the very well known JuliaCon @ Boston Logan hackathon.

PR #7107, opened 2023-07-25
[DSSP] Use upstream pre-built binaries on windows

marcom:
Unfortunately the binaries we build for windows in this build script crash on any non-trivial operation, so use pre-built upstream binaries for now.

Also added are `FileProduct`s for more reliable installation of mmcif dictionaries necessary at runtime.

cc: @lmiq 

Ref: https://github.com/m3g/ProteinSecondaryStructures.jl/issues/8

### DIFF
--- a/D/DSSP/build_tarballs.jl
+++ b/D/DSSP/build_tarballs.jl
@@ -14,8 +14,17 @@ sources = [
               "836aed6ea9a227b37e5b0d9cbcb1253f545d0778"), # v5.1.0 (git-tag v5.1.0.1)
     ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
                   "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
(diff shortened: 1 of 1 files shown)


lmiq:
Does this requires any action on ProteinSecondaryStructures? 

Yesterday the package was failing because of the absence of the dictionary files. 

(We can't really use semver here, can we?)

marcom:
I had all tests for ProteinSecondaryStructures (PSS) passing under windows when I installed mkdssp.exe manually from the upstream installer and modified PSS to use it:
```julia
dssp_executable = `C:\\path\\to\\mkdssp.exe`
```

The upstream binary of mkdssp.exe actually has the  mmcif dictionaries embedded in the executable file, but I have included them in this build script here for good measure and then the code in PSS should still work on windows.

For the compat entry for `DSSP_jll` I am not sure if we can use `= 4.4.0+1`? If yes that would solve the problem.

lmiq:
Well, either the compat will download the latest one, or we can set that explicitly (which would be better). Anyway, when this is merged I can release the new version of ProteinSecondaryStructures and try.

PR #7109, opened 2023-07-25
[Dex]: Support groups in oidc ✨

mdpradeep:
With this it will be possible to support groups in OIDC connector and added few important messages in the logs that will help debug on a production system.


giordano:
Thanks for your contribution!

PR #7119, opened 2023-07-27
Added an automated binary builder for mujoco

JamieMair:
Adds automated building of binary artifacts for MuJoCo using the updated open source library from DeepMind.


M/MuJoCo/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "MuJoCo"
+version = v"2.3.7"
+
+# Collection of sources required to build libX11
```
fingolfin, line 8:
```suggestion
# Collection of sources required to build mujoco
```

M/MuJoCo/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/mujoco
+sed -i '/Werror/d' cmake/MujocoOptions.cmake
+CPPFLAGS="-I${prefix}/include"
+CXXFLAGS="-I${prefix}/include"
+apk update
+apk upgrade
```
fingolfin, line 20: Why?

M/MuJoCo/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+sources = [
+    GitSource("https://github.com/deepmind/mujoco", "c9246e1f5006379d599e0bcddf159a8616d31441")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/mujoco
+sed -i '/Werror/d' cmake/MujocoOptions.cmake
+CPPFLAGS="-I${prefix}/include"
+CXXFLAGS="-I${prefix}/include"
```
fingolfin, line 18: Hmm, I thought the compiler automatically searched in these directories anyway? Why are they necessary here?

M/MuJoCo/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+apk update
+apk upgrade
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release .
+cmake --build .
+cmake --install .
+"""
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [p for p in supported_platforms() if !Sys.isapple(p) || p.tags["arch"] != "aarch64"]
```
fingolfin, line 27: The upstream repository ships macOS binaries as well as aarch64 binaries, so why are they excluded here? Compilation trouble?

M/MuJoCo/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+    BuildDependency("GLFW_jll"),
+    BuildDependency("Xorg_libXrandr_jll"),
+    BuildDependency("Libglvnd_jll"),
+    BuildDependency("Xorg_libXi_jll"),
+    BuildDependency("Xorg_libXinerama_jll"),
+    BuildDependency("Xorg_libXcursor_jll")
+]
+
+# Build the tarballs.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"11.1.0")
```
fingolfin, line 52: That's a really new GCC version, which may limit who can use the resulting JLL (I think?). Is there a specific reason to pick that instead of an older one?

fingolfin (commented):
Hi there, left a few comments meant to be helpful, but feel free to ignore me.

M/MuJoCo/build_tarballs.jl
JamieMair: It didn't compile without these unfortunately.

M/MuJoCo/build_tarballs.jl
JamieMair: Yeah, we ran into compilation errors with aarch on Apple, and we plan to come back to this once we have some updated bindings.

M/MuJoCo/build_tarballs.jl
JamieMair: The libraries require C++ 17, but it seems like GCC 8+ supports this so perhaps it will work with that? We picked this one as we knew it worked on my own machine.

M/MuJoCo/build_tarballs.jl
JamieMair: Removed! it was an artifact of and older install script

JamieMair:
We are going to try as many platforms supported as possible today, and then we will push all the changes to finish this off. This should probably be considered a WIP until the next few commits.

M/MuJoCo/build_tarballs.jl
imciner2: > The libraries require C++ 17

Do you know which features are C++17? The language features can be found in GCC 7, but some library features are GCC 8 or GCC 9.

JamieMair: I tried GCC 8 and it wouldn't compile on any platforms, I don't know the specific features that is required, they just say C++ 17 in the docs. Actually, we are having quite a lot of issues cross-compiling for Mac and Windows as their cmake file doesn't seem to handle cross-compilation very well. We know this compiles fine natively on macos and they also have windows binaries too, but we will work on creating the bindings for it, and get back to adding other platform support soon. For now, we have all non-aarch64 builds on Linux.

traversaro: > Actually, we are having quite a lot of issues cross-compiling for Mac and Windows as their cmake file doesn't seem to handle cross-compilation very well.

Just FYI, I mantain the mujoco recipe on conda-forge ( https://github.com/conda-forge/mujoco-feedstock ) and there we are able to cross-compile mujoco on arm64 on Linux and macOS. Unfortunatly we have quite a few local patches at the moment, and perhaps the problem you are seeing could be solved by some of those patches. Indeed we would like to fix upstream as much as possible (see https://github.com/conda-forge/mujoco-feedstock/issues/25 and https://github.com/deepmind/mujoco/pull/937).

JamieMair: @traversaro  Thanks! That's very helpful, we'll take a look soon!

JamieMair:
@fingolfin I have just managed to get compilation on all of the following platforms:
1. Linux x86_64 {libc=glibc}
2. Linux aarch64 {libc=glibc}
3. Linux powerpc64le {libc=glibc}
4. macOS x86_64
5. macOS aarch64
6. Windows x86_64

This should match the support offered by the upstream library (see [releases](https://github.com/deepmind/mujoco/releases/tag/2.3.7)).

@traversaro Thanks for the tips on the patches and @imciner2 thanks for the suggestion.

M/MuJoCo/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+CPPFLAGS="-I${prefix}/include"
+CXXFLAGS="-I${prefix}/include"
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release .
+cmake --build .
+cmake --install .
+"""
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+function platform_filer(p)
+    return contains(p.tags["arch"], "64") && (!haskey(p.tags, "libc") || p.tags["libc"] != "musl") && (!Sys.isbsd(p) || Sys.isapple(p))
```
fingolfin, line 33: Please don't use p.tags like this, and in particular don't try to parse the `arch` value -- instead use the appropriate tester functions like `nbits(p) == 64`. You can find plenty examples by grepping around in other recipes

M/MuJoCo/patches/mingw.patch
```diff
@@ -0,0 +1,149 @@
...
++++ b/CMakeLists.txt
+@@ -38,10 +38,10 @@ enable_language(CXX)
+ 
+ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+ 
+-option(MUJOCO_BUILD_EXAMPLES "Build samples for MuJoCo" ON)
+-option(MUJOCO_BUILD_SIMULATE "Build simulate library for MuJoCo" ON)
+-option(MUJOCO_BUILD_TESTS "Build tests for MuJoCo" ON)
+-option(MUJOCO_TEST_PYTHON_UTIL "Build and test utility libraries for Python bindings" ON)
++option(MUJOCO_BUILD_EXAMPLES "Build samples for MuJoCo" OFF)
```
fingolfin, line 28: I wonder why these options are patched to "off" instead of just setting them to "off" via command line?

M/MuJoCo/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "MuJoCo"
+version = v"2.3.7"
+
+# Collection of sources required to build mujoco
+sources = [
+    GitSource("https://github.com/deepmind/mujoco", "c9246e1f5006379d599e0bcddf159a8616d31441"),
+    DirectorySource(joinpath(@__DIR__, "patches"))
```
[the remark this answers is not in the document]
JamieMair: Agreed, I have changed to use the "bundled" approach.

M/MuJoCo/build_tarballs.jl
JamieMair: Agreed, this wasn't the best approach. Instead, I have just listed the platforms explicitly, as this is much easier to read.

M/MuJoCo/patches/mingw.patch
JamieMair: I have switched to the command line version.

M/MuJoCo/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+if [[ "${target}" == *-mingw* ]]; then
+    atomic_patch -p1 ../patches/mingw.patch
+elif [[ "${target}" == *-apple-darwin* ]]; then
+    atomic_patch -p1 ../patches/macos.patch
+else
+    atomic_patch -p1 ../patches/other.patch
+fi
+CPPFLAGS="-I${prefix}/include"
+CXXFLAGS="-I${prefix}/include"
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DMUJOCO_BUILD_EXAMPLES=OFF -DMUJOCO_BUILD_SIMULATE=OFF -DMUJOCO_BUILD_TESTS=OFF -DMUJOCO_TEST_PYTHON_UTIL=OFF .
```
fingolfin, line 26: Optional suggestion, feel free to ignore, but: I personally like to split these into multiple lines, as that makes it (a) more readable, and (b) if any arguments are added/modified/removed later on, it is easier to see in the diff what changed and how
```suggestion
cmake \
  -DCMAKE_INSTALL_PREFIX=$prefix \
  -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
  -DCMAKE_BUILD_TYPE=Release \
  -DMUJOCO_BUILD_EXAMPLES=OFF \
  -DMUJOCO_BUILD_SIMULATE=OFF \
  -DMUJOCO_BUILD_TESTS=OFF \
  -DMUJOCO_TEST_PYTHON_UTIL=OFF \
  .
```

JamieMair: Thanks for the suggestion, this is much better!

M/MuJoCo/bundled/patches/mingw.patch
```diff
@@ -0,0 +1,149 @@
...
+ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+ 
+-option(MUJOCO_BUILD_EXAMPLES "Build samples for MuJoCo" ON)
+-option(MUJOCO_BUILD_SIMULATE "Build simulate library for MuJoCo" ON)
+-option(MUJOCO_BUILD_TESTS "Build tests for MuJoCo" ON)
+-option(MUJOCO_TEST_PYTHON_UTIL "Build and test utility libraries for Python bindings" ON)
++option(MUJOCO_BUILD_EXAMPLES "Build samples for MuJoCo" OFF)
++option(MUJOCO_BUILD_SIMULATE "Build simulate library for MuJoCo" OFF)
++option(MUJOCO_BUILD_TESTS "Build tests for MuJoCo" OFF)
++option(MUJOCO_TEST_PYTHON_UTIL "Build and test utility libraries for Python bindings" OFF)
```
fingolfin, line 31: still necessary ?

M/MuJoCo/bundled/patches/macos.patch
```diff
@@ -0,0 +1,389 @@
...
++enable_language(OBJCXX)
++
+ 
+ option(SIMULATE_BUILD_EXECUTABLE "Build the simulate executable binary." ON)
+ option(SIMULATE_GLFW_DYNAMIC_SYMBOLS "Whether to resolve GLFW symbols dynamically." OFF)
+@@ -136,10 +136,10 @@ target_compile_options(libsimulate PRIVATE ${MUJOCO_SIMULATE_COMPILE_OPTIONS})
+ target_link_libraries(libsimulate PUBLIC lodepng mujoco::platform_ui_adapter mujoco::mujoco)
+ target_link_options(libsimulate PRIVATE ${MUJOCO_SIMULATE_LINK_OPTIONS})
+ 
+-if(APPLE)
```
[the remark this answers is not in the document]
imciner2: I wouldn't trust it. Internally, CMake has some unset/set calls for `APPLE`, so those might override the variable in certain cases

@JamieMair do you know why it needs to be patched to ignore the check? I thought it should have already been set on the macOS targets.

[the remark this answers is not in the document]
JamieMair: I was having issues when compiling the other targets that has now been set to "off", but I didn't remove them. Now I should have cleaned up the patch files so that the only changes remaining are necessary.

M/MuJoCo/bundled/patches/mingw.patch
JamieMair: No, it has been removed

M/MuJoCo/bundled/patches/macos.patch
```diff
@@ -0,0 +1,389 @@
...
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -38,15 +38,14 @@ enable_language(CXX)
+ 
+ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+ 
+-option(MUJOCO_BUILD_EXAMPLES "Build samples for MuJoCo" ON)
+-option(MUJOCO_BUILD_SIMULATE "Build simulate library for MuJoCo" ON)
+-option(MUJOCO_BUILD_TESTS "Build tests for MuJoCo" ON)
+-option(MUJOCO_TEST_PYTHON_UTIL "Build and test utility libraries for Python bindings" ON)
```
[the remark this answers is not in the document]
JamieMair: Removed :)

M/MuJoCo/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+#     Platform("aarch64", "linux", libc="glibc"),
+#     Platform("powerpc64le", "linux", libc="glibc"),
+#     Platform("x86_64", "macos"),
+#     Platform("aarch64", "macos"),
+#     Platform("x86_64", "windows")
+# ]
+platforms = [
+    Platform("x86_64", "macos"),
+    Platform("aarch64", "macos"),
+    Platform("x86_64", "windows")
```
fingolfin, line 51: You dropped Linux now?

JamieMair: Ah good spot, this was a mistake from testing the patch changes.

JamieMair:
@fingolfin Do you think this PR is ready to go in? Is there anything else I should do?

M/MuJoCo/build_tarballs.jl
```diff
@@ -0,0 +1,68 @@
...
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux", libc="glibc"),
+    Platform("aarch64", "linux", libc="glibc"),
+    Platform("powerpc64le", "linux", libc="glibc"),
+    Platform("x86_64", "macos"),
+    Platform("aarch64", "macos"),
+    Platform("x86_64", "windows")
+]
```
fingolfin, line 47: I still wonder why you are doing this instead of, say
```suggestion
platforms = filter!(p -> nbits(p) == 64, supported_platforms())
```
so that e.g. Linux with muslc and possibly FreeBSD are also supported... But you may have your reasons (albeit undocumented), so, your decision *shrug*

M/MuJoCo/build_tarballs.jl
fingolfin: Here, too, I still wonder what the actual errors are with older GCC... But if works for you as-is for now, let's move ahead anyway (this can be revisited in a future update, if desired)

fingolfin (approved):

PR #7122, opened 2023-07-28
[rustfft] version 0.2.0

Taaitaaiger:


Taaitaaiger:
This looks like a Julia issue. `small_typeof` is called by `jl_type`, and I think it needs to be exported to work correctly on Windows.

giordano:
Mind opening an issue?

Taaitaaiger:
Of course not, I just think I should reproduce it in isolation first to confirm it's not an issue in jlrs

PR #7118, opened 2023-07-26
Bump MParT to 2.1.0

dannys4:

PR #7126, opened 2023-07-29
Bump Kokkos to 3.7.2

dannys4:
Dual purpose to bump version and rebuild package on FreeBSD (see #7118 )


giordano:
You probably wanted to branch out of `master` instead of your MparT branch

dannys4:
I know I deleted the patch in `bundled`, but it is now unnecessary in this version of Kokkos, due to the code now at kokkos `core/src/Kokkos_Macros.cpp:679`

PR #7127, opened 2023-07-29
relax LibCURL_jll compat to include v8

### LINKED ISSUE #6857: Relax `LibCURL` version bounds

visr:
Fixes #6857.
Already fixed in the registry in https://github.com/JuliaRegistries/General/pull/88557.
[skip build]


giordano:
Thanks! :pray:

PR #7129, opened 2023-07-29
Rebuild zlib, nghttp2, and libssh2

ararslan:
The saga continues (#7036, #7121, #7125) to build all of Julia's binary dependencies with FreeBSD 13.2


ararslan:
Note that Julia is currently using LibSSH2_jll v1.10.2 but this rebuilds v1.11.0. So as part of upgrading the build numbers for the stdlib JLLs I guess I'll also upgrade LibSSH2_jll to v1.11.0. Assuming libssh2 follows semver, it should be safe.

PR #7133, opened 2023-07-29
Rebuild GMP, libgit2, and libcurl

ararslan:
I feel like Julia didn't always have quite so many binary dependencies to update 🤔

### DIFF
--- a/G/GMP/GMP@6.2.1/build_tarballs.jl
+++ b/G/GMP/GMP@6.2.1/build_tarballs.jl
@@ -6,4 +6,4 @@ include("../common.jl")
 build_tarballs(ARGS, configure(version)...;
                preferred_gcc_version=v"6", julia_compat="1.7")
 
-# Trigger build!!
+# Trigger build!!!
--- a/L/LibCURL/LibCURL@8/build_tarballs.jl
+++ b/L/LibCURL/LibCURL@8/build_tarballs.jl
@@ -2,4 +2,4 @@ include("../common.jl")
 
 build_libcurl(ARGS, "LibCURL", v"8.0.1")
 
-# Build trigger: 1
+# Build trigger: 2


giordano:
> I feel like Julia didn't always have quite so many binary dependencies to update thinking

Maybe 10 years ago?

ararslan:
Hm, shouldn't LibGit2 be picking up the updated LibSSH2? The `compat` for the dependency is listed as 1.10.1 but that should be compatible with the recently rebuilt 1.11.0, right?

G/GMP/GMP@6.2.1/build_tarballs.jl
```diff
@@ -6,4 +6,4 @@ include("../common.jl")
 build_tarballs(ARGS, configure(version)...;
                preferred_gcc_version=v"6", julia_compat="1.7")
 
-# Trigger build!!
+# Trigger build!!!
```
fingolfin, line 9: just out of curiosity: why do we need a rebuild?

ararslan: Our toolchain for FreeBSD is now based on FreeBSD 13.2 rather than 12.2 (https://github.com/JuliaPackaging/Yggdrasil/pull/6865). Julia's CI is running 13.2 (https://github.com/JuliaCI/julia-buildkite/pull/266) so I want to update all of Julia's binary dependencies to use the FreeBSD 13.2 toolchain. Once everything is updated, I'm going to submit a PR to Julia to update the build numbers for the JLL dependencies.

ararslan:
I dropped the commits related to libgit2 because it was using the old libssh2 and was failing on Windows whereas GMP and libcurl were fine, so we might as well move forward with those and address libgit2 separately.

PR #7136, opened 2023-07-30
Upgrade x86_64 and i686 linux GCC shares to Glibc v2.17+

staticfloat:
This is necessary to build for LLVM 16 and libuv.

### DIFF
--- a/0_RootFS/gcc_common.jl
+++ b/0_RootFS/gcc_common.jl
@@ -332,11 +332,16 @@ function gcc_script(compiler_target::Platform)
             atomic_patch -p1 ${p} || true;
         done
 
+        # Patch bad `movq` argument in glibc 2.17, adapted from:
+        # https://github.com/bminor/glibc/commit/b1ec623ed50bb8c7b9b6333fa350c3866dbde87f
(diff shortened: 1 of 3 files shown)


.ci/Manifest.toml
```diff
@@ -1,7 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.3"
+julia_version = "1.9.2"
```
giordano, line 3: I don't think we should change the version of Julia here, there are some tests that are failing for Julia v1.8+ in BBB

0_RootFS/gcc_common.jl
```diff
@@ -332,11 +332,14 @@ function gcc_script(compiler_target::Platform)
             atomic_patch -p1 ${p} || true;
         done
 
+        # Patch bad `movq` argument in glibc 2.17
```
giordano, line 335: Can you please add a comment with a link to where the the patch was found, for future reference?

giordano (changes requested):
Looks overall good, but we shouldn't touch the manifest here

giordano (approved):

PR #7137, opened 2023-07-30
[devspace] add recipe

### LINKED ISSUE #6536: request: devspace
This is mostly (entirely?) written in go and has a automated multi-os/arch build/release  script here: https://github.com/devspace-sh/devspace/blob/main/hack/build-all.bash

M-PERSIC:
[devspace](https://github.com/devspace-sh/devspace.git) v6.3.2

closes #6536

### DIFF
--- a/D/devspace/build_tarballs.jl
+++ b/D/devspace/build_tarballs.jl
@@ -1,0 +1,46 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message
+using BinaryBuilder, Pkg
+
+name = "devspace"
+version = v"6.3.2"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource(
+        "https://github.com/devspace-sh/devspace.git",
+        "294a386008d52a6ed58448f3c73e5f3ef3caeb00",
+    ),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/devspace
+install_license LICENSE
+mkdir -p ${bindir}
+go build -o ${bindir}
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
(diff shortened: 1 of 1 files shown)


D/devspace/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
...
+cd ${WORKSPACE}/srcdir/devspace
+install_license LICENSE
+mkdir -p ${bindir}
+go build -o ${bindir}
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = filter(
+    p -> arch(p) != "powerpc64le" && !((arch(p) == "armv6l" || arch(p) == "armv7l" || arch(p) == "aarch64") && !(Sys.isapple())),
```
giordano, line 27: Is it easier to say
```suggestion
    p -> proc_family(p) == "intel",
```
?  I'm not sure what you wanted to do with `!(Sys.isapple())`, since that always returns `true` (it's checking that Julia's host platorm is not macOS) and build for macOS is successful

M-PERSIC: Devspace does not support arm64 builds outside of non-macos platforms. There is no mention of armv6l, armv7l, or i686 support, however they do build successfully. In this case, `proc_family` works perfectly and I can add `&& (arch(p) == "aarch64" && Sys.isapple(p))` to include arm64 support for darwin.

fingolfin: Better include something -- at worst it is broken (then someone can figure that out and report it, and perhaps it can be fixed next time around), at worst it just works. I can see no serious downsides.

Also, the check using `Sys.isapple()` is just wrong. It tests the platform on which the builder is executed (which is always Linux). I am guessing you mean `Sys.isapple(p)` ?

But I would recommend to just try building it for all platforms first. And then only remove platforms that don't work.

M-PERSIC: Ahhh okay yep I meant to write `Sys.isapple(p)`, that's a mistake on my part. I'll retry for all supported platforms and add them if they can be built.

PR #7144, opened 2023-08-01
[abPOA] add recipe

M-PERSIC:
- **Name**: abPOA
- **Version**: 1.4.1 (16c64e5163cbd0502792ca20f1069fe40c929b0d)
- **Source**: https://github.com/yangao07/abPOA.git
- **Language(s)**: C
- **Build System**: CMake
- **Depends**: SIMDe (JLL available: yes), Zlib (yes)
- **Depends (optional)**: N/A
- **Purpose**:  BioJulia community request
- **Description**: "abPOA: an SIMD-based C library for fast partial order alignment using adaptive band"

Patched out the -march option (disallowed in BinaryBuilder) and enabled shared libraries

### DIFF
--- a/A/abPOA/build_tarballs.jl
+++ b/A/abPOA/build_tarballs.jl
@@ -34,14 +34,17 @@ cmake --build . --target install
 platforms = filter(p -> arch(p) == "x86_64", supported_platforms(; exclude = Sys.iswindows))
 
 # The products that we will ensure are always built
-products = [ExecutableProduct("abpoa", :abpoa), LibraryProduct("libabpoa", :libabpoa)]
+products = [
+    ExecutableProduct("abpoa", :abpoa), 
+    LibraryProduct("libabpoa", :libabpoa)
+]
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
     BuildDependency("SIMDe_jll"),
     Dependency(
         PackageSpec(; name = "Zlib_jll", uuid = "83775a58-1f1d-513f-b197-d71354ab007a");
-        compat = "1.2.13",
+        compat = "1.2.12",
     ),
 ]
 
--- a/A/abPOA/build_tarballs.jl
+++ b/A/abPOA/build_tarballs.jl
@@ -1,0 +1,61 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message
+using BinaryBuilder, Pkg
+
+name = "abPOA"
+version = v"1.4.1"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource(
+        "https://github.com/yangao07/abPOA.git",
+        "c6c9dacf92414cd1a358cbd1d28b6f2ca37b30ed",
+    ),
+    DirectorySource("./bundled"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/abPOA
+install_license LICENSE
+atomic_patch -p1 ../patches/no_march_native.patch
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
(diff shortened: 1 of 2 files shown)


A/abPOA/build_tarballs.jl
```diff
@@ -0,0 +1,59 @@
...
+    name,
+    version,
+    sources,
+    script,
+    platforms,
+    products,
+    dependencies;
+    julia_compat = "1.6",
+    preferred_gcc_version = v"6",
+)
```
giordano, line 59: Missing newline 
```suggestion
)

```

A/abPOA/bundled/patches/no_march_and_shared_libs_on.patch
```diff
@@ -0,0 +1,16 @@
...
++++ b/CMakeLists.txt
+@@ -10,10 +10,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
+ 
+ set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
+ 
+-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=native") # SIMD
+-
+ # build abPOA as a static library by default
+-set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build all libraries as shared")
++set(BUILD_SHARED_LIBS ON CACHE BOOL "Build all libraries as shared")
```
giordano, line 13: Can't you set this option on the command line instead of changing it in the code?

M-PERSIC: Yeah unfortunately this build is causing me a lot of issues (as evidenced with all that red below :/) so somewhere I tried patching to see if something would work. I'll come back to this pr over the week.

imciner2:
Try this patch to fix the linking on x86_64-apple-darwin: https://github.com/yangao07/abPOA/pull/51.

A/abPOA/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+
+# The products that we will ensure are always built
+products = [ExecutableProduct("abpoa", :abpoa), LibraryProduct("libabpoa", :libabpoa)]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency("SIMDe_jll"),
+    Dependency(
+        PackageSpec(; name = "Zlib_jll", uuid = "83775a58-1f1d-513f-b197-d71354ab007a");
+        compat = "1.2.13",
```
imciner2, line 44: Is there a reason you need this tight a version bound on zlib?

M-PERSIC: image
It's better to have a tighter compat bound than no bound as it might happen that another contributor might try to update this build recipe with a Zlib 2.x that breaks compatibility. Whenever we bump to a next version we usually try to update all of the dependencies anyway.

giordano: I don't think that really answers the question. Restricting zlib version by extension means restricting Julia versions. I believe it'd be better to relax this compat at least what was used in Julia v1.6

M-PERSIC: I see. In this case, based on https://github.com/JuliaPackaging/Yggdrasil/pull/2031, should I set compat to v1.2.12?

giordano: To determine the version of stdlibs in Julia v1.6.0 I usually look at their Project.toml file: https://github.com/JuliaLang/julia/blob/f9720dc2ebd6cd9e3086365f281e62506444ef37/stdlib/Zlib_jll/Project.toml#L3

M-PERSIC: That's good advice, thank you! Looks like I can set compat to 1.2.12

M-PERSIC:
@giordano If this compat is fine, then this should be ready to go!

PR #7145, opened 2023-08-01
[WFA2] add recipe

M-PERSIC:
- **Name**: WFA2-lib
- **Version**: 2.3.3 (d1116831a3514fff39321a73b8b526f17f7a2c18)
- **Source**: https://github.com/smarco/WFA2-lib.git
- **Language(s)**: C
- **Build System(s)**: CMake, Make (former chosen)
- **Depends**: None
- **Depends (optional)**: CompilerSupportLibraries (non-BSD platforms, JLL available: yes), LLVMOpenMP (BSD, yes) for OpenMP support
- **Platform(s)**: All excluding Windows (see below)
- **Purpose**:  BioJulia community request
- **Description**: "Wavefront alignment algorithm library v2 "

Will look into attempting a future pr to work around missing sys/mman headers for MingW

### DIFF
--- a/W/WFA2/build_tarballs.jl
+++ b/W/WFA2/build_tarballs.jl
@@ -1,0 +1,74 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message
+using BinaryBuilder, Pkg
+
+name = "WFA2"
+version = v"2.3.3"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource(
+        "https://github.com/smarco/WFA2-lib.git",
+        "d1116831a3514fff39321a73b8b526f17f7a2c18",
+    ),
+    DirectorySource("./bundled"),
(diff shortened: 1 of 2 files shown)


giordano:
sys/mman is probably a Unix only header file. Does this package support Windows at all? Or they are just using some preprocessor guards incorrectly?

M-PERSIC:
Yep, sys/mman is Unix-only, had this issue come up with another build that I think went through but I dropped Windows support because of it. 

 I did check for guards but none in this software are platform-specific, only compiler or feature-specific. Almost no mention of Windows at all in the code, not even in any issue or pull request to be honest. I was thinking of filing an upstream issue, and there is also a mirror of an open-source mman library for Windows called [mman-win32](https://github.com/alitrack/mman-win32) that might prove to be plug-and-play that I will look into. 
 
 For now, we can include most other platforms and I can come back to building Windows.

PR #7152, opened 2023-08-02
[OpenCV] Rebuild JLL for 1.10 compatibility

stemann:
Cf. https://github.com/JuliaImages/OpenCV.jl/issues/23


stemann:
> You could also add support for julia 1.11 nightlies by adding it to the `julia_versions` [line](https://github.com/JuliaPackaging/Yggdrasil/pull/7152/files#diff-6af99b6147c9a6bdcccb510174ee5e0e0ac16468cfd84d617474573930bc9084R15), or by removing that line since the version list is set by `libjulia/common.jl` automatically if it doesn't exist (https://github.com/JuliaPackaging/Yggdrasil/blob/master/L/libjulia/common.jl#L10).

Removed the line.

vchuravy:
I would almost recommend moving the libjulia dependent part into a separate jll

PR #7154, opened 2023-08-02
[micromamba] update to  to v1.4.9

behinger:
This is the first time I am trying to update an Yggdrasil binary, I hope I got it right :)

with v1.4.7 I ran into severe issues[1], even on github actions, where PythonCall simply stops working. According to the issue-tracker this issue should be fixed in 1.4.8 - but 1.4.9 is the newest..

[1] https://github.com/mamba-org/mamba/issues/2663

tagging @tylerjthomas9 because you did the earlier releases

### DIFF
--- a/M/micromamba/build_tarballs.jl
+++ b/M/micromamba/build_tarballs.jl
@@ -3,29 +3,29 @@
 using BinaryBuilder
 
 name = "micromamba"
-version = v"1.4.7"
+version = v"1.4.9"
 build = "0"
 
 # Collection of sources required to build micromamba
 # These are actually just the conda packages for each platform
 sources = [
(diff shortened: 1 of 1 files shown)


tylerjthomas9:
Wow that's a massive issue. v1.4.3 had some issues with windows + pip. Hopefully these issues stop popping up this frequently. It's probably worth delaying builds for future releases until we know that they're stable. Thank you for creating this PR.

behinger:
yeah, that make sense. Is this ready to be merged or can I still do something?

tylerjthomas9:
I pinged the slack to see if anyone with merge rights can get this pushed.

PR #7168, opened 2023-08-05
[openblas] Build the BFloat16 kernels in OpenBLAS

ViralBShah:
Build the BFloat16 kernels listed in https://github.com/xianyi/OpenBLAS/wiki/OpenBLAS-Extensions


ViralBShah:
@giordano @staticfloat It would be really nice to be able to build with BFloat16. I've bumped to the highest GCC, but still failing all over the place.

imciner2:
The x86_64 errors are a build system problem, upstream fix in https://github.com/xianyi/OpenBLAS/pull/4192. It doesn't apply cleanly to the 0.3.23 source though, but I have a local patch that adds it (tested on the `x86_64-apple-darwin-libgfortran3` and `x86_64-unknown-freebsd-libgfortran5` targets, but it should work on the other x86_64 targets where the problem was the cooperlake detection). I can't push the patch to the branch though, how would you like me to add it?

ViralBShah:
My PR is actually quite small. So I'll close it - can you open a new PR?

PR #7197, opened 2023-08-13
Update CImGuiPack

JamesWrigley:
Other changes:
- Use GCC 5, the default GCC 4 was giving compiler errors.
- ~~Enable 32bit vertex indices, as recommended by ImPlot: https://github.com/epezent/implot#extremely-important-note~~

~~*Do not merge this yet!* I'm opening it for transparency, https://github.com/JuliaImGui/cimgui-pack/pull/1 needs to be merged first and the repo link in `build_tarballs.jl` set back to the original repo instead of my fork.~~

CC @Gnimuc

### DIFF
--- a/C/CImGuiPack/build_tarballs.jl
+++ b/C/CImGuiPack/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "CImGuiPack"
-version = v"0.1.2"
+version = v"0.1.3"
 
 # Collection of sources required to build CImGuiPack
 sources = [
     GitSource("https://github.com/JuliaImGui/cimgui-pack.git",
-              "ecc93265e2f0f916462a2bef427f5f6b024ad6b2")
+              "af96d7d92c8b430af35e4b87c1260b234ccf9b5a")
 ]
 
 # Bash recipe for building across all platforms
@@ -37,4 +37,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"5")


JamesWrigley:
Ok, I think this is ready to be merged now :crossed_fingers:

C/CImGuiPack/build_tarballs.jl
```diff
@@ -3,20 +3,20 @@
...
+              "af96d7d92c8b430af35e4b87c1260b234ccf9b5a")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/cimgui-pack
 git submodule update --init --recursive
 mkdir build && cd build
-cmake .. -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+cmake .. -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DImDrawIdx='unsigned int'
```
imciner2, line 19: Adding `-DImDrawIdx='unsigned int'` in the CMake command like this doesn't automatically append it to the CFLAGS/CXXFLAGS that are used by the compiler, this just passes it into CMake. So therefore you aren't actually compiling your source file with this flag currently. You can see that CMake is just ignoring it if you look into the build logs and see:

```
-- Configuring done

CMake Warning:
  Manually-specified variables were not used by the project:

    ImDrawIdx

Build files have been written to: /workspace/srcdir/cimgui-pack/build
```

You should either add another `add_definitions` statement in your CMakeLists.txt (since I guess you control it and you aren't using the upstream one), or manually set it on the command line by doing `-DCMAKE_CXX_FLAGS="-DImDrawIdx='unsigned int" -DCMAKE_C_FLAGS="-DImDrawIdx='unsigned int"` (I would say the first is probably the best way).

imciner2 (changes requested):

imciner2 (approved):

wsphillips:
@Gnimuc are we good to go?

PR #7210, opened 2023-08-15
[umesimd] add recipe

M-PERSIC:
- **Name**: umesimd
- **Version**: 0.8.1 (6f72452743151b76ecb9ddf4dd696d263b45cb74)
- **Source**: https://github.com/edanor/umesimd.git
- **Language(s)**: C++
- **Build System**: CMake
- **Depends**: N/A
- **Depends (optional)**: N/A
- **Purpose**: Dependency for another package
- **Description**: "UME::SIMD A library for explicit simd vectorization."

Header-only library. Set to last merged commit

### DIFF
--- a/U/umesimd/build_tarballs.jl
+++ b/U/umesimd/build_tarballs.jl
@@ -1,0 +1,51 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message
+using BinaryBuilder, Pkg
+
+name = "umesimd"
+version = v"0.8.1"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource(
+        "https://github.com/edanor/umesimd.git",
+        "6f72452743151b76ecb9ddf4dd696d263b45cb74",
+    ),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/umesimd
+install_license LICENSE 
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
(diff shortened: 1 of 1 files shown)


U/umesimd/build_tarballs.jl
```diff
@@ -0,0 +1,51 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/umesimd
+install_license LICENSE 
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+      -DCMAKE_BUILD_TYPE=Release ..
+cmake --build . --parallel ${nproc} --target all
```
imciner2, line 25: Is this CMake invocation actually needed? The CMake seems to just have configure-time logic and install statements, so there doesn't appear to be anything to actually build.

M-PERSIC: If you mean `cmake --build` that's just an alternative to running make. I did overlook that there is no `all` target and I can simplify the script a bit so I'll push another commit (`--target all` will move the files to the $includedir).

imciner2: > If you mean cmake --build that's just an alternative to running make.

Well, yes, I know that, this isn't my first CMake rodeo (and it isn't just a make alternative, it is a build system driver that will also launch Ninja builds, Visual Studio solution builds, etc., and is basically the only way to have an actual cross-platform executable build script without caring about the underlying OS/build system).

> I did overlook that there is no all target and I can simplify the script

This was the main point of the comment - there wasn't anything for that particular build command to actually do, since install is excluded from the all target and no compilation steps were configured in the CMake, so thank you for fixing that.

M-PERSIC: Indeed, thank you for your review. As evidenced by my other pull requests, build systems are not my domain :P

PR #7228, opened 2023-08-19
[NEW] Add Siconos_jll build files

avik-pal:


S/Siconos/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd siconos/
+mkdir build
+cd build/
+python3 -m pip install wheel packaging
+cmake -DUSER_OPTIONS_FILE=../../only_numerics.cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DWITH_FCLIB=0 -DWITH_PYTHON_WRAPPER=0 -DWITH_CXX=0 ..
```
imciner2, line 21: You specify the `WITH_CXX`, `WITH_PYTHON_WRAPPER` and `WITH_FCLIB` parts in the custom options file already, so do they really need to be passed on the command line as well?

Also, the values of `WITH_FCLIB`  and `WITH_CXX` here are different from the ones in the `only_numerics.cmake` file in the tree, so these will take priority over those, should you just update the file with these values?

S/Siconos/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+make -j${nproc}
+make install
+"""
```
imciner2, line 24:
```suggestion
install_license ${WORKSPACE}/srcdir/siconos/COPYING
"""
```

You need to install the license file.

imciner2 (changes requested):

imciner2 (approved):

PR #7230, opened 2023-08-20
[libpolars] Add libpolars

Pangoraw:
The source is from https://github.com/pangoraw/polars.jl/tree/main/c-polars and will be used to power the Julia package at the same location.

### DIFF
--- a/L/libpolars/build_tarballs.jl
+++ b/L/libpolars/build_tarballs.jl
@@ -29,6 +29,8 @@ if [[ "${target}" == x86_64-apple-darwin* ]]; then
     popd
 fi
 
+ls ${WORKSPACE}/srcdir
+ls ${WORKSPACE}/srcdir/Polars.jl/
 cd ${WORKSPACE}/srcdir/Polars.jl/c-polars/
 
 cargo build --release
--- a/L/libpolars/build_tarballs.jl
+++ b/L/libpolars/build_tarballs.jl
@@ -29,10 +29,7 @@ if [[ "${target}" == x86_64-apple-darwin* ]]; then
     popd
 fi
 
-ls ${WORKSPACE}
-ls ${WORKSPACE}/srcdir
-ls ${WORKSPACE}/srcdir/Polars.jl/
-cd ${WORKSPACE}/srcdir/Polars.jl/c-polars/
+cd ${WORKSPACE}/srcdir/c-polars/
 
 cargo build --release
 
--- a/L/libpolars/build_tarballs.jl
+++ b/L/libpolars/build_tarballs.jl
@@ -7,17 +7,34 @@ version = v"0.1.0"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/Pangoraw/Polars.jl/",
-              "dd00d8db4e5ddfa19cb4fdd43df48957491052a5"),
+    GitSource(
+        "https://github.com/Pangoraw/Polars.jl/",
+        "da4a4569a8ac7368aab7ae1c589b768b1e5ec36c",
+    ),
+    ArchiveSource(
+        "https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.13.sdk.tar.xz",
+        "a3a077385205039a7c6f9e2c98ecdf2a720b2a819da715e03e0630c75782c1e4",
+    ),
 ]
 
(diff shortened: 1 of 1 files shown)


matbesancon:
in best practices: the library should maybe build using the original repo https://github.com/pola-rs/polars to be able to control the version of the original polar-rs?
In that case, the C header could be added directly to the Rust repo:
https://github.com/Pangoraw/Polars.jl/blob/main/c-polars/include/polars.h

Pangoraw:
The polars version can be specified in [the Cargo.toml file of `c-polars`](https://github.com/Pangoraw/Polars.jl/blob/9ad0fea96a4ba0d367a70bd48e643df54671b153/c-polars/Cargo.toml#L10-L13). I do not specifically plan on upstreaming `c-polars` to limit the friction needed to make changes to `c-polars` for Polars.jl so maybe it can use its own versionning (similarly to [`libcrossterm`](https://github.com/JuliaPackaging/Yggdrasil/pull/7239) for example) independent of polars versions.

matbesancon:
I see. It's just a bit odd for the wrapper that will depend on the JLL to contain the source for the JLL that will be built (sort of a circular dependency).
Could c-polars have maybe its own repository?

For the dependency on polars itself:
```
[dependencies.polars]
git = "https://github.com/pola-rs/polars"
rev = "c6a301e3141d40c6c6febdebe0b95632ac8da2ed" # https://github.com/pola-rs/polars/pull/10568
features = ["parquet", "lazy", "performant"]
```

I assume polars is registered on Cargo so the github repository is not necessary no? That way you can also specify a released version for polars instead of a commit

Pangoraw:
>It's just a bit odd for the wrapper that will depend on the JLL to contain the source for the JLL that will be built (sort of a circular dependency).
Could c-polars have maybe its own repository?

It is similar to the Makie monorepo where GLMakie and Makie live in the repository for example.

> I assume polars is registered on Cargo so the github repository is not necessary no? That way you can also specify a released version for polars instead of a commit

This is unfortunately currently needed to build without a nightly compiler. But once https://github.com/pola-rs/polars/commit/2dbf5cde8dc0cebd8c3092824ff85d37aaba524c is part of a tagged polars release, we should be able to pin a real release.

giordano:
This is failing to build for x86_64 macOS with
```
[18:03:26]   = note: Undefined symbols for architecture x86_64:
[18:03:26]             "_kCFURLVolumeAvailableCapacityForImportantUsageKey", referenced from:
[18:03:26]                 __ZN71_$LT$sysinfo..apple..disk..Disk$u20$as$u20$sysinfo..traits..DiskExt$GT$7refresh17h154ef94082600884E in libsysinfo-a93b532d96641e7b.rlib(sysinfo-a93b532d96641e7b.sysinfo.46e16331-cgu.5.rcgu.o)
[18:03:26]                 __ZN7sysinfo5apple4disk9get_disks17ha272e46a4b5401e5E in libsysinfo-a93b532d96641e7b.rlib(sysinfo-a93b532d96641e7b.sysinfo.46e16331-cgu.5.rcgu.o)
[18:03:26]           ld: symbol(s) not found for architecture x86_64
[18:03:26]           clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
```
[_kCFURLVolumeAvailableCapacityForImportantUsageKey](https://developer.apple.com/documentation/corefoundation/kcfurlvolumeavailablecapacityforimportantusagekey) requires the SDK 10.13, you have to use that one.

Pangoraw:
> [kCFURLVolumeAvailableCapacityForImportantUsageKey](https://developer.apple.com/documentation/corefoundation/kcfurlvolumeavailablecapacityforimportantusagekey) requires the SDK 10.13, you have to use that one.

Does that mean I have to include it in the build script like libdeflate does?

https://github.com/JuliaPackaging/Yggdrasil/blob/af723c87726d61dfffd23ba5ce56cf9cf5a326ee/L/libdeflate/build_tarballs.jl#L22-L30

Thank you for the help.

giordano:
Yes, exactly, and you need to add the sdk as source as well

Pangoraw:
I tried to build locally with the SDK but I encountered this error:

```
[23:31:57] rm: can't remove '/opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/System/Library/Printers/Libraries': I/O error
```

PR #7231, opened 2023-08-21
Add mrfDepth, fix #7224

### LINKED ISSUE #7224: Migrating Fortran part of an R package (mrfDepth) into Julia
Hey guys,

I hope this issue helps with the migration of some legacy code to Julia and also paves the way for the addition of similar packages in the future.

The issue is about migrating the Fortran part of an R package, namely mrfDepth, which is hosted as an R package in [mrfDepth](https://github.com/cran/mrfDepth).

The main interest is the `rdepthmedian` function which is used for estimating the Deepest Regression parameters, a well known robust regression estimator. It is simply used like 

```R
R> rdepthmedian(maxit = 10000, x = stackloss)
$deepest
   intercept slope var. 1 slope var. 2 slope var. 3 
-35.37610619   0.82522124   0.44247788  -0.07964602 
```

where the first parameter is intercept and the remaining part represents the partial slops. 

When I compile all of the fortran code in a single binary using 

```shell
$ gfortran -shared -fPIC *.f
```

I can bring all of the functionality into Julia using the FFI features. The estimated parameters can be re-calculated in Julia using 

```Julia
X = convert(Matrix, stackloss) 
n, p = size(X)
n = Int32(n)
p = Int32(p)
A = zeros(Float64, p)
maxit = Int32(10000)
iter = Int32(1)
MDEPAPPR = Int32(3)
result = ccall((:sweepmedres_, "./a.out"), 
    Cint, 
    (Ref{Float64},      # X
        Ref{Int32},           # n
        Ref{Int32},           # np
        Ref{Float64},   # A
        Ref{Cint},           # maxit 
        Ref{Cint},           # iter
        Ref{Cint}            # MDEPAPPR
        ), X, n, p, A, maxit, iter, MDEPAPPR)

println(A)
```

and the codes returns the same estimates as we obtained using R:

```julia
[0.8252212389746946, 0.44247787604338223, -0.0796460177080886, -35.37610619462] 
```

where the last term is intercept and first 3 estimates are partial slopes. 

The problem is, the package is an R package and I failed to make it done using BinaryBuilder.jl. 

- Can any talented friends in this area help migrating this functionality into Julia?
- Can we standardize the toolchain that inputs an R package and prepare a jdll output in an easy way?

Sorry if I am bothering you for a specific issue. 

Thank you for considering my message.

mkitti:
Add the Fortran code from R package mrfDepth, cc: @jbytecode

Fix #7224

### DIFF
--- a/M/mrfDepth/build_tarballs.jl
+++ b/M/mrfDepth/build_tarballs.jl
@@ -1,0 +1,38 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "mrfDepth"
+version = v"1.0.14"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/cran/mrfDepth.git", "6954d87f8ea3c87a2d945676233bf849af5aacd1")
+]
+
+# Bash recipe for building across all platforms
+# License is GPL ( >= 2 ) according to DESCRIPTION
+script = raw"""
+cd ${WORKSPACE}/srcdir/mrfDepth/src
+gfortran -shared -fPIC *.f -o libmrfDepth.${dlext}
+cp libmrfDepth.so ${libdir}
+install_license ${WORKSPACE}/srcdir/mrfDepth/DESCRIPTION
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+platforms = expand_gfortran_versions(platforms)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libmrfDepth", :libmrfDepth)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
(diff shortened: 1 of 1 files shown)


M/mrfDepth/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+    GitSource("https://github.com/cran/mrfDepth.git", "6954d87f8ea3c87a2d945676233bf849af5aacd1")
+]
+
+# Bash recipe for building across all platforms
+# License is GPL ( >= 2 ) according to DESCRIPTION
+script = raw"""
+cd ${WORKSPACE}/srcdir/mrfDepth/src
+gfortran -shared -fPIC *.f -o libmrfDepth.${dlext}
+cp libmrfDepth.${dlext} ${libdir}
+install_license ${WORKSPACE}/srcdir/mrfDepth/DESCRIPTION
```
aviks, line 19: License parsing tools will choke on this file. Might be better to do something like `install_license /usr/share/licenses/GPL-2.0+` ? With a comment linking to the description file?

M/mrfDepth/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/cran/mrfDepth.git", "6954d87f8ea3c87a2d945676233bf849af5aacd1")
+]
+
+# Bash recipe for building across all platforms
+# License is GPL ( >= 2 ) according to DESCRIPTION
+script = raw"""
+cd ${WORKSPACE}/srcdir/mrfDepth/src
+gfortran -shared -fPIC *.f -o libmrfDepth.${dlext}
```
imciner2, line 17:
```suggestion
$FC -shared -fPIC *.f -o libmrfDepth.${dlext}
```

Don't hardcode the compiler.

imciner2 (changes requested):

mkitti:
Preview at https://github.com/mkitti/mrfDepth_jll.jl

mkitti:
I believe all issues have been addressed here.

PR #7253, opened 2023-08-25
[GMP] Add version 6.3.0

fingolfin:
A few weeks ago, GMP 6.3.0 was released. Besides integrating some of the patches we already used, it has some other nice improvements.

However, since this is used by Julia itself, handling it is a bit delicate... or at least that's how I remember it... We also need to coordinate this with the Julia side. So for now this is just a draft.

### DIFF
--- a/G/GMP/GMP@6.3.0/bundled/patches/gmp-exception.patch
+++ b/G/GMP/GMP@6.3.0/bundled/patches/gmp-exception.patch
@@ -1,0 +1,14 @@
+diff --git a/errno.c b/errno.c
+index b4be555..3f772a5 100644
+--- a/errno.c
++++ b/errno.c
+@@ -68,5 +68,8 @@ __gmp_sqrt_of_negative (void)
+ void
+ __gmp_divide_by_zero (void)
+ {
++  /* try to force a division by zero system exception */
++  __gmp_junk = 10 / __gmp_0;
++
+   __gmp_exception (GMP_ERROR_DIVISION_BY_ZERO);
+ }
+
--- a/G/GMP/GMP@6.3.0/bundled/patches/gmp_alloc_overflow_func.patch
+++ b/G/GMP/GMP@6.3.0/bundled/patches/gmp_alloc_overflow_func.patch
@@ -1,0 +1,111 @@
+diff --git a/errno.c b/errno.c
+index d71c146..40d83bd 100644
+--- a/errno.c
++++ b/errno.c
+@@ -38,8 +38,11 @@ see https://www.gnu.org/licenses/.  */
+ 
+ #include "gmp-impl.h"
+ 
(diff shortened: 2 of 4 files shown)


G/GMP/GMP@6.3.0/bundled/patches/gmp-exception.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/errno.c b/errno.c
+index b4be555..3f772a5 100644
+--- a/errno.c
++++ b/errno.c
+@@ -68,5 +68,8 @@ __gmp_sqrt_of_negative (void)
+ void
+ __gmp_divide_by_zero (void)
+ {
++  /* try to force a division by zero system exception */
++  __gmp_junk = 10 / __gmp_0;
```
fingolfin, line 10: Note that by default, GMP now uses `__gmp_exception` which in turn does this:
```
/* Use SIGFPE on systems which have it. Otherwise, deliberate divide
   by zero, which triggers an exception on most systems. On those
   where it doesn't, for example power and powerpc, use abort instead. */
void
__gmp_exception (int error_bit)
{
  gmp_errno |= error_bit;
#ifdef SIGFPE
  raise (SIGFPE);
#else
  __gmp_junk = 10 / __gmp_0;
#endif
  abort ();
}
```
I wonder: instead of patching GMP to force division by zero here, could/should we catch SIGFPE instead?

Note that this also affects the overflow func patch, see my other comment

G/GMP/GMP@6.3.0/bundled/patches/gmp_alloc_overflow_func.patch
```diff
@@ -0,0 +1,111 @@
...
+ void
+ __gmp_overflow_in_mpz (void)
++{
++  (*__gmp_alloc_overflow_func)();
++}
++
++static void
++__gmp_default_alloc_overflow(void)
+ {
+   __gmp_exception (GMP_ERROR_MPZ_OVERFLOW);
```
fingolfin, line 28: So the overflow patch could be omitted if we instead caught SIGFPE and handled `gmp_errno` suitably.

Alternatively (and perhaps better), we could just use a single patch which adds an optional callback to `__gmp_exception`, and then it is up to us how want to handle the exception. Perhaps we could even convince the GMP folks to accept such a patch properly?

fingolfin:
Note to self: maybe also patch libtool, see https://gmplib.org/list-archives/gmp-bugs/2023-July/005296.html

fingolfin:
Another note to self: also update MPFR to 4.2.1, see https://gmplib.org/list-archives/gmp-discuss/2023-August/006913.html

ViralBShah:
Done in https://github.com/JuliaPackaging/Yggdrasil/pull/7872

PR #7268, opened 2023-08-27
[CImGuiPack] Fix version

Gnimuc:
The new src introduced breaking changes.

### DIFF
--- a/C/CImGuiPack/build_tarballs.jl
+++ b/C/CImGuiPack/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder
 
 name = "CImGuiPack"
-version = v"0.1.3"
+version = v"0.2.0"
 
 # Collection of sources required to build CImGuiPack
 sources = [


giordano:
Are you going to yank the previous version? Otherwise I don't see how this duplicate helps

Gnimuc:
https://github.com/JuliaImGui/LibCImGui.jl/blob/release-1.82.x/Project.toml 

CImGuiPack_jll is only used by LibCImGui so my plan is to drop 0.13 by forcing the version of CImGuiPack_jll to =0.1.2

There is no plan to maintain 0.1.x in the future.

giordano:
I still think you may want to yank the broken version to ensure it's never ever used.

Gnimuc:
@giordano Could you elaborate more about how to "yank" the broken version? Do you mean to open a PR to the registry?

giordano:
> Do you mean to open a PR to the registry?

Yes, see for example https://github.com/JuliaRegistries/General/pulls?q=is%3Apr+sort%3Aupdated-desc+label%3Ayank+is%3Aclosed

Gnimuc:
Thanks!

PR #7269, opened 2023-08-27
[libcxxwrap-julia] Version v0.10.0

barche:
This introduces a breaking change because of a change in type mapping storage, dependent JLLs will need to be recompiled when moving to the to-be-released CxxWrap 0.14.

@fingolfin Given Friday's fiasco I'm going to hold off merging this until you had a chance to take a look ;)


fingolfin:
I can't really say much, other than: if this is 0.10, at least it shouldn't break older CxxWrap releases, we "just" have to recompile everything in order to switch to the newer CxxWrap. And of course while doing so, we might discover issues... (hopefully not).

Perhaps @benlorenz has better insight?

barche:
Correct, ideally the JLL packages should have a compat specification for the libcxxwrap JLL directly when rebuilding against 0.10. I'm going to merge this later today, then.

PR #7270, opened 2023-08-29
Build XyceWrapper for Julia v1.11

staticfloat:


X/XyceWrapper/build_tarballs.jl
```diff
@@ -2,10 +2,10 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
-julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10"]
+julia_versions = [v"1.7", v"1.8", v"1.9", v"1.10", v"1.11"]
```
fingolfin, line 5: This also drop 1.6. Intentionally?

giordano: But in principle nothing should have changed since last build, right?

staticfloat: As usual, I am investigator, victim and perpetrator: https://github.com/JuliaPackaging/Yggdrasil/pull/6180/files

The issue isn't Julia v1.6, it's `aarch64` linux.

staticfloat: It would be nice to have a way to tell BB "filter my own platforms to match this dependency"

giordano:
And compat entries would probably be good anyway :slightly_smiling_face:

staticfloat:
This PR has _not_ been as smooth as I thought it would be.

PR #7277, opened 2023-09-01
[Xyce] Upgrade to v7.7

staticfloat:
Also, link against LBT and force Julia v1.11+ compat, as we're linking against libsuitesparse v7.5

### DIFF
--- a/X/Xyce/build_tarballs.jl
+++ b/X/Xyce/build_tarballs.jl
@@ -62,11 +62,11 @@ products = [
 dependencies = [
     BuildDependency("flex_jll"),
     Dependency(PackageSpec(name="Trilinos_jll", uuid="b6fd3212-6f87-5999-b9ea-021e9cd21b17")),
-    Dependency(PackageSpec(name="SuiteSparse_jll", uuid="bea87d4a-7f5b-5778-9afe-8cc45184846c"); compat="7.2.0"),
+    Dependency(PackageSpec(name="SuiteSparse_jll", uuid="bea87d4a-7f5b-5778-9afe-8cc45184846c"); compat="7.5.1"),
     Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"); compat="5.8.0"),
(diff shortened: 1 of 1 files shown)


X/Xyce/build_tarballs.jl
```diff
@@ -52,12 +60,13 @@ products = [
...
-                    Dependency(PackageSpec(name="Trilinos_jll", uuid="b6fd3212-6f87-5999-b9ea-021e9cd21b17"))
-                    Dependency(PackageSpec(name="SuiteSparse_jll", uuid="bea87d4a-7f5b-5778-9afe-8cc45184846c"))
-                    Dependency(PackageSpec(name="OpenBLAS32_jll", uuid="656ef2d0-ae68-5445-9ca0-591084a874a2"))
-                    Dependency(PackageSpec(name="FFTW_jll", uuid="f5851436-0d7a-5f13-b9de-f02708fd171a"))
-                    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
-                ]
+    BuildDependency("flex_jll"),
+    Dependency(PackageSpec(name="Trilinos_jll", uuid="b6fd3212-6f87-5999-b9ea-021e9cd21b17")),
+    Dependency(PackageSpec(name="SuiteSparse_jll", uuid="bea87d4a-7f5b-5778-9afe-8cc45184846c"); compat="7.2.0"),
+    Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93")),
```
giordano, line 66: Also for lbt it'd be good to specify the compat as it has had some incompatible versions in the past for Windows

staticfloat:
I can't reproduce these failures locally :(

ViralBShah:
The failure is weird in that it is looking for `amd_valid` in `libamd` which is part of SuiteSparse. The library and the symbol are very much present, but SuiteSparse did go through a big version update in 1.10 to SuiteSparse 7 - but there are no real API breakages.

ViralBShah:
Shouldn't Trilinos be pinned to 12.12.1, which is what Xyce recommends is the only supported version?

imciner2:
Closing as stale.

PR #7278, opened 2023-09-01
Add builder for OR-Tools

dourouc05:
~~Not tested nor finished! I am using this PR for test purposes.~~

Failures:
* macOS:
  * ARM64:
    * Some `-march` parameter in OR-Tools' CMake files that Yggdrasil doesn't like. **Won't fix**: not a bug in OR-Tools; [Abseil cannot be built without this parameter](https://github.com/search?q=repo%3Aabseil%2Fabseil-cpp%20-march=armv8-a+crypto&type=code).
    * ~~Undefined symbols for Abseil Time. Will look into this.~~
  * x86_64:
    * ~~Undefined symbols for Abseil Time. Will look into this.~~
    * C++17 is required by default (Abseil won't configure it), i.e. Clang 16+.
* Linux:
  * ARM64:
    * ~~protoc failures. Seems to be related to a 32 vs 64 bit issue (https://stackoverflow.com/questions/57446579/executable-says-line-1-elf-not-found-when-starts).~~
    * Some `-march` parameter in OR-Tools' CMake files that Yggdrasil doesn't like. **Won't fix**: not a bug in OR-Tools; [Abseil cannot be built without this parameter](https://github.com/search?q=repo%3Aabseil%2Fabseil-cpp%20-march=armv8-a+crypto&type=code).
  * x86_64:
    * **Success!**
* FreeBSD:
  * x86_64:
    * ~~Failure of cast around pthread. Bug in Abseil, likely.~~
    * C++17 is required by default (Abseil won't configure it), i.e. Clang 16+.
* Windows:
  * x86_64:
    * ~~Failure finding pthread for OR-Tools while it's found for dependencies.~~
    * ~~zlib using `-rdynamic`, it does not make sense (apart from ELF).~~
    > Cross-compiling to Windows is not supported in OR-Tools (for now, at least). MinGW without cross-compilation is not supported either. 
    > Need an implementation of `dlfcn.h` like https://github.com/dlfcn-win32/dlfcn-win32.

> There is no support for the C++03 string ABI, at least explicitly.

Major TODOs:
* ~~Dependency on SCIP: I'd really like to avoid building SCIP in OR-Tools, because that's of zero use for Julia users, but the support for IIS in MathOpt depends on some features of the SCIP wrapper. Needs a fix in OR-Tools.~~ Solved by https://github.com/google/or-tools/commit/2f04afb7c0f01db8e78fca8411b10a8c01826653.
* ~~Paths for cross-compilation: OR-Tools assumes a different pattern for output directories.~~ Solved by https://github.com/google/or-tools/commit/4fb8de32cbee2773405882956b1dc5237103ef47


O/ORTools/build_tarballs.jl
```diff
@@ -0,0 +1,54 @@
...
+sources = [
+    GitSource("https://github.com/google/or-tools.git",
+              "35d56a4b07db8ea135b83762289fbc0e7d229221"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/or-tools*
+mkdir build
+cmake -S. -Bbuild \
```
giordano, line 18: https://docs.binarybuilder.org/stable/build_tips/#CMake-builds You have to pass `-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}` otherwise you're building for the wrong platform.

dourouc05: Thanks! I didn't find that section on that page (as I'm not on Linux — or without Docker/VM access when on Linux —, I couldn't use the wizard).

dourouc05: Actually, using this option makes things worse: it messes up with paths in such a way that `${CMAKE_CURRENT_LIST_DIR}` isn't pointing to the expected thing, so that the mechanism inside OR-Tools to apply patches fails (https://github.com/google/or-tools/blob/1d696f9108a0ebfd99feb73b9211e2f5a6b0812b/cmake/host.CMakeLists.txt#L106).

giordano: Sounds like an upstream bug then, the toolchain file ***has*** to be used.

imciner2: > Actually, using this option makes things worse: it messes up with paths in such a way that ${CMAKE_CURRENT_LIST_DIR} isn't pointing to the expected thing

It's pointing to the correct place - binary builder cross compiles for all architectures, so the or-tools build chain gets split between a host section (executables used during the build process) and the or-tools section. The error is happening in the host section. In cross compilation, it tries to compile the host tools in the `build/host_tools` directory using a CMakeLists.txt file rooted there. That means all the patches for the host code are only 2 levels down from the root repository instead of 3 levels, so that is a bug with or-tools.

My recommendation is that or-tools not try fussing with relative paths at all, and instead pass the correct paths into their cmake build system for the host tools that they configure. e.g., I could see the following working best:
* Modify the CMake command that generates the host tools to pass an additional variable `-DORTOOLS_GIT_ROOT=${CMAKE_CURRENT_SOURCE_DIR}` here https://github.com/google/or-tools/blob/1d696f9108a0ebfd99feb73b9211e2f5a6b0812b/cmake/host.cmake#L53.
* They then modify the three patch lines (https://github.com/google/or-tools/blob/1d696f9108a0ebfd99feb73b9211e2f5a6b0812b/cmake/host.CMakeLists.txt#L106, https://github.com/google/or-tools/blob/1d696f9108a0ebfd99feb73b9211e2f5a6b0812b/cmake/host.CMakeLists.txt#L120 and https://github.com/google/or-tools/blob/1d696f9108a0ebfd99feb73b9211e2f5a6b0812b/cmake/host.CMakeLists.txt#L136) to instead be `${ORTOOLS_GIT_ROOT}/patches/<patch_name>` instead.

dourouc05: Thanks for stopping by! I had hoped that our CMake infrastructure was ready for cross-compilation, but it clearly is not. Based on the errors I got at other places, it's not just a question of relative paths (looks like `protoc` is compiled for ARM64 while run on x86_64 for Linux ARM64, for instance). Fixing that small issue will only work in one case, but a truckload of others seem to wait just after…

Given how understaffed we are regarding open-source support in general and build systems in particular, the only feasible solution I see in the short term is to forego cross-compilation and focus on Linux x64 (using our existing prebuilt binaries for the other platforms). It would be best if we could support cross-compilation, but this is the first time it's being required for something. (And we don't have a huge buy-in regarding Julia support, making it harder to justify such a time investment.)

O/ORTools/build_tarballs.jl
```diff
@@ -0,0 +1,120 @@
...
+]
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libortools", :libortools),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [Dependency("ProtoBuf")]
```
giordano, line 117: This looks wrong, no?

O/ORTools/build_tarballs.jl
```diff
@@ -0,0 +1,120 @@
...
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"12", preferred_llvm_version=v"16", julia_compat="1.6")
```
giordano, line 120: https://github.com/JuliaPackaging/Yggdrasil/blob/ecbf83d1706228126efa0cd8faeb8eb02bc8f6ed/CONTRIBUTING.md#compatibility-tips

giordano: There's also no llvm 16 in BinaryBuilder at the moment, and llvm defaults to latest available version anyway.

O/ORTools/build_tarballs.jl
```diff
@@ -0,0 +1,120 @@
...
+    -DBUILD_EXAMPLES:BOOL=OFF \
+    -DBUILD_SAMPLES:BOOL=OFF \
+    -DUSE_SCIP:BOOL=OFF \
+    -DUSE_HIGHS:BOOL=OFF \
+    -DUSE_COINOR:BOOL=OFF \
+    -DUSE_GLPK:BOOL=OFF
+cmake --build build
+cmake --build build --target install
+
+# Automatically generate the Julia bindings.
```
giordano, line 70: I'm confused why you can't do this in post processing outside of this recipe.

dourouc05: To ensure that the generated code has exactly the same version as the binary. This is just for generated code (using ProtoBuf.jl), nothing written by hand. 

Where would you do this step?

O/ORTools/build_tarballs.jl
dourouc05: I've enabled the usual way to do cross-compilation, with a fix for the underlying issue I had here. Now, it fails at other places.

O/ORTools/build_tarballs.jl
dourouc05: Fixed.

O/ORTools/build_tarballs.jl
dourouc05: I've bumped the version of GCC to the lowest one where compilation doesn't miserably fail (thanks Abseil…), with an accompanying Julia update.

dourouc05:
@giordano @imciner2 The PR is now ready for review! I only managed to have it compile under Linux; there are more patches required for other OSes, I'll get to that in later versions of the builder.

O/ORTools/build_tarballs.jl
```diff
@@ -0,0 +1,181 @@
...
+    FileProduct("include/operations_research/course_scheduling.proto", :proto_scheduling_course_scheduling),
+    FileProduct("include/operations_research/scheduling/rcpsp/rcpsp.proto", :proto_scheduling_rcpsp),
+    FileProduct("include/operations_research/scheduling/jssp/jobshop_scheduling.proto", :proto_scheduling_jobshop_scheduling),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"11", preferred_llvm_version=v"16", julia_compat="1.9")
```
giordano, line 181: At the moment there's no LLVM 16 in BinaryBuilder, and contrary to GCC for LLVM we use the _latest_ version available anyway
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"11", julia_compat="1.9")
```

O/ORTools/build_tarballs.jl
```diff
@@ -0,0 +1,181 @@
...
+cp ortools/packing/vector_bin_packing.proto ${prefix}/include/operations_research/packing/vbp/
+cp ortools/pdlp/solve_log.proto ${prefix}/include/operations_research/pdlp/
+cp ortools/pdlp/solvers.proto ${prefix}/include/operations_research/pdlp/
+cp ortools/sat/cp_model.proto ${prefix}/include/operations_research/sat/
+cp ortools/sat/sat_parameters.proto ${prefix}/include/operations_research/sat/
+cp ortools/sat/boolean_problem.proto ${prefix}/include/operations_research/sat/
+cp ortools/util/optional_boolean.proto ${prefix}/include/operations_research/
+cp ortools/scheduling/course_scheduling.proto ${prefix}/include/operations_research/
+cp ortools/scheduling/rcpsp.proto ${prefix}/include/operations_research/scheduling/rcpsp/
+cp ortools/scheduling/jobshop_scheduling.proto ${prefix}/include/operations_research/scheduling/jssp/
```
giordano, line 104:
```suggestion
install -Dvm 644 ortools/bop/bop_parameters.proto ${includedir}/operations_research/bop/bop_parameters.proto
install -Dvm 644 ortools/constraint_solver/assignment.proto ${includedir}/operations_research/assignment.proto
install -Dvm 644 ortools/constraint_solver/demon_profiler.proto ${includedir}/operations_research/demon_profiler.proto
install -Dvm 644 ortools/constraint_solver/search_limit.proto ${includedir}/operations_research/search_limit.proto
install -Dvm 644 ortools/constraint_solver/search_stats.proto ${includedir}/operations_research/search_stats.proto
install -Dvm 644 ortools/constraint_solver/solver_parameters.proto ${includedir}/operations_research/solver_parameters.proto
install -Dvm 644 ortools/constraint_solver/routing_enums.proto ${includedir}/operations_research/routing_enums.proto
install -Dvm 644 ortools/constraint_solver/routing_parameters.proto ${includedir}/operations_research/routing_parameters.proto
install -Dvm 644 ortools/glop/parameters.proto ${includedir}/operations_research/glop/parameters.proto
install -Dvm 644 ortools/graph/flow_problem.proto ${includedir}/operations_research/flow_problem.proto
install -Dvm 644 ortools/gscip/gscip.proto ${includedir}/operations_research/gscip.proto
install -Dvm 644 ortools/linear_solver/linear_solver.proto ${includedir}/operations_research/linear_solver.proto
install -Dvm 644 ortools/math_opt/callback.proto ${includedir}/operations_research/math_opt/callback.proto
install -Dvm 644 ortools/math_opt/infeasible_subsystem.proto ${includedir}/operations_research/math_opt/infeasible_subsystem.proto
install -Dvm 644 ortools/math_opt/model.proto ${includedir}/operations_research/math_opt/model.proto
install -Dvm 644 ortools/math_opt/model_parameters.proto ${includedir}/operations_research/math_opt/model_parameters.proto
install -Dvm 644 ortools/math_opt/model_update.proto ${includedir}/operations_research/math_opt/model_update.proto
install -Dvm 644 ortools/math_opt/result.proto ${includedir}/operations_research/math_opt/result.proto
install -Dvm 644 ortools/math_opt/solution.proto ${includedir}/operations_research/math_opt/solution.proto
install -Dvm 644 ortools/math_opt/parameters.proto ${includedir}/operations_research/math_opt/parameters.proto
install -Dvm 644 ortools/math_opt/solvers/glpk.proto ${includedir}/operations_research/math_opt/solvers/glpk.proto
install -Dvm 644 ortools/math_opt/solvers/gurobi.proto ${includedir}/operations_research/math_opt/solvers/gurobi.proto
install -Dvm 644 ortools/math_opt/solvers/highs.proto ${includedir}/operations_research/math_opt/solvers/highs.proto
install -Dvm 644 ortools/math_opt/sparse_containers.proto ${includedir}/operations_research/math_opt/sparse_containers.proto
install -Dvm 644 ortools/packing/multiple_dimensions_bin_packing.proto ${includedir}/operations_research/packing/multiple_dimensions_bin_packing.proto
install -Dvm 644 ortools/packing/vector_bin_packing.proto ${includedir}/operations_research/packing/vbp/vector_bin_packing.proto
install -Dvm 644 ortools/pdlp/solve_log.proto ${includedir}/operations_research/pdlp/solve_log.proto
install -Dvm 644 ortools/pdlp/solvers.proto ${includedir}/operations_research/pdlp/solvers.proto
install -Dvm 644 ortools/sat/cp_model.proto ${includedir}/operations_research/sat/cp_model.proto
install -Dvm 644 ortools/sat/sat_parameters.proto ${includedir}/operations_research/sat/sat_parameters.proto
install -Dvm 644 ortools/sat/boolean_problem.proto ${includedir}/operations_research/sat/boolean_problem.proto
install -Dvm 644 ortools/util/optional_boolean.proto ${includedir}/operations_research/optional_boolean.proto
install -Dvm 644 ortools/scheduling/course_scheduling.proto ${includedir}/operations_research/course_scheduling.proto
install -Dvm 644 ortools/scheduling/rcpsp.proto ${includedir}/operations_research/scheduling/rcpsp/rcpsp.proto
install -Dvm 644 ortools/scheduling/jobshop_scheduling.proto ${includedir}/operations_research/scheduling/jssp/jobshop_scheduling.proto
```

O/ORTools/build_tarballs.jl
```diff
@@ -0,0 +1,181 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+# Prepare the source directory.
+cd $WORKSPACE/srcdir/or-tools*
+atomic_patch -p1 "${WORKSPACE}/srcdir/patches/cmake_dependencies_CMakeLists.txt.patch"
+mkdir build
+cmake --version
+
+# Make the host compile tools easily accessible when cross-compiling.
+# Otherwise, CMake will use the cross-compiler for host tools.
```
giordano, line 24: I'm very confused by this statement: CMake has no distinction between host and target tools (which is a rather annoying limitation of this build system), we want to compile for the target, not the host, and `CMAKE_TARGET_TOOLCHAIN` should already be setting the right toolchain.

dourouc05: And that's exactly the problem: without this, CMake only knows about the target tools and cannot use the host toolchain (which is required for tools that are supposed to run on the host).

giordano: The usual way to deal with this bootstrap problem is to a first build with the host toolchain (BTW, there's `CMAKE_HOST_TOOLCHAIN` without setting everything manually) to build the needed host tools, and then another one with the target toolchain for building the rest.

dourouc05: There's no real bootstrap issue here, as I'm using the existing tool chains (it's just that the build needs a few uncommon tools).

I tried to use the undocumented `CMAKE_HOST_TOOLCHAIN`: unlike the global environment variables, Abseil doesn't detect that this version of GCC defaults to C++17. (At least, the current script builds OR-Tools.)

O/ORTools/build_tarballs.jl
```diff
@@ -0,0 +1,169 @@
...
+    FileProduct("include/operations_research/course_scheduling.proto", :proto_scheduling_course_scheduling),
+    FileProduct("include/operations_research/scheduling/rcpsp/rcpsp.proto", :proto_scheduling_rcpsp),
+    FileProduct("include/operations_research/scheduling/jssp/jobshop_scheduling.proto", :proto_scheduling_jobshop_scheduling),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"11", julia_compat="1.9")
```
giordano, line 169: Would GCC 10 work? https://github.com/JuliaPackaging/Yggdrasil/blob/ac99062a33464ad1396134ba415a97e7fd435a24/CONTRIBUTING.md#compatibility-tips

dourouc05: Unfortunately, no, because one dependency (Abseil) doesn't know how to configure the version of C++ (and it requires C++17 at least, available by default since GCC 11).

O/ORTools/build_tarballs.jl
```diff
@@ -0,0 +1,169 @@
...
+
+# Build OR-Tools.
+cmake -S. -Bbuild \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DBUILD_DEPS:BOOL=ON \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBUILD_EXAMPLES:BOOL=OFF \
+    -DBUILD_SAMPLES:BOOL=OFF \
+    -DUSE_SCIP:BOOL=OFF \
```
Mizux, line 50: disabling SCIP support...

O/ORTools/bundled/patches/cmake_dependencies_CMakeLists.txt.patch
Mizux, line 1: but patching the scip build :thinking:

PR #7284, opened 2023-09-04
[Glib] Update to 2.76.5

jwahlstrand:
Bump version and remove patch applied upstream.


imciner2:
All the platforms (surprisingly) built.

PR #7291, opened 2023-09-06
[DataEcon] upgrade to v0.2.4

bbejanov:


D/DataEcon/build_tarballs.jl
```diff
@@ -8,7 +8,7 @@ version = v"0.2.3"
 # Collection of sources required to complete build
 sources = [
     GitSource("https://github.com/bankofcanada/DataEcon.git",
-        "e940b33a33f1ff173495af5e19a022733fc7b159")
+        "8c37f65f803c7fc5d3c5df961c4bdefdfcd7ff3a")
```
giordano, line 11: Don't you need to update the version number?

bbejanov:
I was hoping for a new build of the same version. Would that work?

________________________________

PR #7294, opened 2023-09-06
Update build_tarballs.jl

joa-quim:
Update GMT
Update DCW to 2.1.2


imciner2:
Are you sure you want the main package version to stay 6.4.2 even though you are bumping the commit to a newer one on the main GMT source?

joa-quim:
Thanks for reminding me that, but it actually doesn't matter that much because that version number is not really true because I building from master and will likely submit another PR in next couple of weeks, but for know I would like to have a new GMT_jll with the current bug fixes and new features that I need for the tutorials that I am writing.

PR #7304, opened 2023-09-07
[Bonmin] Update to 1.8.9

claud10cv:
Changes made to build tarballs for Bonmin 1.8.9

### DIFF
--- a/C/Coin-OR/Bonmin/build_tarballs.jl
+++ b/C/Coin-OR/Bonmin/build_tarballs.jl
@@ -1,3 +1,4 @@
+#Changes made to build tarballs for v1. 8.9
 include("../coin-or-common.jl")
 
 sources = [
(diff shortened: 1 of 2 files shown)


imciner2:
You need to update the build_tarballs.jl file in the bonmin directory in order to do a new build with the new version. I would suggest just adding a version of the comment in the Coin-OR symphony build script here https://github.com/JuliaPackaging/Yggdrasil/blob/master/C/Coin-OR/SYMPHONY/build_tarballs.jl#L1 to the bonmin build script to trigger it.

claud10cv:
Will do, thank you for Your feedback

claud10cv:
Now ONE build out of 16 is failing, apparently due to some linking issue associated with openblas32. 

There are several errors as follows:
```julia
[22:18:41] /workspace/destdir/lib/libopenblas.so: undefined reference to `omp_get_place_num_procs@OMP_4.5'[22:18:41] /workspace/destdir/lib/libopenblas.so: undefined reference to `omp_get_num_places@OMP_4.5'
```

giordano:
Use gcc 6 by adding the keyword argument `preferred_gcc_version=v"6"` to `build_tarballs(...)`

claud10cv:
Will try that thanks

PR #7323, opened 2023-09-12
[XGBoost] Upgrade to v2.0

tylerjthomas9:
[Release notes](https://github.com/dmlc/xgboost/releases/tag/v2.0.0)

Here are the changes that I have made:
- Bump the XGBoost version to 2.0 and updated the build to the corresponding commit 
- Bump `preferred_gcc_version` from `v8` to `v9` for `std::filesystem` support (from my research, this was the best fix that I found, and the builds worked with this change)
- Add MacOS upgrade to support new C++ features such as `any_cast` in macOS 10.14+ and `std::filesystem` in macOS 10.15+
- Patch to fix MacOS compiling with `std::make_shared`

### DIFF
--- a/X/XGBoost/build_tarballs.jl
+++ b/X/XGBoost/build_tarballs.jl
@@ -22,13 +22,15 @@ script = raw"""
 cd ${WORKSPACE}/srcdir/xgboost
 git submodule update --init
 
-# Patch dmlc-core to use case-sensitive windows.h includes: https://github.com/dmlc/dmlc-core/pull/673
+# Patch dmlc-core to use case-sensitive windows.h includes: 
+# https://github.com/dmlc/dmlc-core/pull/673
 (cd dmlc-core; atomic_patch -p1 "../../patches/dmlc_windows.patch")
 
 mkdir build && cd build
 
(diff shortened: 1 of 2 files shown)


tylerjthomas9:
I had to increase the CUDA 11 build from 11.0 to 11.4 due to the `CUDA_Runtime_jll` rebuild [here](https://github.com/JuliaPackaging/Yggdrasil/pull/7213)

X/XGBoost/build_tarballs.jl
```diff
@@ -3,27 +3,51 @@ using BinaryBuilderBase
...
 cd ${WORKSPACE}/srcdir/xgboost
 git submodule update --init
 
-# Patch dmlc-core to use case-sensitive windows.h includes: https://github.com/dmlc/dmlc-core/pull/673
+# Patch dmlc-core to use case-sensitive windows.h includes: 
+# https://github.com/dmlc/dmlc-core/pull/673
 (cd dmlc-core; atomic_patch -p1 "../../patches/dmlc_windows.patch")
 
+# XGBoost fails to compile on mac due to `std::make_shared`
+# being bugged with some apple clang versions
```
giordano, line 30: For what is worth, we don't use Apple Clang, but vanilla Clang.

PR #7325, opened 2023-09-12
Add more AWS libraries

giordano:


giordano:
Sounds like s2n-tls really wants libcrypto from aws-lc from all platforms :smiling_face_with_tear:

PR #7332, opened 2023-09-14
[PRIMA] add recipe

amontoison:
@zaikunzhang

### DIFF
--- a/P/PRIMA/build_tarballs.jl
+++ b/P/PRIMA/build_tarballs.jl
@@ -7,7 +7,7 @@ version = v"0.7.0"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/libprima/prima.git", "7565f6f42358062e56734a91fc62e0aa66ab7575")
+    GitSource("https://github.com/libprima/prima.git", "e601a731133cbc6317b06b568c04e307daf0691b")
 ]
 
 # Bash recipe for building across all platforms
@@ -31,6 +31,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/P/PRIMA/build_tarballs.jl
+++ b/P/PRIMA/build_tarballs.jl
@@ -19,7 +19,7 @@ cmake --build build --target install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = supported_platforms(exclude=Sys.iswindows)
 platforms = expand_gfortran_versions(platforms)
 platforms = filter(p -> libgfortran_version(p) != v"3", platforms)
 
--- a/P/PRIMA/build_tarballs.jl
+++ b/P/PRIMA/build_tarballs.jl
@@ -21,7 +21,7 @@ cmake --build build --target install
 # platforms are passed in on the command line
 platforms = supported_platforms()
 platforms = expand_gfortran_versions(platforms)
-
+platforms = filter(p -> libgfortran_version(p) != v"3", platforms)
 
 # The products that we will ensure are always built
 products = [
--- a/P/PRIMA/build_tarballs.jl
+++ b/P/PRIMA/build_tarballs.jl
@@ -1,0 +1,37 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "PRIMA"
+version = v"0.7.0"
+
+# Collection of sources required to complete build
+sources = [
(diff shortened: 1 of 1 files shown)


P/PRIMA/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libprimac", :libprimac),
+    LibraryProduct("libprimaf", :libprimaf)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
```
ararslan, line 34: As suggested by the build log:
```
│
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/wIfP6/src/auditor/compiler_abi.jl:4
```

```suggestion
dependencies = Dependency[
    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
]
```

ararslan:
The build failure on Windows seems to be because it's building static libraries (`.a`) but not dynamic (`.dll`). Is there an additional build flag that needs to be passed?

The ARM failures on musl Linux are strange: it's complaining that it can't find an `ieee_arithmetic` intrinsic module. Apparently that was new in GFortran 5, and yet the `libgfortran4` builds succeeded for all other platforms but `libgfortran5` still fails on ARM musl Linux. Even more confusingly, i686 and x86_64 musl Linux builds succeeded.

giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/490f462fd968fa4ce97e9e9f288888c5dfa7f613/O/OpenFAST/build_tarballs.jl#L45-L46

zaikunzhang:
> The ARM failures on musl Linux are strange: it's complaining that it can't find an `ieee_arithmetic` intrinsic module. Apparently that was new in GFortran 5, and yet the `libgfortran4` builds succeeded for all other platforms but `libgfortran5` still fails on ARM musl Linux. Even more confusingly, i686 and x86_64 musl Linux builds succeeded.

No worry about this one. `ieee_arithmetic` has been removed. It should not occur if the compilation is done again. Thanks.

amontoison:
@zaikunzhang 
Should we remove the Windows platforms for now?
https://github.com/libprima/prima/issues/58

zaikunzhang:
> @zaikunzhang Should we remove the Windows platforms for now? [libprima/prima#58](https://github.com/libprima/prima/issues/58)

Yes, if this is the only way at the moment.

zaikunzhang:
Hi @amontoison , Windows is supported now. Could you update the recipe accordingly? Thanks.

amontoison:
Done in #7343

zaikunzhang:
Merci!

emmt:
In the `Project.toml` file of `PRIMA_jll`, version is `v0.7.0+0` or  `v0.7.0+1` but this cannot be specified in the `compat` section of my own `Project.toml` file:

```toml
[compat]
julia = "1.6"
PRIMA_jll = "0.7.0+1"
```

yields the following error `invalid version specifier: "0.7.0+1"` in `Pkg.instantiate()`. 

Specifying `0.7.0` for the `compat` section works but since the API of the C functions have chaged between 0.7.0+0 and 0.7.0+1, I really want to distinguish the two versions.

Maybe I am not correctly specifying the version. Otherwise, I think that at least the minor version number should be incremented when the API changes. 

Since the C API of the PRIMA library has changed again since then, it may be a good idea to rebuild the artifact (with the new library and suitable version number).

amontoison:
Hi @emmt!
You just need to provide `PRIMA_jll = "0.7"`.

amontoison:
It's not relevant to distinguish the two versions because the version 0.7.0+0 is replaced by 0.7.0+1. The version 0.7.0+1 will be automatically download.
The number `+i` is a just a way to recompile a package without bumping the version number of the artifact.

amontoison:
> Since the C API of the PRIMA library has changed again since then, it may be a good idea to rebuild the artifact (with the new library and suitable version number).

The C API of the PRIMA library didn't changed since the version 0.7.
Did I miss something?

emmt:
@amontoison thank you for the help. You may be intersted in [PRIMA.jl](https://github.com/emmt/PRIMA.jl) a first attempt I have made to have a Julia interface to the algorithms of `PRIMA` through the artifact `PRIMA_jll`. I did search for such an interface but found none.

emmt:
> The C API of the PRIMA library didn't changed since the version 0.7.
> Did I miss something?

The C API did change: parameters `npt` and `ftarget` were added to `bobyqa`, `newuoa`, and `lincoa` and `ftarget` was added to `cobyla`.

zaikunzhang:
> The C API did change: parameters npt and ftarget were added to bobyqa, newuoa, and lincoa and ftarget was added to cobyla.

Yes. We decided to change a bit the API. This should not happen so often after some iterations. They will be announced if they do happen.

zaikunzhang:
> Yes. We decided to change a bit the API. This should not happen so often after some iterations. They will be announced if they do happen.

Indeed, the Fortran API has been fixed. It was the C API that changed. The C API has not existed long yet.

PR #7345, opened 2023-09-15
[HIP] Add 5.4.4

jpsamaroo:


H/HIP/HIP@5.4.4/build_tarballs.jl
```diff
@@ -0,0 +1,8 @@
+using Pkg
+using BinaryBuilder
+
+include("../common.jl")
+
+build_tarballs(
+    ARGS, configure_build(v"5.2.3")...;
```
amontoison, line 7: @jpsamaroo Is it `v"5.4.4"`?

PR #7347, opened 2023-09-16
[Discussion] Add Mousetrap (DO NOT MERGE)

Clemapfel:
Hi, I am fully aware that this PR is incredibly bad style and should not be merged, but I was send here by @fingolfin (cf. https://github.com/Clemapfel/Mousetrap.jl/issues/18) and they said to open a PR to get help. So far, I deployed my jlls to a private repo because I am incapable of solving the issues on my own, I have genuinely tried for months now and had to give up and do it this way to be able to publish the project that depends on these jlls, [`Mousetrap.jl`](https://github.com/clemapfel/mousetrap.jl), at all. 

There are three different `build_tarballs.jl`s for each OS, even though each of them export the same library output. The underlying C++ project is sound, it works on any machine I have ever tested it on, but I have been unable to make it work inside the binary builder sandbox, unless I split the OSs up like this.

## 1. Each system has Different Dependencies
See this table for which system needs which jll dependency

| jll | Windows | Linux | Apple| 
|-----|---------|-------|------|
| GTK4 | yes | yes | yes |
| libadwaita | yes | yes | yes |
| OpenGLMathematics | yes | yes | yes |
| libcxxwrap_julia | yes | yes | yes |
| libjulia | yes | yes | yes | 
| GLEW | yes | yes | no |
| GLU | no | yes | no |

I assume this can be fixed by just adding the entire set of jlls regardless of OS, since afaik there is no OS-dependend branching outside of the bash-script part of `build_tarballs.jl`.

## 2. Windows needs an external DLL

My project uses OpenGL, to link against it, on Windows, I need access to `OPENGL32.dll`, which is not available in the binary builder sandbox, even though it is available on every Windows computer, it comes pre-installed with the system. There is no jll that supplies it, and I cannot compile it myself because windows is closed source. Because of this, I was forced to download the .dll as a file and use it during building, cf. https://github.com/Clemapfel/Yggdrasil/blob/master/M/mousetrap_windows/build_tarballs.jl#L12.

Similarly, macOS is supposed to have a similar shared library that acts as a wrapper between OpenGL and Apples Metal API. Again, this library comes pre-installed on every physical macOS machine, but is not available in the binary builder sandbox. Unlike the windows `OPENGL32.dll`, it is also not available online. The fact I cannot get access to this shared library during building is the sole reason why any OpenGL-based part of my Julia project cannot be run on macOS, and why I [added a toggle `MOUSETRAP_ENABLE_OPENGL_COMPONENT`](https://github.com/Clemapfel/Yggdrasil/blob/master/M/mousetrap_apple/build_tarballs.jl#L22) to disable OpenGL-related functionality at compile time.

## 3. Folder Structure of the Sandbox Varies depending on OS

This is probably honestly because I'm not very good at CMake, but with the default setup the [linux `CMakeLists.txt`](https://github.com/Clemapfel/mousetrap/blob/main/CMakeLists.txt) was unable to properly locate include directories relating to GTK4, which is why [I hardcoded them](https://github.com/Clemapfel/Yggdrasil/blob/master/M/mousetrap_linux/build_tarballs.jl#L20C129-L21C18). I think this is due to a `Find*.cmake` file missing from `GTK4_jll`, on a proper computer these are present and my `CmakeLists.txt` locates everything automatically. 

Similarly, Windows was unable to locate `glew32.dll`, which is why [I hardcoded that too](https://github.com/Clemapfel/Yggdrasil/blob/master/M/mousetrap_windows/build_tarballs.jl#L25). 

This would be fine, except that the folder structure changes, the path of a file on a linux-only binary builder sandbox is different from that of a linux+windows sandbox, or linux+windows+apple sandbox.

Running the exact same `CMakeLists.txt` on the following physical machines I have available irl:
+ Ubuntu laptop
+ Fedora Laptop
+ Windows 10 Laptop
+ Windows 11 PC
+ Debian VServer
+ OSX 12 Laptop

my `CMakeLists.txt` is able to locate all dependencies automatically without any additional arguments, on all machines. It only fails to do so in the binary builder sandbox, even though I [linked the CMake toolchain file](https://github.com/Clemapfel/Yggdrasil/blob/master/M/mousetrap_windows/build_tarballs.jl#L25) as is recommended, and even manually setting `CMAKE_INCLUDE_PATH` or `CMAKE_LIBRARY_PATH` still causes it to fail. Because I cannot reproduce this issue anywhere else, I don't know how to address it.

Furthermore, `libcxxwrap_julia_jll`, which is the shared library behind `CxxWwrap` and is not maintained by me, fails to locate the Julia include directory in only the "windows-only" sandbox, which caused me to have to [hardcode that path too](https://github.com/Clemapfel/Yggdrasil/blob/master/M/mousetrap_windows/build_tarballs.jl#L34). It works in the "only linux" and "only apple" versions.

---

Since any solution to this would likely require my `CMakeLists.txt` to be modified, I will explain the structure of my project in this section:

`mousetrap` is a shared library compiled from [this C++ project](https://github.com/clemapfel/mousetrap). It has the following dependencies:
| dependency | supplied by jll |
| -----------|-----------------|
| `gtk4, pango, cairo, harfbuzz, gdk_pixbuf, gobject, graphene, gio, gobject, glib`   | `GTK4_jll` |
| `libadwaita-1` | `libadwaita_jll` |
| `glm` | `OpenGLMathematics_jll` |
| `glew` | Windows: `GLEW_jll`, Linux: `GLEW_jll`, Apple: Not available |
| `OpenGL` | Windows: `OPENGL32.dll`, Linux: `GLU_jll`, Apple: Not Available | 

The CMake package of OpenGL is called `GL` or `OpenGL` on linux, but `OPENGL32` on windows. Similarly, glew is called `glew` on linux, but `glew32` on windows. 

Next is `mousetrap_julia_binding`, which is another C++ shared library and completely separate from `mousetrap`, compiled from [this project](https://github.com/clemapfel/mousetrap_julia_binding), which is also managed by me. It has the following dependencies:

| shared library | supplied by jll |
|----------------|------------------|
| `libmousetrap` | see above |
| `libjulia`     | `libjulia_jll` |
| `jlcxx`        | `libcxxwrap_julia_jll` |

Both `libmousetrap.so` an `libmousetrap_julia_binding.so` are runtime dependencies of my Julia project, these two are the only output of my own jll. 

Taking all three of my `build_tarballs.jl` together, the only operating system that is not supported is 32-bit windows. This is due to `OPENGL32.dll` not being available for that platform online.

---

TL;DR To be able to make these 3 `build_tarballs.jl` a single one, I need to be able to do the following:
+ Access `OPENGL32.dll` on Windows
+ Properly locate include directories that should be supplied by `GTK4_jll` and `libadwaita_jll` on linux
+ Properly locate `glew32.dll` on windows
+ Access the MacOS OpenGL wrapper library, which is not present in the BinaryBuilder Mac SDK, or available as part of any dll or online

I asked on both the Julia discord and discourse, but nobody has been able to help me.

Thank you,
C.


fingolfin:
I've converted this into a "draft" since you don't want it to be merged for now.

I looked at the three build recipes, and they are very nearly exactly the same. It should be no problem to merge them. That is:

1. Merge the `dependencies` into a single list. The only downside is that in some cases there will be dependencies that are not necessary on all OSes, but that should be mostly harmless.

2. It is OK to have "source files" that are not used on all OSes. The real problem IMHo is that you are using https://dllfile.net/ to download a random file. How do you know this DLL is genuine and not injected with malware that we then distribute? Have you verified this? Also, are you *sure* it is legal to just distribute that DLL? Are you away of licensing terms anywhere permitting this?

As to using OpenGL on macOS, it is still there in the latest macOS version. You need the compiler&linker flag `-framework OpenGL` to use it, and possibly adjust some `#include` statements but I am not aware of any fundamental issues otherewise. 

3. Regarding your "hardcoded paths" problem, I don't know yet what's going on there, but in any case it doesn't seem like something a simple "if OS is windows then do this  else do that"  couldn't handle for now? There are plenty of other build recipes in here that show how to do that, and of course the BinaryBuilder documentation also discusses it [here](https://docs.binarybuilder.org/stable/build_tips/#Initiating-different-shell-commands-based-on-target)

M/mousetrap_linux/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libmousetrap", :mousetrap),
+    LibraryProduct("libmousetrap_julia_binding", :mousetrap_julia_binding)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="GLEW_jll", uuid="bde7f898-03f7-559e-8810-194d950ce600"))
```
fingolfin, line 54: I think you can have GLEW_jll as a dependency on macOS and Windows, too; if you don't use it, that's not a problem either way.

Well, that is, unless it tries to load something that's not available on macOS and/or Windows (I don't know if it is, I haven't checked). In that case, you can restrict it to loading only on supported systems by using a method that is also used in e.g. https://github.com/JuliaPackaging/Yggdrasil/blob/fbd25b9d6f59e3d2478a1e771b9856af92d16f36/G/GLFW/build_tarballs.jl#L42 : before the start of the `dependencies` definition, add something like
```julia
x11_platforms = filter(p ->Sys.islinux(p) || Sys.isfreebsd(p), platforms)
```
and then here use
```suggestion
    Dependency("GLEW_jll"; platforms=x11_platforms),
```

M/mousetrap_linux/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+    Dependency(PackageSpec(name="GTK4_jll", uuid="6ebb71f1-8434-552f-b6b1-dc18babcca63"))
+    Dependency(PackageSpec(name="OpenGLMathematics_jll", uuid="cc7be9be-d298-5888-8f50-b85d5f9d6d73"))
+    Dependency(PackageSpec(name="GLU_jll", uuid="bd17208b-e95e-5925-bf81-e2f59b3e5c61"))
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"))
+    BuildDependency(PackageSpec(name="libjulia_jll", uuid="5ad3ddd2-0711-543a-b040-befd59781bbf"))
```
fingolfin, line 59: If you link against libjulia, then most likely you need to build multiple versions of the JLL, one for each supported minor Julia version (so e.g. for 1.6, 1.7, 1.8, 1.9). This is because the `libjulia` ABI often changes between between minor versions.

See for example the `julia_versions` list in https://github.com/JuliaPackaging/Yggdrasil/blob/fbd25b9d6f59e3d2478a1e771b9856af92d16f36/L/libcxxwrap_julia/build_tarballs.jl and code using it.

However, I would recommend that you *first* focus on "merging" the three JLL files, and once that works, we can talk about this again.

M/mousetrap_windows/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+using BinaryBuilder, Pkg
+
+name = "mousetrap_windows"
+version = v"0.2.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/Clemapfel/mousetrap.git", "ce80b9ac063170419ffd989921b7d975b74fda72"),
+    GitSource("https://github.com/Clemapfel/mousetrap_julia_binding.git", "4f389b9a4d293c71c36b055dc2316905a33fca7e"),
+    FileSource("https://dllfile.net/download/18762", "b8df8a508817b5f4d8cac6365e3fbf0e8952d50aa58ed9b5032ec5767e530aaa")
```
fingolfin, line 12: As I said, from a purely technical point of view it is perfectly fine to have this `FileSource` present in all builds, and just ignore it on non-Windows build.

That doesn't change the legal and security issues, though (for the latter, at least once you are sure the DLL is legit and malware free, then the checksum should protect against the source server changing the file later on)

Clemapfel:
I fixed this on my end by switching to meson, I'll open a new PR so people don't have to read my novel of problems which are no longer relevant

PR #7364, opened 2023-09-18
[skip build] Fix GR compat for Qt6Base

barche:


barche:
OK, I can make a new one unless there is another way to trigger the CI, changing the title and close/opening didn't help.

giordano:
Push another commit with only `[skip build]` (of force-push the same).  Buildkite ignores completely close/reopen, it only looks at changes in the tip of the branch

PR #7371, opened 2023-09-19
[GR] bump version

jheinen:
bump GR version to 0.72.10

### DIFF
--- a/G/GR/build_tarballs.jl
+++ b/G/GR/build_tarballs.jl
@@ -106,6 +106,6 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-# GCC version 9 because of Qt6
+# GCC version 10 because of Qt6.5
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               preferred_gcc_version = v"9", julia_compat="1.6")
+               preferred_gcc_version = v"10", julia_compat="1.6")
--- a/G/GR/build_tarballs.jl
+++ b/G/GR/build_tarballs.jl
@@ -3,13 +3,13 @@
 using BinaryBuilder
 
 name = "GR"
-version = v"0.72.9"
+version = v"0.72.10"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/sciapp/gr.git", "623b90911a3fc68b253ae26fb0f259ca97d8df7d"),
+    GitSource("https://github.com/sciapp/gr.git", "ffedfe702fdc3872cb3172fddb1dfb8f0d8986c5"),
     FileSource("https://github.com/sciapp/gr/releases/download/v$version/gr-$version.js",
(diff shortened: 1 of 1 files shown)


jheinen:
No idea how to fix this. Or should I stay with the old Qt 6.4.2 version?

G/GR/build_tarballs.jl
```diff
@@ -99,7 +99,7 @@ dependencies = [
     Dependency("Libtiff_jll"; compat="~4.5.1"),
     Dependency("Pixman_jll"),
     HostBuildDependency("Qt6Base_jll"),
-    Dependency("Qt6Base_jll"; compat="~6.4.2"), # Never allow upgrading more than the minor version without recompilation
+    Dependency("Qt6Base_jll"; compat="~6.5.2"), # Never allow upgrading more than the minor version without recompilation
```
giordano, line 102: @barche is it OK? Do you know what was the problem yesterday?

barche: For sure it should be 6.5.2 so GR.jl and QML.jl can be used together, I'll take a look at the errors.

The problem yesterday was using GR_jll binaries compiled against Qt 6.4.2 together with Qt6Base_jll version 6.5.2, which normally should have been prevented by the compat entry, except that the compat entry was not strict enough.

barche:
I set GCC to version 10, this should fix the error.

barche:
Joy oh joy, the new Qt6.5 package is missing the XCB platform and can't work on Linux X11. I missed that because I did all this on my other machine, which is on Wayland.

jheinen:
@barche : Do you think the draft status can be removed now with the successful build of Qt6 (including XCB extensions)?

barche:
Testing now with the new Qt6Base binaries and the compiled artifact from this PR plotting with GR gives the following error. I have no immediate ideas on how to solve this:

```
.julia/dev/GR_jll/override/bin/gksqt: /usr/bin/../lib/julia/libcurl.so.4: version `CURL_OPENSSL_4' not found (required by /usr/lib/libproxy/libpxbackend-1.0.so)
connect: Connection refused
GKS: can't connect to GKS socket application
```

giordano:
You're using a system library, that's the problem.

barche:
Correct, I was accidentally using the Julia from arch linux. Works fine with the official binary for me now.

giordano:
How did that happened though?

barche:
Also the downloaded julia 1.9.3 official binary works fine with the artifacts from this PR, so whatever it is, it is something in my installation getting mixed up. I think it's safe to merge this PR.

jheinen:
LGTM - I will bump `GR.jl` as soon as #7371 is merged.

jheinen:
Thanks!

charperflow:
@barche  Did you ever figure out how your system libraries were getting pulled in? I am currently running into the same issue and only have julia-bin installed from the AUR!

PR #7378, opened 2023-09-19
difmap: upd version + add macos

aplavin:


giordano:
Needless to say the build system is awful and ins't compatible with cross-compilation.

giordano:
That's not very reproducible

giordano:
If we can't cross-compile we can't build from source. For a brief time we did what you suggested with hdf5, it bitrot very quickly and reverted back because it was causing more headaches than anything.

giordano:
I'm worried about reproducibility, but do I understand correctly they removed the source of v2.5k?

giordano:
Untested, but something like
```diff
--- configure
+++ configure
@@ -347,7 +347,7 @@
 ;;
 
   intel-osx-gcc)   # Intel Mac with OSX and the Gnu C compiler.
-    CC=gcc
+    CC=cc
     FC=gfortran
     CFLAGS="$CFLAGS -Dintel_osx"
 #
@@ -368,7 +368,7 @@
 ;;
 
   arm-osx-gcc)   # Arm Mac with OSX and the Gnu C compiler.
-    CC=gcc
+    CC=cc
     FC=gfortran
     CFLAGS="$CFLAGS -Darm_osx"
 #
--- cpg_src/makefile.distrib
+++ cpg_src/makefile.distrib
@@ -29,7 +29,7 @@
 default: pgbind $(LIBCPGPLOT) $(INCDIR)/cpgplot.h
 
 pgbind: pgbind.c
-	$(CC) $(CFLAGS) -I$(INCDIR) -o pgbind pgbind.c
+	$(BUILD_CC) $(CFLAGS) -I$(INCDIR) -o pgbind pgbind.c
 
 $(LIBCPGPLOT): $(CPGPLOT_OBS)
 	ar ru $(LIBCPGPLOT) $(CPGPLOT_OBS)
```
should be just barely saner than what this is now.  But after having looked at the code and I'm more and more convinced the build system is terrible and tries to reinvent the wheel for absolutely no good reason.

giordano:
> Is there a simple way to transfer everything that's required for the jll from that machine to the mac laptop?

PR #7383, opened 2023-09-20
Wizard recipe: charon-v2.2.0

Keno:
This pull request contains a new build recipe I built using the BinaryBuilder.jl wizard:

* Package name: charon
* Version: v2.2.0

@staticfloat please review and merge.

PR #7384, opened 2023-09-20
Trilinos: Build Kokkos and Teko packages

Keno:
These are dependencies of charon, which I'd like to build (https://github.com/JuliaPackaging/Yggdrasil/pull/7383).


Keno:
Alright, so we don't have Kokkos on 32bit, which we can probably make conditional. The build error in atomics is in a file that was removed in https://github.com/kokkos/kokkos/commit/52953c824bac56ba6031c260901247f7259e9ef9, so can probably be addressed by a version bump.

Keno:
Superseded by https://github.com/JuliaPackaging/Yggdrasil/pull/7386.

PR #7386, opened 2023-09-20
Trilinos: Update to 14.4.0

Keno:
This is the latest version. Before I try to mess with the Kokkos build (#7384), better update, since the Kokkos integration has changed a fair bit on Trilinos master.

[DONOTMERGE for now, since we probably want to do any potential Kokkos refactor at the same time]


Keno:
This now includes #7384 also.

PR #7389, opened 2023-09-21
[aws_c_cal] [s2n_tls] Build shared library

giordano:


giordano:
For the record, the linking errors
```
[10:01:26] /opt/x86_64-linux-gnu/bin/../lib/gcc/x86_64-linux-gnu/5.2.0/../../../../x86_64-linux-gnu/bin/ld: /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root/usr/local/lib/libcrypto.a(bcm.c.o): unrecognized relocation (0x2a) in section `.text'
[10:01:26] /opt/x86_64-linux-gnu/bin/../lib/gcc/x86_64-linux-gnu/5.2.0/../../../../x86_64-linux-gnu/bin/ld: final link failed: Bad value
[10:01:26] collect2: error: ld returned 1 exit status
```
were fixed by using the same compiler toolchain as the one used to build the static library (aws_lc was compiled with GCC 7).  We generally don't have these problems because we don't do much static linking, and instead we only link together object files compiled with homogeneous toolchains in one go, and do dynamic linking for the rest.

PR #7405, opened 2023-09-21
[aws_lc] Force compiling with position independent code

giordano:
Let's hope this fixes failures in #7389


giordano:
Nevermind, we were already compiling with `-fPIC`, the trick is linking downstream shared libraries with the same compiler toolchain as the one used to build the static library (gcc 7 in this case)

PR #7412, opened 2023-09-22
[GDAL] fix version_offset

visr:
See https://github.com/JuliaPackaging/Yggdrasil/pull/7050#discussion_r1334462064

Registration of the previous version was blocked: https://github.com/JuliaRegistries/General/pull/91991

### DIFF
--- a/G/GDAL/build_tarballs.jl
+++ b/G/GDAL/build_tarballs.jl
@@ -4,7 +4,7 @@ using BinaryBuilder, Pkg
 
 name = "GDAL"
 upstream_version = v"3.7.2"
-version_offset = v"0.0.0"
+version_offset = v"1.0.0"
 version = VersionNumber(upstream_version.major * 100 + version_offset.major,
                         upstream_version.minor * 100 + version_offset.minor,
                         upstream_version.patch * 100 + version_offset.patch)


visr:
CC @jeremiahpslewis, perhaps related to #7053.

giordano:
Sounds like stricter compat bounds have to be set somewhere?

visr:
Wait no sorry, this should be fine. If I use Julia 1.9, LibPQ_jll and GDAL_jll both work fine, all GDAL tests pass.

I've run into this before, so I think this is a 1.10 issue, but the LibPQ_jll artifact was installed with read-only flag set, leading to this access denied error. This seems to happen sometimes on 1.10. I cannot reproduce it though, if I remove the artifact and install it again, normally it's fine. On Julia 1.9 this never happens.

visr:
I saved both versions of the same artifact. Their contents are exactly the same, but the mode is not:

```
julia> stat("icudt69.dll")
StatStruct for "icudt69.dll"
   size: 28960428 bytes
 device: 2664978069
  inode: 1001914
   mode: 0o100444 (-r--r--r--)
  nlink: 1
    uid: 0
    gid: 0
   rdev: 0
  blksz: 4096
 blocks: 56568
  mtime:  (25 minutes ago)
  ctime:  (25 minutes ago)

julia> stat("icudt69.dll")
StatStruct for "icudt69.dll"
   size: 28960428 bytes
 device: 2652370264
  inode: 1612352
   mode: 0o100666 (-rw-rw-rw-)
  nlink: 1
    uid: 0
    gid: 0
   rdev: 0
  blksz: 4096
 blocks: 56568
  mtime:  (7 minutes ago)
  ctime:  (7 minutes ago)
```

PR #7418, opened 2023-09-23
[LAST] add recipe

M-PERSIC:
- Name: LAST
- Version: 1499 (2cc68d3ba8ae5ca46ceeb69302aef18b9db04f46)
- Source: https://gitlab.com/mcfrith/last.git
- Language(s): C++/Python
- Build System: Makefile
- Depends: Zlib (JLL available: yes)
- Depends (optional): N/A
- Purpose: Bioinformatics software port
- Description: "LAST: find & align related regions of sequences"

Python dependency unused (do not build last-dotplot)
Remove Windows support due to missing sys/mman header
LAST is versioned based on tag numbering

### DIFF
--- a/L/LAST/build_tarballs.jl
+++ b/L/LAST/build_tarballs.jl
@@ -1,0 +1,74 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message
+using BinaryBuilder, Pkg
+
+name = "LAST"
+version = v"1499"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource(
+        "https://gitlab.com/mcfrith/last.git",
+        "2cc68d3ba8ae5ca46ceeb69302aef18b9db04f46",
+    ),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/last
+install_license COPYING.txt
+export CXXFLAGS=" -O3 -Wall -g -std=c++11 -pthread"
+if [[ "${target}" != aarch64-* ]]; then
+    CXXFLAGS="${CXXFLAGS} -msse4"
+fi
+make CXXFLAGS="${CXXFLAGS}" -j${nproc}
+make bindir=${bindir} install
+"""
+
+# These are the platforms we will build for by default, unless further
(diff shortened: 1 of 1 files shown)


L/LAST/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+    ),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/last
+install_license COPYING.txt
+export CXXFLAGS=" -O3 -Wall -g -std=c++11 -pthread"
+if [[ "${target}" != aarch64-* ]]; then
+    CXXFLAGS="${CXXFLAGS} -msse4"
```
imciner2, line 22: Is there a reason you are manually setting the `sse4` architecture flag?

L/LAST/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+fi
+make CXXFLAGS="${CXXFLAGS}" -j${nproc}
+make bindir=${bindir} install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+# NOTE: Windows builds require Cygwin (sys/mman dependency)
+platforms = filter(
+    p -> !Sys.iswindows(p) && !any(arch(p) .== ["powerpc64le", "armv6l", "armv7l"]),
```
giordano, line 32: Isn't
```suggestion
    p -> !Sys.iswindows(p) && !(arch(p) in ("powerpc64le", "armv6l", "armv7l")),
```
simpler (and non-allocating)? Or use `∉` if you fancy Unicode operators.

L/LAST/build_tarballs.jl
M-PERSIC: It's what was written in their Makefile, I just copy/pasted. I definitely need this program to be vectorized + they rely on SSE4 intrinsics directly. Unless GCC can automatically enable `-msse4`?

giordano: I agree in general we don't want to do stuff like this for generic binaries, but in this specific case it may not be _too_ bad: sse4 is fairly old and if it makes a big difference for this package it could be beneficial to enable the extension here.

PR #7421, opened 2023-09-24
[CImGui] Rebuild 1.79 with Apple Silicon, as 1.79.1

IanButterworth:
After this, this should be reverted back to 1.82 without rebuilding 1.82


C/CImGui/build_tarballs.jl
```diff
@@ -44,4 +45,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies, julia_compat="1.7")
```
giordano, line 48: Why 1.7 and not 1.6? Also, I guess we may want to do the same for v1.82?

IanButterworth: I thought the point was that 1.6 couldn't handle experimental platforms? Maybe I misunderstood Max's comment about "after 1.6"

> after 1.6 those checks were improved/relaxed so that we can introduced further platforms w/o breaking things, but this issues was not recognized as such before the attempt to add a new platform (Apple Silicon) and things going boom

PR #7428, opened 2023-09-26
Update patchelf to v0.18.0

sloede:


sloede:
@giordano IMHO this is ready to merge 🙂

PR #7433, opened 2023-09-27
[CombBLAS] add recipe

M-PERSIC:
- Name: CombBLAS
- Version: v2.0.0 (e1c7faad8d5a918d5671b794c04672f30c6bec29)
- Source: https://github.com/PASSIONLab/CombBLAS.git
- Language(s): C++/C
- Build System: CMake
- Depends: OpenMP (jll available: yes - CompilerSupportLibraries), MPI (yes - MPItrampoline)
- Depends (optional): None
- Purpose: Bioinformatics software dependency
- Description: "The Combinatorial BLAS (CombBLAS) is an extensible distributed-memory parallel graph library offering a small but powerful set of linear algebra primitives specifically targeting graph analytics"

gcc recommended for target platforms (see [README](https://github.com/PASSIONLab/CombBLAS/blob/1b1178ef382ed945a3c4e02e3ce98825ab91e4e1/README.md?plain=1#L38))
Need guidance if it is possible to also enable MPI on other platforms like windows

### DIFF
--- a/C/CombBLAS/build_tarballs.jl
+++ b/C/CombBLAS/build_tarballs.jl
@@ -1,0 +1,83 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message
+using BinaryBuilder, Pkg
+using Base.BinaryPlatforms
+const YGGDRASIL_DIR = "../.."
+include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
+
+name = "CombBLAS"
+version = v"2.0.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource(
+        "https://github.com/PASSIONLab/CombBLAS.git",
+        "e1c7faad8d5a918d5671b794c04672f30c6bec29",
+    ),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/CombBLAS/
+install_license LICENSE
+mkdir build
+cd build
+# Prefer gcc over clang due to OpenMP concerns
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake \
+      -DBUILD_SHARED_LIBS=ON \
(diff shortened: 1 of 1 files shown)


C/CombBLAS/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/CombBLAS/
+install_license LICENSE
+mkdir build
+cd build
+# Force gcc on all platforms
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake \
```
giordano, line 24: Why forcing GCC everywhere?

M-PERSIC: Maybe my comment is a bit too brute/lacking some context 😅 There have been [some unaddressed issues with clang](https://github.com/search?q=repo%3APASSIONLab%2FCombBLAS+clang&type=issues) + CombBLAS[ is built with GCC in mind](https://github.com/search?q=repo%3APASSIONLab%2FCombBLAS%20gcc&type=code)

giordano: When you build with GCC also on BSD platforms you need to expand c++ string ABI also there.

M-PERSIC: Do you mean `expand_cxxstring_abis`? If so, that is included. Also, CombBLAS does not build on FreeBSD, so that platform's removed. I just changed up the MPI implementation to match other packages so the build recipe should follow convention.

giordano: Yes, but by default [`expand_cxxstring_abis`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.expand_cxxstring_abis-Tuple{AbstractPlatform}) doesn't expand the ABI for BSD platforms.

M-PERSIC: `expand_cxxstring_abis(skip = Returns(false))`?

ViralBShah:
Should we merge?

M-PERSIC:
If everything checks out, than we are good to merge!

PR #7438, opened 2023-09-27
[libwebp] Bump to v1.2.4 with CVE fix

jeremiahpslewis:
CVE reference: https://nvd.nist.gov/vuln/detail/CVE-2023-5129

Good news is that the binaries which depend on this one are not version pinned...

Wonder if the previous versions should be yanked from the registry??


giordano:
> Wonder if the previous versions should be yanked from the registry??

I don't think we've done that before for security reasons.

PR #7451, opened 2023-09-29
[Tesseract] Fix compat issue

jeremiahpslewis:


T/Tesseract/build_tarballs.jl
```diff
@@ -38,14 +38,14 @@ dependencies = [
...
-    Dependency("Libtiff_jll"; compat="4.3.0"),
+    Dependency("Libtiff_jll"; compat="~4.5.1"),
     Dependency("Zlib_jll"),
     Dependency("Leptonica_jll"; compat="~1.82"),
     Dependency("CompilerSupportLibraries_jll"),
     # Optional dependencies
     # Dependency("ICU_jll"),
     Dependency("Cairo_jll"; compat="1.16.1"),
-    Dependency("Pango_jll"; compat="1.47.0"),
+    Dependency("Pango_jll"; compat="1.50.0"),
```
jeremiahpslewis, line 48:
```suggestion
    Dependency("Pango_jll"; compat="1.47.0"),
```

T/Tesseract/build_tarballs.jl
```diff
@@ -38,7 +38,7 @@ dependencies = [
     Dependency("Giflib_jll"),
     Dependency("JpegTurbo_jll"),
     Dependency("libpng_jll"),
-    Dependency("Libtiff_jll"; compat="4.3.0"),
+    Dependency("Libtiff_jll"; compat="~4.5.1"),
```
giordano, line 41: You can't change the compat bound without changing the version number

giordano (changes requested):

T/Tesseract/build_tarballs.jl
jeremiahpslewis: I'm trying to fix this, which failed...https://github.com/JuliaRegistries/General/pull/92439#issuecomment-1740532930

jeremiahpslewis: (So I don't think the version was registered?)

giordano: That's because compat bounds have always to be manually fixed in the registry first in a consistent way

giordano: This PR is only going to make things worse in the registry

jeremiahpslewis: ok. two questions: 1) should we add a registry test to the ygg CI to prevent this from happening in the future? 2) what do I do now to fix things?

giordano: For consistency whit what it's in the registry: https://github.com/JuliaRegistries/General/blob/1e194764e2c90ca1c8b95463a9aeb528c4cfec2b/jll/T/Tesseract_jll/Compat.toml#L5
```suggestion
    Dependency("Libtiff_jll"; compat="~4.3, ~4.4"),
```

PR #7453, opened 2023-09-29
[HiGHS] up version

matbesancon:
v1.6.0 cc @odow


odow:
To actually trigger a build you need to edit
https://github.com/JuliaPackaging/Yggdrasil/blob/35cb312b4bb5e61b64e48989c51959a12f8021a3/H/HiGHS/HiGHS/build_tarballs.jl#L3
https://github.com/JuliaPackaging/Yggdrasil/blob/35cb312b4bb5e61b64e48989c51959a12f8021a3/H/HiGHS/HiGHSstatic/build_tarballs.jl#L3

PR #7461, opened 2023-10-01
[Mousetrap] Add mousetrap

Clemapfel:
Supercedes #7347. 

I neglected to expand for C++ string ABIs because my project requires gcc 11.0 or newer anyway, so there is no way to build with the old standard. Please correct me if this is improper.


Clemapfel:
I had to add the following line to the `MESON_TARGET_TOOLCHAIN` in the `[binaries]` section:

```ini
make='/usr/bin/cmake'
```

Otherwise it would not detect the cmake executable for me. Should I open an issue for this?

M/mousetrap/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = Platform[]
+include("../../L/libjulia/common.jl")
+for version in [v"1.7.0", v"1.8.2", v"1.9.0", v"1.10", v"1.11"]
+    for platform in libjulia_platforms(version)
+        if !(arch(platform) == "i686" || contains(arch(platform), "arm"))
+            push!(platforms, platform)
+        end
+    end
```
imciner2, line 39: Are you just trying to not use any 32-bit platforms? (I note that this filter will leave only the 64-bit platforms).

imciner2 (changes requested):

M/mousetrap/build_tarballs.jl
Clemapfel: yes, exactly. I couldn't find a is64bit function so I did it with pattern matching

imciner2: You can use `nbits(platform)` to find out the platform, so to only get the 64-bit platforms you can use the comparison `nbits(platform) == 64`.

Clemapfel: that is much nicer, fixed by https://github.com/JuliaPackaging/Yggdrasil/pull/7461/commits/e6ef38bb1477de1a55f5e14193a74ffe091d61f1

staticfloat (approved):

M/mousetrap/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
...
+    Dependency("GTK4_jll")
+    Dependency("libadwaita_jll")
+    Dependency("OpenGLMathematics_jll")
+    Dependency("libcxxwrap_julia_jll")
+    BuildDependency("libjulia_jll")
+    BuildDependency("Xorg_xorgproto_jll"; platforms = x11_platforms)
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.7", preferred_gcc_version = v"12.1.0")
```
giordano, line 59: Please don't use gcc 12, you should instead use the oldest compiler possible: https://github.com/JuliaPackaging/Yggdrasil/blob/7093500a13b031888e5263aa3fcf7c63490bbe2b/CONTRIBUTING.md#compatibility-tips

giordano (commented):
> I neglected to expand for C++ string ABIs because my project requires gcc 11.0 or newer anyway, so there is no way to build with the old standard. Please correct me if this is improper.

Quoting from the [docs](https://docs.binarybuilder.org/stable/building/#Expanding-C-string-ABIs-or-libgfortran-versions)

> This ABI does _not_ have to do with the C++ standard used by the source code, in fact you can build a C++03 library with the C++11 `std::string` ABI and a C++11 library with the C++03 `std::string` ABI. This is achieved by appropriately setting the `_GLIBCXX_USE_CXX11_ABI` macro.

Emphasis not mine.

M/mousetrap/build_tarballs.jl
Clemapfel: I added two commits to expand for string ABIs and to use GCC 9, which is the oldest version supported. Thank you

stemann:
Is support for Julia 1.7 and 1.8 needed? You are spending a lot of CI hours for building this...

Clemapfel:
I agree, but [I was told by fingolfin](https://github.com/JuliaPackaging/Yggdrasil/pull/7347#discussion_r1328071397) to expand for different Julia versions. This and the string ABI expansion makes for a ridiculous amount of versions.

If I only use Julia 1.9, will that mean 1.8 and 1.7 users cannot use the jll package? I'd like to not introduce a version dependency when there isn't any reason for it other than for build time.

stemann:
> I agree, but [I was told by fingolfin](https://github.com/JuliaPackaging/Yggdrasil/pull/7347#discussion_r1328071397) to expand for different Julia versions. This and the string ABI expansion, both of which I didn't have initially but were requested during review, makes for a ridiculous amount of versions.
> 
> If I only use Julia 1.9, will that mean 1.8 and 1.7 users cannot use the jll package? I'd like to not introduce a version dependency when there isn't any reason for it other than for build time.

Indeed.

Yes - that would drop support for 1.8 and 1.7 users.

Clemapfel:
I would be fine to change:
```julia
platforms = Platform[]
for version in [v"1.7.0", v"1.8.2", v"1.9.0", v"1.10", v"1.11"]
    for platform in libjulia_platforms(version)
        if nbits(platform) != 32
            push!(platforms, platform)
        end
    end
end
```
To 
```
platforms = expand_cxxstring_abis(filter!(p -> nbits != 64, supported_platforms())
```
Since I'm not positive any of the Julia api changes between versions affect my project, but I would delegate that decision to you guys since I don't have any experience with deploying jlls.

Clemapfel:
I copied the versions expansion mostly form https://github.com/JuliaPackaging/Yggdrasil/blob/cefcd399d9ef68bced747dd2d3c9ec59f32d169f/L/libcxxwrap_julia/build_tarballs.jl which also has that many versions and is already merged.

ufechner7:
@Clemapfel Are you still waiting for further input, or do you think it can be merged?

ufechner7:
@imciner2 Can you please approve?

Clemapfel:
@imciner2 is there something else I can do? I'm planning on opening a PR with the Julia registry for a package that depends on this jll, but I cannot do so until this is merged and online.

imciner2 (approved):

imciner2:
This is good now. It should appear in the registry in a few hours.

Clemapfel:
thank you!

PR #7465, opened 2023-10-02
New Recipe: HDF4 v2.16.0

meggart:
HDF4 is still used by quite a few legacy datasets in the Earth observation community and it would be good to be able to add HDF4 support to some downstream libraries like NetCDF_jll or GDAL_jll. This is my first attempt at making a build script using the wizard form BinaryBuilder, not all platforms succeeded for reasons that were not completely obvious to me, but it looks like we cover the most important ones.


ViralBShah:
HDF has generally been known as tough to cross compile. @mkitti @simonbyrne

mkitti:
This is HDF 4, the original file format, not HDF5. BB checks seem to be working. Does this work?

meggart:
Yes, building HDF4 is much easier than HDF5. The builds for the platforms included in this PR finish without errors and produce executables as well as a `libhdf.so`. However, I did not yet test using the generated libraries downstream as I did not know how to get the resulting libs out of the build environment.

ViralBShah:
I suppose if BB checks have worked, it may be ok to merge. Let's see what @giordano might have to say.

H/HDF4/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "HDF4"
+version = v"2.16.0"
```
giordano, line 6: Isn't the version number 
```suggestion
version = v"4.2.16"
```
?

H/HDF4/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://support.hdfgroup.org/ftp/HDF/releases/HDF4.2.16-2/src/hdf-4.2.16-2.tar.gz", "a24b18312d421686031c2d66635f7d5abb2fe879f8a182b7e02797b0da8d1f6c")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd hdf-4.2.16-2/
```
giordano, line 16:
```suggestion
cd $WORKSPACE/srcdir/hdf*
```

H/HDF4/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DTEST_LFS_WORKS_RUN=0 -DH4_PRINTF_LL_TEST_RUN=0 -DH4_PRINTF_LL_TEST_RUN__TRYRUN_OUTPUT= ..
```
giordano, line 19:
```suggestion
cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DTEST_LFS_WORKS_RUN=0 \
    -DH4_PRINTF_LL_TEST_RUN=0 \
    -DH4_PRINTF_LL_TEST_RUN__TRYRUN_OUTPUT= \
    ..
```

H/HDF4/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+make
```
giordano, line 20:
```suggestion
make -j${nproc}
```

H/HDF4/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+make install
+exit
```
giordano, line 22:
```suggestion
```

H/HDF4/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("i686", "windows"; ),
+    Platform("x86_64", "windows"; )
+]
+
```
giordano, line 42:
```suggestion
platforms = supported_platforms()
```

H/HDF4/build_tarballs.jl
meggart: I am not sure about this, since we are explicitly talking about HDF4 which I would consider to be a completely different data format compared to HDF5, and not just a different version. Also, practically I don't know if we can assume backwards compatibility if ever a new minor version gets released (which I think will not happen).

giordano: I opened the tarball and saw references to the version number 4.2.16 everywhere.  Also Linux distributions call it 4.2.X, for example Debian https://packages.debian.org/source/sid/libhdf4, and all other distributions: https://repology.org/project/hdf/versions

H/HDF4/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+    -DH4_PRINTF_LL_TEST_RUN=0 \
+    -DH4_PRINTF_LL_TEST_RUN__TRYRUN_OUTPUT= \
+    ..
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
```
giordano, line 31: I suspect these are upstream bugs, but unless someone has the time/will to track them down we can exclude them
```suggestion
platforms = supported_platforms()
# /opt/x86_64-unknown-freebsd13.2/x86_64-unknown-freebsd13.2/sys-root/usr/include/rpc/xdr.h:125:2: error: unknown type name 'u_int'
filter!(!Sys.isfreebsd, platforms)
# /workspace/srcdir/hdf-4.2.16-2/mfhdf/xdr/types.h:68:18: error: conflicting types for ‘u_quad_t’
filter!(p -> !(libc(p) == "musl" && nbits(p) == 64), platforms)
```

meggart: Ok, I agree excluding them for now is good. Thanks for having a look at the error message anyway

PR #7479, opened 2023-10-03
SuiteSparse - Use 7.2.1 release from upstream

ViralBShah:
cc @Wimmerer

PR #7480, opened 2023-10-03
SuiteSparse_GPU - Build 7.2 GPU libs

ViralBShah:
cc @Wimmerer 

This is rebased on top of #7479.


imciner2:
Needs GCC 9+ to get the C++17 support (according to their docs that is when all the C++17 support went ABI stable in GCC), and probably needs a rebase onto current master to get the updated CUDA helpers file so that it gets the CUDA dependencies right for the 12.2 SDK.

PR #7489, opened 2023-10-04
[libcxxwrap_julia] rebuild for Julia 1.10.0-beta3

fingolfin:
CC @benlorenz @barche


barche:
Looks like this requires fixing libcxxwrap for an API change in `jl_apply_tuple_type`, I'll take a look at that later today.

barche:
This will also need a version bump after https://github.com/JuliaInterop/libcxxwrap-julia/pull/136 is merged.

PR #7493, opened 2023-10-05
[sentencepiece] Bump version to v0.1.99

findmyway:
By the way, any hint why it can't be precompiled on Mac?

```
julia> versioninfo()
Julia Version 1.9.3
Commit bed2cd540a1 (2023-08-24 14:43 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 8 × Apple M2
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, apple-m1)
  Threads: 1 on 4 virtual cores
Environment:
  JULIA_PKG_SERVER = https://mirrors.pku.edu.cn/julia

julia> using SentencePiece_jll

[ Info: Precompiling SentencePiece_jll [0cc836cc-86cb-5b0e-b290-caf6c31794d8]

[38556] signal (11.2): Segmentation fault: 11
ERROR: Failed to precompile SentencePiece_jll [0cc836cc-86cb-5b0e-b290-caf6c31794d8] to ...
```


findmyway:
> > By the way, any hint why it can't be precompiled on Mac?
> 
> Stacktrace?

```
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] compilecache(pkg::Base.PkgId, path::String, internal_stderr::IO, internal_stdout::IO, keep_loaded_modules::Bool)
   @ Base ./loading.jl:2300
 [3] compilecache
   @ ./loading.jl:2167 [inlined]
 [4] _require(pkg::Base.PkgId, env::String)
   @ Base ./loading.jl:1805
 [5] _require_prelocked(uuidkey::Base.PkgId, env::String)
   @ Base ./loading.jl:1660
 [6] macro expansion
   @ ./loading.jl:1648 [inlined]
 [7] macro expansion
   @ ./lock.jl:267 [inlined]
 [8] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:1611
```

giordano:
Stacktrace of the segmentation fault.  Who's throwing the segfault?  The library being loaded?  Julia?  Something else?

findmyway:
> Stacktrace of the segmentation fault.

I can only see the error message `[40942] signal (11.2): Segmentation fault: 11` when running `using SentencePiece_jll`. Nothing else.


> Who's throwing the segfault? The library being loaded? Julia? Something else?

TBH, I'm not very sure...

findmyway:
Hi @giordano ,

Is it possible to skip the check of several `LibraryProduct` on some platforms?

For `SentencePiece` here, it won't build shared library on windows platform.

https://github.com/google/sentencepiece/blob/8cbdf13794284c30877936f91c6f31e2c1d5aef7/CMakeLists.txt#L43C1-L45

```
# Disable shared build on windows
if(WIN32)
  set(SPM_ENABLE_SHARED OFF)
endif()
```

giordano:
You'd have to do different `build_tarballs(...)`  calls with different arguments, but I feel like disabling building the shared library on Windows is an upstream mistake...

findmyway:
> You'd have to do different `build_tarballs(...)` calls with different arguments, but I feel like disabling building the shared library on Windows is an upstream mistake...

Strange, it seems binary builder is still looking for the existence of libproducts though I've exclude them in the second `build_tarballs`.

findmyway:
> > You'd have to do different `build_tarballs(...)` calls with different arguments, but I feel like disabling building the shared library on Windows is an upstream mistake...
> 
> Strange, it seems binary builder is still looking for the existence of libproducts though I've exclude them in the second `build_tarballs`.

Hi @giordano ,

It would be much appreciated if you could take a look into this. 🤗

giordano:
Sorry, I was referring to something like https://github.com/JuliaPackaging/Yggdrasil/blob/fbae3dcd1fb377d028445fd36e4157985f475033/Q/qwt/build_tarballs.jl#L48-L53 if you call `build_tarballs` multiple times with same platforms you're just repeating everything multiple times.  But I keep saying that
```cmake
# Disable shared build on windows
if(WIN32)
  set(SPM_ENABLE_SHARED OFF)
endif()
```
might just be wrong and commenting that out could be the easiest solution if that works.

giordano:
As I imagined, building the shared libraries on Windows just works.  We can simply do that, without any further complication.

findmyway:
Great thanks!

PR #7495, opened 2023-10-06
[libsingular_julia] rebuild for Julia 1.10.0-beta3

fingolfin:


barche:
Closing and reopening doesn’t trigger rebuilds with Buildkite, Mosé told me.

L/libsingular_julia/build_tarballs.jl
```diff
@@ -50,7 +50,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    BuildDependency("libjulia_jll"),
+    BuildDependency(PackageSpec(;name="libjulia_jll", version=v"1.10.6")),
     BuildDependency("GMP_jll"),
     BuildDependency("MPFR_jll"),
     Dependency("libcxxwrap_julia_jll"; compat = "~0.11.0"),
```
fingolfin, line 56:
```suggestion
    Dependency("libcxxwrap_julia_jll"; compat = "~0.11.1"),
```

fingolfin: So it keeps building against 0.11.0. Alas, if I change the compat here, then this JLL also need a version increment...

fingolfin: I will be on the road now and can take of that later, if anyone else wants to modify this PR in the meantime, be my guest ;-)

PR #7502, opened 2023-10-06
[SCOTCH] Fix the compilation on FreeBSD

amontoison:


amontoison:
Hummm, It works locally because I have the old version of Clang.

PR #7505, opened 2023-10-07
[rustfft] version 0.3.1

Taaitaaiger:


R/rustfft/build_tarballs.jl
```diff
@@ -8,8 +8,8 @@ uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "rustfft"
-version = v"0.3.0"
-julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9"]
+version = v"0.3.1"
```
giordano, line 11: To be clear: you only want to build for Julia v1.10 without changing the upstream source?

Taaitaaiger:
I released v0.19.1 of jlrs yesterday, rustfft-jl depends on 0.19, so I
don't need to change anything in rustfft-jl to use it.

Op zo 8 okt. 2023 07:51 schreef Mosè Giordano ***@***.***>:

PR #7509, opened 2023-10-09
Recompile SCOTCH for PTSCOTCH

amontoison:
The shared library `libptesmumps` and the associated header should be only available in the tarballs generated by PTSCOTCH (#6563).
We also don't want to store the executable in the tarballs.

### DIFF
--- a/S/SCOTCH/build_tarballs.jl
+++ b/S/SCOTCH/build_tarballs.jl
@@ -64,13 +64,20 @@ CFLAGS=$FLAGS cmake .. \
     -DINSTALL_METIS_HEADERS=OFF
 
 make -j${nproc}
-make install
+# make install
 
 if [[ "${target}" == *mingw* ]]; then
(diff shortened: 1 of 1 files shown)


giordano:
When merged on `master`, builds on `aarch64-apple-darwin` and `i686-linux-gnu` (two seemingly unrelated platforms) [initially failed](https://buildkite.com/julialang/yggdrasil/builds/5882#018b15bb-4b27-49ff-b6e7-5978f1beecbb/628-1920) with
```
[18:40:01] Error copying file "/workspace/srcdir/scotch/src/libscotchmetis/library_metis.h" to "/workspace/srcdir/scotch/build/src/libscotchmetis".
```
It worked when the builds have been restarted.  This smells badly of race conditions in the build systems.

amontoison:
Thanks Mosè,  I will compile PTSCOTCH without `-j${nproc}` to avoid this issue.

PR #7500, opened 2023-10-06
[P4est] Enable OpenMPI

JoshuaLampert:
Currently, OpenMPI is disabled for P4est_jll.

PR #7510, opened 2023-10-09
[P4est] disable OpenMPI again

JoshuaLampert:
Enabling OpenMPI for P4est.jl in #7500 was breaking the JLL, see e.g. https://github.com/trixi-framework/P4est.jl/pull/88. As a hot fix, we first revert the change and disable OpenMPI again.

cc @ranocha, @sloede


sloede (approved):

JoshuaLampert:
This should be ready to merge @giordano.

giordano:
Yanking the broken version in the registry is a better solution than creating a new version identical to an old one

JoshuaLampert:
Thanks, see https://github.com/JuliaRegistries/General/pull/93078.

PR #7512, opened 2023-10-09
[open62541] update to version 1.3.7

thomvet:


O/open62541/build_tarballs.jl
```diff
@@ -29,6 +26,7 @@ cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DUA_ENABLE_AMALGAMATION=ON \
     -DUA_ENABLE_IMMUTABLE_NODES=ON \
     -DBUILD_SHARED_LIBS=ON \
+    -UA_FORCE_WERROR=OFF \
```
giordano, line 29: Did you mean 
```suggestion
    -DUA_FORCE_WERROR=OFF \
```
?  But I'm confused: https://github.com/open62541/open62541/blob/46659f739138d02d48d00979075f95063e8e5806/CMakeLists.txt#L300-L303 the option is set to ON only if doing a Debug build, but this is a Release build: https://github.com/JuliaPackaging/Yggdrasil/blob/36adb7b1ac8c92c6a26e438464fe30de1d47d4e6/O/open62541/build_tarballs.jl#L23 why is `UA_FORCE_WERROR` ON at all?

thomvet: I don't know, but the fact that this kind of error aborts the build means that somehow it's on?

giordano: I don't know, I tend to prefer to understand _why_ things happen when possible, instead of handwaving.

giordano: Ah, that was introduced in https://github.com/open62541/open62541/commit/0b3b53bfe739bda756f533aece967a5524731280, which is after the version you're building.  At least this explains the mismatch.

thomvet: Doing this the first time - sorry for tinkering around here. I would also like to understand this and not just dirty fix it.

O/open62541/build_tarballs.jl
```diff
@@ -3,20 +3,17 @@
...
 
 name = "open62541"
-version = v"1.2.2"
+version = v"1.3.7"
 
 # Collection of sources required to complete build
 sources = [
     GitSource("https://github.com/open62541/open62541.git",
-              "ecf5a703785877a8719a0cda863a98455f7d5d12"),
-    DirectorySource("./bundled"),
```
giordano, line 12: Can you please delete the directory with the now-useless patches?

giordano:
Thanks for your contribution!

thomvet:
Thank YOU for your help with it.

PR #7517, opened 2023-10-09
[libcellml] Update to version 0.5.0

hsorby:


L/libcellml/build_tarballs.jl
```diff
@@ -6,24 +6,11 @@ version = v"0.4.0"
...
-        "697e0b2d07e5e68d147a42c63305d7f72ce26c78"),
-    ArchiveSource(
-        "https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.14.sdk.tar.xz",
-        "0f03869f72df8705b832910517b47dd5b79eb4e160512602f593ed243b28715f"),
+        "9482382acc150cfd3c046292975853cd76e57f43"),
 ]
 
 # https://libcellml.org/documentation/guides/latest/installation/build_from_source
 script = raw"""
-# This requires macOS 10.14
```
giordano, line 17: Sounds like this package still requires macOS 10.14 :slightly_smiling_face: 
```
[23:18:18] Undefined symbols for architecture x86_64:
[23:18:18]   "__ZTISt12bad_any_cast", referenced from:
```

hsorby: :( CMAKE_OSX_DEPLOYMENT_TARGET is set to 10.15, so trying to not require macOS 10.14.

giordano: Ok, but we still need to manually install the newer SDK, by default we ship 10.12 :disappointed:

hsorby: Oh crikey, I thought it was us who was trying to support older versions.  Thanks for the info.

giordano:
Good to go now?

hsorby:
Not just yet, currently this is synced with the 0.5.0 release candidate(s).  When the official 0.5.0 is release I will update this and then we should be good to go.  0.5.0 needs a little housekeeping, nothing that will effect the build.

PR #7520, opened 2023-10-11
build libspatialindex for experimental platforms, bump to 1.9.3

mattwigway:
This updates the LibSpatialIndex_jll package to build for experimental platforms, and additionally updates to version 1.9.3. It should fix JuliaGeo/LibSpatialIndex.jl#15


mattwigway:
Thanks for the suggestions @giordano, will integrate those!

I'm not sure why the Windows builds are failing. On BuildKite it's not showing any errors other than that the libraries are missing, but it also shows just above that that the libraries were found, so I guess something to do with the auditor? When I run locally with `--verbose` I get `Warning: Minimum instruction set detected for libspatialindex.dll is avx, not x86_64 as desired.` which is probably related. I'm not sure how to fix that - shouldn't the CMake toolchain file specify the machine architecture?

imciner2:
> I'm not sure why the Windows builds are failing. On BuildKite it's not showing any errors other than that the libraries are missing

The installation log seems to suggest they are being installed wrong. Here is where they are installed:

```
[17:21:56] -- Installing: /workspace/destdir/lib/libspatialindex.dll.a
[17:21:56] -- Installing: /workspace/destdir/./libspatialindex.dll
[17:21:56] -- Installing: /workspace/destdir/lib/libspatialindex_c.dll.a
[17:21:56] -- Installing: /workspace/destdir/./libspatialindex_c.dll
```
But they should instead be inside `${destdir}/bin`, so something in the toolchain has converted the `bin` to `.`

giordano:
Don't ask me why: https://github.com/libspatialindex/libspatialindex/blob/3be67583852c741d0357f7ef5148c5fe8b484076/CMakeLists.txt#L181

mattwigway:
I think this is good to go on the technical side, and all the LibSpatialIndex.jl tests pass on my M1, but it would be great if one of the LibSpatialIndex.jl maintainers could comment before merge (cc @yeesian @visr @evetion) since there's a version bump as well as the compile change, and it increases the compat entry for Julia to 1.6. Based on the LibSpatialIndex changelog, it looks like the only changes from 1.8.5 are bugfixes and new features, so I don't think anything should stop working.

visr (approved):
Thanks for taking this on!

evetion:
Thanks from a happy M1 user! Libspatialindex already requires Julia 1.6, so this should be no problem.

L/LibSpatialIndex/build_tarballs.jl
```diff
@@ -1,44 +1,41 @@
...
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release ..
+cmake --build .
+cmake --build . --target install
+install_license ../COPYING
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(supported_platforms())
+platforms = expand_cxxstring_abis(supported_platforms(; experimental=true))
```
giordano, line 38:
```suggestion
platforms = expand_cxxstring_abis(supported_platforms())
```

L/LibSpatialIndex/build_tarballs.jl
```diff
@@ -1,44 +1,41 @@
...
-./configure --help
 
-./configure --prefix=${prefix} --host=$target --build=${MACHTYPE} --enable-static=no
-make
-make install
-install_license COPYING
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release ..
+cmake --build .
```
giordano, line 31:
```suggestion
cmake --build . -j${nproc}
```

giordano:
Thanks for your contribution!

PR #7526, opened 2023-10-13
[yosys] update to 34

sjkelly:
This is a simple version bump, but the dependencies list is quite old at this point. Is there a good way to update the dependency `compat`s automatically?

### DIFF
--- a/Y/Yosys/build_tarballs.jl
+++ b/Y/Yosys/build_tarballs.jl
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "Yosys"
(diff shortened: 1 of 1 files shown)


giordano:
> Is there a good way to update the dependency compats automatically?

No, also because I don't that's a well defined process

imciner2:
@sjkelly do you want to update the compats, or should we just merge this in and that be done later?

sjkelly:
Since it builds, I suppose we can merge in the interim. I have been prototyping an update script to automate this.

ViralBShah:
Merge?

PR #7532, opened 2023-10-14
[picotool] Add picotool_jll

Seelengrab:
This is for building the [`picotool`](https://github.com/raspberrypi/picotool) binary, a program used to flash & inspect the Raspberry Pi family of microcontrollers.

The binary is linked statically with libstdc & libstdc++, because _apparently_ their SDK can only be built with GCC13, and pretty much noone is going to already have that around. For my purposes, this will (likely) be good enough.

For future reference, the flags for linking came from [here](https://github.com/arduino/rp2040tools/blob/d6e368bde164286b24a6b647f4a78d49622af457/.github/workflows/release.yml#L77-L91).

### DIFF
--- a/P/picotool/build_tarballs.jl
+++ b/P/picotool/build_tarballs.jl
@@ -14,6 +14,8 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/picotool/
+cp udev/99-picotool.rules $prefix/99-picotool.rules
+install_license LICENSE.TXT
 mkdir build
 cd build
 PICO_SDK_PATH=$WORKSPACE/srcdir/pico-sdk cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_EXE_LINKER_FLAGS="-static-libgcc -static-libstdc++" -DCMAKE_BUILD_TYPE=Release ..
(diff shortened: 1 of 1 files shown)


Seelengrab:
Well, looks like `musl` won't be built then.

P/picotool/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/picotool/
+cp udev/99-picotool.rules $prefix/99-picotool.rules
+install_license LICENSE.TXT
+mkdir build
+cd build
+PICO_SDK_PATH=$WORKSPACE/srcdir/pico-sdk cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_EXE_LINKER_FLAGS="-static-libgcc -static-libstdc++" -DCMAKE_BUILD_TYPE=Release ..
```
giordano, line 21: You need to pass the toolchain file: https://docs.binarybuilder.org/stable/build_tips/#CMake-builds

Seelengrab: Not sure what that should have helped with - `musl` breaks anyway https://buildkite.com/julialang/yggdrasil/builds/5984#018b2f3c-9e98-45ff-a3f7-a7835b0a3b1e/662-842

Seelengrab: I've removed musl again, since I don't really have a use for that. If someone with a musl system wants to use this, we can always debug it later.

PR #7535, opened 2023-10-15
[CompilerSupportLibraries] Upgrade to GCC 13 and move Win `*.a` files to subdir

giordano:
CC: @vtjnash


vtjnash (commented):
A folder name of julia seems odd for being gcc. I think `lib/gcc/<triple>/<major>` would be better (where they normally are located when using a vanilla build of gcc).

giordano:
Ok, I moved it to `lib/gcc/<triple>`.  I excluded `<major>` just because we'd have to update the number in julia's `Makefile` every time we update CSL there, unless that's seen as a feature?

vtjnash:
That is where other tools, such as gcc itself, will expect to find it

giordano:
Alright, moved to the suggested directory, including the major version number.

PR #7541, opened 2023-10-17
[TropicalGemmC_jll] added support of cuda v11

ArrogantGao:


imciner2:
I think you can actually cleanup the CUDA support here if you use the helper platforms provided in this repo now. Take a look at the SuiteSparse build script: https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SuiteSparse/SuiteSparse_GPU%405/build_tarballs.jl, and note its use of `CUDA.supported_platforms()` and `CUDA.required_dependencies()`.

ArrogantGao:
Thank you very much for you advice, I tried the function and found a problem.
The `CUDA.required_dependencies()` only returns the `CUDA_SDK_jll` and my binary also need `CUDA_SDK_static_jll`, if only use the previous one the compiling will fail.
Is there any method to make the `CUDA_SDK_static_jll` also included?

imciner2:
> Is there any method to make the CUDA_SDK_static_jll also included?

There is now :grinning:. Rebase your branch onto the current master branch. Then, if you add the `static_sdk=true` keyword argument to the `CUDA.required_dependencies` call, it should add the static libraries as well.

ArrogantGao:
Thanks, it works now!

imciner2:
Excellent. Merged in, so you should have the new CUDA versions once it does the build and register.

PR #7553, opened 2023-10-17
[platforms] Add CUDA_SDK_static as an optional dependency in the generator

imciner2:
This will allow easy addition of the static libraries when using the platform dependency generator, since the consuming build_tarballs can just do `CUDA.required_dependencies(platform, static_sdk=true)` instead of having to build the `BuildDependency` themselves.

@maleadt, this look good to you?

### DIFF
--- a/platforms/cuda.jl
+++ b/platforms/cuda.jl
@@ -153,20 +153,27 @@ function is_supported(platform)
 end
 
 """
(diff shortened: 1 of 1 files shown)


maleadt (approved):
LGTM, thanks!

PR #7557, opened 2023-10-20
edfplusdcnv [new package]

palday:
Riffing off pandoc as a template for grabbing a project-supplied blob.

https://www.teuniz.net/edfplusd-converter/index.html

### DIFF
--- a/E/edfplusdcnv/build_tarballs.jl
+++ b/E/edfplusdcnv/build_tarballs.jl
@@ -13,9 +13,9 @@ sources = [
 script = raw"""
 cd ${WORKSPACE}/srcdir/
 mkdir -p "${bindir}"
-cp ${target}/edfplusdcnv-* ${bindir}
+cp edfplusdcnv-* ${bindir}
 chmod +x ${bindir}/edfplusdcnv
-install_license  ${bindir}/README
+install_license ${bindir}/README
 """
 
 # These are the platforms we will build for by default, unless further
--- a/E/edfplusdcnv/build_tarballs.jl
+++ b/E/edfplusdcnv/build_tarballs.jl
@@ -13,6 +13,7 @@ sources = [
 script = raw"""
 cd ${WORKSPACE}/srcdir/
 mkdir -p "${bindir}"
+tar -xvzf edfplusdcnv-*.tar.gz
 cp edfplusdcnv-* ${bindir}
 chmod +x ${bindir}/edfplusdcnv
 install_license ${bindir}/README
--- a/E/edfplusdcnv/build_tarballs.jl
+++ b/E/edfplusdcnv/build_tarballs.jl
@@ -13,12 +13,9 @@ sources = [
 script = raw"""
 cd ${WORKSPACE}/srcdir/
 mkdir -p "${bindir}"
-if [[ "${target}" != *-mingw* ]]; then
-    subdir="bin/"
-fi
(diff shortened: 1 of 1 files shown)


E/edfplusdcnv/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+version = VersionNumber(edfplusdcnv_ver)
+
+sources = [
+    ArchiveSource("https://www.teuniz.net/edfplusd-converter/edfplusdcnv_100.tar.gz", "eeb6033a2f21dc052852e3dd779da7d368f65d88e45b7e40fa14f4bb514e62b9"; unpack_target = "x86_64-linux-gnu"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/
+mkdir -p "${bindir}"
```
giordano, line 15:
```suggestion
```

E/edfplusdcnv/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+cp edfplusdcnv_*/* "${bindir}"
```
giordano, line 16: The path of the file you want to install is clearly wrong, but I'm away from computer, I don't know what's inside and can't check myself. In the meantime, suggesting a better way to install file (still with the wrong path, so it won't work anyway):
```suggestion
install -Dvm 755 edfplusdcnv_*/edfplusdcnv "${bindir}/edfplusdcnv"
```

E/edfplusdcnv/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+chmod +x ${bindir}/edfplusdcnv
```
giordano, line 17:
```suggestion
```

E/edfplusdcnv/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
+using BinaryBuilder
+
+# Collection of pre-build edfplusdcnv binaries
+name = "edfplusdcnv"
+edfplusdcnv_ver = "1.0.0"
+version = VersionNumber(edfplusdcnv_ver)
+
+sources = [
+    ArchiveSource("https://www.teuniz.net/edfplusd-converter/edfplusdcnv_100.tar.gz", "eeb6033a2f21dc052852e3dd779da7d368f65d88e45b7e40fa14f4bb514e62b9"; unpack_target = "x86_64-linux-gnu"),
```
giordano, line 9: With `unpack_target = "x86_64-linux-gnu"` you're saying "call the directory `x86_64-linux-gnu`", so you either don't do this or in the build script below you install the file `${target}/edfplusdcnv`:
```suggestion
    ArchiveSource("https://www.teuniz.net/edfplusd-converter/edfplusdcnv_100.tar.gz", "eeb6033a2f21dc052852e3dd779da7d368f65d88e45b7e40fa14f4bb514e62b9"),
```

giordano (changes requested):
As far as I can tell, the binary you're trying to redistribute has no license at all, which means we aren't allowed to redistribute it.

palday:
There is a GPL distribution of the entire source of this program's "ecosystem" (this is just a util function in a larger piece of software) but I can't figure out how to build this binary from that source. I've worked out a stopgap for my needs and probably won't get around to addressing the license/build issue, so I'm closing for now.

Thanks for the guidance and review though! :heart:

PR #7560, opened 2023-10-22
[MUMPS] Upgrade to v5.6.2

amontoison:


amontoison:
With the new release 5.6.2, we have the opportunity to change the name of the shared libraries such that `MUMPS_seq_jll` and `MUMPS_jll` can coexist.

`PETSC_jll` is the only package that is compiled with the parallel version of MUMPS. So weed to update the General registry to add a compat entry.

ViralBShah:
Also assume this is good to merge now?

imciner2:
@amontoison is this good to merge? The CI builds all seem to be passing.

amontoison:
@imciner2 The issue is with #7626, I changed to the name of the shared libraries of `MUMPS_jll.jl` and I don't want to break `PETCs_jll.jl`.

PR #7573, opened 2023-10-26
Update build_tarballs.jl for `QuantumEspresso`

singularitti:


Q/QuantumEspresso/build_tarballs.jl
```diff
@@ -116,4 +108,4 @@ append!(dependencies, platform_dependencies)
 
 # Build the tarballs, and possibly a `build.jl` as well
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               augment_platform_block, julia_compat="1.6", preferred_gcc_version=v"6")
+    augment_platform_block, julia_compat="1.6", preferred_gcc_version=v"12.1.0")
```
imciner2, line 111: Why does this need a bump to GCC 12.1 instead of staying on GCC 6?

singularitti: Please see [this comment](https://github.com/JuliaPackaging/Yggdrasil/blob/6740359c7e6b2ee048001fa4b649c11cd45b9831/Q/QuantumEspresso/build_tarballs.jl#L11). I was trying to see if higher versions of GCC works.

PR #7574, opened 2023-10-26
Update to librdkafka 2.3.0 and openssl 3.0

bluesmoon:
This MR bumps the librdkafka version to 2.3.0 and also bumps the openssl dependency to 3.0.


L/librdkafka/build_tarballs.jl
```diff
@@ -48,7 +48,7 @@ dependencies = [
     Dependency(PackageSpec(name="Lz4_jll", uuid="5ced341a-0733-55b8-9ab6-a4889d929147")),
     Dependency(PackageSpec(name="Zstd_jll", uuid="3161d3a3-bdf6-5164-811a-617609db77b4")),
     Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a")),
-    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="1.1.10"),
+    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="3.0.11"),
```
giordano, line 51: Can relax to
```suggestion
    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="3.0.8"),
```

bluesmoon:
Waiting for https://github.com/JuliaPackaging/Yggdrasil/pull/7575

bluesmoon:
Ok, I guess we have to wait for CyrusSASL_jll 2.1.29 to get into the registry.

L/librdkafka/build_tarballs.jl
```diff
@@ -45,11 +45,11 @@ products = [
...
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="Lz4_jll", uuid="5ced341a-0733-55b8-9ab6-a4889d929147")),
-    Dependency(PackageSpec(name="Zstd_jll", uuid="3161d3a3-bdf6-5164-811a-617609db77b4")),
-    Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a")),
-    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="1.1.10"),
-    Dependency(PackageSpec(name="CyrusSASL_jll", uuid="6422fedd-75a7-50c2-a7c3-a11dad25a896")),
+    Dependency(PackageSpec(name="Lz4_jll", uuid="5ced341a-0733-55b8-9ab6-a4889d929147"); compat="1.9.4"),
+    Dependency(PackageSpec(name="Zstd_jll", uuid="3161d3a3-bdf6-5164-811a-617609db77b4"); compat="1.5.0"),
+    Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a"); compat="1.3.0"),
```
giordano, line 50: Zlib is a Julia dep, if you restrict to superlatest version you're going to make this uninstallable in older versions of Julia. In general, for compat you should really use the oldest compatible version, not the superlatest one.

bluesmoon:
I can't tell why the build is failing on FreeBSD. It's possible we need some CPPFLAGS but I'll have to look into that tomorrow.

giordano:
```
[04:01:44] /workspace/srcdir/librdkafka/src/rdkafka_metadata.c:1809:18: error: call to undeclared library function 'alloca' with type 'void *(unsigned long)'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
[04:01:44]         topics = rd_alloca(sizeof(*topics) * topic_cnt);
[04:01:44]                  ^
[04:01:44] /workspace/srcdir/librdkafka/src/rdposix.h:73:22: note: expanded from macro 'rd_alloca'
[04:01:44] #define rd_alloca(N) alloca(N)
[04:01:44]                      ^
[04:01:44] /workspace/srcdir/librdkafka/src/rdkafka_metadata.c:1809:18: note: include the header <stdlib.h> or explicitly provide a declaration for 'alloca'
```
Problem is https://github.com/confluentinc/librdkafka/issues/1962, which was closed without resolution.  As the error message suggests, they should include `<stdlib.h>` for `alloca` (glibc also [agrees](https://www.gnu.org/software/libc/manual/html_node/Variable-Size-Automatic.html)).

bluesmoon:
Cool, I'll make a patch.

bluesmoon:
@giordano doesn't look like it made a difference.  Still investigating.

bluesmoon:
Looks like the patch shouldn't be required since stdlib.h is already included in rd.h

bluesmoon:
I don't know what else to do about FreeBSD, so I've decided to exclude FreeBSD support. If librdkafka cannot be built on freebsd, then it cannot be used there.

bluesmoon:
I've added an issue on librdkafka: https://github.com/confluentinc/librdkafka/issues/4487

L/librdkafka/build_tarballs.jl
```diff
@@ -35,7 +33,7 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(supported_platforms())
+platforms = expand_cxxstring_abis(supported_platforms(exclude = x -> Sys.isbsd(x) && !Sys.isapple(x)))
```
mkitti, line 36: Do you want to exclude FreeBSD mainly?

```suggestion
platforms = expand_cxxstring_abis(supported_platforms(exclude = Sys.isfreebsd)
```

giordano:
It isn't great to reduce the supported platforms, but if upstream broke some of them and we don't have the bandwidth to fix it, I guess we don't have many options.

bluesmoon:
@giordano agreed. I'll keep track of the upstream issue and if/when it's fixed or alternately someone suggests a way to fix it ourselves, I'll re-add support and push out a new version.

mkitti:
@arsla Any ideas to rescue FreeBSD support here?

PR #7575, opened 2023-10-26
Update CyrusSASL to use OpenSSL 3.0

bluesmoon:


C/CyrusSASL/build_tarballs.jl
```diff
@@ -3,11 +3,12 @@
 using BinaryBuilder, Pkg
 
 name = "CyrusSASL"
-version = v"2.1.28"
+version = v"2.1.28+openssl3"
```
giordano, line 6: You can't do this, you must use a different X.Y.Z version number, build numbers aren't acceptable.

bluesmoon:
@giordano any idea why this fails on mingw32? From the configure output it looks like it can't find openssl & libcrypto:
https://buildkite.com/julialang/yggdrasil/builds/6179#018b6c88-928e-4b8a-a418-c4f61c40d583  (see lines 2271 & 2272 in the output)

PR #7589, opened 2023-10-28
[librdkafka] rebuild with freebsd support

benlorenz:
So the reason for the build failure seems to be that librdkafka [defines](https://github.com/confluentinc/librdkafka/blob/95a542c87c61d2c45b445f91c73dd5442eb04f3c/src/rd.h#L44) `_POSIX_C_SOURCE` everywhere except windows:
```
#ifndef _POSIX_C_SOURCE
#define _POSIX_C_SOURCE 200809L /* for timespec on solaris */
#endif
```

But this define causes FreeBSD to hide various other defines, including `gettimeofday` and `alloca` (via `__XSI_VISIBLE` and `__BSD_VISIBLE`). We probably need to end up in the default block here to have these available: https://github.com/freebsd/freebsd-src/blob/61b15e6dfc963a0c67dbaeae7f4590674976111f/sys/sys/cdefs.h#L765

PS: The logs from the previous jll version (2.0.3) contain:
```
/workspace/srcdir/librdkafka/src/rdtime.h:118:9: warning: implicit declaration of function 'gettimeofday' is invalid in C99 [-Wimplicit-function-declaration]
        rd_gettimeofday(&tv, NULL);
        ^
/workspace/srcdir/librdkafka/src/rdposix.h:155:33: note: expanded from macro 'rd_gettimeofday'
#define rd_gettimeofday(tv, tz) gettimeofday(tv, tz)
                                ^
/workspace/srcdir/librdkafka/src/rdkafka_metadata.c:1353:18: warning: implicitly declaring library function 'alloca' with type 'void *(unsigned long)' [-Wimplicit-function-declaration]
        topics = rd_alloca(sizeof(*topics) * topic_cnt);
                 ^
/workspace/srcdir/librdkafka/src/rdposix.h:73:22: note: expanded from macro 'rd_alloca'
#define rd_alloca(N) alloca(N)
                     ^
/workspace/srcdir/librdkafka/src/rdkafka_metadata.c:1353:18: note: include the header <stdlib.h> or explicitly provide a declaration for 'alloca'
/workspace/srcdir/librdkafka/src/rdposix.h:73:22: note: expanded from macro 'rd_alloca'
#define rd_alloca(N) alloca(N)
                     ^
2 warnings generated.
```
Same issues but non-fatal, so I guess this was also broken for FreeBSD?

x-ref: #7574
cc: @bluesmoon

### DIFF
--- a/L/librdkafka/build_tarballs.jl
+++ b/L/librdkafka/build_tarballs.jl
@@ -9,6 +9,7 @@ version = v"2.3.0"
 sources = [
     # git rev-list -n 1 v2.3.0
     GitSource("https://github.com/confluentinc/librdkafka.git", "95a542c87c61d2c45b445f91c73dd5442eb04f3c",),
+    DirectorySource("./bundled"),
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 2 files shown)


bluesmoon:
Yes it was always broken on FreeBSD, but the change to Clang16 changes them from warnings to errors.

Issue #7605, opened 2023-10-31
`PMIx_jll` 4.2.7 seems to be broken (or something has changed)


Sbozzolo:
On two remote clusters I have access to, `PMIx_jll` 4.2.7 cannot be loaded with the following error:
```
ERROR: InitError: could not load library "/home/gbozzola/.julia/artifacts/b64ad2ff5bdc8b41a59009d2b5547838cfde4f5d/lib/libpmix.so"
libevent_core-2.1.so.7: cannot open shared object file: No such file or directory
Stacktrace:
  [1] dlopen(s::String, flags::UInt32; throw_error::Bool)
    @ Base.Libc.Libdl ./libdl.jl:117
  [2] dlopen(s::String, flags::UInt32)
    @ Base.Libc.Libdl ./libdl.jl:116
  [3] macro expansion
    @ ~/.julia/packages/JLLWrappers/pG9bm/src/products/library_generators.jl:63 [inlined]
  [4] __init__()
    @ PMIx_jll ~/.julia/packages/PMIx_jll/UrLDH/src/wrappers/x86_64-linux-gnu.jl:12
  [5] register_restored_modules(sv::Core.SimpleVector, pkg::Base.PkgId, path::String)
    @ Base ./loading.jl:1115
  [6] _include_from_serialized(pkg::Base.PkgId, path::String, ocachepath::String, depmods::Vector{Any})
    @ Base ./loading.jl:1061
  [7] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String, build_id::UInt128)
    @ Base ./loading.jl:1506
  [8] _require(pkg::Base.PkgId, env::String)
    @ Base ./loading.jl:1783
  [9] _require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:1660
 [10] macro expansion
    @ ./loading.jl:1648 [inlined]
 [11] macro expansion
    @ ./lock.jl:267 [inlined]
 [12] require(into::Module, mod::Symbol)
    @ Base ./loading.jl:1611
during initialization of module PMIx_jll
```
To be noted that
```
$ find .julia -name "*libevent_core*"
.julia/artifacts/e9feeabfa989549ac1dac61e4df5f2849149ad6d/lib/libevent_core-2.1.so.7
.julia/artifacts/e9feeabfa989549ac1dac61e4df5f2849149ad6d/lib/libevent_core-2.1.so.7.0.1
[...]

PR #7609, opened 2023-11-01
[libevent] Add more products

giordano:
These are apparently needed by `PMIx_jll`, see #7605.


giordano:
Wait, I'm not sure this is necessary.

PR #7619, opened 2023-11-02
[Qt6Base] Upgrade to OpenSSL 3

### LINKED ISSUE #7531: Bump Qt6Base to use OpenSSL 3
Are there any plans to do so?

simonbyrne:
I'm not exactly sure if this is the correct fix?

Fixes #7531.

cc: @barche

### DIFF
--- a/Q/Qt6Base/build_tarballs.jl
+++ b/Q/Qt6Base/build_tarballs.jl
@@ -13,7 +13,7 @@ const host_build = true
 # Collection of sources required to build qt6
 sources = [
     ArchiveSource("https://download.qt.io/official_releases/qt/$(version.major).$(version.minor)/$version/submodules/qtbase-everywhere-src-$version.tar.xz",
-                  "7cda4d119aad27a3887329cfc285f2aba5da85601212bcb0aea27bd6b7b544cb"),
+                  "df2f4a230be4ea04f9798f2c19ab1413a3b8ec6a80bef359f50284235307b546"),
     ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/11.0-11.1/MacOSX11.1.sdk.tar.xz",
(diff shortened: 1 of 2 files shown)


simonbyrne:
Okay, well we're a couple of versions behind (6.5.3 and 6.6.0 are both available: https://download.qt.io/archive/qt/), so should we just update those?

giordano:
Yup, that'd do it.

Q/Qt6Base/build_tarballs.jl
```diff
@@ -176,7 +176,7 @@ dependencies = [
     Dependency("Glib_jll", v"2.59.0"; compat="2.59.0"),
     Dependency("Zlib_jll"),
     Dependency("CompilerSupportLibraries_jll"),
-    Dependency("OpenSSL_jll"; compat="1.1.10"),
+    Dependency("OpenSSL_jll"; compat="3"),
```
giordano, line 179:
```suggestion
    Dependency("OpenSSL_jll"; compat="3.0.8"),
```

simonbyrne:
Okay, no idea if this is the right way to do this. @barche does this look right?

simonbyrne:
okay, i'm out of ideas...

barche:
I’m away from my computers until Sunday, but 6.5.3 is the right version to choose, going to 6.6 would require rebuilding GR.

barche:
Looks like the new version requires some extra patching to deal with the AVX stuff.

simonbyrne:
thanks @barche!

giordano (approved):

yuyichao:
I assume it's unintentional that this removes everything but x64 linux musl? https://github.com/JuliaBinaryWrappers/Qt6Base_jll.jl/commit/382b0ad96f67ebe1c55dc92a921fbb48d76a5a3b

yuyichao:
OK I saw https://github.com/JuliaPackaging/Yggdrasil/pull/7628#issuecomment-1793832399

PR #7621, opened 2023-11-02
[OpenVAF] Add v23.5.0

topolarity:


giordano:
You may want to look at the [Enzyme recipe](https://github.com/JuliaPackaging/Yggdrasil/blob/fea91b601cd059a29d0604ae1638eb0806c8afd3/E/Enzyme/build_tarballs.jl) for how to target llvm, which is important because each julia minor version has a different llvm

topolarity:
> You may want to look at the [Enzyme recipe](https://github.com/JuliaPackaging/Yggdrasil/blob/fea91b601cd059a29d0604ae1638eb0806c8afd3/E/Enzyme/build_tarballs.jl) for how to target llvm, which is important because each julia minor version has a different llvm

Hmm, this is a standalone executable that statically links LLVM.

Is there anything I still need to worry about in that case?

giordano:
Ah, excellent point, probably for a standalone executable with static linking we don't need to care about the exact version of LLVM.

O/OpenVAF/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+cargo build --release --bin openvaf
+
+install -Dvm 0755 "target/${rust_target}/release/openvaf${exeext}" "${bindir}/openvaf${exeext}"
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+filter!(p -> !(Sys.islinux(p) && libc(p) == "glibc"), platforms)
```
giordano, line 30: Double checking: you did mean this to exclude glibc, and so only build for x86_64-linux-musl?

topolarity:
That `llvm-config` needs quite the dance, huh. I've had a chance to return to this now, and I think we've got a decent set of platforms up and running.

This is ready for review + merge at your leisure @giordano

O/OpenVAF/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+filter!(p -> !(Sys.iswindows(p) && arch(p) == "i686"), platforms) # undefined reference to `_Unwind_Resume'
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("openvaf", :openvaf),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(get_addable_spec("LLVM_full_jll", v"15.0.7+9")),
```
giordano, line 46: If you're doing static linking do you still need to install `LLVM_full_jll` as dependency at runtime?

giordano: Actually
```console
% otool -L openvaf
openvaf:
	@rpath/libLLVM.dylib (compatibility version 1.0.0, current version 15.0.7)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.3.1)
	@rpath/libiconv.2.dylib (compatibility version 9.0.0, current version 9.1.0)
```
this doesn't seem to be statically linking llvm?

topolarity: Yeah, I had to switch this away from static linking due to the `llvm-config` woes. The host LLVM build has a different set of components and static libraries, but it cannot run the target llvm-config so it's quite hard to get a list of the necessary static libs

My workaround for now was to dynamically link, which only has a single libLLVM to worry about.

Probably a more proper fix for the future would be to recompile LLVM fresh in this recipe, since the additional components are needed for OpenVAF to do support cross-compilation between x86 and AArch64 anyway

giordano: Ok, I _think_ this should work as long as LLVM_full_jll@v15.0.7 can be installed, but you also need to specify the compat bound with something like (untested!)
```suggestion
    Dependency(get_addable_spec("LLVM_full_jll", v"15.0.7+9"); compat="15.0.7"),
```

PR #7623, opened 2023-11-03
CUDA: Restore v10.2

stemann:


platforms/cuda.jl
```diff
@@ -87,6 +87,7 @@ platform(cuda::String) = cuda
 # BinaryBuilder.jl currently does not allow selecting a BuildDependency by compat,
 # so we need the full version for CUDA_SDK_jll (JuliaPackaging/BinaryBuilder.jl#/1212).
 const cuda_full_versions = [
+    v"10.2.89",
```
imciner2, line 90: If we want to have the 10.2 version listed here, then I think we need to modify the `supported_platforms()` function below to allow specifying CUDA major versions to target. Build recipes are starting to transition to use `CUDA.supported_platforms()` to get the platform strings, and I don't think all CUDA packages we have will support 10.x.

stemann: Right, I have added a suggested implementation change in the last push.

stemann: Bump

stemann: Bump :-)

stemann: Bump...

stemann:
@maleadt How does these changes look to you?

platforms/cuda.jl
```diff
@@ -108,11 +114,16 @@ function full_version(ver::VersionNumber)
...
 
 Return a list of supported platforms to build CUDA artifacts for.
+
+# Arguments
+- `cuda_major_versions=cuda_major_versions`: CUDA major versions to target (given provided `cuda_full_versions`) - if `cuda_minor_versions` is empty.
+- `cuda_minor_versions=VersionNumber[]`: CUDA minor versions to target (given provided `cuda_full_versions`) - if non-empty.
+- `cuda_full_versions=cuda_full_versions`.
 """
-function supported_platforms()
+function supported_platforms(; cuda_major_versions=cuda_major_versions, cuda_minor_versions=VersionNumber[], cuda_full_versions=cuda_full_versions)
```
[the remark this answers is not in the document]
stemann: The cuda_full_versions arg was in order to make the function a bit more hackable … e.g. for someone who would like to override the CUDA full versions used.

Will re-consider API style with your comment in mind.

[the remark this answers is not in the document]
stemann: The argument for the array arguments:
1. (the not very strong argument): It mirrors the exclude kwarg of BinaryBuilder.supported_platforms
2. (the slightly better argument): It is more flexible than min/max bounds (... but perhaps unnecessarily so (too flexible)?)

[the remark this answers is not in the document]
maleadt: It's not hard to roll your own version of `CUDA.supported_platforms` if you need the flexibility, so I don't think it's valuable to offer that level of customization.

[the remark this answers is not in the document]
stemann: Right - fixed.

PR #7624, opened 2023-11-03
[SuiteSparse_GPU] Build GPU libraries for SuiteSparse 7.3

ViralBShah:
@maleadt Could you review this PR to make sure the updated GPU build is ok, in case anything needs to be changed to keep pace with the cuda ecosystem.

Also cc @rayegun


maleadt (approved):
Looks great, exactly how I intended the CUDA platform support to be used!

PR #7626, opened 2023-11-05
[PETSc] Update the compat entry for MUMPS_jll

amontoison:
#7560

### DIFF
--- a/P/PETSc/build_tarballs.jl
+++ b/P/PETSc/build_tarballs.jl
@@ -6,7 +6,7 @@ include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
 name = "PETSc"
 version = v"3.18.6"
 petsc_version = v"3.18.6"
-MUMPS_COMPAT_VERSION = "5.5.1"
+MUMPS_COMPAT_VERSION = "5.5.1 - 5.6.1"
 SUITESPARSE_COMPAT_VERSION = "5.10.1"
 SUPERLUDIST_COMPAT_VERSION = "8.1.2"   
 MPItrampoline_compat_version="5.2.1"    
--- a/P/PETSc/build_tarballs.jl
+++ b/P/PETSc/build_tarballs.jl
@@ -29,11 +29,7 @@ script = raw"""
 cd $WORKSPACE/srcdir/petsc*
 atomic_patch -p1 $WORKSPACE/srcdir/patches/petsc_name_mangle.patch
 
-if [[ "${target}" == *-apple* ]]; then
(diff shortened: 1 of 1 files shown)


amontoison:
I just changed the compat entry of `MUMPS_jll` and many builds are failing...
I don't think that it's related to my modification.

ViralBShah:
Seems to be unable to find BLAS.

ViralBShah:
Seems like my commits make it worse.

amontoison:
@giordano I wanted to regenerate `PETSc_jll.jl` to add a compat entry for `MUMPS_jll.jl` directly in the `Project.toml` of `PETSc_jll.jl` but it seems that PETSc downloads additional files during the compilation and we can't easily rebuild it with the same setup.
Is it enough if we just add the compat entry in the `General` repository (done with this [PR](https://github.com/JuliaRegistries/General/pull/93917))?

@ViralBShah The idea of this PR was to rebuild `PETSc_jll` without updating the dependencies.
LBT is great but here we need to rebuild a few packages with it before (`SCALAPACK`, `MUMPS`, ...).
It will break a lot of Julia packages because LP64 BLAS / LAPACK must manually loaded and it's what is required here.

giordano:
My understanding of
```
ERROR: LoadError: Unsatisfiable requirements detected for package LazyArtifacts [4af54fe1]:
 LazyArtifacts [4af54fe1] log:
 ├─possible versions are: 1.3.0 or uninstalled
 └─found to have no compatible versions left with MUMPS_jll [ca64183c]
   └─MUMPS_jll [ca64183c] log:
     ├─possible versions are: 5.2.1-5.6.1 or uninstalled
     └─restricted to versions 5.5.1 by an explicit requirement, leaving only versions 5.5.1
```
is that there's something wrong in the registry which has to be fixed.

ViralBShah:
@amontoison Agree. Aplogoes for polluting the PR. Would it be possible for you to revert or force push to get rid of my commits?

ViralBShah:
cc @KristofferC

giordano:
https://github.com/JuliaRegistries/General/pull/92622#issuecomment-1813255248. Same problem as in #7687

ViralBShah:
aarch64 mac libgfortran5 seems to be complaining about unable to link to `divdc3`.

giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/80da7de30c6632f05b7e15dc9f779ba16111fb32/L/LLVMOpenMP/build_tarballs.jl#L38-L42

amontoison:
#7865 will fix the issue.

PR #2856, opened 2021-04-17
[Qt] Qt6Base builder for all platforms

barche:
Second stage for adding the Qt6Base package, now for all platforms. 🤞

PR #7628, opened 2023-11-05
[Qt6Base] Targets build for v6.3.5

barche:
CC @simonbyrne

### DIFF
--- a/Q/Qt6Base/build_tarballs.jl
+++ b/Q/Qt6Base/build_tarballs.jl
@@ -75,13 +75,8 @@ case "$bb_full_target" in
         make -j${nproc}
         make install
 
-        if [[ "${target}" == x86_64-* ]]; then
-           # On x86_64 mingw32 the import libraries of OpenSSL are in `lib64/`.
-           export LDFLAGS="-L${prefix}/lib64"
-        fi
-
         cd $WORKSPACE/srcdir/build
-        ../qtbase-everywhere-src-*/configure -prefix $prefix $commonoptions -opengl dynamic -- $commoncmakeoptions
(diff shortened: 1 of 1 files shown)


barche:
I just realized that the host-then-target build sequence temporarily breaks GR in minor point upgrades like this, since QtBase 6.5.3 is already in the registry but only for x64 MUSL.

barche:
Windows is failing because CMake can't find OpenSSL, I have no immediate idea on how to fix this, maybe someone else encountered this before using OpenSSL 3? CC @giordano

barche:
Removing `-openssl-linked` on mingw is the only thing I can think of right now. For context, see: https://doc.qt.io/qt-6/ssl.html#enabling-and-disabling-ssl-support-when-building-qt-from-source

I *think* this should still pick up our openssl at runtime, and if not and someone complains (I don't think anyone is actually using the SSL-dependent Qt classes from within Julia) we can try to find a better fix.

barche:
Feel free to merge if this builds and if you approve @giordano

giordano:
Import libraries of openssl on x86_64 win are in `${prefix}/lib64` instead of `${prefix}/lib`, but setting LDFLAGS like in https://github.com/JuliaPackaging/Yggdrasil/blob/b524ac3c03e12e162435e5886e0b619fa32c8737/C/CyrusSASL/build_tarballs.jl#L46-L49 doesn't seem to help here.  I don't think I'll have much time to debug this, also because Qt build system is incredibly messed up and I'm not familiar with it, but in general we should have a better way to deal with this upgrades.  Why not doing the native build in all builds?  It's wasteful, but at least we don't have the problem of having broken packages between updates.

giordano:
Also, why don't we hit ccache here?  Rebuilds with what should be warm cache take way too long.

giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/bca4972591bcae84eecf051c8039a2f491a38601/Q/Qt6Base/build_tarballs.jl#L181 what is LLVM needed for?

giordano:
I found https://github.com/JuliaPackaging/Yggdrasil/pull/2856#discussion_r615281072 but I don't think we should use LLVM_full_jll, but rather the compiler-rt stuff.

barche:
Continuing discussion in PR #7629

PR #7629, opened 2023-11-06
[Qt6Base] Some improvements to build recipe

giordano:
Tested only on 64-bit win


barche:
Mac error is `library not found for -lclang_rt.osx`, does that mean we do need LLVM_full? I don't remember if I tried multiple options when adding this.

barche:
It seems that building only the host tools is not supported, so I'd like to propose the introduction of `Qt6Base_host_jll` that will only contain the host build, which can then be reused in the target builds. We would have to do the same for the other Qt6 packages that require a host version. The advantages of this approach are that we don't need to list a huge amount of separate `HostBuildDependency` declarations and the host tools are still built only once, but no longer break the dependent packages. The downside is that we introduce a few redundant JLLS.

If you agree @giordano then this is ready to merge for me and it should create the `Qt6Base_host_jll` package.

giordano:
While that'd work, I'm not a big fan of this solution.  I haven't seen why we can't do the host build, it's a bit surprising as we should have all the tools to do it.

barche:
Right now, I'm stuck on getting link errors like these:

https://buildkite.com/julialang/yggdrasil/builds/6696#018be863-729c-4966-9a93-d892b3332b2c/6-659

giordano:
That's because we need the host build dependencies 🙂

barche:
I added all dependencies as `HostBuildDependency` using the for loop at line 210. Or at least that's what I hope that loop does.

PR #7632, opened 2023-11-06
[GDAL] Bump to 3.7.3

jeremiahpslewis:


G/GDAL/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "GDAL"
-upstream_version = v"3.7.2"
+upstream_version = v"3.7.3"
```
giordano, line 6: Need to change the source as well, no?

jeremiahpslewis: Definitely! Oof.

giordano:
A linker error is unlikely to be magically fixed by changing compiler version, especially if this is linking to OpenMPI 5 which deosn't exist at all for 32-bit systems (but I can't see the log, so can't know whether that's the case)

jeremiahpslewis:
Error: ``/opt/i686-linux-gnu/bin/../lib/gcc/i686-linux-gnu/9.1.0/../../../../i686-linux-gnu/bin/ld: warning: libmpi.so.40, needed by /workspace/destdir/lib/libhdf5.so.310, not found (try using -rpath or -rpath-link)``

PR #7637, opened 2023-11-07
Bump fzf to 0.43.0

Moelf:


giordano:
```
[08:36:04] cgo: C compiler "cc" not found: exec: "cc": executable file not found in $PATH
```
Wut

Moelf:
super confused

Moelf:
oh `gcc` doens't giev you cc? oh well, will look into it after

giordano:
We generate `cc` symlinks on all platforms.

Moelf:
you're saying this is not my problem ;)?

giordano:
The missing `:c` compiler was perhaps part of the problem, but now that's explicitly there I don't see why `cc` can't be found (especially since this works everywhere else).  I'll need to debug this when I get some time later this week.

giordano:
Ah, it was the missing `:c`, and I didn't realise it was misspelled before :slightly_smiling_face:

PR #7642, opened 2023-11-08
Wizard recipe: ginkgo-v1.6.0

youwuyou:

* Package name: ginkgo
* Version: v1.6.0


youwuyou:
The current source code used for this specific JLL build script contains a C API that is yet under continuous development for [Ginkgo.jl](https://github.com/youwuyou/Ginkgo.jl).

NOTE: The previous [ginkgo library](https://ginkgo-project.github.io/) v1.4.0 built was wrongly misspelled as [Gingko](https://github.com/JuliaPackaging/Yggdrasil/tree/master/G/Gingko).

G/ginkgo/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+mkdir build && cd build
+cmake -DGINKGO_BUILD_TESTS=OFF -DGINKGO_BUILD_BENCHMARKS=ON -DGINKGO_BUILD_EXAMPLES=ON -DGINKGO_DOC_GENERATE_EXAMPLES=OFF -G "Ninja" ../
+ninja -j${nproc} 
+DESTDIR=${WORKSPACE}/destdir/shared ninja install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc")
```
giordano, line 26: Why only one platform?

G/ginkgo/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libginkgo", :libginkgo, "shared/usr/local/lib"),
+    LibraryProduct("libginkgo_device", :libginkgo_device, "shared/usr/local/lib"),
+    LibraryProduct("libginkgo_cuda", :libginkgo_cuda, "shared/usr/local/lib"),
+    LibraryProduct("libginkgo_reference", :libginkgo_reference, "shared/usr/local/lib"),
+    LibraryProduct("libginkgo_omp", :libginkgo_omp, "shared/usr/local/lib"),
+    LibraryProduct("libginkgo_dpcpp", :libginkgo_dpcpp, "shared/usr/local/lib"),
+    LibraryProduct("libginkgo_hip", :libginkgo_hip, "shared/usr/local/lib")
```
giordano, line 38: Why not using standard locations (`libdir`) for the shared libraries?

G/ginkgo/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"12.1.0")
```
giordano, line 46: https://github.com/JuliaPackaging/Yggdrasil/blob/dc3c29259b032a7cd5db6cf5a7b8165e17951451/CONTRIBUTING.md#compatibility-tips

G/ginkgo/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/youwuyou/ginkgo.git", "c4128a21b1114bedfc19153a9508b2bd3b54954f")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd ginkgo/
```
giordano, line 16:
```suggestion
cd $WORKSPACE/srcdir/ginkgo/
```

G/ginkgo/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+mkdir build && cd build
+cmake -DGINKGO_BUILD_TESTS=OFF -DGINKGO_BUILD_BENCHMARKS=ON -DGINKGO_BUILD_EXAMPLES=ON -DGINKGO_DOC_GENERATE_EXAMPLES=OFF -G "Ninja" ../
+ninja -j${nproc} 
+DESTDIR=${WORKSPACE}/destdir/shared ninja install
```
giordano, line 20: If anything this should be
```suggestion
DESTDIR=${prefix}/shared ninja install
```
but why changing the prefix at all?

G/ginkgo/build_tarballs.jl
youwuyou: Added platform support

G/ginkgo/build_tarballs.jl
youwuyou: Changing the install path

G/ginkgo/build_tarballs.jl
youwuyou: Using v7.1.0 compiler

G/ginkgo/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/youwuyou/ginkgo.git", "c4128a21b1114bedfc19153a9508b2bd3b54954f")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/ginkgo/
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=/workspace/destdir -DGINKGO_BUILD_TESTS=OFF -DGINKGO_BUILD_BENCHMARKS=OFF -DGINKGO_BUILD_EXAMPLES=OFF -DGINKGO_DOC_GENERATE_EXAMPLES=OFF -G "Ninja" ../
```
imciner2, line 17: You need to include `-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}` in your CMake command to get everything setup properly for the build environment.

Also, it would be cleaner to put each option on its own line (makes it easier for reading and also diffs in the future).

G/ginkgo/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
```
imciner2, line 35: Why this choice of platforms? The release notes suggest that it should build on Windows and MacOS as well (https://github.com/ginkgo-project/ginkgo/releases/tag/v1.6.0).

youwuyou: For the rest of the platform build there are still some issues, we will try to fix that out and fix them later.

G/ginkgo/build_tarballs.jl
youwuyou: Fixed the alignment and added more flags explicitly

G/ginkgo/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+sources = [
+    GitSource("https://github.com/youwuyou/ginkgo.git", "c4128a21b1114bedfc19153a9508b2bd3b54954f")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/ginkgo/
+mkdir build && cd build
+cmake \
+    -DCMAKE_INSTALL_PREFIX=/workspace/destdir \
```
imciner2, line 18:
```suggestion
    -DCMAKE_INSTALL_PREFIX=${prefix} \
```

G/ginkgo/build_tarballs.jl
imciner2: But you basically only want the Linux platforms, right?

If so, this list can be simplified to be
```
platforms = filter!(Sys.islinux, supported_platforms())
```

G/ginkgo/build_tarballs.jl
youwuyou: It unfortunately did not work out well with ${prefix}, so we switched to the absolute path. A `usr/local` prefix would be added when using prefix and it seems like some packages underlying BinaryBuilder is managing this?

G/ginkgo/build_tarballs.jl
youwuyou: Changed to the simplified expression

G/ginkgo/build_tarballs.jl
imciner2: It definitely should work with `${prefix}`, because that is what all other packages are using for this. Can you push this change to here so we can see the build logs?

G/ginkgo/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libginkgo", :libginkgo),
+    LibraryProduct("libginkgo_device", :libginkgo_device),
+    LibraryProduct("libginkgo_cuda", :libginkgo_cuda),
+    LibraryProduct("libginkgo_reference", :libginkgo_reference),
+    LibraryProduct("libginkgo_omp", :libginkgo_omp),
+    LibraryProduct("libginkgo_dpcpp", :libginkgo_dpcpp),
+    LibraryProduct("libginkgo_hip", :libginkgo_hip)
```
vchuravy, line 48: What are these libraries used for? The build currently has these turned off? Does libginko function like a forwarder/trampoline? If so you may not want to declare these as library product since that means they are getting eagerly loaded into the process.

vchuravy: Answered my own question by running `lld` over `libginko`.

giordano:
@youwuyou thanks for your contribution!

PR #7649, opened 2023-11-09
CDE [new package]

joegilkes:
Adding an executable binding for the [Chemistry Discovery Engine](https://github.com/HabershonLab/cde) (CDE)

### DIFF
--- a/C/CDE/build_tarballs.jl
+++ b/C/CDE/build_tarballs.jl
@@ -5,7 +5,12 @@ version = v"1.0.0"
 sources = [GitSource("https://github.com/HabershonLab/cde", "049cd0d2da73a0c5bb340ebdb9a78e35348fe49a")]
 
 script = raw"""
-export LDFLAGS="-L${libdir}"
+if [[ ${target} == *w64* ]]; then
+    exe_out="${bindir}/cde.exe"
+else
+    exe_out="${bindir}/cde"
+fi
+
 cd cde
 gfortran -O3 -c src/constants.f90 -o src/constants.o -J src
 gfortran -O3 -c src/globaldata.f90 -o src/globaldata.o -J src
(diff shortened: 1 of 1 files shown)


C/CDE/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+name = "CDE"
+version = v"1.0.0"
+sources = [GitSource("https://github.com/HabershonLab/cde", "049cd0d2da73a0c5bb340ebdb9a78e35348fe49a")]
+
+script = raw"""
+if [[ ${target} == *w64* ]]; then
+    exe_out="${bindir}/cde.exe"
+else
+    exe_out="${bindir}/cde"
+fi
```
giordano, line 12: `exeext` is our custom env variable for the extension of executables:
```suggestion
exe_out="${bindir}/cde${exeext}"
```

joegilkes: Ah thanks, that's exactly what I was looking for!

giordano:
Wouldn't using the [`Makefile`](https://github.com/HabershonLab/cde/blob/049cd0d2da73a0c5bb340ebdb9a78e35348fe49a/Makefile) work?

joegilkes:
Yes, I was just having a lot of issues getting OpenBLAS to link properly when I was testing locally so it was easier to have it line-by-line. I suppose the only functional difference would be that make can compile in parallel, so happy to move back to the Makefile if that's the desired behaviour.

giordano:
I think it'd be more reliable going forward because you'd have to keep the list of files up to date in a single place (the Makefile) rather than here as well.

joegilkes:
You're right, that makes much more sense. I've got it working locally but the Makefile needs some changes to support parallel compilation, which I'll have to get to tomorrow.

joegilkes:
Parallel compilation still isn't working, I can't get the object file dependencies to correctly prevent files depending on other modules to not be compiled until they are ready, resulting in compilation failures on anything higher than 2 cores. 

However, CDE is a relatively small program to compile so this shouldn't be too much of an issue. I'll update the package here once I've got parallel compilation working, for now is it okay if we leave it with serial compilation?

giordano:
Yeah, it's fine for the time being, but fixing dependencies for parallel compilation can be useful to others as well :slightly_smiling_face: 

Thanks for your contribution!

PR #7650, opened 2023-11-09
Build the aeron media driver executable

sefffal:
Aeron is a low latency IPC/UDP communication system. 
The "client" is a library that is linked in to a program, but the actual coordination and sending of data is done by a component called the "media driver".
We now include the media driver as an executable so that users on supported platforms can use the complete basic Aeron stack without having to build anything themselves.

This PR doesn't improve platform support for the client; it just adds the media driver executable for the already working platforms.


A/Aeron/build_tarballs.jl
```diff
@@ -15,46 +15,61 @@ script = raw"""
 apk update
 apk add openjdk11 hdrhistogram-c-dev libbsd-dev util-linux-dev
 cd $WORKSPACE/srcdir/aeron
-mkdir build && cd build
-CMAKE_FLAGS=(-DCMAKE_INSTALL_PREFIX=$prefix
--DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}
+mkdir -p /sys/devices/system/cpu/
+echo 1-${nproc} > /sys/devices/system/cpu/possible
```
giordano, line 19: Uhm, why this?

A/Aeron/build_tarballs.jl
```diff
@@ -15,46 +15,61 @@ script = raw"""
...
 -DAERON_INSTALL_TARGETS=ON)
+mkdir build && cd build
+export LDFLAGS="-L${libdir}"
 cmake .. "${CMAKE_FLAGS[@]}"
-make -j${nproc}
+make -j${nproc} aeron_client aeron_driver_static
 make install
+cp /usr/local/lib/*aeron* ${libdir}
+mkdir -p ${bindir}
+cp /usr/local/bin/aeronmd ${bindir}
```
giordano, line 46: Maybe there's a way to change the installation directory instead of manually moving files around afterwards.

A/Aeron/build_tarballs.jl
```diff
@@ -63,4 +78,7 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"11.1.0")
+build_tarballs(
+    ARGS, name, version, sources, script, platforms, products, dependencies;
+    julia_compat="1.6", preferred_gcc_version = v"11.1.0")
+# BinaryBuilder.runshell(Platform("x86_64", "linux", libc="glibc"))
```
giordano, line 84:
```suggestion
# BinaryBuilder.runshell(Platform("x86_64", "linux", libc="glibc"))

```

A/Aeron/build_tarballs.jl
```diff
@@ -63,4 +78,7 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"11.1.0")
+build_tarballs(
+    ARGS, name, version, sources, script, platforms, products, dependencies;
+    julia_compat="1.6", preferred_gcc_version = v"11.1.0")
```
giordano, line 83: This is a C++ library, so if you use GCC 11 it may not work in practice with Julia v1.6 (depending on the `GLIBCXX` symbols used): https://github.com/JuliaPackaging/Yggdrasil/blob/5a60de3a72c2a07a2ddd74781a12e6173a279c39/CONTRIBUTING.md#compatibility-tips

sefffal: Thanks! 
I think we are actually using only the c library, but it's obvious how to turn off the C++ build. It always wants to build both.

Regardless I'll try without the preferred gcc version.

A/Aeron/build_tarballs.jl
giordano: Which recipe?  Can you share the piece of code which is using that file?

sefffal: Looks like this is not needed in the end; thanks for catching it!

A/Aeron/build_tarballs.jl
sefffal: I removed the preferred_gcc_version.

A/Aeron/build_tarballs.jl
sefffal: I spent some time on this without any luck. At least, I couldn't find a cleaner way then just moving them.

sefffal:
C11 is required so I put back `preferred_gcc_version = v"11.1.0",`

A/Aeron/build_tarballs.jl
giordano: What happened to the line `CMAKE_FLAGS=(-DCMAKE_INSTALL_PREFIX=$prefix` whose goal was exactly to set the destination prefix?

sefffal: It doesn't do what we want exactly, it places it in $prefix/lib and $prefix/binaries

giordano: It places _what_ in where?

sefffal: I'll check through this again

sefffal: Ok fixed now. Thanks for your patience!

PR #7663, opened 2023-11-12
[Ipopt] Upgrade to version 3.14.3

amontoison:
I also removed the static library `libipopt.a` to reduce the size of the tarballs.


odow:
I don't understand

amontoison:
What do you not understand Oscar?
The errors on Mac are related to the new linker `lld`. (@gbaraldi)

odow:
The cause of https://github.com/jump-dev/AmplNLWriter.jl/pull/163#issuecomment-1807223107

amontoison:
Me neither but it could help us to solve the issue in `AMPLNLWriter.jl`.

PR #7665, opened 2023-11-12
[SCS] update to 3.2.4 and fix dependencies of SCS_GPU and SCS_MKL

odow:
cc @kalmarek 

SCS.jl currently has an unusual setup for lazy loading.

It has a package dependency on `SCS_GPU_jll` and `SCS_MKL_jll`:
https://github.com/jump-dev/SCS.jl/blob/e24ac32e585357a584c379eec41a601fb93b7cb9/Project.toml#L18-L19
and then lazily loads them using `Requires` if `CUDA_jll` or `MKL_jll` is loaded:
https://github.com/jump-dev/SCS.jl/blob/e24ac32e585357a584c379eec41a601fb93b7cb9/src/SCS.jl#L17-L36

This is back-to-front. The `SCS_xxx_jll` packages should have dependencies on CUDA_jll and MKL_jll, and the user should load them via `using SCS_GPU_jll`, not `using CUDA_jll`.

I think this design was chosen in the olde-times so that we wouldn't have to download the heavy CUDA_jll or MKL_jll, and so that we could also enforce strict compat bounds on which versions were supported.

My work-around is to add `SCS_jll` as a dependency, so that SCS.jl can declare a compat on SCS_jll and it will transitively apply to SCS_GPU_jll and SCS_MKL_jll.

This will require some changes to SCS.jl, but we're safe to do this because SCS is pinned to `3.2.3`:
https://github.com/jump-dev/SCS.jl/blob/e24ac32e585357a584c379eec41a601fb93b7cb9/Project.toml#L20

I'm not sure if one can build dependent packages like this, so I probably need to build 3.2.4 or SCS_jll first, and then the GPU and MKL versions?

### DIFF
--- a/S/SCS_GPU/build_tarballs.jl
+++ b/S/SCS_GPU/build_tarballs.jl
@@ -40,7 +40,6 @@ products = [
 
 dependencies = [
     Dependency("OpenBLAS32_jll", v"0.3.10"),
-    Dependency("CUDA_jll"),
 ]
 
 for platform in platforms
--- a/S/SCS_GPU/build_tarballs.jl
+++ b/S/SCS_GPU/build_tarballs.jl
@@ -41,7 +41,6 @@ products = [
 dependencies = [
     Dependency("OpenBLAS32_jll", v"0.3.10"),
     Dependency("CUDA_jll"),
-    Dependency("SCS_jll"; compat = "=$version"),
 ]
 
 for platform in platforms
--- a/S/SCS_GPU/build_tarballs.jl
+++ b/S/SCS_GPU/build_tarballs.jl
@@ -6,11 +6,11 @@ include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
 include(joinpath(YGGDRASIL_DIR, "platforms", "cuda.jl"))
 
 name = "SCS_GPU"
-version = v"3.2.3"
+version = v"3.2.4"
 
 # Collection of sources required to build SCSBuilder
 sources = [
-    GitSource("https://github.com/cvxgrp/scs.git", "f5f054be7dd71ee0d80c4c0eec0df1e9f0ccb123")
+    GitSource("https://github.com/cvxgrp/scs.git", "9024b8ccc1bba6ee797440fb22354cadb9c81839")
 ]
 
 # Bash recipe for building across all platforms
@@ -39,7 +39,9 @@ products = [
 ]
 
 dependencies = [
-    Dependency("OpenBLAS32_jll", v"0.3.10")
+    Dependency("OpenBLAS32_jll", v"0.3.10"),
+    Dependency("CUDA_jll"),
+    Dependency("SCS_jll"; compat = "=$version"),
 ]
 
 for platform in platforms
--- a/S/SCS_MKL/build_tarballs.jl
+++ b/S/SCS_MKL/build_tarballs.jl
@@ -39,7 +39,6 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency(PackageSpec(;name="MKL_jll", version=v"2023.1")),
-    Dependency("SCS_jll"; compat = "=$version"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well
--- a/S/SCS_MKL/build_tarballs.jl
+++ b/S/SCS_MKL/build_tarballs.jl
@@ -2,11 +2,11 @@ using Pkg
 using BinaryBuilder
 
 name = "SCS_MKL"
-version = v"3.2.3"
+version = v"3.2.4"
 
 # Collection of sources required to build SCSBuilder
 sources = [
-    GitSource("https://github.com/cvxgrp/scs.git", "f0c23340da03bcd663072dc4a90cf1aab8968c61")
+    GitSource("https://github.com/cvxgrp/scs.git", "9024b8ccc1bba6ee797440fb22354cadb9c81839")
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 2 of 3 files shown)


kalmarek:
yes, that's precisely the solution I had in mind

imciner2:
> My work-around is to add SCS_jll as a dependency, so that SCS.jl can declare a compat on SCS_jll and it will transitively apply to SCS_GPU_jll and SCS_MKL_jll.

Does `SCS_GPU_jll` or `SCS_MKL_jll` need the libraries from `SCS_jll` to actually function, or are they standalone libraries? If not, I think the linkage between them should happen at a higher level than the jll packages, such as the SCS.jl package (or any package extensions it might have), and that is where the compat entries should live.

As for the GPU-based version, it would be good to update it to use the new CUDA build methods and dependencies now, since that has matured since this recipe was first written and it has actually become quite nice to use.

odow:
> I think the linkage between them should happen at a higher level than the jll packages, such as the SCS.jl package

That is the current design.

But I don't want SCS.jl to depend on SCS_GPU_jll, and I want this to work on Julia 1.6 (no package extensions).

Once 1.9+ is the LTS, then yes, we could implement this as a package extension and drop the extra linkage between SCS_GPU_jll and SCS_jll.

But also: likely no one is (or ever will) using SCS_GPU_jll or SCS_MKL_jll without SCS.jl.

odow:
> As for the GPU-based version, it would be good to update it to use the new CUDA build methods and dependencies now

Any pointers? I'm well out of my wheelhouse on this.

imciner2:
> Any pointers? I'm well out of my wheelhouse on this.

Take a look at the new build recipe for SuiteSparseGPU: https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SuiteSparse/SuiteSparse_GPU%407/build_tarballs.jl.

odow:
I think I need more pointers. They look pretty similar? What's wrong with our current script?

imciner2:
Will the higher level SCS package require any glue code for using the `SCS_GPU_jll` or `SCS_MKL_jll` package? I am building a similar system for OSQP (where we have separate MKL and CUDA libraries coming for 1.0), and I am doing this by making three high-level packages: the existing `OSQP`, `OSQPCUDA` and `OSQMKL`. Those three then have the dependencies on the jll packages, and so the user just does `using OSQPCUDA` to get the CUDA-based jll and its dependencies. Those three packages will then have the various compat bounds with each other (and this doesn't require package extensions).

odow:
Yes, they have glue code.

But, yes, we could create `SCSCUDA` and `SCSMKL` which each depend one `SCS.jl`.

That feels pretty clunky though.

S/SCS_GPU/build_tarballs.jl
```diff
@@ -39,7 +39,9 @@ products = [
 ]
 
 dependencies = [
-    Dependency("OpenBLAS32_jll", v"0.3.10")
+    Dependency("OpenBLAS32_jll", v"0.3.10"),
+    Dependency("CUDA_jll"),
+    Dependency("SCS_jll"; compat = "=$version"),
```
odow, line 44:
```suggestion
```

S/SCS_MKL/build_tarballs.jl
```diff
@@ -38,7 +38,8 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    BuildDependency(PackageSpec(;name="MKL_jll", version=v"2023.1")),
+    Dependency(PackageSpec(;name="MKL_jll", version=v"2023.1")),
+    Dependency("SCS_jll"; compat = "=$version"),
```
odow, line 42:
```suggestion
```

amontoison:
For `SCC_jll.jl` and `SCS_MKL_jll.jl` packages, is it not easier to compile `SCS` with `LBT`?
For Ipopt, a MKL version was developed [here](https://github.com/JuliaPackaging/Yggdrasil/blob/master/C/Coin-OR/IpoptMKL/build_tarballs.jl) but it's not used anymore.

odow:
I haven't taken a look at what is involved to build SCS with LBT

odow:
It's also currently limited to Linux. @kalmarek might be able to give pointers on why

S/SCS_GPU/build_tarballs.jl
```diff
@@ -39,7 +39,8 @@ products = [
 ]
 
 dependencies = [
-    Dependency("OpenBLAS32_jll", v"0.3.10")
+    Dependency("OpenBLAS32_jll", v"0.3.10"),
+    Dependency("CUDA_jll"),
```
odow, line 43:
```suggestion
```

imciner2:
@giordano any ideas about the pipeline launch error?

> Generating meta.json...
> ERROR: LoadError: SystemError: opening file "/tmp/jl_5gUQ3N/SCS_GPU.meta.json": No such file or directory

giordano:
No, I've never seen that error message and restarting the job fails with a different error:
```
2023-11-13 11:06:39 ERROR  Unrecoverable error, skipping retries
2023-11-13 11:06:39 FATAL  Failed to upload and process pipeline: POST https://agent.buildkite.com/v3/jobs/018bc85b-cd50-452b-9ac0-c15767a7d330/pipelines: 422 The key "S-SCS--i686-linux-gnu" has already been used by another step in this build
```
:confused:

kalmarek:
@amontoison SCS_MKL uses essentially different sparse solver (PARDISO) which is not available with openblas. We could link scs against lbt (when the new LTS supports it), but SCS_MKL is just a separate solver:
https://github.com/cvxgrp/scs/tree/master/linsys

At some point I actually contemplated using julia wrappers to PARDISO/oneMKL to create a linsys compatible solver, but never found enough time to give it a real try.

as for linux-only: I believed scs_mkl failed to compile on windows and since I neither use nor even have access to a windows machine I disabled it. If somebody is willing to spend time on this, give it a go ;

amontoison:
@kalmarek Ok I understand now, MKL PARDISO is one of the main linear solver. It makes senses to create a `SCS_MKL`. The best solution would be able to `dlopen` at runtime `libmkl_rt` to use the additional linear solver but it should be implemented upstream.

Ipopt is doing that with HSL / PARDISO solvers.

odow:
Perhaps I need to try the builds separately.

odow:
I've split this into multiple PRs: #7671, #7672, #7673.

It looks like the issue is with the GPU build. But I don't understand what or why. (I've never actually used a GPU...)

PR #7669, opened 2023-11-13
New recipe: X13as

bbejanov:
Add recipe for [X13-ARIMA-SEATS](https://www.census.gov/data/software/x13as.X-13ARIMA-SEATS.html#list-tab-635278563)  seasonal adjustment program.

### DIFF
--- a/X/X13as/build_tarballs.jl
+++ b/X/X13as/build_tarballs.jl
@@ -28,7 +28,7 @@ then
 else
     make -j${nproc} -f makefile.gf
 fi
-mkdir ${bindir}
+mkdir -p ${bindir}
 install -Dvm 755 "x13as_ascii${exeext}" "${bindir}/x13as_ascii${exeext}"
 popd
 ########## Second, build and install the html version
@@ -47,7 +47,7 @@ install -Dvm 755 "x13as_html${exeext}" "${bindir}/x13as_html${exeext}"
 popd
 ########## Third, install the docs and the test examples
 docsdir="${prefix}/docs"
-mkdir "${docsdir}"
+mkdir -p "${docsdir}"
 install -Dvm 644 "x13as/testairline.spc" "${docsdir}/testairline.spc"
 install -Dvm 644 "x13as/docs/docx13as.pdf" "${docsdir}/docx13as.pdf"
 install -Dvm 644 "x13as/docs/qrefX13ASunix.pdf" "${docsdir}/qrefX13ASunix.pdf"
--- a/X/X13as/build_tarballs.jl
+++ b/X/X13as/build_tarballs.jl
@@ -77,4 +77,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"13.2.0")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
--- a/X/X13as/build_tarballs.jl
+++ b/X/X13as/build_tarballs.jl
@@ -24,9 +24,9 @@ if [[ "${target}" == *-apple-* ]]
 then
     # the -static flag doesn't work in macos; also disable -s linker flag
     sed -e "s/\-static//" -i makefile.gf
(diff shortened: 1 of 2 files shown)


X/X13as/build_tarballs.jl
```diff
@@ -0,0 +1,80 @@
...
+    FileProduct("docs/qrefX13ASHTMLunix.pdf", :qrefX13ASHTMLunix_pdf)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"13.2.0")
```
giordano, line 80: Please see https://github.com/JuliaPackaging/Yggdrasil/blob/82657bf5b3e19935dddbd39ec0370616fc627209/CONTRIBUTING.md#compatibility-tips.  The fact that this works for libgfortran v3 and v4 (gcc 6 and 7) is a strong evidence that gfortran 13 isn't needed.

X/X13as/build_tarballs.jl
```diff
@@ -0,0 +1,80 @@
...
+
+name = "X13as"
+version = v"1.1.60"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://www2.census.gov/software/x-13arima-seats/x13as/unix-linux/program-archives/x13as_asciisrc-v1-1-b60.tar.gz", "2bd53953a1bdd238a128b89e95e8e5fc14f33aa4a9e4c7f0fc3fe7323b73131c", unpack_target="./x13as_asciisrc"),
+    ArchiveSource("https://www2.census.gov/software/x-13arima-seats/x13as/unix-linux/program-archives/x13as_htmlsrc-v1-1-b60.tar.gz", "642f6b6a969c5c311252ca83845ea391ab8c3d59840c4dc2508f9c86095c7757", unpack_target="./x13as_htmlsrc"),
+    ArchiveSource("https://www2.census.gov/software/x-13arima-seats/x13as/unix-linux/program-archives/x13as_ascii-v1-1-b60.tar.gz", "593e6b63024181c9550c281587bfeb28b2dcf1658f4495d42eb8933c6868ee36"),
+    ArchiveSource("https://www2.census.gov/software/x-13arima-seats/x13as/unix-linux/program-archives/x13as_html-v1-1-b60.tar.gz", "a9f55329373092c33ed5b1908094c0dc037cc05bf8789bd4f7156abcaa2200ed"),
```
giordano, line 13: Any particular reason why these aren't independent packages?

X/X13as/build_tarballs.jl
bbejanov: Thanks for pointing that out. I have removed the `preferred_gcc_version=` option altogether.

X/X13as/build_tarballs.jl
bbejanov: Yes. The first two are two source distributions for two versions of the same executable. They provide the same algorithms, the only difference is that one writes its output in ascii formatted tables, the other in html files. They are always both released at the same time by the US Census Bureau and have the same version numbers. I don't think this is a significant difference to justify maintaining two separate packages for them. 

The third and the fourth files are included here only for the pdf files with documentation in them. For reason unknown to me, the US Census Bureau doesn't include the pdf documentation files with the source distributions, it publishes them separately. 

I hope this answers your question. If you insist, I can separte the ascii and the html versions; otherwise I prefer to keep them together in one package. Please let me know.

bbejanov:
Hi @giordano,

Thanks for your review and apologies for the delay in my followup - it has been a crazy week for me.  I think I have addressed all your concerns. Please take a look and let me know if anything else needs fixing.

giordano:
The license file is missing: https://docs.binarybuilder.org/stable/build_tips/#Installing-the-license-file

bbejanov:
Done.  Thanks for catching this.

bbejanov:
Hi @giordano, 
Please let me know if there's anything else that needs fixing and I'll be happy to address it.  Otherwise, I think this PR is ready to merge.  Thanks for your help!
Boyan

PR #7672, opened 2023-11-13
[SCS_MKL] update to v3.2.4

odow:
x-ref https://github.com/JuliaPackaging/Yggdrasil/pull/7665

@kalmarek I've updated the MKL version. Is that okay?

We could also try building this on other platforms.

### DIFF
--- a/S/SCS_MKL/build_tarballs.jl
+++ b/S/SCS_MKL/build_tarballs.jl
@@ -26,9 +26,9 @@ cp out/libscs*.${dlext} ${libdir}
 platforms = [
     Platform("x86_64", "linux"; libc="glibc"),
     # Platform("i686", "linux"; libc="glibc"),
-    Platform("x86_64", "macos"),
+    # Platform("x86_64", "macos"),
     # Platform("i686", "windows"),
-    Platform("x86_64", "windows"),
+    # Platform("x86_64", "windows"),
 ]
 
 # The products that we will ensure are always built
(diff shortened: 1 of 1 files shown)


S/SCS_MKL/build_tarballs.jl
```diff
@@ -38,7 +38,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    BuildDependency(PackageSpec(;name="MKL_jll", version=v"2023.1")),
+    Dependency(PackageSpec(;name="MKL_jll", version=v"2023.2"); compat = "=2023.2.0"),
```
odow, line 41:
```suggestion
    Dependency("MKL_jll"; compat = "=2023.2.0"),
```

amontoison:
`MKL_jll.jl` is only available on five platforms (windows x2, Mac, Linux x2).
The error on Mac is related to the new version of `clang` on Yggdrasil. It could be fixed with a different version of llvm (see [here](https://github.com/JuliaPackaging/Yggdrasil/blob/82657bf5b3e19935dddbd39ec0370616fc627209/S/SCOTCH/build_tarballs.jl#L110)).
For the failure on Windows, it's because they use `Clong` for 64 bits integer but it's a 32 bits integer in Windows.
A quick fix is to replace all `Clong` by `Clonglong`.
The best solution is to replace `Clong` by `C_INT64_T` upstream.

odow:
let's just leave as linux for now and come back to it

amontoison:
Can I merge the PR?

odow:
Yes, I think so

amontoison (approved):

PR #7673, opened 2023-11-13
[SCS_GPU] update to v3.2.4

odow:
x-ref https://github.com/JuliaPackaging/Yggdrasil/pull/7665

### DIFF
--- a/S/SCS_GPU/build_tarballs.jl
+++ b/S/SCS_GPU/build_tarballs.jl
@@ -6,11 +6,11 @@ include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
 include(joinpath(YGGDRASIL_DIR, "platforms", "cuda.jl"))
 
 name = "SCS_GPU"
-version = v"3.2.3"
+version = v"3.2.4"
 
 # Collection of sources required to build SCSBuilder
 sources = [
-    GitSource("https://github.com/cvxgrp/scs.git", "f5f054be7dd71ee0d80c4c0eec0df1e9f0ccb123")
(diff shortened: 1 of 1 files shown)


imciner2:
Looking at this, I think the issue is the CUDA version that is being requested. It looks like SCS is wanting CUDA 11.3, but that is not in Ygg, only versions 11.4+ are available in CUDA_SDK.

Reading the note in the recipe, it says

> ## scs uses CUSPARSE_SPMV_CSR_ALG1 which requires CUDA-11.3

Looking at the CUDA docs, that is available in CUDA 11.3 or later, so SCS should build with the newer versions of CUDA. My suggestion is to remove lines 33 and 34 to let it build on all Ygg supported CUDA versions and see what happens.

imciner2:
Yep, that fixed it. Thanks for your patience!

kalmarek:
> Looking at this, I think the issue is the CUDA version that is being requested. It looks like SCS is wanting CUDA 11.3, but that is not in Ygg, only versions 11.4+ are available in CUDA_SDK.
> 
> Reading the note in the recipe, it says
> 
> > scs uses CUSPARSE_SPMV_CSR_ALG1 which requires CUDA-11.3
> 
> Looking at the CUDA docs, that is available in CUDA 11.3 or later, so SCS should build with the newer versions of CUDA. My suggestion is to remove lines 33 and 34 to let it build on all Ygg supported CUDA versions and see what happens.

@imciner2 That's why I find it very useful to include those rationale comments in the code! Just so that somebody like more knowledgeable look at them and say -- that doesn't make sense anymore ;) Thanks!

PR #7674, opened 2023-11-13
[OpenBLAS] Upgrade to v0.3.25

giordano:


giordano:
@vchuravy I don't know if you have the strength to look at this, but powerpc builds are falling 😞

imciner2:
```
[22:49:16] dynamic_power.c:205:63: error: missing binary operator before token "("
[22:49:16]  #if defined(_AIX) || (defined(__has_builtin) && !__has_builtin(__builtin_cpu_is))
[22:49:16]                                                                ^
[22:49:16] dynamic_power.c:230:63: error: missing binary operator before token "("
[22:49:16]  #if defined(_AIX) || (defined(__has_builtin) && !__has_builtin(__builtin_cpu_supports))
[22:49:16]                                                                ^
[22:49:16] make[1]: *** [<builtin>: dynamic_power.o] Error 1
```

Which was caused by https://github.com/OpenMathLib/OpenBLAS/pull/4298.

I think it doesn't like the `__has_builtin` being in the same statement as the others? All other uses of `__has_builtin` in the code are on their own in the `#if` statement.

imciner2:
Actually, another thing is GCC apparently only added `__has_builtin` in GCC 10, so I wonder if the use of GCC 6 here is breaking their checks since `__has_builtin` doesn't actually exist, and therefore the preprocessor doesn't recognize the statement as valid.

PR #7676, opened 2023-11-14
Delete wrongly named ginkgo recipe

vchuravy:


vchuravy:
#7642 added the jll with it's correct name.

PR #7685, opened 2023-11-15
[Infernal] Update to v1.1.5

marcom:
- use all-in-one tarball from http://eddylab.org/infernal/
- adds support for ARM processors

### DIFF
--- a/I/Infernal/build_tarballs.jl
+++ b/I/Infernal/build_tarballs.jl
@@ -1,29 +1,18 @@
-# Note that this script can accept some limited command-line arguments, run
-# `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, BinaryBuilderBase, Pkg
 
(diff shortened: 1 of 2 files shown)


I/Infernal/build_tarballs.jl
```diff
@@ -93,10 +78,8 @@ products = [
     ExecutableProduct("esl-weight", :esl_weight),
 ]
 
-# Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency(PackageSpec(name="GSL_jll", uuid="1b77fbbe-d8ee-58f0-85f9-836ddc23a7a4"))
```
giordano, line 82: #3024
```suggestion
    Dependency(PackageSpec(name="GSL_jll", uuid="1b77fbbe-d8ee-58f0-85f9-836ddc23a7a4"); compat="~2.7.2")
```

PR #7689, opened 2023-11-16
[SPRAL] Upgrade to v2023.11.15

amontoison:
@jfowkes

### DIFF
--- a/S/SPRAL/build_tarballs.jl
+++ b/S/SPRAL/build_tarballs.jl
@@ -3,41 +3,34 @@
 using BinaryBuilder, Pkg
 
 name = "SPRAL"
-version = v"2023.09.07"
+version = v"2023.11.15"
 
 # Collection of sources required to complete build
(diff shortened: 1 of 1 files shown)


amontoison:
Meson is unable to determine a dependency order when we cross-compile modern Fortran files...
I will open an issue.

jfowkes:
@amontoison `meson.build` version bumped and `v2023.11.15` re-released as requested.

jfowkes:
> Meson is unable to determine a dependency order when we cross-compile modern Fortran files... I will open an issue.

But it builds fine with meson locally in our tests?

amontoison:
> > Meson is unable to determine a dependency order when we cross-compile modern Fortran files... I will open an issue.
> 
> But it builds fine with meson locally in our tests?

Yes, you are right.
It's probably related to the cross-compilation.
It's not a good news for GALAHAD...

PR #7693, opened 2023-11-16
[BLAS32] Add a recipe

amontoison:
I will add a `BLAS_jll.jl` with the release `3.12.0`.
The new release will add an option to add the  suffix `_64` automatically.

With `BLAS32_jll.jl`, we will be able to compare the performances of the reference BLAS with OpenBLAS, Intel MKL, BLIS or AppleAccelerate for packages compiled with LBT.

### DIFF
--- a/B/BLAS/BLAS32/build_tarballs.jl
+++ b/B/BLAS/BLAS32/build_tarballs.jl
@@ -1,0 +1,12 @@
+name = "BLAS32"
+
+include("../common.jl")
+
+script = blas_script(blas32=true)
+
+products = [
+    LibraryProduct("libblas", :libblas),
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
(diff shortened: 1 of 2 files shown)


ViralBShah (approved):

ViralBShah:
How about just calling this package `ReferenceBLAS` and having both the 32-bit and 64-bit versions in there??

amontoison:
I prefer `BLAS` than `ReferenceBLAS` but having the two libraries is a good idea.
The issue is that we already have `LAPACK32_jll.jl` / `LAPACK_jll.jl`, `OpenBLAS32_jll` / `OpenBLAS_jll` and `blis32_jll` / `blis_jll`.

ViralBShah:
Well, this all sounded a bit familiar. So we already have it: https://github.com/JuliaPackaging/Yggdrasil/pull/7167

amontoison:
Why did you use `ReferenceBLAS` and not `ReferenceLPACK`?
I was not aware that you already precompiled `BLAS`!

ViralBShah:
I don't have a good explanation 😕 I just feel like BLAS is a little bit more general with multiple implementations whereas there is usually only one LAPACK.

amontoison:
I think that the best thing to do is to update `ReferenceBLAS` and modernize its `build_tarballs.jl` with the future release 0.3.12.

PR #7698, opened 2023-11-19
Update liblinear

ericphanson:
So it builds for m1, but can also bump the version at the same time

### DIFF
--- a/L/liblinear/build_tarballs.jl
+++ b/L/liblinear/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "liblinear"
-version = v"2.30.0"
+version = v"2.47.0"
 
 # Collection of sources required to complete build
 sources = [
--- a/L/liblinear/build_tarballs.jl
+++ b/L/liblinear/build_tarballs.jl
@@ -7,7 +7,7 @@ version = v"2.30.0"
 
 # Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://github.com/cjlin1/liblinear/archive/v230.tar.gz", "9b57710078206d4dbbe75e9015d4cf7fabe4464013fe0e89b8a2fe40038f8f51"),
+    GitSource("https://github.com/cjlin1/liblinear", "8dc206b782e07676dc0d00678bedd295ce85acf3"),
     DirectorySource("./bundled")
 ]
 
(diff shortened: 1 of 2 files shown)


ericphanson:
The patch seems to be updating the extensions in the Makefile, so I've tried to make the same adjustments to the [v247 makefile](https://github.com/cjlin1/liblinear/blob/v247/Makefile) in the updated patch here.

L/liblinear/build_tarballs.jl
```diff
@@ -17,7 +17,7 @@ cd $WORKSPACE/srcdir
 for f in ${WORKSPACE}/srcdir/patches/*.patch; do
     atomic_patch -p1 ${f}
 done
-cd liblinear-230/
+cd liblinear/
 mkdir -p ${prefix}/bin
 mkdir -p ${prefix}/lib
 if [[ "${target}" == *-freebsd* ]] || [[ "${target}" == *-apple-* ]]; then
```
giordano, line 23: Uhm, I don't remember why were using gcc, it'd be good to see if we can avoid that

L/liblinear/build_tarballs.jl
```diff
@@ -17,7 +17,7 @@ cd $WORKSPACE/srcdir
 for f in ${WORKSPACE}/srcdir/patches/*.patch; do
     atomic_patch -p1 ${f}
 done
-cd liblinear-230/
+cd liblinear/
 mkdir -p ${prefix}/bin
```
giordano, line 21:
```suggestion
mkdir -p "${bindir}"
```

L/liblinear/build_tarballs.jl
```diff
@@ -17,7 +17,7 @@ cd $WORKSPACE/srcdir
...
 mkdir -p ${prefix}/lib
```
giordano, line 22:
```suggestion
mkdir -p "${libdir}"
```

L/liblinear/build_tarballs.jl
ericphanson: should I try just deleting this `if` block?

ericphanson: trying that in https://github.com/JuliaPackaging/Yggdrasil/pull/7698/commits/724addf649f26a9cea343551d1e0b83c0b17e0b8

ericphanson:
success!

PR #7699, opened 2023-11-20
Update open62541 to v1.3.8

thomvet:


giordano:
As a general remark working with Git and GitHub, opening a PR from a branch with the same name as the target branch is an [anti-pattern](https://blog.jasonmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/). You should always create a ***new*** branch.

thomvet:
Noted. Will change my pattern for future PRs and create them from a patch branch instead of master.

PR #7700, opened 2023-11-20
Update `quarto` version to current latest (`v1.3.450`)

MichaelHatherly:


MichaelHatherly:
Ah, some hashes on those other files have changed as well, one moment.

PR #7701, opened 2023-11-20
Bump XML2 to 2.12.0

bluesmoon:


giordano:
Opened an issue for failed compilation on i686-w64-mingw32: https://gitlab.gnome.org/GNOME/libxml2/-/issues/625

bluesmoon:
I'll open a separate MR for 2.11.6 which should also fix the sec vuln in 2.11.5.

PR #7717, opened 2023-11-26
Upgrade RootFS

amontoison:
It should fix some errors when we compile Fortran packages.

Companion PR in BBB: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/353

### DIFF
--- a/0_RootFS/Rootfs/bundled/libs/libc/download_libcs.sh
+++ b/0_RootFS/Rootfs/bundled/libs/libc/download_libcs.sh
@@ -12,5 +12,5 @@ done
 for arch in i686 x86_64; do
     rm -rf musl-${arch}
     mkdir -p musl-${arch}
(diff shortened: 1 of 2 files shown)


0_RootFS/Rootfs/bundled/libs/libc/download_libcs.sh
```diff
@@ -12,5 +12,5 @@ done
 for arch in i686 x86_64; do
     rm -rf musl-${arch}
     mkdir -p musl-${arch}
-    curl -L "https://github.com/JuliaBinaryWrappers/Musl_jll.jl/releases/download/Musl-v1.2.2%2B2/Musl.v1.2.2.${arch}-linux-musl.tar.gz" | tar --wildcards --strip-components=1 -zxv -C musl-${arch} "lib*/"
+    curl -L "https://github.com/JuliaBinaryWrappers/Musl_jll.jl/releases/download/Musl-v1.2.4%2B0/Musl.v1.2.4.${arch}-linux-musl.tar.gz" | tar --wildcards --strip-components=1 -zxv -C musl-${arch} "lib*/"
```
giordano, line 15: Don't update this if the base Alpine image is the same.

PR #7719, opened 2023-11-26
[libcxxwrap-julia] Version 0.11.2

barche:
Mainly releasing now for compatibility fixes with Julia master.

CC @fingolfin @benlorenz


fingolfin:
Thanks!

PR #6740, opened 2023-05-11
Fix GDAL and bump version of DGGRID7

danlooo:
This PR aims to fix a version mismatch of DGGRID7_jll and GDAL_jll by updating DGGRID7_jll to the latest official release 7.7 of [DGGRID](https://github.com/sahrk/DGGRID/) that works with BinaryBuilder. @allixender what do you think?

## Error

Executing in docker image julia:1.8.5-bullseye :

```{julia}
using Pkg
Pkg.add("DGGRID7_jll")

using DGGRID7_jll

download("https://raw.githubusercontent.com/sahrk/DGGRID/4f24e9de46f9e669af4cba94d9d148b557a1de11/examples/icosahedron/icosahedron.meta", "meta.txt")
mkdir("outputfiles")

DGGRID7_jll.dggrid() do dggrid_path
    run(`$dggrid_path meta.txt`)
end
```

results in error

```
/root/.julia/artifacts/3381b9f883af7e984906191545c4fa939e07e91b/bin/dggrid: error while loading shared libraries: libgdal.so.28: cannot open shared object file: No such file or directory
```

Indeed, dggrid requires libgdal.so.28, but package GDAL_jll provides libgdal.so.32.3.6.2:

```
root@6e8b0f893490:~/.julia/artifacts# find ~/.julia/artifacts/6f97fbb4bedf23ccd1d7757586cb7659222a0465/lib
/root/.julia/artifacts/6f97fbb4bedf23ccd1d7757586cb7659222a0465/lib
/root/.julia/artifacts/6f97fbb4bedf23ccd1d7757586cb7659222a0465/lib/pkgconfig
/root/.julia/artifacts/6f97fbb4bedf23ccd1d7757586cb7659222a0465/lib/pkgconfig/gdal.pc
/root/.julia/artifacts/6f97fbb4bedf23ccd1d7757586cb7659222a0465/lib/libgdal.so.32
/root/.julia/artifacts/6f97fbb4bedf23ccd1d7757586cb7659222a0465/lib/gdalplugins
/root/.julia/artifacts/6f97fbb4bedf23ccd1d7757586cb7659222a0465/lib/gdalplugins/drivers.ini
/root/.julia/artifacts/6f97fbb4bedf23ccd1d7757586cb7659222a0465/lib/libgdal.so
/root/.julia/artifacts/6f97fbb4bedf23ccd1d7757586cb7659222a0465/lib/cmake
[...]

PR #7721, opened 2023-11-27
Fix version of GDAL_jll

danlooo:
`GDAL_jll` has been updated to v301.700.200+0 that produces the library file `libgdal.so.33`.
However, `DGGRID7_jlll` requires  `libgdal.so.32`:

```
julia> DGGRID7_jll.dggrid() do x run(`$x /dev/null`) end
/root/.julia/artifacts/f4aac71056880a075dd62fa9b9a65d44c6de441e/bin/dggrid: error while loading shared libraries: libgdal.so.32: cannot open shared object file: No such file or directory
ERROR: failed process: Process(`/root/.julia/artifacts/f4aac71056880a075dd62fa9b9a65d44c6de441e/bin/dggrid /dev/null`, ProcessExited(127)) [127]
```

This PR fixes the compat of GDAL that has non-standard version numbers.
See PR #6740 for previous work.

### DIFF
--- a/D/DGGRID7/build_tarballs.jl
+++ b/D/DGGRID7/build_tarballs.jl
@@ -36,7 +36,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="GDAL_jll", uuid="a7073274-a066-55f0-b90d-d619367d196c"); compat="v301.600.200")
+    Dependency(PackageSpec(name="GDAL_jll", uuid="a7073274-a066-55f0-b90d-d619367d196c"); compat="=v301.600.200")
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.


giordano:
This stuff is fixed by fixing the compat bound in the registry

danlooo:
@giordano thanks for the info! Do you mean I should do a PR for https://github.com/JuliaRegistries/General/blob/master/jll/D/DGGRID7_jll/Compat.toml instead?

giordano:
More than "instead", it's "before opening a PR here".  Also, I think you want to use `~` rather that `=`, builds within the same minor version should be compatible, right?

danlooo:
Ok, all registry checks passed, see https://github.com/JuliaRegistries/General/pull/96005

D/DGGRID7/build_tarballs.jl
```diff
@@ -36,7 +36,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="GDAL_jll", uuid="a7073274-a066-55f0-b90d-d619367d196c"); compat="v301.600.200")
+    Dependency(PackageSpec(name="GDAL_jll", uuid="a7073274-a066-55f0-b90d-d619367d196c"); compat="=v301.600.200")
```
giordano, line 39: For consistency with https://github.com/JuliaRegistries/General/pull/96005#discussion_r1406282649
```suggestion
    Dependency(PackageSpec(name="GDAL_jll", uuid="a7073274-a066-55f0-b90d-d619367d196c"); compat="~v301.600.200")
```

PR #7728, opened 2023-11-28
Update Armadillo, ensmallen, and mlpack to latest stable versions.

rcurtin:
mlpack -> 4.3.0
ensmallen -> 2.21.0
Armadillo -> 12.6.6

Hopefully everything will build correctly... let's find out!

### DIFF
--- a/A/armadillo/build_tarballs.jl
+++ b/A/armadillo/build_tarballs.jl
@@ -5,10 +5,10 @@ using BinaryBuilder
 
 
 name = "armadillo"
-version = v"12.2.0"
+version = v"12.6.6"
 sources = [
     ArchiveSource("http://sourceforge.net/projects/arma/files/armadillo-$(version).tar.xz",
-                  "b0dce042297e865add3351dad77f78c2c7638d6632f58357b015e50edcbd2186")
+                  "3858b0fd431772af032ad3f35c2aeb54e8dabea59169e7d1e9fccd78bc82ad35")
 ]
 
 script = raw"""
--- a/A/armadillo/build_tarballs.jl
+++ b/A/armadillo/build_tarballs.jl
@@ -78,4 +78,4 @@ dependencies = [
     Dependency("OpenBLAS_jll")
 ]
 
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"6", julia_compat="1.7")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"6", julia_compat="1.8")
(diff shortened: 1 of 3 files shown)


rcurtin:
```
[ Info: lib/libmlpack_julia_linear_svm.so already has SONAME "libmlpack_julia_linear_svm.so"
[ Info: Checking shared library lib/libmlpack_julia_lmnn.so
ERROR: could not load library "/cache/build/yggy-amdci7-10/julialang/yggdrasil/M/mlpack/build/x86_64-linux-gnu-cxx11/m8fTdi5x/x86_64-linux-gnu-libgfortran4-cxx11/destdir/lib/libmlpack_julia_lmnn.so"
libmlpack_julia_util.so: ELF load command address/offset not properly aligned
┌ Warning: lib/libmlpack_julia_lmnn.so cannot be dlopen()'ed
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/XPzgk/src/Auditor.jl:171
```

Is it possible BinaryBuilder or some other piece is stripping these incorrectly?  I'll try and dig a little bit but without deep knowledge of the internals of the Yggdrasil or Julia binary build system, I may not get very far.

giordano:
Uhm, no, we don't strip the binaries, but we do run patchelf on them

rcurtin:
I spent a while trying to reproduce locally but had no luck, even using the same Julia version as buildkite appears to be using.  As far as I can tell, the .so is valid before patchelf/Auditor.jl does its thing, and the best I can conclude without any local reproduction is that something patchelf is doing is causing a misalignment.  However, we are wandering into tooling I am pretty unfamiliar with, so I am not quite sure what to do to debug this and would really appreciate some help.

At this point I am just throwing spaghetti at the wall to see what sticks.  (Maybe a Julia version bump will help?)

giordano:
Can you try to revert https://github.com/JuliaPackaging/Yggdrasil/commit/b23cf128e248aaf1c3f4f513f65a997de29f3c11 and see if that changes anything? I won't have much time to debug this this week

giordano:
@haampie have you seen errors like https://github.com/JuliaPackaging/Yggdrasil/pull/7728#issuecomment-1830467643 with recent versions of patchelf?

rcurtin:
Sure, let's see what happens.  Also, for sure no hurry on the PR. :+1:

giordano:
It appears the upgrade of patchelf was the culprit, sigh. On, we'll have to bring that one back.

rcurtin:
Ah, ok, well, glad you were able to figure it out for me (thank you! :smile:).  Let me know if you want that revert handled in a different PR or something, happy to handle it. :+1:

rcurtin:
Thanks! :rocket:

PR #7741, opened 2023-12-01
[GR] bump version

jheinen:
bump GR version to 0.72.11

### DIFF
--- a/G/GR/build_tarballs.jl
+++ b/G/GR/build_tarballs.jl
@@ -3,13 +3,13 @@
 using BinaryBuilder
 
 name = "GR"
-version = v"0.72.10"
+version = v"0.72.11"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/sciapp/gr.git", "ffedfe702fdc3872cb3172fddb1dfb8f0d8986c5"),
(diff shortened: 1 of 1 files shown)


jheinen:
@giordano : How can I specify, that `grplot` should **not** be built for MinGW 32-Bit platforms?
```
ExecutableProduct("grplot", :grplot),
```
Due to a limitation regarding C++ exception handling, we cannot build/provide `grplot` on these platforms.

giordano:
In principle we could have platform-specific products, as we do for dependencies, but platform-specific products are a bit problematic because the consumer will have to check whether the desired product exists or not. Instead we don't have this concern for dependencies of the jll, those aren't typically directly accessible by consumers.

If after this caveat you still want to go ahead, you should basically do two build_tarballs calls with different products and platforms arguments, something like in https://github.com/JuliaPackaging/Yggdrasil/blob/89647232e453190464fb449f7fde96b5cf3e5fd0/G/GlibNetworking/build_tarballs.jl#L43-L73

jheinen:
> If after this caveat you still want to go ahead, you should basically do two build_tarballs calls with different products and platforms arguments, something like in

I think, we will fix the problem on "our" side and provide a dummy `grplot` binary for MinGW 32-Bit systems. Since the `grplot` software component is not needed in Julia anyway, this should not be a problem.

ViralBShah:
I think this can be closed since we have newer versions published.

PR #7743, opened 2023-12-01
New Recipe: GTPSA v1.0.0

mattsignorelli:
This recipe generates binaries for the gtpsa shared library wrapped by the GTPSA.jl package (https://github.com/bmad-sim/GTPSA.jl) presently in development.

### DIFF
--- a/G/GTPSA/build_tarballs.jl
+++ b/G/GTPSA/build_tarballs.jl
@@ -14,8 +14,15 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir
 cd gtpsa/
-if [[ $target == *"-apple-"* ]]; then CC=gcc; CXX=g++; fi
-if [[ $target == *"-apple-"* ]]; then ln /workspace/destdir/lib/liblapack32.dylib /workspace/destdir/lib/liblapack.dylib; elif [[ $target == *"linux"* ]]; then ln /workspace/destdir/lib/liblapack32.so /workspace/destdir/lib/liblapack.so; fi
+if [[ $target == *"-apple-"* ]]
+then
+CC=gcc
+CXX=g++
+ln /workspace/destdir/lib/liblapack32.dylib /workspace/destdir/lib/liblapack.dylib
+elif [[ $target == *"linux"* ]]
+then
+ln /workspace/destdir/lib/liblapack32.so /workspace/destdir/lib/liblapack.so
+fi
 cmake . -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake -DCMAKE_BUILD_TYPE=Release
 make
 make install
--- a/G/GTPSA/build_tarballs.jl
+++ b/G/GTPSA/build_tarballs.jl
@@ -26,7 +26,6 @@ fi
 cmake . -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake -DCMAKE_BUILD_TYPE=Release
 make
 make install
-exit
 """
 
 # These are the platforms we will build for by default, unless further
--- a/G/GTPSA/build_tarballs.jl
+++ b/G/GTPSA/build_tarballs.jl
@@ -30,23 +30,9 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = [
-    Platform("i686", "linux"; libc = "glibc"),
-    Platform("x86_64", "linux"; libc = "glibc"),
-    Platform("aarch64", "linux"; libc = "glibc"),
-    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
-    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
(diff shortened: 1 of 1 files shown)


G/GTPSA/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="ReferenceBLAS_jll", uuid="ee697234-451c-51c9-b102-303d89a9c3a0"))
+    Dependency(PackageSpec(name="LAPACK32_jll", uuid="17f450c3-bd24-55df-bb84-8c51b4b939e3"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"11.1.0")
```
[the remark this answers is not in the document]
mattsignorelli: The source code uses the two argument version of the attribute malloc, which requires GCC 11

G/GTPSA/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+sources = [
+    GitSource("https://github.com/mattsignorelli/gtpsa.git", "ed2c14eb096a9ecac3f483f9914c0c679765496d")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd gtpsa/
+if [[ $target == *"-apple-"* ]]; then CC=gcc; CXX=g++; fi
+if [[ $target == *"-apple-"* ]]; then ln /workspace/destdir/lib/liblapack32.dylib /workspace/destdir/lib/liblapack.dylib; elif [[ $target == *"linux"* ]]; then ln /workspace/destdir/lib/liblapack32.so /workspace/destdir/lib/liblapack.so; fi
```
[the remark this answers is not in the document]
mattsignorelli: Fixed!

G/GTPSA/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("i686", "windows"; ),
+    Platform("x86_64", "windows"; )
+]
```
imciner2, line 50: It looks like you are basically building for all platforms here, right? If so, just condense it down into this statement.
```suggestion
platforms = supported_platforms()
```

G/GTPSA/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+cd $WORKSPACE/srcdir
+cd gtpsa/
+if [[ $target == *"-apple-"* ]]
+then
+CC=gcc
+CXX=g++
+ln /workspace/destdir/lib/liblapack32.dylib /workspace/destdir/lib/liblapack.dylib
+elif [[ $target == *"linux"* ]]
+then
+ln /workspace/destdir/lib/liblapack32.so /workspace/destdir/lib/liblapack.so
```
imciner2, line 24: While I guess this is fine, I believe an alternate way would be to add
```
set( BLA_SIZEOF_INTEGER 4 )
```
to the upstream CMake before your `find_package( LAPACK )` call, and then ensure that the most recent CMake is installed in the build root here. You can do that by adding
```
# we need a more recent cmake version for suitesparse and use the jll instead
apk del cmake
```
to the start of the build script, and then adding a new dependency for
```
HostBuildDependency(PackageSpec(; name="CMake_jll", version = v"3.24.3"))
```
(or another CMake version greater than 3.22).

G/GTPSA/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+fi
+cmake . -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake -DCMAKE_BUILD_TYPE=Release
+make
+make install
+exit
```
imciner2, line 29:
```suggestion
```
Not needed in the build script.

G/GTPSA/build_tarballs.jl
mattsignorelli: I tried this, and ensured the CMake version was > 3.22, and it was giving an error in the linking stage of make because it could not find -llapack, even though CMake was able to find it. So I believe the link with the expected liblapack name is necessary

G/GTPSA/build_tarballs.jl
mattsignorelli: FreeBSD is not included in that list - the source code is unable to compile on this platform with the error: ‘__builtin_tgmath’ is not a function pointer . I have had difficulty resolving this and decided, given the purpose/audience/use cases of this package, and compatibility of packages in the same realm as this one, that FreeBSD support is not necessary.

imciner2: In that case, you can use the following code:
```julia
# The code does not compile on FreeBSD due to an error with __builtin_tgmath
```

imciner2 (approved):

imciner2:
Looks good now. Thanks for your work on this.

PR #7757, opened 2023-12-05
[libjulia] bump nightly for jl_alloc_array_*, update to 1.10.0-rc2

benlorenz:


giordano:
Any clue of why that wasn't caught by CI?

fingolfin:
This builds now with the crc32 patch, great!. Since that patch doesn't affect the ABI, can't we just merge this PR now so we can finally update all the other JLLs depending on this? We can drop the crc32 later once it (or some other fix) is merged into Julia master.

PR #7775, opened 2023-12-06
t8code: New version 1.6.1

jmark:
t8code just released a new version. This PR updates the associated JLL package.


T/t8code/build_tarballs.jl
```diff
@@ -6,24 +6,23 @@ const YGGDRASIL_DIR = "../.."
...
-cd $WORKSPACE/srcdir/t8code
+script = """
+cd \$WORKSPACE/srcdir/t8-$(version)
+""" * raw"""
 atomic_patch -p1 "${WORKSPACE}/srcdir/patches/mpi-constants.patch"
 
 # Set default preprocessor and linker flags
 # Note: This is *crucial* for Windows builds as otherwise the wrong libraries are picked up!
-export CPPFLAGS="-I${includedir}"
+export CPPFLAGS="-I${includedir} -I${includedir}"
```
imciner2, line 25:
```suggestion
export CPPFLAGS="-I${includedir}"
```
You shouldn't need the include directory listed twice.

imciner2 (changes requested):

jmark:
@imciner2 Is it possible to restart the failed test. It seems that is just stalled. I manually canceled the test.

T/t8code/build_tarballs.jl
```diff
@@ -6,19 +6,18 @@ const YGGDRASIL_DIR = "../.."
...
-]
+# Collection of sources required to complete build
+sources = [ArchiveSource(tarball, sha256sum), DirectorySource("./bundled")]
 
 # Bash recipe for building across all platforms
-script = raw"""
-cd $WORKSPACE/srcdir/t8code
+script = """
+cd \$WORKSPACE/srcdir/t8-$(version)
+""" * raw"""
```
giordano, line 20: We usually do something like
```suggestion
script = raw"""
cd $WORKSPACE/srcdir/t8*
```
which would work with any version

imciner2 (approved):

PR #7780, opened 2023-12-07
[MKL] Require same version of IntelOpenMP

giordano:
MKL should really go hand-in-hand with Intel OpenMP.  See for example https://github.com/JuliaMath/FFTW.jl/issues/281.

Opening as draft because this should be fixed in the registry first, but don't have the time to do that right now.  ***Edit***: done in https://github.com/JuliaRegistries/General/pull/96745.

### DIFF
--- a/M/MKL/build_tarballs.jl
+++ b/M/MKL/build_tarballs.jl
@@ -43,7 +43,11 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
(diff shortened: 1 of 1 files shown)


amontoison:
Mosè, I'm not sure that they don't break things between releases of the same year like `2023.0.0 and `2023.2.0`.
The release number is the just the year / semester of the release for them. We already encountered this issue with libraries for GPUs.
It could be safe to add to do `compat="="*string(version)`.

giordano:
If that becomes necessary, yes, we may want to do that.  But also, in the past we didn't always have the same versions of IntelOpenMP and MKL (actually never until 2023)

PR #7785, opened 2023-12-09
Update Spglib's build_tarballs.jl to v2.2.0

singularitti:


S/spglib/build_tarballs.jl
```diff
@@ -19,7 +19,6 @@ if [[ ! -z "${CMAKE_TARGET_TOOLCHAIN}" ]]; then
 fi
 cmake -B ./build \
       -DCMAKE_INSTALL_PREFIX=${prefix} \
-      -DSPGLIB_WITH_TESTS=OFF \
```
imciner2, line 22: Why did you remove this line? The tests really don't serve any purpose when building in Yggdrasil because this is all cross compiled, so the test executables can't even be run here.

singularitti: OK, from the upstream, I heard the Julia packaging issue is [fixed](https://github.com/spglib/spglib/pull/341). So I want to see if I can run these tests now.

imciner2 (approved):

S/spglib/build_tarballs.jl
imciner2: Running the tests will always be a problem I believe. The only way you could do that is have a virtual machine system (like Qemu) that they run in. But that is getting way more complicated than we really want to do here in Yggdrasil.

Usually, the way the JLLs are tested is just through the tests in the wrapper .jl package (if there is one).

PR #7787, opened 2023-12-09
Add a recipe for GALAHAD

amontoison:


G/GALAHAD/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+    Dependency(PackageSpec(name="METIS_jll", uuid="d00139f3-1899-568f-a2f0-47f597d42d70")),
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
+    Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"), compat="5.4.0"),
+    Dependency(PackageSpec(name="Hwloc_jll", uuid="e33a78d0-f292-5ffc-b300-72abe9b543c8")),
+    Dependency(PackageSpec(name="MUMPS_seq_jll", uuid="d7ed1dd3-d0ae-5e8e-bfb4-87a502085b8d")),
+    # Dependency(PackageSpec(name="PaStiX_jll", uuid="46e5285b-ff06-5712-adf2-cc145d39f096")),
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"8.1.0", julia_compat="1.9")
```
giordano, line 56: libgfortran3 corresponds to gcc 4-6: https://github.com/JuliaPackaging/Yggdrasil/blob/ef3bbebbc8f13a326e5a49e960a09054f3573dad/RootFS.md#compiler-shards.  Asking gcc 8 for fixing problems with libgfortran3 doesn't make sense, at most you'd need 6 for that (of course gcc 8 implies gcc 6 for libgfortran3)

amontoison: I asked GCC 8 to use the oldest g++ compiler with `libgfortran5` and the most recent one with `libgfortran3`.
It makes sense because I know that it will use GCC 6 :)

amontoison: I also didn't remember what was the most recent GCC 6.x.x compiler.

PR #7798, opened 2023-12-13
Recipe to make WrapIt_jll package

grasph:
Recipe to make WrapIt_jll package providing the [wrapit](https://github.com/grasph/wrapit) binaries. WrapIt is an application written in C++ to generate automatically from the headers the code to make Julia binding using CxxWrap.

The _jll package ships an executable that uses LLVM and clang libraries. As explained in [this discourse  thread](), we don't want the version of the used libraries to depend on the julia version used to install the _jll.

The version decorrelation is achieved by adding a dependency directly on the artifacts of Clang_jll and libLLVM_jll of the desired version (v. 16) instead of depending on the _jll packages.

Two points could be improved in the recipe:

  * The libclang and libLLVM shared library are copied in our artifact, while a symbolic to the files provided by the Clang_jll and libLLVM_jll artifacts would have been preferred; 
  * The produces _jll misses the Artificats we depend on in its Artifact.toml file. We need to put them in the julia helper module that will use this _jll.
 
I've not found how to include the symbolic links to the regular files in the generated tarball, neither how to customize the Artifacts.toml (probably not possible).

Recommendation to improve these two points is welcome or on the general approach to achieve the goal is welcome. 

The points are not blocking as in practice the recipe is working.

Philippe.


W/WrapIt/build_tarballs.jl
```diff
@@ -0,0 +1,111 @@
...
+    rm -rf /opt/${target}/${target}/sys-root/System
+    cp -ra usr/* "/opt/${target}/${target}/sys-root/usr/."
+    cp -ra System "/opt/${target}/${target}/sys-root/."
+    export MACOSX_DEPLOYMENT_TARGET=10.15
+    popd
+fi
+
+######################################################################
+# Build the software
+clang_resource_dir=clang/res
```
giordano, line 31: I think these directory is a broken symlink in the resulting tarball.

giordano (commented):
To be honest I'm not a big fan of vendoring libllvm and clang in this package, but since the only product is an executable, rather than a shared library that would need to be loaded in the same session as another libllvm, this may be a _not too awful_ solution and at least avoids the combinatorial explosion of llvm versions.

W/WrapIt/build_tarballs.jl
grasph: It is intended. The link points to a directory from the Clang@16 artifact that needs to be installed alongside. This _jll package is expected to be used through the [WrapIt](https://github.com/grasph/WrapIt.jl) package (whose version number will match with the _jll one). 

I would have preferred to put the dependency in the Artifacts.toml of the _jll package, but  I have not found how to make it happens.

giordano: > The link points to a directory from the Clang@16 artifact that needs to be installed alongside.

Uhm, that's not how this works.  Artifacts can be installed in different depots, we can't guarantee at all that the desired Clang_jll package will be where you expect it, actually, that's also a wrong expectation which since Clang_jll isn't even a runtime dependency.

grasph: I hadn't consider the case of multiple depot. 

So I we to copy the files instead of making link. The files to copy represent an overhead on the artifact size of 2.8% only (*). Therefore, as we already copy the libraries which are the biggest pieces, we don't save much disk space with this link and a copy is fine. 

(*) computed for x86_64 with Linux, glibc, and cxx11 string ABI.

grasph:
Hello @giordano , can we proceed with the merging ? Let me know if I need to make any other modification.

Philippe.

giordano:
I think this could be improved, but ok, let's go with it.

grasph:
Thanks Mosè. It's nice to have a working version in. Write me if you have suggestions for improvements, keeping in mind I need llvm/clang version choice independent of the Julia version.

giordano:
Sounds like [registration job failed](https://buildkite.com/julialang/yggdrasil/builds/8260) (my fault, I should have restarted CI before merging), hardcoding llvm v16 wasn't a great idea after all.

grasph:
> Sounds like [registration job failed](https://buildkite.com/julialang/yggdrasil/builds/8260) (my fault, I should have restarted CI before merging), hardcoding llvm v16 wasn't a great idea after all.

see  PR #8125 for the fix.

PR #7801, opened 2023-12-14
Wizard recipe: OpenFOAM_com-v2312

thealanjason:

* Package name: OpenFOAM_com
* Version: v2312

### DIFF
--- a/O/OpenFOAM_com/build_tarballs.jl
+++ b/O/OpenFOAM_com/build_tarballs.jl
@@ -214,6 +214,7 @@ dependencies = [
     Dependency(PackageSpec(name="METIS_jll", uuid="d00139f3-1899-568f-a2f0-47f597d42d70"))
     Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a"))
 ]
+append!(dependencies, platform_dependencies)
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; 
--- a/O/OpenFOAM_com/build_tarballs.jl
+++ b/O/OpenFOAM_com/build_tarballs.jl
@@ -1,0 +1,188 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "OpenFOAM_com"
+version = v"2306.0.0"
+
+# Collection of sources required to complete build
+sources = [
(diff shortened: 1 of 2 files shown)


imciner2:
What is the difference between this version of OpenFOAM and the version we already have as a JLL here https://github.com/JuliaPackaging/Yggdrasil/tree/master/O/OpenFOAM?

thealanjason:
Hello @imciner2 thanks for looking into this pull request, even though it does not pass all the pipelines. 
This version is compiled from the source at openfoam.com (ESI), and the the version for which we already have a JLL is built from the source at openfoam.org. Although they sound the same, they diverged from the same source some time ago. I am particularly interested in the openfoam.com version, because it has an additional module, avalanche, for granular shallow flow simulation, which I would like to use in a benchmark package for my research. Furthermore, keeping reproducibility of my package in mind, I would like to build this JLL for additional platforms, MacOS amd64 and aarch64, Windows amd64 and Linux amd64 and aarch64. 

The build_tarballs.jl for the already OpenFOAM version that you linked really helped in getting started. However, the advanced copying and moving of the lib, bin, folder fails in the in the buildkite pipeline. I'm not sure why. Is there a support channel where I can get help on slack/discourse?

ViralBShah:
It does make sense to have this version. The naming is rather unfortunate. It would help if the projects had more differentiated names. Any suggestions for an improved name for the `.com` version? Is this what the linux distros do?

thealanjason:
@ViralBShah I find the naming unfortunate as well. I'm not sure what linux distros do and chose the "_com" arbitrarily.

thealanjason:
For the naming, in Ubuntu, the package for the openfoam.com version is named as openfoam in the PPA (see [here](https://packages.ubuntu.com/search?keywords=openfoam)). The openfoam.org version, for which a build script already exists in Yggdrasil, does not exist on the ubuntu PPA and the repository has to be added separately [see here](https://openfoam.org/download/11-ubuntu/). In the installation instruction the package is then refered to as `openfoam11`, i.e. openfoam + version number (see corresponding binary names [here](http://dl.openfoam.org/ubuntu/dists/jammy/main/binary-amd64/)). 

I think it would be inefficient to include the package version in the name of the JLL package. One way to handle this: since both releases have different version numbering, they would not clash for now, and we could keep the same name of the package, i.e. OpenFOAM_jll, and have different non-clashing version numbers to indicate the two different packages. However, this could be confusing for the user. 

What should be the way forward? @ViralBShah @giordano 

Also, I'll be closing this pull request in favour of https://github.com/JuliaPackaging/Yggdrasil/pull/7976, because a new version 2312 was released after I started with this recipe.

ViralBShah:
I have given it some thought and couldn't come up with something better. OpenFOAM_com is perhaps the best option to go forward with.

PR #7803, opened 2023-12-14
[SuiteSparse_GPU] Update products to only contain GPU libraries

imciner2:
The registration for the v7 SuiteSparse GPU JLL failed because some SuiteSparse libraries (such as AMD) were declared as products in the recipe, but we weren't actually building them in the GPU JLL because we were going to be using the CPU-built ones in the normal JLL (since there isn't a GPU-specific version of those libraries.

Log file of the failure: https://buildkite.com/julialang/yggdrasil/builds/6438#018baa65-8284-4658-b64a-39baf25356f7, error:
```Julia
ERROR: LoadError: Unable to locate LibraryProduct(["libamd"], :libamd, String[], false, Symbol[]) within /tmp/jl_C9Wk2N for x86_64-linux-gnu-cuda+11.4
```

Instead, we need to declare only the GPU-accelerated libraries in the GPU recipe.

### DIFF
--- a/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
+++ b/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
@@ -4,8 +4,6 @@ name = "SuiteSparse"
 version = v"7.4.0"
 
 sources = suitesparse_sources(version)
-
-sources = suitesparse_sources(version)
 push!(sources, DirectorySource("./bundled"))
 
 # Bash recipe for building across all platforms
--- a/S/SuiteSparse/SuiteSparse_GPU@7/build_tarballs.jl
+++ b/S/SuiteSparse/SuiteSparse_GPU@7/build_tarballs.jl
@@ -99,10 +99,6 @@ install_license LICENSE.txt
 platforms = CUDA.supported_platforms()
 filter!(p -> arch(p) == "x86_64", platforms)
 
-# Add products
-push!(products, LibraryProduct("libgpuqrengine", :libgpuqrengine))
-push!(products, LibraryProduct("libsuitesparse_gpuruntime", :libsuitesparse_gpuruntime))
-
 # Add dependency on SuiteSparse_jll
 push!(dependencies, Dependency("SuiteSparse_jll"))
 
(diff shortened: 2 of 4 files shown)


imciner2:
The 7.3.0 version we have doesn't play nicely with our CMake right now because it wants the `cuda::NvToolsExt` target, but that has been replaced with a `cuda::nvtx3` target. This has been fixed in what will become the upstream 7.4.0 release, and there are quite a few major changes to the build system between 7.3.0 and 7.4.0 that makes it impossible to backport things to our 7.3.0 build easily. Since 7.4.0 has some beta releases right now, I am just going to hold off on finishing this until 7.4.0 is released and we rebase onto 7.4.0.

(note 7.4.0 also seems to simplify how we can do the CMake builds, so we shouldn't need the for loop over targets and instead just pass a list of targets to a root CMake instead).

ViralBShah:
@imciner2 Would you be able to update this to 7.4? Would be nice to get it merged.

imciner2:
Yes, this is on my list to revisit later this week. I believe it is close to working and that I just need to rebase it.

imciner2:
Ok, this is now rebased to 7.4 fully, and it builds. I haven't tried any tests on it though, so I can't say if it actually works.

That said, apparently SuiteSparse 7.5 will be released today, so since this also touched the actual SuiteSparse build script to cleanup some commonalities, we could do that update in here and not have an extra build of the CPU libraries.

ViralBShah:
Yeah, makes sense to wait for 7.5. The first step to making all this work on the GPU is to have a working build. I hope to test it out once it gets merged.

imciner2:
Ok, I put up the 7.5 update in a separate PR. Unfortuently, I don't think we can properly bump the GPU and non-GPU libraries in the same PR due to dependency issues. The GPU libraries will use the AMD compiled in the normal SuiteSparse_jll, and so they have a specific version dependency apparently (e.g. I just tested building the GPU version of 7.5.0 against the 7.4.0 jll in the general registry right now, and 7.5.0 wants AMD 3.3.1 minimum, whereas 7.4.0 has 3.3.0).

I think we might also need to add a strict version requirement to the SuiteSparse dependency in this package also - to pin the SuiteSparse and SuiteSparse_GPU packages to be the exact same SuiteSparse version.

rayegun:
They do need to be the exact same. SPQR_GPU and CHOLMOD_GPU are technically just part of their respective parent packages.

Same with SPQR <=> CHOLMOD. I would not trust trying to separate them even if it's possible for some versions by accident

imciner2:
> They do need to be the exact same. SPQR_GPU and CHOLMOD_GPU are technically just part of their respective parent packages.

Great, thanks for confirming. I've added a strict version requirement on the updated PR (https://github.com/JuliaPackaging/Yggdrasil/pull/7932/files#diff-331b2e16e9da7ac54597cf58b1baae5feedef23d91ce85981be8fe645a15474cR99), so that should force the same version when installed.

Issue #7139, opened 2023-07-31
C++17 not fully supported on MacOS: C++ shared pointers to arrays don't compile 


Downsite:
I am currently failing to build C++ binaries for a project using features from C++17 on MacOS.
One example is the usage of C++ shared pointers to arrays.

Downloading a newer SDK does not seem to help. The issue has been discussed before (#5851), but there the issue was not fixed generally, but for the specific project by forking & modifying upstream code.

Is there a known workaround that does not require changing upstream code?
BinaryBuilder can build the code for MacOS using the GCC toolchain without issue, but on the single test machine we got, the resulting binaries could not be opened.

PR #7812, opened 2023-12-15
Wizard recipe: MAiNGO-v0.7.2

MAiNGO-github:
This pull request contains a new build recipe we built using the BinaryBuilder.jl wizard:

    Package name: MAiNGO
    Version: v0.7.2

## Auditor complains about AVX2

Without setting march the Auditor detects avx2.
However, after using `march="avx2"`, avx512 is detected.
Thus, we build without `march` as it is either a false detection or some dependency adapts to the `march` setting.
In both cases the selected setting should result in the most compatible binaries.

## Why only Windows and Linux ?

MacOS support is held back by the following issue (XCode seems to be too old (?)): https://github.com/JuliaPackaging/Yggdrasil/issues/7139 
Note that we tried the suggestion in https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1263 locally under Linux, but we received many errors for

```
rm -rf /opt/${target}/${target}/sys-root/System
cp -raf usr/* "/opt/${target}/${target}/sys-root/usr/."
cp -raf System "/opt/${target}/${target}/sys-root/."
```

similar to what was described in the issue comments.

FreeBSD was once successfully compiled using ` -DCMAKE_TOOLCHAIN_FILE=${CMAKE_/TARGET_TOOLCHAIN%.*}_gcc.cmake`, but MAiNGO is not properly tested on that platform.


M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://git.rwth-aachen.de/avt-svt/public/maingo.git", "252733413a29dbe5b84a4cdaf53e60e9934f372f"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd maingo/
```
giordano, line 16:
```suggestion
cd $WORKSPACE/srcdir/maingo/
```

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+git remote set-url origin https://git.rwth-aachen.de/avt-svt/public/maingo.git
```
giordano, line 17: Why?  Isn't this already the case?
```suggestion
```

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+mkdir build
+cd build
+git submodule init
+git submodule update -j 1
+if [[ "${target}" == x86_64-apple-darwin* ]]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.15
+    export CXXFLAGS="-mmacosx-version-min=10.15"
```
giordano, line 24: This is already the case when you set `MACOSX_DEPLOYMENT_TARGET`
```suggestion
```

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir
+cd maingo/
+git remote set-url origin https://git.rwth-aachen.de/avt-svt/public/maingo.git
+mkdir build
+cd build
+git submodule init
+git submodule update -j 1
+if [[ "${target}" == x86_64-apple-darwin* ]]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.15
```
giordano, line 23: I'm surprised this actually works: by default we use the SDK 10.12, if you really need a different version you need to add that to the sources and replace the default SDK with that one.  If you aren't doing that and the build is successful, maybe 10.15 isn't needed after all?

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+    export CXXFLAGS="-mmacosx-version-min=10.15"
+fi
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DMAiNGO_build_standalone=True -DMAiNGO_build_shared_c_api=True -DMAiNGO_build_parser=True -DMAiNGO_use_cplex=False -DMAiNGO_use_melon=False  ..
```
giordano, line 26: More readable
```suggestion
cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DMAiNGO_build_standalone=True \
    -DMAiNGO_build_shared_c_api=True \
    -DMAiNGO_build_parser=True \
    -DMAiNGO_use_cplex=False \
    -DMAiNGO_use_melon=False \
    ..
```

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+cmake --build . --config Release --parallel ${nproc} > make_out.txt
+mkdir -p ${libdir}
+mkdir -p ${bindir}
+find . -type f -name "*maingo-c-api.*" ! -name "*.cpp*" -exec cp '{}' ${libdir}/ \;
+find . -type f -name "MAiNGO*" !  -name "*.cpp*" -exec cp '{}' ${bindir}/ \;
+install_license ../LICENSE
+exit
```
giordano, line 33:
```suggestion
```

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libmaingo-c-api", :libmaingo_c_api),
+    ExecutableProduct("MAiNGOcpp", :MAiNGOcpp),
+    ExecutableProduct("MAiNGO", :MAiNGO)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
```
giordano, line 62: From the audit log:
```
│ 
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/XPzgk/src/auditor/co
```
```suggestion
dependencies = [
    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),
```

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+if [[ "${target}" == x86_64-apple-darwin* ]]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.15
+    export CXXFLAGS="-mmacosx-version-min=10.15"
+fi
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DMAiNGO_build_standalone=True -DMAiNGO_build_shared_c_api=True -DMAiNGO_build_parser=True -DMAiNGO_use_cplex=False -DMAiNGO_use_melon=False  ..
+cmake --build . --config Release --parallel ${nproc} > make_out.txt
+mkdir -p ${libdir}
+mkdir -p ${bindir}
+find . -type f -name "*maingo-c-api.*" ! -name "*.cpp*" -exec cp '{}' ${libdir}/ \;
+find . -type f -name "MAiNGO*" !  -name "*.cpp*" -exec cp '{}' ${bindir}/ \;
```
giordano, line 31:
```suggestion
install -Dvm 755 "MAiNGO${exeext}" "${bindir}/MAiNGO${exeext}"
install -Dvm 755 "MAiNGOcpp${exeext}" "${bindir}/MAiNGOcpp${exeext}"
install -Dvm 755 "libmaingo-c-api.${dlext}" "${libdir}/libmaingo-c-api.${dlext}"
```
should work as well, no?

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+mkdir build
+cd build
+git submodule init
+git submodule update -j 1
+if [[ "${target}" == x86_64-apple-darwin* ]]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.15
+    export CXXFLAGS="-mmacosx-version-min=10.15"
+fi
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DMAiNGO_build_standalone=True -DMAiNGO_build_shared_c_api=True -DMAiNGO_build_parser=True -DMAiNGO_use_cplex=False -DMAiNGO_use_melon=False  ..
+cmake --build . --config Release --parallel ${nproc} > make_out.txt
```
giordano, line 27: Why hiding the output?  That's very useful
```suggestion
cmake --build . --config Release --parallel ${nproc}
```

giordano:
> but we received many errors for
> ```
> rm -rf /opt/${target}/${target}/sys-root/System
> cp -raf usr/* "/opt/${target}/${target}/sys-root/usr/."
> cp -raf System "/opt/${target}/${target}/sys-root/."
> ```
> similar to what was described in the issue comments.

You need a privileged runner to be able to delete files from the root file system: https://docs.binarybuilder.org/stable/environment_variables/

M/MAiNGO/build_tarballs.jl
Downsite: I ran into this problem before, at least in my tests locally the git remote is changed for caching.
See https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/355

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+# platforms are passed in on the command line
+
+
+#Auditor complains about avx1.
+#Without march the Auditor detects avx2
+#but with march="avx2" avx512 is detected, so we build without march
+#We expand only to gfortan 4 and 5 (others seem not to have std::variant)
+#MacOS is in principle supported but requires newer SDK
+#see https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1263 (wee need std::shared_ptr and std::weak_ptr with array support that needs XCode 12.0)
+#FreeBsd builds only with gcc, that platform has not yet been sufficiently tested for inclusion.
```
fingolfin, line 46: But how do you plan to test it if you don't provide the binaries?

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+platforms = [
+    Platform("x86_64", "linux", libgfortran_version=v"4"),
+    Platform("x86_64", "linux", libgfortran_version=v"5"),
```
fingolfin, line 49: No `v3` on purpose?

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,66 @@
...
+    Platform("x86_64", "Windows", libgfortran_version=v"4"),
+    Platform("x86_64", "Windows", libgfortran_version=v"5")]
+platforms = expand_cxxstring_abis(platforms)
```
fingolfin, line 52: Personally I would recommend to start with
`platforms = supported_platforms()` and then filter out anything that really doesn't work. This way you get variants you are not thinking of, like muslc (instead of glibc). So something like this:
```julia
filter!(!Sys.isfreebsd, platforms)   #although I would just leave that in if it compiles
platforms = expand_gfortran_versions(platforms)   # if you really don't want/support v3, I'd filter it out after this
```

This way, if new platforms or Fortran variants or whatever, you'll automatically pick them up on the next rebuild.

M/MAiNGO/build_tarballs.jl
MAiNGO-github: We expand only to gfortan 4 and 5 (3 seems not to have std::variant)

M/MAiNGO/build_tarballs.jl
MAiNGO-github: I will add the complete code for MacOS and FreeBSD later once I get access to that machine where this was tried again.
Then we can check if the build succeeds.

From what I read, updating the SDK will likely not be enough, but lets see.

With FreeBSD we will have to force building with GCC. Would that be ok also for the MacOS case?

M/MAiNGO/build_tarballs.jl
MAiNGO-github: Good idea!

I have to look up what the corresponding command would be for gfortran 3.
Will add later.

M/MAiNGO/build_tarballs.jl
MAiNGO-github: MacOS was added, FreeBSD failed, likely because the way Ipopt is integrated upstream using CMake.

M/MAiNGO/build_tarballs.jl
MAiNGO-github: We tested again on here. FreeBSD did not build successfully, we thus excluded it for now.
To answer the question:
The original software is not yet tested on FreeBSD. Even if the binaries would be built, we do not have a FreeBSD testing machine ready. Likely, we won't have time to check for and fix platform specific bugs.

M/MAiNGO/build_tarballs.jl
```diff
@@ -0,0 +1,85 @@
...
+                     -DMAiNGO_build_shared_c_api=True \
+                     -DMAiNGO_build_parser=True \
+                     -DMAiNGO_use_cplex=False \
+                     -DMAiNGO_use_melon=False"
+
+# GCC used because of https://github.com/JuliaPackaging/Yggdrasil/issues/7139
+if [[ "${target}" == x86_64-apple-darwin* ]]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.15
+    cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+          -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake \
```
giordano, line 34: Why not clang?  Also, at this point you could avoid the conditional and use gcc everywhere, but I'd prefer to understand what's wrong with clang, that's usually the best choice for macOS.

giordano:
The error message on x86_64 macOS is (multiple times)
```
[13:29:02] In file included from /workspace/srcdir/maingo/dep/libale/src/symbol.cpp:12:
[13:29:02] In file included from /workspace/srcdir/maingo/dep/libale/src/symbol.hpp:14:
[13:29:02] In file included from /workspace/srcdir/maingo/dep/libale/src/value.hpp:16:
[13:29:02] /workspace/srcdir/maingo/dep/libale/src/tensor.hpp:460:31: error: no member named 'to_string' in namespace 'std'
[13:29:02]               "index " + std::to_string(index) + " out of bounds for shape = " + std::to_string(shape(0)));
[13:29:02]                          ~~~~~^
[13:29:02] /workspace/srcdir/maingo/dep/libale/src/tensor.hpp:460:87: error: no member named 'to_string' in namespace 'std'
[13:29:02]               "index " + std::to_string(index) + " out of bounds for shape = " + std::to_string(shape(0)));
[13:29:02]                                                                                  ~~~~~^
[13:29:02] 2 errors generated.
```
According to https://github.com/pytorch/pytorch/issues/3192#issuecomment-338229506 there may be a missing
```c++
#include <string>
```
Also C++ docs shows that `std::to_string` requires the `string` header file: https://en.cppreference.com/w/cpp/string/basic_string/to_string.  I don't see this header explicitly loaded in https://git.rwth-aachen.de/avt-svt/public/libale/-/blob/49bacdeaa0ae4c2236d01a65c8b31f87a8576926/src/tensor.hpp.  This looks an upstream bug to me.

MAiNGO-github:
> According to [pytorch/pytorch#3192 (comment)](https://github.com/pytorch/pytorch/issues/3192#issuecomment-338229506) there may be a missing
> 
> ```c++
> #include <string>
> ```
> 
> Also C++ docs shows that `std::to_string` requires the `string` header file: https://en.cppreference.com/w/cpp/string/basic_string/to_string. I don't see this header explicitly loaded in https://git.rwth-aachen.de/avt-svt/public/libale/-/blob/49bacdeaa0ae4c2236d01a65c8b31f87a8576926/src/tensor.hpp. This looks an upstream bug to me.

I agree that not including `<string>` is a mistake. In my understanding it is fragile as it implicitly assumes `<string>`  will be included at the place where the template is used. 

This has been noted and upstream will fix it in the next release.

 However, I believe this is not the reason the compilation failed, but an outdated SDK version.

 The first reason is that compilation outside of Binary Builder works with Clang on Linux.
 Secondly, further down in the log it complained that std::variant is not yet supported (and here gives explicity that that requires SDK 10.15).
After changing to 10.15, we no longer see the error about std::to_string.

Instead, the problem is, as we had seen previously, https://github.com/JuliaPackaging/Yggdrasil/issues/7139, that is shared_pointer to array is not yet supported.

For now I  suggest building for MacOS with GCC, which we have seen before works.
Is that fine?

MAiNGO-github:
@giordano Would you still have access to the logs of the failed pipeline of the lastest commit (8c2ce78) that I reference above, even after I add another commit ? 
I would like to put the pull request to the suggested state (=reverting to the situation of  d56546f) with passing tests that might be accepted.

PR #7829, opened 2023-12-20
[GTPSA] bump v1.1.1 (fix Windows bug not finding LAPACK)

mattsignorelli:


mattsignorelli:
Even though all Windows tests were previously passed, GTPSA_jll could not find liblapack on Windows machines (determined by using DependencyWalker.jl on the libgtpsa.dll generated). This patch should fix that bug

mattsignorelli:
BinaryBuilder does not work, so it is not easy to test this locally on a Windows machine. I generated the Windows binary via BinaryBuilder running on a Linux machine, and then moved the tar to my Windows machine and attempted to access. I am now getting the error upon dlopen "InitError: could not load library "C:\Users\Matt\.julia\artifacts\ec09de38eaa16266d5b4dd4b3bf972cead42f94b\bin\libgtpsa.dll"
%1 is not a valid Win32 application." So based on this (https://discourse.julialang.org/t/loading-a-shared-library-on-windows/10346/3) it seems LAPACK32_jll , which is 32 bit, is unable to run on Windows with Julia 64 bit installed.

giordano:
> it seems LAPACK32_jll , which is 32 bit, is unable to run on Windows with Julia 64 bit installed.

32-bit means the library uses the [LP64](https://en.wikipedia.org/wiki/64-bit_computing) interface, i.e. matrix indices are always 32-bit integers, this doesn't have anything to do with 32- vs 64-bit operating system.

mattsignorelli:
> > it seems LAPACK32_jll , which is 32 bit, is unable to run on Windows with Julia 64 bit installed.
> 
> 32-bit means the library uses the [LP64](https://en.wikipedia.org/wiki/64-bit_computing) interface, i.e. matrix indices are always 32-bit integers, this doesn't have anything to do with 32- vs 64-bit operating system.

I understand this. I've looked into this more and I think the error message I was getting is a bit misleading. Nonetheless I think I have at least an idea of what was going on. On MacOS and on Linux Julia v1.9+, I found cmake in BinaryBuilder had trouble finding LAPACK unless I created a `ln` for `liblapack32.dylib/so` to `liblapack.dylib/so`. So this is done on those operating systems, and it works. On Windows, all the tests passed for BinaryBuilder so I just assumed it worked. However on Julia v1.9+, `using GTPSA_jll` throws an error.

On Julia v1.8, it seems `using GTPSA_jll` does not work on MacOS and Linux, but *does* work on Windows. On Linux and MacOS, the library was` ldd`-ed to `liblapack32.so`, but included in the artifact was only `liblapack.so`. So cmake in BinaryBuilder on Julia v1.8 found `liblapack32.so` and linked to that, but the wrong `liblapack.so` file was included. 

On Windows Julia v1.8, because I did nothing, it found `liblapack32.dll` and seems to have worked in some cases. However using Juliaup v1.8.5 it doesn't work, for some reason, I've found.

Regardless, it seems the solution is to `ln` for `liblapack32.dll` to `liblapack.dll` also in Windows and up the compatibility to Julia v1.9+. I am unable to test this locally because BinaryBuilder doesn't work on Windows, but I believe this will work for Windows.

mattsignorelli:
OK, I investigated this more and have solved the problem.

I removed `ReferenceBLAS_jll` and instead used `OpenBLAS32_jll`. Before the first pull request to Yggdrasil I had trouble with `OpenBLAS_jll`,  and only now am realizing I probably should have been using `OpenBLAS32_jll`. With both `LAPACK32_jll` and `OpenBLAS32_jll`, I no longer need to do any `ln` in the build script.

On another note, I found `CMake_jll` does not work on `aarch64-linux-gnu`. I haven't checked other platforms, but the executable does not run in the workspace. I will submit an issue, but I don't actually need `CMake_jll` as a dependency for `GTPSA_jll` with the above fixes.

In any case, the build script much significantly cleaner now. No `ln`-ing and Julia compat is still 1.6. 

This pull request is ready  @imciner2

PR #7830, opened 2023-12-20
Wizard recipe: obabel-v3.1.0

vcanogil:

* Package name: obabel
* Version: v3.1.0


vcanogil:
I need this specific version of obabel because using version 3.1.1. results in bugs when trying to use as a julia package.

PR #7848, opened 2023-12-21
[LibSSH2] Apply upstream patch for CVE-2023-48795

giordano:


giordano:
Of course the patch doesn't apply cleanly.  And of course things where easier if they made releases after patching security vulnerabilities.

L/LibSSH2/build_tarballs.jl
```diff
@@ -16,6 +16,8 @@ cd $WORKSPACE/srcdir/libssh2*/
 
 # Apply patch from https://github.com/libssh2/libssh2/pull/1054
 atomic_patch -p1 ../patches/0001-mbedtls-use-more-size_t-to-sync-up-with-crypto.h.patch
+# https://github.com/libssh2/libssh2/pull/1291
+atomic_patch -p1 ../patches/0001-mbedtls-use-more-size_t-to-sync-up-with-crypto.h.patch
```
imciner2, line 20:
```suggestion
atomic_patch -p1 ../patches/0001-src-add-strict-KEX-to-fix-CVE-2023-48795-Terrapin-At.patch
```

imciner2: It probably helps to apply the patch you are adding instead of the existing one a second time :wink:

imciner2:
Could use a local rebase on top of the new version without MbedTLS, since the GitHub web UI isn't letting me resolve the conflict nicely and only keep this new patch.

imciner2:
Ok, @giordano it builds now after doing a bit of manual tweaking to fix the merge. Anything else you wanted to do on this or is it good to merge?

giordano:
I think it's good-ish, the only thing is that this got mangled with the move to OpenSSL, so we can't backport the security patch to previous julia versions :disappointed:

imciner2:
Yea, it was that mangling that actually broke this PR... Looking at the CVE, it is rated Medium and is a MITM attack, so we probably can get away with not rebuilding the old version - otherwise we have to go through the song and dance of reverting commits to get back the old MBedTLS version, which is probably going to be painful.

PR #7852, opened 2023-12-22
[finufft] Bump Finufft to v2.2.0 and use GCC12

ludvigak:
- Use latest release of FINUFFT library
- Switch to GCC12. Build was using GCC8 which is explicitly discouraged by FINUFFT team
- Switch from `ArchiveSource` to `GitSource` since GitHub archive hashes no longer are stable

### DIFF
--- a/F/finufft/build_tarballs.jl
+++ b/F/finufft/build_tarballs.jl
@@ -6,14 +6,17 @@ using BinaryBuilderBase
 include(joinpath(@__DIR__, "..", "..", "platforms", "microarchitectures.jl"))
 
 name = "finufft"
-version = v"2.1.0"
-
+version = v"2.2.0"
+commit_hash = "51892059a4b457a99a2569ac11e9e91cd2e289e7";
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


F/finufft/build_tarballs.jl
```diff
@@ -70,4 +72,4 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               preferred_gcc_version=v"8", julia_compat="1.6", augment_platform_block)
+               preferred_gcc_version=v"12", julia_compat="1.6", augment_platform_block)
```
giordano, line 75: GCC 12 with Julia v1.6 is going to be problematic for C++ code: https://github.com/JuliaPackaging/Yggdrasil/blob/c3f5c0ce1e1c2a275a3785022b08dcda1de56a0a/CONTRIBUTING.md#compatibility-tips.  Can we use GCC 10?

ludvigak: Should work! Updated now

PR #7853, opened 2023-12-22
[MKL_Headers] Rebuild 2023.2.0 to fix missing CMake files

imciner2:
We have already built 2024.0.0, and that has been fixed to include these CMake/pkgconfig files, but because 2023.2.0 is the last version that is installable on macOS, in order to compile for macOS targets, we should also add these build system files to the 2023.2.0 headers package as well.

I believe this should work with the registry and create a new version for 2023.2.0 and not kill 2024.0.0. I have also restored the old Julia compatibility that 2023.2.0 had, so this should be a transparent change.

### DIFF
--- a/M/MKL_Headers/build_tarballs.jl
+++ b/M/MKL_Headers/build_tarballs.jl
@@ -2,53 +2,68 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
+# Note: 2024.0.0 has already been built, but 2023.2.0 is being built to fix a bug in the packaging where the CMake/pkgconfig scripts were missing.
+# When updating to the next 2024.x.x release, the following must be done again (in addition to updating the sources):
+# * Bump the Julia compat to 1.6
+# * Remove the macos x86_64 platform and source files
+
 name = "MKL_Headers"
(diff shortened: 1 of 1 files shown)


M/MKL_Headers/build_tarballs.jl
```diff
@@ -2,54 +2,69 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
+# Note: 2024.0.0 has already been built, but 2023.2.0 is being built to fix a bug in the packaging where the CMake/pkgconfig scripts were missing.
+# When updating to the next 2024.x.x release, the following must be done again (in addition to updating the sources):
+# * Bump the Julia compat to 1.6
+# * Remove the macos x86_64 platform and source files
```
giordano, line 8: We can also revert the merge commit (with `[skip ci]`) after this PR is merged, no?

giordano (approved):

M/MKL_Headers/build_tarballs.jl
imciner2: We can do it that way, I hadn't thought of just reverting it right away.

PR #7856, opened 2023-12-25
[Doxygen] Update to 1.9.8

JamesWrigley:
In an attempt to fix weird segfaults in another package :octopus: 

The previous link was pointing to the auto-generated Github source which apparently can change its hash, so now it points to the tarball built by Doxygen.

### DIFF
--- a/D/Doxygen/bundled/patches/convert-path-to-std-string.patch
+++ b/D/Doxygen/bundled/patches/convert-path-to-std-string.patch
@@ -1,0 +1,22 @@
+diff --git a/src/portable.cpp b/src/portable.cpp
+index a14afdb95..853d9d74c 100644
+--- a/src/portable.cpp
++++ b/src/portable.cpp
+@@ -642,7 +642,7 @@ std::ofstream Portable::openOutputStream(const QCString &fileName,bool append)
+ #if defined(__clang__) && defined(__MINGW32__)
+   return std::ofstream(fs::path(fileName.str()).wstring(), mode);
+ #else
+-  return std::ofstream(fs::path(fileName.str()), mode);
++  return std::ofstream(fs::path(fileName.str()).string(), mode);
+ #endif
+ }
+ 
+@@ -654,7 +654,7 @@ std::ifstream Portable::openInputStream(const QCString &fileName,bool binary, bo
+ #if defined(__clang__) && defined(__MINGW32__)
+   return std::ifstream(fs::path(fileName.str()).wstring(), mode);
+ #else
+-  return std::ifstream(fs::path(fileName.str()), mode);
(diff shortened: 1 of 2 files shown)


JamesWrigley:
Hmph, looks like Doxygen versions after 1.9.2 use a filesystem library that uses `utimensat`, which is apparently only available for newer versions of OSX: https://github.com/ziglang/zig/issues/5023

Is there a way to set the minimum OSX version? Or would we need to remove support for OSX entirely?

giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/18413b89feaf52ebfa9afea2ca2c2cfa238e6369/L/libpolars/build_tarballs.jl#L14-L17
https://github.com/JuliaPackaging/Yggdrasil/blob/18413b89feaf52ebfa9afea2ca2c2cfa238e6369/L/libpolars/build_tarballs.jl#L22-L30

JamesWrigley:
Thanks, let's see if that works :crossed_fingers: When testing locally I got IO errors for some reason when running `rm -rf /opt/${target}/${target}/sys-root/System`, but I assume that's something fishy with my system. Also hopefully fixed the windows build errors with the new patch, I'll look into getting that upstreamed (see: https://github.com/doxygen/doxygen/pull/9854).

JamesWrigley:
Ok, builds all pass but I see this warning:
```
┌ Warning: Minimum instruction set detected for bin/doxygen.exe is avx2, not x86_64 as desired.
│   platform = "x86_64-w64-mingw32-cxx11"
└ @ BinaryBuilder.Auditor ~/.julia/dev/BinaryBuilder/src/Auditor.jl:323
```

Which I assume means that this will crash on windows machines without avx2? On the other hand, the 1.9.2 build also had that warning so at least this version isn't making things worse... If we're ok with the warning, then I think this can be merged.

D/Doxygen/bundled/patches/convert-path-to-std-string.patch
giordano, line 1: I don't understand this patch, are you basically undoing https://github.com/doxygen/doxygen/pull/9854?

JamesWrigley: I also don't fully understand why it was failing, since the patch author tested it on gcc :shrug: Maybe something that's fixed on a newer gcc version since the original compilation error was on gcc 12: https://github.com/doxygen/doxygen/pull/9757#issuecomment-1422640043

But it's not fully undoing the PR, it's only adding back the conversion to `std::string` for anything that's not clang on mingw32.

imciner2:
We have Doxygen 1.10.0 as of https://github.com/JuliaPackaging/Yggdrasil/pull/8294, so closing this one.

PR #7857, opened 2023-12-25
[MKL_Headers] Retrigger build

imciner2:
The previous build appears to have used build number +1, but that already existed for the 2023.2.0 version, so we need to retrigger the build to try to get build number +2 to get a unique version of the package.

@giordano I guess there is a bug in the version string generation somewhere? It appears that it recognized there was a previous 2023.2.0 release, so it tried to use the version string 2023.2.0+1 (see the logs here: https://buildkite.com/julialang/yggdrasil/builds/7341#018c9e04-e974-4c55-8b95-e8f68a36c17c). However 2023.2.0+1 is actually already a version, so it should have been 2023.2.0+2 that it used the first time.

### DIFF
--- a/M/MKL_Headers/build_tarballs.jl
+++ b/M/MKL_Headers/build_tarballs.jl
@@ -2,6 +2,8 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
+# Build counter: 1
+
 # Note: 2024.0.0 has already been built, but 2023.2.0 is being built to fix a bug in the packaging where the CMake/pkgconfig scripts were missing.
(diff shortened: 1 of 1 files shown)


imciner2:
Although it appears that 2023.2.0+1 was never actually registered in the General Registry: https://github.com/JuliaRegistries/General/blob/master/jll/M/MKL_Headers_jll/Versions.toml, which is due to https://github.com/JuliaRegistries/General/pull/96415.

So, what is the right way to go about this then, will the registry complain if we try to register a +2 without the +1 being in the versions list?

giordano:
Fixed by deleting the old `2023.2.0+1` release

imciner2:
> Fixed by deleting the old 2023.2.0+1 release

Excellent, thanks for that!

PR #7858, opened 2023-12-25
Add support for latest cuQuantum release

kshyatt:
https://docs.nvidia.com/cuda/cuquantum/latest/overview.html latest version is 23.10.0.6, which also seems to need CUTENSOR 2.0

cc @maleadt


C/CUDA/cuQuantum/build_tarballs.jl
```diff
@@ -8,8 +8,8 @@ include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
 include(joinpath(YGGDRASIL_DIR, "platforms", "cuda.jl"))
 
 name = "cuQuantum"
-version = v"23.6.1"
-version_str = "23.06.1"
+version = v"23.10.0"
+version_str = "23.10.0"
```
giordano, line 12:
```suggestion
version_str = "23.10.0"
version = VersionNumber(version_str)
```

PR #7865, opened 2023-12-28
[PETSc] rebuild 3.18.6 with suitesparse 7.2.1

boriskaus:
PETSc 3.18.6 does not work on Julia 1.10 because of an incompatibility with SuiteSparse (appears to require version >7.2.0), as can be seen [here](https://github.com/JuliaGeodynamics/LaMEM.jl/actions/runs/7349953191/job/20010762129?pr=21)

This is an attempt to rebuild 3.18.6 with version 7.2.1


P/PETSc/build_tarballs.jl
```diff
@@ -7,7 +7,7 @@ name = "PETSc"
 version = v"3.18.6"
 petsc_version = v"3.18.6"
 MUMPS_COMPAT_VERSION = "5.5.1"
-SUITESPARSE_COMPAT_VERSION = "5.10.1"
+SUITESPARSE_COMPAT_VERSION = "7.2.1"  # copmpiled with this version; compat removed below
```
giordano, line 10: You need to change julia compat, to require 1.10 then

P/PETSc/build_tarballs.jl
```diff
@@ -280,7 +280,7 @@ dependencies = [
     Dependency("OpenBLAS32_jll"; platforms=filter(!Sys.isapple, platforms)),
     Dependency("CompilerSupportLibraries_jll"),
     Dependency("SuperLU_DIST_jll"; compat=SUPERLUDIST_COMPAT_VERSION),
-    Dependency("SuiteSparse_jll"; compat=SUITESPARSE_COMPAT_VERSION),
+    Dependency("SuiteSparse_jll", compat=SUITESPARSE_COMPAT_VERSION),
```
giordano, line 283: Why this change?  In my opinion `;` is better when separating positional arguments from keyword ones.

P/PETSc/build_tarballs.jl
```diff
@@ -312,5 +322,8 @@ append!(dependencies, platform_dependencies)
 ENV["MPITRAMPOLINE_DELAY_INIT"] = "1"
 
 # Build the tarballs.
+# NOTE: llvm15 seems to have an issue with PETSc 3.18.x as on apple architectures it doesn't know how to create dynamic libraries
```
giordano, line 325: What's the problem exactly? Also, I think we don't have llvm 15 nor 14 at all, we jumped from v13 to v16, you can see list of compilers in https://github.com/JuliaPackaging/Yggdrasil/tree/29fc7d7beafd210b26fc2c4f3c6cdce76080bfab/0_RootFS

boriskaus: Builds for all apple systems were failing with this warning:
```
[20:31:09] *******************************************************************************
[20:31:09]          UNABLE to CONFIGURE with GIVEN OPTIONS    (see configure.log for details):
[20:31:09] -------------------------------------------------------------------------------
[20:31:09] Exhausted all shared linker guesses. Could not determine how to create a shared library!
[20:31:09] *******************************************************************************
```
when using the  `--with-shared-libraries=1` compilation flag. If this flag is not specified (which was the status-quo so far), it compiles a static library instead and, as a result, the audit [fails](https://buildkite.com/julialang/yggdrasil/builds/7414).

That's somewhat puzzling as previous compilations of PETSc_jll (e.g., [here](https://github.com/JuliaPackaging/Yggdrasil/pull/6765)) worked fine. It turns out that these versions use [clang 13.0.1](https://buildkite.com/julialang/yggdrasil/builds/3071#0188387b-a063-4ba0-b7db-f7590cf3b6aa/6-962), whereas the latest build ends up using [clang 16.0.6](https://buildkite.com/julialang/yggdrasil/builds/7414#018cb6bb-fb7f-45f8-a34d-3243c057942c/6-749), and with clang 13 it would create dynamic libraries (even when not specifying the compilation flag).

I expect that there might be a hickup in the PETSc installation system for clang 16 (perhaps that has been resolved in more recent versions of PETSc).
In any case, specifying `preferred_llvm_version=v"14"` results in [clang 13.0.1](https://buildkite.com/julialang/yggdrasil/builds/7437#018cbb9f-9ee0-485b-a8df-6c98d2c222a6/6-953). 

From what you suggests, it should rather be `preferred_llvm_version=v"13"`, not?

P/PETSc/build_tarballs.jl
```diff
@@ -4,12 +4,13 @@ const YGGDRASIL_DIR = "../.."
 include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
 
 name = "PETSc"
-version = v"3.18.6"
+version = v"3.18.7"
 petsc_version = v"3.18.6"
 MUMPS_COMPAT_VERSION = "5.5.1"
```
amontoison, line 9:
```suggestion
MUMPS_COMPAT_VERSION = "5.6.2"
```

amontoison (changes requested):
@boriskaus 
Can you also replace `libdmumps.${dlext}` by `libdmumpspar.${dlext}` at line 112?

https://github.com/JuliaPackaging/Yggdrasil/pull/7865/files#diff-7e36d662fb19edb9f7bab826e8c04fb5fb779932f5e66a1323e3d923a021866dR112

P/PETSc/build_tarballs.jl
boriskaus: sure thing

boriskaus (commented):
implemented

amontoison (approved):

PR #7869, opened 2023-12-29
Add OpenFHE library

sloede:
Build JLL package for [OpenFHE](https://github.com/openfheorg/openfhe-development).

### DIFF
--- a/O/OpenFHE/bundled/patches/macos-include-cmath.patch
+++ b/O/OpenFHE/bundled/patches/macos-include-cmath.patch
@@ -1,0 +1,10 @@
+--- a/src/core/include/utils/utilities.h
++++ b/src/core/include/utils/utilities.h
+@@ -36,6 +36,7 @@
+ #include "utils/inttypes.h"
+
+ #include <climits>  // CHAR_BIT
++#include <cmath>
+ #include <limits>   // std::numeric_limits
+ #include <string>
+ #include <type_traits>  // std::is_integral
--- a/O/OpenFHE/build_tarballs.jl
+++ b/O/OpenFHE/build_tarballs.jl
@@ -31,9 +31,11 @@ make install
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = supported_platforms()
-# platforms = [
-#     Platform("x86_64", "linux"; libc = "glibc")
-# ]
+
+# We cannot build with musl since OpenFHE requires the `execinfo.h` header for `backtrace`
+platforms = filter(p -> libc(p) != "musl", platforms)
+
+# Expand C++ string ABIs since we use std::string
 platforms = expand_cxxstring_abis(platforms)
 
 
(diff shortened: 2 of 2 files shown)


sloede:
Ready for review and & merge 🙏 @giordano

O/OpenFHE/bundled/patches/macos-include-cmath.patch
giordano, line 1: Can this patch be upstreamed? It doesn't look specific to Yggdrasil.

sloede: Unfortunately, I failed to reproduce the error on my Macbook. I thus figured it is maybe a specific combination of Xcode and/or Clang. But I will try to get it upstreamed and amend the build recipe here accordingly once (if) a new, fixed upstream version is available.

It would be great if we could still get this PR merged already now, to allow us to continue working on further downstream packages.

giordano: It's not a matter of being able to reproduce, but correctness: if the standard declares that a function is defined in a certain header file, the header file should always be included, regardless of whether some toolchain happen to accidentally define said function through other includes. And my understanding is that `abs` is supposed to be defined by `cmath`: https://cplusplus.com/reference/cmath/

sloede: https://github.com/openfheorg/openfhe-development/pull/633

sloede: > It's not a matter of being able to reproduce the error, but of correctness: if the standard declares that a function is defined in a certain header file, the header file should always be included, regardless of whether some toolchains happen to accidentally define said function through other includes. And my understanding is that `abs` is supposed to be defined by `cmath`: https://cplusplus.com/reference/cmath/

Sorry, just saw your reply now. Totally agree. I've created an upstream PR. It would still be great though if we can proceed here and I create a new release once the upstream fix has been published.

giordano: Yeah, that's fine.

sloede: Note though that the problematic `abs` here is the one for `int` arguments (see error message in https://github.com/openfheorg/openfhe-development/pull/633). Rereading the references, [`cmath`](https://en.cppreference.com/w/cpp/header/cmath) only defines the float versions of `abs`, while [`cstdlib`](https://en.cppreference.com/w/cpp/header/cstdlib) seems to be indeed the source for the `int` version (and is included in the header in question). However, including `cstdlib` does not prevent the error either, so I'm waiting to see if the upstream folks have a take on this 🤷‍♂️

imciner2: > Note though that the problematic abs here is the one for int arguments

Doesn't look like it. The error message emanating from line 118 in that header is trying to do `std::abs` on a double value since the code is
```
inline bool is64BitOverflow(double d) {
    return std::abs(d) > static_cast<double>(Max64BitValue());
}
```
so it needs `cmath` to have that `abs`. (and similarly for the error from line 137, which is also doing an abs of a double).

sloede: Thanks for your insights. I'll pass this on to the upstream devs should it come up.

Issue #7132, opened 2023-07-29
llama_cpp: Build with METAL support for apple sillicon


svilupp:
@marcom , I was wondering if it would make sense to change the Llama.cpp binaries to build with metal support for AARCH64?
Apple silicon macs all have metal support and the text generation is 20x faster.

Currently, it prefers Apple Accelerate: https://github.com/JuliaPackaging/Yggdrasil/blob/14c399600723fdc69b845af526ca06fdc35ad4ae/L/llama_cpp/build_tarballs.jl#L20

I would like to propose adding the following: 
```
# use METAL on Apple devices
if [[ "${proc_family}" == "arm" && "${nbits}" == 64 ]]; then
    EXTRA_CMAKE_ARGS="$EXTRA_CMAKE_ARGS -DLLAMA_METAL=ON"
fi
```
based on the Llama.cpp [README.md](https://github.com/ggerganov/llama.cpp/blob/8a88e5855c3b93024be0f93290b01a4206b65b38/README.md?plain=1#L278)

FYI. When I try it, I get a warning on import:
```
objc[63847]: Class GGMLMetalClass is implemented in both /Users/xxx/.julia/artifacts/9b75f107cfc237d11b901342b84e724445754599/lib/libllama.dylib (0x133ec8688) and /Users/xxx/.julia/artifacts/9b75f107cfc237d11b901342b84e724445754599/lib/libggml_shared.dylib (0x13f7643d0). One of the two will be used. Which one is undefined.
```
(Probably because both are provided as libraries)

PR #7874, opened 2023-12-30
[llama_cpp] Update to v0.0.14 (b1767, 2024-01-04)

svilupp:
- Update to v0.0.14 (b1729, 2023-12-30)
- Fixes `GGMLMetalClass` error on import when using Metal on Apple Silicon (referenced in https://github.com/JuliaPackaging/Yggdrasil/issues/7132)
- Provides support for Llava (images), Mixtral, Phi and many more new models and architectures

Changes to the build script:
- removed armv{6,7} specific CMAKE ARGS as the flag `LLAMA_K_QUANTS` is no longer available
- removed Product "embd_input_test" as it's no longer part of the project
- removed Library "libembdinput" as it's no longer part of the project
- disabled METAL (LLAMA_METAL=OFF) on Intel-based MacOS as it's not supported (supported on Apple Silicon only)

Build succeeds locally and inference on Metal finally works without any errors.

cc: @marcom


svilupp:
With the latest change, x86_64 MacOS builds again (locally).

Suspected root cause: Llama.cpp has enabled METAL by default for all MacOS devices, however, only Apple Silicon can run with Metal support.

Solution: Disable Metal support for all MacOS architectures except for aarch64.

svilupp:
I'm puzzled by the failure of "aarch64-apple-darwin". Locally it builds without a problem and I cannot find any weird flags or indication where it's coming from.

The only difference seems to be Clang 13 locally vs Clang 16 in the buildkite, but that shouldn't be a problem, right?

giordano:
> The only difference seems to be Clang 13 locally vs Clang 16 in the buildkite, but that shouldn't be a problem, right?

What version of BinaryBuilderBase do you have?  LLVM 16 was added in [v1.26.0](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/releases/tag/v1.26.0), back in October, you seem to have an older version?

Also, since the error message was
```
[12:06:36] /workspace/srcdir/llama.cpp/ggml-quants.c:413:25: error: redefinition of 'vdotq_s32'
[12:06:36] inline static int32x4_t vdotq_s32(int32x4_t acc, int8x16_t a, int8x16_t b) {
[12:06:36]                         ^
[12:06:36] /opt/x86_64-linux-musl/lib/clang/16/include/arm_neon.h:33859:51: note: previous definition is here
[12:06:36] __ai __attribute__((target("dotprod"))) int32x4_t vdotq_s32(int32x4_t __p0, int8x16_t __p1, int8x16_t __p2) {
[12:06:36]                                                   ^
[12:06:36] 1 error generated.
```
it wasn't unlikely that the error came precisely from using a newer version of the compiler.

> EDIT: I cannot get Windows to build, so temporarily disabling.

For the record, the error message is
```
[12:07:59] /workspace/srcdir/llama.cpp/ggml.c: In function ‘ggml_init’:
[12:07:59] /workspace/srcdir/llama.cpp/ggml.c:193:35: error: implicit declaration of function ‘_aligned_malloc’; did you mean ‘_aligned_realloc’? [-Werror=implicit-function-declaration]
[12:07:59]   193 | #define GGML_ALIGNED_MALLOC(size) _aligned_malloc(size, GGML_MEM_ALIGN)
[12:07:59]       |                                   ^~~~~~~~~~~~~~~
```
1. developers should maybe stop forcing `-Werror` onto packagers
2. this seems to be https://sourceforge.net/p/mingw-w64/bugs/192/, which should have a simple fix instead of just removing the platform completely, no?

svilupp:
Thank you, Mose! That explains a lot.

Re. Windows build, thank you for the tip! I haven't managed to google anything sensible and playing with flags didn't help!
I've tried simply disabling the -Werror but that fails anyway.
Given the linked article, I've tried reordering headers locally before I open the issue in the library - no success. It does finish, but none of the products/libraries are in the destdir...

I'll open the issue in the library, but for now, I don't see any immediate resolution for the Windows builds, because any future fix would bring back the bug for Metal (that issue is still unresolved).

svilupp:
As a side note, have you ever encountered an issue where the build process finishes but none of the products/libraries are available in the destdir _just for windows builds_?

It seems like a silly error with copying the final products, but it's unlikely to be an issue in the makefile since this library is built by many downstream packages with Windows binaries. It must be some silly mistake on my end.
It works fine for all the other platforms.

When I patch the header order, the build process finishes and I can even see all executables are being made:

> [18:39:22] [ 74%] Linking CXX executable ../../bin/perplexity.exe
> [18:39:22] cd /workspace/srcdir/llama.cpp/build/examples/perplexity && /usr/bin/cmake -E cmake_link_script CMakeFiles/perplexity.dir/link.txt --verbose=true
> [18:39:22] /usr/bin/cmake -E rm -f CMakeFiles/perplexity.dir/objects.a
> [18:39:22] /opt/bin/x86_64-w64-mingw32-libgfortran5-cxx03/x86_64-w64-mingw32-ar qc CMakeFiles/perplexity.dir/objects.a @CMakeFiles/perplexity.dir/objects1.rsp
> [18:39:22] /opt/bin/x86_64-w64-mingw32-libgfortran5-cxx03/x86_64-w64-mingw32-g++ --sysroot=/opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/ -O3 -DNDEBUG -Wl,--whole-archive CMakeFiles/perplexity.dir/objects.a -Wl,--no-whole-archive -o ../../bin/perplexity.exe -Wl,--out-implib,libperplexity.dll.a -Wl,--major-image-version,0,--minor-image-version,0 @CMakeFiles/perplexity.dir/linklibs.rsp
> [18:39:23] make[2]: Leaving directory '/workspace/srcdir/llama.cpp/build'

But they stay in srcdir/llama.cpp/build/bin/perplexity.exe (same for all the other executables).

So the final check of products fails:

>  Info: ["perplexity"] does not exist, reporting unsatisfied
> ┌ Error: Built llama_cpp but perplexity still unsatisfied:
> └ @ BinaryBuilder ~/.julia/packages/BinaryBuilder/MoPsh/src/AutoBuild.jl:946

Have you ever encountered a similar issue?

giordano:
https://github.com/ggerganov/llama.cpp/issues/4654 should have been fixed.

giordano:
And moving
```c
#include <stdlib.h>
```
before any other includes in `ggml.c` fixes the issue for me.  `make install` installs all files in `${bindir}` correctly for me, also on Windows.

svilupp:
Thank you for the help, Mose! Awesome trick to patch the source code - I’m saving it for later! 
(I don’t think I’ve seen it in the BinaryBuilder docs?)

I’m also waiting for the official fix upstream - once that’s merged, I’ll open a new PR to bump up the version.

giordano:
> (I don’t think I’ve seen it in the BinaryBuilder docs?)

Using `DirectorySource` for patches is briefly mentioned at https://docs.binarybuilder.org/stable/building/#Sources, and the `atomic_patch` utility is described at https://docs.binarybuilder.org/stable/build_tips/#utils_build_env.  Then there are hundreds of recipes here in Yggdrasil using patches.

> EDIT2: So the windows issue has been fixed upstream, but I'd still prefer to merge this PR to have a record of your workaround -- I think we'll need that trick in the future. Once merged, I'll open a new PR with the latest release.

I'm not sure having the fix hidden in the git history would help anything.  I'd prefer we update this PR to the latest version and move on.  There's already a record of the fix at https://github.com/JuliaPackaging/Yggdrasil/pull/7874#issuecomment-1875936339 (and in the upstream repository, too), and I easily found https://sourceforge.net/p/mingw-w64/bugs/192/ with a simple search, it should be quite easy for future users to find the solution to the problem.

PR #7876, opened 2023-12-30
Use new keyword argument `clang_use_lld` where possible

giordano:
There are some cases where using `lld` as linker together with Clang is proving to be problematic (for example #7745), this new keyword argument lets you easily switch back to using `ld`.


giordano:
It worked everywhere, yay!

PR #7877, opened 2023-12-30
[LLVMCompilerRT] Build for more platforms

giordano:


barche:
If it's OK to add LLVM_full as a build dependency then this should be good to go. I couldn't find any other way to fix the build on the non-Apple platforms, only on Apple it seems LLVM_full is not needed for the build.

barche:
Hi @giordano is this OK for you please?

PR #7878, opened 2023-12-30
[SuiteSparse] Update to v7.4.0

imciner2:
This release had a large refactor in how the build system was configured, with the introduction of an overarching CMake file in the root directory that now can select the projects to build, and also the renaming of a lot of the SuiteSparse options we need to configure.

### DIFF
--- a/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
+++ b/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
@@ -1,7 +1,7 @@
 include("../common.jl")
 
 name = "SuiteSparse"
-version = v"7.3.0"
+version = v"7.4.0"
 
 sources = suitesparse_sources(version)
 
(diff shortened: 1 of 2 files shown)


amontoison (approved):

S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
```diff
@@ -37,29 +37,32 @@ else
...
+      -DSUITESPARSE_USE_STRICT=ON \
+      -DSUITESPARSE_USE_CUDA=OFF \
+      -DSUITESPARSE_USE_FORTRAN=OFF \
+      -DSUITESPARSE_USE_OPENMP=OFF \
+      -DNPARTITION=OFF \
+      -DBLAS_FOUND=1 \
+      -DBLAS_LIBRARIES="${libdir}/lib${BLAS_NAME}.${dlext}" \
+      -DBLAS_LINKER_FLAGS="${BLAS_NAME}" \
+      -DBLAS_UNDERSCORE=ON \
+      -DBLA_VENDOR="${BLAS_NAME}" \
```
giordano, line 58: `BLA_VENDOR` is a typo?

imciner2: No, SuiteSparse references this to determine the vendor of the BLAS, so since we supply all the flags ourselves, we need to set it ourselves.

This lets it print it out in the CMake log as well, e.g. https://buildkite.com/julialang/yggdrasil/builds/7473#018cc579-e8d8-4282-b92c-fef9a361a31b/630-935 has
```
[14:45:18] -- Found blastrampoline 64-bit BLAS
[14:45:18] -- BLAS64_suffix: _64
[14:45:18] -- BLAS64 suffix has an underscore
[14:45:18] -- BLAS libraries:      /workspace/destdir/lib/libblastrampoline.dylib
[14:45:18] -- BLAS linker flags:   blastrampoline
[14:45:18] -- BLAS include:
```

giordano: I was just wondering whether the name `BLA` instead of `BLAS` was a typo on our side, but it's the name used upstream.

imciner2: Yea, everything starting with `BLAS` is the CMake-internal variables, so SuiteSparse used the `BLA` suffix. I think I recall a comment recently about how that is confusing for other people, but it doesn't look like this was renamed during the recent cleanup of the options.

PR #7880, opened 2023-12-31
OpenFHE: fix library install directory in CMake

sloede:
cc @giordano 

I will also attempt to figure out if this can be fixed upstream, i.e., that OpenFHE directly installs the `.dll` files to the correct location (`bin/`) such that this postfix is not necessary anymore.

### DIFF
--- a/O/OpenFHE/bundled/patches/windows-fix-cmake-libdir.patch
+++ b/O/OpenFHE/bundled/patches/windows-fix-cmake-libdir.patch
@@ -1,41 +1,39 @@
---- a/workspace/destdir/CMake/OpenFHETargets-release.cmake
-+++ b/workspace/destdir/CMake/OpenFHETargets-release.cmake
-@@ -9,31 +9,31 @@ set(CMAKE_IMPORT_FILE_VERSION 1)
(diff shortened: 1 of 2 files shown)


sloede:
I believe the culprit is here:
https://github.com/openfheorg/openfhe-development/blob/b2869aef5cf61afd364b3eaea748dcc8a7020b9c/src/binfhe/CMakeLists.txt#L38
(similar for the other two libraries created by OpenFHE). It forces all library products to be installed in the `lib` dir, which is wrong on Windows systems. See also the sections on the `DESTINATION` and the `GNUInstallDirs` in https://cmake.org/cmake/help/latest/command/install.html#targets.

O/OpenFHE/bundled/patches/windows-fix-cmake-libdir.patch
giordano, line 1: It's presumably easier to patch the CMake file which installs the file?

sloede: Sorry, yes. Once I figured this out, I submitted an upstream patch (https://github.com/openfheorg/openfhe-development/pull/634) and updated the PR here with the same patch. It would be great if we can merge this already now, and I'll update the `build_tarballs.jl` once an upstream fix has been released.

PR #7881, opened 2023-12-31
[SuiteSparse@7] Update some CMake configuration options

giordano:
Ref https://github.com/JuliaLang/julia/pull/52577#issuecomment-1872963230 and https://github.com/JuliaLang/julia/pull/52577#issuecomment-1872967550.  Most of the other options were fixed by #7878, but some were missed.

### DIFF
--- a/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
+++ b/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
@@ -89,4 +89,4 @@ install_license LICENSE.txt
 """
 
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat="1.10")
+               julia_compat="1.11")
--- a/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
+++ b/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
@@ -39,9 +39,6 @@ fi
 
 PROJECTS_TO_BUILD="suitesparse_config;amd;btf;camd;ccolamd;colamd;cholmod;klu;ldl;umfpack;rbio;spqr"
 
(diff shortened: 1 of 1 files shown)


imciner2:
Let's wait to see if any of those other options should be removed/changed. `LAPACK_FOUND` is probably a CMake option, not SuiteSparse (so it is a bit weird it isn't being used, but that might be because of their build system changing how they check for BLAS/LAPACK), but `NPARTITION` I think is SuiteSparse, but when I looked through there code last night that flag was still in use in the C code, but I couldn't see what the CMake for it was now. `BLAS_UNDERSCORE` I am not sure if it is CMake or SuiteSparse.

giordano:
@mmuetzel how does it look now?

mmuetzel:
@DrTimothyAldenDavis was writing that the Fortran name mangling should be detected automatically by the build system of SuiteSparse. Afaict, that requires that the Fortran compiler that was used to build the Fortran libraries (like BLAS or LAPACK) is installed when SuiteSparse is built (or at least a Fortran compiler that uses compatible name mangling).
If that is the case, you could probably remove setting that preprocessor variable.

On the other hand: If you are sure that the Fortran name mangling is one single trailing underscore, you could also set that variable.

Other than that, it looks good to me. 👍

mmuetzel:
Also, I don't know if your build environment already sets `CFLAGS` and/or `CXXFLAGS`. If you prefer to retain their values, you might want to append to them instead. E.g., `CFLAGS="$CFLAGS -DBLAS_UNDERSCORE"

giordano:
No, we don't set the compiler `*FLAGS` environment variables by default, so the script starts from a clean slate.  And it looks like the builds after removing `-DBLAS_UNDERSCORE` are by-by-bit identical to when setting that variable, which suggests the build system was inferring correctly the setting.  Thanks for the quick review!

PR #7890, opened 2024-01-02
[SuperLU_DIST] version 8.2.1

boriskaus:
increments the version

### DIFF
--- a/S/SuperLU_DIST/build_tarballs.jl
+++ b/S/SuperLU_DIST/build_tarballs.jl
@@ -6,12 +6,12 @@ const YGGDRASIL_DIR = "../.."
 include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
 
 name = "SuperLU_DIST"
-version = v"8.1.2"
-superlu_dist_version = v"8.1.2"
+version = v"8.2.1"
+superlu_dist_version = v"8.2.1"
 
 # Collection of sources required to complete build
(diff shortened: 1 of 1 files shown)


S/SuperLU_DIST/build_tarballs.jl
```diff
@@ -89,6 +89,17 @@ platforms, platform_dependencies = MPI.augment_platforms(platforms; MPItrampolin
 # Avoid platforms where the MPI implementation isn't supported
 # OpenMPI
 platforms = filter(p -> !(p["mpi"] == "openmpi" && arch(p) == "armv6l" && libc(p) == "glibc"), platforms)
+
+# OpenMPI 5 supports only 64-bit systems
+platforms = filter(p -> !(p["mpi"] == "openmpi" && nbits(p) != 64), platforms)
```
giordano, line 94: Target OpenMPI v4.1.6 instead

boriskaus:
There appears to be a bigger issue with `SuperLU_DIST` and `PETSc` (compiles fine, but does not actually work), so I'm putting this on hold for now

boriskaus:
ok, I've added a testing framework to check that the installation works on different systems (see [here](https://github.com/boriskaus/test_SuperLU_DIST_jll)). It now passes the CI for a number of systems:
https://github.com/boriskaus/test_SuperLU_DIST_jll/actions/runs/7422000918

PR #7896, opened 2024-01-03
[AMGCL_C] new build_tarballs for C Wrapper to AMGCL

j-fu:
Build script for [AMGCL_C](https://github.com/j-fu/amgcl_c) which is a C wrapper to [AMGCL](https://github.com/ddemidov/amgcl), an algebraic multigrid (AMG) based iterative solver library.

Intended to serve as the backend to [AMGCLWrap.jl](https://github.com/j-fu/AMGCLWrap.jl) intended to be registered once this PR is approved.

### DIFF
--- a/A/AMGCL_C/build_tarballs.jl
+++ b/A/AMGCL_C/build_tarballs.jl
@@ -10,10 +10,6 @@
 #
 # During the build, AMGCL_C downloads AMGCL via ExternalProject_Add() 
 #
-# AMGCL_C uses OpenMP. Unfortunately, on Apple, CMake is unable to find it
-# so currently, OpenMP probably won't be available on MacOS.
-#
-#
 # CMake parameters:
 #   BUILD_DI_INTERFACE: Build interface for double + int
 #   BUILD_DL_INTERFACE: Build interface for double + long
--- a/A/AMGCL_C/build_tarballs.jl
+++ b/A/AMGCL_C/build_tarballs.jl
@@ -55,7 +55,9 @@ cmake -DCMAKE_INSTALL_PREFIX=$prefix\
       $(interfaces)\
       -DBLOCKSIZES="BLOCKSIZE(2) BLOCKSIZE(3) BLOCKSIZE(4) BLOCKSIZE(5) BLOCKSIZE(6) BLOCKSIZE(7) BLOCKSIZE(8)"\
       -B build .
-make -C build install
+cd build
+make -j${nprocs}
+make install
 """
 
 # These are the platforms we will build for by default, unless further
--- a/A/AMGCL_C/build_tarballs.jl
+++ b/A/AMGCL_C/build_tarballs.jl
@@ -30,7 +30,7 @@ version = v"0.1.0"
 # Collection of sources required to complete build
 # This accesses AMGCL version 1.4.4 and AMGCL_C 0.1.2
 sources = [
-    GitSource("https://github.com/j-fu/amgcl_c.git", "afe76c46d2bd68aadccd04f56d55e5f7ce78a4b1")
+    GitSource("https://github.com/j-fu/amgcl_c.git", "aab4d220948c0ea269ac624f540f836c8047fb0f")
 ]
 
 # Bash recipe for building across all platform
--- a/A/AMGCL_C/build_tarballs.jl
+++ b/A/AMGCL_C/build_tarballs.jl
@@ -30,7 +30,7 @@ version = v"0.1.0"
 # Collection of sources required to complete build
 # This accesses AMGCL version 1.4.4 and AMGCL_C 0.1.2
 sources = [
-    GitSource("https://github.com/j-fu/amgcl_c.git", "6604079edbb76f18a5a73a07b7638eb26b39fafe")
+    GitSource("https://github.com/j-fu/amgcl_c.git", "afe76c46d2bd68aadccd04f56d55e5f7ce78a4b1")
 ]
 
 # Bash recipe for building across all platform
--- a/A/AMGCL_C/build_tarballs.jl
+++ b/A/AMGCL_C/build_tarballs.jl
@@ -30,7 +30,7 @@ version = v"0.1.0"
 # Collection of sources required to complete build
 # This accesses AMGCL version 1.4.4 and AMGCL_C 0.1.2
 sources = [
-    GitSource("https://github.com/j-fu/amgcl_c.git", "aab4d220948c0ea269ac624f540f836c8047fb0f")
+    GitSource("https://github.com/j-fu/amgcl_c.git", "6c8c3bcdd793cf3098b866e261ee3008a555a41c")
 ]
 
 # Bash recipe for building across all platform
--- a/A/AMGCL_C/build_tarballs.jl
+++ b/A/AMGCL_C/build_tarballs.jl
@@ -76,7 +76,10 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
+    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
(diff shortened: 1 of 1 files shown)


A/AMGCL_C/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+   interfaces="-DBUILD_DI_INTERFACE=True  -DBUILD_DL_INTERFACE=True"
+fi
+
+cmake -DCMAKE_INSTALL_PREFIX=$prefix\
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}\
+      -DCMAKE_BUILD_TYPE=Release\
+      $(interfaces)\
+      -DBLOCKSIZES="BLOCKSIZE(2) BLOCKSIZE(3) BLOCKSIZE(4) BLOCKSIZE(5) BLOCKSIZE(6) BLOCKSIZE(7) BLOCKSIZE(8)"\
+      -B build .
+make -C build install
```
imciner2, line 58:
```suggestion
cd build
make -j${nprocs}
make install
```

A/AMGCL_C/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libamgcl_c", :libamgcl_c),
+    FileProduct("include/amgcl_c/amgcl_c.h",:amgcl_c_h)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    Dependency(PackageSpec(name="boost_jll", uuid="28df3c45-c428-5900-9ff8-a3135698ca75"))
```
imciner2, line 78: Boost is notoriously bad about ABI with versions, so please include a compat entry for which Boost versions this library can support.

A/AMGCL_C/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libamgcl_c", :libamgcl_c),
+    FileProduct("include/amgcl_c/amgcl_c.h",:amgcl_c_h)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
```
imciner2, line 77:
```suggestion
    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
```
This is the standard way to do OpenMP in Yggdrasil, and may fix the problem of not finding OpenMP on Apple platforms.

A/AMGCL_C/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+    Dependency(PackageSpec(name="boost_jll", uuid="28df3c45-c428-5900-9ff8-a3135698ca75"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"10.2.0")
```
imciner2, line 82: Does this really need GCC 10? In Yggdrasil, we aim to build with the oldest possible GCC version to promote compatibility with more systems (see https://github.com/JuliaPackaging/Yggdrasil/blob/master/RootFS.md).

imciner2 (changes requested):

A/AMGCL_C/build_tarballs.jl
j-fu: Thanks, will try. The part of boost used here seems to be rather stable (I am not a boost specialist, though).

A/AMGCL_C/build_tarballs.jl
j-fu: I had problems on 4.8.5, will retry.

j-fu: This is the error I get:
```
[17:21:00] /workspace/srcdir/amgcl_c/src/default_params.cxx: In function ‘boost::property_tree::ptree boost_params(char*)’:
[17:21:00] /workspace/srcdir/amgcl_c/src/default_params.cxx:6:92: error: no matching function for call to ‘regex_replace(std::string, std::regex,const char [2])’
```

j-fu: So minimum gcc is now v4

j-fu:
OpenMP is now required by cmake - so the hint helped to fix it on apple.

j-fu:
Please wait with merging, found another glitch...

j-fu:
Will be ok if build is ok.

j-fu:
... still refactoring a bit...

A/AMGCL_C/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+    FileProduct("include/amgcl_c/amgcl_c.h",:amgcl_c_h)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
+    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
+    Dependency(PackageSpec(name="boost_jll", uuid="28df3c45-c428-5900-9ff8-a3135698ca75"),v"1.79.0")
```
giordano, line 78: You have to use `compat` to specify you the compatibility bound:
```suggestion
    Dependency(PackageSpec(name="boost_jll", uuid="28df3c45-c428-5900-9ff8-a3135698ca75"); compat="=1.79.0")
```
Also, boost breaks the ABI in every single version, so you have to force compatibility with that single version (the `=` sigil).

giordano (commented):
Git history seems to be messed up

j-fu:
Seems to be good now.

A/AMGCL_C/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+#
+# Handling of integer types attempts to be similar to umfpack: build both DL and DI on
+# 64 bit systems and only DI on 32bit. See
+# https://github.com/JuliaSparse/SparseArrays.jl/blob/feb54ee5e49008bd157227099cafe604a67c36fb/src/solvers/umfpack.jl#L145
+# https://github.com/JuliaSparse/SparseArrays.jl/blob/feb54ee5e49008bd157227099cafe604a67c36fb/src/solvers/umfpack.jl#L578
+#
+using BinaryBuilder, Pkg
+
+name = "AMGCL_C"
+version = v"0.1.0"
```
imciner2, line 28: Your comment says 0.1.2 below, but this version is 0.1.0, and there doesn't appear to be a 0.1.2 release yet (the below commit SHA is just the most recent commit on main.

It would be good to do another release and then update this version number before we merge this.

imciner2 (commented):
The git history also seems to include things that aren't related to this, so some git merge seems to have gone awry. Can you do a local rebase to clean up the parts of the history that are unrelated to your script?

A/AMGCL_C/build_tarballs.jl
j-fu: Relased AMGCL_C 0.1.2 and set script version to 0.1.2 .

j-fu:
Hopefully ok now...

imciner2 (approved):

imciner2:
Yep, looks good. Thanks for your work on this.

PR #7898, opened 2024-01-03
[SCIP] no optimization on windows

matbesancon:


imciner2:
I guess this is fine. Is there a specific reason why even O1 is problematic on Windows? O1 optimizations are supposed to be pretty safe.

matbesancon:
The reason is trying to finally nail down the Windows issue that's been running forever:
https://github.com/JuliaPackaging/Yggdrasil/pull/6054#issuecomment-1373633154

matbesancon:
thanks!

PR #7899, opened 2024-01-03
Update openfhe-julia to v0.2.0

sloede:


sloede:
@giordano ready for review & merge from my POV

barche:
I see this PR uses `libjulia_jll` version `1.10.8`, but the latest `libcxxwrap_julia_jll` is built using version `1.10.7`, but I think this is harmless since libsingular does the same.

sloede:
> I see this PR uses `libjulia_jll` version `1.10.8`, but the latest `libcxxwrap_julia_jll` is built using version `1.10.7`, but I think this is harmless since libsingular does the same.

True, but libsingular does not build on Windows, so there's no telling if it would have an impact there.

Just to be on the safe side, I changed the `libjulia_jll` and preferred GCC version to be identical to libcxxwrap-julia's settings in [60c2711](https://github.com/JuliaPackaging/Yggdrasil/pull/7899/commits/60c2711581ccb60027e6d6d505ffef38b73acd1b).

Could you [applying `RTLD_GLOBAL`](https://github.com/JuliaPackaging/Yggdrasil/blob/1210bc5fde582783d6dd87b5f36d4ad36fea42c8/L/libcxxwrap_julia/build_tarballs.jl#L46C59-L46C86) have any effect that I should consider here as well?

sloede:
@barche Any idea what's going on here:
https://buildkite.com/julialang/yggdrasil/builds/7543#018cd43a-c16b-42d9-90f2-087d4cf1a5f5/604-807

sloede:
@barche I think I figured it out: With the newest build of CSL from Oct 2023 (only for Julia v1.11+), the location of the offending libraries has been moved to `lib/gcc/<triple>/<major>` (see #7535). I ended up using the following, somewhat defensive code that seems to work for all Julia versions (including pre-1.9 versions that do not have these libraries at all). If you find the time, it would be great if you could confirm that the build products indeed work for you.

@giordano Since all builds succeed now, it would be great if this could be merged.

PR #7900, opened 2024-01-03
[RocksDB] Add library

jeremiahpslewis:


R/RocksDB/build_tarballs.jl
```diff
@@ -0,0 +1,59 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd rocksdb
+mkdir build && cd build
+
+# build flags for arm
+if [[ "${target}" = arm* ]] || [[ "${target}" == aarch* ]]; then
+    export ARMCRC_SOURCE=1
+    export CXXFLAGS="-march=armv8-a+crc+crypto"
+    export CFLAGS="-march=armv8-a+crc+crypto"
```
imciner2, line 22: Why are you forcing these flags?

Also, the `arm*` platforms are actually armv7 and armv6, so you shouldn't try to use an armv8 architecture there.

R/RocksDB/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
+    DirectorySource("./bundled"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd rocksdb
+
+# Apply patch for arm and mingw
+if [[ "${target}" = arm* ]] || [[ "${target}" == aarch* ]]; then
```
imciner2, line 21: You are filtering out the armv7 and armv6 targets below, so you shouldn't need the first condition in this check. Also, the comment above this references mingw, so I guess that needs updating.

R/RocksDB/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+    -DWITH_GFLAGS=0 \
+    -DROCKSDB_BUILD_SHARED=1 \
+    -DPORTABLE=1 \
+    -DWITH_TOOLS=0 \
+    -DWITH_TESTS=0 \
+    -DWITH_BENCHMARK_TOOLS=0 \
+    -DCMAKE_BUILD_TYPE=Release ..
+
+cmake --build . --parallel ${nproc}
+cmake --install .
```
imciner2, line 50: > Unable to find valid license file in "${prefix}/share/licenses/RocksDB"

The auditor is saying it can't find a license in the installed files, so please add a call to `install_license` at the end of the script to install the license provided with the library.

R/RocksDB/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+    DirectorySource("./bundled"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd rocksdb
+
+# Apply patch for arm and mingw
+if [[ "${target}" = arm* ]] || [[ "${target}" == aarch* ]]; then
+    atomic_patch -p1 ../patches/arm64.patch
```
imciner2, line 22: I would change this to `${WORKSPACE}/srcdir/patches/arm64.patch` to be clearer.

R/RocksDB/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
+filter!(p -> arch(p) ∉ ("armv7l", "armv6l"), platforms)
+filter!(!Sys.iswindows, platforms)
+
+# The products that we will ensure are always built
+products = Product[
```
giordano, line 61:
```suggestion
products =[
```

R/RocksDB/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
+                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
+    DirectorySource("./bundled"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd rocksdb
+
+# Apply patch for arm and mingw
```
giordano, line 20: The reference to mingw in this comment doesn't match what the code below does.

PR #7911, opened 2024-01-05
[OpenBLAS] Update to v0.3.26

giordano:


giordano:
With https://github.com/OpenMathLib/OpenBLAS/pull/4394 Apple M series CPUs use NeoverseN1 kernels when OpenBLAS is compiled with `DYNAMIC_ARCH` (which is what we do here).  Some quick benchmarks on M1: with OpenBLAS 0.3.25:
```julia
julia> using LinearAlgebra, BenchmarkTools

julia> peakflops()
1.7049262964078418e11

julia> x = randn(1_000); y = randn(size(x));

julia> @btime dot($x, $y);
  402.705 ns (0 allocations: 0 bytes)
```
with OpenBLAS v0.3.26:
```julia
julia> using LinearAlgebra, BenchmarkTools

julia> peakflops()
1.8042546290642157e11

julia> x = randn(1_000); y = randn(size(x));

julia> @btime dot($x, $y);
  143.777 ns (0 allocations: 0 bytes)
```

PR #7917, opened 2024-01-07
[Tasmanian] - using recommended flags

MichelJuillard:
- O3
- link with strampoline
- link with OMP

### DIFF
--- a/T/Tasmanian/build_tarballs.jl
+++ b/T/Tasmanian/build_tarballs.jl
@@ -10,8 +10,6 @@ sources = [
     GitSource("https://github.com/ORNL/TASMANIAN.git", "10a762e036c58b2aee4dbf21137aff8401acf0a3")
 ]
 
-BLAS="blastrampoline"
-LAPACK="blastrampoline"
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/TASMANIAN
@@ -23,8 +21,6 @@ fi
 cmake -DCMAKE_INSTALL_PREFIX=$prefix \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
-    -DBLAS_LIBRARIES="-l${BLAS}" \
-    -DLAPACK_LIBRARIES="-l${LAPACK}" \
     -DTasmanian_ENABLE_RECOMMENDED=ON \
     -DTasmanian_ENABLE_PYTHON=OFF \
     ..
@@ -46,8 +42,10 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[
-    Dependency("libblastrampoline_jll"),
-    Dependency("CompilerSupportLibraries_jll"),
+    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
+    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
 ]
 
 # License file
--- a/T/Tasmanian/build_tarballs.jl
+++ b/T/Tasmanian/build_tarballs.jl
@@ -10,13 +10,23 @@ sources = [
     GitSource("https://github.com/ORNL/TASMANIAN.git", "10a762e036c58b2aee4dbf21137aff8401acf0a3")
 ]
 
+BLAS="blastrampoline"
+LAPACK="blastrampoline"
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/TASMANIAN
 mkdir build && cd build
+export CXXFLAGS=-Dsgemm_=sgemm_64
+if [[ "${target}" == *-freebsd* ]]; then
+    export LDFLAGS="-lpthread"
+fi
 cmake -DCMAKE_INSTALL_PREFIX=$prefix \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
+    -DBLAS_LIBRARIES="-l${BLAS}" \
+    -DLAPACK_LIBRARIES="-l${LAPACK}" \
+    -DTasmanian_ENABLE_RECOMMENDED=ON \
+    -DTasmanian_ENABLE_PYTHON=OFF \
     ..
 make -j${nproc}
 make install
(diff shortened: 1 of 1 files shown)


T/Tasmanian/build_tarballs.jl
```diff
@@ -35,6 +43,8 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[
+    Dependency("libblastrampoline_jll"),
+    Dependency("CompilerSupportLibraries_jll"),
```
imciner2, line 47:
```suggestion
    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
```
To use OpenMP, you need to use the LLVM OpenMP on BSD platforms, so you need this dependency scheme.

MichelJuillard:
Thank you very much Ian. Could I have found the information about FreeBSB and pthread anywhere?

imciner2:
Hmm, so it appears that the pthreads error is not related to that OpenMP depedency definition.

From what I can tell, it looks like the library itself uses OpenMP as its threading layer when requested, however some of the examples will always use `std::thread` for its threading. It looks like for some reason the FreeBSD compiler is not getting a dependency on pthreads added when it is building the `std::thread` code, so it is unable to link against that standard library construct.

I took a look inside the TASMANIAN repo to see what is happening, and it has this line https://github.com/ORNL/TASMANIAN/blob/6ff2a0ac55e0ec5b167a80c1bcb6ddc73098a35c/Config/CMakeIncludes/sanity_check_and_xsdk.cmake#L67C1-L70C8
```
# multi-threading is required by the Addons module even if OpenMP is not enabled
if (NOT Tasmanian_ENABLE_OPENMP)
    find_package(Threads REQUIRED)
endif()
```
which seems a bit confusing - it is saying in the comment that `Threads` is always required, but yet is not actually looking for it unless OpenMP is not used.

I also can't seem to find where it adds the threading library as a dependency inside the Addons file here https://github.com/ORNL/TASMANIAN/blob/6ff2a0ac55e0ec5b167a80c1bcb6ddc73098a35c/Addons/CMakeLists.txt, even though it seems it would always be required. I do see it being added as a dependency for the core library when OpenMP is disabled here though https://github.com/ORNL/TASMANIAN/blob/6ff2a0ac55e0ec5b167a80c1bcb6ddc73098a35c/Config/CMakeIncludes/add_dependencies_target.cmake#L21C62-L21C84, which then transitively applies it to the addons I believe.
 
This is looking like an upstream bug where they need to fixup the CMake processing to add the `${CMAKE_THREAD_LIBS_INIT}` variable to the addons target and always search for `Threads` as the comment says.

MichelJuillard:
> Hmm, so it appears that the pthreads error is not related to that OpenMP depedency definition.
> 
> From what I can tell, it looks like the library itself uses OpenMP as its threading layer when requested, however some of the examples will always use `std::thread` for its threading. It looks like for some reason the FreeBSD compiler is not getting a dependency on pthreads added when it is building the `std::thread` code, so it is unable to link against that standard library construct.
> 
> I took a look inside the TASMANIAN repo to see what is happening, and it has this line https://github.com/ORNL/TASMANIAN/blob/6ff2a0ac55e0ec5b167a80c1bcb6ddc73098a35c/Config/CMakeIncludes/sanity_check_and_xsdk.cmake#L67C1-L70C8
> 
> ```
> # multi-threading is required by the Addons module even if OpenMP is not enabled
> if (NOT Tasmanian_ENABLE_OPENMP)
>     find_package(Threads REQUIRED)
> endif()
> ```
> 
> which seems a bit confusing - it is saying in the comment that `Threads` is always required, but yet is not actually looking for it unless OpenMP is not used.
> 
> I also can't seem to find where it adds the threading library as a dependency inside the Addons file here https://github.com/ORNL/TASMANIAN/blob/6ff2a0ac55e0ec5b167a80c1bcb6ddc73098a35c/Addons/CMakeLists.txt, even though it seems it would always be required. I do see it being added as a dependency for the core library when OpenMP is disabled here though https://github.com/ORNL/TASMANIAN/blob/6ff2a0ac55e0ec5b167a80c1bcb6ddc73098a35c/Config/CMakeIncludes/add_dependencies_target.cmake#L21C62-L21C84, which then transitively applies it to the addons I believe.
> 
> This is looking like an upstream bug where they need to fixup the CMake processing to add the `${CMAKE_THREAD_LIBS_INIT}` variable to the addons target and always search for `Threads` as the comment says.

I added `LDFLAGS="-lpthread"` as in https://github.com/JuliaPackaging/Yggdrasil/blob/master/A/Arpack/build_tarballs.jl and it seems to do the trick

imciner2:
It looks like there are a lot of other commits in this PR now. Can you rebase your changes onto master and force push onto this branch to clean that up.

ViralBShah:
This PR is pulling in all sorts of other commits.

MichelJuillard:
Oops! I have squashed my commits and rebased

imciner2:
I think you lost the `LLVMOpenMP_jll` dependency configuration in the most recent rebase. (And actually, I am surprised the FreeBSD build worked in that case).

MichelJuillard:
Yes, as adding the `LLVOpenMP_jll` dependency didn't fix the   `pthread` error and as Arpack doesn't use it, I wanted to test if it was necessary for `Tasmanian`. Ideally one should test the product but I don't have access to a FreeBSD box

giordano:
> And actually, I am surprised the FreeBSD build worked in that case

FreeBSD sdk may have libomp already, the main problem is macOS which doesn't come with any openmp library at all.

MichelJuillard:
@giordano How come then that the build works also for macos?

imciner2:
It looks like macOS isn't being told to use OpenMP:

The build log has https://buildkite.com/julialang/yggdrasil/builds/7659#018ced16-1e07-436f-bfc0-9d3c3d537465/620-788:
```
07:20:38] -- Tasmanian 8.0: summary of build options
[07:20:38] --  -D CMAKE_BUILD_TYPE:STRING=Release
[07:20:38] --  -D CMAKE_INSTALL_PREFIX:PATH=/workspace/destdir
[07:20:38] --  -D CMAKE_CXX_FLAGS:STRING=-Dsgemm_=sgemm_64
[07:20:38] --  -D BUILD_SHARED_LIBS=ON
[07:20:38] --  -D Tasmanian_ENABLE_OPENMP:BOOL=FALSE
[07:20:38] --  -D Tasmanian_ENABLE_BLAS:BOOL=OFF
[07:20:38] --  -D Tasmanian_ENABLE_MPI:BOOL=OFF
[07:20:38] --  -D Tasmanian_ENABLE_PYTHON:BOOL=OFF
[07:20:38] --  -D Tasmanian_ENABLE_CUDA:BOOL=OFF
[07:20:38] --  -D Tasmanian_ENABLE_MAGMA:BOOL=OFF
[07:20:38] --  -D Tasmanian_ENABLE_HIP:BOOL=OFF
[07:20:38] --  -D Tasmanian_ENABLE_DPCPP:BOOL=OFF
[07:20:38] --  -D Tasmanian_ENABLE_SWIG:BOOL=OFF
[07:20:38] --  -D Tasmanian_ENABLE_FORTRAN:BOOL=OFF
[07:20:38] --  -D Tasmanian_ENABLE_DOXYGEN:BOOL=OFF
```

MichelJuillard:
Thanks Ian. Yes, if it doesn't find the corresponding library, Cmake just turns off the feature. I will check all the tests

PR #7919, opened 2024-01-08
[P4est] Fix linking to MSMPI when targeting Windows

giordano:
The old recipe was creating a broken symlink, a better solution is to tell the linker the name of the library to search (and where).  Ref: https://github.com/JuliaIO/Tar.jl/pull/165#issuecomment-1880295075.

CC: @JoshuaLampert @sloede


JoshuaLampert:
Nice. thanks @giordano!

giordano:
Uhm, compilation of all executables (some of which do link to msmpi) is successful, but libraries aren't compiled.  Ok, debugging this will be more complicated than what I hoped.

PR #7921, opened 2024-01-08
[llama_cpp] Update to v0.0.15 (b1796, 2024-01-09)

svilupp:
- Update Llama.cpp to version v1794
- New functionality: 2bit SOTA quantization


svilupp:
It seems that the new 2-bit quantization functionality causes problems for arm 6/7.

svilupp:
I've tried different GCCs but no luck.

Unlike previously, there is no way to disable quants at compilation (LLAMA_K_QUANTS=OFF), so they might left behind until it's fixed upstream.

I've opened the issue here: https://github.com/ggerganov/llama.cpp/issues/4827

L/llama_cpp/build_tarballs.jl
```diff
@@ -50,11 +51,12 @@ version = v"0.0.14"  # fake version number
 # 0.0.12          2023-07-24       master-41c6741    https://github.com/ggerganov/llama.cpp/releases/tag/master-41c6741
 # 0.0.13          2023-07-29       master-11f3ca0    https://github.com/ggerganov/llama.cpp/releases/tag/master-11f3ca0
 # 0.0.14          2024-01-04       b1767             https://github.com/ggerganov/llama.cpp/releases/tag/b1767
+# 0.0.15          2024-01-08       b1794             https://github.com/ggerganov/llama.cpp/releases/tag/b1794
 
 
 sources = [
     GitSource("https://github.com/ggerganov/llama.cpp.git",
-        "012cf349aec8ffb47c9def5dc018240fa3721e8b"),
+        "1fc2f265ff9377a37fd2c61eae9cd813a3491bea"),
```
giordano, line 59:
```suggestion
        "27afe2992741d559fe074cf609c95ba65a124bcb"),
```

svilupp:
Arm67 included back in. All passes.
Updated to the latest tag.

We should be good to merge.

PR #7925, opened 2024-01-09
[elmerfem] New package v9.0.20231229

tpgillam:
New (unofficial) build recipe for [elmerfem](https://github.com/ElmerCSC/elmerfem/)

Currently this will only build artifacts for certain Linux systems. Hopefully someone can help fix the issues on Windows & MacOS & FreeBSD in due course.

### DIFF
--- a/E/elmerfem/build_tarballs.jl
+++ b/E/elmerfem/build_tarballs.jl
@@ -79,12 +79,18 @@ install_license ../elmerfem/LICENSE.md
 #   
 #   The linux cmake_install.cmake doesn't include this line at all, so perhaps we can just remove it
 #   before running `make`...?
+# XXX: build fails on FreeBSD with compilation error... suggests we need to use an older C++ standard:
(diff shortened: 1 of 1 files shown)


E/elmerfem/build_tarballs.jl
```diff
@@ -0,0 +1,151 @@
...
+build_tarballs(
+    ARGS,
+    name,
+    version,
+    sources,
+    script,
+    platforms,
+    products,
+    dependencies;
+    julia_compat="1.6",
```
giordano, line 148: libblastrampoline 5.4 requires julia v1.9: https://github.com/JuliaLang/julia/blob/v1.9.0/stdlib/libblastrampoline_jll/Project.toml (1.8 had 5.1.1: https://github.com/JuliaLang/julia/blob/v1.8.0/stdlib/libblastrampoline_jll/Project.toml).  Are you fine with requiring julia v1.9?
```suggestion
    julia_compat="1.9",
```
Otherwise supporting different versions of libblastrampoline before 5.4 is a pain.

tpgillam: Yep Julia >=1.9 is fine for us, thanks!

giordano:
Thanks for your contribution!

PR #7926, opened 2024-01-09
[iso_codes] Copy symlinks to avoid installation issues.

nhz2:
This should fix the Windows installation issue.
The tarball is about 6 MB larger. I have tested this version locally with the automated tests in `Gtk.jl` and `Countries.jl` and they both passed, though I don't know what those packages are doing.
The new artifact tree hash is `1719102ef324df43b9fe123a231c2ee630931a03` which matches the tree hash of the old artifact if it is extracted with `copy_symlinks=true`

### DIFF
--- a/I/iso_codes/build_tarballs.jl
+++ b/I/iso_codes/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder
 
 name = "iso_codes"
-version = v"4.15.0"
+version = v"4.15.1"
 
 # the git tag used for versioning has changed format
 if version < v"4.8"
@@ -12,6 +12,8 @@ if version < v"4.8"
     else
         tag = "iso-codes-$version"
     end
+elseif version == v"4.15.1" # for fake patch version to fix windows install.
+    tag = "v4.15.0"
 else
     tag = "v$version"
 end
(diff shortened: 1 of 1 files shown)


mkitti:
Should we do this for everyone? Or should we make a Windows specific version?

giordano:
Yeah, I'd prefer doing it only for Windows

vtjnash (commented):
Sgtm

staticfloat:
Agreed, I'd rather do this only for windows.

nhz2:
I updated the build to only copy symlinks on Windows. I also set `julia_compat="1.6"` to build on all platforms.

I/iso_codes/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder
 
 name = "iso_codes"
-version = v"4.15.0"
+version = v"4.15.1"
```
mkitti, line 6: @giordano do we still need to do this version bump or will the build number take care of the issue?

I/iso_codes/build_tarballs.jl
```diff
@@ -34,7 +36,7 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = [AnyPlatform()]
+platforms = supported_platforms(; experimental=true)
```
giordano, line 39: `experimental` does nothing, it's no-op.

@staticfloat would
```julia
[AnyPlatform(); supported_platforms(; exclude=!Sys.iswindows)]
```
do the right thing, instead of having 14 identical artifacts?

I/iso_codes/build_tarballs.jl
giordano: If we change the julia compat below we must have a different version number

I/iso_codes/build_tarballs.jl
staticfloat: Sadly, if you have a platformless binding in an `Artifacts.toml` file, you can't also have other platform-specific bindings in that same `Artifacts.toml` file.  So the best we can do here is to just have the same tarball listed for a bunch of different platforms, then a different tarball for the windows platforms.

mkitti (approved):

PR #7931, opened 2024-01-11
[SuiteSparse] Update to 7.5.0

imciner2:
Update to the recently released SuiteSparse 7.5.0.

This allows for us to drop the build system patch, since that was already upstream, so now we link against blastrampoline properly with no patches. Looking through the Windows build logs, I did see the warning
```
<command-line>:0:20: warning: ISO C99 requires whitespace after the macro name [enabled by default]
```
which ends up being because SuiteSparse tries to make a define for `-DBLAS_blastrampoline-5`, and the `-` is an illegal character. I modified it so the library name is passed in without the version suffix to the define, but the linkers still get the suffix, so this command line warning is gone now.

Also, @rayegun, it looks like this update will be good for your work to add the Float32 support to the SuiteSparse wrapper:

> CHOLMOD 5.1.1 appears in SuiteSparse 7.5.0 (a stable release). It's essentially finished, in terms of support for single precision. It could use a method for creating a copy of a sparse matrix, dense matrix, triplet matrix, or factorization, while at the same time converting between single and double precision (see issue https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/720, which I'll consider for a future version of CHOLMOD), but otherwise it's in good shape.


rayegun:
Great thank you. I will also look at the CUDA version so we can add function stubs. I think I can use Cyclops to test

PR #7951, opened 2024-01-14
Add mlir-jl-tblgen

vchuravy:


M/mlir_jl_tblgen/build_tarballs.jl
```diff
@@ -0,0 +1,117 @@
...
+        augment_llvm!(platform)
+    end"""
+
+# determine exactly which tarballs we should build
+builds = []
+for llvm_version in llvm_versions, llvm_assertions in (false, true)
+    # Dependencies that must be installed before this package can be built
+    llvm_name = llvm_assertions ? "LLVM_full_assert_jll" : "LLVM_full_jll"
+    dependencies = [
+        BuildDependency(PackageSpec(name=llvm_name, version=llvm_version))
```
vchuravy, line 67: Do we need to add MLIR_jll here?

M/mlir_jl_tblgen/build_tarballs.jl
```diff
@@ -0,0 +1,117 @@
...
+include(joinpath(YGGDRASIL_DIR, "platforms", "llvm.jl"))
+
+name = "MLIR_JL_TBLGEN"
+repo = "https://github.com/JuliaLabs/MLIR.jl.git"
+version = v"0.0.1"
+
+llvm_versions = [v"14.0.6", v"15.0.7", v"16.0.6"]
+
+sources = [
+    GitSource(repo, "fb6673ba60c1a82582b286eb954ad759659558d5")
```
vchuravy, line 15:
```suggestion
    GitSource(repo, "ff7e7e9f342b5fd1932efbee506d19bec6cf3159")
```

M/mlir_jl_tblgen/build_tarballs.jl
mofeing: The MLIR lib and headers are already in `LLVM_full_jll`, so... nope

M/mlir_jl_tblgen/build_tarballs.jl
```diff
@@ -0,0 +1,117 @@
...
+
+llvm_versions = [v"14.0.6", v"15.0.7", v"16.0.6"]
+
+sources = [
+    GitSource(repo, "ff7e7e9f342b5fd1932efbee506d19bec6cf3159")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd MLIR.jl/deps/tblgen
```
mofeing, line 20: Shouldn't we move it to another dir? We can do it later, just an opinion

vchuravy: I don't think there is a need for it, we could make it follow the structure a bit more, but right now that seems over the top for two cc files.

M/mlir_jl_tblgen/build_tarballs.jl
```diff
@@ -0,0 +1,117 @@
...
+non_platform_ARGS = filter(arg -> startswith(arg, "--"), ARGS)
+
+# `--register` should only be passed to the latest `build_tarballs` invocation
+non_reg_ARGS = filter(arg -> arg != "--register", non_platform_ARGS)
+
+for (i,build) in enumerate(builds)
+    build_tarballs(i == lastindex(builds) ? non_platform_ARGS : non_reg_ARGS,
+                   name, version, build.sources, script,
+                   build.platforms, build.products, build.dependencies;
+                   preferred_gcc_version=v"8", julia_compat="1.6",
```
giordano, line 114: We may need to compile with gcc 10 to match what we used to build llvm 16 (which is currently failing here for the platforms where we use gcc): https://github.com/JuliaPackaging/Yggdrasil/blob/b7f5e3c48f292078bbed4c9fdad071da7875c0bc/L/LLVM/LLVM_full%4016/build_tarballs.jl#L6
```suggestion
                   preferred_gcc_version=v"10", julia_compat="1.6",
```

PR #7969, opened 2024-01-18
update devcontainer

M-PERSIC:
- Replace the Dockerfile image with the [Julia development container feature](https://github.com/JuliaLang/devcontainer-features.git)
- Set one project environment (.ci)
- New `privileged` and `hostRequirements.gpu` properties

A few notes:
- We can keep a separate project environment for devcontainers if desired
- With Codespaces,  xz, bzip2, etc. are already included ([common-utils](https://github.com/devcontainers/features/pkgs/container/features%2Fcommon-utils) feature)
- The julia vscode extension is included with the feature (along with even-better-toml :) )
- I've kept the same Julia version (v1.7.3), not sure if this should be bumped or kept as is
- This devcontainer config works best with Codespaces. On a local machine, one would need to specify the image as well as include the common-utils feature.  I can create a second devcontainer config better suited for local development if desired


vchuravy:
> I've kept the same Julia version (v1.7.3), not sure if this should be bumped or kept as is

Needs to be kept as is.

.devcontainer/devcontainer.json
```diff
@@ -1,8 +1,15 @@
...
-    "runArgs": ["--privileged"],
-    "dockerFile": "Dockerfile",
-    "onCreateCommand": "julia --project=.ci -e 'using Pkg; Pkg.instantiate(); Pkg.precompile();'"
+  "features": {
+    "ghcr.io/julialang/devcontainer-features/julia:1": {
+      "channel": "1.7"
+    },
+  },
+  "containerEnv": {
+    "JULIA_PROJECT": ".ci"
```
giordano, line 8: This has to be an absolute path, otherwise if you move from the default directory `.ci` will not exist.

stemann:
Why favor Codespaces?

I would personally only run locally.

giordano:
Personally I use codespaces a lot when I cannot run BinaryBuilder locally. Locally I don't need the devcontainer to start with, but maybe it's just me

stemann:
I would definitely support Mose’s use case over mine :-)

M-PERSIC:
@stemann it's not so much favouring Codespaces as it is dealing with how Codespaces handles certain properties implicitly. For example, if we were to run the same devcontainer config locally we would need, at a minimum, to specify both the image and common-utils feature:
```
"image": "ubuntu:latest",
...
"features": {
    "ghcr.io/devcontainers/features/common-utils:2": {}
}
```
It's more of an annoyance than anything to have to re-include these lines when developing on our own machines.

stemann:
Understood.

I use dev. containers extensively - but “only” locally or via SSH.

As Mose, I would probably not use the **Yggdrasil** dev. container locally, but I guess it would be very convenient for e.g. Windows users to have a dev container setup ready to go.

I would not assume that someone would bother tweaking the dev. container spec. for a local setup.

My suggestion is to either have a universal setup or have two dev. container specs  - it sounds like that is an option.

But that is just my personal preference - feel free to disregard.

M-PERSIC:
That's what I wrote in my pr description :) I'll take a stab at some different configs tonight

.devcontainer/cuda/devcontainer.json
giordano, line 1: Why do we need this here?

M-PERSIC: Do you mean within the cuda directory or the devcontainer itself?

giordano: I was referring to the cuda devcontainer, didn't feel like it fits in here, but if it's for testing until the PR is ready I guess it's fine and we keep this conversation open as a reminder

M-PERSIC:
For now, I am satisfied with an universal devcontainer config. I would like to revisit this again once Codespaces begins offering GPU machines/we can bump the Julia version as there are some minute improvements in mind, however if there are no objections than I feel this is good to go!

PR #7974, opened 2024-01-18
[Sundials] Upgrade to v6.7.0

staticfloat:
Also, require Julia v1.11+, since Julia v1.11 has a new SuiteSparse version.  This will be the first version that supports Julia v1.11.

### DIFF
--- a/S/Sundials/Sundials@6/build_tarballs.jl
+++ b/S/Sundials/Sundials@6/build_tarballs.jl
@@ -1,12 +1,12 @@
 using BinaryBuilder
 
 name = "Sundials"
-version = v"6.6.0"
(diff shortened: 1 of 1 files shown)


ViralBShah:
The real challenge is to upgrade the wrappers in Sundials.jl. My latest effort is in: https://github.com/SciML/Sundials.jl/pull/415

staticfloat:
Oh, I didn't realize that `Sundials.jl` does not yet work with Sundials v6+.  Okay, this is less useful then.

imciner2:
We can still merge this and at least get the JLLs built so you can continue working on that. Is there anything else you need to do here, or is it ready to merge?

PR #7976, opened 2024-01-18
Wizard recipe: OpenFOAM_com-v2312.0.0

thealanjason:

* Package name: OpenFOAM_com
* Version: v2312.0.0


ViralBShah:
That's one impressive build script!

O/OpenFOAM_com/bundled/patches/bashrc-compilerflags.patch
```diff
@@ -0,0 +1,88 @@
...
+diff --git before/openfoam/wmake/rules/General/Gcc/c after/openfoam/wmake/rules/General/Gcc/c
+index e43a85b..0db22f1 100644
+--- before/openfoam/wmake/rules/General/Gcc/c
++++ after/openfoam/wmake/rules/General/Gcc/c
+@@ -3,7 +3,7 @@
+ #------------------------------------------------------------------------------
+ SUFFIXES   += .c
+ 
+-cc         := gcc$(COMPILER_VERSION)
++cc         := gcc$(COMPILER_VERSION) -Wl,-rpath-link=/workspace/srcdir/openfoam/platforms/linux64GccDPInt32Opt/lib -Wl,-rpath-link=/workspace/srcdir/openfoam/platforms/linux64GccDPInt32Opt/lib/dummy -Wl,-rpath-link=/workspace/srcdir/openfoam/platforms/linux64GccDPInt32Opt/lib/mpi-system
```
giordano, line 72: These rpaths don't seem correct as these directories won't exist at runtime.

thealanjason: I took inspiration form [this](https://github.com/JuliaPackaging/Yggdrasil/blob/ac0620361ec81f0ed398641b494475100713c448/O/OpenFOAM/build_tarballs.jl#L34C1-L34C51). In the .org version of OpenFOAM the directory structure and the commands are a bit different. Since I couldn't get it working with `sed` I directly patched the files based on what would be the result if I ran the linked command along with the OpenFOAM_org. Without this, I had linker errors. 

Would it make sense to set the `-rpath` as `${libdir}` instead?

thealanjason:
Imagine a case where the JLL package needs to have many more `ExecutableProduct`s or `LibraryProduct`s in a newer version. Is there a way for track down the newly generated ones and automatically update the `build_tarballs.jl`? Or do we have to manually add these? 

I would like to add two of the submodules, `avalanche` and `cfmesh`, [from here](https://develop.openfoam.com/Development/openfoam/-/tree/master/modules?ref_type=heads) because these are the ones I'm most interested in. If I add these to the project, there will be a large number of new libraries and binaries. It won't be fun to track them down manually and add to the `build_tarballs.jl`

thealanjason:
> Imagine a case where the JLL package needs to have many more `ExecutableProduct`s or `LibraryProduct`s in a newer version. Is there a way for track down the newly generated ones and automatically update the `build_tarballs.jl`? Or do we have to manually add these?

Answering my own question, I did it manually. For this, in the log files of the buildkite CI, I searched for the strings "Checking bin/" and "Checking lib/". Through these, I found the list of all the built binaries and libraries.

O/OpenFOAM_com/bundled/patches/bashrc-compilerflags.patch
thealanjason: For now I've added `${libdir}` to `-rpath-link`

thealanjason:
For the record, this is my first collaborative open source contribution. I'll be very happy when this is merged :)

O/OpenFOAM_com/build_tarballs.jl
```diff
@@ -0,0 +1,577 @@
...
+## For linux64
+LDFLAGS=""
+for dir in "" "/dummy" "/mpi-system"; do
+    LDFLAGS="${LDFLAGS} -Wl,-rpath-link=${PWD}/platforms/linux64GccDPInt32Opt/lib${dir}"
+done
+LDFLAGS="${LDFLAGS} -Wl,-rpath-link=${libdir}"
+
+# Set rpath-link in all C/C++ compilers
+sed -i "s|cc         := gcc\$(COMPILER_VERSION)|cc         := gcc\$(COMPILER_VERSION) ${LDFLAGS}|" wmake/rules/General/Gcc/c
+sed -i "s|CC         := g++\$(COMPILER_VERSION) -std=c++14|CC         := g++\$(COMPILER_VERSION) -std=c++14 ${LDFLAGS}|" wmake/rules/General/Gcc/c++
```
giordano, line 37: It'd be better to use `cc`/`c++` as compilers, instead of forcing `gcc`/`g++`:
```suggestion
sed -i "s|cc         := gcc\$(COMPILER_VERSION)|cc         := cc\$(COMPILER_VERSION) ${LDFLAGS}|" wmake/rules/General/Gcc/c
sed -i "s|CC         := g++\$(COMPILER_VERSION) -std=c++14|CC         := c++\$(COMPILER_VERSION) -std=c++14 ${LDFLAGS}|" wmake/rules/General/Gcc/c++
```

O/OpenFOAM_com/build_tarballs.jl
```diff
@@ -0,0 +1,577 @@
...
+dependencies = [
+    Dependency(PackageSpec(name="flex_jll", uuid="48a596b8-cc7a-5e48-b182-65f75e8595d0"))
+    Dependency(PackageSpec(name="SCOTCH_jll", uuid="a8d0f55d-b80e-548d-aff6-1a04c175f0f9"); compat=SCOTCH_VERSION)
+    Dependency(PackageSpec(name="FFTW_jll", uuid="f5851436-0d7a-5f13-b9de-f02708fd171a"); compat=FFTW_VERSION)
+    Dependency(PackageSpec(name="PTSCOTCH_jll", uuid="b3ec0f5a-9838-5c9b-9e77-5f2c6a4b089f"))
+    Dependency(PackageSpec(name="METIS_jll", uuid="d00139f3-1899-568f-a2f0-47f597d42d70"))
+    Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a"))
+    Dependency(PackageSpec(name="CGAL_jll", uuid="8fcd9439-76b0-55f4-a525-bad0597c05d8"))
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    RuntimeDependency(PackageSpec(name = "M4_jll", uuid = "9051c120-a745-5e86-aaa7-8cbc404dba28"))
```
giordano, line 571: What is m4 needed for at runtime? :thinking:

O/OpenFOAM_com/build_tarballs.jl
```diff
@@ -0,0 +1,577 @@
...
+    ExecutableProduct("XiFoam", :XiFoam),
+    ExecutableProduct("zipUpMesh", :zipUpMesh),
+    
+    FileProduct("share/openfoam/etc", :openfoam_etc), 
+    FileProduct("share/openfoam/bin", :openfoam_bin)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="flex_jll", uuid="48a596b8-cc7a-5e48-b182-65f75e8595d0"))
```
giordano, line 563: How is flex used here?  It's usually a `HostBuildDependency`, because it's a tool used only during the build, and has to run on the host, does openfoam does something special with it at runtime?

O/OpenFOAM_com/build_tarballs.jl
```diff
@@ -0,0 +1,577 @@
...
+    Dependency(PackageSpec(name="SCOTCH_jll", uuid="a8d0f55d-b80e-548d-aff6-1a04c175f0f9"); compat=SCOTCH_VERSION)
+    Dependency(PackageSpec(name="FFTW_jll", uuid="f5851436-0d7a-5f13-b9de-f02708fd171a"); compat=FFTW_VERSION)
+    Dependency(PackageSpec(name="PTSCOTCH_jll", uuid="b3ec0f5a-9838-5c9b-9e77-5f2c6a4b089f"))
+    Dependency(PackageSpec(name="METIS_jll", uuid="d00139f3-1899-568f-a2f0-47f597d42d70"))
+    Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a"))
+    Dependency(PackageSpec(name="CGAL_jll", uuid="8fcd9439-76b0-55f4-a525-bad0597c05d8"))
```
giordano, line 569: `CGAL` is a header-only package now, it's unlikely you need it at runtime?

O/OpenFOAM_com/build_tarballs.jl
```diff
@@ -0,0 +1,577 @@
...
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    RuntimeDependency(PackageSpec(name = "M4_jll", uuid = "9051c120-a745-5e86-aaa7-8cbc404dba28"))
+]
+append!(dependencies, platform_dependencies)
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; 
+    preferred_gcc_version = v"12")
```
giordano, line 577: https://github.com/JuliaPackaging/Yggdrasil/blob/f06c3a47a4773cdeaa3bd74f7930746e4a8cc553/CONTRIBUTING.md#compatibility-tips

O/OpenFOAM_com/build_tarballs.jl
thealanjason: I remember seeing some tutorials in the repository of openfoam using the m4 command, but cannot find it now. So I add it as a dependency. I could take it out for now, and add it later if needed

O/OpenFOAM_com/build_tarballs.jl
thealanjason: Okay. I shall try with v"8", because the minimum required `gcc` version is 7.5.0. In that case the 7.1.0 version in BinaryBuilder will not work.

thealanjason (commented):
I've made the changes suggested.

O/OpenFOAM_com/build_tarballs.jl
thealanjason: When I change it to `HostBuildDependency` I get the following error:
`[23:16:27] /workspace/srcdir/openfoam/build/linux64GccDPInt32Opt/src/fileFormats/stl/STLAsciiParseFlex.L.C:319:10: fatal error: FlexLexer.h: No such file or directory
[23:16:27]   319 | 
[23:16:27]       |          ^            
[23:16:27] compilation terminated.`
When it is a normal `Dependency` it works fine.

O/OpenFOAM_com/build_tarballs.jl
```diff
@@ -0,0 +1,576 @@
...
+    ExecutableProduct("XiFoam", :XiFoam),
+    ExecutableProduct("zipUpMesh", :zipUpMesh),
+    
+    FileProduct("share/openfoam/etc", :openfoam_etc), 
+    FileProduct("share/openfoam/bin", :openfoam_bin)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency(PackageSpec(name="flex_jll", uuid="48a596b8-cc7a-5e48-b182-65f75e8595d0"))
```
giordano, line 563: To reiterate the question in https://github.com/JuliaPackaging/Yggdrasil/pull/7976#discussion_r1478728642, how's flex used?  Is it needed at runtime?  Or only at build time?  If you mark it as `BuildDependency` and then it's needed at runtime, no one will tell you until you need it and things crash badly.

PR #7978, opened 2024-01-19
Sundials 5: Update SuiteSparse dep

ViralBShah:


staticfloat:
The problem is that SuiteSparse v7.3 is incompatible with SuiteSparse v7.2, so if you build a version of Sundials against SuiteSparse v7.3, it won't work with SuiteSparse v7.2.  So I think this bound was purposeful, as it matches the version of SS shipped with a certain Julia version.

ViralBShah:
So should we bump the Sundials version to 5.2.3, and use Julia 7.5 that will ship with Julia 1.11?

Keno:
> So should we bump the Sundials version to 5.2.3, and use Julia 7.5 that will ship with Julia 1.11?

Yes

ViralBShah:
Can someone recommend how to make the build process find klu.h? It is part of `SuiteSparse_jll`, but we don't copy the header files when we include the JLL in the Julia build.

PR #7980, opened 2024-01-19
[FLINT] redo the patch backport for 2.9 with llvm 13 (like the old builds)

benlorenz:
With the new llvm16 default we have some weird errors on x86_64 darwin.
My test-rebuild with llvm13 seems to work fine again.

cc: @thofma


thofma (approved):
I have not checked it locally, but I am happy to hear that you tested it.

PR #7983, opened 2024-01-19
openfhe_julia: fix version for libcxxwrap_julia to avoid bug

sloede:
OpenFHE-julia does not work properly with libcxxwrap-julia v0.12, which has been released 3 days ago (see https://github.com/JuliaInterop/libcxxwrap-julia/issues/143). We therefore build explicitly against the latest version known to work, v0.11.2, until a fix has been released.

@giordano I hope I specified the version correctly and it is not interpreted as a compat setting (with semver upgradeability) but as an exact version.

### DIFF
--- a/O/openfhe_julia/build_tarballs.jl
+++ b/O/openfhe_julia/build_tarballs.jl
@@ -78,7 +78,7 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     BuildDependency(PackageSpec(;name="libjulia_jll", version=v"1.10.7")),
-    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"); compat="0.11"),
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"); compat="0.11.2"),
     Dependency(PackageSpec(name="OpenFHE_jll", uuid="a2687184-f17b-54bc-b2bb-b849352af807")),
     # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
     # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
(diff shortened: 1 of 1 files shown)


O/openfhe_julia/build_tarballs.jl
```diff
@@ -78,7 +78,7 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     BuildDependency(PackageSpec(;name="libjulia_jll", version=v"1.10.7")),
-    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7")),
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7", version=v"0.11.2")),
```
giordano, line 81: Why not using compat instead? I always find it dodgy to specify the version for building abut not the compat: if the build version is important it's very likely compat would be very useful too

sloede: Sorry, I didn't know how to specify the compat version essentially I want something like `"< 0.12“`. Would that be sufficient to set as version or is the correct keyword `compat`?

sloede: Let's try it...

O/openfhe_julia/build_tarballs.jl
```diff
@@ -78,7 +78,7 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     BuildDependency(PackageSpec(;name="libjulia_jll", version=v"1.10.7")),
-    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7")),
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"); compat="0.11"),
```
sloede, line 81:
```suggestion
    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"); compat="0.11.2"),
```
It seems like it's not like a regular compat bound, thus I need to specify the version I want precisely.

sloede:
@giordano this seems to build now as desired. Sorry for the noise!

PR #7989, opened 2024-01-22
[PETSc] fixes for 3.18.7

boriskaus:
It turns out that PETSc 3.18.6/3.18.7 have a number of problems that prevent it from running on Julia 1.10 and on windows machines, even when they can be compiled fine. I summarise them below:

#### *1. MPI is broken on Windows* 
PETSc 3.18.6, released in May 2023, works fine on Julia 1.9 and passes [tests](https://github.com/boriskaus/test_PETSc_jll/actions/runs/7444942534). 
If I, however, recompile the same 3.18.6 (while making sure to use the same compiler/package versions for everything), it compiles fine but gives the following [runtime error](https://github.com/boriskaus/test_PETSc_jll/actions/runs/7594048475/job/20685150138#step:7:220):
```
Allocations: 4998341 (Pool: 4993543; Big: 4798); GC: 9
Mingw-w64 runtime failure:
32 bit pseudo relocation at 00000000090C136A out of range, targeting 00007FFCF4376530, yielding the value 00007FFCEB2B51C2.
ERROR: LoadError: Package test_PETSc_jll errored during testing (exit code: 3)
Stacktrace:
```
This is mysterious; a comparison between the log files suggests that they use the same versions for everything. The same was found in [HDF5_jll](https://github.com/eschnett/Yggdrasil/pull/6) where it was solved by deactivating MPI support.

Interesting, a recent [SuperLU_dist](https://github.com/boriskaus/test_SuperLU_DIST_jll/actions/runs/7422000918) build does work fine with MPI support on windows, a workaround is to deactivate MPI for windows.

Since this breaks the code for windows users (as well as downstream packages such as LaMEM), and I failed to find a solution for this during the last few weeks.

#### *2. SuiteSparse never worked* 
The previous versions of PETSc shipped with SuiteSparse support. Yet, that was not tested thoroughly enough. It turns out that this never worked, most likely because it requires a very specific version of SuiteSparse (5.13.0), for which there is currently no Julia build.  See [here](https://github.com/boriskaus/test_PETSc_jll/actions/runs/7602965554).
There are a few issues with that: 
a) SuiteSparse is notoriously difficult to compile
b) Nobody complained about this so far (so I suppose it wasn't used)
c) I am not allowed to use the PETSc build system to download and install the correct versions of SuiteSparse and ship it along with PETSc (for example under a modified name to avoid name collision with julia).
The simplest way forward is therefore to remove support. If there is a future version of PETSc that is compatible with an available version of SuiteSparse in the Julia ecosystem we can out it back again.

The proposed solution is to compile PETSc without MPI support on windows, and remove SuiteSparse support.

A local build of this was [tested succesfully on various systems](https://github.com/boriskaus/test_PETSc_jll/actions/runs/7613537765).


ViralBShah:
SuiteSparse is unfortunately hard since you are forced to use the version that Julia ships, and we can't change it in the near future.

Is it possible to compile Petsc with LBT as the BLAS rather than directly linking to openblas?

boriskaus:
I tried LBT, which caused the same MPI runtime issues that have been driving me nuts for the last few weeks. I can give it another shot though.

What about downloading and compiling suitesparse through the PETSc system and renaming the libraries so it won't clash (`libumfpack_petsc` for example)?

ViralBShah:
I don't think that changing the names is sufficient. What could work is if you can build SuiteSparse as static libraries and link them statically into the petsc shared libs.

Building windows without MPI for now would make sense.

It would be nice to see if we can use LBT. I see you just pushed a commit.

boriskaus:
> I don't think that changing the names is sufficient. What could work is if you can build SuiteSparse as static libraries and link them statically into the petsc shared libs.

I see. I'm not sure whether that can be done with petsc (perhaps something for an upcoming release if it works).

boriskaus:
> It would be nice to see if we can use LBT. I see you just pushed a commit for LBT, but that seems like it is only for Apple ARM.
yes, my mistake. Let me try locally.

P/PETSc/build_tarballs.jl
```diff
@@ -326,6 +361,8 @@ ENV["MPITRAMPOLINE_DELAY_INIT"] = "1"
...
 # NOTE: llvm16 seems to have an issue with PETSc 3.18.x as on apple architectures it doesn't know how to create dynamic libraries  
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               augment_platform_block, julia_compat="1.10", 
-               preferred_gcc_version = v"9",  
-               preferred_llvm_version=v"13")
+               augment_platform_block, 
+               julia_compat="1.6", 
+               preferred_gcc_version = v"9",
+               preferred_llvm_version = v"13",
+               clang_use_lld=false)
```
giordano, line 368: Missing newline
```suggestion
               clang_use_lld=false)

```

P/PETSc/build_tarballs.jl
```diff
@@ -7,14 +7,13 @@ name = "PETSc"
 version = v"3.18.7"
 petsc_version = v"3.18.6"
 MUMPS_COMPAT_VERSION = "5.6.2"
-SUITESPARSE_COMPAT_VERSION = "7.2.1" 
 SUPERLUDIST_COMPAT_VERSION = "8.1.2"   
 MPItrampoline_compat_version="5.2.1"
 BLASTRAMPOLINE_COMPAT_VERSION="5.8.0"
```
ViralBShah, line 12: I think you want this to be 5.4.0, and `julia_compat` to be `1.9` for using LBT.

P/PETSc/build_tarballs.jl
```diff
@@ -326,6 +361,8 @@ ENV["MPITRAMPOLINE_DELAY_INIT"] = "1"
 # Build the tarballs.
 # NOTE: llvm16 seems to have an issue with PETSc 3.18.x as on apple architectures it doesn't know how to create dynamic libraries  
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               augment_platform_block, julia_compat="1.10", 
-               preferred_gcc_version = v"9",  
-               preferred_llvm_version=v"13")
+               augment_platform_block, 
+               julia_compat="1.6",
```
ViralBShah, line 365: You want `julia_compat` to be `1.9`, since we are using LBT 5.8 in this build (which is the right thing to do).

imciner2:
> Does Petsc patch SuiteSparse? Presumably when they release a version compatible with SuiteSparse 7.4 or 7.5, we can just update to that and it will work out of the box.

The issue with this will be SuiteSparse ABI making PETSc versions then tied to specific Julia versions instead of being more separate (since we ship the SuiteSparse JLL in the Julia image). I don't know if that is something that would be a problem for PETSc, but it is something to keep in mind.

They are working to update to a newer SuiteSparse version at least: https://gitlab.com/petsc/petsc/-/merge_requests/7104.

boriskaus:
> > Does Petsc patch SuiteSparse? Presumably when they release a version compatible with SuiteSparse 7.4 or 7.5, we can just update to that and it will work out of the box.
> 
> The issue with this will be SuiteSparse ABI making PETSc versions then tied to specific Julia versions instead of being more separate (since we ship the SuiteSparse JLL in the Julia image). I don't know if that is something that would be a problem for PETSc, but it is something to keep in mind.
> 

yes, that is precisely the issue; I would like to be able to have different versions of PETSc running on different versions of Julia.
 
> They are working to update to a newer SuiteSparse version at least: https://gitlab.com/petsc/petsc/-/merge_requests/7104.

yes; the current PR is for an older version of PETSc (needed for downstream codes).

ViralBShah:
It would be good to know where LBT causes segfaults in the libblastrampoline repo, so we can get them fixed.

boriskaus:
In case this got lost in the editing of the comments above:

1. I was unable to compile PETSc successfully with LBT as it results in segfaults (including on apple aarch), so have to leave it with OpenBLAS for now (perhaps this works better in a more recent version of PETSc).
2. SuiteSparse 5.13 is not available for the different Julia versions (1.9 - 1.11), which is a requirement for PETSc. More recent versions of PETSc are upgrading SuiteSparse, but backwards compatibility of PETSc remains important (linking PETSc to a specific version of Julia is no good). I can give the linking of 

As far as I can tell, this is therefore ready to be merged; various downstream packages (LaMEM, InteractiveGeodynamics) are currently broken on Julia 1.10, and this would fix that.

boriskaus:
The results for tests that have LBT and segfault are [here](https://github.com/boriskaus/test_PETSc_jll/actions/runs/7629633632/job/20783458285#step:7:179). 
`ex19_int64_deb` is the example that is compiled wit debug flags active.
I'm afraid it is not very informative:
```
[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range
[0]PETSC ERROR: Try option -start_in_debugger or -on_error_attach_debugger
[0]PETSC ERROR: or see https://petsc.org/release/faq/#valgrind and https://petsc.org/release/faq/
[0]PETSC ERROR: ---------------------  Stack Frames ------------------------------------
[0]PETSC ERROR: The line numbers in the error traceback are not always exact.
[0]PETSC ERROR: #1 BLASdot()
[0]PETSC ERROR: #2 VecNorm_Seq() at /workspace/srcdir/petsc-3.18.6/src/vec/vec/impls/seq/bvec2.c:216
[0]PETSC ERROR: #3 VecNorm() at /workspace/srcdir/petsc-3.18.6/src/vec/vec/interface/rvector.c:237
[0]PETSC ERROR: #4 SNESSolve_NEWTONLS() at /workspace/srcdir/petsc-3.18.6/src/snes/impls/ls/ls.c:167
[0]PETSC ERROR: #5 SNESSolve() at /workspace/srcdir/petsc-3.18.6/src/snes/interface/snes.c:4692
```

and 
```
[0]PETSC ERROR: The line numbers in the error traceback are not always exact.
[0]PETSC ERROR: #1 VecNorm_MPI() at /workspace/srcdir/petsc-3.18.6/src/vec/vec/impls/mpi/pvec2.c:43
```

ViralBShah:
Is the error only on Apple? How does Petsc call the BLAS? Do we do the name mangling by adding `_64` as the suffix to the BLAS/LAPACK calls? Without that it is calling the LP64 interface instead of the ILP64 interface.

boriskaus:
> Is the error only on Apple? 

I only tested it on apple with 3.20.

> How does Petsc call the BLAS? Do we do the name mangling by adding `_64` as the suffix to the BLAS/LAPACK calls? Without that it is calling the LP64 interface instead of the ILP64 interface.

[This](https://gitlab.com/petsc/petsc/-/blame/main/src/vec/vec/impls/seq/bvec2.c?ref_type=heads#L206) is one of the offending lines in PETSc.

ViralBShah:
> > How does Petsc call the BLAS? Do we do the name mangling by adding `_64` as the suffix to the BLAS/LAPACK calls? Without that it is calling the LP64 interface instead of the ILP64 interface.
> 
> [This](https://gitlab.com/petsc/petsc/-/blame/main/src/vec/vec/impls/seq/bvec2.c?ref_type=heads#L206) is one of the offending lines in PETSc.

Basically, in this file https://gitlab.com/petsc/petsc/-/blob/main/include/petscblaslapack_mangle.h?ref_type=heads, `PETSC_BLASLAPACK_SUFFIX` should get set to `_64`. I don't know the Petsc build system enough to know how to do that, but that will make it compatible with ILP64 (and hence remove the need for OpenBLAS32). But shouldn't hold this relase.

P/PETSc/CreatePatch_Boris/petsc
imciner2, line 1: Why was the submodule added to PETSc here?

boriskaus: That was a mixup from my side and I somehow seem to be unable to remove it.

imciner2: I've removed it in https://github.com/JuliaPackaging/Yggdrasil/pull/8009.

PR #7990, opened 2024-01-22
libsingular_julia for Singular_jll 403.214.1400

hannes14:
@fingolfin


imciner2:
> is this gonna be an issue? (i.e. building with the old version but using the new one at runtime)

Did the ABI for Singular_jll change between the two versions? If so, then there will be a problem and the compat needs to be updated. If the ABI is the same, then it shouldn't be an issue with the compat entry.

PR #7996, opened 2024-01-22
[gdk_pixbuf] bump patch version and libtiff compat

visr:
This library has not yet been rebuild against libtiff 4.5.1, since https://github.com/JuliaRegistries/General/pull/86099. That means the GTK stack cannot update to these newer libtiff builds that other deps require.

### DIFF
--- a/G/gdk_pixbuf/build_tarballs.jl
+++ b/G/gdk_pixbuf/build_tarballs.jl
@@ -68,4 +68,5 @@ dependencies = [
 ]
 
 # Build the tarballs.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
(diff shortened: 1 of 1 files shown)


G/gdk_pixbuf/build_tarballs.jl
```diff
@@ -78,7 +78,7 @@ dependencies = [
     Dependency("Glib_jll"; compat="2.68.3"),
```
jwahlstrand, line 78: Setting Glib_jll compat to 2.76.5 gets x86_64-linux-gnu compiling for me locally. I think it's related to GObject introspection, which is enabled just for that platform.

jwahlstrand:
Adding `clang_use_lld=false` to `build_tarballs()` and removing lines 19-35 gets the apple builds working for me.

PR #7997, opened 2024-01-22
[Modflow6] build version 6.4.2

visr:


giordano:
When there are [weird linker errors](https://buildkite.com/julialang/yggdrasil/builds/7899#018d32b5-36c3-4a11-9e57-622a3fc75c08/617-1722) on BSD systems like
```
[20:13:56] ninja: job failed: /opt/bin/x86_64-apple-darwin14-libgfortran3-cxx11/x86_64-apple-darwin14-gfortran  -o src/mf6 src/mf6.p/mf6.f90.o -L/workspace/destdir/lib -Wl,-dead_strip_dylibs -Wl,-headerpad_max_install_names -Wl,-undefined,error src/libmf6core.a src/libmf6_external.a
[20:13:56] ld: warning: ignoring file src/libmf6_external.a, building for macOS-x86_64 but attempting to link with file built for unknown-unsupported file format ( 0x21 0x3C 0x61 0x72 0x63 0x68 0x3E 0x0A 0x2F 0x2F 0x20 0x20 0x20 0x20 0x20 0x20 )
[20:13:56] ld: warning: ignoring file src/libmf6core.a, building for macOS-x86_64 but attempting to link with file built for unknown-unsupported file format ( 0x21 0x3C 0x61 0x72 0x63 0x68 0x3E 0x0A 0x2F 0x2F 0x20 0x20 0x20 0x20 0x20 0x20 )
[20:13:56] Undefined symbols for architecture x86_64:
[20:13:56]   "___mf6coremodule_MOD_mf6run", referenced from:
[20:13:56]       _main in mf6.f90.o
[20:13:56] ld: symbol(s) not found for architecture x86_64
[20:13:56] collect2: error: ld returned 1 exit status
```
the problem is likely using lld as linker, which you can opt out with `build_tarballs(...; clang_use_lld=false)`

visr:
Thanks for the tip. Unfortunately in this case `clang_use_lld=false` doesn't seem to have an effect. I tried various GCC versions but that also doesn't help.

Going back to the last built Modflow version gets the same error, so it seems like something that changed in the build environment causes this issue.

visr:
Closing this as I don't need it right now, and new Modflow versions have come out since.

PR #8004, opened 2024-01-24
[PETSc] version 3.18.8

boriskaus:
In `PETSc_jll v3.18.7+1`, merged yesterday, we removed SuiteSparse as a dependency.  Because it was a dependency in version `PETSc_jll 3.18.7+0`, we have problems linking it with other codes [see here](https://github.com/JuliaGeodynamics/LaMEM.jl/actions/runs/7636728953/job/20804282693#step:6:427) as it is looking for SuiteSparse files.

I should really have bumped the version number which is done with this PR.

### DIFF
--- a/P/PETSc/build_tarballs.jl
+++ b/P/PETSc/build_tarballs.jl
@@ -4,7 +4,7 @@ const YGGDRASIL_DIR = "../.."
 include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
 
 name = "PETSc"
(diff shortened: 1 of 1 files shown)


imciner2:
This is good for the PETSc recipe, and then once this is merged Lamem should be updated to have a compat for `~3.18.8` on PETSc and be rebuilt.

We can also yank the 3.18.7+1 version in the general repo to at least let 3.18.7 "work" again.

Issue #8001, opened 2024-01-23
Patch GCC 11 build to fix AMX instructions


imciner2:
To build the OpenBLAS BFloat16 extensions on Windows, we need to patch the GCC AMX intrinsics header to fix two sets of typos:

* https://github.com/gcc-mirror/gcc/commit/7fcc22dae70af4202fe83b9ecb642fd6333464a2
* https://github.com/gcc-mirror/gcc/commit/aeb9b58225916bc84a0cd02c6fc77bbb92167e53

These two patches are both in GCC 12.3 and GCC 13, but we can't switch to using that for OpenBLAS yet, so we need to live with GCC 11.

The first one is probably harmless, but was still a typo none the less. The second I believe is actively breaking the OpenBLAS build with BFloat16 enabled for the Windows 64-bit platform in https://github.com/JuliaPackaging/Yggdrasil/pull/7202 with the errors

```
[18:55:08] /tmp/ccNhJljp.s: Assembler messages:
[18:55:08] /tmp/ccNhJljp.s:108: Error: `(%rax,%r10d,1)' is not a valid base/index expression
[18:55:08] /tmp/ccNhJljp.s:111: Error: `(%rdx,%r11d,1)' is not a valid base/index expression
[18:55:08] /tmp/ccNhJljp.s:125: Error: `(%r8,%r10d,1)' is not a valid base/index expression
[18:55:08] /tmp/ccNhJljp.s:131: Error: `(%rcx,%eax,1)' is not a valid base/index expression
[18:55:08] /tmp/ccNhJljp.s:175: Error: `(%r8,%ebx,1)' is not a valid base/index expression
[18:55:08] /tmp/ccNhJljp.s:178: Error: `(%rax,%r11d,1)' is not a valid base/index expression
[18:55:08] /tmp/ccNhJljp.s:195: Error: `(%rdx,%eax,1)' is not a valid base/index expression
[18:55:08] /tmp/ccNhJljp.s:203: Error: `(%rcx,%eax,1)' is not a valid base/index expression
[18:55:08] make[1]: *** [Makefile.L3:637: sbgemm_oncopy_SAPPHIRERAPIDS.obj] Error 1
```

More details in https://github.com/OpenMathLib/OpenBLAS/issues/4450#issuecomment-1905945228.

PR #8005, opened 2024-01-24
[GCCBootstrap] Add patches to fix AMX intrinsics and rebuild for Mingw 11.0.1

imciner2:
* The GCC patches are needed to fix typos present in GCC 11+12 that affect the use of the _tile_loadd and _tile_stored intrinsics and can cause assembler errors on Windows. These are already in GCC 13, so only need to be backported and not carried further.
* Rebuild GCC 11, 12 and 13 for the new Mingw version, since this was missed during the rebuilds for the other GCC versions.

Corresponding PR to update the artifacts in BinaryBuilderBase: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/365

Fixes https://github.com/JuliaPackaging/Yggdrasil/issues/8001/


0_RootFS/GCCBootstrap@11/bundled/patches/mingw/gcc1110-add-__udivmoddi4.patch
giordano, line 1: Just for the record, why were these patches deleted?

giordano (approved):
Looks good to me, I only left a comment to understand why some patches were deleted

0_RootFS/GCCBootstrap@11/bundled/patches/mingw/gcc1110-add-__udivmoddi4.patch
imciner2: These changes are now part of the Mingw 11 release (commits https://sourceforge.net/p/mingw-w64/mingw-w64/ci/6a0e9165008f731bccadfc41a59719cf7c8efc02/ and https://sourceforge.net/p/mingw-w64/mingw-w64/ci/ab0fa5ad3c67d634fd7754d9473d1bb7051f9bf0/), so we don't need to carry the patches anymore because we build with Mingw 11 now.

PR #8008, opened 2024-01-24
[LaMEM] version 2.1.3 

boriskaus:
This updates LaMEM to use `PETSc_jll` version 3.18.8.

A test with a locally compiled version suggests that it runs fine on [Windows/Mac/linux](https://github.com/JuliaGeodynamics/LaMEM.jl/actions/runs/7644174772/job/20827882495)

### DIFF
--- a/L/LaMEM/build_tarballs.jl
+++ b/L/LaMEM/build_tarballs.jl
@@ -65,13 +65,21 @@ augment_platform_block = """
 # platforms are passed in on the command line
 platforms = expand_gfortran_versions(supported_platforms(exclude=[Platform("i686", "windows"),
                                                                   Platform("i686","linux"; libc="musl"),
+                                                                  Platform("i686","linux"; libc="gnu"),
+                                                                  Platform("x86_64","freebsd"),
+                                                                  Platform("armv6l","linux"; libc="musl"),
+                                                                  Platform("armv7l","linux"; libc="musl"),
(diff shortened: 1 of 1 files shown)


L/LaMEM/build_tarballs.jl
```diff
@@ -6,16 +6,16 @@ const YGGDRASIL_DIR = "../.."
 include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
 
 name = "LaMEM"
-version = v"2.1.2"
+version = v"2.1.3"
 
 
-PETSc_COMPAT_VERSION = "3.18.6" # Note: this is the version of the PETSc_jll package, which is sometimes larger than the PETSc version  
+PETSc_COMPAT_VERSION = "3.18.8" # Note: this is the version of the PETSc_jll package, which is sometimes larger than the PETSc version
```
imciner2, line 12: This compat line will give it a compat entry of [3.18.8 - 4.0.0) I believe, but I think you said that PETSc changes ABI every release, so 3.19 won't be compatible with a build using 3.18. I think it would be better then to make this `~3.18.8` because that will limit it to only PETSc 3.18, and not try to automatically install PETSc 3.20 when you update to that.

boriskaus: yes good point

imciner2:
Looks good.

PR #8019, opened 2024-01-26
pgplot: support macos and png

aplavin:
Seems to cross-compile fine! Some build steps adapted from https://github.com/kazuakiyama/homebrew-pgplot, patches from macports https://trac.macports.org/browser/trunk/dports/graphics/pgplot/files.

I personally only need pgplot for difmap, hope this build works there as well. Will update difmap script after this one gets through, I guess it's not possible to combine into one PR.


P/PGPLOT/build_tarballs.jl
```diff
@@ -5,30 +5,54 @@ using BinaryBuilder, Pkg
...
+if [[ "${target}" == *-apple-* ]]; then
+    ../pgplot/makemake ../pgplot/ darwin
+    make lib SHELL=/bin/bash
+else
+    ../pgplot/makemake ../pgplot/ linux g77_gcc
+    sed -i 's|FCOMPL=g77|FCOMPL=gfortran|' makefile
+    make lib SHARED_LD="${FC} -shared  -o libpgplot.${dlext}"
+fi
+
+cp libpgplot.${dlext} "${libdir}/libpgplot.${dlext}"
```
giordano, line 39: Just to make the command more verbose
```suggestion
install -Dvm 755 libpgplot.${dlext} "${libdir}/libpgplot.${dlext}"
```

P/PGPLOT/build_tarballs.jl
```diff
@@ -5,30 +5,54 @@ using BinaryBuilder, Pkg
...
+if [[ "${target}" == *-apple-* ]]; then
+    cd pgplot
+    for f in ../patch-*;
+    do
+    patch -p0 <$f;
+    done
+    mkdir sys_darwin
+    cp ../bb.conf ./sys_darwin/bb.conf
+    cd ..
+fi
```
giordano, line 25: I thought I left a comment about this, but looks like I didn't save
```suggestion
if [[ "${target}" == *-apple-* ]]; then
    pushd pgplot
    for f in ../patch-*; do
        atomic_patch -p0 "${f}"
    done
    mkdir sys_darwin
    cp ../bb.conf ./sys_darwin/bb.conf
    popd
fi
```

giordano (commented):
> How to bump new JLL version correctly? It's the same pgplot 5.2.

Since you're changing the julia compat you must change the version number too.  Maybe use a fake 5.2.1, if they don't seem to use all three parts of the version number?

PR #8024, opened 2024-01-26
[libsingular_julia] update to latest, drop Singular_jll compat entry

fingolfin:
CC @hannes14 @ederc @benlorenz The change to turn Singular_jll into a bdep is as discussed previously:

Decoupling libsingular_julia_jll from Singular_jll is OK because Singular.jl also specifies a version of Singular_jll. The drawback is that one could end up accidentally using the wrong Singular_jll version. But the big advantage is that we are freed to work with newer Singular_jll release in Singular.jl which is safe when using the JLL override mechanism there (when changes in the C++ code is detected while in dev mode, the C++ code just gets compiled and used to replace the code in the JLL).

Dropping the explicit version for Singular_jll also removes the chance for forgetting to update that. Of course all this means that we have to pay close attention in Singular.jl to match what was actually used to build any `libsingular_julia` updates...

### DIFF
--- a/L/libsingular_julia/build_tarballs.jl
+++ b/L/libsingular_julia/build_tarballs.jl
@@ -9,11 +9,11 @@ uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "libsingular_julia"
-version = v"0.40.11"
+version = v"0.41.0"
 
 # Collection of sources required to build libsingular-julia
 sources = [
-    GitSource("https://github.com/oscar-system/Singular.jl.git", "99dc5160176fb81ff69064fe7a3fdddf84c95968"),
+    GitSource("https://github.com/oscar-system/Singular.jl.git", "b3295f802f1638bf85a5fd4fe127443b72727cce"),
 ]
 
 # Bash recipe for building across all platforms
@@ -54,7 +54,11 @@ dependencies = [
     BuildDependency("GMP_jll"),
     BuildDependency("MPFR_jll"),
     Dependency("libcxxwrap_julia_jll"; compat = "~0.11.2"),
-    Dependency("Singular_jll", compat = "^403.214.1400"),
+    # we do not set a compat entry for Singular_jll -- instead we leave it to
+    # Singular.jl to ensure the right versions of libsingular_julia_jll and
+    # Singular_jll are paired. This gives us flexibility in the development
+    # setup there.
+    Dependency("Singular_jll", v"403.214.1400"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.


imciner2:
> Dropping the explicit version for Singular_jll also removes the chance for forgetting to update that. Of course all this means that we have to pay close attention in Singular.jl to match what was actually used to build any libsingular_julia updates...

So you will always want to build with the most recent Singular_jll version?

fingolfin:
Precisely, we always want to build against the latest `Singular_jll`. But more important for us that our mechanism to test updates to the code used in `libsingular_julia_jll` will be able to run tests with newer `Singular_jll` versions than what was used to build `libsingular_julia_jll` if we remove the version in the dependency here, and only enforce it in `Singular.jl`.

But the PR in the current form has a mistake: `Singular_jll` should remain a dependency, it is not just a build dependency; the crucial bit is to not add the version to `compat`.

I'll update the PR shortly.

hannes14:
Singular_jll is both build time and run time dependency.
Building always with the newest version of Singular_jll is fine with me

giordano:
I'm a bit confused by how you plan to handle incompatible versions if you don't specify compat bounds.  Unless libsingular is super API/ABI-stable.

fingolfin:
There is a tight coupling between `libsingular_julia_jll` and `Singular.jl` (and their releases): the source code for the former is part of the repository for the latter.

To this end, `Singular.jl` has a strict compat for `libsingular_julia_jll`. E.g. right now.
```
libsingular_julia_jll = "=0.40.11"
```


Our development works like this: first you edit the C++ and Julia code in the Singular.jl repository as necessary. When you then load this modified dev version of Singular.jl it detects that the C++ code was modified compared to what is in `libsingular_julia_jll` (by comparing the tree hash of the C++ source repository). If that is the case, it doesn't load the `libsingular_julia` from the JLL; instead it compiles the C++ code and uses the result of that.

This makes it MUCH simpler to evolve the code (a single PR can contain all the necessary changes, including updates for the C++ code). The downside is that we need to remember to update the JLL before the release (as we don't want users to have a C++ compiler etc. etc.), but that's OK.

But we also sometimes need a newer version of `Singular_jll` than what is used in `libsingular_julia_jll`. This is where this PR comes into play: it allows us to do just that: before this PR here, we had to first make an update to `libsingular_julia_jll` to use the newer `Singular_jll`, with no other changes, then could start writing a PR to use the new `Singular_jll`, then update `libsingular_julia_jll` *again*, now with our new C++ code. With this PR, we can simplify the process: first make a PR to Singular.jl which edits the C++ and Julia code, *and* updates the `Singular_jll` compat entry in the `Project.toml` there. Once all is good, merge that; then update `libsingular_julia_jll`, just once.

Note that 

The only serious downside I can think of is that one might rebuild `libsingular_julia_jll` against a newer (incompatible) `Singular_jll` but forgets to update the version of `libsingular_julia_jll` (that could potentially break existing releases of `Singular.jl`). If that happens, we can fix it by either yanking that update or rebuilding `libsingular_julia_jll` again against an older `Singular_jll`.

The other potential mistake is to rebuild `libsingular_julia_jll` against a newer `Singular_jll`, including changing the veresion of `libsingular_julia_jll`; but then forgetting to update the version of `Singular_jll` in `Singular.jl`'s `Project.toml`. But in 99% that will just lead to CI crashing (and in the other cases... well, it might actually be fine).

PR #8044, opened 2024-01-30
[strace] Don't build static anymore, avoid `aarch64-linux-musl`

staticfloat:
It appears that old glibc has a bug when reading locale data that makes this particular version of glibc useless for building statically with. Sigh.  Instead, we'll just build normal versions and build for musl as well, but it appears that `aarch64-linux-musl` has a problem that I'm not going to patch right now.


giordano:
You can also build with newer glibc, I think crun (by memory) recipe does that

PR #8046, opened 2024-01-31
[GR] bump GR version

jheinen:
bump GR run-time version to 0.73.2


jheinen:
This MR fixes the `GKSTerm.app not found` problem.

PR #8056, opened 2024-02-02
New Package: Hydrological Model VIC5

kongdd:
This model is under the MIT license.
https://github.com/CUG-hydro/VIC5.c/blob/classic/LICENSE.txt

### DIFF
--- a/V/VIC5/build_tarballs.jl
+++ b/V/VIC5/build_tarballs.jl
@@ -13,7 +13,11 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/VIC5.c/vic
 
-target=${libdir}/libvic5_classic.${dlext} make dll
+mkdir -p "${libdir}"
+mkdir -p "${bindir}"
+
+target=${libdir}/libvic5_classic.${dlext} make dll -j${nproc}
+install_license ${WORKSPACE}/srcdir/VIC5.c/LICENSE.txt
 """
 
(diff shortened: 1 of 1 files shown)


V/VIC5/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/VIC5.c/vic
+
+mkdir -p "${libdir}"
+mkdir -p "${bindir}"
+
+target=${libdir}/libvic5_classic.${dlext} make dll -j${nproc}
```
giordano, line 19:
```suggestion
target=${libdir}/libvic5_classic.${dlext} make dll CC=${CC} -j${nproc}
```

V/VIC5/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+install_license ${WORKSPACE}/srcdir/VIC5.c/LICENSE.txt
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = expand_cxxstring_abis(supported_platforms())
```
giordano, line 25: Why `expand_cxxstring_abis`?  This looks like a pure C package, and the auditor doesn't say products contain `std::string` symbols.
```suggestion
platforms = supported_platforms()
```

PR #8057, opened 2024-02-02
[MUMPS_seq] Add the 64-bit integer version

amontoison:
Some packages like PETSc can be compiled with the 64-bit integer version.
I propose to add it in `MUMPS_seq_jll.jl`.

### DIFF
--- a/M/MUMPS/MUMPS_seq@5/build_tarballs.jl
+++ b/M/MUMPS/MUMPS_seq@5/build_tarballs.jl
@@ -108,7 +108,7 @@ make_args_64+=(PLAT="64"
                LIBBLAS="${BLAS_LAPACK}"
                LAPACK="${BLAS_LAPACK}")
 
-make -j${nproc} allshared "${make_args_64[@]}"
+make allshared "${make_args_64[@]}"
 
 cp lib/*.${dlext} ${libdir}
 """
--- a/M/MUMPS/MUMPS_seq@5/build_tarballs.jl
+++ b/M/MUMPS/MUMPS_seq@5/build_tarballs.jl
@@ -135,4 +135,4 @@ dependencies = [
 ]
 
 # Build the tarballs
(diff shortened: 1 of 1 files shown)


M/MUMPS/MUMPS_seq@5/build_tarballs.jl
```diff
@@ -15,7 +15,7 @@ using BinaryBuilder, Pkg
 
 name = "MUMPS_seq"
 upstream_version = v"5.6.2"
-version_offset = v"0.0.1" # reset to 0.0.0 once the upstream version changes
+version_offset = v"0.0.2" # reset to 0.0.0 once the upstream version changes
```
ViralBShah, line 18: Won't Yggdrasil just publish new versions as +1, +2, etc? Unless we are patching the package, I thought we wouldn't need to change the version_offset.

amontoison:
@ViralBShah The issue is with GCC 7, you can't use another compiler because it's the only one with `libgfortran4`.
It's a compiler bug with the flag `-ffixed-line-length-none`...

M/MUMPS/MUMPS_seq@5/build_tarballs.jl
amontoison: Yes, but for Coin packages and its dependencies, we multiply all version numbers by 100:
https://github.com/JuliaPackaging/Yggdrasil/blob/master/M/MUMPS/MUMPS_seq%405/build_tarballs.jl#L19-L21
We often recompile a package (like Ipopt) with updated dependencies (MUMPS, SPRAL, ...).
Without this trick, if the new version of the dependency breaks the package, all the Coin ecosystem is broken... (cc @odow :stuck_out_tongue:).

amontoison:
Superseded by https://github.com/JuliaPackaging/Yggdrasil/pull/8548

PR #8066, opened 2024-02-05
Upgrade Clp -- v1.17.9

amontoison:
@odow

PR #8067, opened 2024-02-05
Upgrade CoinUtils -- 2.11.11

amontoison:
@odow

### DIFF
--- a/C/Coin-OR/CoinUtils/build_tarballs.jl
+++ b/C/Coin-OR/CoinUtils/build_tarballs.jl
@@ -1,7 +1,7 @@
 # In addition to coin-or-common.jl, we need to modify this file to trigger a
 # rebuild.
 #
-# Last updated: 2022-10-26
+# Last updated: 2024-02-05
 
 include("../coin-or-common.jl")
 
@@ -36,6 +36,13 @@ elif [[ ${target} == *linux* ]]; then
     export LDFLAGS="-ldl -lrt"
 fi
 
+# BLAS and LAPACK
+if [[ "${target}" == *mingw* ]]; then
+  LBT="-lblastrampoline-5"
+else
+  LBT="-lblastrampoline"
+fi
+
 ../configure \
     --prefix=${prefix} \
     --build=${MACHTYPE} \
@@ -45,8 +52,10 @@ fi
     --disable-debug \
     --enable-shared \
     lt_cv_deplibs_check_method=pass_all \
-    --with-blas --with-blas-lib="-lopenblas" \
-    --with-lapack --with-lapack-lib="-lopenblas"
+    --with-blas \
+    --with-blas-lib="-L${libdir} ${LBT}" \
+    --with-lapack \
+    --with-lapack-lib="-L${libdir} ${LBT}"
 
 make -j${nproc}
 make install
@@ -58,9 +67,9 @@ products = [
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[
-    Dependency("OpenBLAS32_jll", OpenBLAS32_version),
-    Dependency("CompilerSupportLibraries_jll"),
+dependencies = [
+    Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"), compat="5.4.0"),
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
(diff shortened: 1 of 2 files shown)


odow (changes requested):
Wait for resolution of https://github.com/JuliaPackaging/Yggdrasil/pull/8066#discussion_r1477815422

odow (approved):
Okay try this now

odow:
A list of the order is:

[CoinUtils_jll]
[Osi_jll] 
[Clp_jll] 
[Cgl_jll] 
[Cbc_jll] 
[Bonmin_jll] 
[SYMPHONY_jll] 
[ALPS_jll] 
[BiCePS_jll] 
[CHIPPS_BLIS_jll] 
[MibS_jll]

ViralBShah:
I believe you have to merge CoinUtils first, and then use it in the other packages in the dependency chain - and keep doing that repeatedly.

amontoison:
Ipopt_jll is already linked with LBT so we could remove the folder for the artifact IpoptMKL_jll.

Yes, I discussed with Oscar a few days ago and we will start with CoinUtils, OsiUtils and Clp.

ViralBShah:
We can just let IpOptMKL be around, but just not update it going forward.

odow:
> I believe you have to merge CoinUtils first, and then use it in the other packages in the dependency chain - and keep doing that repeatedly.

Yeah, @amontoison has this under control, but they're traveling etc over the next few weeks. There is no rush.

imciner2:
> Ipopt_jll is already linked with LBT so we could remove the folder for the artifact IpoptMKL_jll.

Has anyone tested running the LBT build with MKL selected? We should make sure we don't see the same problem that SLICOT had when using MKL with LBT (e.g. https://github.com/JuliaLinearAlgebra/MKL.jl/issues/138).

imciner2:
> But we compile all COIN-OR packages with LP64 BLAS / LAPACK and not ILP64 BLAS / LAPACK so we can't have this problem.
> 
> I also checked with nm -D libCoinUtils.so.3.11.10 and only two symbols (related to dense LU) are used:

Excellent, then it should be safe to deprecate the old IpOptMKL and just use the main build from now on. IMO, I agree we can leave the build scripts there, but I think we should  add a `DEPRECATED` file to the folder explaining the deprecation and that the normal IpOpt should be used instead.

ViralBShah:
Should we merge this and then merge the Clp PR?

odow:
There's now 2.11.11 so we should probably first update to that 
https://github.com/coin-or/CoinUtils/releases/tag/releases%2F2.11.11

amontoison:
I will update the PR tomorrow. I forgot that I started to work on the LBT support in COIN-OR packages.

C/Coin-OR/coin-or-common.jl
```diff
@@ -45,8 +45,8 @@ Clp_gitsha = "1c2586a08d33ecc59ed67d319c29044802c0866b"
 Osi_version = offset_version(v"0.108.7", v"0.0.0")
 Osi_gitsha = "c67528fc8b2a36dc936b285fe34030439b9147c3"
 
-CoinUtils_version = offset_version(v"2.11.6", v"0.0.0")
-CoinUtils_gitsha = "26e9639ed9897e13e89169870dbe910296a9783b"
+CoinUtils_version = offset_version(v"2.11.10", v"0.0.0")
+CoinUtils_gitsha = "5b98c4bb87ceebedb86c6e921dc4ad0af6166c8e"
```
odow, line 49:
```suggestion
CoinUtils_gitsha = "15a819b7e8763b2557e4bf440f8cb62ee6734f36"
```

C/Coin-OR/coin-or-common.jl
```diff
@@ -45,8 +45,8 @@ Clp_gitsha = "1c2586a08d33ecc59ed67d319c29044802c0866b"
 Osi_version = offset_version(v"0.108.7", v"0.0.0")
 Osi_gitsha = "c67528fc8b2a36dc936b285fe34030439b9147c3"
 
-CoinUtils_version = offset_version(v"2.11.6", v"0.0.0")
-CoinUtils_gitsha = "26e9639ed9897e13e89169870dbe910296a9783b"
+CoinUtils_version = offset_version(v"2.11.10", v"0.0.0")
```
odow, line 48:
```suggestion
CoinUtils_version = offset_version(v"2.11.11", v"0.0.0")
```

C/Coin-OR/CoinUtils/build_tarballs.jl
```diff
@@ -1,7 +1,7 @@
 # In addition to coin-or-common.jl, we need to modify this file to trigger a
 # rebuild.
 #
-# Last updated: 2022-10-26
+# Last updated: 2024-02-05
```
amontoison, line 4:
```suggestion
# Last updated: 2024-06-19
```

PR #8081, opened 2024-02-09
Update libuv to v2-1.48.0

vtjnash:
Smoke build for https://github.com/JuliaLang/julia/pull/49937 to test if this update will go through smoothly


giordano:
> What is wrong with the sanitize build? This snippet of code seems to be copied into every build recipe, but is just broken here?

We don't have compiler-rt for llvm 16 for x86_64-linux-gnu (#7877)

PR #8082, opened 2024-02-09
Update casacore dependency cfitsio 4.0.0 -> 4.3.0

torrance:
Need to trigger a rebuild with the new dependency requirement since CFITSIO_jll seem to have broken semver.

### DIFF
--- a/C/casacore/build_tarballs.jl
+++ b/C/casacore/build_tarballs.jl
@@ -4,7 +4,7 @@
 using BinaryBuilder, Pkg
 
 name = "casacore"
-version = v"3.5.0+1"
+version = v"3.5.1"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


giordano (changes requested):
You can't just change the compat bound, to make the change effective you must also change the version number of the package you're building. Is there a newer version of casacore?

torrance:
@giordano No there's no later version of Casacore yet.

The problem I'm seeing is that CFITSIO_jll went from emitting libcfitsio.so.9 to libcfits.so.10 while keeping major version 4.x. The current build of Casacore links against libcfitsio.so.9, and if the most recent version of CIFTSIO_jll is installed, Casacore will fail.

I'm not sure the best way to fix this. Open to ideas.

torrance:
I've retained the version as 3.5.0 but I've bumped the build number to 3.5.0+1. Is this sufficient?

torrance:
Yggdrasil doesn't like the build number being set, so I'm (unfortunately) setting the version to 3.5.1, which diverges from the actual underlying library version.

giordano (approved):

PR #8090, opened 2024-02-11
Update FFMPEG to 6.1.1

theogf:
- `libresample` is deprecated and was therefore removed.
- `CUDA_ARGS` was set but `ffnvcodec` (I could not find any jll except for https://github.com/JuliaPackaging/Yggdrasil/blob/master/N/nv_codec_headers/build_tarballs.jl ?) is not provided, resulting in an error when trying to compile it.

There seems to be an error on MacOS, mentioned [here](https://trac.macports.org/ticket/68720l) and indicate that MacOS > 10.13 is needed?

### DIFF
--- a/F/FFMPEG/FFMPEG/build_tarballs.jl
+++ b/F/FFMPEG/FFMPEG/build_tarballs.jl
@@ -12,7 +12,6 @@ products = [
     LibraryProduct(["libavdevice", "avdevice"], :libavdevice),
     LibraryProduct(["libavfilter", "avfilter"], :libavfilter),
     LibraryProduct(["libavformat", "avformat"], :libavformat),
-    LibraryProduct(["libavresample", "avresample"], :libavresample),
     LibraryProduct(["libavutil", "avutil"], :libavutil),
     LibraryProduct(["libpostproc", "postproc"], :libpostproc),
     LibraryProduct(["libswresample", "swresample"], :libswresample),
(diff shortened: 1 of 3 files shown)


theogf:
Still throws an error around https://buildkite.com/julialang/yggdrasil/builds/8409#018dae14-f188-440b-a2bf-b486d91ab22c/6-1702

I think I will just give up on this PR at this point and wait for a fix from FFmpeg

F/FFMPEG/FFMPEG/build_tarballs.jl
```diff
@@ -12,36 +12,46 @@ products = [
...
+    name,
+    version,
+    sources,
+    script(; ffplay = false),
+    platforms,
+    products,
+    dependencies;
+    julia_compat = "1.6",
+    preferred_gcc_version = preferred_gcc_version,
+)
```
giordano, line 57:
```suggestion
    clang_use_lld=false,
)
```

theogf: Wow that's fast!

F/FFMPEG/FFMPEG/build_tarballs.jl
```diff
@@ -12,36 +12,47 @@ products = [
...
 ]
 
 # Dependencies that must be installed before this package can be built.
-# TODO: Theora once it's available
+# TODO: Theora once it's available 
 dependencies = [
     HostBuildDependency("YASM_jll"),
     BuildDependency("nv_codec_headers_jll"),
-    Dependency("libass_jll"; compat="0.15.1"),
+    Dependency("libass_jll"; compat = "0.15.1"),
```
giordano, line 26: Can you please remove all the whitespace changes?  It makes review needlessly more complicated.

theogf: Sure! Sorry about that, forgot to deactivate format on save

JoshuaLampert:
We get [Error building `FFMPEG`](https://github.com/trixi-framework/Trixi.jl/actions/runs/7926442642/job/21641220749#step:7:45) errors, which is probably related to this change?

giordano:
Only partially true, the problem is that the package manager is giving you an awfully old version of FFMPEG.jl

giordano:
@theogf would you mind following up this PR in [`FFMPEG.jl`](https://github.com/JuliaIO/FFMPEG.jl) to allow major version 6?

theogf:
https://github.com/JuliaIO/FFMPEG.jl/pull/59

PR #8101, opened 2024-02-13
[DuckDB] Bump to v0.10.0

Mytherin:


Mytherin:
Thanks for having a look - I'll patch the include and update the PR with a new git hash later today.

PR #8119, opened 2024-02-16
Update FFplay to 6.1.1

theogf:
Follow up to #8090 

Hopefully this should compile for mac + aarch

### DIFF
--- a/F/FFMPEG/FFplay/build_tarballs.jl
+++ b/F/FFMPEG/FFplay/build_tarballs.jl
@@ -18,6 +18,7 @@ dependencies = [
     BuildDependency("LibVPX_jll"), # We use the static archive
     Dependency("FFMPEG_jll"; compat=string(version)),
     Dependency("SDL2_jll"),
+    Dependency("OpenSSL_jll"; compat="3.0.8"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/F/FFMPEG/FFplay/build_tarballs.jl
+++ b/F/FFMPEG/FFplay/build_tarballs.jl
@@ -23,4 +23,4 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script(; ffplay=true), platforms, products, dependencies;
-               julia_compat="1.6", preferred_gcc_version=preferred_gcc_version, clang_use_lld=false)
+               julia_compat="1.6", preferred_gcc_version, clang_use_lld=false)
--- a/F/FFMPEG/FFplay/build_tarballs.jl
+++ b/F/FFMPEG/FFplay/build_tarballs.jl
@@ -20,9 +20,6 @@ dependencies = [
     Dependency("SDL2_jll"),
 ]
 
-# FFplay 4.4.4 does not build on aarch64-apple
-filter!(p -> !(Sys.isapple(p) && arch(p) == "aarch64"), platforms)
-
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script(; ffplay=true), platforms, products, dependencies;
                julia_compat="1.6", preferred_gcc_version=preferred_gcc_version)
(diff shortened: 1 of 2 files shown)


imciner2:
macOS aarch64 build failure is this:

> 11:55:01]  ---> mv "ffplay${exeext}" "${bindir}/ffplay${exeext}"
> [11:55:01] mv: can't rename 'ffplay': No such file or directory

It looks like it isn't building ffplay anymore, because also the configure output only shows two programs being built:
> Programs:
> ffmpeg                  ffprobe

but the configure output for the x86_64 linux build has

> Programs:
> ffmpeg                  ffplay                  ffprobe

theogf:
From reading a couple of issues around ffplay builds, it sounds like a dependency could be missing (which is weird for it to happen only for apple arm64...
But this could be connected to the following warning
```
Dependency SDL2_jll does not have a mapping for artifact SDL2 for platform aarch64-apple-darwin20-libgfortran5-cxx11
```

theogf:
Wild guess, but apple aarch64 is special cased when building SDL2:
https://github.com/JuliaPackaging/Yggdrasil/blob/d6ed3cd569a1662749bbe8ea22bb7a016bab5341/S/SDL2/build_tarballs.jl#L66

giordano:
> Wild guess, but apple aarch64 is special cased when building SDL2:

Yes, that's a wild guess because

1. you're looking at a dependency of SDL2, but the package has been built for aarch64-apple-darwin
2. the version of SDL2 that was built for aarch64-apple-darwin was v2.24.2 (#5812), but according to the build log of this PR you're using SDL2_jll v2.0.12, so whatever you're looking at in https://github.com/JuliaPackaging/Yggdrasil/blob/d6ed3cd569a1662749bbe8ea22bb7a016bab5341/S/SDL2/build_tarballs.jl is irrelevant since that's a different version than the one used here.  Which is the entire problem.

giordano:
More specifically the problem is
```
(jl_CWPdaN) pkg> add SDL2_jll@v2.24 FFMPEG_jll@v6
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package OpenSSL_jll [458c3c95]:
 OpenSSL_jll [458c3c95] log:
 ├─possible versions are: 1.1.1 - 3.0.13 or uninstalled
 ├─restricted to versions * by an explicit requirement, leaving only versions: 1.1.1 - 3.0.13
 ├─restricted by compatibility requirements with FFMPEG_jll [b22a6f82] to versions: 3.0.9 - 3.0.13
 │ └─FFMPEG_jll [b22a6f82] log:
 │   ├─possible versions are: 4.1.0 - 6.1.1 or uninstalled
 │   └─restricted to versions 6 by an explicit requirement, leaving only versions: 6.1.1
 └─restricted by compatibility requirements with PulseAudio_jll [02771fc1] to versions: 1.1.10 - 1.1.23 — no versions left
   └─PulseAudio_jll [02771fc1] log:
     ├─possible versions are: 14.2.0 - 15.0.0 or uninstalled
```

giordano:
#8168 #8169

F/FFMPEG/FFplay/build_tarballs.jl
```diff
@@ -20,9 +20,6 @@ dependencies = [
     Dependency("SDL2_jll"),
```
giordano, line 20:
```suggestion
    Dependency("SDL2_jll"),
    Dependency("OpenSSL_jll"; compat="3.0.8"),
```

F/FFMPEG/FFplay/build_tarballs.jl
```diff
@@ -18,11 +18,9 @@ dependencies = [
     BuildDependency("LibVPX_jll"), # We use the static archive
     Dependency("FFMPEG_jll"; compat=string(version)),
     Dependency("SDL2_jll"),
+    Dependency("OpenSSL_jll"; compat="3.0.8"),
```
giordano, line 21:
```suggestion
    Dependency("OpenSSL_jll"; compat="3.0.9"),
```

theogf:
Maybe one needs to add a compat on SDL2 ?

PR #8121, opened 2024-02-16
[finufft] Library build using CMake

jkrimmer:
- Perform library builds using CMake
- Fix broken library and failing Tests on Apple Silicon (observed on M3 Pro)
- Remove obsolete(?) DFFTW_PLAN_SAFE

EDIT: Oops, I messed up and changed the version number... Should probably read 2.2.0+1 instead?

### DIFF
--- a/F/finufft/build_tarballs.jl
+++ b/F/finufft/build_tarballs.jl
@@ -6,38 +6,28 @@ using BinaryBuilderBase
 include(joinpath(@__DIR__, "..", "..", "platforms", "microarchitectures.jl"))
 
 name = "finufft"
-version = v"2.2.0"
+version = v"2.3.0"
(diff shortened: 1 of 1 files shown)


F/finufft/build_tarballs.jl
```diff
@@ -6,38 +6,28 @@ using BinaryBuilderBase
 include(joinpath(@__DIR__, "..", "..", "platforms", "microarchitectures.jl"))
 
 name = "finufft"
-version = v"2.2.0"
+version = v"2.3.0"
```
giordano, line 9: Why changing the version number if you aren't changing the source?

jkrimmer: Late night mess-up, thanks for making me aware. Reverted to 2.2.0

PR #8125, opened 2024-02-18
Fix WapIt_jll build issue.

grasph:
The clang `BuildDependency` was missing the version specification, while the build assumes a specific one.

This has lead to build failure in CI after an update of the default version: https://buildkite.com/julialang/yggdrasil/builds/8260 .

The update script was tested locally.

Philippe.

### DIFF
--- a/W/WrapIt/build_tarballs.jl
+++ b/W/WrapIt/build_tarballs.jl
@@ -1,7 +1,6 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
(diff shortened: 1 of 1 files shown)


W/WrapIt/build_tarballs.jl
```diff
@@ -1,10 +1,14 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
-
+
```
giordano, line 4:
```suggestion

```

grasph:
thanks Mosè. -- Philippe.

PR #8134, opened 2024-02-20
[NodeJS] Add NodeJS v20 as a JLL

asinghvi17:
This commit duplicates the `NodeJS_18_jll` script for `build_tarballs.jl` and updates the hashes for the new executables.

This would be used initially by [DocumenterVitepress.jl](https://github.com/LuxDL/DocumenterVitepress.jl) to allow `makedocs` to build a fully functioning website without the user having to add extra steps.

### DIFF
--- a/N/NodeJS/NodeJS_20/build_tarballs.jl
+++ b/N/NodeJS/NodeJS_20/build_tarballs.jl
@@ -1,0 +1,69 @@
+using BinaryBuilder
+
+_name = "NodeJS_20"
+version = v"20.11.1"
+
+url_prefix = "https://nodejs.org/dist/v$version/node-v$version"
+sources = [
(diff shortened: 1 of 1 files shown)


N/NodeJS/NodeJS_20/build_tarballs.jl
```diff
@@ -0,0 +1,69 @@
...
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("node", :node),
+    FileProduct("bin/npm", :npm),
+    FileProduct("bin/npx", :npx),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
```
giordano, line 65:
```suggestion
dependencies = Dependency[
```

asinghvi17: Thanks!  Should I update the other scripts as well?

giordano: Probably not worth it now, if we aren't going to touch them in the future.

PR #8137, opened 2024-02-21
[ImageMagick] Update to 6.9.13

ViralBShah:
I am relaxing the compat on libtiff (which was set to v4.4), since it looks like [ubuntu and others are using libtiff >=4.5](https://packages.ubuntu.com/noble/libmagickcore-6.q16-7).

### DIFF
--- a/I/ImageMagick/ImageMagick@6/build_tarballs.jl
+++ b/I/ImageMagick/ImageMagick@6/build_tarballs.jl
@@ -67,4 +67,5 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="1.6", clang_use_lld=false)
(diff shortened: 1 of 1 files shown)


I/ImageMagick/ImageMagick@6/build_tarballs.jl
```diff
@@ -61,7 +61,7 @@ dependencies = [
     Dependency("Zlib_jll"),
     Dependency("libpng_jll"),
     Dependency("JpegTurbo_jll"),
-    Dependency("Libtiff_jll"; compat="4.3.0"),
+    Dependency("Libtiff_jll"),
```
giordano, line 64: You must keep compat bounds for libtiff: #3024. If anything, you can increase it to 4.5.1

ViralBShah: How about just going with `compat="4.6"`?

giordano: In general we want to have _more_ compat bounds, not less.  Not having compat bounds is a mess, if they turn out to be too tight they can be relaxed afterwards a little bit in the registry, but doing the opposite (tightening non-existing compat bounds) is much more complicated.

ViralBShah: Ok, so if I say `compat="4.5.1"`, will it pick that as the minumum and pick 4.6 etc?

giordano: > How about just going with compat="4.6"?

If you do `compat="4.5.1"` you are compatible with _both_ 4.5.1 _and_ 4.6.0 (v4.6.0 doesn't seem to have broken ABI: https://github.com/JuliaBinaryWrappers/Libtiff_jll.jl/commit/cb52abbbd649e4c39270355a57ebcbf3d57597ea, but breakages are generally unpredictable in advance)

PR #8155, opened 2024-02-23
add XicTools build script

kpobrien:
XicTools contains the electronic design editor Xic and the circuit simulator WRspice. This build_tarballs builds the Linux version without a GUI.

### DIFF
--- a/X/XicTools/build_tarballs.jl
+++ b/X/XicTools/build_tarballs.jl
@@ -103,4 +103,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
--- a/X/XicTools/build_tarballs.jl
+++ b/X/XicTools/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder
 
 name = "XicTools"
-version = v"0.1.0"
+version = v"4.3.19"
 
 # Collection of sources required to complete build
 sources = [
--- a/X/XicTools/build_tarballs.jl
+++ b/X/XicTools/build_tarballs.jl
@@ -25,29 +25,53 @@ make prefix="${prefix}/usr" install
 install_license ${WORKSPACE}/srcdir/xictools/license/LICENSE-2.0.txt
 cd ${prefix}/usr/xictools
 install -Dvm 755 wrspice.current/bin/mmjco "${bindir}/mmjco"
+rm wrspice.current/bin/mmjco
 install -Dvm 755 wrspice.current/bin/multidec "${bindir}/multidec"
+rm wrspice.current/bin/multidec
 install -Dvm 755 wrspice.current/bin/printtoraw "${bindir}/printtoraw"
+rm wrspice.current/bin/printtoraw
 install -Dvm 755 wrspice.current/bin/proc2mod "${bindir}/proc2mod"
+rm wrspice.current/bin/proc2mod
 install -Dvm 755 wrspice.current/bin/wrspice "${bindir}/wrspice"
+rm wrspice.current/bin/wrspice
 install -Dvm 755 wrspice.current/bin/wrspiced "${bindir}/wrspiced"
+rm wrspice.current/bin/wrspiced
 install -Dvm 755 bin/admsXml "${bindir}/admsXml"
+rm bin/admsXml
 install -Dvm 755 bin/busgen "${bindir}/busgen"
+rm bin/busgen
 install -Dvm 755 bin/capgen "${bindir}/capgen"
+rm bin/capgen
 install -Dvm 755 bin/cubegen "${bindir}/cubegen"
+rm bin/cubegen
 install -Dvm 755 bin/fastcap "${bindir}/fastcap"
+rm bin/fastcap
 install -Dvm 755 bin/fasthenry "${bindir}/fasthenry"
+rm bin/fasthenry
 install -Dvm 755 bin/fcpp "${bindir}/fcpp"
+rm bin/fcpp
 install -Dvm 755 bin/lstpack "${bindir}/lstpack"
+rm bin/lstpack
 install -Dvm 755 bin/lstunpack "${bindir}/lstunpack"
+rm bin/lstunpack
 install -Dvm 755 bin/mrouter "${bindir}/mrouter"
+rm bin/mrouter
 install -Dvm 755 bin/pipedgen "${bindir}/pipedgen"
+rm bin/pipedgen
 install -Dvm 755 bin/pyragen "${bindir}/pyragen"
+rm bin/pyragen
 install -Dvm 755 bin/vl "${bindir}/vl"
+rm bin/vl
 install -Dvm 755 xic.current/bin/wrdecode "${bindir}/wrdecode"
+rm xic.current/bin/wrdecode
 install -Dvm 755 xic.current/bin/wrencode "${bindir}/wrencode"
+rm xic.current/bin/wrencode
 install -Dvm 755 xic.current/bin/wrsetpass "${bindir}/wrsetpass"
+rm xic.current/bin/wrsetpass
 install -Dvm 755 xic.current/bin/xic "${bindir}/xic"
+rm xic.current/bin/xic
 install -Dvm 755 bin/zbuf "${bindir}/zbuf"
+rm bin/zbuf
 """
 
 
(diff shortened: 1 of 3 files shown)


X/XicTools/build_tarballs.jl
```diff
@@ -0,0 +1,106 @@
...
+dependencies = [
+    Dependency("Libtiff_jll"),
+    Dependency("libpng_jll"),
+    Dependency("JpegTurbo_jll"),
+    Dependency("GSL_jll"),
+    Dependency("CompilerSupportLibraries_jll"),
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
```
giordano, line 106:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```

X/XicTools/build_tarballs.jl
```diff
@@ -0,0 +1,106 @@
...
+	    ExecutableProduct("xic", :xic),
+	    ExecutableProduct("zbuf", :zbuf),
+	    ]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("Libtiff_jll"),
+    Dependency("libpng_jll"),
+    Dependency("JpegTurbo_jll"),
+    Dependency("GSL_jll"),
```
giordano, line 101: See #3024
```suggestion
    Dependency("GSL_jll; compat="~2.7.2""),
```

X/XicTools/build_tarballs.jl
```diff
@@ -0,0 +1,106 @@
...
+	    ExecutableProduct("vl", :vl),
+	    ExecutableProduct("wrdecode", :wrdecode),
+	    ExecutableProduct("wrencode", :wrencode),
+	    ExecutableProduct("xic", :xic),
+	    ExecutableProduct("zbuf", :zbuf),
+	    ]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("Libtiff_jll"),
```
giordano, line 98:
```suggestion
    Dependency("Libtiff_jll""; compat="4.5.1"),
```

X/XicTools/build_tarballs.jl
```diff
@@ -0,0 +1,106 @@
...
+sed -i "s|enable-itopok=no|enable-itopok=yes|g" Makefile
+sed -i "s|GFXLOC = --enable-gtk2=yes|#GFXLOC = --enable-gtk2=yes|g" Makefile
+#sed -i "s|^SUBDIRS =.*|SUBDIRS = xt_base adms KLU vl wrspice|g" Makefile
+sed -i "766c\#include <malloc.h>" xt_base/malloc/malloc-2.8.6.c
+update_configure_scripts
+make config
+make all -j$(nproc)
+make prefix="${prefix}/usr" install
+install_license ${WORKSPACE}/srcdir/xictools/license/LICENSE-2.0.txt
+cd ${prefix}/usr/xictools
```
giordano, line 26: Should we delete the files that you're copying from here below? Otherwise we have duplicate files in the tarball, no?

giordano (commented):
Thanks for your contribution!

X/XicTools/build_tarballs.jl
kpobrien: Good point, thank you, yes, I do see those duplicated in the tarballs. I'll give deleting them a shot and test if the program works as expected (or try to get BinaryBuilder to recognize those executables without copying).

kpobrien: I made these changes, tested locally and it looks ok, and it appears to successfully build on buildkite and pass all the checks on github. If there is anything else to do on my end before merging, please let me know. Thanks!

X/XicTools/build_tarballs.jl
```diff
@@ -0,0 +1,130 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/xictools/
+cp Makefile.sample Makefile
+sed -i "s|enable-itopok=no|enable-itopok=yes|g" Makefile
+sed -i "s|GFXLOC = --enable-gtk2=yes|#GFXLOC = --enable-gtk2=yes|g" Makefile
+#sed -i "s|^SUBDIRS =.*|SUBDIRS = xt_base adms KLU vl wrspice|g" Makefile
+sed -i "766c\#include <malloc.h>" xt_base/malloc/malloc-2.8.6.c
```
giordano, line 20: This way of patching files with `sed` is very fragile, if the file is changed upstream then the line number doesn't match anymore and no one will tell you.  It's better to have proper patches and apply them with [`atomic_patch`](https://docs.binarybuilder.org/stable/build_tips/#utils_build_env), there are loads of similar examples in this repository.

X/XicTools/build_tarballs.jl
```diff
@@ -0,0 +1,130 @@
...
+update_configure_scripts
+make config
+make all -j$(nproc)
```
giordano, line 23: Please use the [env var `${nproc}`](https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables), not the coomand `nproc` which would give the wrong number of cores (`${nproc}` is set by BinaryBuilder and we use it to reduce parallelism, since we run multiple build jobs in parallel)

X/XicTools/build_tarballs.jl
```diff
@@ -0,0 +1,130 @@
...
+make prefix="${prefix}/usr" install
+install_license ${WORKSPACE}/srcdir/xictools/license/LICENSE-2.0.txt
+cd ${prefix}/usr/xictools
+install -Dvm 755 wrspice.current/bin/mmjco "${bindir}/mmjco"
```
giordano, line 27: This can probably be simplified with something like
```bash
for file in wrspice.current/bin/mmjco wrspice.current/bin/multidec ...; do
    install -Dvm 755 "${file}" "${bindir}/$(basename ${file})"
    rm "${file}"
done
```

X/XicTools/build_tarballs.jl
```diff
@@ -0,0 +1,130 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "XicTools"
+version = v"0.1.0"
```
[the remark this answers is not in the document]
kpobrien: There is a strong reason in this case because builds without a GUI (such as gtk2) were only introduced in the latest commit as discussed in this issue: https://github.com/wrcad/xictools/issues/18 . I plan to shift to the latest release as soon as they tag one.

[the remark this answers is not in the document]
giordano: I see, but it'd be nicer to have a version number "close" to upstream, like 4.3.18 or 4.3.19?

X/XicTools/build_tarballs.jl
kpobrien: fixed

kpobrien: Based on this PR https://github.com/wrcad/xictools/pull/7 it might be safer to not build in parallel at all. What do you think?

X/XicTools/build_tarballs.jl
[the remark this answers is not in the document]
kpobrien: Sure, no problem. I'll change to 4.3.19.

X/XicTools/build_tarballs.jl
kpobrien: Fixed with something similar to above. Also added symlinks to deleted files because I noticed BinaryBuilder warnings (and more importantly may cause breakage in the program).

X/XicTools/build_tarballs.jl
kpobrien: Nevermind, it looks like that race condition was fixed in the commit we are building.

X/XicTools/bundled/patches/malloc-2.8.6.c.patch
```diff
@@ -0,0 +1,11 @@
+--- ../xictools-master/xt_base/malloc/malloc-2.8.6.c	2024-02-18 20:19:53.000000000 -0500
++++ xt_base/malloc/malloc-2.8.6.c	2024-02-23 08:57:12.595927930 -0500
+@@ -763,7 +763,7 @@
+ /* #define HAVE_USR_INCLUDE_MALLOC_H */
+ 
+ #ifdef HAVE_USR_INCLUDE_MALLOC_H
+-#include "/usr/include/malloc.h"
```
giordano, line 7: Oh wow, this should really fixed upstream, hardcoding the full absolute path of a header file is nonsense.

X/XicTools/build_tarballs.jl
kpobrien: Totally agree, fixed this.

giordano:
Looks good now, thanks!

PR #8158, opened 2024-02-23
update libsingular_julia to 0.43.0

hannes14:
@fingolfin


fingolfin:
Fails to build the Julia 1.12 variants because cxxwrap is not yet updated; so I think we basically need to wait for PR #8159 to be in. Or if it is super urgent, then manually filter out 1.12 from the `julia_versions` list for the time being.

PR #8159, opened 2024-02-23
[libcxxwrap_julia] rebuild for new libjulia

fingolfin:
... to get support for Julia nightly (which is now 1.12).


fingolfin:
Ah, I wasn't even aware of that at all. *sigh*

So what's the hold up for CxxWrap 0.15, @barche ?

If that could be released soon, this PR here might be sufficient for us (if we then update Singular and Polymake to use the new CxxWrap).

In the meantime I guess we can temporarily restore libcxxwrap-julia 0.11.2, bump it to 0.11.3, then later switch it back to 0.12.x

barche:
Indeed, depending on what is needed downstream this rebuild should be for libcxxwrap_julia version `v"0.11.2"` or 0.12, but using 0.12 will require also the use of the `testjll` branch of CxxWrap. We still need to work out a few issues before the next CxxWrap base on 0.12 can be released.

barche:
> So what's the hold up for CxxWrap 0.15, @barche ?

The hold-up is this: https://github.com/JuliaInterop/libcxxwrap-julia/pull/145#issuecomment-1944326600

This distinction between const and non-const is actually quite important, but also breaking, so I'd like to still push it into the 0.12 series of libcxxwrap to avoid two breaking changes in a short interval of time.

PR #8165, opened 2024-02-25
[llama_cpp] Update to v0.0.16 (b2256, 2024-02-25)

VaiTon:


imciner2:
Build errors are only on ARM 6 and 7:

> [15:36:41] /workspace/srcdir/llama.cpp/ggml-quants.c: In function ‘ggml_vec_dot_iq3_s_q8_K’:
[15:36:41] /workspace/srcdir/llama.cpp/ggml-quants.c:9463:32: error: implicit declaration of function ‘vld1q_u8_x2’; did you mean ‘vld1q_u32’? [-Werror=implicit-function-declaration]
[15:36:41]  9463 |     const uint8x16x2_t mask1 = vld1q_u8_x2(k_mask1);
[15:36:41]       |                                ^~~~~~~~~~~
[15:36:41]       |                                vld1q_u32
[15:36:41] /workspace/srcdir/llama.cpp/ggml-quants.c:9463:32: error: invalid initializer
[15:36:41] /workspace/srcdir/llama.cpp/ggml-quants.c:9491:34: error: implicit declaration of function ‘vqtbl1q_u8’; did you mean ‘vtbl1_u8’? [-Werror=implicit-function-declaration]
[15:36:41]  9491 |             vs.val[1] = vandq_u8(vqtbl1q_u8(vs.val[0], mask1.val[1]), mask2);
[15:36:41]       |                                  ^~~~~~~~~~
[15:36:41]       |                                  vtbl1_u8


Do you know what compilers LLAMA supports? Because the NEON intrinsic `vld1q_u8_x2` that it is trying to use was only just being discussed on the GCC patches mailing list in early January: https://gcc.gnu.org/pipermail/gcc-patches/2024-January/641647.html.


Also, the error about `vqtbl1q_u8` seems to suggest this codepath isn't compatible with ARMv6 or ARMv7, because that intrinsic is listed as being A64 only: https://developer.arm.com/architectures/instruction-sets/intrinsics/vqtbl1q_u8. Does upstream still support ARMv6 and ARMv7?

giordano:
Sounds similar to https://github.com/ggerganov/llama.cpp/issues/4827 but with a different function?  With these issues coming up every other day and no practical use for armv{6,7}l platforms from a julia point of view I'm not sure it's worth keep building this package for those architectures.

imciner2:
Yea, that is the same error, just with the signed 8-bit function instead of the unsigned 8-bit function.

> With these issues coming up every other day and no practical use for armv{6,7}l platforms from a julia point of view I'm not sure it's worth keep building this package for those architectures.

Hard to say, because I think some of the popular Raspberry Pi OS versions are still ARMv7 even though the chip is an aarch64-capable one. I can't think of very many cases where ARMv6 would be used still though, so maybe we could sunset that set of binaries. I don't doubt though that the ARMv6/7 packages would probably be infrequently downloaded though.

 I wonder if this is a case where upstream doesn't have any ARMv7 CI that can catch these issues, so maybe sending a report upstream again and see if they want to continue supporting ARMv7.

imciner2:
Looking at upstream:

The first issue (missing `vld1q_u8_x2`) probably needs a patch like https://github.com/ggerganov/llama.cpp/pull/4828 to use `ggml_vld1q_u8_x2` instead. The problematic code was just add 2 days ago.

It appears the second issue (missing `vqtbl1q_u8`) is fixed by this PR: https://github.com/ggerganov/llama.cpp/pull/5711 (albeit with the comment `// NOTE: not tested` on the function that was added).

To get the patch for the latest one, we should try bumping the version to something released today.

svilupp:
Oh, very nice! I was just about to open a PR when I noticed this :) 

Btw. there is already release v2275 available, so it might fix some issues.

svilupp:
> Sounds similar to [ggerganov/llama.cpp#4827](https://github.com/ggerganov/llama.cpp/issues/4827) but with a different function? With these issues coming up every other day and no practical use for armv{6,7}l platforms from a julia point of view I'm not sure it's worth keep building this package for those architectures.

I'm fully onboard to temporarily disable the ARM6,7 builds until we have someone explicitly asking for it or until it becomes officially supported by llama.cpp.

Most of the llama.cpp functionality requires quite a bit of memory and compute, so it won't be a frequent use case for Raspberry Pi'rs (and frankly with the limited resources, ppl might be using llama.cpp directly and customize their builds).

svilupp:
Is anyone willing and capable to try the `ggml_*` fixes?
The llamacpp devs have asked us to do the PR, so it requires someone who can write some cpp.
I have no idea how to declare the functions.

Alternatively, I'd suggest to exclude ARM6,7 from this release. 

@VaiTon would you mind updating the PR, please? I've updated my suggestions to the latest release as of now.

VaiTon:
https://github.com/ggerganov/llama.cpp/commit/e25fb4b18fcedb9bed6be4585cf842e9a669b28b should have fixed this :)

svilupp:
I believe the above-linked fix still does NOT fix the support for ARM, so we should temporarily move ahead without it.

I've opened a new PR to get things moving - we're almost 600 versions behind master now.

New PR here: https://github.com/JuliaPackaging/Yggdrasil/pull/8271

PR #8167, opened 2024-02-25
New Recipe: rwkv_cpp v0.0.1

VaiTon:


R/rwkv_cpp/build_tarballs.jl
```diff
@@ -0,0 +1,67 @@
...
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "freebsd"; ),
+    Platform("i686", "windows"; ),
+    Platform("x86_64", "windows"; )
+]
```
giordano, line 50: What's the problem with the other platforms?

VaiTon:
For `i686`, `armv6l`, `armv7l`

```
[17:54:51] /workspace/srcdir/rwkv.cpp/rwkv.cpp:37:1: error: static assertion failed: File offsets should be 64-bit or else rwkv.cpp will not be able to load model files over 2 GB
```

For `aarch64`

```
[17:55:58] In file included from /workspace/srcdir/rwkv.cpp/ggml/src/ggml.c:3:0:
[17:55:58] /workspace/srcdir/rwkv.cpp/ggml/include/ggml/ggml.h:283:13: error: unknown type name ‘__fp16’
[17:55:58]      typedef __fp16 ggml_fp16_t;
[17:55:58]              ^
```

R/rwkv_cpp/build_tarballs.jl
VaiTon: For `i686`, `armv6l`, `armv7l`

```
[17:54:51] /workspace/srcdir/rwkv.cpp/rwkv.cpp:37:1: error: static assertion failed: File offsets should be 64-bit or else rwkv.cpp will not be able to load model files over 2 GB
```

For `aarch64`

```
[17:55:58] In file included from /workspace/srcdir/rwkv.cpp/ggml/src/ggml.c:3:0:
[17:55:58] /workspace/srcdir/rwkv.cpp/ggml/include/ggml/ggml.h:283:13: error: unknown type name ‘__fp16’
[17:55:58]      typedef __fp16 ggml_fp16_t;
[17:55:58]              ^
```

fingolfin: You say `i686` has error but it is actually in the list of platforms?

What about macOS and freebsd, why are they missing?

In general I would recommend to start with the full list of platforms, then filtering it down. So for example:
```
platforms = filter!(p -> arch(p) != "armv6l", platforms)
platforms = filter!(p -> arch(p) != "armv7l", platforms)
```
(It may also be possible to fix the first error by adding `-D_FILE_OFFSET_BITS=64` to `CPPFLAGS`).

PR #8181, opened 2024-02-26
[OpenJpeg] Allow libtiff >= 4.5.1 rather than pinning it to that version

ViralBShah:
Make it possible to use newer versions of Libtiff with openjpeg.

### DIFF
--- a/O/OpenJpeg/build_tarballs.jl
+++ b/O/OpenJpeg/build_tarballs.jl
@@ -46,4 +46,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", clang_use_lld=false)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"6")
--- a/O/OpenJpeg/build_tarballs.jl
+++ b/O/OpenJpeg/build_tarballs.jl
@@ -18,7 +18,8 @@ cd $WORKSPACE/srcdir/openjpeg/
 for f in ${WORKSPACE}/srcdir/patches/*.patch; do
     atomic_patch -p1 ${f}
 done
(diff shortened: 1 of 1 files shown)


O/OpenJpeg/build_tarballs.jl
```diff
@@ -37,10 +40,10 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="LittleCMS_jll", uuid="d3a379c0-f9a3-5b72-a4c0-6bf4d2e8af0f"))
-    Dependency(PackageSpec(name="libpng_jll", uuid="b53b4c65-9356-5827-b1ea-8c7a1a84506f"))
-    Dependency("Libtiff_jll"; compat="~4.5.1")
+    Dependency("LittleCMS_jll")
+    Dependency("libpng_jll")
+    Dependency("Libtiff_jll"; compat="4.5.1")
```
giordano, line 45: This won't work in the registry because you're _changing_ an existing compat bound.  You have first to fix it there and then you can merge this with `[skip ci]` (it's useless to actually make a new release)

ViralBShah: If I understood correctly, this is what you suggested: https://github.com/JuliaRegistries/General/pull/101698 

If I do `[skip ci]` in a new comment here and then merge, will it actually do the right thing given that binaries have been built in this PR?

ViralBShah:
[skip ci]

O/OpenJpeg/build_tarballs.jl
giordano: No, after https://github.com/JuliaRegistries/General/pull/101698 is merged, then you can merge this PR and add `[skip ci]` to the commit message of the merge commit: https://github.com/JuliaPackaging/Yggdrasil/blob/f6ee203984887233121ccfa578d87c65ebbd42b9/CONTRIBUTING.md#information-for-maintainers


> will it actually do the right thing given that binaries have been built in this PR?

What I'm suggesting is to avoid creating a new release with new binaries, since nothing should have actually changed, no?

ViralBShah: Right, I understood what I was doing wrong. I just wasn't sure how exactly I should avoid the binaries from being built. I got it now - thanks.

PR #8187, opened 2024-02-26
[COSTA] New build recipe

abussy:
This is a build recipe for the COSTA (Distributed Communication-Optimal Shuffle and Transpose Algortihm) library: https://github.com/eth-cscs/COSTA.

Note the following line in the script: https://github.com/abussy/Yggdrasil/blob/12c14cbdc661d70f9de2d7365a142cf087336ffe/C/COSTA/build_tarballs.jl#L19. This is a hack in order to be able to build with the native CMake 3.21.7 version despite the project requiring 3.22. It has proven impossible to update CMake in the usual way (see Issue: https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1317).

### DIFF
--- a/C/COSTA/build_tarballs.jl
+++ b/C/COSTA/build_tarballs.jl
@@ -1,0 +1,93 @@
+# Note that this script can accept some limited command-line arguments, run                          
+# `julia build_tarballs.jl --help` to see a usage message.                                           
+using BinaryBuilder, Pkg                                                                             
+using Base.BinaryPlatforms                                                                           
+const YGGDRASIL_DIR = "../.."                                                                        
+include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
+
+name = "COSTA"
+version = v"2.2.2"
+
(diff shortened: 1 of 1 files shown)


C/COSTA/build_tarballs.jl
```diff
@@ -0,0 +1,94 @@
...
+sources = [
+   GitSource("https://github.com/eth-cscs/COSTA/", "bb84528d023db9a6b00ad729fb44b8c3cef8c981")
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir
+
+#Hack, because very hard to change CMake version in here
+sed -i 's/VERSION 3.22 FATAL_ERROR/VERSION 3.21.7 FATAL_ERROR/' ./CMakeLists.txt
+
```
giordano, line 20: Can remove it now, right?
```suggestion
```

abussy: Yes, indeed.

PR #8188, opened 2024-02-26
[VMEC] Version 1.3.0

benjaminfaber:


giordano:
Not a blocker, but this PR seems to have lots of unrelated commits.  You should delete branches after the corresponding PRs are merged, instead of keeping merging target branch of top of them, that creates lots of unnecessary noise.

benjaminfaber:
Thanks, Mosé, I will do that from now on.

PR #8195, opened 2024-02-27
[Pango] Add `libpangofc` as a LibraryProduct to Pango_jll

asinghvi17:
I'm not entirely sure how to create a version bump like "v1.1.1+1` here, but this is needed since libpangoft2 is deprecated in favour of libpangofc.


giordano:
Doesn't look like this library is built at all?

imciner2:
> I'm not entirely sure how to create a version bump like "v1.1.1+1` here

That is done automatically by the builder when it is merged.

And, I don't see that anywhere in the upstream git repo: https://gitlab.gnome.org/search?search=libpangofc&nav_source=navbar&project_id=1515&group_id=8&search_code=true&repository_ref=main

PR #8206, opened 2024-02-28
[SDPA] rebuild with fixed compat

odow:


giordano:
I should write thisddown in the CONTRIBUTING.md file as I keep repeating it very often, but if compat bounds are important they must be fixed in the registry first. The registry is what really matters. Also, if you add a new version with compat bounds without fixing the registry the package manager is still free to give you incompatible combination of packages, if for some reasons it prefers the newer version of the other package, downgrading the jll to a version without compat bounds.

Once this is fixed in the registry then we can merge this PR with [skip ci], as an actual new version would be useless.

odow:
Yeah I get you. Although in this case, we're going to pin the version of SDPA_jll in SDPA.jl, which is the only real consumer of this.

I always forget. Can I make this a `+1` build if we're adding new compat bounds? Ideally, the current build should never be used.

S/SDPA/build_tarballs.jl
```diff
@@ -27,7 +27,7 @@ julia_compat = join("~" .* string.(getfield.(julia_versions, :major)) .* "." .*
 
 name = "SDPA"
 upstream_version = v"7.3.17"
-version_offset = v"0.0.0" # reset to 0.0.0 once the upstream version changes
+version_offset = v"0.0.1" # reset to 0.0.0 once the upstream version changes
```
odow, line 30:
```suggestion
version_offset = v"0.0.0" # reset to 0.0.0 once the upstream version changes
```

barche:
Note that this now builds (correctly) against libcxxwrap 0.11 instead of libcxxwrap 0.12, so I don’t think you can skip ci here.

fingolfin:
While you are at it, I recommend that you remove the definition of `julia_versions`. Then it will switch to the default list, meaning you get newer Julia versions automatically -- right now you are missing out on 1.11 and 1.12. Both are still unreleased, of course, but the former already branched off so a release is not so far away, and the latter allows running CI tests again Julia nightly. Plus you can't forget to enable new versions this way.

odow:
So like this? I'll freely admit I have no idea what I'm doing with this libcxxwrap stuff.

S/SDPA/build_tarballs.jl
```diff
@@ -131,9 +127,11 @@ platforms = expand_cxxstring_abis(platforms)
 platforms = expand_gfortran_versions(platforms)
 filter!(p -> libgfortran_version(p) >= v"4", platforms)
 
+julia_compat = join("~" .* string.(getfield.(julia_versions, :major)) .* "." .* string.(getfield.(julia_versions, :minor)), ", ")
+
```
fingolfin, line 131: This can also go if you change the last line of the file from
```
    julia_compat)
```
to
```
    julia_compat = "1.6")
```

S/SDPA/build_tarballs.jl
```diff
@@ -131,9 +127,11 @@ platforms = expand_cxxstring_abis(platforms)
...
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("libcxxwrap_julia_jll"),
+    Dependency("libcxxwrap_julia_jll"; compat="=0.11"),
```
fingolfin, line 134:
```suggestion
    Dependency("libcxxwrap_julia_jll"; compat="~0.11.2"),
```

odow:
I think this is good to merge?

PR #8209, opened 2024-02-28
[systemd] Bump to v255

jeremiahpslewis:


S/systemd/build_tarballs.jl
```diff
@@ -23,7 +23,6 @@ cd systemd
 install_license LICENSE.GPL2
 
 # build-time dependencies that aren't packaged as JLLs
-apk add coreutils
 pip install jinja2
 
 meson --cross-file=${MESON_TARGET_TOOLCHAIN} build \
```
fingolfin, line 28: The build logs have lots of warnings from meson, e.g.:
```
[08:04:39] DEPRECATION: Option 'nss-resolve' value 'false' is replaced by 'disabled'
[08:04:39] DEPRECATION: Option 'nss-mymachines' value 'false' is replaced by 'disabled'
```
and
```
[08:04:39] WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
```

fingolfin: The actual error is
```
[08:04:39] meson.build:28:23: ERROR: Command `/usr/bin/realpath --relative-to=/workspace/srcdir/systemd/build /workspace/srcdir/systemd` failed with status 1.
```
but I can't tell from that if that means `/usr/bin/realpath ` is missing or it gave an error or something else

imciner2:
We have systemd 256.7 from https://github.com/JuliaPackaging/Yggdrasil/pull/9679, so closing this PR.

PR #8210, opened 2024-02-28
Update CoverM to 0.7.0

jakobnissen:
This update has bumped dependencies, which should allow it to compile on aarch64 Linux.

### DIFF
--- a/C/CoverM/build_tarballs.jl
+++ b/C/CoverM/build_tarballs.jl
@@ -26,7 +26,7 @@ platforms = filter!(supported_platforms()) do platform
     # Windows can't compile the dep htslib-rust for now, and I don't care about
     # other platforms
     (Sys.isapple(platform) || Sys.islinux(platform)) &&
-    in(arch(platform), ("x86_64", "aarch64")) &&
+    in(arch(platform), ("x86_64", "aarch64"))
 end
 
 # The products that we will ensure are always built
--- a/C/CoverM/build_tarballs.jl
+++ b/C/CoverM/build_tarballs.jl
@@ -8,8 +8,8 @@ version = v"0.7.0"
 # Collection of sources required to complete build
 sources = [
     GitSource(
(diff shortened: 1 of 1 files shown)


C/CoverM/build_tarballs.jl
```diff
@@ -3,13 +3,13 @@
...
 name = "CoverM"
-version = v"0.6.1"
+version = v"0.7.0"
 
 # Collection of sources required to complete build
 sources = [
     GitSource(
-        "https://github.com/wwood/CoverM",
-        "0056336d4b245064a23e5b40197b29607fb1fe56"),
+        "https://github.com/jakobnissen/CoverM",
```
giordano, line 11: Why a fork, instead of applying patches?

jakobnissen: This fork has downgraded a bunch of packages to make it work with Rust 1.73. I don't think the author wants that on the main repo.

giordano: Is the problem that we use an old (as in less than 6-month old) version of Rust?  We can update that, if that helps.

jakobnissen: Yes, that should fix it! :) I did consider trying to make a PR to BB, but it does seem excessive to have to keep up with Rust's many releases.
Nonetheless, if you have time to either upgrade Rust, or teach me how to, that'd be great!

jakobnissen: @giordano If you have time to upgrade the Rust toolchain to the latest version that's great, else having this functional patch out is the second best.

giordano: We have rustc 1.77.1 now.

giordano: If only Rust projects weren't aggressively requiring versions of rust more recent than 6 months ago... :smiling_face_with_tear:

PR #8213, opened 2024-02-28
Wizard recipe: SoftHSM2-v2.6.1

el-oso:

* Package name: SoftHSM2
* Version: v2.6.1

### DIFF
--- a/S/SoftHSM2/build_tarballs.jl
+++ b/S/SoftHSM2/build_tarballs.jl
@@ -29,7 +29,7 @@ cmake --install build
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(supported_platforms())
+platforms = expand_cxxstring_abis(supported_platforms(; exclude=Sys.iswindows))
 
 
 # The products that we will ensure are always built
--- a/S/SoftHSM2/build_tarballs.jl
+++ b/S/SoftHSM2/build_tarballs.jl
@@ -1,0 +1,58 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "SoftHSM2"
+version = v"2.6.1"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://dist.opendnssec.org/source/softhsm-$(version).tar.gz", "61249473054bcd1811519ef9a989a880a7bdcc36d317c9c25457fc614df475f2")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd softhsm-2.6.1/
+export LDFLAGS="-L${libdir}"
+export CPPFLAGS="-I${includedir}"
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j${nprocs}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "freebsd"; )
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libsofthsm2", :libsofthsm2, "lib/softhsm"),
+    ExecutableProduct("softhsm2-keyconv", :softhsm2_keyconv),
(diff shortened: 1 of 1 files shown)


el-oso:
not compiling in Windows 64-bit. Any ideas on how to fix that?

S/SoftHSM2/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+sources = [
+    ArchiveSource("https://dist.opendnssec.org/source/softhsm-$(version).tar.gz", "61249473054bcd1811519ef9a989a880a7bdcc36d317c9c25457fc614df475f2")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd softhsm-2.6.1/
+export LDFLAGS="-L${libdir}"
+export CPPFLAGS="-I${includedir}"
```
giordano, line 18: I don't think we need to set these flags nowadays
```suggestion
cd $WORKSPACE/srcdir/softhsm-*
```

S/SoftHSM2/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "freebsd"; )
+]
```
giordano, line 41: > not compiling in Windows 64-bit. Any ideas on how to fix that?

Let's see what's the error
```suggestion
platforms = expand_cxxstring_abis(supported_platforms())
```

S/SoftHSM2/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+products = [
+    LibraryProduct("libsofthsm2", :libsofthsm2, "lib/softhsm"),
+    ExecutableProduct("softhsm2-keyconv", :softhsm2_keyconv),
+    ExecutableProduct("softhsm2-util", :softhhsm2_util),
+    ExecutableProduct("softhsm2-dump-file", :softhsm2_dump_file)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"))
```
giordano, line 54:
```suggestion
    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="3.0.8")
```

S/SoftHSM2/build_tarballs.jl
```diff
@@ -0,0 +1,40 @@
...
+    ExecutableProduct("softhsm2-dump-file", :softhsm2_dump_file)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="3.0.8")
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
giordano, line 40: I think for Musl we need GCC v6:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
               julia_compat="1.6", preferred_gcc_version=v"6")
```

S/SoftHSM2/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = expand_cxxstring_abis(supported_platforms())
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libsofthsm2", :libsofthsm2, "lib/softhsm"),
```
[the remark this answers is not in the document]
el-oso: This package indeed has a shared library. It is a software emulated for PKCS11 compliant devices like TPMs. The workflow is that you use this .so for testing, most of the time that is all you need.
It requires to setup a small database with cryptographic keys.
It also has some other binaries to interact with such database and query the emulated device.

giordano:
Windows support is complicated:

* the autotools build doesn't seem to support Windows at all for the way it works
* there's a cmake build system, but it tries to run tons of `try_run` in https://github.com/opendnssec/SoftHSMv2/blob/913e7bfd463194fadcdd28f578087cc9c15643ee/cmake/modules/CompilerOptions.cmake, which is incompatible with cross-compilation.

I don't see any easy option related to Windows (short of them supporting cross-compilation).

el-oso:
Yes, I see the problem. Well, thanks for at least trying. For now we can make a binary for the platforms where it works and will try to solve it for the other ones at a later time.

el-oso:
What else needs to be done to merge and register this? 
I accepted that windows will not work for now

S/SoftHSM2/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir/softhsm-*
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j${nprocs}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = expand_cxxstring_abis(supported_platforms())
```
giordano, line 23: To exclude windows:
```suggestion
platforms = expand_cxxstring_abis(supported_platforms(; exclude=Sys.iswindows))
```

giordano:
Ok, the problem with the wrong extension on macOS is because they hardcode the `.so` extension in https://github.com/opendnssec/SoftHSMv2/blob/913e7bfd463194fadcdd28f578087cc9c15643ee/configure.ac with no way to override that (short of editing the file).  The CMake build system seems to handle that correctly (probably they just not hardcode the extension anywhere and let CMake infer it correctly).

el-oso:
Thanks for the help!

PR #8218, opened 2024-02-29
[ginkgo] Add CUDAsupport

upsj:
This adds CUDA support to the compilation. Due to the structure of Ginkgo, this doesn't need to be a runtime dependency, as we can also use the stub library instead.

Still some TODOs before I can mark this ready to review.

- [ ] Check for consistency in `libginkgo.so` stub and non-stub build
- [ ] Test interoperability between stub and backend libraries
- [ ] HIP support

### DIFF
--- a/G/ginkgo/build_tarballs.jl
+++ b/G/ginkgo/build_tarballs.jl
@@ -1,6 +1,13 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
-using BinaryBuilder, Pkg
+using BinaryBuilder, Pkg, BinaryBuilderBase
+using Base.BinaryPlatforms
+
+using Base.BinaryPlatforms: arch, os, tags
+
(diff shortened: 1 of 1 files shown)


G/ginkgo/build_tarballs.jl
```diff
@@ -12,30 +19,72 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
+# nvcc writes to /tmp, which is a small tmpfs in our sandbox.
+# make it use the workspace instead
+export TMPDIR=${WORKSPACE}/tmpdir
+mkdir ${TMPDIR}
+export PATH=$PATH:${prefix}/cuda/bin
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${prefix}/cuda/lib64
```
upsj, line 27:
```suggestion
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${prefix}/cuda/lib
```

PR #8224, opened 2024-03-01
[SCIP_PaPILO] Update version to scip 9

matbesancon:


matbesancon:
Oh no the "too many sections" error again:
https://buildkite.com/julialang/yggdrasil/builds/8779#018df958-87b5-45a9-a93b-e2d573070caf/6-9183
@giordano do you know if there is something actionable in these cases?

imciner2:
I think you can try adding
```
-Wa,-mbig-obj
```
to the compiler command line for mingw builds to get it to use a larger object size and not be limited.

matbesancon:
Thanks! Trying now

S/SCIP_PaPILO/build_tarballs.jl
```diff
@@ -4,25 +4,23 @@ using BinaryBuilder, Pkg
...
 cd scipoptsuite*
 
-# for soplex threadlocal
-export CXXFLAGS="-DTHREADLOCAL=''"
-
-# can be removed for scip 805
-echo "target_link_libraries(clusol gfortran)" >> papilo/CMakeLists.txt
-
+# for soplex threadlocal and large files
+export CXXFLAGS="-DTHREADLOCAL='' -Wa,-mbig-obj"
```
imciner2, line 22: I think you need to make adding the `-Wa,-mbig-obj` only added on the mingw builds.

imciner2, line 22:
```suggestion
# for soplex threadlocal
export CXXFLAGS="-DTHREADLOCAL=''"

# Enable large file support on mingw
if [[ "${target}" == *-mingw* ]]; then
    export CXXFLAGS="-Wa,-mbig-obj"
fi
```

matbesancon:
thanks!

imciner2:
Well, the error has at least changed, it is now a bunch of gfortran library linking errors, e.g.

```
[22:49:01] /opt/i686-w64-mingw32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld: ../libclusol.a(lusol6b.f.obj):lusol6b.f:(.text+0x530): undefined reference to `_gfortran_st_write'
[22:49:01] /opt/i686-w64-mingw32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld: ../libclusol.a(lusol6b.f.obj):lusol6b.f:(.text+0x552): undefined reference to `_gfortran_transfer_integer_write'
[22:49:01] /opt/i686-w64-mingw32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld: ../libclusol.a(lusol6b.f.obj):lusol6b.f:(.text+0x574): undefined reference to `_gfortran_transfer_integer_write'
[22:49:01] /opt/i686-w64-mingw32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld: ../libclusol.a(lusol6b.f.obj):lusol6b.f:(.text+0x59d): undefined reference to `_gfortran_transfer_integer_write'
[22:49:01] /opt/i686-w64-mingw32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld: ../libclusol.a(lusol6b.f.obj):lusol6b.f:(.text+0x5a5): undefined reference to `_gfortran_st_write_done'
...
```

I haven't dug into the CMake of the project, but based on this https://github.com/trilinos/Trilinos/issues/8632, it is missing the `gfortran` library in the linker flags for the executable.

imciner2:
So... in today's edition of banging my head against the wall while fighting we CMake, we have a doozy.

Basically, the link line for all the papilo executables on mingw was missing the `lgfortran` library, and since they were being compiled with `gcc` instead of `gfortran` since papilo is mainly a C/C++ library, the compiler wasn't adding that on its own. CMake is supposed to add that through what is known as an "implicit link library", where when it sees it is linking Fortran object files to C/C++ object files using the C/C++ compiler it automatically adds the relevant Fortran libraries (and vice-versa if it links C/C++ with the Fortran compiler).

However, this was only working for the unit test executables and the soplex executables. The reason: the location in the file where SCIP enabled the Fortran language support was inside the `papilo/CMakeLists.txt` file when creating the `clusol` target.

In CMake, many things are only applied to the current directory and its children, and not to its parent, which means that enabling Fortran in the `papilo/` directory enabled it for all the targets in that CMake file and the CMake files that it includes directly (the various libraries and the unit tests). The problem is, papilo isn't building the binaries using the codepath in the `papilo/CMakeLists.txt` file, instead it uses the top-level `CMakeLists.txt` file to directly load `papilo/binaries/CMakeLists.txt` - which means Fortran language support isn't present when creating the binary targets (since it wasn't enabled at the top-level CMake), and so CMake has no idea what the flags for Fortran are and just doesn't put any - so we end up with the missing link libraries.

The `soplex` directory was actually immune to this because it uses the `find_package(papilo)` system to load the config definition for papilo, which will enable the language properly for all uses of the target in that directory.

So to fix this, we have to enable the Fortran language before creating the binary targets in `papilo/binaries/CMakeLists.txt` if the `clusol` target is going to be used. I added a patch that does that, and it seems to have worked on my local build, so now let's see how it works on the rest of them.

As for why this seems to have only affected the mingw builds and not the linux builds with gcc/gfortran - :shrug:.

imciner2:
Ok, all the builds seem to be working in the CI here now. @matbesancon anything else you want to do here, or is this good to merge?

imciner2:
There may be better ways to fix it upstream actually, but this was the easiest to do as a patch in Yggdrasil.

Perhaps a better way to fix it upstream would be to use the same mechanism as soplex for loading the targets in the binaries subfolder when the `PAPILO_NO_BINARIES` flag is set to `ON` (that is the case where this is breaking right now).

PR #8230, opened 2024-03-04
[OpenFOAM_com] Add package environment variables to `init_block`

thealanjason:
Moving etc folder to `artifact_dir` and pointing `WM_PROJECT_DIR` to `artifact_dir`.  With this, the package could be used directly without setting any ENV on target machine.


thealanjason:
@giordano 
Could you please have a look and merge :)

PR #8234, opened 2024-03-05
[cppcheck] init at v2.13.0 

sjkelly:
This adds a build for cppcheck: https://github.com/danmar/cppcheck
cppcheck is useful for static analysis of C/C++ code. From the manual:

> Cppcheck is an analysis tool for C/C++ code. It provides unique code analysis
> to detect bugs and focuses on detecting undefined behaviour and dangerous
> coding constructs. The goal is to detect only real errors in the code, and generate
> as few false positives (wrongly reported warnings) as possible. Cppcheck is
> designed to analyze your C/C++ code even if it has non-standard syntax, as is
> common in for example embedded projects.

### DIFF
--- a/C/cppcheck/build_tarballs.jl
+++ b/C/cppcheck/build_tarballs.jl
@@ -39,7 +39,7 @@ products = Product[
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[
+dependencies = [
     Dependency("PCRE_jll")
 ]
 
--- a/C/cppcheck/build_tarballs.jl
+++ b/C/cppcheck/build_tarballs.jl
@@ -40,7 +40,7 @@ products = Product[
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("PCRE_jll")
+    Dependency("PCRE_jll"; compat="8.44.0")
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/C/cppcheck/build_tarballs.jl
+++ b/C/cppcheck/build_tarballs.jl
@@ -31,7 +31,7 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(exclude=[Platform("x86_64", "freebsd"; )])
+platforms = supported_platforms(exclude=Sys.isfreebsd)
 
 # The products that we will ensure are always built
 products = Product[
--- a/C/cppcheck/build_tarballs.jl
+++ b/C/cppcheck/build_tarballs.jl
@@ -31,7 +31,7 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(exclude=Sys.isfreebsd)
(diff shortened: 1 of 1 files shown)


imciner2:
It looks like you need to patch out the `-pedantic` flag in the build system to work around the MUSL bug (https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/387).

C/cppcheck/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+      -DBUILD_TESTS=OFF \
+      -DHAVE_RULES=ON \
+      ..
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(exclude=[Platform("x86_64", "freebsd"; )])
```
giordano, line 34:
```suggestion
platforms = supported_platforms(exclude=Sys.isfreebsd)
```

C/cppcheck/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+
+# The products that we will ensure are always built
+products = Product[
+    ExecutableProduct("cppcheck", :cppcheck),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
```
giordano, line 42:
```suggestion
dependencies = [
```

C/cppcheck/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+    Dependency("PCRE_jll")
```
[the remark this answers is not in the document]
sjkelly: Hmm, I thought since Julia uses PCRE also this would be restrictive?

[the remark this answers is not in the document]
giordano: Julia doesn't use `PCRE_jll`.

C/cppcheck/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+      -DBUILD_TESTS=OFF \
+      -DHAVE_RULES=ON \
+      ..
+make -j${nproc}
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(exclude=Sys.isfreebsd)
```
giordano, line 34:
```
┌ Warning: /cache/build/yggy-amdci7-8/julialang/yggdrasil/C/cppcheck/build/x86_64-linux-gnu/6IKqoQva/x86_64-linux-gnu-libgfortran4-cxx11/destdir/bin/cppcheck contains std::string values!  This causes incompatibilities across the GCC 4/5 version boundary.  To remedy this, you must build a tarball for both GCC 4 and GCC 5.  To do this, immediately after your `platforms` definition in your `build_tarballs.jl` file, add the line:
│ 
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/L8tBh/src/auditor/compiler_abi.jl:247
```
```suggestion
platforms = expand_cxxstring_abis(supported_platforms(exclude=Sys.isfreebsd))
```

PR #8242, opened 2024-03-06
Update casacorecxx 0.2.1 -> 0.3.0

torrance:


fingolfin:
@torrance if you update this JLL again, may I suggest getting rid of the custom `julia_versions` here, so that it automatically makes use of the default one provided by libjulia_jll and hence automatically supports all versions available? Then also e.g. Julia dev versions can already be tested.

Also the custom `julia_compat` could just be deleted, replacing it by `julia_compat="1.6"` in the `build_tarballs` call.

Issue #7904, opened 2024-01-04
Not a valid application error on windows


barche:
On PR's #2339 and #7899 the executables or DLL for Windows result in "not a valid Win32 application" errors. The problem comes from the `libgcc.a`, `libgcc_s.a` and `libmsvcrt.a` that is placed in `/workspace/destdir/lib` by CompilerSupportLibraries_jll. I'm not sure under which circumstances it is triggered (it seems quite rare), but when this occurs removing these files solves the problem.

sloede:
AFAICT, the location of the offending libraries depends on the Julia version. For the moment, it is
* no libraries for Julia <= 1.8
* `$prefix/lib/` for Julia 1.9 and 1.10
* `$prefix/lib/gcc/<triple>/<major>/` for Julia >= 1.11
where `<triple>` is the usual BB triplet (e.g., `x86_64-w64-mingw32`) and `<major>` is the GCC compiler major version (e.g., `13`) (see also #7535).

vtjnash:
So this should be fixed for Julia >= 1.11 I think, since we now hide those `.a` files in subfolders

giordano:
I'm not entirely sure, but I believe https://github.com/JuliaPackaging/Yggdrasil/blob/d4808d4916dab74e6ec758d54e3bef9bbf699de7/S/SDPA/build_tarballs.jl#L67-L69 was used to work around this problem.  Maybe @blegat can confirm.

odow:
That doesn't appear to have worked around the issue because we're now hitting this:
https://github.com/jump-dev/SDPA.jl/pull/57

odow:
Just confirming that https://github.com/JuliaPackaging/Yggdrasil/pull/8249 fixed this for SDPA.

https://github.com/jump-dev/SDPA.jl/actions/runs/8197171105/job/22418710526?pr=61

PR #2339, opened 2021-01-02
Add Gnuplot builder

VarLad:
This again is incomplete! 
@giordano Your kind help is yet again needed my friend!

PR #8249, opened 2024-03-07
[SDPA] fix Windows build

odow:
x-ref https://github.com/JuliaPackaging/Yggdrasil/issues/7904#issuecomment-1982683296

Code take from https://github.com/JuliaPackaging/Yggdrasil/pull/2339/


odow:
Now trying the code from https://github.com/JuliaPackaging/Yggdrasil/pull/7899

PR #8254, opened 2024-03-08
[Glib] Update to 2.78.3

jwahlstrand:


giordano:
My understanding is that we don't build with `--enable-sdk=all`: https://github.com/JuliaPackaging/Yggdrasil/blob/39e47bb370e24fcf7b5e2ac7f1712c63b273a7d7/0_RootFS/gcc_common.jl#L419-L439 But I don't see why we shouldn't do it in our toolchain.

jwahlstrand:
Would it be possible to get this merged before #8274? In version 2.80 they pull libgirepository into GLib and I half expect trouble with gobject-introspection. No big deal if we jump straight to 2.80, I will make do.

PR #8257, opened 2024-03-08
Wizard recipe: libzenohc-v0.10.1

n0wis:

* Package name: libzenoh
* Version: v0.10.1

### DIFF
--- a/L/libzenohc/build_tarballs.jl
+++ b/L/libzenohc/build_tarballs.jl
@@ -12,8 +12,7 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir
-cd zenoh-c
+cd $WORKSPACE/srcdir/zenoh-c
 cmake -B build -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DZENOHC_CARGO_CHANNEL=$RUSTUP_TOOLCHAIN -DZENOHC_CUSTOM_TARGET=$CARGO_BUILD_TARGET
 cmake --build build --parallel ${nproc}
 cmake --install build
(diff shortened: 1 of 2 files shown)


L/libzenohc/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+products = [
+    LibraryProduct("libzenohc", :libzenohc)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", compilers = [:rust, :c], preferred_gcc_version = v"13.2.0")
```
giordano, line 43: Please read https://github.com/JuliaPackaging/Yggdrasil/blob/a9121f7c51adcc4006dba1095143a1ecf1193b2a/CONTRIBUTING.md#compatibility-tips

giordano (commented):
Thanks for your contribution!

L/libzenohc/build_tarballs.jl
n0wis: Thanks for your feedback!

L/libzenohc/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "freebsd"; )
+]
+
```
giordano, line 30: What's the problem with the other platforms?

n0wis: They did not compile successfully with the given cmake commands. Unfortunately, I don't have the capacity to follow through and check / fix the build for every platform. Taking i686-linux-gnu as an example, apparently there is a compilation problem with a dependency (see https://github.com/briansmith/ring/issues/1795).

Is supporting all platforms a requirement for merging into main?

giordano: No, it isn't mandatory, but I was curious because for the most part we should be able to build Rust for all platforms (except for 32-bit Windows)

PR #8259, opened 2024-03-08
[GR] bump version

jheinen:
bump GR run-time version to 0.73.3

### DIFF
--- a/G/GR/build_tarballs.jl
+++ b/G/GR/build_tarballs.jl
@@ -3,13 +3,13 @@
 using BinaryBuilder
 
 name = "GR"
-version = v"0.73.2"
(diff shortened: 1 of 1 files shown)


jheinen:
@giordano : I don't habe any idea why the macOS jobs fail? Our "own" binary builder CI pipelines passed without any errors!?

giordano:
Try to add the keyword argument `clang_use_lld=false` to the `build_tarballs(...)` function.

jheinen:
> Try to add the keyword argument `clang_use_lld=false` to the `build_tarballs(...)` function.

Unfortunately, the problem still persists ...

PR #8263, opened 2024-03-08
[GR] bump version

jheinen:
bump GR run-time version

PR #8268, opened 2024-03-09
[libpng] Disable building the framework on macOS

giordano:
Ref #8259, #8263, https://github.com/JuliaPackaging/Yggdrasil/pull/8110#issuecomment-1986447387


jheinen:
@giordano : Thanks for the fixes!!!

PR #8278, opened 2024-03-10
[GtkSourceView] update to 5.10.0

jwahlstrand:
Update to v5, which depends on GTK4.

### DIFF
--- a/G/GtkSourceView/build_tarballs.jl
+++ b/G/GtkSourceView/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "GtkSourceView"
-version = v"4.4.0"
+version = v"5.10.0"
 
 # Collection of sources required to build GtkSourceView
 sources = [
-    "https://download.gnome.org/sources/gtksourceview/4.4/gtksourceview-$(version).tar.xz" =>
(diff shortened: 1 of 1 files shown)


G/GtkSourceView/build_tarballs.jl
```diff
@@ -17,34 +17,47 @@ script = raw"""
...
+    LibraryProduct("libgtksourceview-5", :libgtksourceview),
 ]
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Glib_jll", v"2.59.0"; compat="2.59"),
-    Dependency("GTK3_jll"),
+    Dependency("Glib_jll", v"2.76.5"; compat="2.76"),
+    Dependency("GTK4_jll"),
+    Dependency("Cairo_jll"),
```
giordano, line 57: #3024
```suggestion
    Dependency("Cairo_jll"; compat="1.16.1"),
```

giordano (commented):
This will break the ABI of the package, do the packages depending on GtkSourceView fix the compat bounds correctly already?

jwahlstrand:
The only package in the general registry that uses this JLL is [GtkSourceWidget,](https://github.com/JuliaGtk/GtkSourceWidget.jl) and it does have compat bounds correctly set.

PR #8283, opened 2024-03-12
Wrap binaries for biber

### LINKED ISSUE #5633: build request: biber
Biber has pretty straightforward build scripts (e.g. https://github.com/plk/biber/blob/dev/dist/darwin_arm64/build.sh), and having a biber jll would allow Tectonic.jl to not vendor its own artifacts and fix https://github.com/MichaelHatherly/Tectonic.jl/issues/13

iagobaapellaniz:
UPDATE: Wrapper around binaries obtained from CTAN.ORG

Initially, even though it doesn't work, I think it's worth sharing this draft. Can we make it work somehow? It would close #5633

### DIFF
--- a/B/biber/build_tarballs.jl
+++ b/B/biber/build_tarballs.jl
@@ -34,9 +34,7 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd ${WORKSPACE}/srcdir/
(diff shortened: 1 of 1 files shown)


iagobaapellaniz:
Thank you very much @giordano 🎉 🎉 🎉

PR #8291, opened 2024-03-12
Add libdatachannel

VarLad:
~~Windows build seems to not be working because it apparently can't find OpenSSL~~

For some reason that only happens when building locally.

### DIFF
--- a/L/libdatachannel/build_tarballs.jl
+++ b/L/libdatachannel/build_tarballs.jl
@@ -6,7 +6,9 @@ name = "libdatachannel"
 version = v"0.20.2"
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/paullouisageneau/libdatachannel.git", "9cbe6a2a1f21cde901bca9571581a96c6cda03cf")
+    GitSource("https://github.com/paullouisageneau/libdatachannel.git", "9cbe6a2a1f21cde901bca9571581a96c6cda03cf"),
+    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.14.sdk.tar.xz", 
+               "0f03869f72df8705b832910517b47dd5b79eb4e160512602f593ed243b28715f"))
 ]
 
 # Bash recipe for building across all platforms
@@ -16,7 +18,6 @@ if [[ ${target} == x86_64*mingw* ]]; then
 fi
 
 if [[ "${bb_full_target}" == x86_64-apple-darwin* ]]; then
-    wget -qO- https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.14.sdk.tar.xz | tar -xJ
     # LLVM 15+ requires macOS SDK 10.14.
     pushd $WORKSPACE/srcdir/MacOSX10.*.sdk
     rm -rf /opt/${target}/${target}/sys-root/System
--- a/L/libdatachannel/build_tarballs.jl
+++ b/L/libdatachannel/build_tarballs.jl
@@ -22,6 +22,7 @@ mkdir -p ${prefix}/include/sys
 touch ${prefix}/include/sys/random.h
 cat >> ${prefix}/include/sys/random.h <<EOF
 #include <sys/syscall.h>
+#define _XOPEN_SOURCE 700
 #define getrandom(buf, sz, flags) syscall(SYS_getrandom, buf, sz, flags)
 #endif
 
--- a/L/libdatachannel/build_tarballs.jl
+++ b/L/libdatachannel/build_tarballs.jl
@@ -11,8 +11,7 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir
-cd libdatachannel
+cd $WORKSPACE/srcdir/libdatachannel
 git submodule update --init --recursive --depth 1
 cmake -B build -DUSE_GNUTLS=0 -DUSE_NICE=0 -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
 cmake --build build --parallel ${nproc}
--- a/L/libdatachannel/build_tarballs.jl
+++ b/L/libdatachannel/build_tarballs.jl
@@ -18,13 +18,15 @@ if [[ ${target} == x86_64*mingw* ]]; then
 fi
 
 if [[ ${target} == *musl* ]]; then
-    mkdir -p ${prefix}/include/sys
-    touch ${prefix}/include/sys/random.h
-    cat >> ${prefix}/include/sys/random.h <<EOF
-        #include <sys/syscall.h>
-        #define getrandom(buf, sz, flags) syscall(SYS_getrandom, buf, sz, flags)
-        #endif
-    EOF
+mkdir -p ${prefix}/include/sys
+touch ${prefix}/include/sys/random.h
+cat >> ${prefix}/include/sys/random.h <<EOF
+#include <sys/syscall.h>
+#define getrandom(buf, sz, flags) syscall(SYS_getrandom, buf, sz, flags)
+#endif
+
+EOF
+
 fi
 
 if [[ "${bb_full_target}" == x86_64-apple-darwin* ]]; then
--- a/L/libdatachannel/build_tarballs.jl
+++ b/L/libdatachannel/build_tarballs.jl
@@ -53,6 +53,8 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="3.0.8")
+    # Links to libgcc_s on linux for something
+    Dependency("CompilerSupportLibraries_jll"; platforms=filter(p -> Sys.islinux(p),  platforms))
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/L/libdatachannel/build_tarballs.jl
+++ b/L/libdatachannel/build_tarballs.jl
@@ -11,6 +11,9 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
+if [[ ${target} == x86_64*mingw* ]]; then
+    export OPENSSL_ROOT_DIR=$WORKSPACE/destdir/lib64
+fi
 cd $WORKSPACE/srcdir/libdatachannel
 git submodule update --init --recursive --depth 1
 cmake -B build \
--- a/L/libdatachannel/build_tarballs.jl
+++ b/L/libdatachannel/build_tarballs.jl
@@ -13,7 +13,12 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/libdatachannel
 git submodule update --init --recursive --depth 1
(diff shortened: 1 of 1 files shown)


VarLad:
x86-64-w64-mingw32 build seems to not be able to find OpenSSL.
For x86 MacOS, can we somehow use MacOS 10.14 toolkit to support the remaining few devices? :sweat_smile: 

As for musl, any idea what to do about the "`sys/random.h`" not found in musl libc variants?

L/libdatachannel/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libdatachannel", :libdatachannel)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"))
```
giordano, line 34:
```suggestion
    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="3.0.8")
```

L/libdatachannel/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+version = v"0.20.2"
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/paullouisageneau/libdatachannel.git", "9cbe6a2a1f21cde901bca9571581a96c6cda03cf")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd libdatachannel
```
giordano, line 15:
```suggestion
cd $WORKSPACE/srcdir/libdatachannel
```

giordano (commented):
> x86-64-w64-mingw32 build seems to not be able to find OpenSSL.

OpenSSL is under lib64 on windows

VarLad:
@imciner2 I believe that the musl builds can be fixed by
```
apk add musl-dev
```
and adding `/usr/include` to CMAKE's list of include dirs.
Any idea what command do I need to pass to do that?

giordano:
We do ***not*** want to use system header files.

VarLad:
@giordano Wouldn't runtime deps would already be present in musl systems so we should  be able to use the header files freely? :thinking:

If thats not acceptable, got any other ideas? Should I look into getting those header files into https://github.com/JuliaBinaryWrappers/Musl_jll.jl/ instead? (from what I can see it doesn't have the header files)

giordano:
> @giordano Wouldn't runtime deps would already be present in musl systems so we should be able to use the header files freely? 🤔

Header files may differ for different architectures for example, if we use the headers of the x86_64 system we'd have broken builds for all other arches.

L/libdatachannel/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+version = v"0.20.2"
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/paullouisageneau/libdatachannel.git", "9cbe6a2a1f21cde901bca9571581a96c6cda03cf")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+if [[ ${target} == x86_64*mingw* ]]; then
+    export OPENSSL_ROOT_DIR=$WORKSPACE/destdir/lib64
```
giordano, line 15:
```suggestion
    export OPENSSL_ROOT_DIR="${prefix}/lib64"
```

imciner2:
Hmm, it is odd it can't find the header, since that was added in Musl 1.1.19 or so, and we claim to be on Musl 1.2.2 in the buildroot, however when I do make a session there really isn't a `sys/random` inside the sys-root. One way to work around this might be to do something similar to this patch https://github.com/JuliaPackaging/Yggdrasil/blob/ff745311dc40be8fd3f046dc39237482c70b2c7b/T/TracyProfiler/bundled/patches/TracyProfiler-nfd-extended-1.0.2.patch#L32, where a test for the include is made and if the include doesn't exist the equivalent syscall formulation is used instead.

VarLad:
@imciner2 By buildroot I guess you mean the environment in which we run our commands?
If so, you can get the sys/random.h with `musl-dev` package :thinking:

I'm wondering why the Musl_jll library packaged by binary builder doesn't have header files.

giordano:
> I'm wondering why the Musl_jll library packaged by binary builder doesn't have header files.

Because no one is telling it to install any header files: https://github.com/JuliaPackaging/Yggdrasil/blob/ff745311dc40be8fd3f046dc39237482c70b2c7b/M/Musl/build_tarballs.jl#L48.  Our Musl header files come from the GCC build: https://github.com/JuliaPackaging/Yggdrasil/blob/ff745311dc40be8fd3f046dc39237482c70b2c7b/0_RootFS/gcc_common.jl#L382-L394

giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/ff745311dc40be8fd3f046dc39237482c70b2c7b/M/mlir_jl_tblgen/build_tarballs.jl#L91-L92
https://github.com/JuliaPackaging/Yggdrasil/blob/ff745311dc40be8fd3f046dc39237482c70b2c7b/M/mlir_jl_tblgen/build_tarballs.jl#L22-L30

VarLad:
@giordano 

- For MacOS, I'll try adding support for x86_64 in a future update
- As for musl, any idea for what the idea solution would be in this case? Can this be fixed from BinaryBuilder's side?

Also, any idea why does `x86-64-linux-gnu-cxx1` decide to fail with:

> ERROR: could not load library "/cache/build/yggy-amdci7-13/julialang/yggdrasil/L/libdatachannel/build/x86_64-linux-gnu-cxx11/XMdeXSJe/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libdatachannel.so.0.20.2"
/cache/julia-buildkite-plugin/julia_installs/bin/linux/x64/1.7/julia-1.7-latest-linux-x86_64/bin/../lib/julia/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by /cache/build/yggy-amdci7-13/julialang/yggdrasil/L/libdatachannel/build/x86_64-linux-gnu-cxx11/XMdeXSJe/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libdatachannel.so.0.20.2)
┌ Warning: lib/libdatachannel.so.0.20.2 cannot be dlopen()'ed

Is that a BinaryBuilder error?

VarLad:
Is the `wget` fix acceptable?
Lets see if we can get the musl thingy fixed trivially in a few hours.
If not, I'll go for it without musl for now.

giordano:
> Is the wget fix acceptable?

No, add the archive to `sources`

L/libdatachannel/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libdatachannel", :libdatachannel)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="3.0.8")
+]
```
imciner2, line 55:
```suggestion
    # Links to libgcc_s on linux for something
    Dependency("CompilerSupportLibraries_jll"; platforms=filter(p -> Sys.islinux(p),  platforms))
]
```
I am not sure exactly why it is complaining about linking against `libgcc_s` in the Linux build, but this adds that library into the buildroot.

@giordano is that a standard error we see on the auditor and usually ignore?

L/libdatachannel/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+git submodule update --init --recursive --depth 1
+cmake -B build \
+    -DUSE_GNUTLS=0 \
+    -DUSE_NICE=0 \
+    -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release
+cmake --build build --parallel ${nproc}
+cmake --install build
+"""
```
imciner2, line 40:
```suggestion
install_license $WORKSPACE/srcdir/libdatachannel/LICENSE
"""
```

The auditor was flagging a missing license.

L/libdatachannel/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+    LibraryProduct("libdatachannel", :libdatachannel)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="OpenSSL_jll", uuid="458c3c95-2e84-50aa-8efc-19380b2a3a95"); compat="3.0.8")
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"13.2.0")
```
imciner2, line 58: GCC version 13.2.0 is very restrictive, which is what causes this error:
```
ERROR: could not load library "/cache/build/yggy-amdci7-15/julialang/yggdrasil/L/libdatachannel/build/x86_64-linux-gnu-cxx11/5iGi6pG4/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libdatachannel.so.0.20.2"
/cache/julia-buildkite-plugin/julia_installs/bin/linux/x64/1.7/julia-1.7-latest-linux-x86_64/bin/../lib/julia/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by /cache/build/yggy-amdci7-15/julialang/yggdrasil/L/libdatachannel/build/x86_64-linux-gnu-cxx11/5iGi6pG4/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libdatachannel.so.0.20.2)
```

Here it is trying to use the C++ runtime library version 3.4.32, but that is the most recent one and so will only work on the most recent systems. The general rule is that libraries built in Yggdrasil should use the oldest possible GCC version to build with to ensure they can be used on the largest number of systems without compatibility issues.

I would suggest starting with GCC 7 and increasing the version number until you find a compiler version that works without error (we have a mix of GCC 7, 8 and 9 in use for C++17 programs, depending on the exact C++17 features they use).

L/libdatachannel/build_tarballs.jl
giordano: I usually ignore it because I'm lazy, but I believe `CompilerSupportLibraries_jll` as dependency is the correct course of action, yes.

VarLad: @imciner2 That should be `BuildDependency` right? 
(or does it not matter?)

imciner2: `Dependency` says it is needed at both the build and runtime, `BuildDependency` means it is only needed at build time. In this case, the library is needed at runtime as well (hence why it is an error in the auditor phase).

L/libdatachannel/build_tarballs.jl
VarLad: @imciner2 Wait, is the `CompilerSupportLibraries_jll` still needed if this works? :thinking:

giordano: `CompilerSupportLibraries_jll` is always needed if the library links to `libgcc_s`, regardless of the version.

VarLad:
@giordano Is this fix acceptable? 
If so, this is ready to be merged!

PR #8314, opened 2024-03-19
Wizard recipe: coreblasmkl-v23.8.2

Rabab53:

* Package name: coreblasmkl
* Version: v23.8.2


imciner2:
This library doesn't use any non-BLAS MKL routine, correct? If so, then I think it would be better to modify the coreblas builder in https://github.com/JuliaPackaging/Yggdrasil/pull/8335 to use libblastrampoline to switch between OpenBLAS and MKL without having to have a separate JLL package for each.

imciner2:
Replaced by https://github.com/JuliaPackaging/Yggdrasil/pull/8357, which uses libblastrampoline now.

PR #8317, opened 2024-03-19
[GAP_pkg_anupq] update for GAP 4.13.0

fingolfin:


vchuravy:
For next time. I think you can just open a single (or a few PRs) instead of opening a PR for every package.

fingolfin:
OK. I recalled that there is (was?) a limit on how many build jobs can be run for a single PR, and I was hitting that limit at some point. But presumably this is gone then?

PR #8337, opened 2024-03-20
Add Maximum Independent Set library

claud10cv:
Tarballs for MaximumIndependentSet (originally in C++)

### DIFF
--- a/M/MaximumIndependentSet/build_tarballs.jl
+++ b/M/MaximumIndependentSet/build_tarballs.jl
@@ -34,7 +34,7 @@ products = Product[
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[
 	Dependency("libcxxwrap_julia_jll"),
-	BuildDependency("libjulia_jll")
+	Dependency("libjulia_jll")
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/M/MaximumIndependentSet/build_tarballs.jl
+++ b/M/MaximumIndependentSet/build_tarballs.jl
@@ -3,7 +3,7 @@ using BinaryBuilder
 name = "MaximumIndependentSet"
 version = v"0.1.1"
 sources = [
-    GitSource("https://github.com/claud10cv/MaximumIndependentSet.git", "acf502d3e991952f554eda01c8b94a1114e50d02"),
+    GitSource("https://github.com/claud10cv/MaximumIndependentSet.git", "fe2e7ccc79f4de024e798c4daf9dcb3b8e99b906"),
 ]
 
 script = raw"""
--- a/M/MaximumIndependentSet/build_tarballs.jl
+++ b/M/MaximumIndependentSet/build_tarballs.jl
@@ -7,7 +7,7 @@ version = v"0.1.2"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/claud10cv/MaximumIndependentSet.git", "d16dda9b8cb2c1011ba0ba9fc3f68be36bbe1c7b")
+    GitSource("https://github.com/claud10cv/MaximumIndependentSet.git", "26faa2c3422c7b18f090d9865b16e272d9a6d90e")
 ]
 
 # Bash recipe for building across all platforms
--- a/M/MaximumIndependentSet/build_tarballs.jl
+++ b/M/MaximumIndependentSet/build_tarballs.jl
@@ -1,37 +1,35 @@
-using BinaryBuilder
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
 
 name = "MaximumIndependentSet"
-version = v"0.1.1"
+version = v"0.1.2"
+
+# Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/claud10cv/MaximumIndependentSet.git", "950f57e52d98dc511491a92eac12d487a427ed27"),
+    GitSource("https://github.com/claud10cv/MaximumIndependentSet.git", "d16dda9b8cb2c1011ba0ba9fc3f68be36bbe1c7b")
 ]
 
+# Bash recipe for building across all platforms
 script = raw"""
-cd ${WORKSPACE}/srcdir/MaximumIndependentSet/build
-cmake -B . -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
-cmake --build . --parallel ${nproc}
+cd $WORKSPACE/srcdir/MaximumIndependentSet/src/
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+cmake --build build --parallel ${nproc}
 cmake --install build
 """
 
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
 platforms = supported_platforms()
-#filter!(t -> os(t) in ["linux"], platforms)
-#filter!(t -> arch(t) == "x86_64", platforms)
-#filter!(t -> isnothing(libc(t)) || libc(t) == "glibc", platforms)
 
-products = [
-    LibraryProduct("libmis", :libmis),
+# The products that we will ensure are always built
+products = Product[
+	LibraryProduct(["libmis", "mislib"], :mis)
 ]
 
-dependencies = [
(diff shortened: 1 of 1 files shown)


M/MaximumIndependentSet/build_tarballs.jl
```diff
@@ -0,0 +1,36 @@
...
+cd ${WORKSPACE}/srcdir/MaximumIndependentSet
+make -j${nproc}
+make install
+"""
+
+platforms = supported_platforms()
+filter!(t -> os(t) in ["linux"], platforms)
+filter!(t -> arch(t) == "x86_64", platforms)
+filter!(t -> isnothing(libc(t)) ||  libc(t) == "glibc", platforms)
+
```
imciner2, line 19:
```suggestion
platforms = expand_cxxstring_abis(platforms)

```
The auditor reports that `std::string` is used, so we need to handle the ABI differences.

imciner2:
The upstream makefile seems to have `-lcxxwrap_julia` on the link line for the dynamic library, I don't see that on the command line for the Yggdrasil build you did (and you also don't include the `libcxxwrap_julia_jll` as a dependency). Are you packaging the most recent version of the library, or purposefully holding it back?

M/MaximumIndependentSet/build_tarballs.jl
```diff
@@ -0,0 +1,36 @@
...
+script = raw"""
+cd ${WORKSPACE}/srcdir/MaximumIndependentSet
+make -j${nproc}
+make install
+"""
+
+platforms = supported_platforms()
+filter!(t -> os(t) in ["linux"], platforms)
+filter!(t -> arch(t) == "x86_64", platforms)
+filter!(t -> isnothing(libc(t)) ||  libc(t) == "glibc", platforms)
```
imciner2, line 18: This is a slightly awkward way of just specifying the one platform, but it works. Out of curiosity though, is there a reason to not build on the other platforms?

claud10cv: The makefile is a hard-coded, hand-written one. It only works on linux for now. I will work on creating a cmake script to enable cross-compile later this week

M/MaximumIndependentSet/build_tarballs.jl
```diff
@@ -0,0 +1,36 @@
...
+build_tarballs(ARGS,
+               name,
+               version,
+               sources,
+               script,
+               platforms,
+               products,
+               dependencies,
+               julia_compat = "1.9";
+               preferred_gcc_version=v"11")
```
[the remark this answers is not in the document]
claud10cv: According to https://gcc.gnu.org/projects/cxx-status.html#cxx17 full support for c++17 is introduced as for g++-9. I will modify the line accordingly

M/MaximumIndependentSet/build_tarballs.jl
claud10cv: I am testing a cmake build and will see if I can get to coompile on more platforms

claud10cv:
> The upstream makefile seems to have `-lcxxwrap_julia` on the link line for the dynamic library, I don't see that on the command line for the Yggdrasil build you did (and you also don't include the `libcxxwrap_julia_jll` as a dependency). Are you packaging the most recent version of the library, or purposefully holding it back?

Originally I had considered adding the wrapper within the C++ lib but ultimately decided to decouple both things, so I removed all references to libcxxwrap. I will create a separate wrapper package

M/MaximumIndependentSet/build_tarballs.jl
claud10cv: The tarballs now build for all platforms

M/MaximumIndependentSet/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+cmake --build build --parallel ${nproc}
+cmake --install build
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; 
+	exclude=[Platform("i686", "linux"; libc = "musl")
+		]
+)
```
giordano, line 27: https://buildkite.com/julialang/yggdrasil/builds/9118#018e62ae-0e18-4c32-bb7b-99ef8051ce83/663-878
```
┌ Warning: /cache/build/yggy-amdci7-15/julialang/yggdrasil/M/MaximumIndependentSet/build/x86_64-linux-gnu/clphCAVm/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libmis.so contains std::string values!  This causes incompatibilities across the GCC 4/5 version boundary.  To remedy this, you must build a tarball for both GCC 4 and GCC 5.  To do this, immediately after your `platforms` definition in your `build_tarballs.jl` file, add the line:
│ 
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/L8tBh/src/auditor/compiler_abi.jl:247
```
```suggestion
platforms = supported_platforms(; exclude=p->libc(p)=="musl" && arch(p)=="i686")
platforms = expand_cxxstring_abis(platforms)
```

giordano (commented):
If you link to libjulia you really must follow what other packages do here, see for example: https://github.com/JuliaPackaging/Yggdrasil/blob/5dd8bee7317d0c5d0f797ae2328a069890eda5c4/L/libsingular_julia/build_tarballs.jl, since julia itself breaks the ABI in each minor release.

Highlights are
https://github.com/JuliaPackaging/Yggdrasil/blob/5dd8bee7317d0c5d0f797ae2328a069890eda5c4/L/libsingular_julia/build_tarballs.jl#L39-L44
(without the exclusion of Windows, if it isn't relevant for your case)
https://github.com/JuliaPackaging/Yggdrasil/blob/5dd8bee7317d0c5d0f797ae2328a069890eda5c4/L/libsingular_julia/build_tarballs.jl#L53
https://github.com/JuliaPackaging/Yggdrasil/blob/5dd8bee7317d0c5d0f797ae2328a069890eda5c4/L/libsingular_julia/build_tarballs.jl#L56

M/MaximumIndependentSet/build_tarballs.jl
claud10cv: I see. I believe I implemented all the changes suggested (except for the last one that clashes with on of the other suggestions above)

giordano:
I forgot you need also # See https://github.com/JuliaLang/Pkg.jl/issues/2942
# Once this Pkg issue is resolved, this must be removed

claud10cv:
> I forgot you need also
> 
> https://github.com/JuliaPackaging/Yggdrasil/blob/5dd8bee7317d0c5d0f797ae2328a069890eda5c4/L/libsingular_julia/build_tarballs.jl#L6-L9

Yep indeed without that the compiler complains about that UUID. Thank you for pointing out to the solution (or more like a workaround apparently)

M/MaximumIndependentSet/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/claud10cv/MaximumIndependentSet.git", "8a4d5f24f818800eb41ac4c7e8a3f663b96500d3")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/MaximumIndependentSet/src/
+Julia_PREFIX=${prefix}
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DJulia_PREFIX=${Julia_PREFIX} -DJlCxx_DIR=$prefix/lib/cmake/JlCxx
```
giordano, line 22: For readability:
```suggestion
cmake -B build \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DJulia_PREFIX=${Julia_PREFIX} \
    -DJlCxx_DIR=$prefix/lib/cmake/JlCxx
```

M/MaximumIndependentSet/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+
+include("../../L/libjulia/common.jl") 
+platforms = vcat(libjulia_platforms.(julia_versions)...) 
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = Product[
+	LibraryProduct(["libmis", "mislib"], :mis)
```
giordano, line 36:
```suggestion
products = [
    LibraryProduct(["libmis", "mislib"], :mis)
```

PR #8340, opened 2024-03-21
MAGEMin v1.4.0

NicolasRiel:
- Added computation of dQFM
- Corrected computation of oxides fugacity (FeO, Al2O3, H2O, SiO2, TiO2, MgO)
- Added routines to compute specific heat capacity by Gibbs_system differentiation (using MAGEMinApp)
- Improved performances when using legacy solver at suprasolidus conditions


NicolasRiel:
Thanks Mosè!

PR #8348, opened 2024-03-25
[nlcglib] New package recipe

abussy:
This is a new build recipe for the [nlcglib](https://github.com/simonpintarelli/nlcglib) library, providing a set of nonlinear CG (conjugated gradient) methods for wave-function minimization in DFT (density functional theory). We made the conscious decision not to support Windows or FreeBSD.

### DIFF
--- a/N/nlcglib/build_tarballs.jl
+++ b/N/nlcglib/build_tarballs.jl
@@ -1,0 +1,96 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+using Base.BinaryPlatforms
+const YGGDRASIL_DIR = "../.."
+include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
+
+name = "nlcglib"
+version = v"1.0.2" #technically v1.0b, but format not allowed
+
+sources = [
+   GitSource("https://github.com/simonpintarelli/nlcglib/", "de851f03b403f33cbe0bae8aac013194f8de3ff4")
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir
+
+mkdir build
+cd build
+
+CMAKE_ARGS="-DUSE_OPENMP=ON \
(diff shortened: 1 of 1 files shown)


N/nlcglib/build_tarballs.jl
```diff
@@ -0,0 +1,96 @@
...
+  CMAKE_ARGS="${CMAKE_ARGS} \
+               -DMPI_C_LIB_NAMES='mpi;pmpi;hwloc' \
+               -DMPI_CXX_LIB_NAMES='mpicxx;mpi;pmpi;hwloc' \
+               -DMPI_mpicxx_LIBRARY=${libdir}/mpich/lib.libmpicxx.a \
+               -DMPI_mpi_LIBRARY=${libdir}/mpich/lib/libmpi.a \
+               -DMPI_pmpi_LIBRARY=${libdir}/mpich/lib/libpmpi.a \
+               -DMPI_hwloc_LIBRARY=${libdir}/libhwloc.dylib"
+fi
+
+cmake .. ${CMAKE_ARGS} || cmake .. ${CMAKE_ARGS}
```
imciner2, line 49: Why are you running CMake twice?

abussy: I am not sure how it ended up there, it works fine with a single CMake call... Thanks for pointing it out, I cleaned it up.

imciner2:
I don't have much experience with MPI-based recipes here, so @giordano @eschnett either of you able to take a look at the MPI parts?

N/nlcglib/build_tarballs.jl
```diff
@@ -0,0 +1,96 @@
...
+
+products = [
+   LibraryProduct("libnlcglib", :libnlcglib)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("OpenBLAS32_jll"),
+    Dependency("Kokkos_jll"),
+    Dependency("nlohmann_json_jll"),
```
giordano, line 76: You said this is header-only, right?  See https://docs.binarybuilder.org/stable/build_tips/#Dependencies-for-the-target-system-vs-host-system
```suggestion
    BuildDependency("nlohmann_json_jll"),
```

PR #8352, opened 2024-03-25
[p7zip] Update to 17.05

fxcoudert:
Released: Feb 20, 2023


giordano:
Problem is that we don't have msan libraries in our compiler-rt build of LLVM 16, so we need to pin llvm 13

PR #8354, opened 2024-03-25
[SuiteSparse] Update to 7.6.1

fxcoudert:
Released 2nd March 2024: https://github.com/DrTimothyAldenDavis/SuiteSparse/releases/tag/v7.6.1

### DIFF
--- a/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
+++ b/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
@@ -1,7 +1,7 @@
 include("../common.jl")
 
 name = "SuiteSparse"
-version = v"7.6.0"
+version = v"7.6.1"
 
 sources = suitesparse_sources(version)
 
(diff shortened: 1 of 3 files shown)


fxcoudert:
I get:

```
Unable to find project for S/SuiteSparse/common.jl
```

and for the GPU version only:

```
ERROR: LoadError: Unsatisfiable requirements detected for package SuiteSparse_jll [bea87d4a]:
  | SuiteSparse_jll [bea87d4a] log:
  | ├─possible versions are: 5.4.0-7.6.0 or uninstalled
  | └─restricted to versions 7.6.1 by an explicit requirement — no versions left
 ```

Is it because it needs to be done in two separate PRs?

imciner2:
> Is it because it needs to be done in two separate PRs?

Yea, due to the use of the system AMD and common libraries by the SuiteSparse GPU package, we have to build the SuiteSparse CPU package first and have it be registered/available, then we can build the SuiteSparse GPU package with a dependency on that CPU package. It is a bit awkward, but it means we don't have to rebuild AMD or the common libraries in the GPU package.

PR #8359, opened 2024-03-27
Wizard recipe: svg2pdf-v0.10.0

disberd:

* Package name: svg2pdf
* Version: v0.10.0

### DIFF
--- a/S/svg2pdf/build_tarballs.jl
+++ b/S/svg2pdf/build_tarballs.jl
@@ -1,0 +1,52 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "svg2pdf"
+version = v"0.10.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/typst/svg2pdf.git", "04424512b519bc0b141775b2aee1e82c68293b5b")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/svg2pdf
+cargo build -p svg2pdf-cli --release
+install -Dvm 755 "target/${rust_target}/release/svg2pdf${exeext}" "${bindir}/svg2pdf${exeext}"
+install_license ${WORKSPACE}/srcdir/svg2pdf/LICENSE-MIT
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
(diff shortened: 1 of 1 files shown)


disberd:
I have tried this locally and it succeeded while it seems to fail here.
I adapted the script contents from https://github.com/JuliaPackaging/Yggdrasil/blob/master/T/Typst/build_tarballs.jl but have no experience with rust itself, so if anyone can help me fix this it would be appreciated

S/svg2pdf/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "freebsd"; ),
+    Platform("x86_64", "windows"; )
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("svg2pdf", :svg2pdf)
```
imciner2, line 44:
```suggestion
    ExecutableProduct("svg2pdf", :svg2pdf)
```
You have this listed as a library, but this appears to be an executable.

S/svg2pdf/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "freebsd"; ),
+    Platform("x86_64", "windows"; )
+]
```
imciner2, line 39:
```suggestion
platforms = supported_platforms()
```
This list appears to just be every platform we build on, correct? If not, there are enough here that it is better to do a filtering out instead of adding explicitly I think.

S/svg2pdf/build_tarballs.jl
disberd: Thanks a lot for this, I did not notice the issue, I fixed it in the latest commit

S/svg2pdf/build_tarballs.jl
disberd: Indeed it was only missing windows i686 as that failed on the wizard. I also fixed this using filter as suggested in the latest commit

disberd:
Thanks a lot for the very fast feedback @imciner2

disberd:
One additional comment on this.
The library itself is shipped with dual license as mentioned in the [README](https://github.com/typst/svg2pdf#license)

As user the README speicifes that either of the two licenses can be selected (MIT and Apache 2.0).
I did pick the MIT one in the build script just because I see it more often and it seems more permissive but have no real knowledge about licenses.

I am open to changing this if the other license is deemed preferable

imciner2:
I think installing only the MIT license is fine.

PR #8364, opened 2024-03-27
MAGEMin v1.4.1

NicolasRiel:
- Corrected issue with minimum number of pseudocompounds for fluid which was leading to segmentation fault when computing system specific heat capactity


NicolasRiel:
Thanks!

PR #8377, opened 2024-03-28
Migrate libcurl/libgit2/libssh2 to OpenSSL

fxcoudert:
Thinking about https://github.com/JuliaLang/julia/issues/48799 and trying to see how to migrate things in here first.


L/LibCURL/LibCURL@7/build_tarballs.jl
giordano, line 1: We aren't going to touch this old version of libcurl, I think we can exclude it from this.

L/LibSSH2/build_tarballs.jl
```diff
@@ -57,7 +53,7 @@ llvm_version = v"13.0.1"
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("MbedTLS_jll"; compat="~2.28.0"),
+    Dependency("OpenSSL_jll"; compat="3.0.12")
```
giordano, line 56:
```suggestion
    Dependency("OpenSSL_jll"; compat="3.0.12"),
```

L/LibGit2/build_tarballs.jl
```diff
@@ -63,8 +62,8 @@ llvm_version = v"13.0.1"
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("MbedTLS_jll"; compat="~2.28.0"),
     Dependency("LibSSH2_jll"; compat="1.11.0"),
+    Dependency("OpenSSL_jll"; compat="3.0.12")
```
giordano, line 66:
```suggestion
    Dependency("OpenSSL_jll"; compat="3.0.12"),
```

giordano:
For libgit2 and libssh2 we need https://github.com/JuliaPackaging/Yggdrasil/blob/fa8242bb8ab99dc836a30cb63088e0999cc6e34f/L/libssh/build_tarballs.jl#L15-L18 because openssl puts the 64-bit windows library under lib64/

giordano (changes requested):
This PR is dealing with building the packages which directly depend on mbedtls, which is great! The good news is that they're all stdlibs, so we'll just need to move update all of them together in Julia and that's it.

However there are a bunch of packages which link to mbedtls just because it's an indirect dependency through other packages, like libcurl or libgit2: https://github.com/search?q=repo%3AJuliaPackaging%2FYggdrasil%20mbedtls_jll&type=code. Before merging this PR we need to figure out what to do with them. In an ideal world we wouldn't need to link them to mbedtls/openssl at all if that's only an indirect dependency, but the build systems of those packages didn't seem to agree.

L/LibSSH2/build_tarballs.jl
fxcoudert: Oops

fxcoudert:
- Fixed typos
- Removed LibCURL@7
- Added `OPENSSL_ROOT_DIR` on `x86_64-*-mingw*`

fxcoudert:
```
[07:51:19] -- Found OpenSSL: /opt/i686-linux-gnu/i686-linux-gnu/sys-root/usr/local/lib/libcrypto.so (found version "3.0.12")
[07:51:19] -- Found mbedTLS:
[07:51:19] --   version 2.28.6
[07:51:19] --   TLS: /opt/i686-linux-gnu/i686-linux-gnu/sys-root/usr/local/lib/libmbedtls.so
[07:51:19] --   X509: /opt/i686-linux-gnu/i686-linux-gnu/sys-root/usr/local/lib/libmbedx509.so
[07:51:19] --   Crypto: /opt/i686-linux-gnu/i686-linux-gnu/sys-root/usr/local/lib/libmbedcrypto.so
```

OK we'll need to force mbedTLS off, because it's picking it up from `MbedTLS_jll v2.28.6+0`. I am wondering if this is installed from Julia itself, or from the `.ci/Manifest.toml` (which would mean I have to regenerate that file, but I don't know how).

fxcoudert:
OK I'm trying something: removing all the independent `MbedTLS_jll` dependencies.

```
    Aria2: remove MbedTLS_jll dependency
    AzStorage: remove MbedTLS_jll dependency
    IOAPI: remove MbedTLS_jll dependency
    NetCDFF: remove MbedTLS_jll dependency
    samtools: remove MbedTLS_jll dependency
    TempestModel: remove MbedTLS_jll dependency
    VMEC: remove MbedTLS_jll dependency
```

They shouldn't be necessary, and maybe the issue that made them required does not occur with OpenSSL.

fxcoudert:
Joy.

```
ERROR: LoadError: ArgumentError: The archive automatically generated by GitHub
  https://github.com/Unidata/netcdf-fortran/archive/refs/tags/v4.6.0.tar.gz
may not have a stable checksum in the future, thus cannot be used as a reliable source, see
<https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/>.
Use a different source, for example a `GitSource`, or an official release artifact uploaded
by the maintainers of the package (*not* the automatic archive produced by GitHub).
```

Updating to 4.6.1 and switching source URL.

fxcoudert:
State of the build:

- TempestModel fails on 6/9 targets. The error is: `/workspace/destdir/lib/libiomp5.so: undefined reference to 'memcpy@GLIBC_2.14'` and I don't know where it comes from. There are, however, warnings before that, related to netcdf.

```
[10:43:21] /opt/x86_64-linux-gnu/bin/../lib/gcc/x86_64-linux-gnu/4.8.5/../../../../x86_64-linux-gnu/bin/ld: warning: libmbedtls.so.12, needed by /workspace/destdir/lib/libnetcdf.so, not found (try using -rpath or -rpath-link)
[10:43:21] /opt/x86_64-linux-gnu/bin/../lib/gcc/x86_64-linux-gnu/4.8.5/../../../../x86_64-linux-gnu/bin/ld: warning: libmbedx509.so.0, needed by /workspace/destdir/lib/libnetcdf.so, not found (try using -rpath or -rpath-link)
[10:43:21] /opt/x86_64-linux-gnu/bin/../lib/gcc/x86_64-linux-gnu/4.8.5/../../../../x86_64-linux-gnu/bin/ld: warning: libmbedcrypto.so.3, needed by /workspace/destdir/lib/libnetcdf.so, not found (try using -rpath or -rpath-link)
```

Those might mean that NetCDF should be rebuilt to be based on the new CURL? If so, are there other packages depending on `LibCURL` that we should rebuild? How do we audit that?

- NetCDFF has a single failure on `x86_64-linux-gnu-libgfortran3`. It says `configure: error: Could not link to netcdf C library. Please set LDFLAGS; for static builds set LIBS to the results of nc-config --libs`… but why does it work on other targets?
- IOAPI has a single failure on `x86_64-linux-gnu-libgfortran4`, seeming to pick up mbedtls through NetCDF (as above).

```
ERROR: could not load library "/cache/build/yggy-amdci7-15/julialang/yggdrasil/I/IOAPI/build/x86_64-linux-gnu-libgfortran4/qRO2jGHy/x86_64-linux-gnu-libgfortran4-cxx11/destdir/lib/libioapi.so"
libmbedtls.so.12: cannot open shared object file: No such file or directory
```

- samtools has two failures, on `x86_64-linux-gnu` and `x86_64-linux-musl`. First one is:

```
configure: error: HTSlib development files not found
```

and second one is:

```
[10:48:07] configure: error: C compiler cannot create executables
```

Both might be fixed with a htslib rebuild with the new LibCURL.

giordano:
> Should I regenerate `.ci/Manifest.toml`? How is it done?

We can do that with the [Update Manifest](https://github.com/JuliaPackaging/Yggdrasil/actions/workflows/update_manifest.yml) GitHub action ([workflow file](https://github.com/JuliaPackaging/Yggdrasil/blob/8247c75836e45cb9004123aaf472b2bf33a4a4de/.github/workflows/update_manifest.yml)), but that's the environment for starting up BinaryBuilder, it doesn't affect what's present inside the build environment, so that's not a concern for this PR.

fxcoudert:
OK, based on my understand above, I'm adding triggers for three new builds:

```
    HDF5: new build
    htslib: new build
    NetCDF: new build
```

giordano:
Unfortunately I'm not sure just removing `MbedTLS_jll` from the deps of the extra packages is sufficient, for example I downloaded the build of [Aria2 for x86_64-linux-gnu-cxx11](https://buildkite.com/julialang/yggdrasil/builds/9227#018e89cc-e01d-478b-89bc-64554b698550):
```console
% readelf -d lib/libaria2.so

Dynamic section at offset 0x51f5000 contains 34 entries:
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/../lib64]
 0x0000000000000001 (NEEDED)             Shared library: [libxml2.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libssl.so.3]
 0x0000000000000001 (NEEDED)             Shared library: [libcrypto.so.3]
 0x0000000000000001 (NEEDED)             Shared library: [libssh2.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libmbedcrypto.so.7]
 0x0000000000000001 (NEEDED)             Shared library: [libcares.so.2]
 0x000000000000000e (SONAME)             Library soname: [libaria2.so.0]
```
It's _still_ linking to mbedtls :disappointed:

giordano:
That's probably because libssh2 forces that: https://buildkite.com/julialang/yggdrasil/builds/9225#018e89b6-012c-4518-bcb5-0207dc268aed/657-2813
```
[10:19:25] Libssh2:        yes (CFLAGS='-I/workspace/destdir/include ' LIBS='-L/workspace/destdir/lib -lssh2 -lmbedcrypto ')
```
Sigh.

fxcoudert:
Capture d’écran 2024-03-29 à 12 10 37

I think I know what's happening: it's not rebuilding LibCURL actually! Probably because I made no change to `build_tarballs.jl`? That's a guess about how the builder works, but I'm going to try to add a trigger line there.

fxcoudert:
Hum, I definitely don't understand how the builder works. In the latest run, it is rebuilding LibCURL, but it built LibGit2 before LibSSH2, even though that's a dependency. Does it actually ignore dependencies altogether? Or have I not figured out how to tell it correctly?

giordano:
Ok, I think we need to do something like https://github.com/JuliaPackaging/Yggdrasil/blob/9b72607ebca505d699a6b391a93edb91389e195e/A/Aria2/build_tarballs.jl#L18-L24 to override info from pkg-config files and force packages not to link directly to mbedtls

giordano:
> They won't link to mbedtls if the new version of libssh2 is picked.

Sure, but it'd be good to solve the issue without linking to the new builds of libcurl/libgit2/libssh2, so that these packages will be compatible with all versions of Julia, not just the newer ones (since libcurl/libgit2/libssh2 are stdlibs, and so bound to a specific version of Julia).

> I just can't figure out how to make the builder aware of the correct dependency chain. Help on that would be appreciated.

Yeah, that's not possible at the moment, we use information in the [General registry](https://github.com/JuliaRegistries/General), which is where new version are recorded _after_ a PR is merged here.

fxcoudert:
Yeah that won't work. The build artefacts for libgit2 still have:

```
/Users/fx/Downloads/LibGit2.v1.8.0.x86_64-apple-darwin/lib/libgit2.1.8.0.dylib:
	@rpath/libgit2.1.8.dylib (compatibility version 1.8.0, current version 1.8.0)
	@rpath/libssl.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	@rpath/libcrypto.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	@rpath/libssh2.1.dylib (compatibility version 1.0.0, current version 1.0.1)
	@rpath/libmbedcrypto.7.dylib (compatibility version 7.0.0, current version 2.28.6)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
```

because of the libssh2 dependency on a previous version. I suppose they'll need to be done one after another.

giordano:
This should do the trick: https://github.com/JuliaPackaging/Yggdrasil/blob/ab296450b4fa6903d4ac31a393892280629f7093/L/LibGit2/build_tarballs.jl#L26-L31

giordano (approved):
With #8386 merged, my concerns are addressed.  Thank you so much for your work!

PR #8386, opened 2024-03-29
Remove indirect MbedTLS_jll dependency from a few packages

giordano:
@fxcoudert I'm taking out of #8377 the commits which are only removing the unneeded dependency of MbedTLS from some packages, so that we can keep there only switching to OpenSSL for julia's stdlibs.  The two changesets are independent, what we have here doesn't require changing the SSL backend for the other packages.

This also reduces CI time, when I push new commits.

### DIFF
--- a/A/Aria2/build_tarballs.jl
+++ b/A/Aria2/build_tarballs.jl
@@ -15,13 +15,13 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/aria2-*/
 
-export CPPFLAGS="-I${includedir}"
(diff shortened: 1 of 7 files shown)


giordano:
This is finally all green!  I need to double check mbedtls was effectively removed everywhere, but at a quick glance this seems to be working as advertised.

giordano:
Alright, I verified that all the packages in this PR were clean from linking to mbedtls libraries, so this is good to go.  @fxcoudert thanks a lot for kickstarting this, it was extremely helpful!

PR #8392, opened 2024-03-31
[Harfbuzz] update to 8.3.1

jwahlstrand:
Updating HarfBuzz and enabling GDI and DWrite fixes font issues on Windows:

https://github.com/JuliaGtk/Gtk4.jl/issues/32
https://github.com/JuliaGtk/Gtk4.jl/issues/42
https://github.com/JuliaGraphics/Cairo.jl/issues/349

~~I haven't a clue how to fix the Mac build.~~ Bypassing a broken check for CoreText support lets the Mac build go ahead, and I have verified locally that the JLL works on my Mac.

Also, because of the huge version jump there may be compatibility issues with JLL's that depend on HarfBuzz. The soname hasn't changed, but: https://github.com/harfbuzz/harfbuzz/issues/3216. I don't actually think this is an issue, since the API's that changed were experimental and not likely used by the libraries in Yggdrasil.


giordano:
> I haven't a clue how to fix the Mac build.

For the benefit of future (and current) readers, the error is
```
[00:41:51] ../meson.build:334:6: ERROR: Problem encountered: CoreText was enabled explicitly, but required headers or frameworks are missing.
```
I guess one option is to not request CoreText explicitly?  Another option is to [read the relevant log](https://docs.binarybuilder.org/stable/troubleshooting/#General-comments) and understand why meson claims the CoreText framework isn't available.

giordano:
Is this good to go now?

jwahlstrand:
I think so -- for this to be used we'll have to rebuild Pango_jll and libass_jll since their compat is set for the old version.

giordano:
>  for this to be used we'll have to rebuild Pango_jll and libass_jll since their compat is set for the old version.

Why?

jwahlstrand:
Maybe I'm wrong, but I had to modify the compat in the JLL's to test them locally, otherwise they wouldn't install. Anyway, the answer to your original question is "yes".

giordano:
Ah, I see, we're jumping from v2.8.1 to v8.3.1.  And there's already v9.0.0

jwahlstrand:
Thanks for merging -- yes newer versions including 9 are available but I thought it might be smart to wait for 9.x because these font libraries often seem problematic.

PR #8393, opened 2024-03-31
OpenSSL: install libraries in lib not lib64

fxcoudert:
To be consistent with other libraries, libcrypto and libssl should be installed in `lib/` (not `lib64/`) on x86_64-linux. I found out about this because of https://github.com/JuliaLang/julia/pull/53891

### DIFF
--- a/O/OpenSSL/OpenSSL@3.0/build_tarballs.jl
+++ b/O/OpenSSL/OpenSSL@3.0/build_tarballs.jl
@@ -20,3 +20,5 @@ products = [
 
 # Build the tarballs.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
+
+# Build trigger: 1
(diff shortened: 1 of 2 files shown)


fxcoudert:
This would also make it possible to remove these:

```
./N/nats_c/build_tarballs.jl:    export OPENSSL_ROOT_DIR=${prefix}/lib64/
./M/mosquitto_client/build_tarballs.jl:    export OPENSSL_ROOT_DIR=${prefix}/lib64/
./L/LibSSH2/build_tarballs.jl:    export OPENSSL_ROOT_DIR=${prefix}/lib64
./L/libssh/build_tarballs.jl:    export OPENSSL_ROOT_DIR=${prefix}/lib64
./L/LibGit2/build_tarballs.jl:    export OPENSSL_ROOT_DIR=${prefix}/lib64
./L/libdatachannel/build_tarballs.jl:    export OPENSSL_ROOT_DIR="${prefix}/lib64"
```

giordano:
While we're at it, is there a way to disable building the static library?  A quick look at the `Configure` script didn't catch anything relevant there.

giordano:
Ok, no, it doesn't seem to be possible automatically: https://github.com/openssl/openssl/issues/8823.

PR #8399, opened 2024-04-01
LibGit2: avoid OpenSSL on Windows and macOS

fxcoudert:
I made a mistake in #8377: I moved Windows and macOS builds to OpenSSL, but they shouldn't. Moving back to previous situation.


giordano:
While we are at it it'd be good to look at https://github.com/JuliaLang/julia/pull/53891#issuecomment-2029491035: we aren't getting the versioned symbols anymore.

PR #8400, opened 2024-04-01
LibSSH2: move to native crypto on Windows

fxcoudert:
In Julia builds from source, the native Windows crypto backend is used instead of OpenSSL. This brings the binary builds in line (and I think it is good, one less dependency).

### DIFF
--- a/L/LibSSH2/build_tarballs.jl
+++ b/L/LibSSH2/build_tarballs.jl
@@ -64,7 +64,7 @@ llvm_version = v"13.0.1"
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("OpenSSL_jll"; compat="3.0.8"),
+    Dependency("OpenSSL_jll"; compat="3.0.8", platforms=filter(!Sys.iswindows, platforms)),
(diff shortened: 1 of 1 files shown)


L/LibSSH2/build_tarballs.jl
```diff
@@ -37,6 +36,13 @@ BUILD_FLAGS=(
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}
 )
 
+# Use native backend on Windows, OpenSSL on others
+if [[ ${target} == *-mingw* ]]; then
+    BUILD_FLAGS+=(--DCRYPTO_BACKEND=WinCNG)
+else
+    BUILD_FLAGS+=(--DCRYPTO_BACKEND=OpenSSL)
```
giordano, line 43:
```suggestion
    BUILD_FLAGS+=(-DCRYPTO_BACKEND=WinCNG)
else
    BUILD_FLAGS+=(-DCRYPTO_BACKEND=OpenSSL)
```

fxcoudert: Oops. Fixed

PR #8410, opened 2024-04-02
[bwa] Set verbosity to 2 (shows warnings/error only)

### LINKED ISSUE #8300: Reduce verbosity of BWA ?
I'm working on BWA bindings (https://github.com/jonathanBieler/BurrowsWheelerAligner.jl) and the code is printing a lot of stuff. As I understand the verbosity level is set at compile time here, so there's no way to change it from Julia's side :

https://github.com/lh3/bwa/blob/139f68fc4c3747813783a488aef2adc86626b01b/bwa.c#L42

Would it be possible to do that with a patch ? I think setting it to 2 would do the trick, level 1,2 seem to be about error handling, while 3 prints more general informations.

https://github.com/JuliaPackaging/Yggdrasil/tree/master/B/BWA

jonathanBieler:
This adds a little patch to BWA build to reduce the verbosity (the library prints a lot of stuff), should fix #8300.


B/BWA/build_tarballs.jl
```diff
@@ -17,6 +17,7 @@ script = raw"""
 cd $WORKSPACE/srcdir/bwa*/
 atomic_patch -p1 ../patches/makefile.patch
 atomic_patch -p1 ../patches/0001-Fix-building-against-GCC-10.patch
+atomic_patch -p1 ../patches/set_verbosity.patch
```
imciner2, line 20:
```suggestion

# Reduce verbosity of the output to only have warnings and errors
sed -i -e 's/int bwa_verbose = 3;/int bwa_verbose = 2/' bwa.c

```
Since all you want to do is change the one value, this find/replace should do the trick without the need for a separate patch.

PR #8426, opened 2024-04-03
[Valhalla] Rebuild with libcurl with unversioned symbols

fxcoudert:

I think it was missing many commas in the dependencies array?


fxcoudert:
clang error messages, probably from the libc++ version being higher than last time the build was attempted:

```
[19:07:57] /workspace/srcdir/valhalla/third_party/robin-hood-hashing/src/include/robin_hood.h:1445:33: error: builtin __has_trivial_copy is deprecated; use __is_trivially_copyable instead [-Werror,-Wdeprecated-builtins]
[19:07:57]         Cloner<Table, IsFlat && ROBIN_HOOD_IS_TRIVIALLY_COPYABLE(Node)>()(o, *this);
[19:07:57]                                 ^
```

This is on darwin and freebsd. Might be fixed in latest version.

giordano:
Since it's a `-Werror` failure (sigh), presumably passing [`-DENABLE_SINGLE_FILES_WERROR=OFF`](https://github.com/valhalla/valhalla/blob/4c4dd99eeab9572a54fa5fb4dcf158107e9bb8f5/CMakeLists.txt#L35C8-L35C34) to `cmake` would do it?

jeremiahpslewis:
Want to try the latest release? v3.4.0, hash: cbabe7cfbad2225090e4c006c778f1f0a7b3ec4e

PR #8433, opened 2024-04-03
[libgeotiff] Rebuild with libcurl with unversioned symbols

fxcoudert:

We can rebuild this now that PROJ was done in https://github.com/JuliaPackaging/Yggdrasil/pull/8424


imciner2:
If you have dependencies in the chain, you'll need to wait for the new versions to propagate through the package registration process as well probably: https://github.com/JuliaRegistries/General/pulls.

fxcoudert:
OK, will rerun after https://github.com/JuliaRegistries/General/pull/104196

fxcoudert:
This is good to go!

PR #8434, opened 2024-04-03
[NetCDF] Rebuild with libcurl with unversioned symbols

fxcoudert:

This needed https://github.com/JuliaPackaging/Yggdrasil/pull/8423 to go in first, now we can rebuilb

### DIFF
--- a/N/NetCDF/build_tarballs.jl
+++ b/N/NetCDF/build_tarballs.jl
@@ -11,7 +11,7 @@ name = "NetCDF"
 upstream_version = v"4.9.2"
 
 # Offset to add to the version number.  Remember to always bump this.
-version_offset = v"0.2.9"
+version_offset = v"0.2.10"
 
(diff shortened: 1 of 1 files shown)


fxcoudert:
Will need to wait for https://github.com/JuliaRegistries/General/pull/104197

giordano:
Ah, that's because `compat` in
```julia
    Dependency("HDF5_jll"; compat = "~1.14"),
```
has a two-fold effect: specifying the compatibility bounds, _and_ indicating the version of the package to use for the build, we automatically infer we should use the lowest version compatible with the compat specification (see https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.Dependency).  So we're building with HDF5_jll@v1.14.0, see https://buildkite.com/julialang/yggdrasil/builds/9418#018ea5a2-d783-4037-8f74-4c2c02fbda90/664-3010.

PR #8437, opened 2024-04-04
[TracyProfiler] Update to v0.10

IanButterworth:
v0.10 released Oct 16, 2023 https://github.com/wolfpld/tracy/releases/tag/v0.10

@topolarity @KristofferC this look good to you?


T/TracyProfiler/build_tarballs.jl
```diff
@@ -3,12 +3,12 @@
...
 
 name = "TracyProfiler"
-version = v"0.9.1"
+version = v"0.10.0"
 
 # Collection of sources required to complete build
 sources = [
     GitSource("https://github.com/wolfpld/tracy.git",
-              "897aec5b062664d2485f4f9a213715d2e527e0ca"), # v0.9.1
+              "37aff70dfa50cf6307b3fee6074d627dc2929143"), # v0.19
```
IanButterworth, line 11:
```suggestion
              "37aff70dfa50cf6307b3fee6074d627dc2929143"), # v0.10
```

IanButterworth:
Thanks to @giordano for [publishing](https://github.com/giordano/TracyProfiler_jll.jl) this pre-release. I tested it and tracy loads, but complains that julia needs to be built against a compatible lib version


Screenshot 2024-04-04 at 11 36 13 AM

topolarity:
Yeah, unfortunately Tracy is not backwards-compatible so this will need a coordinated bump of LibTracyClient_jll

giordano:
Oh, I see TracyProfiler and LibTracyClient technically don't depend on each other, so we can't set compat bounds to avoid these issues.  That's annoying.

topolarity:
Yeah we don't want to make TracyProfiler depend on LibTracyClient since that would cause the client to auto-load.

We could potentially use a dummy package TracyProtocol that they both depend on?

giordano:
> We could potentially use a dummy package TracyProtocol that they both depend on?

How about the opposite: a simple Julia package which loads both of them, and can set compat bounds.  End users would use that one directly, instead of the jlls.

topolarity:
That would continue to have the problem of loading the client in a process where you only want to run the GUI application though, which is very confusing for users because it means an extra `julia` appears in menu to connect to and it generally sends no events.

IanButterworth:
I might be missing something, but you can depend on a package and set compat for it, but not load it. Wouldn't that work?

giordano:
In JLLs we don't have "depend on but not load" dependencies, not at the moment.

IanButterworth:
I've been trying to use the current TracyProfiler_jll on MacOS and it is basically unusable because of the hypersensitive zoom and lack of scrolling via mouse and no scrollbars.

It'd be good to get this in and julia updated.

topolarity:
Are those issues fixes in 0.10?

IanButterworth:
I've tried to figure that out from Tracy and CImGUI PRs but haven't managed.

Do you know if I can load this jll with some dummy profile data?

IanButterworth:
I posted here https://github.com/wolfpld/tracy/issues/631#issuecomment-2227420490

topolarity:
You can use the sample file used for Tracy's web demo: https://tracy.nereid.pl/Tracy-release.data

Despite the file extension, it's really a `.tracy` file

IanButterworth:
Ah good tip.

No luck, unfortunately.

Screenshot 2024-07-14 at 7 33 08 PM


I'll post back on the issue.

KristofferC:
https://github.com/wolfpld/tracy/issues/631#issuecomment-2476979922 suggests this have been fixed. There is also a 0.11 now.

IanButterworth:
Feel free to take this over

giordano:
#11089

PR #8464, opened 2024-04-08
MAGEMin v1.4.3

NicolasRiel:
- largely improved general performances (~40% faster)
- for mac, matrix inversion now uses dgetrs from the accelerate framework
- fixes cases of ill-conditionned matrix due to duplicated pseudocompounds
- fixes minimization status output


NicolasRiel:
Thank you!

PR #8467, opened 2024-04-09
Update mpi.jl to respect platform argument

joshday:
This PR makes the MPI platform augmentation respect the `platform` argument.

The change fixes a rather niche bug, which is that e.g. `Pkg.instantiate(platform=Windows(:x86_64))` from a non-windows machine doesn't install any artifacts that use this augmentation (I don't know if this claim is fully true, but it is true for HDF5).  I discovered this through https://github.com/JuliaComputing/DepotDelivery.jl, a package for bundling a depot as a software deliverable.

Looks like this would [affect quite a few JLLs](https://github.com/search?q=org%3AJuliaBinaryWrappers+augment_mpi%21&type=code).

Is it possible to trigger all these builds without the dummy e.g. `# Trigger build` changes to the `build_tarballs.jl` files?

### DIFF
--- a/platforms/mpi.jl
+++ b/platforms/mpi.jl
@@ -11,7 +11,7 @@ const augment = raw"""
     function augment_mpi!(platform)
         # Doesn't need to be `const` since we depend on MPIPreferences so we
         # invalidate the cache when it changes.
-        binary = get(preferences, "binary", Sys.iswindows() ? "MicrosoftMPI_jll" : "MPICH_jll")
+        binary = get(preferences, "binary", Sys.iswindows(platform) ? "MicrosoftMPI_jll" : "MPICH_jll")
 
         abi = if binary == "system"
             let abi = get(preferences, "abi", nothing)
--- a/platforms/mpi.jl
+++ b/platforms/mpi.jl
@@ -11,6 +11,7 @@ const augment = raw"""
     function augment_mpi!(platform)
         # Doesn't need to be `const` since we depend on MPIPreferences so we
         # invalidate the cache when it changes.
+        # Note: MPIPreferences uses `Sys.iswindows()` without the `platform` argument.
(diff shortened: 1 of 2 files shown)


platforms/mpi.jl
```diff
@@ -11,7 +11,7 @@ const augment = raw"""
     function augment_mpi!(platform)
         # Doesn't need to be `const` since we depend on MPIPreferences so we
         # invalidate the cache when it changes.
-        binary = get(preferences, "binary", Sys.iswindows() ? "MicrosoftMPI_jll" : "MPICH_jll")
+        binary = get(preferences, "binary", Sys.iswindows(platform) ? "MicrosoftMPI_jll" : "MPICH_jll")
```
vchuravy, line 14: Can you add a brief comment. E.g. the logic will no longer be in-sync with MPIPreferences

https://github.com/JuliaParallel/MPI.jl/blob/4ae3df31c2aa521dab8603f974363e965646c795/lib/MPIPreferences/src/MPIPreferences.jl#L26

vchuravy (approved):

vchuravy:
> Is it possible to trigger all these builds without the dummy e.g. # Trigger build changes to the build_tarballs.jl files?

No not really. Maybe @giordano has an idea

giordano:
> Maybe @giordano has an idea

No :smiling_face_with_tear:.  But maybe we can start with rebuilding the package(s) needed by @joshday right now (only hdf5?), others will be automatically fixed when they'll be rebuilt next for whatever other reason.  This is a corner case of the build system, I wouldn't expect many users to target a different platform than their host.

joshday:
I've added a comment about the difference in MPIPreferences and bumped HDF5/build_tarballs.jl as well.

Waiting on the other JLLs to be rebuilt (as @giordano suggested) sounds like a good plan to me.

PR #8473, opened 2024-04-10
[SCS] enable openmp

kalmarek:
Since SCS_GPU_jll and SCS_MKL_jll are available only on x86_64 platforms the LLVMOpenMP dep is disabled for them.

MKL can be also updated to v2024, but let's do it in another pull request.

@odow

### DIFF
--- a/S/SCS/build_tarballs.jl
+++ b/S/SCS/build_tarballs.jl
@@ -1,3 +1,4 @@
+using Pkg
 using BinaryBuilder
 
 name = "SCS"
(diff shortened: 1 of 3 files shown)


odow:
LGTM, but I'm not an expert here.

kalmarek:
does this require a version bump, or are we fine with `+1`?

odow:
I think +1 is okay.

kalmarek:
@giordano I believe this is ready to be merged, unless you have some comments?

PR #8479, opened 2024-04-13
[libspatialite] Update to v5.1.0

asinghvi17:
Also removes the GEOS compat constraint here, since that was restricting LibGEOS installation.

### DIFF
--- a/L/libspatialite/build_tarballs.jl
+++ b/L/libspatialite/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "libspatialite"
-version = v"5.0.1"
+version = v"5.1.0"
 
 # Collection of sources required to complete build
 sources = [
@@ -88,7 +88,7 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("SQLite_jll")
-    Dependency("GEOS_jll"; compat="~3.9")
+    Dependency("GEOS_jll")
     Dependency(PackageSpec(name="PROJ_jll", uuid="58948b4f-47e0-5654-a9ad-f609743f8632"))
     Dependency(PackageSpec(name="Libiconv_jll", uuid="94ce4f54-9a6c-5748-9c1c-f9c7231a4531"))
     Dependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a"))


L/libspatialite/build_tarballs.jl
```diff
@@ -88,7 +88,7 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("SQLite_jll")
-    Dependency("GEOS_jll"; compat="~3.9")
+    Dependency("GEOS_jll")
```
giordano, line 91: Compat bounds are there to avoid real errors (runtime incompatibilities), please restore it

giordano (changes requested):

L/libspatialite/build_tarballs.jl
asinghvi17: Is there an example of this that I can test?  GEOS_jll is on v3.12 now, and this compat requirement restricts that.  Alternatively if this is just meant to avoid versions below 3.9 I could change it instead.

giordano: And BTW, you can also see the code at https://github.com/JuliaBinaryWrappers/GEOS_jll.jl/blob/cb824a85a217e13cb352f24d7d83e57ee781dc3e/src/wrappers/x86_64-linux-gnu-cxx11.jl#L6. For some reason they thought it was a sensible idea to hardcode the full version number in the soname, so if you move to v3.12.0 your compat bound must be "=3.12.0", no other options possible.

asinghvi17: Thanks for the tip!  What a mess.  I can definitely change compat to 3.12 then...is there a way to release multiple versions, each compatible with a different version of GEOS_jll?  Or is it just easier to keep syncing them up>

asinghvi17:
```suggestion
    Dependency("GEOS_jll"; compat="~3.12")
```
Fix GEOS_

L/libspatialite/build_tarballs.jl
```diff
@@ -88,7 +88,7 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("SQLite_jll")
-    Dependency("GEOS_jll"; compat="~3.9")
+    Dependency("GEOS_jll"; compat="~3.12")
```
giordano, line 91: https://github.com/JuliaPackaging/Yggdrasil/pull/8479#discussion_r1564117908

fingolfin:
@asinghvi17 are you still interested in this? if so, there is a merge conflict and an unaddressed comment by Mosè

L/libspatialite/build_tarballs.jl
asinghvi17:
```suggestion
    Dependency("GEOS_jll"; compat="=3.12")
```

PR #8483, opened 2024-04-15
Added Faiss

stemann:


F/Faiss/bundled/patches/mingw32.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/faiss/impl/platform_macros.h b/faiss/impl/platform_macros.h
+index 2aecc512..9c2e416c 100644
+--- a/faiss/impl/platform_macros.h
++++ b/faiss/impl/platform_macros.h
+@@ -11,7 +11,7 @@
+ #include <cstdint>
+ #include <cstdio>
+ 
+-#ifdef _MSC_VER
++#ifdef _MSC_VER || defined(__MINGW32__)
```
giordano, line 10: This should simply check `_WIN32`: https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

F/Faiss/bundled/patches/mingw32.patch
giordano, line 1: Please upstream whatever patches you have.

stemann:
How do we want this?

v1.7.4, without windows, musl and freebsd, builds just fine: https://github.com/JuliaPackaging/Yggdrasil/pull/8483/commits/3fe4eb836f57b032ff79cd176c8d6e3a4e7ad797

Complications:
* v1.8.0: Fixes musl and freebsd, but breaks aarch64-linux due to ARM NEON optimization (which uses vld1q_f32_x2 etc.) included in v1.8.0: https://github.com/facebookresearch/faiss/pull/3166
* Windows: Linking fails to find some symbols

To-do (future?)
* CUDA support

giordano:
> v1.8.0: Fixes musl and freebsd, but breaks aarch64-linux due to ARM NEON optimization (which uses vld1q_f32_x2 etc.) included in v1.8.0: https://github.com/facebookresearch/faiss/pull/3166

Open an issue?

stemann:
> > v1.8.0: Fixes musl and freebsd, but breaks aarch64-linux due to ARM NEON optimization (which uses vld1q_f32_x2 etc.) included in v1.8.0: [facebookresearch/faiss#3166](https://github.com/facebookresearch/faiss/pull/3166)
> 
> Open an issue? _**Edit**_: ah, I guess that may require the neon extension.

I might just have gotten too lazy at first - too lazy to create a patch for the aarch64/neon bits...; it was easier rolling back to v1.7.4 - and skipping musl and freebsd at first - better than not having aarch64-linux support.

stemann:
> How do we want this?
> 
> ...
> 
> To-do (future?)
> 
> * CUDA support

As there is already CUDA-support in Faiss.jl (using the Faiss Python API via PythonCall), CUDA-support might be worth giving a go in the first iteration of a Faiss JLL (this PR): https://github.com/zsz00/Faiss.jl/issues/33#issuecomment-2061530093

Cf. https://github.com/zsz00/Faiss.jl/blob/v0.3.5/src/Faiss.jl#L43

PR #8486, opened 2024-04-16
[alsa_plugins] Bump to v1.2.7, build for experimental platforms

gnadt:
Uses the plugin directory option [here](https://github.com/alsa-project/alsa-plugins/blob/master/configure.ac#L189-L192)

### DIFF
--- a/A/alsa_plugins/build_tarballs.jl
+++ b/A/alsa_plugins/build_tarballs.jl
@@ -21,7 +21,7 @@ make install
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = filter(Sys.islinux, supported_platforms(;experimental=true))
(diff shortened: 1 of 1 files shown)


A/alsa_plugins/build_tarballs.jl
```diff
@@ -3,32 +3,32 @@
...
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
 make
 make install
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = filter!(Sys.islinux, supported_platforms())
-
+platforms = filter(Sys.islinux, supported_platforms(;experimental=true))
```
imciner2, line 25: `experimental` is empty right now, so this doesn't actually change anything.

giordano:
This PR is mixing changes from #8485?

PR #8489, opened 2024-04-16
[alsa_plugins] Bump to v1.2.7

gnadt:


A/alsa_plugins/build_tarballs.jl
```diff
@@ -3,32 +3,31 @@
...
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
 make
 make install
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = filter!(Sys.islinux, supported_platforms())
-
+platforms = filter(p -> Sys.islinux(p) & (arch(p) != "armv6l"), supported_platforms(;experimental=true))
```
giordano, line 24:
```suggestion
platforms = filter(p -> Sys.islinux(p) & (arch(p) != "armv6l"), supported_platforms())
```

A/alsa_plugins/build_tarballs.jl
```diff
@@ -3,32 +3,31 @@
...
 
 # The products that we will ensure are always built
 products = [
-    LibraryProduct("libasound_module_pcm_a52", :libasound_module_pcm_a52, "lib/alsa-lib"),
+    LibraryProduct("libasound_module_pcm_a52", :libasound_module_pcm_a52, "lib/alsa-lib"; dont_dlopen=true),
```
giordano, line 28: Why don't dlopen?

A/alsa_plugins/build_tarballs.jl
```diff
@@ -45,10 +44,10 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency(PackageSpec(name="FFMPEG_jll", uuid="b22a6f82-2f65-5046-a5b2-351ab43fb4e5"))
```
giordano, line 46:
```suggestion
    Dependency(PackageSpec(name="FFMPEG_jll", uuid="b22a6f82-2f65-5046-a5b2-351ab43fb4e5"); compat="6.1.1")
```

A/alsa_plugins/build_tarballs.jl
giordano: Doesn't sound very useful if we can't dlopen it?  Why it used to work?  What's the problem exactly?

PR #8490, opened 2024-04-16
Recompile qr_mumps with recent SuiteSparse_jll.jl

amontoison:


Q/qr_mumps/build_tarballs.jl
```diff
@@ -58,12 +58,12 @@ products = [
...
+    Dependency(PackageSpec(name="SuiteSparse_jll", uuid="bea87d4a-7f5b-5778-9afe-8cc45184846c")),
     Dependency(PackageSpec(name="METIS_jll", uuid="d00139f3-1899-568f-a2f0-47f597d42d70")),
     Dependency(PackageSpec(name="SCOTCH_jll", uuid="a8d0f55d-b80e-548d-aff6-1a04c175f0f9"); compat="6.1.3")
     # Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93")) for Julia ≥ v"1.9"
     # Dependency(PackageSpec(name="StarPU_jll", uuid="e3ad0b27-b140-5312-a56e-059adfc55eb4"))
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"9.1.0", clang_use_lld=false)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.10", preferred_gcc_version=v"9.1.0", clang_use_lld=false)
```
giordano, line 69: You can't change compat bounds without changing version number of a package, this PR should be reverted

amontoison: Because with SuiteSparse 7.7.0 they changed the soname of `libamd.so`, it's not anymore `libamd.so.2` but `libamd.so.3` now.
Because this version of SuiteSparse is only available on Julia 1.10 and we can't dynamically changed SuiteSparse_jll.Jl, I bumped the version number of `qr_mumps_jll.jl`.

amontoison: https://github.com/JuliaSmoothOptimizers/QRMumps.jl/issues/86

giordano: But you didn't specify the version of SuiteSparse, v7.7 is not what's used in Julia v1.10

amontoison: Because I wasn't certain which version of SuiteSparse introduced the new SONAME, and also because not all SONAMEs of SuiteSparse libraries are updated when Tim Davis increments the major version number.

However, recompiling qr_mumps with SuiteSparse 7.7.0 resolved the issue.
So the SONAME probably changed with SuiteSparse 7.x.x.
I added a compatibility entry in Julia to install it only on Julia versions >= 1.10.

Issue #8465, opened 2024-04-08
CeresSolver_jll leads to `libcholmod.so.3` not found -- julia 1.10+


avik-pal:
```julia
julia> using CeresSolver_jll
ERROR: InitError: could not load library "/home/avikpal/.julia/artifacts/d137110a8d2540ce7c5d544b481ce9aff7a5594a/lib/libceres.so"
libcholmod.so.3: cannot open shared object file: No such file or directory
Stacktrace:
  [1] dlopen(s::String, flags::UInt32; throw_error::Bool)
    @ Base.Libc.Libdl ./libdl.jl:117
  [2] dlopen(s::String, flags::UInt32)
    @ Base.Libc.Libdl ./libdl.jl:116
  [3] macro expansion
    @ ~/.julia/packages/JLLWrappers/pG9bm/src/products/library_generators.jl:63 [inlined]
  [4] __init__()
    @ CeresSolver_jll ~/.julia/packages/CeresSolver_jll/nq3AI/src/wrappers/x86_64-linux-gnu-cxx11.jl:12
  [5] run_module_init(mod::Module, i::Int64)
    @ Base ./loading.jl:1134
  [6] register_restored_modules(sv::Core.SimpleVector, pkg::Base.PkgId, path::String)
    @ Base ./loading.jl:1122
  [7] _include_from_serialized(pkg::Base.PkgId, path::String, ocachepath::String, depmods::Vector{Any})
    @ Base ./loading.jl:1067
  [8] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String, build_id::UInt128)
    @ Base ./loading.jl:1581
  [9] _require(pkg::Base.PkgId, env::String)
    @ Base ./loading.jl:1938
 [10] __require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:1812
 [11] #invoke_in_world#3
    @ ./essentials.jl:926 [inlined]
 [12] invoke_in_world
    @ ./essentials.jl:923 [inlined]
 [13] _require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:1803
 [14] macro expansion
    @ ./loading.jl:1790 [inlined]
 [15] macro expansion
    @ ./lock.jl:267 [inlined]
 [16] __require(into::Module, mod::Symbol)
    @ Base ./loading.jl:1753
 [17] #invoke_in_world#3
    @ ./essentials.jl:926 [inlined]
[...]

PR #8414, opened 2024-04-02
[libblastrampoline] Make using libblastrampoline easier from CMake

imciner2:
This PR adds a new CMake file that setups the environment for libblastrampoline to be used with the CMake FindBLAS and FindLAPACK by setting up the various CMake variables those find modules provide to point to the appropriate library in the buildroot.

This replaces the need for setting all the `-DBLAS*` and `-DLAPACK*` variables on the CMake command line each time. Instead, each CMake invocation that will use libblastrampoline should add
```
-DCMAKE_PROJECT_TOP_LEVEL_INCLUDES="${libdir}/cmake/blastrampoline/yggdrasilenv.cmake" \
```
to the command line, and ensure that they declare a dependency on at least version 3.24 of the CMake JLL.

This does expose two extra variables that aren't actually defined by the find scripts normally, `BLAS_INCLUDE_DIRS` and `LAPACK_INCLUDE_DIRS`, which point to the target-specific include directory provided by libblastrampoline. While not defined by the docs for the targets, I felt it was useful to include because libblastrampoline does put all the include files into a bit of a weird place (and some CMake scripts do refer to these variables, such as the SuiteSparse CMake, even if they aren't required).

There are two configuration points for this:
* `BLASTRAMPOLINE_LIB` - set to the desired library name (without the `lib` prefix). This is automatically set appropriately based on the platform if left blank.
* `BLASTRAMPOLINE_INTEGER` - set to either `ILP64` or `LP64` to force a specific interface spec. This is automatically set based on the `nbits` of the target, so a 64 bit target gets `ILP64` and a 32-bit target gets `LP64`. This only affects the directory used in the include path though, since the actual library is the same for both.

I am tempted to add sets of variables for `LAPACKE` and `CBLAS` as well, but there aren't actually upstream find modules for those, but they seem like they might be useful anyway.

I have tested this by modifying the SuiteSparse recipe locally, and it finds the BLAS/LAPACK libraries properly still during the build phase:

64-bit Linux:

```
[16:44:29] -- LAPACK libraries:    /workspace/destdir/lib/libblastrampoline.so
[16:44:29] -- LAPACK include:      /workspace/destdir/include/libblastrampoline/ILP64/x86_64-linux-gnu
[16:44:29] -- LAPACK linker flags: blastrampoline
[16:44:29] -- BLAS libraries:      /workspace/destdir/lib/libblastrampoline.so
[...]

PR #8491, opened 2024-04-16
[CeresSolver] Update to 2.2 (and bump SuiteSparse for Julia 1.10)

avik-pal:


C/CeresSolver/build_tarballs.jl
```diff
@@ -37,17 +37,12 @@ products = Product[
...
-    # Eigen_jll v0.3.4 throws errors on powerpc64le with older GCC versions
     BuildDependency("Eigen_jll"),
     Dependency("glog_jll"),
-    # Metis replaces SuiteSparse on Windows
     Dependency("METIS_jll"),
-    Dependency("OpenBLAS32_jll"),
-    # Hard code the version now as the latest v7.0.1 does not get recognized
-    Dependency("SuiteSparse_jll", v"5.10.1")
+    Dependency("libblastrampoline_jll"),
+    Dependency("SuiteSparse_jll", v"7.2.1")
```
giordano, line 44: Please use compat bounds, not just build version numbers, they're almost useless
```suggestion
    Dependency("libblastrampoline_jll"; compat="5.4.0"),
    Dependency("SuiteSparse_jll"; compat="7.2.1"),
```

ViralBShah: Ok - now I understand what you mean.

C/CeresSolver/build_tarballs.jl
```diff
@@ -37,17 +37,12 @@ products = Product[
...
-    # https://github.com/ceres-solver/ceres-solver/issues/886
-    # Eigen_jll v0.3.4 throws errors on powerpc64le with older GCC versions
     BuildDependency("Eigen_jll"),
     Dependency("glog_jll"),
-    # Metis replaces SuiteSparse on Windows
     Dependency("METIS_jll"),
-    Dependency("OpenBLAS32_jll"),
-    # Hard code the version now as the latest v7.0.1 does not get recognized
-    Dependency("SuiteSparse_jll", v"5.10.1")
+    Dependency("libblastrampoline_jll"; compat="5.4.0"),
```
giordano, line 43: Ok, SuiteSparse_jll@7.2.1 requires libblastrampoline_jll@5.8.0: https://github.com/JuliaRegistries/General/blob/d92ae979ea8b36c45993586e6fbca2db188f500a/jll/S/SuiteSparse_jll/Compat.toml#L24
```suggestion
    Dependency("libblastrampoline_jll"; compat="5.8.0"),
```

C/CeresSolver/build_tarballs.jl
```diff
@@ -16,8 +16,8 @@ CMAKE_FLAGS+=(-DCMAKE_INSTALL_PREFIX=${prefix}
               -DBUILD_SHARED_LIBS=ON
               -DBUILD_EXAMPLES=OFF
               -DBUILD_TESTING=OFF
-              -DBLAS_LIBRARIES=${libdir}/libopenblas.${dlext}
-              -DLAPACK_LIBRARIES=${libdir}/libopenblas.${dlext}
+              -DBLAS_LIBRARIES=${libdir}/libblastrampoline.${dlext}
+              -DLAPACK_LIBRARIES=${libdir}/libblastrampoline.${dlext}
```
giordano, line 20: With #8414 this should be simpler, but I'm not sure about what we have to do exactly.  @imciner2?

imciner2: To use that, we need a dependency on libblastrampoline that is new enough, a HostBuildDependency on CMake (either unversioned or at least 3.24), do `apk del cmake` in the script, add `-DCMAKE_PROJECT_TOP_LEVEL_INCLUDES="${libdir}/cmake/blastrampoline/yggdrasilenv.cmake"` to the CMake options, and remove these existing options.

C/CeresSolver/build_tarballs.jl
```diff
@@ -1,11 +1,11 @@
 using BinaryBuilder
```
imciner2, line 1:
```suggestion
using BinaryBuilder, Pkg
```

imciner2:
Hmm, it seems CMake here isn't happy for some reason:

> [00:07:55] -- Did not find LAPACK library, disabling LAPACK support.
[00:07:55] -- Failed to find SuiteSparse - Did not find BLAS library (required for SuiteSparse).
[00:07:55] -- Failed to find SuiteSparse - Did not find LAPACK library (required for SuiteSparse).

I'll need to dig into that, because I don't see any obvious errors being shown with the new system.

C/CeresSolver/build_tarballs.jl
```diff
@@ -16,11 +16,12 @@ CMAKE_FLAGS+=(-DCMAKE_INSTALL_PREFIX=${prefix}
...
-              -DLAPACK_LIBRARIES=${libdir}/libopenblas.${dlext}
+              -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES="${libdir}/cmake/blastrampoline/yggdrasilenv.cmake"
               -DMETIS_LIBRARY=${libdir}/libmetis.${dlext}
               )
 
+apk del cmake
+
 cd $WORKSPACE/srcdir/ceres-solver/
 mkdir build && cd build
 cmake .. ${CMAKE_FLAGS[@]}
```
imciner2, line 27:
```suggestion
cmake -S .. -B . "${CMAKE_FLAGS[@]}"
```
Lets try this since it has a quoted string in it.

C/CeresSolver/build_tarballs.jl
```diff
@@ -16,14 +16,15 @@ CMAKE_FLAGS+=(-DCMAKE_INSTALL_PREFIX=${prefix}
               -DBUILD_SHARED_LIBS=ON
               -DBUILD_EXAMPLES=OFF
               -DBUILD_TESTING=OFF
-              -DBLAS_LIBRARIES=${libdir}/libopenblas.${dlext}
-              -DLAPACK_LIBRARIES=${libdir}/libopenblas.${dlext}
+              -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES="${libdir}/cmake/blastrampoline/yggdrasilenv.cmake"
```
imciner2, line 19:
```suggestion
              -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=${libdir}/cmake/blastrampoline/yggdrasilenv.cmake
```

ViralBShah (approved):

imciner2:
Not ready yet, CMake is still unhappy:

> [01:00:10]  ---> cmake -S .. -B . "${CMAKE_FLAGS[@]}"
[01:00:10] Re-run cmake no build system arguments
[01:00:10] -- The C compiler identification is GNU 8.1.0
[01:00:10] -- The CXX compiler identification is GNU 8.1.0

imciner2:
I believe we are hitting a CMake bug with the FindBLAS and FindLAPACK not working properly with our overrides here. I filed an upstream issue here https://gitlab.kitware.com/cmake/cmake/-/issues/25913 for that. I don't think I saw this in my original testing because I used SuiteSparse, which has its own FindBLAS and FindLAPACK logic that didn't do this.

Now, I guess the question is, do we need to switch over in this PR? I know we link SuiteSparse against blastrampoline, so I think we probably will need to, otherwise we risk having library linking problems. The other option is we don't link against BLAS/LAPACK right now and let it fallback to its internal routines, which might be the safest option while I work with upstream CMake to understand this and get it fixed.

ViralBShah:
Would using something like `-L${libdir} -lblastrampoline` not work? We can also just link it to OpenBLAS32 like before.

giordano:
We can go back to https://github.com/JuliaPackaging/Yggdrasil/blob/be4f220c9d063ab76436261638ce74713468e023/C/CeresSolver/build_tarballs.jl#L19-L20 (but with the right names, that's wrong for Windows).  In the long term it'd be nice to get proper cmake support for libblastrampoline though, so simplify this.

ViralBShah:
That looks like it worked. Should be good to merge.

C/CeresSolver/build_tarballs.jl
```diff
@@ -37,17 +45,13 @@ products = Product[
...
-    # Eigen_jll v0.3.4 throws errors on powerpc64le with older GCC versions
     BuildDependency("Eigen_jll"),
     Dependency("glog_jll"),
-    # Metis replaces SuiteSparse on Windows
     Dependency("METIS_jll"),
-    Dependency("OpenBLAS32_jll"),
-    # Hard code the version now as the latest v7.0.1 does not get recognized
-    Dependency("SuiteSparse_jll", v"5.10.1")
+    Dependency("libblastrampoline_jll"; compat="5.8.0"),
+    Dependency("SuiteSparse_jll"; compat="7.2.1"),
```
giordano, line 52: I think that based on https://github.com/JuliaPackaging/Yggdrasil/issues/8465#issuecomment-2044682817 this should be
```suggestion
    Dependency("SuiteSparse_jll"; compat="~7.2.1"),
```
?

PR #8492, opened 2024-04-16
[gmsh]: Update to 4.13.0 and build with updated HDF5 version

fredrikekre:
This is an attempt to make `gmsh_jll` work on Julia 1.12. The only HDF5 version that works on Julia 1.12 is 1.14, but `gmsh_jll` is locked to 1.12.

IIRC this PR would only result in a new build number for `gmsh_jll` so it might interfere with the existing compat bounds in the registry, right?

### DIFF
--- a/G/gmsh/build_tarballs.jl
+++ b/G/gmsh/build_tarballs.jl
@@ -69,7 +69,7 @@ dependencies = [
     Dependency("GLU_jll"; platforms=x11_platforms),
     Dependency("GMP_jll"; compat="6.2"),
     # Updating to a newer HDF5 version requires rebuilding this package
-    Dependency("HDF5_jll"; platforms=hdf5_platforms, compat="~1.12.1"),
+    Dependency("HDF5_jll"; platforms=hdf5_platforms, compat="~1.14.3"),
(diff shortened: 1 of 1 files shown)


giordano:
Yes, you need to change the version number to change existing compat bounds.

fredrikekre:
Whats the recommended version to use here then?

giordano:
Can you update to v4.13.0?

fredrikekre:
Ah, it exist [here](https://gmsh.info/src/) but not [here](https://gitlab.onelab.info/gmsh/gmsh/-/releases).

fredrikekre:
Do you typically update the Julia compat here too or just implicitly through the HDF5 compat?

giordano:
The compat of HDF5 https://github.com/JuliaPackaging/Yggdrasil/blob/55bff4fe203da46b6436d72f6df417800d6cb66f/H/HDF5/build_tarballs.jl#L305 is the same as this package, no?

fredrikekre:
Right, I misremembered what the origianl problem was -- previous HDF5 versions are not allowed on 1.12.

PR #7454, opened 2023-09-29
[Cairo] v1.18.0

jeremiahpslewis:

PR #8497, opened 2024-04-18
[Pango] Update to v1.52.2

giordano:
Also, update dependency on Cairo_jll.  #7454 changed the build system to the newer meson, but this changed the compatibility version in the macOS binaries, which means that libraries built against Cairo 1.16 require a compatibility version way larger than what's recorded in our builds of Cairo 1.18: https://github.com/JuliaGraphics/Cairo.jl/issues/362.  This is non-sense, but this is a reminder that changing build systems has a cost: most of the time they don't produce 100% compatible binaries.

For the record, the problem with the macOS compatibility arises only on macOS < 12 and `-mmacosx-version-min` is set to a value < 10.14 (we use 10.12 for x86_64-darwin builds, and 11.0 for aarch64-darwin builds): https://github.com/giordano/macos-compatibility-version

### DIFF
--- a/P/Pango/build_tarballs.jl
+++ b/P/Pango/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "Pango"
-version = v"1.52.1"
+version = v"1.52.2"
 
-# Collection of sources required to build Pango
+# Collection of sources required to build Pango: https://download.gnome.org/sources/pango/
 sources = [
     ArchiveSource("http://ftp.gnome.org/pub/GNOME/sources/pango/$(version.major).$(version.minor)/pango-$(version).tar.xz",
-                  "58728a0a2d86f60761208df9493033d18ecb2497abac80ee1a274ad0c6e55f0f"),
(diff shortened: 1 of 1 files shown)


P/Pango/build_tarballs.jl
```diff
@@ -68,9 +68,9 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Cairo_jll"; compat="1.16.1"),
+    Dependency("Cairo_jll"; compat="1.18.0"),
     Dependency("Fontconfig_jll"),
-    Dependency("FreeType2_jll"; compat="2.10.4"),
+    Dependency("FreeType2_jll"; compat="2.13.1"),
```
giordano, line 73: For the record, this is because #7454 changed the compat bound of `FreeType2_jll` to 2.13.1, but I have no clue of why: https://github.com/JuliaPackaging/Yggdrasil/pull/7454/files#diff-e2d8f4bfb63a6c0200a9e9beeb578adef317bf2a540738ea6b7858df24e1115fR76

PR #8500, opened 2024-04-20
Add `SuperCollider` recipe

mofeing:
This PR adds the `scsynth` and `supernova` audio servers.

### DIFF
--- a/S/SuperCollider/build_tarballs.jl
+++ b/S/SuperCollider/build_tarballs.jl
@@ -1,0 +1,43 @@
+using BinaryBuilder, Pkg
+
+name = "SuperCollider"
+version = v"3.13.0"
+
+sources = [
+    GitSource("https://github.com/supercollider/supercollider.git", "31885032db394f4d2b7d39eaf90e17927383e39d")
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/supercollider
+
+cmake -G Ninja \
+    -S . \
+    -B build \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DBUILD_TESTING=OFF \
+    -DENABLE_TESTSUITE=OFF \
+    -DLIBSCSYNTH=ON \
+    -DSC_IDE=OFF \
+    -DSC_QT=OFF \
+    -DSC_USE_QTWEBENGINE=OFF \
+    -DSUPERNOVA=OFF
+
+cmake --build build --parallel ${nproc} --target all
+cmake --build build --target install
+"""
+
+platforms = supported_platforms()
+
+products = [
+    ExecutableProduct("scsynth", :scsynth),
+    LibraryProduct("libscsynth", :libscsynth),
+    ExecutableProduct("supernova", :supernova),
+]
+
+dependencies = [
+    Dependency("libportaudio_jll", v"19.7.0"),
+]
+
+build_tarballs(ARGS, name, version, sources, script, products, dependencies)


S/SuperCollider/build_tarballs.jl
```diff
@@ -0,0 +1,95 @@
+using BinaryBuilder, Pkg
+
+name = "SuperCollider"
+version = v"3.14.0"
+
+sources = [
+    GitSource("https://github.com/supercollider/supercollider.git", "d263b8cc9905302bb6cfc26e8e68f22bb43092e9"),
```
fingolfin, line 7: There is a release tarball at <https://github.com/supercollider/supercollider/releases/tag/Version-3.14.0> resp. <https://github.com/supercollider/supercollider/releases/download/Version-3.14.0/SuperCollider-3.14.0-Source.tar.bz2>, any particular reason to not just use that? (Perhaps because this PR predates that release... which was in July :-/ so yeah, sorry for no review here for so long :-/ )

S/SuperCollider/build_tarballs.jl
```diff
@@ -0,0 +1,95 @@
...
+    # ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
+    #     "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
```
fingolfin, line 9:
```suggestion
```

S/SuperCollider/build_tarballs.jl
```diff
@@ -0,0 +1,95 @@
...
+#     EXTRA_CMAKE_ARGS+=(-DCMAKE_CXX_FLAGS="-fno-aligned-allocation")
+
+#     pushd ${WORKSPACE}/srcdir/MacOSX10.*.sdk
+#     rm -rf /opt/${target}/${target}/sys-root/System
+#     cp -ra usr/* "/opt/${target}/${target}/sys-root/usr/."
+#     cp -ra System "/opt/${target}/${target}/sys-root/."
+#     export MACOSX_DEPLOYMENT_TARGET=10.15
+#     popd
+# fi
+
```
fingolfin, line 35: I don't like having such commented out sections of code without explanations. If you look at the code in a year and wonder "why is this here again" it might be tough to answer...

So my personally recommendation: Either remove this; or add a comment explaining why it is commented out and why/when it might become useful again?
```suggestion
```

S/SuperCollider/build_tarballs.jl
```diff
@@ -0,0 +1,95 @@
...
+    -DSC_ABLETON_LINK=OFF \
+    -DSC_HIDAPI=OFF \
+    -DSUPERNOVA=ON \
+    -DNO_X11=ON \
+    -DSCLANG_SERVER=OFF \
+    -DINSTALL_HELP=OFF \
+    "${EXTRA_CMAKE_ARGS[@]}"
+
+# $host_bindir/cmake --build build --parallel ${nproc} --target scsynth supernova
+
```
fingolfin, line 56: Again: comment or remove?
```suggestion
```

S/SuperCollider/build_tarballs.jl
```diff
@@ -0,0 +1,95 @@
...
+    Platform("i686", "linux"; libc="glibc"),
+    Platform("x86_64", "linux"; libc="glibc"),
+    Platform("aarch64", "linux"; libc="glibc"),
+    Platform("powerpc64le", "linux"; libc="glibc"),
+    Platform("i686", "linux"; libc="musl"),
+    Platform("x86_64", "linux"; libc="musl"),
+    Platform("aarch64", "linux"; libc="musl"),
+    # Platform("x86_64", "macos"),
+    Platform("aarch64", "macos"),
+]
```
fingolfin, line 81: Instead of an explicit list of platforms, perhaps consider starting with the full list and filtering out unsupported ones (possibly with a comment as to why they are not supported, e.g. you might filter for the OS being Linux or macOS and state that only those are supported but not e.g. Windows or FreeBSD).

The advantage is that if things like Linux-on-RISCV become available, they are automatically added (if they work), or lead to an error and can be dealt with explicit, instead of simply being "forgotten".

S/SuperCollider/build_tarballs.jl
```diff
@@ -0,0 +1,95 @@
...
+
+products = [
+    ExecutableProduct("scsynth", :scsynth),
+    ExecutableProduct("supernova", :supernova),
+    # ExecutableProduct("sclang", :sclang),
```
fingolfin, line 86: Again: remove or add a comment?
```suggestion
    # ExecutableProduct("sclang", :sclang),
```

S/SuperCollider/build_tarballs.jl
mofeing: Exactly because of that. It can be changed but is there a preferences between `GitSource` and `ArchiveSource`?

PR #8501, opened 2024-04-20
[LightGBM] Update to v4.0.0

characat0:
And remove windows build patch since it was already fixed.


imciner2:
Compiler segfaults are mysterious things. Probably just a bug they managed to track down and fix in between the versions.

imciner2:
I also see that upstream is already up to 4.3.0 but this was only 4.0. Do you want to do another PR to update it further to the upstream version?

characat0:
Yes, I will make that PR, now I'm working on adding GPU support

PR #8502, opened 2024-04-21
Reactant macOS

wsmoses:


wsmoses:
@vchuravy @giordano  is there any way to get a later target clang/lld install? See https://github.com/llvm/llvm-project/issues/90283 in which the target toolchain segfaults :'(

wsmoses:
I suppose also cc @staticfloat

giordano (approved):
I had a very quick look, I think this recipe could be slightly cleaned up, but current version is OK, and bazel is so messed anyway

PR #8517, opened 2024-04-23
[Geant4] Upgrade version to 11.2.1 and add windows platform

peremato:
- Needed to add patch file for windows

### DIFF
--- a/G/Geant4/build_tarballs.jl
+++ b/G/Geant4/build_tarballs.jl
@@ -3,14 +3,15 @@
 using BinaryBuilder
 
 name = "Geant4"
-version = v"11.2.0"
+version = v"11.2.1"
 
 # Collection of sources required to build
 sources = [
     ArchiveSource("https://gitlab.cern.ch/geant4/geant4/-/archive/v$(version)/geant4-v$(version).tar.gz",
-                  "9ff544739b243a24dac8f29a4e7aab4274fc0124fd4e1c4972018213dc6991ee"),
+                  "76c9093b01128ee2b45a6f4020a1bcb64d2a8141386dea4674b5ae28bcd23293"),
     ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
                   "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
+    DirectorySource("./bundled")
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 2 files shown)


G/Geant4/bundled/patches/windows.patch
```diff
@@ -0,0 +1,118 @@
...
+--- a/source/processes/hadronic/models/lend/include/nf_specialFunctions.h
++++ b/source/processes/hadronic/models/lend/include/nf_specialFunctions.h
+@@ -11,7 +11,7 @@
+ 
+ #include "nf_utilities.h"
+ 
+-#ifdef WIN32
++#ifdef _WIN32
+ #define M_PI 3.141592653589793238463
+ #endif
```
giordano, line 77: Side note, the way to get `M_PI` on Windows is to define `_USE_MATH_DEFINES` before loading `math.h`/`cmath`: https://learn.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170

G/Geant4/bundled/patches/windows.patch
```diff
@@ -0,0 +1,118 @@
...
+index afa8a06987..e6e8edfe26 100644
+--- a/source/processes/hadronic/models/lend/src/xDataTOM_importXML.cc
++++ b/source/processes/hadronic/models/lend/src/xDataTOM_importXML.cc
+@@ -11,8 +11,8 @@
+ #include <fcntl.h>
+ #include <errno.h>
+ 
+-#if defined(WIN32) || defined(__MINGW32__)
+-#include <BaseTsd.h>
++#if defined(_WIN32) || defined(__MINGW32__)
```
giordano, line 114: This is equivalent to
```suggestion
+#if defined(_WIN32)
```
no?

G/Geant4/bundled/patches/windows.patch
```diff
@@ -0,0 +1,118 @@
...
+ #endif
+ 
+diff --git a/source/global/management/src/G4Threading.cc b/source/global/management/src/G4Threading.cc
+index 3c6bd77b14..fe09e672c7 100644
+--- a/source/global/management/src/G4Threading.cc
++++ b/source/global/management/src/G4Threading.cc
+@@ -35,7 +35,7 @@
+ #include "globals.hh"
+ 
+ #if defined(WIN32) || defined(__MINGW32__)
```
giordano, line 34: Also here
```c
 #if defined(_WIN32)
```
is more sensible

G/Geant4/bundled/patches/windows.patch
giordano, line 1: It'd be great to upstream these changes :slightly_smiling_face:

G/Geant4/bundled/patches/windows.patch
peremato: Yes I know, I used in one occasion.  But I wanted to do minimal patch. They define this several times.

G/Geant4/build_tarballs.jl
```diff
@@ -101,4 +109,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"8", julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"11", julia_compat="1.6")
```
giordano, line 112: GCC 11 can be problematic for Julia v1.6, for C++ code: https://github.com/JuliaPackaging/Yggdrasil/blob/9e93401baf07c15c00d239fc6c4d124f8121451f/CONTRIBUTING.md#compatibility-tips.  Would GCC 10 be sufficient?

peremato: Should be also OK I guess. I'll run a test

G/Geant4/bundled/patches/windows.patch
peremato: Yes of course. I was planing on it.

G/Geant4/build_tarballs.jl
peremato: Done. The release notes says that has also been tested with gcc10

G/Geant4/bundled/patches/windows.patch
peremato: Done

PR #8518, opened 2024-04-23
[LightGBM]: Rebuild v3.3.5

characat0:
The commit for this version is lost in the original repo, so I'm using a fork of the project I made.
I'm rebuilding this since we found a crash in Windows for Julia 1.8, this is related to https://github.com/JuliaLang/julia/issues/48187 and https://discourse.julialang.org/t/issue-with-xgboost-jl-and-libsvm-jl-when-julia-1-8-4/92396.
This build now should work since it was fixed in https://github.com/JuliaLang/julia/pull/50135.

Also, is it okey to just overwrite build_tarballs.jl?
I saw other projects using Package@version folders inside the main package folder but I'm assuming this works too.

### DIFF
--- a/L/LightGBM/build_tarballs.jl
+++ b/L/LightGBM/build_tarballs.jl
@@ -3,11 +3,12 @@
 using BinaryBuilder, Pkg
 
 name = "LightGBM"
-version = v"4.0.0"
+version = v"3.3.5"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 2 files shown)


giordano:
> The commit for this version is lost in the original repo, so I'm using a fork of the project I made.

You should be able to use the hash of the tag https://github.com/microsoft/LightGBM/releases/tag/v3.3.5

characat0:
> You should be able to use the hash of the tag https://github.com/microsoft/LightGBM/releases/tag/v3.3.5

I tried that locally and got a weird error saying something about a inexistent git object, but I cannot reproduce it now, so I'm changing that back, thanks!

PR #8520, opened 2024-04-23
[Geant4-julia] Use latest Geant4_jll (including windows)

peremato:


fingolfin:
@peremato if you update this JLL again, may I suggest getting rid of the custom `julia_versions` here, so that it automatically makes use of the default one provided by libjulia_jll and hence automatically supports all versions available? Then also e.g. Julia dev versions can already be tested.

fingolfin:
Also the custom `julia_compat` could just be deleted, replacing it by `julia_compat="1.6"` in the `build_tarballs` call.

peremato:
@fingolfin Perhaps I do not completely understand what I am doing, but since this wrapper binary package links directly with `libjulia` I need to have one artefact instance per Julia version, I thought this was the way of listing the versions I would like to support.

PR #8522, opened 2024-04-24
[open62541] v1.3.10, split off v1.4

thomvet:
After building open62541 v1.4 with https://github.com/JuliaPackaging/Yggdrasil/pull/7889 rebuilding the wrapper package using Clang.jl proves difficult. Segfaults are encountered when running simple server test cases and figuring out the exact source has been difficult for me so far.

With this PR, I would like to build v.1.3.10 without OpenSSL support and without header amalgamation, while having v.1.4 separately around (unchanged). The hope is that I can first update the generator scripts used in Clang.jl to reflect the lack of amalgamated header file on the patch release 1.3.10 without introducing multiple larger changes (addition of OpenSSL + upgrade to v1.4 (which unfortunately seems to contain some breaking changes - at least as seen from the perspective of my wrapper))

Hopefully I did the splitting of the versions correctly...


O/open62541/open62541@1.3/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+platforms = supported_platforms(; experimental=true)
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libopen62541", :libopen62541)
+]
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
giordano, line 44: Missing newline
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")

```

O/open62541/bundled@1.4/patches/0001-freebsd.patch
giordano, line 1: I think it's easier to move this under `O/open62541/open62541@1.4/bundled/patches`

O/open62541/open62541@1.3/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+    -DBUILD_SHARED_LIBS=ON \
+    -DUA_FORCE_WERROR=OFF \
+    ..
+make -j${nproc}
+make install
+install_license ../LICENSE
+"""
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; experimental=true)
```
giordano, line 35: `experimental` does nothing
```suggestion
platforms = supported_platforms()
```

PR #8525, opened 2024-04-24
[LLVM@17] Update Julia compat to v1.12

mofeing:


giordano:
This isn't super relevant, and will have problems in the registry anyway.  I think we can just ignore this.

PR #8536, opened 2024-04-26
[FLINT] bump to v3.1.3

thofma:


thofma:
CC: @benlorenz @fingolfin

lgoettgens:
I think you are missing the bump of the upstream version in line 28.

thofma:
> I think you are missing the bump of the upstream version in line 28.

Thanks

PR #8546, opened 2024-04-27
[Singular] rebuild for new flint version

thofma:


thofma:
This binary + libsingular_julia_jll v0.44.2+0 makes the Singular test pass for me locally. So I think just rebuilding should be fine?

I also hope that bumping the offset this way gives the correct version.

fingolfin:
This seems plausible to me.

PR #8568, opened 2024-04-29
RustFFT: New version 0.3.2

Taaitaaiger:
Like the previous update, a rebuild is sufficient to pick up the latest version of jlrs.

### DIFF
--- a/R/rustfft/build_tarballs.jl
+++ b/R/rustfft/build_tarballs.jl
@@ -8,7 +8,7 @@ uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "rustfft"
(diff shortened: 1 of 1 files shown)


Taaitaaiger:
Thanks for the advice!

The rustfft-jl crate depends on jlrs, which requires setting a julia-1-x feature to handle breaking changes in the Julia C API (e.g. the recent removal of many `jl_array_*`-functions). Does the default libjulia_jll offer a way to handle these changes in some way?

fingolfin:
I am not sure I understand your question. The changes I suggested should have no effect on any consumers of the JLL: Julia will make sure install the correct artifact, period.

I know nothing about the "rustfft-jl crate" or "jlrs", but assuming what you are saying is that they need to use different code depending on which Julia version is being used, this seems orthogonal to what I suggest, too. I.e. whatever you are doing now should keep working if it is not too crazy.

PR #8578, opened 2024-05-01
[libpolymake_julia] rebuild for flint 3

thofma:


thofma:
@benlorenz please double check the version numbers

L/libpolymake_julia/build_tarballs.jl
```diff
@@ -71,14 +71,14 @@ dependencies = [
     BuildDependency("GMP_jll"),
     BuildDependency("MPFR_jll"),
     Dependency("CompilerSupportLibraries_jll"),
-    Dependency("FLINT_jll", compat = "~200.900.004"),
+    Dependency("FLINT_jll", compat = "~300.100.300"),
     Dependency("TOPCOM_jll"; compat = "~0.17.8"),
     Dependency("lib4ti2_jll"; compat = "^1.6.10"),
     Dependency("libcxxwrap_julia_jll"; compat = "~0.11.2"),
     Dependency("polymake_jll"; compat = "~400.1100.1"),
```
thofma, line 78:
```suggestion
    Dependency("polymake_jll"; compat = "~400.1100.3"),
```

barche:
I recommend going to libcxxwrap_julia_jll v0.11.3

barche:
Sorry for the previous and now deleted comment, I was confused between libcxxwrap 0.11 and 0.12. 0.12.3 should be used if you want to move to CxxWrap 0.15, otherwise the compat should stay as it is.

fingolfin (approved):
Looks good to me but obviously @benlorenz should make the call

fingolfin:
@barche in principle I guess we would like to switch to a newer CxxWrap, but we have to coordinate this between all our packages, and right now the FLINT3 switch is more urgent -- better not to mix the two.

(BTW I think at some point it would be great to talk to you about your view on some potentially breaking changes for CxxWrap by reducing the number of automatic conversions further, as they cause massive method invalidations in any packages loaded after CxxWrap, even those which don't use CxxWrap at all (not even indirectly) -- convenience is great and all, but the cost seems too high to me... -- see also the thread around [this discourse post of mine](https://discourse.julialang.org/t/how-to-help-reduce-package-load-latency/105294/6)).

PR #8587, opened 2024-05-02
CppInterOp: Add build script

Gnimuc:
The script is copied from libclangex and tweaked a bit.🤞🏻


Gnimuc:
Windows build failures are due to that CppInterOp doesn't support a non-msvc compiler.

The failure on macOS x86_64 seems to be a SDK version issue.

I don't know why the cmake config file can not be found on i686-linux-musl.

giordano:
> The failure on macOS x86_64 seems to be a SDK version issue.

What SDK version you need? You can download a different one, we do it in several places.

> I don't know why the cmake config file can not be found on i686-linux-musl.

Because that's true: we can't build llvm (and Julia) for that platform.

mcbarton:
> Windows build failures are due to that CppInterOp doesn't support a non-msvc compiler.

Windows doesn't completly work at the moment even with an MSVC compiler. There are failing tests where the issue has yet to be resolved.

mcbarton:
https://github.com/conda-forge/cppinterop-feedstock - This conda forge feedstock may help with getting it built. It also has several patches which were found to needed to be applied to version 1.2.0 .

Gnimuc:
@giordano I rewrote the CMake scripts and stripped all "musl" platforms. Now I think this is OK to be merged.

L/libCppInterOp/build_tarballs.jl
```diff
@@ -0,0 +1,94 @@
...
+     -DClang_DIR=${prefix}/lib/cmake/clang \
+     -DCMAKE_BUILD_TYPE=Release
+make -j${nproc}
+make install
+install_license ../LICENSE.txt
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = expand_cxxstring_abis(supported_platforms(; experimental=true))
```
giordano, line 36:
```suggestion
platforms = expand_cxxstring_abis(supported_platforms())
```

L/libCppInterOp/build_tarballs.jl
```diff
@@ -0,0 +1,94 @@
...
+filter!(p -> libc(p) != "musl", platforms) # LLVM_full+asserts isn't available for musl
```
giordano, line 37: That's true only for i686-linux-musl, not all musl-based platforms.

Gnimuc: Fixed!

PR #8589, opened 2024-05-02
AlphaMolWrapper: new version 0.3.0

IvanSpirandelli:
Including overlap calculations for unions of balls utilizing AlphaMols Alphacomplex data structure.


fingolfin:
@IvanSpirandelli if you update this JLL again, may I suggest getting rid of the custom `julia_versions` here, so that it automatically makes use of the default one provided by libjulia_jll and hence automatically supports all versions available? Then also e.g. Julia dev versions can already be tested.

IvanSpirandelli:
@fingolfin Will do! Thanks for pointing that out.

IvanSpirandelli:
@fingolfin Sorry to ping you, but I was just updating the JLL again and was wondering, if I am doing what you asked me to correctly: 

This is how my new PR looks. https://github.com/JuliaPackaging/Yggdrasil/pull/9772/files 
Also you said I should put Julia_compat="1.6" in the build_tarballs call, but this was not in the custom compats before, so I just put in "1.7". Is that correct or should is there something going on under the hood that I don't get. 

(I am not entirely sure why I started at 1.7 in the first place, but recall that I had some reason to do so)

PR #8591, opened 2024-05-02
Reactant wip cuda

wsmoses:


R/Reactant/build_tarballs.jl
```diff
@@ -237,12 +282,22 @@ for platform in platforms
...
               ArchiveSource("https://github.com/llvm/llvm-project/releases/download/llvmorg-18.1.4/llvm-project-18.1.4.src.tar.xz",
                             "2c01b2fbb06819a12a92056a7fd4edcdc385837942b5e5260b9c2c0baff5116b"))
+
+    end
+
+    if CUDA.is_supported(platform)
+        cuda_deps = CUDA.required_dependencies(platform, static_sdk=true)
+        push!(cuda_deps, BuildDependency(PackageSpec(name="CUDNN_jll")))
+        push!(cuda_deps, BuildDependency(PackageSpec(name="TensorRT_jll")))
+        push!(cuda_deps, BuildDependency(PackageSpec(name="CUDA_full_jll")))
```
imciner2, line 292:
```suggestion
```
This package isn't used anymore and instead is replaced by the CUDA_SDK that is added with your earlier call to `CUDA.required_dependencies`.

wsmoses: No, it is required presently, sadly (otherwise various headers are not found, and a manual find indicates they are not installed otherwise).

wsmoses: Specifically without it cuda.h among others are missing from the destdir/include and/or destdir/cuda/include, etc (and I did not see anywhere in the checkout).

vchuravy: cc: @maleadt

imciner2: The CUDA headers aren't in `destdir/include`, they are in `destdir/cuda/include`. The `cuda` directory is where CUDA puts everything in the CUDA_SDK archive, so really `CUDA_HOME` needs to be set to `destdir/cuda` instead of just `destdir`. See the SuiteSparse GPU recipe here: https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SuiteSparse/SuiteSparse_GPU%407/build_tarballs.jl.

PR #8593, opened 2024-05-03
[LightGBM] build v4.3.0 with OpenCL/CUDA support

characat0:


L/LightGBM/build_tarballs.jl
```diff
@@ -17,27 +21,49 @@ cd $WORKSPACE/srcdir/LightGBM
...
+configure () {
+  cmake . \
+    -DCMAKE_INSTALL_PREFIX=$prefix\
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    $cmake_extra_args
+}
+
+# Need to call twice due to FindCUDA issues in cmake
+configure || configure
```
imciner2, line 65: What issues? I haven't seen problems with CMake's CUDA support requiring two configuration steps to solve.

characat0: it has to do with cmake finding binaries and libs of cuda https://github.com/JuliaPackaging/Yggdrasil/pull/4554#discussion_r831155780.
CUDA build throw this error the first time cmake is called https://buildkite.com/julialang/yggdrasil/builds/10093#018f3bfb-663a-4a64-a57f-d5148ee1dbdb/660-912
```
[01:16:29] CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
[01:16:29]   Could NOT find CUDA (missing: CUDA_INCLUDE_DIRS CUDA_CUDART_LIBRARY) (found
[01:16:29]   suitable version "11.5", minimum required is "11.0")
[01:16:29] Call Stack (most recent call first):
[01:16:29]   /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
[01:16:29]   /usr/share/cmake/Modules/FindCUDA.cmake:1264 (find_package_handle_standard_args)
[01:16:29]   CMakeLists.txt:201 (find_package)
```

imciner2: They shouldn't be using the FindCUDA module via `find_package(CUDA)`, since it has been deprecated since CMake 3.10: https://cmake.org/cmake/help/latest/module/FindCUDA.html. Instead they just need to enable the language for CUDA and CMake should take care of the rest of the setup for the base CUDA libraries.

characat0: thanks, let me see if I can get the change upstream, do you think we can merge this as is or this need a patch to their CMakeLists.txt?

ViralBShah: Should this be ok to merge?

imciner2: Yea, I don't think this is blocking for it, it is just very odd it takes two tries to pass... but if it works :man_shrugging:

ViralBShah:
@giordano ok to merge?

giordano:
I quickly skimmed through it and it looks ok, but is there an upstream issue to switch to FindCUDAToolkit?  The deprecated FindCUDA module recently [caused me lots of wasted time](https://github.com/spack/spack/pull/42437), it should ***really*** not be used.

characat0:
I will make a patch here and make an issue upstream

characat0:
can we check https://github.com/JuliaPackaging/Yggdrasil/pull/8691?
I think is the reason build are failing

L/LightGBM/build_tarballs.jl
giordano, line 1: Why the mode changed?

L/LightGBM/build_tarballs.jl
```diff
@@ -46,20 +70,43 @@ install_license LICENSE
...
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
-    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, all_platforms)),
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, all_platforms)),
+    
+    # GPU support is enabled via OpenCL and is hard to cross compile for other platforms
+    Dependency(PackageSpec(name="OpenCL_jll", uuid="6cb37087-e8b6-5417-8430-1f242f1e46e4"); platforms=filter(Sys.islinux, all_platforms)),
+    BuildDependency(PackageSpec(name="OpenCL_Headers_jll", uuid="a7aa756b-2b7f-562a-9e9d-e94076c5c8ee"); platforms=filter(Sys.islinux, all_platforms)),
+
+    Dependency(PackageSpec(name="boost_jll", uuid="28df3c45-c428-5900-9ff8-a3135698ca75"); platforms=filter(Sys.islinux, all_platforms)),
```
giordano, line 97: #3024
```suggestion
    Dependency(PackageSpec(name="boost_jll", uuid="28df3c45-c428-5900-9ff8-a3135698ca75"); compat="=1.76.0", platforms=filter(Sys.islinux, all_platforms)),
```

L/LightGBM/bundled/patches/lightgbm-cmake-find-cuda.patch
giordano, line 1: Has the patch been upstreamed?

L/LightGBM/build_tarballs.jl
characat0: tbh, I have no idea, I did some editing on my phone, maybe the web editor messed with the file modes

characat0:
Are we ok to merge @giordano ?

L/LightGBM/build_tarballs.jl
```diff
@@ -17,27 +22,46 @@ cd $WORKSPACE/srcdir/LightGBM
...
+
+if  [[ ("${bb_full_target}" == *-cuda*) && ("${bb_full_target}" != *-cuda+none*) ]]; then
+  # nvcc writes to /tmp, which is a small tmpfs in our sandbox.
+  # make it use the workspace instead
+  export TMPDIR=${WORKSPACE}/tmpdir
+  mkdir ${TMPDIR}
+  export CUDA_PATH=${WORKSPACE}/destdir/cuda
+  export PATH=$PATH:$CUDA_PATH/bin/
+
+  ln -s $CUDA_PATH/lib $CUDA_PATH/lib64
```
giordano, line 53: I think this leaves a broken symlink in the final tarballs?

characat0: I found that the issue was in the enable_language(CUDA) call, it was not using the library path provided, so I'm forcing it now with CMAKE_CUDA_FLAGS, is this a good approach?

L/LightGBM/build_tarballs.jl
```diff
@@ -17,27 +22,43 @@ cd $WORKSPACE/srcdir/LightGBM
 git submodule update --init --depth=1
 git submodule update --checkout --depth=1
 
-
-if [[ $target == *"apple-darwin"* ]]; then
-  cmake_extra_args="-DAPPLE=1 -DAPPLE_OUTPUT_DYLIB=1"
-fi
+for p in $WORKSPACE/srcdir/patches/*.patch; do
+  atomic_patch -p1 "${p}"
+done
```
giordano, line 27:
```suggestion
# https://github.com/microsoft/LightGBM/pull/6457
atomic_patch -p1 "$WORKSPACE/srcdir/patches/lightgbm-cmake-find-cuda.patch"
```

PR #8605, opened 2024-05-03
[GR] bump version

jheinen:
Bump GR run-time version  to 0.73.5


jheinen:
@giordano : There were build errors in 0.73.4 on FreeBSD, so we had to fix it and make a new release.

jheinen:
@giordano : Thanks a lot for the quick reaction 💯

PR #8607, opened 2024-05-03
Reactant fix registration issue

wsmoses:
See https://buildkite.com/julialang/yggdrasil/builds/10149#018f3f9b-9218-438c-8bdd-79880854b42e

```

--> Uploading: Reactant-logs.v0.0.4.aarch64-apple-darwin.tar.gz
--
  | --> Uploading: Reactant-logs.v0.0.4.x86_64-linux-gnu-cxx03.tar.gz
  | ERROR: LoadError: TypeError: in keyword argument augment_platform_block, expected String, got a value of type Vector{String}
  | Stacktrace:
  | [1] top-level scope
  | @ /cache/build/yggy-amdci7-9/julialang/yggdrasil/.ci/register_package.jl:139
  | in expression starting at /cache/build/yggy-amdci7-9/julialang/yggdrasil/.ci/register_package.jl:139


```

### DIFF
--- a/R/Reactant/build_tarballs.jl
+++ b/R/Reactant/build_tarballs.jl
@@ -299,7 +299,7 @@ for platform in platforms
     should_build_platform(triplet(augmented_platform)) || continue
     push!(builds, (;
                    dependencies=[dependencies; cuda_deps], products, sources=platform_sources,
-        platforms=[augmented_platform], augment_platform_block::String, script=prefix*script
+        platforms=[augmented_platform], augment_platform_block=augment_platform_block::String, script=prefix*script
     ))
 end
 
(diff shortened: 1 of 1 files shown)


R/Reactant/build_tarballs.jl
```diff
@@ -292,7 +292,7 @@ for platform in platforms
         push!(cuda_deps, BuildDependency(PackageSpec(name="CUDNN_jll")))
         push!(cuda_deps, BuildDependency(PackageSpec(name="TensorRT_jll")))
         push!(cuda_deps, BuildDependency(PackageSpec(name="CUDA_full_jll")))
-        augment_platform_block=CUDA.augment
+        # augment_platform_block=CUDA.augment
```
vchuravy, line 295: hm this is necessary, otherwise it will load the wrong binaries later.

wsmoses: so ldd on our generated .so doesnt link to cuda (presumably static)

vchuravy: And https://github.com/JuliaPackaging/Yggdrasil/blob/720d1ec37613e02974dd8e548c52bb790247f95c/platforms/cuda.jl#L13 is just a string.

wsmoses:
Added some more type casts and such to see if we can identify concretely what's happening

PR #8615, opened 2024-05-05
[MUMPS] Upgrade to v5.7.0

amontoison:


boriskaus:
@amontoison thanks for upgrading. Are you aware of the fact that this upgrade broke a whole range of downstream packages, including `PETSc_jll` and `LaMEM_jll`?  Unfortunately, I only realised when some of the users pointed out that it no longer works for them.

giordano:
Is that because downstream packages don't set appropriate compat bounds?

boriskaus:
yes that is likely the reason; I put a minimum bound for the packages used to compile `PETSc_jll` but no upper bound.
You can thus blame it all on me.
Yet for me this also reveals a wider problem with BinaryBuilder: the lack of tests implies that there is no way of spotting this at an early stage. IMHO it would be very helpful if we can add tests to the various packages and use github actions to check if downstream dependencies still work if an upstream package is updated.

giordano:
Actually, from https://github.com/JuliaBinaryWrappers/MUMPS_jll.jl/commit/36d77ae093fdd8ed26382c21b73f56c71ca46e67 I can't tell what would break downstream packages.

boriskaus:
the errors are very cryptic indeed, as seen [here](https://github.com/JuliaGeodynamics/LaMEM.jl/actions/runs/8962595642/job/24611683971#step:6:2276) for example. Something internal must have changed between MUMPS 5.6.2 and MUMPS 5.7.0 which breaks the PETSc interface to mumps.

giordano:
Running all downstream packages to catch obscure errors like that will take forever....

boriskaus:
Well it would make for a more robust system. Or if we would at least receive a warning email saying: "hey, we are just about the update `MUMPS_jll`, please test that `PETSc_jll` still works" would already be helpful.

This is not the first time this happens; it basically occurs every time there is a new release of any of the dependencies of PETSc, including MPI, Scalapack, MUMPS, SuiteSparse, ....

If there is a way to use the PETSc installation system to download all required dependencies (e.g. as static libraries) rather than force me to rely on the dynamic libraries on Yggdrasil it would certainly prevent a lot of this. It would also save weeks of my time  ...

boriskaus:
> > Well it would make for a more robust system.
> 
> I mean, aspirationally yes, but that's totally impractical. PkgEval takes in the order of 9-12 _hours_ to run, if we had to wait that long for each single PR, development here would grind to a halt. Not talking about the infrastructure where to run this (are you offering to procure a dedicated high-powered workstation?) and setting up the entire pipeline (are you volunteering to do it?). Besides, even if we had that, the fact is that downstream packages of the package being updated have to react, then what? Wait for the developers of downstream packages to do something? Catching only ABI breakage would _theoretically_ be much simpler (although doing that still requires people to invest their time to figure out all the details and implement it, and that's not going to take a couple of hours), but what you ask is a lot more complex and time consuming.
> 

yes, I understand that this is complicated and may be too time consuming. 
What about an intermediate step: allow at least for optionally adding tests to BinaryBuilder packages? 
In that way users can write: 
```Julia
pkg> test MUMPS_jll
```
and check that the underlying package and dependencies still work on the system. Essentially every normal Julia package has tests at the moment. Many packages such as `SuperLU_Dist` and `PETSc` come with their own build-in test cases, which can be utilised for this. I realise that this may not be the case for every binary package on this planet, so I wouldn't make this obligatory.

With `PETSc_jll` I have also encountered runtime errors, where the library compiled fine on all system, but segfaulted while running. Running the tests through github actions at the end of compiling would have revealed that.

> > Or if we would at least receive a warning email saying: "hey, we are just about the update MUMPS_jll, please test that PETSc_jll still works" would already be helpful.
> 
> Sounds like something you can do yourself already right now with a scheduled github actions workflow.

If the packages had build-in tests, this would be way easier.

amontoison:
@boriskaus I met the MUMPS developers last week at SIAM Linear Algebra 24.

We talked about the new features of MUMPS 5.7.0 but no breaking changes.

Can you provide me the error such that I can send it to them?
We plan to discuss by email about a few other things this week so it should not take long to fix it.

amontoison:
@boriskaus The best solution is to tag you before that I merge a new release of MUMPS_jll.jl such that you can test it before that we merge the PR.

boriskaus:
> @boriskaus I met the MUMPS developers last week at SIAM Linear Algebra 24.
> 
> We talked about the new features of MUMPS 5.7.0 but no breaking changes.
> 
> Can you provide me the error such that I can send it to them? We plan to discuss by email about a few other things this week so it should not take long to fix it.

Thanks. It's not straightforward to reproduce this as the error occurs on our code LaMEM, which calls PETSc, which calls MUMPS. Link to one of the errors that occurs is [here](https://github.com/JuliaGeodynamics/LaMEM.jl/actions/runs/8962595642/job/24611683971#step:6:1042)

boriskaus:
> @boriskaus The best solution is to tag you before that I merge a new release of MUMPS_jll.jl such that you can test it before that we merge the PR.

thanks. PETSc usually relies on a particular version of packages, but if there are no API changes it indeed *should* work, except that it doesn't.

amontoison:
I suspect that they just modified the main C strcuture with the new release 5.7.0 and data are misaligned if we keep the C of MUMPS 5.6.2.
I will check that later this week.

I am wondering if you should not add strict compat in PETCS_jll and LaMEM_jll.
We did that for COIN-OR packages like Ipopt.jl:
https://github.com/JuliaPackaging/Yggdrasil/blob/master/C/Coin-OR/coin-or-common.jl#L3

We are sure that updating a dependencies is not breaking anymore the JuMP ecosystem.

boriskaus:
Yes I should certainly have done that and will do so in a future version. 

My last compilation of PETSc_jll was several weeks of work and I currently lack the time for that, which is why it hasn’t happened.
A quick fix was indeed to make `MUMPS_jll` a dependency of `LaMEM.jl` and fix the compat to 5.6.2.

PR #8629, opened 2024-05-07
[PAPI] Update to v7.1.0 and build for CUDA on aarch64-linux-gnu

giordano:
This version added support for Neoverse V1 and V2, see https://github.com/icl-utk-edu/papi/blob/3ce9001dff49e1b6b1653ffb429808795f71a0bd/RELEASENOTES.txt#L27.  CC: @mofeing.


giordano:
Lots of custom patches, yay :smiling_face_with_tear:

P/PAPI/bundled/patches/0001-CUDA-Add-environment-overrides.patch
giordano, line 1: @vchuravy do you still need this patch?  I was looking into refreshing this patch, but the relevant code changed completely, I couldn't see how to adapt the patch.

giordano:
@mofeing with this build, on neoverse-v2 I get
```
$ papi_avail
[...]
Of 108 possible events, 47 are available, of which 15 are derived.
```
:partying_face:

P/PAPI/bundled/patches/0001-CUDA-Add-environment-overrides.patch
vchuravy: Probably not, if it can find the right CUDA some other way. I did raise the issue upstream

giordano:
With this build
```
julia> using Libdl, CUDA

julia> CUDA.versioninfo()
CUDA runtime 12.4, artifact installation
CUDA driver 12.4
NVIDIA driver 550.54.14

julia> filter(contains("cuda"), dllist())
2-element Vector{String}:
 "/lib64/libcuda.so.1"
 "/home/cceamgi/.julia/artifacts/" ⋯ 26 bytes ⋯ "474b5bee456adf/lib/libcudart.so"

julia> using PAPI_jll

julia> filter(contains("cuda"), dllist())
2-element Vector{String}:
 "/lib64/libcuda.so.1"
 "/home/cceamgi/.julia/artifacts/" ⋯ 26 bytes ⋯ "474b5bee456adf/lib/libcudart.so"
```
I get the same in a fresh session with only `PAP_jll` loaded.  @vchuravy is it working as expected?

PR #8660, opened 2024-05-11
[Qt6Base] Version 6.7.0 targets build

barche:


Q/Qt6Base/build_tarballs.jl
```diff
@@ -157,7 +155,7 @@ products_macos = [
 ]
 
 # We must use the same version of LLVM for the build toolchain and LLVMCompilerRT_jll
-llvm_version = v"13.0.1"
+llvm_version = v"16.0.6"
```
giordano, line 158: I doubt this will work

barche: Fixing the error `/workspace/srcdir/MacOSX13.3.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSString.h:396:97: error: function does not return string type` requires a newer LLVM, do you think this will cause problems at install time?

giordano: The problem is that we don't have compiler rt for llvm 16, so you'd need to fix that first if you need it.

barche: OK, looks like I will need to do that, since the current workaround I tried was setting the deployment target to version 13. The runtime macOS version check gets omitted that way and enabling it requires linking to compiler-rt.

giordano: You want to look at #7877

PR #8665, opened 2024-05-12
[gmsh] Rebuild with Cairo 1.18

giordano:
Same stupid thing as #8497.  Ref https://github.com/JuliaFEM/Gmsh.jl/issues/33.  CC: @j-fu.


giordano:
We can't do this because we'd have to change the compat of FreeType.  Sigh.

j-fu:
Thanks for this! Unfortunatly, there seems to be [more of the same hidden](https://github.com/j-fu/ExtendableGrids.jl/actions/runs/9059108407/job/24886018089?pr=45#step:6:524)

Issue #7660, opened 2023-11-12
request: libflame


ChrisRackauckas:
BLIS does not export a full LAPACK (see https://github.com/JuliaLinearAlgebra/BLIS.jl/issues/3), but is supposed to be a part of flame https://github.com/flame/libflame. In LinearSolve.jl, I don't need (or want) libblastrampoline wrappers on here anyways, so if the jll exists it would be immediately useful to a large community and would finish (https://github.com/SciML/LinearSolve.jl/pull/431).

jd-foster:
`--enable-lapack2flame` provides a LAPACK 3.11 compatibility layer.
Note also this recent commit that updates on this work: https://github.com/flame/libflame/commit/74b56a226bf976be9a89e810b2c7bd0789c6d2f9

ViralBShah:
Does FLAME provide a complete compatiblity with LAPACK? It would be nice to include it in Ygg.

imciner2:
> Does FLAME provide a complete compatiblity with LAPACK? It would be nice to include it in Ygg.

I actually just saw some discussion on this that happened in the BLIS discord server recently, and apparently, libflame is ABI compatible with LAPACK 3.11 when it is built using `--enable-lapack`, but is not ABI compatible using their older f2c LAPACK wrapper layer.

So I think we can get a LAPACK version of libflame into Ygg.

ViralBShah:
And might also then transparently work with LBT. What fun!

ChrisRackauckas:
Any movement on this?

imciner2:
The initial script in https://github.com/JuliaPackaging/Yggdrasil/pull/8671 was working on all the platforms except for macOS - on there is was somehow creating a dylib that had absolutely no functions at all. No one could figure out exactly what was happening, and so it just kind of stalled.

ChrisRackauckas:
That's fine since you almost certainly want to use Apple accelerate on that platform anyways, so if it's a no op there then it's fine

imciner2:
The libflame build script is merged now.

PR #8671, opened 2024-05-14
[libflame] Initial build scripts.

jd-foster:
Build scripts for [libflame](https://github.com/flame/libflame)

Addresses #7660

~Currently MacOS is omitted until linking errors can be resolved cf. https://github.com/flame/libflame/issues/100~


imciner2:
Why is it split into a build_tarballs and a common? This looks simple enough we should just keep it all in a single build_tarballs to make updating the recipe simple.

jd-foster:
Yes, that also occurred to me after the initial commit. Will reduce to a single build_tarball

L/libflame/build_tarballs.jl
```diff
@@ -0,0 +1,78 @@
...
+    # Platform("x86_64", "macos"),
+    # Platform("aarch64", "macos"),
+]
+
+# Enable the following line if trying to link against Fortran code:
+# platforms = expand_gfortran_versions(platforms)
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    RuntimeDependency(PackageSpec(name="blis_jll", uuid="6136c539-28a5-5bf0-87cc-b183200dce32")),
```
imciner2, line 65: Does this actually require BLIS as the BLAS layer, or can this also work on our libblastrampoline system? It would be nicer if we can get it to use libblastrampoline (which then could use BLIS if the user selects it).

L/libflame/build_tarballs.jl
```diff
@@ -0,0 +1,78 @@
...
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc="musl"),
+    Platform("x86_64", "linux"; libc="glibc"),
+    Platform("aarch64", "linux"; libc="glibc"),
+    Platform("x86_64", "freebsd"),
+    Platform("x86_64", "windows"),
+    # Platform("x86_64", "macos"),
+    # Platform("aarch64", "macos"),
+]
```
imciner2, line 58:
```suggestion
platforms = supported_platforms()

filter!(!Sys.isapple(p), platforms)
filter!(p -> arch(p) != "i686", platforms)
filter!(p -> arch(p) != "armv7l", platforms)
filter!(p -> arch(p) != "armv6l", platforms)
filter!(p -> arch(p) != "powerpc64le", platforms)
```

IMO, it is a bit nicer to remove the platforms we don't want versus listing the ones we do like this. (warning, I haven't tested this code though :))

Also, what is keeping it from working on these other platforms? (especially the Apple platforms).

jd-foster: > IMO, it is a bit nicer to remove the platforms we don't want versus listing the ones we do like this. (warning, I haven't tested this code though :))

Ok, will do this.

> Also, what is keeping it from working on these other platforms? (especially the Apple platforms).

Not clear ATM, but it's not linking the dynamic library.

L/libflame/build_tarballs.jl
jd-foster: Will give this a go.

L/libflame/build_tarballs.jl
jd-foster: Alright, Apple platforms are now back. 
Now onto windows... 😫

jd-foster: Seems like we're close with the windows build.
Could you advise on what's going wrong in the build?

jd-foster:
So ` libflame.1.dll` !== `libflame-1.dll` !== `libflame.dll` ?

```
[ Info: Checking lib/libflame.1.dll with RPath list String[]
[ Info: Ignored system libraries msvcrt.dll, KERNEL32.dll
┌ Warning: lib/libflame.1.dll should be in `bin`!
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/L8tBh/src/Auditor.jl:233
┌ Warning: Simple buildsystem detected; Moving all `.dll` files to `bin`!
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/L8tBh/src/Auditor.jl:244
[ Info: Normalising timestamps in import library /cache/build/yggy-amdci7-15/julialang/yggdrasil/L/libflame/build/x86_64-w64-mingw32/NiP3KxjS/x86_64-w64-mingw32-libgfortran5-cxx11/destdir/lib/libflame.dll.a
[ Info: Checking license file
[ Info: Found license file(s): LICENSE
[ Info: Found a valid dl path libatomic-1.dll while looking for libflame
[ Info: Found a valid dl path libblastrampoline-5.dll while looking for libflame
[ Info: Found a valid dl path libflame.1.dll while looking for libflame
[ Info: Found a valid dl path libgcc_s_seh-1.dll while looking for libflame
[ Info: Found a valid dl path libgfortran-5.dll while looking for libflame
[ Info: Found a valid dl path libgomp-1.dll while looking for libflame
[ Info: Found a valid dl path libquadmath-0.dll while looking for libflame
[ Info: Found a valid dl path libssp-0.dll while looking for libflame
[ Info: Found a valid dl path libstdc++-6.dll while looking for libflame
[ Info: Found a valid dl path libwinpthread-1.dll while looking for libflame
[ Info: Could not locate libflame inside ["/cache/build/yggy-amdci7-15/julialang/yggdrasil/L/libflame/build/x86_64-w64-mingw32/NiP3KxjS/x86_64-w64-mingw32-libgfortran5-cxx11/destdir/bin"]
┌ Error: Built libflame but libflame still unsatisfied:
└ @ BinaryBuilder /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/L8tBh/src/AutoBuild.jl:946
ERROR: LoadError: Cannot continue with unsatisfied build products!
```

jd-foster:
Note: now that it's working on `mingw32`, we should see if it's possible to upstream the `configure` script patches, which are based on the newer BLIS scripts anyway.

imciner2:
I see that it is currently building the static library. In general, we do like to avoid having those in the main JLLs because they can really eat up space (the artifact here shows it is 15MB on Linux). Can we disable the static library build?


Also, are we sure the aarch64 macos build is working properly? I downloaded the artifact to check, and the dylib it makes is extremely tiny compared to the static library (and also compared to the size it is on Linux):

```
$ ll libflame.v5.2.0.x86_64-linux-gnu/lib/
/home/im1716/dev/julia/binary
Permissions Size User   Group  Date Modified Date Created Name
.rw-r--r--@  15M im1716 im1716  1 Jan  1970  20 May 10:07 libflame.a
lrwxrwxrwx@    - im1716 im1716  1 Jan  1970  20 May 10:07 libflame.so -> libflame.so.1.0.0
lrwxrwxrwx@    - im1716 im1716  1 Jan  1970  20 May 10:07 libflame.so.1 -> libflame.so.1.0.0
.rwxr-xr-x@ 8.7M im1716 im1716  1 Jan  1970  20 May 10:07 libflame.so.1.0.0

$ ll libflame.v5.2.0.aarch64-apple-darwin/lib/
/home/im1716/dev/julia/binary
Permissions Size User   Group  Date Modified Date Created Name
.rwxr-xr-x@  17k im1716 im1716  1 Jan  1970  20 May 10:07 libflame.1.0.0.dylib
lrwxrwxrwx@    - im1716 im1716  1 Jan  1970  20 May 10:07 libflame.1.dylib -> libflame.1.0.0.dylib
.rw-r--r--@ 7.9M im1716 im1716  1 Jan  1970  20 May 10:07 libflame.a
lrwxrwxrwx@    - im1716 im1716  1 Jan  1970  20 May 10:07 libflame.dylib -> libflame.1.0.0.dylib

```

Can someone on macos confirm what is in this dylib?

jd-foster:
```
$ nm libflame.v5.2.0.x86_64-apple-darwin/lib/libflame.dylib
                 U dyld_stub_binder
```

L/libflame/build_tarballs.jl
```diff
@@ -0,0 +1,89 @@
...
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
+]
+
+products = [
+    LibraryProduct("libflame", :libflame),
+]
+
+# Build the tarballs
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               preferred_gcc_version=v"11", lock_microarchitecture=false, julia_compat="1.6")
```
imciner2, line 89:
```suggestion
               preferred_gcc_version=v"11", clang_use_lld=false, lock_microarchitecture=false, julia_compat="1.6")
```
Lets try not using LLD and instead using the other linker to see if these flags work then.

jd-foster: Mixed up the gcc ld `-whole-archive` vs. clang lld `-force_load` but didn't seem to like it anyway. Might need to try your approach.

giordano: > Mixed up the gcc ld `-whole-archive` vs. clang lld `-force_load`

1. this isn't gcc vs clang, but Linux vs macOS
1. this is why we provide the [`flagon`](https://docs.binarybuilder.org/stable/build_tips/#utils_build_env) utility.

L/libflame/build_tarballs.jl
```diff
@@ -0,0 +1,89 @@
...
+filter!(p -> arch(p) != "armv7l", platforms)
+filter!(p -> arch(p) != "armv6l", platforms)
+filter!(p -> arch(p) != "powerpc64le", platforms)
+
+# Enable the following line if trying to link against Fortran code:
+# platforms = expand_gfortran_versions(platforms)
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"), compat="5.9.0"),
```
imciner2, line 76: Just a note, doing a compat to 5.9.0 here will limit this to Julia 1.11+ I think (LBT 5.9.0 is in 1.12 for sure, and was discussed as a backport for 1.11, but I don't think that has happened yet).

Julia 1.9.1+ and 1.10 had libblastrampoline 5.8 in it, and the only difference should just be new symbols for the new LAPACK, so I think it should be safe to limit to 5.8 instead.

jd-foster, line 76:
```suggestion
    Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"), compat="5.8"),
```

jd-foster: Thanks

jd-foster:
@imciner2 I think I've done as much as I can here with what I know. What do you think should be next steps?

ViralBShah:
Since the build is all green, is it ok to merge, or is mac still broken?

imciner2:
The macOS dylib is still tiny compared to the static library - 17.2kB vs. 7.9MB, so I think it is still broken.

jd-foster:
I think we should try to port over how BLIS does it, since the make script is a downstream improvement on libflame.

L/libflame/build_tarballs.jl
giordano: Haven't looked at the log, but is this a case of #10290?

imciner2: What were the symptoms of that error with single module?

ViralBShah:
Just wondering if there is a new upstream release that fixes the issue and gets this over the line.

ChrisRackauckas:
The use case here doesn't care about Mac. Can we just make it no-op on Mac and merge?

ChrisRackauckas:
I was trying this out https://github.com/SciML/LinearSolve.jl/pull/660 but it seems to have some issues with OpenMP. But more importantly, we were trying to get this as something we could have hardcoded to BLIS 😅, like how we have a way to just load `mkl_jll` and use the binary instaed of requiring global BLAS changes. However, this library seems like it's built to support LBT. Could it be reasonable to have a version of libflame that is just directly built with BLIS?

ViralBShah:
Enabled all the platforms for libflame in https://github.com/JuliaPackaging/Yggdrasil/pull/12328.

PR #8687, opened 2024-05-16
[GTPSA] bump version and use CMAKE_C_FLAGS

mattsignorelli:


giordano:
https://github.com/mattsignorelli/gtpsa/blob/d343b561c44e69d25bb327a656eeca08d2c43808/CMakeLists.txt#L10 [Please, don't](https://simonbyrne.github.io/notes/fastmath/)

mattsignorelli:
> https://github.com/mattsignorelli/gtpsa/blob/d343b561c44e69d25bb327a656eeca08d2c43808/CMakeLists.txt#L10 [Please, don't](https://simonbyrne.github.io/notes/fastmath/)

OK, I have removed fast math.

Any idea why cmake can't find threads?

giordano:
> Any idea why cmake can't find threads?

Quoting from https://docs.binarybuilder.org/stable/troubleshooting/#General-comments

>  The best recommendation is to access the build environment and carefully read the log files generated by the build systems: it is not uncommon that build systems would only print to screen misleading error messages, and the actual problem is completely different (e.g. "library XYZ can't be found", when the problem instead is that the command they run to look for library XYZ fails for unrelated reasons, for example for a wrong compiler flag used in the check).

If you read the log you'd have found that BinaryBuilder doesn't allow you to use fast-math, making the compiler throw an error whenever that flag is used, for example when looking for pthreads libraries.

mattsignorelli:
> If you read the log you'd have found that BinaryBuilder doesn't allow you to use fast-math, making the compiler throw an error whenever that flag is used, for example when looking for pthreads libraries.

Understood, thanks. This is ready for merging

G/GTPSA/build_tarballs.jl
```diff
@@ -32,6 +32,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
     Dependency(PackageSpec(name="OpenBLAS32_jll", uuid="656ef2d0-ae68-5445-9ca0-591084a874a2"))
     Dependency(PackageSpec(name="LAPACK32_jll", uuid="17f450c3-bd24-55df-bb84-8c51b4b939e3"))
```
giordano, line 37: BTW, why do you need lapack? Isn't openblas sufficient?

mattsignorelli: From the GTPSA github: https://github.com/MethodicalAcceleratorDesign/MAD-NG/blob/2adfb27c07313ac7b493d9abb09feb61ab07fea0/src/libgtpsa/README.GTPSA#L49-L50 

I will remove the CompilerSupportLibraries dependency though

giordano: OpenBLAS implements LAPACK, too.

mattsignorelli: Oh, I didn't realize that. It implements the LP64 interface?

Also, why do you suggest adding in CompilerSupportLibraries as a dependency?

giordano: > Also, why do you suggest adding in CompilerSupportLibraries as a dependency?

Did I? :flushed:

mattsignorelli: Nope! I misinterpreted the review, its removed now

mattsignorelli:
@giordano thanks for the input and comments! this should be ready now

mattsignorelli:
> Why GCC?

See here: https://github.com/JuliaPackaging/Yggdrasil/pull/7743#discussion_r1413800627 . Clang doesn't have this two-argument version of the malloc attribute
> Why not building in parallel? That'd save some compilation time.

Sure! I'll update the build script

>Regarding blas/lapack libraries, we can continue building with openblas32 for now, but I'd recommend looking at https://docs.binarybuilder.org/stable/build_tips/#link-blas for moving to libblastrampoline, as it'd make it possible to switch the blas/lapack backend more easily.

I agree, and I will make this change once I have a bit more time

giordano:
> See here: https://github.com/JuliaPackaging/Yggdrasil/pull/7743#discussion_r1413800627 . Clang doesn't have this two-argument version of the malloc attribute

A comment in the build script explaining that, with a reference to the offending call, may be useful, to avoid the same question in the next round of reviews because I'm sure I'll forget this detail :slightly_smiling_face:

mattsignorelli:
> A comment in the build script explaining that, with a reference to the offending call, may be useful, to avoid the same question in the next round of reviews because I'm sure I'll forget this detail 🙂

Sure, I've added a comment

G/GTPSA/build_tarballs.jl
```diff
@@ -3,19 +3,21 @@
...
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/mattsignorelli/gtpsa.git", "70e4c39dffd9e4fdf0192fe4c7ca0ced72bb913f")
+    GitSource("https://github.com/mattsignorelli/gtpsa.git", "394a20847b869a842c6a89f2af1a889c3a1c2813")
 ]
 
 # Bash recipe for building across all platforms
+# GCC >=11 is necessary because the source code uses the two-argument version
+# of the attribute malloc, see lines 73-75 in mad_mem.h (also unsupported by clang)
```
giordano, line 15:
```suggestion
# of the attribute malloc, see https://github.com/mattsignorelli/gtpsa/blob/394a20847b869a842c6a89f2af1a889c3a1c2813/code/mad_mem.h#L73-L75 (also unsupported by clang)
```

PR #8698, opened 2024-05-17
MAGEMin v1.4.5

NicolasRiel:
- added the option to use aH2O, aO2, aTiO2, aMgO, aFeO and aAl2O3 as buffers
- corrected normalization when hitting liquidus


NicolasRiel:
Thanks!

PR #8701, opened 2024-05-17
[LLVM] Ensure libclang_rt.osx.a is built on aarch64 Mac

barche:
This should fix compilation of PR #8660


giordano:
This has nothing to do with the compiler toolchain

barche:
libclang_rt.osx.a is built on x86_64 but not on aarch64, not sure how else I should fix this?

giordano:
This is the LLVM used by Julia itself, it has nothing to do with the compiler toolchain we used inside BinaryBuilder.  As I said in https://github.com/JuliaPackaging/Yggdrasil/pull/8660#discussion_r1604962675, you want to make #7877 work, I never had the time nor interest to finish that.

PR #8707, opened 2024-05-18
[Spot_julia] add versions in libjulia and libcxxwrap_julia dependencies

MaximeBouton:


S/Spot_julia/build_tarballs.jl
```diff
@@ -89,9 +85,9 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("libcxxwrap_julia_jll"),
-    BuildDependency("libjulia_jll")
+    BuildDependency(PackageSpec(;name="libjulia_jll", version=v"1.10.9")),
+    Dependency("libcxxwrap_julia_jll")
```
giordano, line 89: You should add a compat for `libcxxwrap_julia_jll`, depending on whatever version of CxxWrap.jl you want to to match.

PR #8719, opened 2024-05-20
Add CompilerSupportHeaders v1.1 builder

rayegun:
I am not 100% sure this is accurate, but it's fairly simple and should be easy to verify. I just added the C++ headers (I don't *think* these are in Clang_jll), and the omp.h and openacc.h headers. 

The goal here is to make runtime linking against Julia's bundled libgomp easy using Clang_jll.

### DIFF
--- a/C/CompilerSupportHeaders/CompilerSupportHeaders@v0.5/build_tarballs.jl
+++ b/C/CompilerSupportHeaders/CompilerSupportHeaders@v0.5/build_tarballs.jl
@@ -1,5 +1,0 @@
-include("../common.jl")
-
-build_csh(ARGS, v"1.0.5"; preferred_gcc_version=v"12", julia_compat="1.9")
-
-# Build trigger: 1
--- a/C/CompilerSupportHeaders/CompilerSupportHeaders@v0.5/build_tarballs.jl
+++ b/C/CompilerSupportHeaders/CompilerSupportHeaders@v0.5/build_tarballs.jl
@@ -1,0 +1,5 @@
+include("../common.jl")
+
+build_csh(ARGS, v"1.0.5"; preferred_gcc_version=v"12", julia_compat="1.9")
+
+# Build trigger: 1
--- a/C/CompilerSupportHeaders/common.jl
+++ b/C/CompilerSupportHeaders/common.jl
@@ -1,0 +1,36 @@
+using BinaryBuilder
+
+function build_csh(ARGS, version::VersionNumber;
+                   # Note: use preferred_gcc_version=v"100" to always force
+                   # latest compatible version.
+                   preferred_gcc_version::VersionNumber,
+                   julia_compat::String,
+                   )
+    name = "CompilerSupportHeaders"
+
+    script = raw"""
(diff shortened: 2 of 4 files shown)


rayegun:
@gbaraldi does this look right to you? 😇

C/CompilerSupportHeaders/common.jl
```diff
@@ -0,0 +1,36 @@
...
+                   # Note: use preferred_gcc_version=v"100" to always force
+                   # latest compatible version.
+                   preferred_gcc_version::VersionNumber,
+                   julia_compat::String,
+                   )
+    name = "CompilerSupportHeaders"
+
+    script = raw"""
+mkdir -p ${includedir}
+install_license /usr/share/licenses/GPL-3.0+
```
giordano, line 13: Duplicate?
```suggestion
```

C/CompilerSupportHeaders/CompilerSupportHeaders@v0.5/build_tarballs.jl
giordano, line 1: In general, you can't try to register multiple versions of a single package at the same time: it'll fail in the registry

C/CompilerSupportHeaders/CompilerSupportHeaders@v0.5/build_tarballs.jl
```diff
@@ -0,0 +1,5 @@
+include("../common.jl")
+
+build_csh(ARGS, v"1.0.5"; preferred_gcc_version=v"12", julia_compat="1.9")
```
imciner2, line 3: Did you swap the `build_csh` calls between 0.5 and 1.0 here?

rayegun: Yes thanks, I deleted those anyway to register 1.1 for now

C/CompilerSupportHeaders/common.jl
```diff
@@ -0,0 +1,35 @@
...
+    name = "CompilerSupportHeaders"
+
+    script = raw"""
+mkdir -p ${includedir}
+# Copy all the libstdc++ and libgomp headers:
+cp -Rv /opt/${target}/${target}/include/c++/ ${includedir} || true
+# This doesn't grab any of the std C headers, but they are in Clang_jll already
+cp -Rv /opt/${target}/lib/gcc/${target}/*/include/{omp.h,openacc.h} ${includedir} || true
+
+# Install license (we license these all as GPL3, since they're from GCC)
```
fingolfin, line 18: We don't license these files, so, better wording:
```suggestion
# Install license (these files are licensed under GPL3, since they are from GCC)
```

fingolfin (commented):
@rayegun are you still interested in this? If so, it has open questions in the comments about the need to include the `include/c++/` headers. I am also making a code change suggestion.

fingolfin:
@rayegun ping

ViralBShah:
Bump. Should we try to get this in or close?

PR #8728, opened 2024-05-21
[XRootD] updated to latest version 5.6.9 and added MacOS platform

peremato:


X/XRootD/build_tarballs.jl
```diff
@@ -3,68 +3,71 @@
...
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir
 mkdir build && cd build
 install_license ../xrootd-*/LICENSE
-cmake -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DENABLE_PERL=FALSE ../xrootd-*/
+cmake -DCMAKE_INSTALL_PREFIX=$prefix \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DCMAKE_CXX_STANDARD=17 \
```
imciner2, line 21: If you want C+=17, you should really be requesting a newer GCC version. Most projects use GCC 7 with C++17.

peremato: Yes, you are right. In other projects I used v8

PR #8731, opened 2024-05-21
[SuiteSparse] Add ParU / SPEX, move GPU and CXSparse to new common builder

rayegun:
This PR does a few different things:

1. Different builders duplicated quite a bit of code, instead create a `build_script` function which generates a base build script that the builders can prepend and append options to, but otherwise share most options. The new script also reduces build time for some of these new libraries
2. Switches CXSparse and SuiteSparse_GPU to use the common build script.
3. Adds new builders for ParU and SPEX using the common build script.

Does *not* rebuild the base SuiteSparse at this time since it is already at the most recent version. Can be switched to the common builder in the next version bump. 

Does *not* switch GraphBLAS yet or add LAGraph, this requires some downstream work on my part, and a new SuiteSparse release.


S/SuiteSparse/CXSparse/build_tarballs.jl
```diff
@@ -1,38 +1,37 @@
...
-"""
-
-platforms = supported_platforms(;experimental=true)
-
-dependencies = [
-    Dependency("SuiteSparse_jll")
-]
+# Add dependency on SuiteSparse_jll
+dependencies = append!(dependencies, [
+    BuildDependency(PackageSpec(name="LLVMCompilerRT_jll", uuid="4e17d02c-6bf5-513e-be62-445f41c75a11", version=v"13.0.1+1");
```
giordano, line 27: For what is worth, with #7877 now merged you can use v16.0.6 (but note that you always need to match the version of the llvm compiler used below).

S/SuiteSparse/ParU/build_tarballs.jl
```diff
@@ -0,0 +1,34 @@
...
+        -DSUITESPARSE_USE_SYSTEM_CCOLAMD=ON
+        -DSUITESPARSE_USE_SYSTEM_UMFPACK=ON
+        -DSUITESPARSE_USE_SYSTEM_CHOLMOD=ON
+    )
+""" * build_script(true)
+
+# Add dependency on SuiteSparse_jll
+dependencies = append!(dependencies, [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
```
giordano, line 27: Is this for a package which is eventually going to be a Julia dep?  `LLVMOpenMP_jll` isn't currently a stdlib, so we can't nilly willy add it as a dep.

S/SuiteSparse/SPEX/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+    )
+""" * build_script() * raw"""
+rm -f ${libdir}/libspexpython.*
+rm -f ${includedir}/suitesparse/spex_python_connect.h
+""" # remove python libs until a CMake variable is added.
+
+# Add dependency on SuiteSparse_jll
+dependencies = append!(dependencies, [
+    Dependency("GMP_jll"),
+    Dependency("MPFR_jll"),
```
giordano, line 30: Need to add compat for these.

S/SuiteSparse/SSGraphBLAS/build_tarballs.jl
```diff
@@ -1,46 +1,35 @@
...
 ]
 
-# Dependencies that must be installed before this package can be built
-dependencies = [
-    Dependency("CompilerSupportLibraries_jll"),
-]
+# Add dependency on SuiteSparse_jll
+dependencies = append!(dependencies, [
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
```
giordano, line 30: Same question as above about LLVM OpenMP

S/SuiteSparse/ParU/build_tarballs.jl
rayegun: No. CXSparse, ParU, SPEX, GraphBLAS, and LAGraph will remain outside of stdlibs. The plan is not to add any further sparse deps to the stdlib, with most development happening in packages.

S/SuiteSparse/SPEX/build_tarballs.jl
rayegun: There's no lower bounds listed in the source repo for these libs. Just set it to "6" and "4" respectively?

S/SuiteSparse/common.jl
```diff
@@ -71,17 +71,107 @@ products = [
...
+      -DCMAKE_INSTALL_PREFIX=${prefix} \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+      -DBUILD_STATIC_LIBS=OFF \
+      -DBUILD_TESTING=OFF \
+      -DSUITESPARSE_ENABLE_PROJECTS=${PROJECTS_TO_BUILD} \
+      -DSUITESPARSE_DEMOS=OFF \
+      -DSUITESPARSE_USE_STRICT=ON \
+      -DSUITESPARSE_USE_FORTRAN=OFF \
+      -DSUITESPARSE_USE_OPENMP=${USEOMP} \
+      -DSUITESPARSE_USE_CUDA=${USECUDA} \
```
imciner2, line 136: Looking at this common script, why keep this option here instead of making it an extra flag from `SuiteSparse_GPU`? There is nothing else in this script that uses this `USECUDA` environment variable. And actually, you are already passing this in as an additional variable in the GPU script.

S/SuiteSparse/common.jl
```diff
@@ -71,17 +71,107 @@ products = [
...
+else
+    CMAKE_OPTIONS+=(
+        -DSUITESPARSE_USE_64BIT_BLAS=NO
+    )
+fi
+
+# some of these are not used for a particular builder
+# but most are. BLAS handling is the big one which isn't always used.
+# It's likely easier to keep them in common.jl than add them elsewhere.
+# TODO: Can BLAS handling be done by upstream now?
```
imciner2, line 124: Probably not. I need to get stuff upstream into CMake itself to properly support libblastrampoline, otherwise CMake doesn't know anything about our BLAS libraries and their locations.

S/SuiteSparse/common.jl
rayegun: Default is ON and was erroring due to missing CUDA for some reason (rather than turning it off when missing CUDA. I suspect due to the strict setting). I should remove it from the GPU script

rayegun:
I didn't realize there's an existing SPEX builder, although it is fairly old at this point. I'd rather it be part of SuiteSparse since it is directly dependent on a number of SuiteSparse packages and seems to be fully integrated. I'd rather that SPEX not fall behind the other packages like it has right now.

rayegun:
Tim Davis just confirmed they will remain in sync. @giordano could we get this merged soon?

S/SuiteSparse/CXSparse/build_tarballs.jl
```diff
@@ -1,38 +1,37 @@
...
+    Dependency("SuiteSparse_jll"; compat = "=$SS_version_str"),
+    Dependency("CompilerSupportLibraries_jll")
+])
 
 products = [
     LibraryProduct("libcxsparse", :libcxsparse)
 ]
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, 
+               dependencies; julia_compat="1.11", preferred_llvm_version=v"16.0.6")
```
giordano, line 37:
```suggestion
               dependencies; julia_compat="1.11", preferred_llvm_version=LLVM_version)
```

PR #2134, opened 2020-11-18
+ qwtw library

ig-or:
another 2D plotting library, for Linux X64 only yet..

PR #8732, opened 2024-05-21
[qwtw] new version 2.13.2 (spectrogramm functionality improved)

ig-or:
a few useful improvements and features for qwtw library


Q/qwtw/build_tarballs.jl
```diff
@@ -52,4 +52,4 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat="1.6", preferred_gcc_version = v"9.1.0")
+               julia_compat="1.8", preferred_gcc_version = v"9.1.0")
```
giordano, line 55: Any specific reason for this change?

giordano (commented):
@barche this is using Qt_jll, should it be doing something else nowadays?

Q/qwtw/build_tarballs.jl
ig-or: no any specific reason for this. The only change that might be important here is a move from C++14 into C++20..

imciner2:
Why does this build script do `apk add g++`? We provide the g++ version alongside gcc, so this shouldn't be done (and CMake actually is ignoring the system version anyway, so this is just extraneous).

ig-or:
Without those apk add g++, it didn't worked.. More details: 
[]([](https://github.com/JuliaPackaging/Yggdrasil/pull/2134/commits/9eaa610ea83a91633f4b71b136b4d934934e51aa))

ig-or:
@giordano , @barche , do you think we can merge it ?

Q/qwtw/build_tarballs.jl
fingolfin: If there is no reason to change the julia_compat, then maybe just don't change it?

fingolfin:
```suggestion
               julia_compat="1.6", preferred_gcc_version = v"9.1.0")
```

PR #8733, opened 2024-05-22
Add builder for the zstd seekable format subproject

ararslan:
This is an extension of the zstd format and lives in zstd's `contrib/` directory. It's not built by default and has no build system of its own nor integration with the parent build system, so here it's built with a single manual call to `cc`.

There doesn't appear to be a single, official name for this subproject. I've opted to name the library `zstd_seekable` in keeping with its associated header file, `zstd_seekable.h`, and the prefix used for the library's symbols (`ZSTD_seekable` for all but a couple).

This could be built and distributed by Zstd_jll but I made this standalone since the seekable format is not technically part of zstd proper (though it would be more convenient).

### DIFF
--- a/Z/ZstdSeekable/build_tarballs.jl
+++ b/Z/ZstdSeekable/build_tarballs.jl
@@ -1,0 +1,36 @@
+using BinaryBuilder, Pkg
+
+# This is a subproject of zstd that lives in the zstd repository but, like the other
+# subprojects, is not built by default. It also has no Makefile or other build system
+# of its own, so here it's built manually. The version is set to the version of zstd
+# even though this subproject doesn't necessarily change between zstd versions, simply
+# for convenience.
+name = "ZstdSeekable"
+version = v"1.5.6"
+
+sources = [
+    ArchiveSource("https://github.com/facebook/zstd/releases/download/v$version/zstd-$version.tar.gz",
+                  "8c29e06cf42aacc1eafc4077ae2ec6c6fcb96a626157e0593d5e82a34fd403c1"),
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/zstd-*/contrib/seekable_format/
+${CC} -O3 -g zstdseek_compress.c zstdseek_decompress.c -shared -o libzstd_seekable.${dlext} -lzstd -I${includedir} -I../../lib/common/ -L${libdir}
+install -Dvm 644 zstd_seekable.h "${includedir}/zstd_seekable.h"  # might as well I guess?
+install -Dvm 755 libzstd_seekable.${dlext} "${libdir}/libzstd_seekable.${dlext}"
+install_license ../../LICENSE
+"""
+
+platforms = supported_platforms()
+
+products = [
+    LibraryProduct("libzstd_seekable", :libzstd_seekable),
+]
+
+dependencies = [
+    Dependency(PackageSpec(; name="Zstd_jll", uuid="3161d3a3-bdf6-5164-811a-617609db77b4");
+               compat="^$version"),
+]
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="1.6")


ararslan:
The build steps work as expected for me locally on macOS so I'm not sure why CI is complaining about undefined symbols. Perhaps I put the `-lzstd` in a place that macOS and Windows don't like?

Z/ZstdSeekable/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+    ArchiveSource("https://github.com/facebook/zstd/releases/download/v$version/zstd-$version.tar.gz",
+                  "8c29e06cf42aacc1eafc4077ae2ec6c6fcb96a626157e0593d5e82a34fd403c1"),
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/zstd-*/contrib/seekable_format/
+FLAGS=(-O3 -g -fPIC -shared -L${libdir} -I${includedir} -I. -I../../lib/ -I../../lib/common/ -lzstd)
+SOURCES=(zstdseek_compress.c zstdseek_decompress.c)
+LIB=libzstd_seekable.${dlext}
+${CC} ${FLAGS[@]} -o ${LIB} ${SOURCES[@]}
```
ararslan, line 21:
```suggestion
${CC} ${SOURCES[@]} -o ${LIB} ${FLAGS[@]}
```
Worth a try I guess

ararslan:
I rearranged some flags and now Windows is happy but still not macOS. 😕

giordano:
Error is:
```
[15:47:17]  ---> FLAGS=(-O3 -g -fPIC -shared -L${libdir} -I${includedir} -I. -I../../lib/ -I../../lib/common/ -lzstd)
[15:47:17]  ---> SOURCES=(zstdseek_compress.c zstdseek_decompress.c)
[15:47:17]  ---> LIB=libzstd_seekable.${dlext}
[15:47:17]  ---> ${CC} ${SOURCES[@]} -o ${LIB} ${FLAGS[@]}
[15:47:18] Undefined symbols for architecture arm64:
[15:47:18]   "_ZSTD_XXH64_digest", referenced from:
[15:47:18]       _ZSTD_seekable_endFrame in zstdseek_compress-be5564.o
[15:47:18]       _ZSTD_seekable_decompress in zstdseek_decompress-e366ad.o
[15:47:18]   "_ZSTD_XXH64_reset", referenced from:
[15:47:18]       _ZSTD_seekable_initCStream in zstdseek_compress-be5564.o
[15:47:18]       _ZSTD_seekable_endFrame in zstdseek_compress-be5564.o
[15:47:18]       _ZSTD_seekable_decompress in zstdseek_decompress-e366ad.o
[15:47:18]   "_ZSTD_XXH64_update", referenced from:
[15:47:18]       _ZSTD_seekable_compressStream in zstdseek_compress-be5564.o
[15:47:18]       _ZSTD_seekable_decompress in zstdseek_decompress-e366ad.o
[15:47:18] ld: symbol(s) not found for architecture arm64
[15:47:18] clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```
but in libzstd.dylib I see:
```console
% nm libzstd.dylib|grep -E 'ZSTD_XXH64_(digest|reset|update)'
0000000000008d70 t _ZSTD_XXH64_digest
0000000000008b50 t _ZSTD_XXH64_reset
0000000000008ba0 t _ZSTD_XXH64_update
```
I believe the lowercase `t` means these symbols are not visible?

giordano:
The upstream example [links to the static library](https://github.com/facebook/zstd/blob/0e2ceb2d5061f3a8357d124029ebaae16d915a3d/contrib/seekable_format/examples/Makefile#L13), so they don't care about the visibility of these symbols.

giordano:
But try to use `-DXXH_NAMESPACE=ZSTD_`, like in https://github.com/facebook/zstd/blob/0e2ceb2d5061f3a8357d124029ebaae16d915a3d/contrib/seekable_format/examples/Makefile#L16, I've seen that suggested in other places but it wasn't clear to me what it'd do.

ararslan:
Ha, when trying this locally on macOS, I was linking to libzstd from Homebrew:
```
$ nm /opt/homebrew/lib/libzstd.dylib | grep -E 'ZSTD_XXH64_(digest|reset|update)'
0000000000004580 T _ZSTD_XXH64_digest
0000000000004390 T _ZSTD_XXH64_reset
00000000000043d8 T _ZSTD_XXH64_update
```
Whereas the libzstd built with BinaryBuilder has:
```
$ nm ~/.julia/artifacts/3908ba61f3e144a7e7e28b9d676ca770f2f6a955/lib/libzstd.dylib | grep -E 'ZSTD_XXH64_(digest|reset|update)'
0000000000008d70 t _ZSTD_XXH64_digest
0000000000008b50 t _ZSTD_XXH64_reset
0000000000008ba0 t _ZSTD_XXH64_update
```
So those symbols are global in Homebrew's build and local in ours.

giordano:
Interesting.  Do they use Meson as build system for libzstd?

ararslan:
> But try to use `-DXXH_NAMESPACE=ZSTD_`

I had omitted that since it gets set by default in https://github.com/facebook/zstd/blob/0e2ceb2d5061f3a8357d124029ebaae16d915a3d/lib/common/xxhash.h#L18-L20 but it's worth a try.

ararslan:
> Do they use Meson as build system for libzstd?

No, they use CMake: https://github.com/Homebrew/homebrew-core/blob/master/Formula/z/zstd.rb#L37-L48

Z/ZstdSeekable/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+version = v"1.5.6"
+
+sources = [
+    ArchiveSource("https://github.com/facebook/zstd/releases/download/v$version/zstd-$version.tar.gz",
+                  "8c29e06cf42aacc1eafc4077ae2ec6c6fcb96a626157e0593d5e82a34fd403c1"),
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/zstd-*/contrib/seekable_format/
+FLAGS=(-O3 -g -fPIC -shared -L${libdir} -I${includedir} -I. -I../../lib/ -I../../lib/common/ -lzstd)
```
ararslan, line 18:
```suggestion
FLAGS=(-O3 -g -fPIC -shared -DXXH_NAMESPACE=ZSTD_ -L${libdir} -I${includedir} -I. -I../../lib/ -I../../lib/common/ -lzstd)
```

ararslan:
Presumably the symbol visibility is the same across platforms so why would only macOS complain about it? 🤔

giordano:
> Presumably the symbol visibility is the same across platforms

[citation needed] :stuck_out_tongue: 

> why would only macOS complain about it? 🤔

macOS linkers don't allow undefined symbols by default, Linux and FreeBSD ones do :upside_down_face:

ararslan:
Interestingly, it appears the the symbol visibility is set differently depending on the build system:
- The library Makefile sets `-fvisibility=hidden` unconditionally: https://github.com/facebook/zstd/blob/dev/lib/Makefile#L145
- Meson sets `gnu_symbol_visibility: 'hidden'` unconditionally: https://github.com/facebook/zstd/blob/dev/build/meson/lib/meson.build#L121
- CMake does not seem to set it explicitly but allows configuration via `ZSTDLIB_VISIBLE`, defaulting to unset (https://github.com/facebook/zstd/blob/dev/build/cmake/lib/CMakeLists.txt#L20), which then uses default visibility (https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L28)

giordano:
> Interestingly, it appears the the symbol visibility is set differently depending on the build system:

Not the first time we see ABI-incompatible binaries produced by different build systems of the same package (#8677, #8497).  It's a mess.

ararslan:
Welp. 😐

What do you propose we do here then? Amend the zstd builder to patch out the `-fvisibility=hidden` flag, or perhaps switch it to use CMake instead of Meson? Link libzstd statically here?

Also, thoughts on whether this should be part of the zstd builder rather than a separate builder? That would ensure they don't get out of sync in cases where that might matter, but I don't think the seekable format is meaningfully versioned at all so it's possible it could get updated independently of zstd.

giordano:
I'm a bit hesitant about switching to a different build system: as shown by the PRs linked above, whenever we do that we break something because life is horrible.  May first try to patch out zstd's meson build system (only the visibility bit) and see if that achieves what we want.

ararslan:
I tried removing [this line](https://github.com/facebook/zstd/blob/dev/build/meson/lib/meson.build#L121) locally and that did indeed change the visibility from local to global, which then allows this library to compile successfully. Is there any reason not to make that change in the zstd builder then? Relatedly, can `Dependency(...; compat=something)` specify a build number so that this JLL can ensure its zstd artifact has the appropriately exposed symbols? (That finickiness kind of makes me think that including the seekable library in the zstd builder might be better than standalone...)

I tried also building and running the test program for the seekable library and I'm not sure whether I did something wrong with compiling that program because I don't know what I'm doing or if there are some flags I missed when compiling the library. Here's what all I tried:
```
$ cc seekable_tests.c -o seekable_tests -L../ -L../../../build-zstd/lib/ -I.. -I../../../lib/ -I../../../lib/common/ -lzstd_seekable -lzstd
$ ./seekable_tests
dyld[20062]: Library not loaded: libzstd_seekable.dylib
  Referenced from: <8693FD34-D66F-370F-B561-3E1A32A3760D> /Users/alex/Projects/c/zstd/contrib/seekable_format/tests/seekable_tests
  Reason: tried: 'libzstd_seekable.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibzstd_seekable.dylib' (no such file), 'libzstd_seekable.dylib' (no such file), '/Users/alex/Projects/c/zstd/contrib/seekable_format/tests/libzstd_seekable.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/alex/Projects/c/zstd/contrib/seekable_format/tests/libzstd_seekable.dylib' (no such file), '/Users/alex/Projects/c/zstd/contrib/seekable_format/tests/libzstd_seekable.dylib' (no such file)
[1]    20062 abort      ./seekable_tests
$ cd ..
$ cc zstdseek_compress.c zstdseek_decompress.c -o libzstd_seekable.dylib -O3 -g -fPIC -shared -install_name "@rpath/libzstd_seekable.dylib" -L../../build-zstd/lib/ -I. -I../../lib/ -I../../lib/common/ -lzstd
$ cd -
~/Projects/c/zstd/contrib/seekable_format/tests
$ cc seekable_tests.c -o seekable_tests -L../ -L../../../build-zstd/lib/ -I.. -I../../../lib/ -I../../../lib/common/ -lzstd_seekable -lzstd
$ ./seekable_tests
dyld[20191]: Library not loaded: @rpath/libzstd_seekable.dylib
  Referenced from: <3652C391-6608-340A-B538-C17D2F0D714F> /Users/alex/Projects/c/zstd/contrib/seekable_format/tests/seekable_tests
  Reason: no LC_RPATH's found
[1]    20191 abort      ./seekable_tests
$ cc seekable_tests.c -o seekable_tests -L../ -L../../../build-zstd/lib/ -I.. -I../../../lib/ -I../../../lib/common/ -lzstd_seekable -lzstd -rpath ".." -rpath "../../../build-zstd/lib"
$ ./seekable_tests
Beginning zstd seekable format tests...
Test 1 - simple round trip: Success!
Test 2 - check that seekable decompress does not hang: Success!
Test 3 - check #2 that seekable decompress does not hang: Success!
Test 4 - check ZSTD magic in compressing empty string: Success!
Test 5 - multiple decompress calls: Success!
Finished tests
```

giordano:
> Is there any reason not to make that change in the zstd builder then?

Uhm, I guess the risk of such change would be to expose too many symbols. It'd be nice if upstream actually exposed in the shared library the symbols that may be needed by other libraries, this feels like a bug to me.

ararslan:
I think it's reasonable for libzstd to not reexport the xxHash symbols since its use of xxHash is arguably an implementation detail. It did occur to me though that we can use xxHash directly to provide that functionality, and doing so seems to fix the issue!

PR #8738, opened 2024-05-22
[DuckDB] Bump to v0.10.3

Mytherin:


giordano:
```
[14:07:05] In file included from /workspace/srcdir/duckdb/build/src/common/ub_duckdb_common.cpp:20:0:
[14:07:05] /workspace/srcdir/duckdb/src/common/local_file_system.cpp: In member function ‘virtual bool duckdb::LocalFileSystem::Trim(duckdb::FileHandle&, duckdb::idx_t, duckdb::idx_t)’:
[14:07:05] /workspace/srcdir/duckdb/src/common/local_file_system.cpp:489:26: error: ‘FALLOC_FL_PUNCH_HOLE’ was not declared in this scope
[14:07:05]   int res = fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset_bytes, length_bytes);
[14:07:05]                           ^~~~~~~~~~~~~~~~~~~~
[14:07:05] /workspace/srcdir/duckdb/src/common/local_file_system.cpp:489:49: error: ‘FALLOC_FL_KEEP_SIZE’ was not declared in this scope
[14:07:05]   int res = fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset_bytes, length_bytes);
[14:07:05]                                                  ^~~~~~~~~~~~~~~~~~~
```
Based on https://man7.org/linux/man-pages/man2/fallocate.2.html I take you require glibc 2.18 now?  We'd need to update glibc in our build recipe, doable but a bit annoying.

Mytherin:
It's not really required - we could also patch it and change the `#ifdef` where this is used:

```cpp
bool LocalFileSystem::Trim(FileHandle &handle, idx_t offset_bytes, idx_t length_bytes) {
#if defined(__linux__)
	int fd = handle.Cast<UnixFileHandle>().fd;
	int res = fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset_bytes, length_bytes);
	return res == 0;
#else
	return false;
#endif
}
```

Mytherin:
Yes, that is the current code. I'm suggesting we could patch the `#ifdef` to something else to make that work.

giordano:
> Yes, that is the current code.

Ah, sorry, I missed that, I thought that was the suggestion! :smiley: 

I'm also happy to adapt the build recipe to use a newer glibc (2.19 likely) for those platforms, up to you :slightly_smiling_face:

Mytherin:
I have a patch up here - https://github.com/duckdb/duckdb/pull/12193. Whichever is easier (applying the patch vs switching to the glibc version) is fine by me. Alternatively the next DuckDB release should have this patch included as well.

D/DuckDB/build_tarballs.jl
```diff
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "DuckDB"
-version = v"0.10.2"
+version = v"0.10.3"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/duckdb/duckdb.git", "1601d94f94a7e0d2eb805a94803eb1e3afbbe4ed"),
+    GitSource("https://github.com/duckdb/duckdb.git", "70fd6a8a2450c1e2a7d0547d4c0666a649dc378e"),
```
giordano, line 10: Ok, now that the PR has been merged I can guess we can do something like
```suggestion
    GitSource("https://github.com/duckdb/duckdb.git", "3f33d34debb665b37c5e4d479afd7f85a64a94f7"),
```
?

Mytherin: It should point to a nightly build - e.g. `1323e95677e571cfacd56e3dd1fe6d3220a76512` - otherwise there will be no extensions available

mzy2240:
@Mytherin do you mind to bump the compat in DuckDB.jl for DuckDB_jll to v0.10.3 as well? Thanks!

giordano:
As far as I can tell you can already get `DuckDB_jll@0.10.3` out of the box:
```
(jl_afUoWa) pkg> add DuckDB
[...]
(jl_afUoWa) pkg> st -m DuckDB DuckDB_jll
Status `/private/var/folders/v2/hmy3kzgj4tb3xsy8qkltxd0r0000gn/T/jl_afUoWa/Manifest.toml`
  [d2f5444f] DuckDB v0.10.2
  [2cbbab25] DuckDB_jll v0.10.3+0
```

mzy2240:
ok ic. just need to update the registry. thanks!

Mytherin:
I've pushed a new version of `DuckDB.jl` as well

PR #8741, opened 2024-05-23
 [ITK] Add build recipe for v5.3.0 : New Package

divital-coder:
Proposal for a build recipe for the ITK medical imaging toolkit framework binaries.
Specifics:
platform_built: x86_64 (Linux)
libc : glibc
gcc --version = 8

The build script does encompass everything needed to build and link the library.
Currently, the entirety of the Library shared objects were not exposed as Library products,
Only the Image Registration and ImageIO library products are exposed.

refs:
https://itk.org/Wiki/ITK/Configuring_and_Building

### DIFF
--- a/G/GDAL/build_tarballs.jl
+++ b/G/GDAL/build_tarballs.jl
@@ -50,10 +50,10 @@ fi
 mkdir build && cd build
 
 CMAKE_FLAGS=(-DCMAKE_INSTALL_PREFIX=${prefix}
-    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}
-    -DCMAKE_PREFIX_PATH=${prefix}
-    -DCMAKE_FIND_ROOT_PATH=${prefix}
     -DCMAKE_BUILD_TYPE=Release
+    -DCMAKE_FIND_ROOT_PATH=${prefix}
+    -DCMAKE_PREFIX_PATH=${prefix}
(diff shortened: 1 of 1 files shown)


I/ITK/build_tarballs.jl
```diff
@@ -0,0 +1,129 @@
...
+cmake --install build
+install_license ${WORKSPACE}/srcdir/InsightToolkit-5.3.0/LICENSE
+logout
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc")
+]
```
ararslan, line 67: Why only x86-64 glibc Linux? ITK appears to at least also support macOS (x86-64 and ARM) and Windows. It could be worth trying to build for all supported platforms.
```suggestion
platforms = supported_platforms()
```

divital-coder: Hiya Alex, thanks for the review, yes that is the intended plan, but first i needed to test some functionality of the wrapped coded after this, on linux.
If mulitple platform support is a must for a pkg to be published in the registry, i will certainly prioritize it first.

divital-coder: Further, right now i would like to get support for x86_64 platforms as soon as possible in the yggdrasil to support the wrappers around ITK in julia :),
i will ship with windows and macOS too.

PR #8776, opened 2024-05-26
[Spot_julia] upgrade using Spot version 2.12

MaximeBouton:


S/Spot_julia/build_tarballs.jl
```diff
@@ -19,7 +19,8 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/spot-2.9.7/
+export MACOSX_DEPLOYMENT_TARGET=10.15
```
giordano, line 22: Lying about the macOS version you're targeting is not quite enough to achieve anything.  You want to take inspiration from something like https://github.com/JuliaPackaging/Yggdrasil/blob/63d5b20f7dbedd7b7ac4de71b190dcfb4b8353c8/B/BOLT/build_tarballs.jl#L146-L149 https://github.com/JuliaPackaging/Yggdrasil/blob/63d5b20f7dbedd7b7ac4de71b190dcfb4b8353c8/B/BOLT/build_tarballs.jl#L14-L24

S/Spot_julia/build_tarballs.jl
```diff
@@ -3,12 +3,12 @@
 using BinaryBuilder, Pkg
 
 name = "Spot_julia"
-version = v"2.9.7"
+version = v"2.12"
 julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10"]
```
fingolfin, line 7: While you are it, why not update your package to drop this hard coded list, so it'll be compatible with 1.11.0 on the day it is released, and also with 1.12 dev versions?

To do so, it should be sufficient to drop this line.
```suggestion
```

MaximeBouton: Thank you, I don't remember why I had this.

S/Spot_julia/build_tarballs.jl
```diff
@@ -19,7 +20,19 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
...
+    # references therein.
+    pushd $WORKSPACE/srcdir/MacOSX10.*.sdk
+    rm -rf /opt/${target}/${target}/sys-root/System
+    cp -ra usr/* "/opt/${target}/${target}/sys-root/usr/."
+    cp -ra System "/opt/${target}/${target}/sys-root/."
+    export MACOSX_DEPLOYMENT_TARGET=10.14
+    popd
+fi
+
+cd $WORKSPACE/srcdir/spot-2.12/
```
giordano, line 35:
```suggestion
cd $WORKSPACE/srcdir/spot*
```

S/Spot_julia/build_tarballs.jl
```diff
@@ -19,7 +20,19 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
...
+    # references therein.
+    pushd $WORKSPACE/srcdir/MacOSX10.*.sdk
+    rm -rf /opt/${target}/${target}/sys-root/System
+    cp -ra usr/* "/opt/${target}/${target}/sys-root/usr/."
+    cp -ra System "/opt/${target}/${target}/sys-root/."
+    export MACOSX_DEPLOYMENT_TARGET=10.14
+    popd
+fi
+
+cd $WORKSPACE/srcdir/spot*
```
giordano, line 35:
```suggestion
cd $WORKSPACE/srcdir/spot-*
```

S/Spot_julia/build_tarballs.jl
```diff
@@ -19,7 +20,19 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
...
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} --disable-python
 make -j${nproc}
 make install
```
giordano, line 38:
```suggestion
make install
install_license COPYING
```

MaximeBouton: I think I had copied the GNU license from the original src in the wrapper I wrote: https://github.com/MaximeBouton/spot_julia/blob/main/LICENSE.md 

Do you know if I should remove that one?

PR #8788, opened 2024-05-28
update hidapi to use the hidraw backend over libusb

laborg:
The hidraw backend is better suited for linux and gets rid of the libusb dependency.

### DIFF
--- a/H/hidapi/build_tarballs.jl
+++ b/H/hidapi/build_tarballs.jl
@@ -19,10 +19,13 @@ mkdir build
 cd build
 
 if [[ ${target} == *-linux-* ]]; then
-    # We need this only to trick `configure` into thinking that udev is available,
-    # but we aren't going to actually use it
     apk add eudev-dev
-    linuxflags="-DHIDAPI_WITH_HIDRAW=false"
+    linuxflags="-DHIDAPI_WITH_LIBUSB=false"
+    export CPATH=/usr/include
+fi
+
+if [[ ${target} == *-freebsd ]]; then
+    apk add libiconv
 fi
 
 cmake  \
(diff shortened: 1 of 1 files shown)


H/hidapi/build_tarballs.jl
```diff
@@ -41,11 +44,12 @@ install_license ../hidapi/LICENSE*.txt
...
 
 products = [
-    LibraryProduct(["libhidapi", "libhidapi-libusb"], :hidapi)
+    LibraryProduct(["libhidapi", "libhidapi-hidraw","libhidapi-libusb"], :hidapi)
 ]
 
 dependencies = [
-    Dependency("libusb_jll", platforms=filter(p -> Sys.islinux(p) || Sys.isfreebsd(p), platforms))
+    Dependency("libusb_jll", platforms=filter(p -> Sys.isfreebsd(p), platforms))
+    BuildDependency("libusb_jll", platforms=filter(p -> Sys.islinux(p), platforms))
```
giordano, line 52: Why the same dependency twice?

H/hidapi/build_tarballs.jl
```diff
@@ -19,10 +19,13 @@ mkdir build
 cd build
 
 if [[ ${target} == *-linux-* ]]; then
-    # We need this only to trick `configure` into thinking that udev is available,
-    # but we aren't going to actually use it
     apk add eudev-dev
-    linuxflags="-DHIDAPI_WITH_HIDRAW=false"
+    linuxflags="-DHIDAPI_WITH_LIBUSB=false"
+    export CPATH=/usr/include
```
giordano, line 24: This looks wrong.

H/hidapi/build_tarballs.jl
```diff
@@ -19,10 +19,13 @@ mkdir build
...
+fi
+
+if [[ ${target} == *-freebsd ]]; then
+    apk add libiconv
```
giordano, line 28: What do you need of libiconv when targeting FreeBSD? Why not adding iconv_jll?

H/hidapi/build_tarballs.jl
laborg: The idea was to use libusb backend for freebsd, so that a BuildDependency wouldn't suffice.

H/hidapi/build_tarballs.jl
laborg: Without the CPATH the `libudev.h` in `/usr/include` isn't found. But this is probably wrong because host build dependency (`libudev` from the package eudev-dev) would also need to be available for 32bit. Nowadays libudev is provided by systemd but I didn't find a jll for systemd...

H/hidapi/build_tarballs.jl
laborg: Thats an oversight, didn't know that there is an iconv_jll.

laborg:
Given the test failures for 32 bit I think the hidraw backend might not be possible to build.

laborg:
Ah, wait, I missed that we have systemd: https://github.com/JuliaPackaging/Yggdrasil/blob/28d07bbfea9fa22a3f2dee41e565a3c692a3674d/S/systemd/build_tarballs.jl

PR #8792, opened 2024-05-28
[XRootD_cxxwrap] New binary package with the wrapper library to XRootD  

peremato:
The XRootD_cxxwrapper builds the CxxWrap wrapper library to the XRootD package.


X/XRootD_cxxwrap/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libXRootDWrap", :libXRootDWrap),
+    FileProduct("XRootD-export.jl", :XRootD_exports),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency("libjulia_jll"),
+    Dependency("libcxxwrap_julia_jll"; compat="0.11.2"),
```
giordano, line 53: You sure you don't need v0.12?  That's the latest now, but you also need to match the corresponding version of the Julia package CxxWrap.

peremato: yes, I am sure. The tool I use the generate the wrapper code (WrapIt) does not support well the latest version of CxxWrap. They are working on it.

giordano: Ok, just double checking 🙂

PR #8793, opened 2024-05-28
[CppInterOp] Build version v0.1.1

Gnimuc:
switch to the original git repo

### DIFF
--- a/L/libCppInterOp/build_tarballs.jl
+++ b/L/libCppInterOp/build_tarballs.jl
@@ -8,21 +8,24 @@ include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
 include(joinpath(YGGDRASIL_DIR, "platforms", "llvm.jl"))
 
 name = "libCppInterOp"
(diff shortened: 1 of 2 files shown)


L/libCppInterOp/build_tarballs.jl
```diff
@@ -8,21 +8,24 @@ include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
 include(joinpath(YGGDRASIL_DIR, "platforms", "llvm.jl"))
 
 name = "libCppInterOp"
-version = v"0.1.0"
+version = v"0.1.1"
```
giordano, line 11: Now that I notice this: any specific reason for not using upstream version number [`v1.3.0`](https://github.com/compiler-research/CppInterOp/releases/tag/v1.3.0)?

Gnimuc: No. But I'm still experimenting it. Will switch to the official version after everything works.

Issue #8779, opened 2024-05-26
Build cminpack


giordano:
https://github.com/devernay/cminpack.  Ref: https://github.com/sglyon/MINPACK.jl/pull/16#issuecomment-2118540308

imciner2:
Actually, looking at the repo, maybe @sglyon can say what the settings for building MINPACK should be?

sglyon:
It’s has been a while (7 years) , but my compilation notes are here: https://github.com/sglyon/MINPACK.jl/blob/master/notes.md

PR #8794, opened 2024-05-28
cminpack: Add recipe for version 1.3.8

imciner2:
Recipe for cminpack. This builds both the double and single precision libraries, and works on all Yggdrasil-supported platforms. I have backported the relevant patches on the upstream master branch to be on top of the 1.3.8 release, but I ended up disabling BLAS for now (once I get proper CMake support for libblastrampoline it will be easier), but this should match the config the library is currently using.

CC @sglyon

Fixes: https://github.com/JuliaPackaging/Yggdrasil/issues/8779

### DIFF
--- a/C/cminpack/build_tarballs.jl
+++ b/C/cminpack/build_tarballs.jl
@@ -1,0 +1,81 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "cminpack"
+version = v"1.3.8"
+
+# Collection of sources required to complete build
+sources = [
+    # We get 1.3.8, but it needs some patches
+    GitSource("https://github.com/devernay/cminpack.git", "cb7c3f6433ccea7eef58ff57b3a9a4c2563eb375"),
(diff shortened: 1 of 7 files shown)


giordano:
Looks good to me, if that's good for @sglyon.

sglyon:
If it works let’s do it!

horasio:
Hi. I've an optimal control project that uses MINPACK that uses cminlack_jll.

It works well with cminpack v1.3.8. But all broken with v1.3.11 or v1.3.12.
My code is long and I have no MRE yet. Sorry. Here is the symptom :

I call MINPACK::fsolve() without Jacobian, so the jacobian is estimated with N+1 evaluations of the function.
All that is well.
Then a move is computed by minpack to move the evaluation point, based on the Jacobian. Here is the crazy behaviour : the new point has crazy coordinates of the order of 10^100 ...
I obtained this information simply by println(...) in the evaluation function.
The code of MINPACK itself by Deverney does not seem to have changed from 1.3.8 to 1.3.11 so I belive the bug is in the auto-generation of cminpack_jll.

This is probably not the right place to deal with this issue... where should I put the issue ? Yggdrasil/issues ?

PR #8795, opened 2024-05-28
[ITK]-v5.3.0 : New Package

divital-coder:
Build Recipe for Insight Toolkit (ITK)
Version : 5.3.0


mkitti:
This definitely looks promising. Are there any other `ITK_USE_SYSTEM_*` options?

divital-coder:
Nope I included all of them as per the dependencies for itk, thanks for the feedback ur awesome.

divital-coder:
Hiya @giordano I made sure i followed the contributing guidelines and everything, is there going to be a result of this PR?

divital-coder:
@giordano Perhaps you can provide some insight, as to if there is going to be a potential result of this.

I/ITK/build_tarballs.jl
```diff
@@ -0,0 +1,144 @@
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "ITK"
+version = v"5.3.0"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/InsightSoftwareConsortium/ITK/releases/download/v$(version)/InsightToolkit-$(version).tar.gz", "57a4471133dc8f76bde3d6eb45285c440bd40d113428884a1487472b7b71e383")
```
asinghvi17, line 9: If this is a git-repo and what's being downloaded is the source code, then this should be a [`GitSource`](https://docs.binarybuilder.org/dev/reference/#BinaryBuilderBase.GitSource)

divital-coder: Its an archive from a Github repository of ITK, i originally was following the pattern from other tarballs, seems like the latter is more appropriate.

I/ITK/build_tarballs.jl
```diff
@@ -0,0 +1,144 @@
...
+install_license ${WORKSPACE}/srcdir/InsightToolkit-5.3.0/LICENSE
+exit
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+
+    Platform("x86_64", "windows"; ),
+    # Platform("x86_64", "macOS"; ),
```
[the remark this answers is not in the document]
divital-coder: Apparently, macOS have some linking issues with the library, I tried building for aarch and arm, and couldn't succeed for a while. So my mentor suggested to appeal a PR with x86_64 platforms, since the resulting wrapper is supposed to be tested on those target systems for a while.

I/ITK/build_tarballs.jl
divital-coder: Since its a tar file, the docs refer to wrap it in an ArchiveSource function.

asinghvi17: Gotcha.  I'd still recommend a GitSource for this - just because it's a lot easier to update!

imciner2:
The logs are showing that a lot of your CMake variables are not actually being used by your script:

```
  Manually-specified variables were not used by the project:
    DOUBLE_CONVERSION_CORRECT_DOUBLE_OPERATIONS__TRYRUN_OUTPUT
    H5_DISABLE_SOME_LDOUBLE_CONV_RUN
    H5_DISABLE_SOME_LDOUBLE_CONV_RUN__TRYRUN_OUTPUT
    H5_LDOUBLE_TO_LLONG_ACCURATE_RUN
    H5_LDOUBLE_TO_LLONG_ACCURATE_RUN__TRYRUN_OUTPUT
    H5_LDOUBLE_TO_LONG_SPECIAL_RUN
    H5_LDOUBLE_TO_LONG_SPECIAL_RUN__TRYRUN_OUTPUT
    H5_LLONG_TO_LDOUBLE_CORRECT_RUN
    H5_LLONG_TO_LDOUBLE_CORRECT_RUN__TRYRUN_OUTPUT
    H5_LONG_TO_LDOUBLE_SPECIAL_RUN
    H5_LONG_TO_LDOUBLE_SPECIAL_RUN__TRYRUN_OUTPUT
    H5_NO_ALIGNMENT_RESTRICTIONS_RUN
    H5_NO_ALIGNMENT_RESTRICTIONS_RUN__TRYRUN_OUTPUT
    H5_PRINTF_LL_TEST_RUN
    H5_PRINTF_LL_TEST_RUN__TRYRUN_OUTPUT
    HAVE_CLOCK_GETTIME_RUN
    ITK_USE_SYSTEM_TBB
    RUN_RESULT_VAR
    RUN_RESULT_VAR__TRYRUN_OUTPUT
    TEST_LFS_WORKS_RUN
    VCL_HAS_LFS__TRYRUN_OUTPUT
    VXL_HAS_SSE2_HARDWARE_SUPPORT__TRYRUN_OUTPUT
    VXL_SSE2_HARDWARE_SUPPORT_POSSIBLE
    VXL_SSE2_HARDWARE_SUPPORT_POSSIBLE__TRYRUN_OUTPUT
    _libcxx_run_result
    _libcxx_run_result__TRYRUN_OUTPUT
```

Please clean up the script a bit to not have a lot of extra variables.

I/ITK/build_tarballs.jl
```diff
@@ -0,0 +1,119 @@
...
+-DQNANHIBIT_VALUE__TRYRUN_OUTPUT:STRING=0 \
+-DVXL_HAS_SSE2_HARDWARE_SUPPORT:STRING=1 \
+-DVCL_HAS_LFS:STRING=1 \
+-DDOUBLE_CONVERSION_CORRECT_DOUBLE_OPERATIONS:STRING=1 \
+-DHAVE_CLOCK_GETTIME_RUN:STRING=0 \
+InsightToolkit-5.3.0/
+cmake --build build --parallel ${nproc}
+cmake --install build
+install_license ${WORKSPACE}/srcdir/InsightToolkit-5.3.0/LICENSE
+exit
```
imciner2, line 38:
```suggestion
cd $WORKSPACE/srcdir/InsightToolkit*

mkdir build/

cmake -B build -S . \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DBUILD_SHARED_LIBS:BOOL=ON \
    -DITK_USE_SYSTEM_EXPAT:BOOL=ON \
    -DITK_USE_SYSTEM_FFTW:BOOL=ON \
    -DITK_USE_SYSTEM_HDF5:BOOL=ON \
    -DITK_USE_SYSTEM_JPEG:BOOL=ON \
    -DITK_USE_SYSTEM_TIFF:BOOL=ON \
    -DITK_USE_SYSTEM_PNG:BOOL=ON \
    -DITK_USE_SYSTEM_EIGEN:BOOL=ON \
    -DITK_USE_SYSTEM_ZLIB:BOOL=ON \
    -DQNANHIBIT_VALUE:BOOL=0 \
    -DQNANHIBIT_VALUE__TRYRUN_OUTPUT:STRING=0 \
    -DVXL_HAS_SSE2_HARDWARE_SUPPORT:STRING=1 \
    -DVCL_HAS_LFS:STRING=1 \
    -DDOUBLE_CONVERSION_CORRECT_DOUBLE_OPERATIONS:STRING=1 \
    -DHAVE_CLOCK_GETTIME_RUN:STRING=0

cmake --build build --parallel ${nproc}
cmake --install build

install_license ${WORKSPACE}/srcdir/InsightToolkit-5.3.0/LICENSE
```

(The directory command will also need to be updated after the change to GitSource)

I/ITK/build_tarballs.jl
imciner2: Yes, making this a GitSource is better, because the hashes of these tarballs is actually not considered a stable property, and we have seen them accidentally get changed in the past: https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/

Using the GitSource helps precent that because the commit hash we build from should be stable.

imciner2 (changes requested):

divital-coder:
i686 platforms have an open Issue under ITK 
macOS and FeeBSD platforms also are facing issues due to a abrupt Cmake condition checkes in the ITK src.
Rest all platforms are built for.

divital-coder:
hi, i was wondering what are the next steps, as to get the ITK binaries in the julia binary wrappers organization.

fingolfin:
@imciner2 I think your comments were addressed, can you confirm?

I/ITK/build_tarballs.jl
```diff
@@ -0,0 +1,123 @@
...
+    Platform("x86_64","windows";),
+    Platform("aarch64","linux"; libc="glibc"),
+    Platform("x86_64", "linux"; libc="glibc"),
+    Platform("armv6l","linux"; call_abi="eabihf",libc="glibc"),
+    Platform("armv7l","linux"; call_abi="eabihf",libc="glibc"),
+    Platform("powerpc64le","linux"; libc="glibc"),
+    Platform("aarch64","linux"; libc="musl"),
+    Platform("armv6l", "linux"; call_abi="eabihf", libc="musl"),
+    Platform("armv7l", "linux"; call_abi="eabihf", libc="musl")
+]
```
fingolfin, line 52: If it works, it works, but in general I would recommend to start with the *full* list of platforms, then remove any that don't work, ideally with a comment indicating what the problem is.

This way, when new platforms are added in the future, they automatically get picked up. And the comments ensure that anyone who wonders "hey, why is macos not supported" has a place to start off.

So here this could be something like this:
```suggestion
platforms = supported_platforms()
filter!(!Sys.isapple, platforms)  # macOS currently not supported because ...
```
.... which actually would leave in also 32bit Windows, I think; but if that really is not supported, it could also be filtered out.

It also would leave in FreeBSD. If things don't work there, just add something like
```
filter!(!Sys.isfreebsd, platforms)  # FreeBSD currently not supported because ...
```

I/ITK/build_tarballs.jl
```diff
@@ -0,0 +1,125 @@
...
+
+
+function is_problematic_platform(platform)
+    # Extract relevant information from the platform object
+    platform_string = string(platform)
+    # Check conditions to exclude the platform
+    return contains(platform_string, "i686") || #sse2 disabled errors in ITK with open issues on github
+           contains(platform_string, "macOS") || #CMAKE errors for _libcxx_run_result in cross compilation
+           contains(platform_string, "FreeBSD") || #CMAKE error for _libcxx_run_result in cross compilation
+           platform_string == "Linux x86_64 {libc=musl}" #CMAKE error for _libcxx_run_result result in cross compilation
```
fingolfin, line 49: No, please don't do this by string comparison, that's too fragile. 

What was wrong with the code I suggested? It works great in other recipes.

The platform objects have methods to check cleanly for all these properties, like the OS, the bits (33 vs 64), the arch etc

I/ITK/build_tarballs.jl
```diff
@@ -0,0 +1,125 @@
...
+end
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+platforms = filter(platforms) do p
+    return !is_problematic_platform(p)
+end# macOS currently not supported because ...
```
fingolfin, line 56: that comment makes no sense

divital-coder:
@fingolfin I had multiple platforms, so i included a function, will take your review into account and take immediate action, thanks for your feedback.

fingolfin:
> Also, i am leaving comments on reviews, but for some reason they all are marked pending. Forgot to remove the unsensible comment .

That sounds as if you accidentally used the "Review" feature to write your comments as part of a review. Try going to https://github.com/JuliaPackaging/Yggdrasil/pull/8795/files and click on the big green "Review changes" button to open a popover in which you can press "submit review" to release all your pending comments

I/ITK/build_tarballs.jl
```diff
@@ -0,0 +1,116 @@
...
+    -DDOUBLE_CONVERSION_CORRECT_DOUBLE_OPERATIONS:STRING=1 \
+    -DHAVE_CLOCK_GETTIME_RUN:STRING=0
+cmake --build build --parallel ${nproc}
+cmake --install build
+install_license ${WORKSPACE}/srcdir/ITK/LICENSE
+"""
+
+
+
+#sse2 disabled errors in ITK with open issues on github for i686 platforms
```
fingolfin, line 42: What does "sse2 disabled errors in ITK" mean? "The fact that SSE2 is disabled leads to error in ITK" maybe (why is SSE2 disabled?)

If there are relevant open issues, how about including a link / links ?

I/ITK/build_tarballs.jl
```diff
@@ -0,0 +1,116 @@
...
+#CMAKE errors for _libcxx_run_result in cross compilation for macOS, freebsd and x86_64 linux musl
+#
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = platforms = BinaryBuilder.supported_platforms(exclude=x -> (Sys.isapple(x) || Sys.isfreebsd(x) || arch(x) == "i686" || (arch(x) == "x86_64" && libc(x) == "musl")))
+platforms = expand_cxxstring_abis(platforms)
```
fingolfin, line 48: Based on my experience, I really would suggest structuring it like this, i.e.: group the comment with the resulting action; and keep each removal on its own line so it is easier to experiment with it in the future
```suggestion
# These are the platforms we will build for by default, unless further
# platforms are passed in on the command line
platforms = supported_platforms()

#sse2 disabled errors in ITK with open issues on github for i686 platforms
filter!(p -> arch(p) == "i686", platforms)

#CMAKE errors for _libcxx_run_result in cross compilation for macOS, freebsd and x86_64 linux musl
filter!(!Sys.isapple, platforms)
filter!(!Sys.isfreebsd, platforms)
filter!(p -> !(arch(x) == "x86_64" && libc(x) == "musl"), platforms)

platforms = expand_cxxstring_abis(platforms)
```

I/ITK/build_tarballs.jl
```diff
@@ -0,0 +1,116 @@
...
+    -DHAVE_CLOCK_GETTIME_RUN:STRING=0
+cmake --build build --parallel ${nproc}
+cmake --install build
+install_license ${WORKSPACE}/srcdir/ITK/LICENSE
+"""
+
+
+
+#sse2 disabled errors in ITK with open issues on github for i686 platforms
+#CMAKE errors for _libcxx_run_result in cross compilation for macOS, freebsd and x86_64 linux musl
```
fingolfin, line 43: But what is the error, exactly? I googled for `_libcxx_run_result` and `libcxx_run_result` and found no hits. I also had a look at the ITK repository and again no hits. So I wonder where this comes from and the error might be?

You could just submit your PR here with one of the failing targets enabled, so we can see the error in the logs ourselves.

divital-coder:
@fingolfin 
For x86_64 linux with musl platform here is the error (its obscure, doesnt give any red markers for errors, just stops the build but i was able to catch this)
swappy-20240610-132038

divital-coder:
This is the issue with freebsd 
swappy-20240610-135941
https://github.com/InsightSoftwareConsortium/ITK/issues/4565

PR #8796, opened 2024-05-28
[Geant4_julia] New version of wrapper

peremato:


barche:
This is failing on Julia 1.6 because the way in which `ptls` is accessed changed. You can get a cross-version access to it using:

```c++
#if (JULIA_VERSION_MAJOR * 100 + JULIA_VERSION_MINOR) >= 107
  jl_ptls_t ptls = jl_current_task->ptls;
#else
  jl_ptls_t ptls = jl_get_ptls_states();
#endif
```

peremato:
Thanks very much.

PR #8815, opened 2024-05-30
[aws_c_cal] Update to version 0.6.15

quinnj2:
This PR updates aws_c_cal to version 0.6.15. cc: @quinnj @Octogonapus


quinnj:
This one is a bit weird because it seems they have 2 successive releases/tags based on the same commit (took me a while to figure out what was going on, because I thought surely my new automation was doing something wrong). I think we still want to do a new build here though, right? (releases reference: https://github.com/awslabs/aws-c-cal/releases)

PR #8817, opened 2024-05-30
cminpack: Update to 1.3.9 and enable BLAS

imciner2:
Update to the latest release (1.3.9) and remove all our patches. Also enable BLAS support through libblastrampoline using the new CMake support for it that was just added.


avik-pal:
Is there some additional configuration needed for end users? NonlinearSolve CI seems to break https://github.com/SciML/NonlinearSolve.jl/actions/runs/9326385714/job/25674844612?pr=441#step:6:32485

giordano:
Yank the new version, looks like it doesn't work correctly

giordano:
@imciner2 I presume either this library should be adapted to do ilp64 properly, or we add openblas32 as a default lp64 backend?

avik-pal:
yank the jll or the MINPACK.jl release?

giordano:
The latest version of the jll. Why would you yank minpack (and what version)?

avik-pal:
MINPACK 1.3 starts using the jll, before that it had a custom build step https://github.com/sglyon/MINPACK.jl/blob/v1.2.0/deps/build.jl.

imciner2:
cminpack_jll version 1.3.8 didn't use BLAS, that was added in 1.3.9. So we could yank the 1.3.9 JLL release, or alternately, maybe we could just make a new build to flip the flag back to off?

PR #7941, opened 2024-01-12
Add comment to OpenCV to force rebuild

cdsousa:
OpenCV.jl is not working in Julia 1.10 (https://github.com/JuliaImages/OpenCV.jl/issues/36).
It seems to have something to do with OpenCV_jll, something like it not being compiled against the newest CxxWrap and libjulia.
Trying a rebuild.

PR #8820, opened 2024-05-31
Updated OpenCV compatibility for libcxxwrap_julia_jll to version 0.11

Moeasy64:
Good morning,

As we are working with our Professor @marcoxa, I am opening this pull request in order to fix the incompatibility problem for Julia 1.10 of OpenCV package. I'm also going to open a PR on [OpenCV.jl](https://juliahub.com/ui/Packages/General/OpenCV) to update the `Project.toml` with the CxxWrap  version 0.14. I hope this will fix the incompatibility.


O/OpenCV/build_tarballs.jl
```diff
@@ -123,8 +123,8 @@ dependencies = [
     Dependency(PackageSpec(name="Qt5Base_jll", uuid="ea2cea3b-5b76-57ae-a6ef-0a8af62496e1"))
     Dependency(PackageSpec(name="Libglvnd_jll", uuid="7e76a0d4-f3c7-5321-8279-8d96eeed0f29"))
     BuildDependency(PackageSpec(name="libjulia_jll"))
-    Dependency(PackageSpec(name="libcxxwrap_julia_jll"))
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"); compat="0.11")
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"9")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.10", preferred_gcc_version = v"9")
```
giordano, line 130: Any reason why this was changed?

O/OpenCV/build_tarballs.jl
```diff
@@ -123,8 +123,8 @@ dependencies = [
     Dependency(PackageSpec(name="Qt5Base_jll", uuid="ea2cea3b-5b76-57ae-a6ef-0a8af62496e1"))
     Dependency(PackageSpec(name="Libglvnd_jll", uuid="7e76a0d4-f3c7-5321-8279-8d96eeed0f29"))
     BuildDependency(PackageSpec(name="libjulia_jll"))
-    Dependency(PackageSpec(name="libcxxwrap_julia_jll"))
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"); compat="0.11")
```
giordano, line 126: Why not libcxxwrap_julia_jll@0.12?

giordano (changes requested):
You also need to change the version number of the package itself:
https://github.com/JuliaPackaging/Yggdrasil/blob/cc90a42a42a00e8a353a5fba8831aacd720ab5fd/O/OpenCV/build_tarballs.jl#L11
which basically means upgrade to a newer version

O/OpenCV/build_tarballs.jl
fingolfin: Just to say, we are also still at 0.11 for some of our packages because we'd need to coordinate a bunch of packages to simultaneously make the jump to the newer version. Not sure if that's the case here, but it is a problem (and of course gets worse for people who'd like to combine packages, some of which made the jump already, some didn't).

marcoxa:
Hi

my students are doing a great job tracking down these issues.

Let me say that all of this should be ... easier.

O/OpenCV/build_tarballs.jl
marcoxa: @fingolfin apologies for the obtuseness, but what are the 0.11 and 0.12 numbers?  Do they refer to Julia, Yggdrasil or BinaryBuilder?

marcoxa: > Why not libcxxwrap_julia_jll@0.12?

Good point.

O/OpenCV/build_tarballs.jl
marcoxa: Because otherwise you canno use OpenCV with Julia 1.10.x (at least).

marcoxa:
> You also need to change the version number of the package itself:
> 
> https://github.com/JuliaPackaging/Yggdrasil/blob/cc90a42a42a00e8a353a5fba8831aacd720ab5fd/O/OpenCV/build_tarballs.jl#L11
> 
> which basically means upgrade to a newer version

We are not the maintainers of OpenCV; it should be up to them to decide on a new version number.

O/OpenCV/build_tarballs.jl
giordano: I'm confused. Can you elaborate on why?

giordano:
> it should be up to them to decide on a new version number.

It looks like there have been already four releases after 4.6.0: https://opencv.org/releases/

O/OpenCV/build_tarballs.jl
marcoxa: Because it generates several errors?

@Moeasy64 (and friends) has the details.

This appears to be a well known issue with OpenCV and Julia 1.10.x (worked fine in 1.9.x) since January.  Alas, it has not been resolved yet.

Cfr. https://github.com/JuliaPackaging/Yggdrasil/pull/7941

marcoxa:
> > it should be up to them to decide on a new version number.
> 
> It looks like there have been already four releases after 4.6.0: https://opencv.org/releases/

Good point.  But this is the OpenCV release(s).

The bigger issue here is how to bring the Julia binding of OpenCV up to speed with Julia latest incarnations.

O/OpenCV/build_tarballs.jl
giordano: > Cfr. #7941

That PR didn't change the compat bound. I still don't understand why that should change.

O/OpenCV/build_tarballs.jl
barche: The 0.11 and 0.12 are the `libcxxwrap_julia` version, there is a breaking change between 0.11 and 0.12 and you need at least 0.11 to be compatible with Julia 1.10. I would recommend to jump to 0.12 if you can, but it may require some changes to the OpenCV julia wrapper code.

O/OpenCV/build_tarballs.jl
marcoxa: > To clarify, I think this was a misunderstanding, the Julia compat bound oof 1.6 means it will work with Julia 1.6 and above, so leaving this at 1.6 is fine.

I think instead that this is the issue.  The current libcxxwrap library (which is required by OpenCV.jl) does _not_ work in Julia 1.10.

Any suggestions about how to make things right?

giordano: > I think instead that this is the issue. The current libcxxwrap library (which is required by OpenCV.jl) does not work in Julia 1.10.

Changing this line is totally unrelated to that problem and insisting on that will bring us nowhere. I still didn't get a real motivation for why this line should be changed.

barche:
It looks like the actual Julia wrappers are in opencv_contrib, which like OpenCV itself has a tag for version 4.9.0. Unfortunately the code in the Julia module is 3 years old, so it may need some adaptation.

The reason the OpenCV version needs to be incremented is because the compat requirements of this JLL will change by moving to a newer libcxxwrap_julia. The most elegant way to do that is by moving to a more recent OpenCV upstream version. There are workarounds by modifying the version number only here, but this is messy and best avoided.

giordano:
My reason for me to ask to change the version number was [the change of the compat bound for julia](https://github.com/JuliaPackaging/Yggdrasil/pull/8820#discussion_r1622219958), which appears unjustified.  Without that, we may not _need_ to change the version number of the package, but someone still needs to fix the wrapper somehow.

marcoxa:
> The main reason for me to ask to change the version number was [the change of the compat bound for julia](https://github.com/JuliaPackaging/Yggdrasil/pull/8820#discussion_r1622219958), which appears unjustified. Without that, we may not _need_ to change the version number of the package, but someone still needs to fix the wrapper somehow.

That "someone still needs to fix the wrapper somehow" seems reasonable.

Now: **where** should this wrapper be "fixed"?  In the OpenCV main repository?  Meaning a PR to the OpenCV community?  And what about the `libcxxwrap `library?  How to coordinate these changes?

O/OpenCV/build_tarballs.jl
fingolfin, line 130:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"9")
```
I also don't understand this change at all, it seems unnecessary.

barche:
> Now: **where** should this wrapper be "fixed"? In the OpenCV main repository? Meaning a PR to the OpenCV community? And what about the `libcxxwrap `library? How to coordinate these changes?

The wrapper lives here: https://github.com/opencv/opencv_contrib/tree/4.x/modules/julia

Unfortunately, it is a little more complicated than I anticipated, since the wrapper is actually auto-generated using a Python script, so modifying that would probably be the best way forward. Looking at the git commits in that directory it seems the wrappers were added as part of a GSOC project.

marcoxa:
Ahhhhh.  A Python script.  People never learn, do they?

Marco Antoniotti
Somewhere over the Rainbow

barche:
Getting this to work for Julia 1.10 may actually be easier than I initially thought, all that is needed is to modify `bundled/patches/opencv-julia.patch` to replace 

```julia
@wrapmodule(OpenCV_jll.libopencv_julia_path, :cv_wrap)
```

with 

```julia
@wrapmodule(OpenCV_jll.get_libopencv_julia_path, :cv_wrap)
```

The Julia 1.11 and 1.12 errors can probably also be fixed by adding a patch to `opencv_contrib/modules/julia/gen/cpp_files/jlcxx/array.hpp`, backporting some of the changes that were made in libcxxwrap_julia.

likanzhan:
@barche Is there an update to this problem? And where this code should be changed? Thanks.

giordano:
> And where this code should be changed? Thanks.

https://github.com/JuliaPackaging/Yggdrasil/pull/8820#issuecomment-2143045557

O/OpenCV/build_tarballs.jl
ViralBShah: Agree. This should just be left at 1.6.

O/OpenCV/build_tarballs.jl
ViralBShah: What's the latest on this now? Should this go to 0.13?

ViralBShah:
@marcoxa The easiest thing for us to do is to apply the patches we need for `opencv_contrib` in Yggrdasil and build opencv 4.10 first. We can then submit them upstream to `opencv_contrib` so that we don't need to keep carrying them.

It seems that once we can get this over the hump, future updates should hopefully be straightforward since a lot of the machinery on the Julia side has stabilized a fair bit (fingers crossed).

O/OpenCV/build_tarballs.jl
barche: Yes, 0.13.2 is the one used by the latest released CxxWrap.

barche:
I added some changes that will hopefully make this build against the latest libcxxwrap-julia

ViralBShah (approved):

ViralBShah:
@barche That is an insane patch - thank you! I hope going forward it won't need this much work. Should the opencv-contrib patches be upstreamed?

barche:
Before upstreaming the opencv_contrib changes I'd like to see what can be changed in libcxxwrap-julia, since some of the changes (e.g. the `container_has_less_than_operator` specializations) are only needed because of a poor design in libcxxwrap-julia, that causes also problems elsewhere (such as https://github.com/JuliaInterop/CxxWrap.jl/issues/455)

Also, before merging this PR the binaries (downloadable from buildkite) should be tested with OpenCV.jl

barche:
The binaries are not working because the OpenCV build system adds the flags `-fvisibility=hidden -fvisibility-inlines-hidden` to the compilation. I don't immediately see where, so clues welcome :)

barche:
Provided all builds still pass the binaries should work after the build is complete. At least the OpenCV tests start here in Julia 1.11 locally, with no output shown yet at the time of writing this.

So if all builds pass this is good to merge for me.

barche:
Going over the PR one more time before hitting the big green button I see this still uses Qt5Base. Any idea if this is needed? It will conflict with the Qt6Base requirement of GR (and thus Plots).

barche:
I went ahead and tried directly by just changing to Qt6Base, let's hope this recent OpenCV just picks it up. It's best to do this change now, since updating the Qt dep would require a JLL version change otherwise.

PR #8824, opened 2024-06-01
Wizard recipe: WIGXJPF-v1.13.0

wes648:

* Package name: WIGXJPF
* Version: v1.13.0

### DIFF
--- a/W/WIGXJPF/build_tarballs.jl
+++ b/W/WIGXJPF/build_tarballs.jl
@@ -1,0 +1,47 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "WIGXJPF"
+version = v"1.13.0"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("http://fy.chalmers.se/subatom/wigxjpf/wigxjpf-latest.tar.gz", "90ab9bfd495978ad1fdcbb436e274d6f4586184ae290b99920e5c978d64b3e6a")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/wigxjpf-1.13
+
+mkdir -p $WORKSPACE/destdir/lib
+mkdir -p $WORKSPACE/destdir/bin
+mkdir -p $WORKSPACE/destdir/shared/licenses
+
+if [ "$(uname)" == "Darwin" ]; then
+    make lib/libwigxjpf_shared.dylib
+    cp lib/libwigxjpf_shared.dylib $WORKSPACE/destdir/lib/  
+else
+    make lib/libwigxjpf_shared.so
+    cp lib/libwigxjpf_shared.so $WORKSPACE/destdir/lib/
+    cp lib/libwigxjpf_shared.so $WORKSPACE/destdir/bin/libwigxjpf_shared.dll                    
+fi
+
+cp README $WORKSPACE/destdir/shared/licenses/LICENSE
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libwigxjpf_shared", :wigxjpf)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")


W/WIGXJPF/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+if [ "$(uname)" == "Darwin" ]; then
+    make lib/libwigxjpf_shared.dylib
+    cp lib/libwigxjpf_shared.dylib $WORKSPACE/destdir/lib/  
+else
+    make lib/libwigxjpf_shared.so
+    cp lib/libwigxjpf_shared.so $WORKSPACE/destdir/lib/
+    cp lib/libwigxjpf_shared.so $WORKSPACE/destdir/bin/libwigxjpf_shared.dll                    
+fi
+
+cp README $WORKSPACE/destdir/shared/licenses/LICENSE
```
imciner2, line 30:
```suggestion
install_license COPYING COPYING.LESSER
```

Please install the proper license file.

W/WIGXJPF/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+mkdir -p $WORKSPACE/destdir/bin
+mkdir -p $WORKSPACE/destdir/shared/licenses
+
+if [ "$(uname)" == "Darwin" ]; then
+    make lib/libwigxjpf_shared.dylib
+    cp lib/libwigxjpf_shared.dylib $WORKSPACE/destdir/lib/  
+else
+    make lib/libwigxjpf_shared.so
+    cp lib/libwigxjpf_shared.so $WORKSPACE/destdir/lib/
+    cp lib/libwigxjpf_shared.so $WORKSPACE/destdir/bin/libwigxjpf_shared.dll
```
imciner2, line 27: Please only install the needed library on each platform - e.g. the mingw build shouldn't have the .so and the non-mingw shouldn't have the .dll.

W/WIGXJPF/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+    ArchiveSource("http://fy.chalmers.se/subatom/wigxjpf/wigxjpf-latest.tar.gz", "90ab9bfd495978ad1fdcbb436e274d6f4586184ae290b99920e5c978d64b3e6a")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/wigxjpf-1.13
+
+mkdir -p $WORKSPACE/destdir/lib
+mkdir -p $WORKSPACE/destdir/bin
+mkdir -p $WORKSPACE/destdir/shared/licenses
```
imciner2, line 19:
```suggestion
```

W/WIGXJPF/build_tarballs.jl
giordano: Maybe the CMake build system works better?

W/WIGXJPF/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "WIGXJPF"
+version = v"1.13.0"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("http://fy.chalmers.se/subatom/wigxjpf/wigxjpf-latest.tar.gz", "90ab9bfd495978ad1fdcbb436e274d6f4586184ae290b99920e5c978d64b3e6a")
```
giordano, line 10: Isn't there a versioned tarball? A tarball called `-latest.tar.gz` sounds like will be overwriten in place in the future, which means the recipe isn't reproducible.

W/WIGXJPF/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "WIGXJPF"
+version = v"1.13.0"
```
giordano, line 6:
```suggestion
version_string = "1.13"
version = VersionNumber(version_string)
```

W/WIGXJPF/build_tarballs.jl
giordano:
```suggestion
    ArchiveSource("http://fy.chalmers.se/subatom/wigxjpf/wigxjpf-$(version_string).tar.gz", "90ab9bfd495978ad1fdcbb436e274d6f4586184ae290b99920e5c978d64b3e6a")
```

W/WIGXJPF/build_tarballs.jl
imciner2: CMake doesn't appear to include the shared library since it hardcodes the library product as being a STATIC library, and it might have incorrect install paths (it doesn't include the GNUInstallDirs component in the CMake).

So we would need patching to remove that `STATIC` and then we can pass `BUILD_SHARED_LIBS=ON` to build it.

W/WIGXJPF/build_tarballs.jl
wes648: Yes there is a versioned tarball. it is: http://fy.chalmers.se/subatom/wigxjpf/wigxjpf-1.13.tar.gz

wes648:
I don't have a particularly good sense of what I'm doing with this pull request, to be honest. I'm mostly trying to update WIGXJFP to version 1.13 in Yggdrasil and the Julia Binary Builder Wigxjpf_jll package. I really appreciate all the work and help you are providing

W/WIGXJPF/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+mkdir -p $WORKSPACE/destdir/bin
+
+if [ "$(uname)" == "Darwin" ]; then
+    make lib/libwigxjpf_shared.dylib
+    cp lib/libwigxjpf_shared.dylib $WORKSPACE/destdir/lib/  
+else
+    make lib/libwigxjpf_shared.so
+    cp lib/libwigxjpf_shared.so $WORKSPACE/destdir/lib/
+    cp lib/libwigxjpf_shared.so $WORKSPACE/destdir/bin/libwigxjpf_shared.dll                    
+fi
```
imciner2, line 27:
```suggestion

# Patch the CMake workflow to fix the install directories and build the libraries as shared libraries
echo -e "include(GNUInstallDirs)\n$(cat CMakeLists.txt)" > CMakeLists.txt
sed -i 's/STATIC//' CMakeLists.txt

mkdir -p build && cd build

cmake -B . -S .. \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=ON

cmake --build . --parallel ${nproc}
cmake --install .
```

I think this should work to enable a CMake build of this without needing to actually have special patch files.

fingolfin:
@wes648 are you still interested in finishing this up? There are a bunch of unaddressed comments here that would have to be implemented.

wes648:
Sorry for abandoning this project. I just noticed that the dependent package I used has stopped working so I will be picking this back up.

lucifer1004:
Instead of making a new recipe, the update should be applied to https://github.com/JuliaPackaging/Yggdrasil/blob/master/W/Wigxjpf/build_tarballs.jl

lucifer1004:
I think I can take this over, since I am author of the original jll package and the julia wrapper package https://github.com/lucifer1004/Wigxjpf.jl

lucifer1004:
Opened #12846

wes648:
Thank you so much @lucifer1004 !

I'll be closing this request then

PR #8828, opened 2024-06-02
Wizard recipe: helloworld-v0.2.1

mankai-chow:

* Package name: helloworld
* Version: v0.2.1

### DIFF
--- a/H/helloworld/build_tarballs.jl
+++ b/H/helloworld/build_tarballs.jl
@@ -1,0 +1,42 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "helloworld"
+version = v"0.2.1"
+
+# Collection of sources required to complete build
+sources = [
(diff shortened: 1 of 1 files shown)


giordano:
This looks like a test?

staticfloat:
@mankai-chow can you give some more information about what this is?  In particular:

1. The name `helloworld` is not very descriptive, it looks like you have some numerical code being put together here, is this a library of your creation?  If so, you will need a better name for it than "helloworld".
2. The library you're creating is just called `lib.so`, which will be difficult for downstream C/C++/Fortran users to actually use.  You usually want to name it something like `libblasfuncs.so` or something like that so that they can link to it via `-lblasfuncs` on the command line.
3. The software itself should come with a README, a license, etc....

mankai-chow:
Dear Saba :

    Sorry. This generation by the wizard was meant just for I test. Could
you remove it for me ? Sorry for the trouble. Thank you.

Best regards,
Zheng Zhou 周正.

staticfloat:
Yep, no worries.  Thanks for the quite response!

PR #8830, opened 2024-06-03
New Recipe: FuzzifiED v0.5.2

mankai-chow:
This Julia script created by BinaryBuild Wizard intends to wrap Fortran source code stored at [FuzzifiED_Fortran](https://github.com/mankai-chow/FuzzifiED_Fortran) for the Julia package [FuzzifiED](https://github.com/mankai-chow/FuzzifiED.jl) for exact diagonalisation (ED) calculations on fuzzy sphere (hence the name, 'fuzzified' as things made fuzzy, 'ED' as the initials for 'exact diagonalisation').

### DIFF
--- a/F/FuzzifiED/build_tarballs.jl
+++ b/F/FuzzifiED/build_tarballs.jl
@@ -14,7 +14,6 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir
 cd FuzzifiED_Fortran/
-export MKL_THREADING_LAYER=GNU
 gfortran -fPIC -larpack -fopenmp -c ./cfs.f90
 gfortran -fPIC -larpack -fopenmp -c ./bs.f90
 gfortran -fPIC -larpack -fopenmp -c ./op.f90
--- a/F/FuzzifiED/build_tarballs.jl
+++ b/F/FuzzifiED/build_tarballs.jl
@@ -1,0 +1,42 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "FuzzifiED"
+version = v"0.5.2"
+
+# Collection of sources required to complete build
+sources = [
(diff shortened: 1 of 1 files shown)


F/FuzzifiED/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/mankai-chow/FuzzifiED_Fortran.git", "c96719aa5fb0e072612de78a1ebe2cf140515d47")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd FuzzifiED_Fortran/
+export MKL_THREADING_LAYER=GNU
```
imciner2, line 17:
```suggestion
```
I don't see a dependency on MKL, and also it doesn't look like it is used, so lets remove it.

F/FuzzifiED/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libfuzzified", :LibpathFuzzifiED)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"))
```
imciner2, line 37:
```suggestion
    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
    Dependency("CompilerSupportLibraries_jll"; platforms=filter(!Sys.isbsd, platforms)),
    Dependency("LLVMOpenMP_jll"; platforms=filter(Sys.isbsd, platforms)),
```

For the non-BSD systems, we need the libgomp library.

PR #8834, opened 2024-06-03
MAGEMin 1.4.6

NicolasRiel:
- corrected volume fraction nomalization mistake when buffer are used
- set buffer fraction = 0 and renormalize outputed phase fractions accordingly
- added site fraction names for out_matlab option and for the julia wrapper


NicolasRiel:
Thank you Mosè!

PR #8836, opened 2024-06-04
MAGEMin 1.4.7

NicolasRiel:
- correction of segfault for MAGEMin_jll due to wrong number of site fractions names for igneous biotite
- the segfault was not captured during the suite of tests using the locally compiled libMAGEMin.jll  for version 1.4.6 and only appeared when trying to use the released MAGEMin_jll 1.4.6


NicolasRiel:
Thank you Mosè!
Sorry for having to push a new version so quickly.

PR #8839, opened 2024-06-04
Updated the MuJoCo jll to v3.1.6

JamieMair:
Updated the jll build script to release a new 3.1.6 build for MuJoCo

Compilation requires C++17, which was a problem on x86_64 MacOS (due to older tbd files), so we needed to include the libc++ files from https://github.com/phracker/MacOSX-SDKs/releases/download/11.3/MacOSX11.3.sdk.tar.xz (similar to this issue - https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1185)

I tried to fix the MUSL builds, build I couldn't get the library to compile without significant modifications. All platforms that are not included could not be compiled for. The Linux builds all fail with gcc versions below 11 as well.


ViralBShah:
Good to merge?

M/MuJoCo/build_tarballs.jl
```diff
@@ -3,26 +3,38 @@
...
 sources = [
-    GitSource("https://github.com/deepmind/mujoco", "c9246e1f5006379d599e0bcddf159a8616d31441"),
+    GitSource("https://github.com/google-deepmind/mujoco", "21bc6e5ce41b09d80a8b8df30f6fc866b81a152b"),
     DirectorySource("bundled")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/mujoco
+CPPFLAGS="-I${prefix}/include"
```
giordano, line 17: CMake doesn't use `CPPFLAGS`

M/MuJoCo/build_tarballs.jl
```diff
@@ -3,26 +3,38 @@
...
+CXXFLAGS="-I${prefix}/include"
```
giordano, line 18:
```suggestion
CXXFLAGS="-I${includedir}"
```

M/MuJoCo/build_tarballs.jl
```diff
@@ -31,38 +43,35 @@ cmake \
...
     BuildDependency("Xorg_libXrandr_jll"; platforms=linux_platforms),
-    BuildDependency("Libglvnd_jll"; platforms=linux_platforms),
     BuildDependency("Xorg_libXi_jll"; platforms=linux_platforms),
     BuildDependency("Xorg_libXinerama_jll"; platforms=linux_platforms),
     BuildDependency("Xorg_libXcursor_jll"; platforms=linux_platforms)
 ]
 
 # Build the tarballs.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"11.1.0")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"11")
```
giordano, line 77: Missing newlines
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"11")

```

M/MuJoCo/build_tarballs.jl
JamieMair: Thanks, I've removed it now.

JamieMair:
Thanks for the review. I've committed the changes and also manually added licenses (as the audit could not find the license originally). Should be ready to go now.

PR #8853, opened 2024-06-06
[GEOS] Update to v3.12.2

jaakkor2:


jaakkor2:
Looks like for failed builds the registry did not respond fast enough

> ERROR: could not download https://us-east.pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/430a592934e8e6652a6c8f1b674dfbacddf5aab7
> 
> HTTP/2 302 (Operation too slow. Less than 1 bytes/sec transferred the last 20 seconds) while requesting https://us-east.pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/430a592934e8e6652a6c8f1b674dfbacddf5aab7

PR #8855, opened 2024-06-06
[WIP] [QT5Base] Try to update OpenSSL compat

asinghvi17:
This is holding back quite a few packages downstream, including Plots and GR, in some situations.

### DIFF
--- a/Q/Qt5Base/build_tarballs.jl
+++ b/Q/Qt5Base/build_tarballs.jl
@@ -193,7 +193,7 @@ dependencies = [
     Dependency("Glib_jll"; compat="2.68.1"),
     Dependency("Zlib_jll"),
     Dependency("CompilerSupportLibraries_jll"),
-    Dependency("OpenSSL_jll"; compat="^3"),
+    Dependency("OpenSSL_jll"; compat="3.0.13"),
     BuildDependency(PackageSpec(name="LLVMCompilerRT_jll", uuid="4e17d02c-6bf5-513e-be62-445f41c75a11", version=llvm_version);
                     platforms=filter(p -> Sys.isapple(p) && arch(p) == "x86_64", platforms_macos)),
 ]
--- a/Q/Qt5Base/build_tarballs.jl
+++ b/Q/Qt5Base/build_tarballs.jl
@@ -193,7 +193,7 @@ dependencies = [
     Dependency("Glib_jll"; compat="2.68.1"),
     Dependency("Zlib_jll"),
     Dependency("CompilerSupportLibraries_jll"),
-    Dependency("OpenSSL_jll"; compat="3"),
+    Dependency("OpenSSL_jll"; compat="^3"),
     BuildDependency(PackageSpec(name="LLVMCompilerRT_jll", uuid="4e17d02c-6bf5-513e-be62-445f41c75a11", version=llvm_version);
                     platforms=filter(p -> Sys.isapple(p) && arch(p) == "x86_64", platforms_macos)),
 ]
--- a/Q/Qt5Base/build_tarballs.jl
+++ b/Q/Qt5Base/build_tarballs.jl
@@ -193,7 +193,7 @@ dependencies = [
     Dependency("Glib_jll"; compat="2.68.1"),
     Dependency("Zlib_jll"),
     Dependency("CompilerSupportLibraries_jll"),
-    Dependency("OpenSSL_jll"; compat="1.1.10"),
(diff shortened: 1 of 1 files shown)


barche:
While it's no doubt a good idea to update this, GR has switched to Qt6 quite a while ago, so how is this holding packages back?

asinghvi17:
Huh...guess I was just stuck with an old GR version then, and Pkg wasn't able to get around that.  I just tried reinstalling Plots and it did install Qt6.  This may still have value just for people stuck on older versions of Plots or GR, though.

barche:
Even though this builds, it might not work, the full log has this:

```
[19:20:49]   OpenSSL ................................ yes
[19:20:49]     Qt directly linked to OpenSSL ........ yes
[19:20:49]   OpenSSL 1.1 ............................ yes
```

I think these binaries need testing before merging (you can dev the Qt5Base package, do `Qt5Base_jll.dev_jll()` and then extract the archive from buildkite in the override dir)


See also https://gist.github.com/seyedmmousavi/b1f6681eb37f3edbe3dcabb6e89c5d43

ViralBShah:
Please reopen if necessary.

ViralBShah:
@asinghvi17 Can you rebase this on master?

giordano:
Also, the source URL gives 404 now 😐

ViralBShah:
I don't see QT5 there: https://download.qt.io/official_releases/qt/

ViralBShah:
5.15.18 is the latest version.
v5 is now archived: https://download.qt.io/archive/qt/5.15/5.15.18/single/

ViralBShah:
@asinghvi17 Can you rebase this on master?

ViralBShah:
For Darwin x86_64, the error is:

```
Resolving package versions...
ERROR: LoadError: version 13.0.1 of package LLVMCompilerRT_jll is not available. Available versions: 12.0.0+0, 13.0.1+0, 13.0.1+1, 16.0.6+0, 16.0.6+1, 17.0.6+0
```

It says 13.0.1 is not available, but it clearly is as per the message. Does anyone know what to do?

giordano:
Yes, it's always the same: change the `VersionNumber` to a `String` :sweat_smile:

ViralBShah:
Ah ok. Seems like lots of them use v"" if we do `git grep`:

```
C/Coin-OR/coin-or-common.jl:llvm_version = v"13.0.1"
F/faust/build_tarballs.jl:llvm_versions = [v"11.0.1", v"12.0.1", v"13.0.1"]
F/faust/build_tarballs.jl:    v"13.0.1" => [sources; ],
F/finufft/build_tarballs.jl:llvm_version = v"13.0.1+1"
G/GMP/GMP@6.2.0/build_tarballs.jl:llvm_version = v"13.0.1"
G/GMP/GMP@6.2.1/build_tarballs.jl:llvm_version = v"13.0.1"
G/GMP/GMP@6.3.0/build_tarballs.jl:llvm_version = v"13.0.1"
G/GibbsSeaWater/build_tarballs.jl:llvm_version = v"13.0.1"
```

ViralBShah:
It doesn't seem to like the string. What am I doing wrong?

giordano:
That variable is used in two different places:
* Julia v1.12 wants a string as `version` keyword argument to `PackageSpec` (a `VersionNumber` would work as well, but it's a lot more restrictive that it used to be, you must now use the full build number, if any, which is a hassle, at which point the string is easier).  This is the error which popped up in a few places when we moved to v1.12
* the `preferred_llvm_version` keyword argument to `build_tarballs` is still a `VersionNumber`

You should convert only the argument to `PackageSpec` to `String` (`string(llvm_version)`), not the other

ViralBShah:
@barche I do see that it says OpenSSL 1.1. What's a good test case? Would it be easy for you to try it out?

barche:
I don't have a use case, I am not using Qt5 anymore and even with Qt6 I am not using OpenSSL. I'm not even sure what OpenSSL is used for in Qt and I also don't know what the need was for this PR?

ViralBShah:
Ah ok - I thought this PR was because some packages could no longer be installed in Julia 1.12. With the move to QT6, maybe it's just best to no longer worry about any of this.

I was poking at this because of https://github.com/JuliaPackaging/Yggdrasil/issues/12794

barche:
Ah, yes, because of Gnuplot then I think, but that builds with Qt6 now so problem solved :)

PR #8863, opened 2024-06-07
Make alsa_plugin build again.

### LINKED ISSUE #8714: ALSA + FFMPEG compatibility issue
Users have been [reporting some issues](https://github.com/JuliaAudio/PortAudio.jl/issues/126) with installing PortAudio.jl, which I think are related to an update to FFMPEG that isn't reflected in the ALSA build.

`alsa_plugins_jll` loads `libasound_module_pcm_a52.so`, and that that library seems to be dependent on `libavcodec.so.58`, which it's not finding.

It looks like the [latest version of FFMPEG_jll](https://github.com/JuliaBinaryWrappers/FFMPEG_jll.jl/blob/main/src/wrappers/x86_64-linux-gnu.jl) builds `libavcodec.so.60`, so maybe `alsa_plugins_jll` needs to be updated to match.

I'm  pretty out of touch with how the binary build system works these days, is it just a matter of kicking off a new build for `alsa_plugins_jll`?

tp2750:
Trying to fix #8714.

* ~~Identified the version of FFMPEG that provided the required library: `libasound_module_pcm_a52.so`~~
* ~~Identified the architectures, where it works. Looks like it is those that were supported by julia 1.0~~
* Updated alsa_plugins to version 1.2.7.1
* Architecture armv6l does not build


tp2750:
Updated alsa_plugin to 1.2.7.1. Now it builds with current version of FFMPEG.
Only the armv6l architecture does not build.

tp2750:
I don't understand why it fails to build for x86-64. It builds find locally.

sergstesh:
> I don't understand why it fails to build for x86-64. It builds fine locally.

I do not understand the quoted statement. Specifically, regarding "It builds fine locally" - it builds fine on what architecture ?

Regarding "it fails to build for x86-64" - what is the exact failure ? I.e. what are the error/warning messages, etc ? Are the warning/error messages to be looked at presented in https://buildkite.com/julialang/yggdrasil/builds/11024#018ff814-c642-48ef-a8bb-c17089bbf83b ?

ViralBShah:
Close in favor of #9105

PR #8869, opened 2024-06-09
[ColPack] Recompile with OpenMP

amontoison:


C/ColPack/build_tarballs.jl
```diff
@@ -3,23 +3,28 @@
...
-../configure --disable-examples --disable-openmp --prefix=${prefix} --build=${MACHTYPE} --host=${target} --disable-static --enable-shared
+cd build
+CXX=g++
+../configure --enable-examples --build=${MACHTYPE} --host=${target}
 make -j${nproc}
-make install
+
+mkdir -p ${bindir}
+cp ColPack${exeext} ${bindir}/ColPack${exeext}
+g++ -shared $(flagon -Wl,--whole-archive) libcolpack.a $(flagon -Wl,--no-whole-archive) -lgomp -o ${libdir}/libcolpack.${dlext}
```
imciner2, line 27: Using `gomp` on all platforms will lead to problems I believe. Other packages usually use the LLVM OpenMP library on the BSD/Apple systems and gomp on the others. So doing this will unconditionally will lead to two different OpenMP runtime libraries being linked into the Julia process potentially.

amontoison: @imciner2  I compile ColPack with `g++` on Mac / FreeBSD to avoid this issue.
I did that for `SPRAL` and it works fine.

giordano: Ian's point was that if you use llvm openmp for BSD systems then you can use the "more natural" compiler on all platforms instead of forcing gcc everywhere.

imciner2: I was actually thinking the problem would be the fact we have two different OpenMP runtime libraries being loaded by a single Julia process at runtime if you use another package that has OpenMP enabled. E.g. if you load ParU (which opens `LLVMOpenMP_jll` containing `libomp`) and this (which will open `CompilerSupportLibraries_jll` containing `libgomp`), they should both try to provide functions like `_omp_get_max_threads`. We might be alleviating some of the worst problems by loading with `RTLD_DEEPBIND`, but this seems like it could be a fragile system.

amontoison: Thanks for you comments Ian and Mosè!
I understand the issue now.
I opened a PR to upgrade ColPack and I modified how I link with it with OpenMP:
https://github.com/JuliaPackaging/Yggdrasil/pull/9031

@imciner2 A little bit off-topic but I am wondering what we should do if a package rely on OpenMP and also has Fortran code (like SPRAL) where we need `CompilerSupportLibraries_jll` for `libgfortran.$dlext`.

imciner2: > @imciner2 A little bit off-topic but I am wondering what we should do if a package rely on OpenMP and also has Fortran code (like SPRAL) where we always need CompilerSupportLibraries_jll for libgfortran.$dlext.

I don't think this will be a problem. `libgfortran` doesn't link to OpenMP, and just the fact we have `CompilerSupportLibraries_jll` in the environment doesn't mean `gomp` will be loaded. The library should still only be loaded when another library requests it - so we should be able to link LLVM OpenMP with applications using fortran without issues.

PR #8870, opened 2024-06-09
[CImGuiPack] Update to ImGui 1.90.8

JamesWrigley:
Note that the URL changed in https://github.com/Gnimuc/CImGui.jl/pull/119.

CC @Gnimuc

### DIFF
--- a/C/CImGuiPack/build_tarballs.jl
+++ b/C/CImGuiPack/build_tarballs.jl
@@ -1,22 +1,35 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
-using BinaryBuilder
+using BinaryBuilder, Pkg
+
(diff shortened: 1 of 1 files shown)


Gnimuc:
Thanks for following this up!

We could also pick up LLVM.jl's way to support a local-build: https://github.com/maleadt/LLVM.jl/blob/master/deps/build_local.jl

I think we should also move CImGui.jl to JuliaImGui if more people are get involved in.

JamesWrigley:
> We could also pick up LLVM.jl's way to support a local-build: https://github.com/maleadt/LLVM.jl/blob/master/deps/build_local.jl

Hmm, yeah that could be useful :thinking: 

> I think we should also move CImGui.jl to JuliaImGui if more people are get involved in.

Sure, that's fine with me :+1:

JamesWrigley:
(made this a draft for now because I think some more build changes are necessary)

JamesWrigley:
I believe this is ready to merge now.

PR #8872, opened 2024-06-09
ADOLC: new package

TimSiebert1:


A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,97 @@
...
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "windows"; )
```
fingolfin, line 53: Why this specific list? Do others platforms no work?

In general I would recommend to start with the full list of platforms and then use `filter!` to remove anything that is known to not work, with a comment each time as to what doesn't work.

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,97 @@
...
+# It seems we need a separate build for each Julia version
+if isyggdrasil
+    include("../../L/libjulia/common.jl")
+else
+    # For local test, download
+    # https://github.com/JuliaPackaging/Yggdrasil/blob/master/L/libjulia/common.jl
+    # https://github.com/JuliaPackaging/Yggdrasil/blob/master/fancy_toys.jl
+    include("common.jl")
+end
+julia_versions=VersionNumber[v"1.9.0", v"1.10.0"]
```
fingolfin, line 75: Why this restriction? If there is a proper reason, it should be stated. Otherwise, better to get rid of it to maximize where this can be used (e.g. right now you are excluding the 1.11 betas/release candidates).
```suggestion
```

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,97 @@
...
+platforms = vcat(libjulia_platforms.(julia_versions)...)
+
+# Platform for initial testing
+platforms = filter( p-> (arch(p)=="x86_64" && os(p)=="linux" && libc(p)=="glibc")
+                    || os(p)=="macos"
+                    || os(p)=="windows",
+                    platforms)
```
fingolfin, line 82: Since you hard coded a small list of platforms above, this isn't really doing anything. But as stated, I would recommend to instead start with `platforms = supported_platforms()` and then gradually remove things via `filter!`. Alternatively, use `supported_platforms(; exclude=...)`. There are plenty examples for each in Yggdrasil, just search a bit around.

(If you really need to keep this, I would at least advise against using `os(p)=="macos"` -- instead use `Sys.isapple(p)`. Similarly you can use `Sys.iswindows(p)`.)

A/ADOLC/common.jl
```diff
@@ -0,0 +1,420 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+#include("../../fancy_toys.jl") # for get_addable_spec
```
fingolfin, line 5: Why are you duplicating libjulia's `common.jl` as well as `fancy_toys.jl` in here? That seems rather problematic to me.

Everything else I commented is sorta optional, but this one seems serious to me.

fingolfin (changes requested):

A/ADOLC/build_tarballs.jl
j-fu: We had a number of problems to get cxxwrap to work. Generally, recommendations had been taken from https://discourse.julialang.org/t/binarybuilder-jl-cant-dlopen-because-of-libopenlibm-so/108486/3

Will again have a look on this. May be, also @giordano  has an update on this information.

A/ADOLC/build_tarballs.jl
j-fu: See above. We will have to test with 1.11 before modifying this.

A/ADOLC/common.jl
j-fu: Ah, this appears to be a leftover from trying locally.

A/ADOLC/build_tarballs.jl
fingolfin: Looking at the Discourse thread, @giordano recommends https://github.com/JuliaPackaging/Yggdrasil/blob/9387cacc28a4eb26dd5ab0e16a78c60030e31c8a/O/openfhe_julia/build_tarballs.jl as a good example and that module does exactly what I recommended above as best practice :-)

A/ADOLC/build_tarballs.jl
fingolfin: I don't see why you'd have to test? Unless you directly call into libjulia yourself, this should just work. Even in the worst case (i.e., it does not work after all), you can then deal with this by adjusting your Julia package which is using this JLL to either not load it there (if the JLL is for an optional feature), or to refuse to load itself (if the JLL is essential). But again, it usually works everywhere if it works on one Julia version -- except when there is a breaking change in the Julia API/ABI and you have the bad luck of trying to build the JLL before we had a chance to adjust CxxWrap. But that's not currently the case as far as I know.

Hence, I'd really suggest to not artificially limit the list of supported Julia versions here unless you really have explicit knowledge of a failure (i.e. I suggest to assume it works by default, while you seem to assume that "it doesn't work" as default)

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+# See https://github.com/JuliaLang/Pkg.jl/issues/2942
+# Once this Pkg issue is resolved, this must be removed
+uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
+delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
+
+name = "ADOLC"
+version = v"0.0.24"
+
+isyggdrasil = get(ENV, "YGGDRASIL", "") == "true"
+
```
fingolfin, line 17: This is not needed anymore, is it?
```suggestion
```

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+sources = [
+    GitSource("https://github.com/coin-or/ADOL-C.git", "b7ea10536a1e452f7dd47a90ef8c8118ce3e8432"),
+    GitSource("https://github.com/TimSiebert1/libadolccxx.git", "51c23abf9ee2e1566331b440e81503ac9e7e8f5b"),
+]
+
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd ADOL-C/ && autoreconf -fi && ./configure --enable-atrig-erf --prefix=${prefix} --build=${MACHTYPE} --host=${target}
```
fingolfin, line 28:
```suggestion
cd ADOL-C/
autoreconf -fi
./configure --enable-atrig-erf --prefix=${prefix} --build=${MACHTYPE} --host=${target}
```

fingolfin (approved):
Seems good to me now, just have two minor comments but they are not crucial at all

TimSiebert1:
Thank you for the review.

fingolfin (approved):
LGTM

j-fu:
Local tests fail in the moment. It seems we need the julia_versions for getting libjulia to load.

A/ADOLC/build_tarballs.jl
j-fu: Without 
```
julia_versions=VersionNumber[v"1.9", v"1.10"]
```
`libjulia.so.1` cannot be loaded and the wrapper package tests fail. 

Also, for 1.11 I am getting compilation errors, probably due to the new Memory based array layout.
Once @TimSiebert1 merges https://github.com/TimSiebert1/Yggdrasil/pull/1 , the corresponding change will occur here.

fingolfin: Huh? If you don't specify `julia_versions` then all that does is that the default value for it is used, which is a super set of what you use. So your statement that then libjulia cannot be loaded is odd. It is not clear to me what you mean by that either. Who cannot load it? What is the error?

fingolfin: Ahh wait you also removed `platforms = vcat(libjulia_platforms.(julia_versions)...)` but that must stay.

TimSiebert1:
With the change to `platforms = vcat(libjulia_platforms.(julia_versions)...)`, it fails because it can not find the JLCxxConfig.cmake.

j-fu:
Ok, we have three problems occuring during local tests.

Need to exclude linux-musl. The error is
```
linux-musl:
In file included from ../../ADOL-C/include/adolc/internal/common.h:26:0,
[10:26:27]                  from storemanager.h:79,
[10:26:27]                  from taping_p.h:20,
[10:26:27]                  from adouble.cpp:24:
[10:26:27] /opt/aarch64-linux-musl/aarch64-linux-musl/include/c++/7.1.0/cstdlib:144:11: error: ‘::calloc’ has not been declared
[10:26:27]    using ::calloc;
[10:26:27]            ^~~~~~
[10:26:27] /opt/aarch64-linux-musl/aarch64-linux-musl/include/c++/7.1.0/cstdlib:151:11: error: ‘::malloc’ has not been declared
[10:26:27]    using ::malloc;
[10:26:27]            ^~~~~~
[10:26:27] /opt/aarch64-linux-musl/aarch64-linux-musl/include/c++/7.1.0/cstdlib:164:11: error: ‘::realloc’ has not been declared
[10:26:27]    using ::realloc;
```
This appears during compilation of ADOLC and would need an upstream fix. Gut feeling says it is some simple missing include. Fix depends on priorities of upstream authors.

j-fu:
Need to exclude Julia 1.11:
Typical error is
```
[11:00:24] In file included from /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root/usr/local/include/jlcxx/jlcxx.hpp:14:0,
[11:00:24]                  from /workspace/srcdir/libadolccxx/src/array_types.cpp:1:
[11:00:24] /opt/x86_64-linux-gnu/x86_64-linux-gnu/sys-root/usr/local/include/jlcxx/array.hpp:194:66: error: macro "jl_array_data" requires 2 arguments, but only 1 given
[11:00:24]      return iterator(static_cast<julia_t*>(jl_array_data(wrapped())));
```
Not sure if CxxWrap.jl has been adapted to the new Memory based array layout in 1.11. Will open an issue there.

j-fu:
Bugs in ADOLC  runtests.jl (all other tests are ok):
```
Derivative array accessed out of bounds while "setADValue(...)"!!!
jac: Error During Test at /home/fuhrmann/Wias/work/julia/dev/ADOLC/test/test_derivative.jl:36
Got exception outside of a @test
incorrect function call, errorcode=-1
Stacktrace:
[1] setADValue
@ ~/.julia/packages/CxxWrap/5IZvn/src/CxxWrap.jl:624 [inlined]
[2] init_gradient(a::Vector{Adouble{ADOLC.TladoubleModule.TladoubleCxxAllocated}})
@ ADOLC ~/Wias/work/julia/dev/ADOLC/src/Adouble.jl:115
[3] tape_less_forward!(res::CxxWrap.CxxWrapCore.CxxPtr{CxxWrap.CxxWrapCore.CxxPtr{Float64}}, f::var"#f#11", n::Int64, x::Vector{Float64})
@ ADOLC ~/Wias/work/julia/dev/ADOLC/src/derivative.jl:144
[4] jac!(res::CxxWrap.CxxWrapCore.CxxPtr{CxxWrap.CxxWrapCore.CxxPtr{Float64}}, f::Function, m::Int64, n::Int64, x::Vector{Float64}, tape_id::Int64, reuse_tape::Bool)
@ ADOLC ~/Wias/work/julia/dev/ADOLC/src/derivative.jl:124
[5] derivative!(res::CxxWrap.CxxWrapCore.CxxPtr{CxxWrap.CxxWrapCore.CxxPtr{Float64}}, f::Function, m::Int64, n::Int64, x::Vector{Float64}, mode::Symbol; dir::Vector{Float64}, weights::Vector{Float64}, tape_id::Int64, reuse_tape::Bool)
@ ADOLC ~/Wias/work/julia/dev/ADOLC/src/derivative.jl:16
[6] derivative!(res::CxxWrap.CxxWrapCore.CxxPtr{CxxWrap.CxxWrapCore.CxxPtr{Float64}}, f::Function, m::Int64, n::Int64, x::Vector{Float64}, mode::Symbol)
@ ADOLC ~/Wias/work/julia/dev/ADOLC/src/derivative.jl:3
```
If this is on the Julia wrapper side, we could fix this in the wrapper code after this PR is done.

j-fu:
IMHO good to go now.

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+# platforms are passed in on the command line
+# See https://discourse.julialang.org/t/binarybuilder-jl-cant-dlopen-because-of-libopenlibm-so/108486
+# It seems we need a separate build for each Julia version
+include("../../L/libjulia/common.jl")
+
+
+
+# Current restriction: CxxWrap.jl seems to be not
+# adapted to 1.11. Typical error:
+# macro "jl_array_data" requires 2 arguments, but only 1 given
```
giordano, line 57: These comments don't match the julia versions listed below (comments suggests v1.11 doesn't work, but it's there)?  Also, I'm sure @fingolfin would suggest to just delete the line below.

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+filter!(p->  (libc(p)!="musl"), platforms)
+
+products = [
+    LibraryProduct("libadolc_wrap", :libadolc_wrap)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7");compat="0.13"),
+    Dependency(PackageSpec(name="libjulia_jll", uuid="5ad3ddd2-0711-543a-b040-befd59781bbf");compat="=1.10.10"),
```
giordano, line 76: Why such an overly strict compat bound?

giordano: Also, shouldn't this be a BuildDependency only?

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
+delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
+
+name = "ADOLC"
+version = v"0.0.24"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/coin-or/ADOL-C.git", "b7ea10536a1e452f7dd47a90ef8c8118ce3e8432"),
+    GitSource("https://github.com/TimSiebert1/libadolccxx.git", "51c23abf9ee2e1566331b440e81503ac9e7e8f5b"),
```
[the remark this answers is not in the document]
j-fu: ADOLC is an established package. libadolccxx ist code specific for the Julia wrapper.

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd ADOL-C/
+autoreconf -fi
+./configure --enable-atrig-erf --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j${nproc}
+make install
+mkdir ${prefix}/share/licenses/ADOLC
```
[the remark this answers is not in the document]
j-fu: ADOLC: automatic handling of license

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7");compat="0.13"),
+    Dependency(PackageSpec(name="libjulia_jll", uuid="5ad3ddd2-0711-543a-b040-befd59781bbf");compat="=1.10.10"),
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"8", preferred_llvm_version = v"11.0.1")
```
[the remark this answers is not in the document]
j-fu: Not really, just unsure what works together with cxxwrap. Will retry local test without this specific version.

A/ADOLC/build_tarballs.jl
j-fu: Yeah need to delete this as it is working now.

A/ADOLC/build_tarballs.jl
j-fu: ... as said above, let us retry locally without these restrictions.

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,73 @@
...
+# Same for realloc and malloc.
+filter!(p->  (libc(p)!="musl"), platforms)
+
+products = [
+    LibraryProduct("libadolc_wrap", :libadolc_wrap)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"))
```
giordano, line 68: GitHub UI isn't exactly clear, but https://github.com/JuliaPackaging/Yggdrasil/pull/8872#discussion_r1638258789 was only on the `libjulia_jll` line, `libcxxwrap_julia_jll` is still needed at runtime.  Also, you do need to specify the compat for this package, since it's a runtime package with frequent breaking changes.

giordano: To be crystal clear, https://github.com/JuliaPackaging/Yggdrasil/pull/8872#discussion_r1638258789 was referring to the line
```
    Dependency(PackageSpec(name="libjulia_jll", uuid="5ad3ddd2-0711-543a-b040-befd59781bbf");compat="=1.10.10"),
```
being a runtime dependency (libjulia should be only build-time) and forcing ***only*** v1.10.10.  I said nothing about `libcxxwrap_julia_jll`

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,73 @@
...
+    BuildDependency(PackageSpec(name="libjulia_jll", uuid="5ad3ddd2-0711-543a-b040-befd59781bbf"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"8")
```
giordano, line 73:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
               julia_compat="1.6", preferred_gcc_version=v"8", clang_use_lld=false)
```

j-fu:
For whatever reason, removing the `preferred_llvm_version` requirement leads to problems on apple.

giordano:
> For whatever reason, removing the preferred_llvm_version requirement leads to problems on apple.

https://github.com/JuliaPackaging/Yggdrasil/pull/8872#discussion_r1638443766 :point_left:

A/ADOLC/build_tarballs.jl
j-fu: Ok.

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+make install
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+# See https://discourse.julialang.org/t/binarybuilder-jl-cant-dlopen-because-of-libopenlibm-so/108486
+# It seems we need a separate build for each Julia version
+include("../../L/libjulia/common.jl")
+
+julia_versions=VersionNumber[v"1.9", v"1.10", v"1.11"]
```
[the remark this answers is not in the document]
j-fu: Well, it is just to limit the "combinatorial explosion" from the need to build for every Julia minor version. I'd rather restrict compatibility to 1.9+ to be consistent. OTOH if you can tell me that this does not matter much 1.6 would be ok as well.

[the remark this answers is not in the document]
j-fu: 1.9 is just the first version with package extensions, for this reason many packages already have this lower bound in order to save work on maintaining backward compatibility with Requires.jl.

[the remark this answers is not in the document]
giordano: Fair, but then I'd change the julia compat at the bottom to require v1.9 for consistency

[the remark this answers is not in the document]
j-fu: done this in f8cc27e

A/ADOLC/build_tarballs.jl
```diff
@@ -0,0 +1,74 @@
...
+
+platforms = vcat(libjulia_platforms.(julia_versions)...)
```
giordano, line 53: https://buildkite.com/julialang/yggdrasil/builds/11150#01901264-7007-424b-88e6-eca382831201/666-1737
```
┌ Warning: /cache/build/yggy-amdci7-8/julialang/yggdrasil/A/ADOLC/build/x86_64-linux-gnu-julia_version+1.9.0/LNG4nM6x/x86_64-linux-gnu-libgfortran5-cxx11-julia_version+1.9.0/destdir/lib/libadolc_wrap.so.0.0.1 contains std::string values!  This causes incompatibilities across the GCC 4/5 version boundary.  To remedy this, you must build a tarball for both GCC 4 and GCC 5.  To do this, immediately after your `platforms` definition in your `build_tarballs.jl` file, add the line:
│ 
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/fV5qK/src/auditor/compiler_abi.jl:247
```
```suggestion
platforms = vcat(libjulia_platforms.(julia_versions)...)
platforms = expand_cxxstring_abis(platforms)
```

j-fu:
Many thanks!

PR #8873, opened 2024-06-10
New Recipe: nativefiledialog-extended v1.1.1

Suavesito-Olimpiada:


N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "nativefiledialog_extended"
+version = v"1.1.1"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/btzy/nativefiledialog-extended/", "5786fabceeaee4d892f3c7a16b243796244cdddc", "nativefiledialog-extended-1.1.1")
```
fingolfin, line 10: Better to avoid hard coding the version
```suggestion
    GitSource("https://github.com/btzy/nativefiledialog-extended/", "5786fabceeaee4d892f3c7a16b243796244cdddc")
```

N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd nativefiledialog-extended-1.1.1/
```
fingolfin, line 16: Again, better to avoid hard coding the version. With the change to the GitSource, I think you can simply do
```suggestion
cd nativefiledialog-extended
```

N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+if [[ "${target}" == *-mingw* ]]
+then
+    cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS="-Wall -Wextra -pedantic" -DCMAKE_CXX_FLAGS="-Wall -Wextra -pedantic" -DBUILD_SHARED_LIBS=ON .
```
fingolfin, line 19: Writing the `cmake` invocation like this, four times, makes it kinda difficult to understand what's going on and what the differences are. I would suggest to instead focus on the differences in the `if/elif/else` and then use that to adjust the arguments for a single `cmake build`.

Here, it seems dropping the `-Werror` would be one thing. Then perhaps `-DCMAKE_CXX_STANDARD=11` can just always be passed? That just leaves `-DNFD_PORTAL=ON/OFF` and `-DNFD_APPEND_EXTENSION=ON/OFF` as differences. So could invoke `cmake` like this (note the use of a multi-line command which further helps to see what is going on, and also when in the future this needs to be tweaked, the diff is much easier to read.
```
# set defaults
NFD_PORTAL=OFF
NFD_APPEND_EXTENSION=OFF
if [[ "${target}" == *-linux-* ]] ; then
  NFD_PORTAL=ON
elif [[ "${target}" == *-freebsd-* ]] ; then
  NFD_APPEND_EXTENSION=ON
fi

cmake -B build \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_C_FLAGS="-Wall -Wextra -pedantic" \
    -DCMAKE_CXX_FLAGS="-Wall -Wextra -pedantic" \
    -DBUILD_SHARED_LIBS=ON \
    -DCMAKE_CXX_STANDARD=11 \
    -DNFD_PORTAL=${NFD_PORTAL} \
    -DNFD_APPEND_EXTENSION=${NFD_APPEND_EXTENSION} \
    .
```

fingolfin (commented):
Thank you. A few quick comments

N/nativefiledialog_extended/build_tarballs.jl
Suavesito-Olimpiada: Thanks for the comment, I'll update it.

N/nativefiledialog_extended/build_tarballs.jl
Suavesito-Olimpiada: Changed. Still, I took the flags from the [`cmake.yml`](https://github.com/btzy/nativefiledialog-extended/blob/master/.github/workflows/cmake.yml) action. Do you know what version of `cmake` do we use; as in the actions they use the option `-G 'MSYS Makefiles'` which threw errors in the `buildkite`. 😅

fingolfin: I don't think the `cmake` version is relevant here, but rather that we don't support targeting MSYS (which includes a POSIX emulation layer), but rather only MinGW. You could try using `-G 'MingGW Makefiles'` ? (And upstream perhaps also should use that, given that their cmake.yml only references MingGW, not MSYS.

N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+    GitSource("https://github.com/btzy/nativefiledialog-extended/", "5786fabceeaee4d892f3c7a16b243796244cdddc", "nativefiledialog-extended")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd nativefiledialog-extended/
+NFD_APPEND_EXTENSION=OFF
+if [[ "${target}" == *-linux-* ]] ; then
+    NFD_PORTAL=ON
```
fingolfin, line 19: You dropped the default value `NFD_PORTAL=OFF` above, is that intentional?

BTW I used "OFF" as default value for both options in my example but of course I have no idea about the code and whether this is sensible / correct, it is on you to decided that :-)

Suavesito-Olimpiada: Not intentional, thanks for the notice.

I decided to use `OFF`. It is the better default, as it only makes sense on Linux.

N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "nativefiledialog_extended"
+version = v"1.1.1"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/btzy/nativefiledialog-extended/", "5786fabceeaee4d892f3c7a16b243796244cdddc", "nativefiledialog-extended")
```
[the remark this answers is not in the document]
Suavesito-Olimpiada: Didn't know that it was the default name, thanks for the heads-up. Changed it too.

N/nativefiledialog_extended/build_tarballs.jl
Suavesito-Olimpiada: I'll let it out like this right now. As I'm more worried to be able to user to test it, as I only have Linux `x86_64`.

N/nativefiledialog_extended/build_tarballs.jl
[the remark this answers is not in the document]
fingolfin: What is the error?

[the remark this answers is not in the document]
Suavesito-Olimpiada: It did not find the directory `"nativefiledialog-extended"` in the `cd` command. 😅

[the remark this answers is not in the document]
giordano: I'm also confused.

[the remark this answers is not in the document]
Suavesito-Olimpiada: Let me force-push again that change to see if it reproduces. :b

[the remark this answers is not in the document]
Suavesito-Olimpiada: Reproduced here https://buildkite.com/julialang/yggdrasil/builds/11077

N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "nativefiledialog_extended"
+version = v"1.1.1"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/btzy/nativefiledialog-extended/", "5786fabceeaee4d892f3c7a16b243796244cdddc")
```
giordano, line 10:
```suggestion
    GitSource("https://github.com/btzy/nativefiledialog-extended.git",
              "5786fabceeaee4d892f3c7a16b243796244cdddc"),
```

N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd nativefiledialog-extended
```
giordano, line 16:
```suggestion
cd ${WORKSPACE}/srcdir/nativefiledialog-extended
```

Suavesito-Olimpiada: Pushed.

N/nativefiledialog_extended/build_tarballs.jl
Suavesito-Olimpiada: Puched.

N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+    -DNFD_PORTAL=${NFD_PORTAL} \
+    -DNFD_APPEND_EXTENSION=${NFD_APPEND_EXTENSION} \
+    .
+cmake --build build --parallel ${nproc}
+cmake --install build
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = filter!(p -> arch(p) != "armv6l", supported_platforms(; experimental=true))
```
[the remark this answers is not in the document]
giordano: Also, do you need to filter out armv6l at all?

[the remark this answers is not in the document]
Suavesito-Olimpiada: Done too.

Suavesito-Olimpiada:
Thanks for all the reviews! Do you think this is ready for merging?

fingolfin (approved):
Looks good to me

N/nativefiledialog_extended/build_tarballs.jl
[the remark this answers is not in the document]
Suavesito-Olimpiada: > Also, do you need to filter out armv6l at all?

`Dbus`, and `Gtk3` does not mention `armv6l`. I don't know if it will work, but I put it to test the build.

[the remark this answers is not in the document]
giordano: Ah, our dbus build is _old_.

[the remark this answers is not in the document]
Suavesito-Olimpiada: It was a failure for `armv6l` [here](https://buildkite.com/julialang/yggdrasil/builds/11098#01900be3-b49b-4c1b-8da0-846e8cf2eb6b).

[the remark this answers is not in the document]
Suavesito-Olimpiada: I'll try building against `SDL2` in those platforms, to see if it compiles.

N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,102 @@
...
+products = [
+    LibraryProduct("libnfd", :libnfd)
+    LibraryProduct("libnfd_gtk", :libnfd_gtk)
+    LibraryProduct("libnfd_sdl", :libnfd_sdl)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency(PackageSpec(name="Xorg_xorgproto_jll", uuid="c4d99508-4286-5418-9131-c86396af500b"); platforms=filter(p->Sys.islinux(p)||Sys.isfreebsd(p), platforms))
+    BuildDependency(PackageSpec(name="GTK3_jll", uuid="77ec8976-b24b-556a-a1bf-49a033a670a6"); platforms=filter(p->Sys.islinux(p)||Sys.isfreebsd(p), platforms))
```
giordano, line 96: Why gtk3 and sdl2 are build-only dependencies?

N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,102 @@
...
+CMAKE_FLAGS="-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_CXX_STANDARD=11 \
+    -DCMAKE_C_FLAGS=\"-Wall -Wextra -pedantic\" \
+    -DCMAKE_CXX_FLAGS=\"-Wall -Wextra -pedantic\" \
+    -DBUILD_SHARED_LIBS=ON"
+
+# NATIVE and PORTAL
+BUILD="build"
+PREFIX="${prefix}"
```
giordano, line 29: What's the point of setting a variable to another variable with the same name?

N/nativefiledialog_extended/build_tarballs.jl
Suavesito-Olimpiada: Most of the platforms and users are supported only with "native" (Mingw, Mpple) and "portal" (Linux, everything except `armv6l`) options.

giordano: I don't quite understand what that means.

Suavesito-Olimpiada: My plan is to write two package extensions, one for SDL2 and the other for GTK3. In `armv6l` or FreeBSD systems, I plan to show a warning to users; saying that they need to install manually one of the two packages to be able to use this.

N/nativefiledialog_extended/build_tarballs.jl
Suavesito-Olimpiada: Bash variables are case-sensitive (I think), making `prefix` and `PREFIX` different. The point was to make it easier to read only the "configs" (`PREFIX`, `BUILD`, `NFD_PORTAL`, `NFD_APPEND_EXTENSION`) before every `cmake` call.

N/nativefiledialog_extended/build_tarballs.jl
Suavesito-Olimpiada: It looks like I misinterpreted part of the README before. NFDe doesn't have SDL2 backend, only GTK3, Portal (dbus), and Native (Windows, Apple). Dbus now [works](https://github.com/JuliaPackaging/Yggdrasil/pull/8884) for all Linux platforms; then, GTK3 is going to be only an alternative on all supported targets as a package extension.

N/nativefiledialog_extended/build_tarballs.jl
```diff
@@ -0,0 +1,95 @@
...
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libnfd", :libnfd)
+    LibraryProduct("libnfd_gtk", :libnfd_gtk)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    BuildDependency(PackageSpec(name="Xorg_xorgproto_jll", uuid="c4d99508-4286-5418-9131-c86396af500b"); platforms=platforms)
+    BuildDependency(PackageSpec(name="GTK3_jll", uuid="77ec8976-b24b-556a-a1bf-49a033a670a6"); platforms=platforms)
```
giordano, line 90: I still don't understand why this isn't a regular Dependency.

Suavesito-Olimpiada: Given that for all platforms there is an alternative (native API for Windows/Apple, portal for Linux/FreeBSD), I decided to support the GTK3 backend only through a package extension.

I don't want the `NativeFileDialog.jl` (the user facing API of this artifact) to install unnecessary packages (strictly unnecessary for Windows/Apple, not necessary for most Linux/FreeBSD) for users when there is an alternative for basically everyone.

Suavesito-Olimpiada: Though, for the support of Portal in FreeBSD/Linux to be complete, there is still [this](https://github.com/JuliaPackaging/Yggdrasil/pull/8884) PR to update `Dbus` to be merged.

imciner2: Huh? I still don't understand. You are telling it that GTK3 is available, and it can link against it, so it will be. That should lead to a problem at runtime because the dependency is missing.

IMO, you should just provide these for Linux and FreeBSD unconditionally. The weight of these dependencies doesn't seem too large in my opinion.

Suavesito-Olimpiada: There are two artifacts generated here:

 - Portal/Native one, will use Dbus in Linux/FreeBSD and native API's on Windows/Apple, this is `libnfd`.
 - GTK3, will use `GTK3_jll` to call on a file dialog on all supported platform, this is `libnfd_gtk`.

I plant to use the first in the main package, as it covers _most_ of user needs (probably all if you do not count people who are installing a _really extremely terse_ `X` environment on Linux/FreeBSD).

I guess that importing `libnfd_gtk` will fail, because `dlopen`ing at `_init_` time will fail to link to GTK3. Is there a way to not open it at `_init_` time? I plan to program `NativeFileDialog.jl` use the GTK3 backend only as a package extension.

GTK3 is **NOT** necesary, not for basically _all users_. Installing and importing it cost time, disk and network, for no benefit. Having the GTK3 backend to work as an extension _does make this package lighter_, only incurring in _very light_ inconvenience when there is no other alternative for this to work.

I don't understand why this plan is not sound to you. It covers all possible users, benefiting most for an extremely small (`add GTK3`) cost against the very few that decided to not have `dbus` and `portal` installed in their systems (which should be _really, really, few people_).

I do not have my mind set in stone, by any means. I'm sorry if I sound stuborn. I'm just not yet capable of grasping if the solution I propose goes against some design of the registry, or if I'm being unable to transmit what I think is a reasonable compromise. 😅

tecosaur: @giordano friendly poke :slightly_smiling_face:

tecosaur: (p.s. @Suavesito-Olimpiada you could update this to 1.2.1 now)

giordano: > Is there a way to not open it at `_init_` time?

Pass `dont_dlopen`: https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct

> I don't understand why this plan is not sound to you.

Because in my mind library `libA` in package `PackageA` links dynamically to `libB` in `PackageB` implies that `PackageA` depends on `PackageB`.  If this is only a platform-specific issue, we already allow for platform-specific [`Dependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.Dependency)'s.

Suavesito-Olimpiada: > Pass `dont_dlopen`: https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct

Thanks!

> Because in my mind library `libA` in package `PackageA` links dynamically to `libB` in `PackageB` implies that `PackageA` depends on `PackageB`.

I understand that. Maybe this is a hack, but, what I want to do is to deliver both artifacts `libnfd_gtk` and `libnfd`, with and without GTK respectively. `libnfd` is dynamically loaded by default, and `libnfd_gtk` is only loaded and used from an extension triggered by `GTK.jll`.

Suavesito-Olimpiada: Part of the reason I'm thinking about this, instead of simply adding Gtk is to make the default installation (without the extension) smaller. 😅

Suavesito-Olimpiada: > (p.s. @Suavesito-Olimpiada you could update this to 1.2.1 now)

Changed it, thanks for the ping.

Suavesito-Olimpiada: > Because in my mind library `libA` in package `PackageA` links dynamically to `libB` in `PackageB` implies that `PackageA` depends on `PackageB`. If this is only a platform-specific issue, we already allow for platform-specific [`Dependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.Dependency)'s.

Changed it to two recipes: one without Gtk, only with portal support in Linux/FreeBSD, and the second with `GTK3` as a normal dependency.

Is this way fine? @giordano

tecosaur:
Is there a reason why this has been closed?

Suavesito-Olimpiada:
I plan to contribute it back in two PR's, one for `nativefiledialog_extended_portal` and the other one for `nativefiledialog_extended_gtk`. The difference being that the first will only have `dbus` as a dependency (to work only with [portals](https://wiki.archlinux.org/title/XDG_Desktop_Portal)), and the second will have `GTK3`, but not `dbus` (at least not directly).

I'm the maintainer of [NativeFileDialog.jl](https://github.com/JuliaGraphics/NativeFileDialog.jl) (which honestly needs some love). The library which it is build on ([`nativefiledialog`](https://github.com/mlabbe/nativefiledialog)) is no longer maintained. My plan is to change the underlying library to `nativefiledialog_extended`, which is still maintained and have better support, to release a `v1.0`.

I do not want to include `GTK3` by default because it is a big dependency (disk-wise), and it has caused problems for some users. Also, there is a way to make it work for Windows, Mac, and what I think is a big subset of GNU/Linux users, with portals (given the table of support on the arch wiki). Which is why I separated the builds of the packages.

As for the gtk build, I plan to release a new package that uses it under the hood, so users who do not have portals can make use of it. :)

PR #8874, opened 2024-06-10
New Recipe: polyfill_glibc v0.1.0

Suavesito-Olimpiada:
### **UPDATE:**

It was added upstream a `LICENSE` file, making the stated below no longer true.

---

~~There is still no LICENSE file, but I already filled an [issue](https://github.com/corsix/polyfill-glibc/issues/5), though the license is stated in the [`README.md`](https://github.com/corsix/polyfill-glibc?tab=readme-ov-file#license) as `MIT` wit~~

 > "the above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software" clause of the MIT license is explicitly waived for any such pieces.

### DIFF
--- a/P/polyfill_glibc/build_tarballs.jl
+++ b/P/polyfill_glibc/build_tarballs.jl
@@ -14,7 +14,7 @@ sources = [
 script = raw"""
 cd ${WORKSPACE}/srcdir/polyfill-glibc/
 ninja -j${nproc} polyfill-glibc
-install -D ./polyfill-glibc ${prefix}/bin/polyfill-glibc
+install -Dvm 755 ./polyfill-glibc ${prefix}/bin/polyfill-glibc
 """
 
 # These are the platforms we will build for by default, unless further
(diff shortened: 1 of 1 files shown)


Suavesito-Olimpiada:
The platform is only `linux-x86_64-glibc` as it is the only one that is currently supported. 😅

Suavesito-Olimpiada:
The `LICENSE` file was added. I think this is ready.

P/polyfill_glibc/build_tarballs.jl
```diff
@@ -0,0 +1,37 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/corsix/polyfill-glibc.git", "f2005b78d6d0402ddc1079bcfb6db31a424f663b")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/polyfill-glibc/
+ninja -j${nproc} polyfill-glibc
+install -D ./polyfill-glibc ${prefix}/bin/polyfill-glibc
```
imciner2, line 17:
```suggestion
install -Dvm 755 ./polyfill-glibc ${prefix}/bin/polyfill-glibc
```
Just a nit, we like to be explicit with the permissions when installing the binaries.

Suavesito-Olimpiada: Added.

PR #8875, opened 2024-06-10
[Tasmanian] link against libblasstrampoline bump version -> 8.0.1

MichelJuillard:


T/Tasmanian/build_tarballs.jl
```diff
@@ -41,11 +43,13 @@ products = [
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[
+dependencies = [
     # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
     # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
     Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
+    Dependency("libblastrampoline_jll"),
```
giordano, line 51: One thing: libblastrampoline went through a few ABI-breaking changes, oldest version since this was stabilised is libblastrampoline v5.4.0, who is only in Julia v1.9. Are you happy with restricting compatibility of this package to that julia version?
```suggestion
    Dependency("libblastrampoline_jll"; compat="5.4"),
```

MichelJuillard: Thanks @giordano. I was under the impression that `compat="5.4"` was required to use  `DBLA_VENDOR="libblastrampoline"`. What's the alternative? Following the instructions in `BinaryBuilder` documentation to link with `libblastrampoline` ?

giordano: So, it's generally a good idea to specify compat bounds of packages, especially in cases like libblastrampoline which have broken the ABI a few times along the way.  So setting `compat="5.4"` is pretty much necessary in any case if you want to move forward and get something working.  I'm also pointing out that doing this also corresponds with effectively requiring Julia v1.9 for this and newer versions of Tasmanian (past versions will still be compatible with Julia v1.6-1.8 of course).

MichelJuillard: I understand. I believe that the benefit of having BLAS compensate the restriction on Julia version. Thanks for your help.

giordano: Ok, let's do it then

imciner2: > I was under the impression that compat="5.4" was required to use DBLA_VENDOR="libblastrampoline"

These actually aren't related. `DBLA_VENDOR="libblastrampoline"` is dependent on the CMake version (it requires 3.29.3+1, but if you leave off the version you will get the latest, which is this one). This will work with any libblastrampoline with and soversion of 5.

MichelJuillard: Thanks Ian for the clarification

PR #8881, opened 2024-06-12
Patch Z3 to support third party memory allocators

remysucre:
Workaround for https://github.com/ahumenberger/Z3.jl/issues/28


Z/z3/build_tarballs.jl
```diff
@@ -36,9 +35,12 @@ fi
 """
 
 # Bash recipe for building across all platforms
+# Patches Z3 to work around https://github.com/Z3Prover/z3/issues/6890 (closed but not fixed!)
```
imciner2, line 38: This is better placed right above the line doing the patching, and also a description of what the patch does - e.g. why is that sed command patching it? The linked issue doesn't mention this as a fix at all.

imciner2 (dismissed):

Z/z3/build_tarballs.jl
fingolfin, line 38: If this comment refers to the `sed` invocation below, why not move it right next to that?

Z/z3/build_tarballs.jl
```diff
@@ -9,7 +9,6 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "z3"
 version = v"4.13.0"
-julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10", v"1.11"]
```
fingolfin, line 12: Thanks :-)

Z/z3/build_tarballs.jl
```diff
@@ -36,9 +35,12 @@ fi
 """
 
 # Bash recipe for building across all platforms
+# Patches Z3 to work around https://github.com/Z3Prover/z3/issues/6890 (closed but not fixed!)
 script = macfix * raw"""
 cd $WORKSPACE/srcdir/z3-*/core
 
+sed -i '16,30d' src/util/memory_manager.cpp
```
fingolfin, line 42: That's rather opaque and brittle, isn't it? If someone else wants to update the recipe in the future, they'll first have to reverse engineer what exactly this does.

Can't you instead e.g. specify the exact string to be deleted? Or provide an actual patchfile?

Z/z3/build_tarballs.jl
remysucre: I linked to a new issue that documents more details about the problem, and moved the comment down.

Z/z3/build_tarballs.jl
remysucre: Agreed, pushed a more explicit command and linked to a new issue documenting the bug.

PR #8884, opened 2024-06-13
update dbus to 1.14.10

Suavesito-Olimpiada:
And delete the `armv6l` filter for platforms. @giordano

### DIFF
--- a/D/Dbus/build_tarballs.jl
+++ b/D/Dbus/build_tarballs.jl
@@ -43,7 +43,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="Expat_jll", uuid="2e619515-83b5-522b-bb60-26c02a35a201"); compat="2.6.2"),
+    Dependency(PackageSpec(name="Expat_jll", uuid="2e619515-83b5-522b-bb60-26c02a35a201"); compat="2.2.10"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/D/Dbus/build_tarballs.jl
+++ b/D/Dbus/build_tarballs.jl
@@ -1,31 +1,32 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
-using BinaryBuilder
+using BinaryBuilder, Pkg
 
 name = "Dbus"
-version = v"1.12.16"
+version = v"1.14.10"
+
+# NOTE: Next stable version 1.16 will require cmake for Windows and meson for everything
+# else. Check [here](https://cgit.freedesktop.org/dbus/dbus/tree/NEWS?h=master#n21).
 
 # Collection of sources required to build Dbus
 sources = [
-    ArchiveSource("https://dbus.freedesktop.org/releases/dbus/dbus-$(version).tar.gz",
-                  "54a22d2fa42f2eb2a871f32811c6005b531b9613b1b93a0d269b05e7549fec80"),
+    ArchiveSource("https://dbus.freedesktop.org/releases/dbus/dbus-$(version).tar.xz",
+                  "ba1f21d2bd9d339da2d4aa8780c09df32fea87998b73da24f49ab9df1e36a50f"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/dbus-*
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} \
-    --with-xml=expat \
     --with-dbus-user=messagebus \
     --with-system-pid-file=/var/run/dbus.pid \
     --disable-verbose-mode \
     --disable-static \
     --enable-inotify \
-    --disable-dnotify \
     --disable-asserts \
     --enable-user-session \
     --with-session-socket-dir=/tmp \
-    --with-x
+    --with-x=no
 make -j${nproc}
 make install
 """
@@ -33,9 +34,8 @@ make install
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = filter!(p -> Sys.islinux(p) || Sys.isfreebsd(p), supported_platforms())
-# TODO: Remove this restriction for the next build
-filter!(p -> arch(p) != "armv6l", platforms)
 
+# The products that we will ensure are always built
 products = [
     LibraryProduct("libdbus-1", :libdbus),
     ExecutableProduct("dbus-daemon", :dbus_daemon)
@@ -43,8 +43,8 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Expat_jll"; compat="2.2.7"),
+    Dependency(PackageSpec(name="Expat_jll", uuid="2e619515-83b5-522b-bb60-26c02a35a201"); compat="2.6.2"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")


fingolfin:
I wonder what needs dbus and for what. Do we really need X11 support (which is currently causing the failure)

Suavesito-Olimpiada:
What was missing was the `Xorg_libX11`, and `Xorg_xorgproto` (dependency for the first). I think it should be automatable the dependency detection of some `autoconfigure` packages and `cmake` ones. I might try to do it someday.

D/Dbus/build_tarballs.jl
```diff
@@ -33,18 +31,19 @@ make install
...
     LibraryProduct("libdbus-1", :libdbus),
     ExecutableProduct("dbus-daemon", :dbus_daemon)
 ]
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Expat_jll"; compat="2.2.7"),
+    Dependency(PackageSpec(name="Expat_jll", uuid="2e619515-83b5-522b-bb60-26c02a35a201"); platforms=platforms),
+    BuildDependency(PackageSpec(name="Xorg_xorgproto_jll", uuid="c4d99508-4286-5418-9131-c86396af500b"); platforms=platforms),
+    BuildDependency(PackageSpec(name="Xorg_libX11_jll", uuid="4f6342f7-b3d2-589e-9d20-edeb45f2b2bc"); platforms=platforms),
```
imciner2, line 45: This should be a normal dependency - it provides a shared library to link against.

Also, if a dependency is needed for all platforms, then don't specify the `platforms` argument to it.

D/Dbus/build_tarballs.jl
```diff
@@ -33,18 +31,19 @@ make install
...
+# The products that we will ensure are always built
 products = [
     LibraryProduct("libdbus-1", :libdbus),
     ExecutableProduct("dbus-daemon", :dbus_daemon)
 ]
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Expat_jll"; compat="2.2.7"),
+    Dependency(PackageSpec(name="Expat_jll", uuid="2e619515-83b5-522b-bb60-26c02a35a201"); platforms=platforms),
```
imciner2, line 43: Why did you remove the compat entry from expat here?

imciner2 (changes requested):

D/Dbus/build_tarballs.jl
giordano: Also, what's the point of `platforms=platforms`?  You aren't really filtering anything, as far as I can tell.

D/Dbus/build_tarballs.jl
Suavesito-Olimpiada: It looks like the original script _did not_ include `X11` support. Given that leaving `--with-x` would mean `--with-x=auto`, then it would not detect `X11` and just "jump over". Checking `ldd` output on both artifact on `v"1.12.6"` confirms that there is no link to it.

On the new version, `ldd` outputs the _same_ for both artifacts. Meaning that `X11` support is for some other generated file, not for the ones we use. I'll set `--with-x=no` and delete the dependencies.

D/Dbus/build_tarballs.jl
Suavesito-Olimpiada: For the `platforms=platforms`, it was my ignorance on how it interacted with the option on `build_tarballs`. I'll delete it.

As for the first, the only mention I've found about the versions that `dbus` supports for `expat` is [this](https://cgit.freedesktop.org/dbus/dbus/tree/NEWS?h=master#n1603); in which they mention only that "Expat >= 2.1.0 is required". Given that `dbus` builds correctly against the latest version, is there a way to test that this build does not break some other package?

D/Dbus/build_tarballs.jl
Suavesito-Olimpiada: BTW, the change in behavior happened [here](https://gitlab.freedesktop.org/dbus/dbus/-/merge_requests/263).

D/Dbus/build_tarballs.jl
giordano: The compat should still be restored, or adapted, but definitely not removed...

Suavesito-Olimpiada: Update it. Is it OK to set it to `"2.1.0, 3"`, or, do you have any specific suggestion?

Suavesito-Olimpiada: I think that should have done it. Do you think is it related to [`BinaryBuilder.jl#1330`](https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1330) or [`Yggdrasil#8211`](https://github.com/JuliaPackaging/Yggdrasil/issues/8211)?

giordano: I'm totally missing what's the bug here and how anything is possibly related to the linked tickets.

giordano: >  Is it OK to set it to "^2.1"; or, do you have any specific suggestion?

Why don't you just keep the existing one?

Suavesito-Olimpiada: > I'm totally missing what's the bug here and how anything is possibly related to the linked tickets.

According to [#`BinaryBuilderBase.Dependency`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.Dependency)

> The optional keyword argument `compat` can be used to specify a string for use in the `Project.toml` of the generated Julia package.

Doesn't that means that setting `compat="^2.1"` should set the `[compat]` of `Expat_jll` in the `Project.toml` of `nativefiledialog_jll.jl` to `Expat_jll = "^2.1"`? Which, according to [#Caret-specifiers](https://pkgdocs.julialang.org/v1/compatibility/#Caret-specifiers) implies the versions of `Expat_jll` accepted are `[2.1.0, 3.0.0)`?

But, the `buildkite` log says that:

```julia
ERROR: LoadError: Unsatisfiable requirements detected for package Expat_jll [2e619515]:
 Expat_jll [2e619515] log:
 ├─possible versions are: 2.2.7-2.6.2 or uninstalled
 └─restricted to versions 2.1.0 by an explicit requirement — no versions left
```

I'll change it to the current one and left it like that; but, I don't understand where I'm interpreting wrong the docs. 😓

Suavesito-Olimpiada: Given that I don't see where I'm wrong in reading the docs, I searched issues for errors with `Dependency` and `compat` bounds. That is why I landed to those I shared, though I did not research further, which is my mistake.

giordano: Try to read more:

> If compat is non-empty and build_version is not passed, the latter defaults to the minimum version compatible with the compat specifier.

Suavesito-Olimpiada: I changed it to the latest version of `Expat_jll`, as the old one did not have support for `armv6l` as seen in [`buildkite`](https://buildkite.com/julialang/yggdrasil/builds/11261). Now in green for all platforms.

D/Dbus/build_tarballs.jl
```diff
@@ -1,50 +1,50 @@
...
+# The products that we will ensure are always built
 products = [
     LibraryProduct("libdbus-1", :libdbus),
     ExecutableProduct("dbus-daemon", :dbus_daemon)
 ]
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Expat_jll"; compat="2.2.7"),
+    Dependency(PackageSpec(name="Expat_jll", uuid="2e619515-83b5-522b-bb60-26c02a35a201"); compat="2.6.2"),
```
giordano, line 46:
```suggestion
    Dependency(PackageSpec(name="Expat_jll", uuid="2e619515-83b5-522b-bb60-26c02a35a201"); compat="2.2.10"),
```

D/Dbus/build_tarballs.jl
Suavesito-Olimpiada: > Try to read more:
> 
> > If compat is non-empty and build_version is not passed, the latter defaults to the minimum version compatible with the compat specifier.

I'll be more careful, thanks for the answer.

D/Dbus/build_tarballs.jl
Suavesito-Olimpiada: Merged.

imciner2 (commented):
@giordano any more comments. I think all of mine have been addressed sufficently to merge this.

imciner2 (approved):

PR #8888, opened 2024-06-13
Update boost to 1.85.0

mattwigway:
This updates Boost to 1.85.0 (the latest version). The currently packaged boost 1.79.0 is incompatible with gcc 12 (see boostorg/function#42).

NB: this also changes the download URL to `boost.io` from `boostorg.jfrog.io`. This appears to be the current official source linked from [boost.org](https://boost.org).

### DIFF
--- a/B/boost/build_tarballs.jl
+++ b/B/boost/build_tarballs.jl
@@ -20,7 +20,7 @@ cd $WORKSPACE/srcdir/boost*/
 # on PowerPC, apply https://github.com/boostorg/charconv/pull/183
 if [[ $target == *powerpc64le* ]]; then
     atomic_patch ../bundled/patches/183.patch
-end
+fi
 
 ./bootstrap.sh --prefix=$prefix --without-libraries=python --with-toolset="--cxx=${CXX_FOR_BUILD}"
 
--- a/B/boost/build_tarballs.jl
+++ b/B/boost/build_tarballs.jl
@@ -10,11 +10,17 @@ sources = [
     ArchiveSource(
         "https://archives.boost.io/release/$(version)/source/boost_$(version.major)_$(version.minor)_$(version.patch).tar.bz2",
         "7009fe1faa1697476bdc7027703a2badb84e849b7b0baad5086b087b971f8617"),
+    DirectorySource("./bundled")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/boost*/
+
+# on PowerPC, apply https://github.com/boostorg/charconv/pull/183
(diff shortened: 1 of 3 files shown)


mattwigway:
This is the relevant error from the mingw platforms. I'm not sure what to make of this, I guess it's not compiling a valid object? I'm not sure how to debug that.
```
[22:28:23] > windres --input res.rc --output res.o
[22:28:23] > /opt/bin/x86_64-linux-musl-cxx11/x86_64-linux-musl-g++ -DNDEBUG -Wl,res.o bindjam.cpp builtins.cpp class.cpp command.cpp compile.cpp constants.cpp cwd.cpp debug.cpp debugger.cpp execcmd.cpp execnt.cpp execunix.cpp filent.cpp filesys.cpp fileunix.cpp frames.cpp function.cpp glob.cpp hash.cpp hcache.cpp hdrmacro.cpp headers.cpp jam_strings.cpp jam.cpp jamgram.cpp lists.cpp make.cpp make1.cpp md5.cpp mem.cpp modules.cpp native.cpp option.cpp output.cpp parse.cpp pathnt.cpp pathsys.cpp pathunix.cpp regexp.cpp rules.cpp scan.cpp search.cpp startup.cpp tasks.cpp timestamp.cpp value.cpp variable.cpp w32_getreg.cpp mod_jam_builtin.cpp mod_jam_class.cpp mod_jam_errors.cpp mod_jam_modules.cpp mod_order.cpp mod_path.cpp mod_property_set.cpp mod_regex.cpp mod_sequence.cpp mod_set.cpp mod_string.cpp mod_summary.cpp mod_sysinfo.cpp mod_version.cpp -o b2
[22:28:51] res.o: file not recognized: File format not recognized
```

The PowerPC ones (relevant error below) look like they might be an upstream issue. Boost.Charconv was introduced in Boost 1.85.0. The offending code is in a #else block that I suspect is only triggered on PowerPC, and [has since been removed](https://github.com/boostorg/charconv/commit/0a82afd2643d4546651ba0142c4f2466a1336403). [conda-forge applied a patch to workaround](https://github.com/conda-forge/boost-feedstock/pull/196); I'm working on doing the same.

mattwigway:
PowerPC builds are now working, using the patch copied from [conda-forge](https://github.com/conda-forge/boost-feedstock/blob/main/recipe/patches/0002-Merge-pull-request-183-from-boostorg-ppc64le.patch), itself pulled from the boost repository, but with file paths updated to apply to the boost source tarball.

mattwigway:
The compile is working on Windows locally when I run `julia build_tarballs.jl x86_64-w64-mingw32 --debug=error --verbose` . I'll investigate more why it's not working here next week.

PR #8891, opened 2024-06-14
pandoc 3.2.0

palday:


P/pandoc/build_tarballs.jl
```diff
@@ -2,17 +2,17 @@ using BinaryBuilder
 
 # Collection of pre-build pandoc binaries
 name = "pandoc"
-pandoc_ver = "3.1.12"
+pandoc_ver = "3.2"
```
omus, line 5: Odd that this release switched to a 2-digit version: https://github.com/jgm/pandoc/releases/tag/3.2

palday: I think it's more just that they don't feel the need to specify a `.0` and there will be `.1` at some point

PR #8892, opened 2024-06-14
pandoc-crossref 0.3.17.1a

palday:
@giordano should I just bump the version to 0.3.18? That doesn't seem quite right because it's still pandoc-crossref 0.3.17, but the build number is important to distinguish which version of pandoc it was compiled against (and that's why 0.3.17 already exists as a JLL -- it was compiled against an older version of pandoc)

### DIFF
--- a/P/pandoc_crossref/build_tarballs.jl
+++ b/P/pandoc_crossref/build_tarballs.jl
@@ -1,10 +1,12 @@
 using BinaryBuilder
+using Pkg: PackageSpec
 
 # Collection of pre-build pandoc binaries
 name = "pandoc_crossref"
 
 crossref_ver = "0.3.17.1a"
-version = v"0.3.18"
+panddoc_jll_version = v"3.2.0"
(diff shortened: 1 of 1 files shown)


omus:
I've set a `Dependency` on `pandoc_jll` as pandoc-crossref requires a specific version of pandoc to be installed or will produce a "ton of warnings". Having this declared as a dependency in Julia should avoid seeing those warnings.
 
Additionally, with this dependency in place we can leave the version number as `0.3.17` which matches the version set in pandoc-crossref. Now when a user requests a specific version of `pandoc_jll` the correct version of `pandoc_crossref_jll` will be installed.

omus:
Unfortunately Pkg.jl doesn't use separate compat entries for each build which means the latest build updates the compat entries for all of `0.3.17`. We're yanking the problematic release here: https://github.com/JuliaRegistries/General/pull/109007.

It looks like we'll have to define a new version convention for this JLL.

PR #8893, opened 2024-06-14
[OSRM] Add new binary, second try...

jeremiahpslewis:


jeremiahpslewis:
Error on `x86-64-linux-gnu-cxx11`:

```Julia
[ Info: Checking shared library lib/libosrm.so
--
ERROR: could not load library "/cache/build/yggy-amdci7-8/julialang/yggdrasil/O/OSRM/build/x86_64-linux-gnu-cxx11/ZsPx5jB0/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libosrm.so"
  | /cache/julia-buildkite-plugin/julia_installs/bin/linux/x64/1.7/julia-1.7-latest-linux-x86_64/bin/../lib/julia/libstdc++.so.6: version `GLIBCXX_3.4.31' not found (required by /cache/build/yggy-amdci7-8/julialang/yggdrasil/O/OSRM/build/x86_64-linux-gnu-cxx11/ZsPx5jB0/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libosrm.so)
  | ┌ Warning: lib/libosrm.so cannot be dlopen()'ed
```

jeremiahpslewis:
Error on `*-apple-*` builds:

```bash
ld64.lld: error: undefined symbol: osrm::extractor::CompressedEdgeContainer::GetBucketReference(unsigned int) const
```

imciner2:
> Error on x86-64-linux-gnu-cxx11

That is because it is using GCC 13, so it gets linked against a newer glibcxx version because that is what ships with it. You need to lower the GCC version requested to get around this.

jeremiahpslewis:
> > Error on x86-64-linux-gnu-cxx11
> 
> That is because it is using GCC 13, so it gets linked against a newer glibcxx version because that is what ships with it. You need to lower the GCC version requested to get around this.

If I lower GCC below v12, other target builds break. Is there another strategy for dealing with this issue (I don't exactly understand what the problem is...all of the dependencies were built with earlier GCC versions).

mattwigway:
@jeremiahpslewis I've also been hacking on this, though it seems like we are running into a disjoint set of issues. What I've been working on [is here](https://github.com/mattwigway/Yggdrasil/tree/osrm-again). The issues I've run into so far:

- macOS SDK doesn't support c++20 - I was having issues even with SDK 11.3, but seems like you have solved that.
- Project-OSRM/osrm-backend#6704 which is caused by a [gcc bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105438), which I worked around by patching in -Wno-array-bounds to disable the (spurious) warning.
- Boost 1.79 wasn't working with gcc 12 and up for me, I have a PR in progress to update Boost to 1.85 (#8888)

jeremiahpslewis:
Hey @mattwigway!

1) you have write access to this branch and should feel free to modify it as necessary. 
2) I added your gcc modification to my `CMakeList.txt` patch
3) Using clang for all builds seems strictly better
4) The remaining errors which I encounter don't seem related to boost versions. They are as follows:

*-apple-*
```bash
[20:49:20] ld64.lld: error: undefined symbol: osrm::extractor::CompressedEdgeContainer::GetBucketReference(unsigned int) const
```

*-mingw-*
```bash
warning "__STRICT_ANSI__ seems to have been undefined; this is not supported"[39m
```

- This is a problem, as the mingw symbols which we need to build are only available when `__STRICT_ANSI__` is undefined. Until Ygg ships a more recent version of mingw, (maybe with the upcoming BinaryBuilder redo) it's probably best to just drop windows.

x86_64-linux-gnu-cxx11
```bash
<head></head>
[ Info: Checking shared library lib/libosrm_extract.so
--
  | ERROR: could not load library "/cache/build/yggy-amdci7-14/julialang/yggdrasil/O/OSRM/build/x86_64-linux-gnu-cxx11/Sjda2mna/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libosrm_extract.so"
  | /cache/julia-buildkite-plugin/julia_installs/bin/linux/x64/1.7/julia-1.7-latest-linux-x86_64/bin/../lib/julia/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /cache/build/yggy-amdci7-14/julialang/yggdrasil/O/OSRM/build/x86_64-linux-gnu-cxx11/Sjda2mna/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib/libosrm_extract.so)
  | ┌ Warning: lib/libosrm_extract.so cannot be dlopen()'ed
```

- dlopen issues, which I wonder if we can solve by setting the appropriate binaries to `dont_dlopen=true`

giordano:
> dlopen issues, which I wonder if we can solve by setting the appropriate binaries to dont_dlopen=true

Ignoring an issue doesn't solve it. You simply can't build a library with a recent compiler and load it with an older runtime. There's no way around it: just don't do it.

jeremiahpslewis:
> > dlopen issues, which I wonder if we can solve by setting the appropriate binaries to dont_dlopen=true
> 
> Ignoring an issue doesn't solve it. You simply can't build a library with a recent compiler and load it with an older runtime. There's no way around it: just don't do it.

Why do I get an older runtime when I set the build gcc to v12?

giordano:
> Why do I get an older runtime when I set the build gcc to v12?

The runtime is used....at run time, not during the build. It's the gcc runtime shipped with [Julia v1.7](https://github.com/JuliaLang/julia/releases/tag/v1.7.0), which was released months before [GCC 12](https://gcc.gnu.org/gcc-12/) was a thing. This is the whole point of the [compatibility section in the `CONTRIBUTING.md` file](https://github.com/JuliaPackaging/Yggdrasil/blob/c3fdd9d985d6ccbce5a529ab306230584707d79b/CONTRIBUTING.md#compatibility-tips).

mattwigway:
So based on the contributing.md document, it seems like Clang doesn't have the same problem with runtime version incompatibilities that gcc does, so maybe that's a solution? Otherwise I'm okay making this only supported on more recent Julia, as it's a fairly specialized package.

giordano:
> it seems like Clang doesn't have the same problem with runtime version incompatibilities that gcc does, so maybe that's a solution?

Wrong, clang on Linux uses gcc runtime.

mattwigway:
FWIW, here are the missing symbols in the Mac build, demangled (with `llvm-cxxfilt`):

```
tbb::detail::r1::initialize(tbb::detail::d1::task_group_context&)
tbb::detail::r1::allocate_memory(unsigned long)
tbb::detail::r1::set_end_of_input(tbb::detail::d1::base_filter&)
tbb::detail::r1::deallocate_memory(void*)
tbb::detail::r1::parallel_pipeline(tbb::detail::d1::task_group_context&, unsigned long, tbb::detail::d1::filter_node const&)
tbb::detail::r1::destroy(tbb::detail::d1::task_group_context&)
osrm::extractor::SuffixTable::isSuffix(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) const
osrm::extractor::RestrictionGraph::GetRestrictions(unsigned long long) const
osrm::extractor::WayRestrictionMap::IsRestricted(unsigned long long, unsigned int) const
osrm::extractor::WayRestrictionMap::IsViaWayEdge(unsigned int, unsigned int) const
osrm::extractor::WayRestrictionMap::GetRestrictions(unsigned long long, unsigned int) const
osrm::extractor::WayRestrictionMap::DuplicatedNodeIDs(unsigned int, unsigned int) const
osrm::extractor::CompressedEdgeContainer::HasEntryForID(unsigned int) const
osrm::extractor::CompressedEdgeContainer::GetBucketReference(unsigned int) const
```

I'm not sure why these are missing, all the osrm:: ones I've checked are defined in both .cpp and .h files. I don't have time tonight, but I'm going to try the build locally and look at the built artifacts to see if anything looks suspicious.

mattwigway:
So, the Mac build failures do appear to be linker problems. All of the `osrm::*` missing symbols are in fact defined in compiled `.o` files. The tbb ones are not, but I wouldn't expect them to be—they should be in `.so` files (I haven't specifically looked for them there). I'm going to see if I can figure out why the linker isn't finding the symbols.

mattwigway:
So, what these symbols have in common is that they're all symbols that come from other dynamic libraries, rather than from `.o` files being linked into the current dynamic library. So I think this is an `rpath` issue.

mattwigway:
Builds are now succeeding on macOS and Linux. We had to set `dont_dlopen` because OSRM can only be built with GCC 12+, and that's not compatible with the Julia 1.7 used in CI. I confirmed that the built artifacts work on Julia 1.8, and `julia_compat` is set to 1.8. Perhaps in a future BinaryBuilder the `dlopen()` test could happen with the julia version specified in compat—I'm happy to take a look at that if desired.

There are a few more steps before merge:
- We're using a significantly different build configuration than stock OSRM (single monolithic dylib, instead of five or six circularly-dependent dylibs), I'm waiting for upstream feedback on whether this will get mainlined (Project-OSRM/osrm-backend#6955)
- To use OSRM.jl from Julia requires a [C++ shim](https://github.com/mattwigway/OSRM.jl/tree/main/cxx). I'm building packages locally and am going to confirm that the C++ shim works with the built packages.

O/OSRM/build_tarballs.jl
```diff
@@ -0,0 +1,145 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "OSRM"
```
stemann, line 5: Overly terse for something specific.

```suggestion
name = "OpenStreetRoutingMachine"
```

O/OSRM/build_tarballs.jl
```diff
@@ -0,0 +1,145 @@
...
+version = v"5.28.0" # UNTAGGED / ASK FOR NEW RELEASE TAG
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/mattwigway/osrm-backend.git", "f16983e668724d4fea1f248cb3e36509e1b5962e"),
```
stemann, line 10: Why not use upstream, https://github.com/Project-OSRM/osrm-backend ?

mattwigway: Actually I take that back, there is still one outstanding PR. But since it's a fairly minor change we can include it as a patch here and build off upstream.

O/OSRM/build_tarballs.jl
stemann: Alternatively, OpenStreetMapOSRM ?

Does OSRM work on something other than OpenStreetMap data?

Does the OSRM use of the term “routing” generalise to (mathematical) graphs, networks etc. or is it geo/street map specific?

Though the source/upstream/official naming might be X, I do not think it automatically justifies this name (X) in the Julia General registry - if X only makes sense in a certain context.

O/OSRM/build_tarballs.jl
stemann: Ah - sounds good!

O/OSRM/build_tarballs.jl
stemann: But OpenSourceRoutingMachine is also sufficiently specific/verbose to be fine :-)

(Side note: Why are GEOS, MEOS, OSRM naming their projects OpenSourceX ?)

stemann: (Side note: Why are GEOS, MEOS, OSRM naming their projects OpenSourceX ?)

imciner2:
> interpreted the endings of the build names (cxx03, and cxx11) correctly, they are conflict with the cxx20 requirement of OSRM.

No, they aren't. These endings refer to the G++ string ABI that was used for the build. Since GCC changed the ABI of `std::string` for c++11, we build with both the old ABI and new ABI so that we can have libraries that interface with both in case it is needed.

mattwigway:
Thanks both for working on this! I'm away from the computer now, but the osrm-again branch of my Yggdrasil fork was building last I checked, not sure if you've looked at those changes.

jeremiahpslewis:
> Thanks both for working on this! I'm away from the computer now, but the osrm-again branch of my Yggdrasil fork was building last I checked, not sure if you've looked at those changes.

I've included your cmake patch.

this version builds on even more platforms, just seeing if I can get mingw to build; the last failures are risc, etc. due to onetbb not being available for the platform

jrklasen:
Does the resent [OSRM 6.0](https://github.com/Project-OSRM/osrm-backend/releases/tag/v6.0.0) release help here by any chance?

.devcontainer/devcontainer.json
```diff
@@ -11,5 +12,5 @@
     "gpu": "optional"
   },
   "privileged": true,
-  "onCreateCommand": "julia -e 'using Pkg; Pkg.precompile();'"
+  "onCreateCommand": "julia -e 'using Pkg; Pkg.precompile();'",
```
fingolfin, line 15: I don't think this file should be modified in this PR

O/OpenSourceRoutingMachine/build_tarballs.jl
```diff
@@ -0,0 +1,157 @@
...
+name = "OpenSourceRoutingMachine"
+version = v"6.0.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/Project-OSRM/osrm-backend.git", "054b0a6395f689a47a908b26b12e32ed7704c533"),
+    DirectorySource("./bundled"),
+    # OSRM requires C++20, which needs a newer SDK
+    ArchiveSource("https://github.com/realjf/MacOSX-SDKs/releases/download/v0.0.1/MacOSX12.3.sdk.tar.xz",
+                  "a511c1cf1ebfe6fe3b8ec005374b9c05e89ac28b3d4eb468873f59800c02b030"),
```
fingolfin, line 14: Should be switched to the `require_macos_sdk` helper, see e.g. fddd43fd1542cb61c57d16dc12305284aa344ce4 for examples how to do that.

PR #8895, opened 2024-06-14
Add `lexbor` build script

MichaelHatherly:


L/lexbor/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/lexbor/lexbor.git", "26a8819b60801301df92d771f12a01dbe8705a7a")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/lexbor
+cmake -B build -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DLEXBOR_BUILD_SEPARATELY=ON -DLEXBOR_BUILD_STATIC=OFF
```
MichaelHatherly, line 16: `-G "Unix Makefile"` comes from their docs for building on Windows: https://lexbor.com/docs/lexbor/#windows. Wasn't required for other platforms but seems to build fine with it added. Could always do it conditionally for Windows only if required.

`-DLEXBOR_BUILD_SEPARATELY=ON` added to get separate libs for each module that they provide.

`-DLEXBOR_BUILD_STATIC=OFF` added since I figured we don't really need those, the dynlibs only is fine?

giordano: I think unix makefiles is the default anyway for us.

L/lexbor/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/lexbor/lexbor.git", "26a8819b60801301df92d771f12a01dbe8705a7a")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/lexbor
+cmake -B build -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DLEXBOR_BUILD_SEPARATELY=ON -DLEXBOR_BUILD_STATIC=OFF
```
giordano, line 16: Just for readability:
```suggestion
cmake -B build \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DLEXBOR_BUILD_SEPARATELY=ON \
    -DLEXBOR_BUILD_STATIC=OFF
```

MichaelHatherly:
Thanks @giordano.

PR #8896, opened 2024-06-14
[pandoc_crossref] Compat change requires version bump

omus:
Follow up to: https://github.com/JuliaPackaging/Yggdrasil/pull/8892#issuecomment-2168402590

### DIFF
--- a/P/pandoc_crossref/utils.jl
+++ b/P/pandoc_crossref/utils.jl
@@ -1,0 +1,27 @@
+# pandoc-crossref appears to be using the Haskell Package Versioning Policy. The forth
+# component of the version number is maintainer specified and for pandoc-crossref is always
+# a digit followed by an optional letter.
+#
+# https://pvp.haskell.org/
+# https://github.com/lierdakil/pandoc-crossref/releases
+function pandoc_crossref_jll_version(v::AbstractString)
+    m = match(r"^(?<major1>\d+)\.(?<major2>\d{1,2})\.(?<minor>\d+)\.(?<patch>\d)(?<build>[a-z])?$", v)
+    if !isnothing(m)
+        if !isnothing(m[:build])
+            build_letter = only(collect(m[:build]))
+            build_num = build_letter - 'a' + 1
+            1 <= build_num <= 9 || throw(ArgumentError("Found unexpected pandoc-crossref build value in: $v"))
+        else
+            build_num = 0
+        end
+
+        # Note: Version 0.3.16.0a occurs after 0.3.16.0
+        return VersionNumber(
+            parse(Int, m[:major1]) * 100 + parse(Int, m[:major2]),
+            parse(Int, m[:minor]),
+            parse(Int, m[:patch]) * 10 + build_num,
+        )
+    else
+        throw(ArgumentError("Unhandled pandoc-crossref version number: $v"))
+    end
+end
--- a/P/pandoc_crossref/utils.jl
+++ b/P/pandoc_crossref/utils.jl
@@ -10,7 +10,6 @@ function pandoc_crossref_jll_version(v::AbstractString)
         if !isnothing(m[:build])
             build_letter = only(collect(m[:build]))
             build_num = build_letter - 'a' + 1
-            1 <= build_num <= 9 || throw(ArgumentError("Found unexpected pandoc-crossref build value in: $v"))
         else
             build_num = 0
         end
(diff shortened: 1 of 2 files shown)


omus:
One alternative I could see is to translate the pandoc-crossref version and include all of the 4th digit in the patch version:

- `v0.3.171a` -> `v0.3.1710`
- `v0.3.16.0e` -> `v0.3.1604`

I don't think the registry tests will like this though as the version number is non-incrementing. There may be a special case for JLLs though.

omus:
Hold off on merging this until we settle on a convention

omus:
I couldn't remember if we could use non-sequential version numbers with JLL packages and the answer is yes:

- https://github.com/JuliaRegistries/RegistryCI.jl/blob/e4bb3ad236f66b626eb6b4ec8bcaf2b3b3e0246d/src/AutoMerge/guidelines.jl#L1105
- https://github.com/JuliaRegistries/RegistryCI.jl/blob/e4bb3ad236f66b626eb6b4ec8bcaf2b3b3e0246d/src/AutoMerge/pull_requests.jl#L191-L192

palday:
Looking at the version history of pandoc-crossref, it looks like it uses zerover with the schema `0.major.minor.patch[build]` where `build` is a letter. So we could promote everything one level and and combine `patch` and `build` into a single entity: `major.minor.julia_patch` where `julia_patch = patch * 100 + sequence_number(build)`.

Here are a few examples:
- 0.3.17.0 -> 3.17.0
- 0.3.17.0f -> 3.17.006 = 3.17.6
- 0.3.17.1 -> 3.17.100
- 0.3.17.1a -> 3.17.101

omus:
Do we have evidence that [pandoc-crossref](https://github.com/lierdakil/pandoc-crossref) is using [0ver](https://0ver.org/)? From this [comment](https://github.com/lierdakil/pandoc-crossref/issues/253#issuecomment-612074120) it seems like the package may be using [Haskell PvP](https://pvp.haskell.org/)

I also found this Haskell release of pandoc-crossref which seems to drop the letter from the version: https://hackage.haskell.org/package/pandoc-crossref

palday:
The letter is the build -- no real source change, it just reflects which version of pandoc it was compiled against, which is why for haskell sources it doesn't show up.

I think Haskell PvP is perhaps the better description. Haskell PvP is a lot like SemVer, but one key difference is that it has `MAJOR.MAJOR.MINOR.PATCH` versioning. It looks pandoc-crossref has locked the first `MAJOR` to 0, which is the core detail that I was assuming.

palday:
To full incorporate the two `MAJOR`s in PvP, the SemVer major version could be `MAJOR[1]*100 + MAJOR[2]` so that even if pandoc-cross ref increments the first major, the scheme still works.

P/pandoc_crossref/utils.jl
```diff
@@ -0,0 +1,27 @@
...
+            1 <= build_num <= 9 || throw(ArgumentError("Found unexpected pandoc-crossref build value in: $v"))
+        else
+            build_num = 0
+        end
+
+        # Note: Version 0.3.16.0a occurs after 0.3.16.0
+        return VersionNumber(
+            parse(Int, m[:major1]) * 100 + parse(Int, m[:major2]),
+            parse(Int, m[:minor]),
+            parse(Int, m[:patch]) * 10 + build_num,
```
giordano, line 22: Is 10 enough?  What if they got past `l`?

omus: Historically the highest they ever have got to is `f`. There is a check in place to ensure that the build number is always a single numeric digit. I was thinking if they do suddenly go past `i` we could update the convention to use `* 100` and the new version would be larger than any version we can have with the current function `9i` is the current max.

omus: I dug more into this and there is a possible collision:
```
julia> pandoc_crossref_jll_version("0.3.17.1")  # `9i` max
v"3.17.10"

julia> pandoc_crossref_jll_version2("0.3.17.0j")  # `9z` max
v"3.17.10"
```

omus:
Updated examples:
```julia
julia> pandoc_crossref_jll_version("0.3.17.0")
v"3.17.0"

julia> pandoc_crossref_jll_version("0.3.17.0f")
v"3.17.6"

julia> pandoc_crossref_jll_version("0.3.17.1")
v"3.17.100"

julia> pandoc_crossref_jll_version("0.3.17.1a")
v"3.17.101"

julia> pandoc_crossref_jll_version("0.3.16.0")
v"3.16.0"

julia> pandoc_crossref_jll_version("0.3.16.0a")
v"3.16.1"

julia> pandoc_crossref_jll_version("1.3.4.5")
v"103.4.500"
```

PR #8911, opened 2024-06-17
Wizard recipe: libgsd-v3.2.1

ywitzky:

* Package name: libgsd
* Version: v3.2.1

### DIFF
--- a/L/libgsd/build_tarballs.jl
+++ b/L/libgsd/build_tarballs.jl
@@ -7,12 +7,11 @@ version = v"3.2.1"
 
 # Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://github.com/glotzerlab/gsd/releases/download/v$(version)/gsd-$(version).tar.gz", "49a0e16276789b6953ece7e5a678bd22c128f8cc57cb967f4228ff6487770995")
+    GitSource("https://github.com/glotzerlab/gsd/releases/download/v$(version)/gsd-$(version).tar.gz", "49a0e16276789b6953ece7e5a678bd22c128f8cc57cb967f4228ff6487770995")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir
 cd ${WORKSPACE}/srcdir/gsd-*
 mkdir -p "${libdir}"
 ${CC} -std=c99 -fPIC gsd/gsd.c -shared -o "${libdir}/libgsd.${dlext}"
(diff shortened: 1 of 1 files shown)


L/libgsd/build_tarballs.jl
```diff
@@ -0,0 +1,51 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "libgsd"
+version = v"3.2.1"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/glotzerlab/gsd/releases/download/v$(version)/gsd-$(version).tar.gz", "49a0e16276789b6953ece7e5a678bd22c128f8cc57cb967f4228ff6487770995")
```
imciner2, line 10: Since this is a git repo, it is better to use a `GitSource` here instead of an `ArchiveSource`.

L/libgsd/build_tarballs.jl
```diff
@@ -0,0 +1,51 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
```
imciner2, line 15:
```suggestion
```
Not needed with the one below it.

L/libgsd/build_tarballs.jl
```diff
@@ -0,0 +1,51 @@
...
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "freebsd"; )
+]
```
imciner2, line 38:
```suggestion
platforms = supported_platforms()
platforms = filter!(!Sys.iswindows, platforms)
```
It is cleaner to filter out the platforms you don't want instead of listing everything when you have a lot of things.

(I think you just wanted to exclude Windows, right? It is also good to leave a comment explaining why it is excluded).

imciner2 (changes requested):

L/libgsd/build_tarballs.jl
ywitzky: Yes that is the idea. All of this has been created running the wizard and i have selected everything but windows. Therefore it might be worth updating this in the wizard as well. But thank you for the feedback.

imciner2 (approved):

PR #8914, opened 2024-06-17
ADOLC: Update to 1.1.1

TimSiebert1:


imciner2:
You didn't bump the version number when changing the git has here.

TimSiebert1:
Good point 😅

PR #8920, opened 2024-06-18
[GR] bump version

jheinen:
bump GR version to 0.73.6


jheinen:
@giordano : if there's anything against merging, pls let me know. Thanks in advance!

PR #8922, opened 2024-06-18
MAGEMin v1.4.9

NicolasRiel:
- general improvement of the accuracy of the phase equilibrium prediction
- corrects issue with spinel in for the metabasite database when O was set to 0.0
- corrects julia function to export to csv when using buffer
- corrects function to compute zircon saturation
- reorganization of the C code to simplify database modification and addition


NicolasRiel:
As always, thank you Mosè!

PR #8925, opened 2024-06-19
[libgsd-3.2.1]  Added header files

ywitzky:
Added FileProduct of gsd.h to use in CBinding.jl

### DIFF
--- a/L/libgsd/build_tarballs.jl
+++ b/L/libgsd/build_tarballs.jl
@@ -27,7 +27,7 @@ platforms = filter!(!Sys.iswindows, platforms)
 
 # The products that we will ensure are always built
 products = [
-    FileProduct("include/gsd.h", :gsd_h), # header file is needed for the use of CBinding.jl in follow up wrapper of GSD.jl
+    FileProduct("include/gsd.h", :gsd_h), 
     LibraryProduct("libgsd", :libgsd)
 ]
 
--- a/L/libgsd/build_tarballs.jl
+++ b/L/libgsd/build_tarballs.jl
@@ -10,11 +10,14 @@ sources = [
     GitSource("https://github.com/glotzerlab/gsd.git", "ad2417d0dbc455d1fb5aab525a919b36bc7f0851")
 ]
 
+
 # Bash recipe for building across all platforms
 script = raw"""
 cd ${WORKSPACE}/srcdir/gsd/
 mkdir -p "${libdir}"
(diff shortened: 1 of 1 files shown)


L/libgsd/build_tarballs.jl
```diff
@@ -14,6 +14,8 @@ sources = [
 script = raw"""
 cd ${WORKSPACE}/srcdir/gsd/
 mkdir -p "${libdir}"
+mkdir -p ${prefix}/share
+cp ./gsd/gsd.h ${prefix}/share/gsd.h
```
imciner2, line 18:
```suggestion
install -D -m644 ./gsd/gsd.h ${includedir}/gsd.h
```
Header files should be in the include directory, not share.

(and using `install -D` will automatically create the path as required).

L/libgsd/build_tarballs.jl
```diff
@@ -26,6 +28,7 @@ platforms = filter!(!Sys.iswindows, platforms)
 
 # The products that we will ensure are always built
 products = [
+    FileProduct("share/gsd.h", :gsd_h), # header file is needed for the use of CBinding.jl in follow up wrapper of GSD.jl
```
imciner2, line 31:
```suggestion
    FileProduct("include/gsd.h", :gsd_h),
```

imciner2 (changes requested):

L/libgsd/build_tarballs.jl
imciner2: @ywitzky I had removed the comment because it isn't necessary. This package is a library, including header files is a normal thing to do in a library and doesn't need an explanation.

ywitzky: @imciner2  Ok thanks. That makes sense. Ill fix this.

ywitzky:
@imciner2 
Can you please tell me why the build fails to register in the buildkite at [https://buildkite.com/julialang/yggdrasil/builds/11377#01903531-0c24-48c8-ba78-69c6fb368303](https://buildkite.com/julialang/yggdrasil/builds/11377#01903531-0c24-48c8-ba78-69c6fb368303) ?

I am unsure whether it is intended that I open a pull request at  https://github.com/JuliaRegistries/General/pull/new/registrator-libgsd_jll-e4d6e05a-v3.2.1+1-64408fdd61 or if there is something wrong with the commit .

imciner2:
No, nothing needs to be done. There was a random failure, and we triggered the registration again, so it is waiting to be merged in the next round in the general registry https://github.com/JuliaRegistries/General/pull/109585.

PR #8928, opened 2024-06-19
[finufft] CUDA components of FINUFFT library

ludvigak:
Creating a separate build process for `cufinufft`, the CUDA version of FINUFFT, in order to separate it from the microarchitecture expansion business of the CPU lib.

### DIFF
--- a/C/cufinufft/build_tarballs.jl
+++ b/C/cufinufft/build_tarballs.jl
@@ -44,7 +44,7 @@ cmake --install .
 """
 
 # Build for all supported CUDA > v11
-platforms = CUDA.supported_platforms(min_version=v"11.0")
+platforms = expand_cxxstring_abis(CUDA.supported_platforms(min_version=v"11.0"))
 # Cmake toolchain breaks on aarch64, so only x86_64 for now
 filter!(p -> p["arch"]=="x86_64", platforms)
 
--- a/C/cufinufft/build_tarballs.jl
+++ b/C/cufinufft/build_tarballs.jl
@@ -43,8 +43,10 @@ cmake --build . --parallel $nproc
 cmake --install .
 """
 
-# Build for all supported CUDA platforms > v11
+# Build for all supported CUDA > v11
 platforms = CUDA.supported_platforms(min_version=v"11.0")
+# Cmake toolchain breaks on aarch64, so only x86_64 for now
+filter!(p -> p["arch"]=="x86_64", platforms)
 
 # The products that we will ensure are always built
 products = [
@@ -74,7 +76,8 @@ for platform in platforms
     build_tarballs(ARGS, name, version, sources, platform_script, [platform],
                    products, [dependencies; cuda_deps];
                    preferred_gcc_version=preferred_gcc_version,
-                   julia_compat="1.9",
-                   augment_platform_block=CUDA.augment
+                   julia_compat="1.6",
+                   augment_platform_block=CUDA.augment,
+                   lazy_artifacts=true
                    )
 end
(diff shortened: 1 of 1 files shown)


ludvigak:
@giordano could you review this?

C/cufinufft/build_tarballs.jl
```diff
@@ -0,0 +1,83 @@
...
+    -DFINUFFT_USE_CUDA=ON \
+    -DCMAKE_CUDA_ARCHITECTURES="${CUDA_ARCHS}"
+cmake --build . --parallel $nproc
+cmake --install .
+"""
+
+# Build for all supported CUDA > v11
+platforms = CUDA.supported_platforms(min_version=v"11.0")
+# Cmake toolchain breaks on aarch64, so only x86_64 for now
+filter!(p -> p["arch"]=="x86_64", platforms)
```
giordano, line 49: Just slightly nicer:
```suggestion
filter!(p -> arch(p)=="x86_64", platforms)
```

C/cufinufft/build_tarballs.jl
```diff
@@ -0,0 +1,83 @@
...
+]
+
+# Build script: cufinufft, all possible archs available for each CUDA version
+# - CMake toolchain looks for compiler in CUDA_PATH/bin/nvcc
+#   and libs in CUDA_PATH/lib64, so create link.
+script = raw"""
+cd $WORKSPACE/srcdir/finufft*/
+
+export CUDA_PATH="$prefix/cuda"
+ln -s $prefix/cuda/lib $prefix/cuda/lib64
```
giordano, line 30: I think you need to delete this symlink at the end, otherwise it ends up with a broken symlink in the tarball

C/cufinufft/build_tarballs.jl
```diff
@@ -0,0 +1,83 @@
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DFINUFFT_USE_CPU=OFF \
+    -DFINUFFT_USE_CUDA=ON \
+    -DCMAKE_CUDA_ARCHITECTURES="${CUDA_ARCHS}"
+cmake --build . --parallel $nproc
+cmake --install .
+"""
+
+# Build for all supported CUDA > v11
+platforms = CUDA.supported_platforms(min_version=v"11.0")
```
giordano, line 47: https://buildkite.com/julialang/yggdrasil/builds/11379#0190358f-17b6-45b3-805d-06dec3f11260/671-979
```suggestion
platforms = expand_cxxstring_abis(CUDA.supported_platforms(min_version=v"11.0"))
```

giordano (commented):
This looks overall good to me, but are you sure you want to keep this a separate package from finufft?

ludvigak:
> This looks overall good to me, but are you sure you want to keep this a separate package from finufft?

Not entirely, but it sort of makes sense because the cpu/gpu codes are completely separated even though they are in one repo. My main reasons were:

- I wasn't sure how to make it work with the different specializations: finufft.so is built for all microarchitecture variants, while cufinufft.so is built for all the CUDA versions. Building for all permutations of these two seems crazy.
- I want the CUDA functionality to be optional in FINUFFT.jl, which I check with `cufinufft_jll.is_available()`. Not sure how this would work if both products were in the same package

giordano:
Uhm, since the products are different I guess this makes sense.

ludvigak:
Ok, let's merge this then so that I can move forward with the development of the CUDA part. If the double maintenance becomes too much of a headache down the line we'll have to reconsider.

PR #8409, opened 2024-04-01
[HiGHSstatic] remove Zlib dependency

odow:
Closes https://github.com/JuliaPackaging/Yggdrasil/issues/8408

Let's just remove this for now. It was added in https://github.com/JuliaPackaging/Yggdrasil/pull/8071, and it's only needed for cases where the user wants to read compressed MPS files.

PR #8932, opened 2024-06-19
[HiGHS] update to v1.7.1

odow:


imciner2:
You'll need to update the dependencies in the shared library script. https://github.com/JuliaPackaging/Yggdrasil/pull/8409 only added CMake to the static library dependencies but also removed it from the common build script, so it needs adding explicitly to the shared library now.

odow:
For now I've just dropped the `apk del cmake` line if we're building the shared libs. Is that okay?

imciner2:
I would have thought adding the dependency was going to be easier - and it also gets you a more recent CMake version ;)

odow:
Okay, updated to use CMake_jll

PR #8933, opened 2024-06-20
New Recipe: strobealign v0.13.0

jakobnissen:


S/strobealign/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/ksahlin/strobealign.git", "11aaa5cbb60eba080e63dbe49b4506c8308b42b9")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/strobealign
+cmake -B build -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS="-msse4.2" -DCMAKE_CXX_FLAGS="-msse4.2"
```
giordano, line 16: Can we set the -msse4.2 flags only for x86_64 and then build for all platforms?

jakobnissen: I don't think it supports non-x86 architechtures: https://github.com/ksahlin/strobealign/issues/2
Unfortunately, the x86 MacOS (which could then be run with Rosetta on ARM macs) build also fails for me for seemingly unrelated reasons, which is why I disabled it (some symbol is not found by GCC). I don't know how to debug that.

PR #8937, opened 2024-06-21
[Uno] add initial build_tarballs.jl

odow:
x-ref: https://github.com/cvanaret/Uno

cc @cvanaret

I haven't tried this locally, so let's see what happens. ... okay, not immediately straight forward. I'll try this locally.

### DIFF
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -4,12 +4,12 @@ using BinaryBuilder, Pkg
 
 name = "Uno"
 
-version = v"1.0.0"
+version = v"1.1.0"
 
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "888114ce7d5e1d208ed8865dc836e4b94017e6a8",
+        "67b2c20ce95dc725ce5bc68a90cc1a95fe94a46a",
     ),
 ]
 
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "bf92ac7e756200c9be5adfb147b9158dd7def219",
+        "3bd224443a905c3d41b058a108db2b382a537ea4",
     ),
 ]
 
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "23056f985b3a9e5f5acc72f881c3c8dca23ac0c7",
+        "e5940ef2f42c8ab4a820d3f1641af820c077899a",
     ),
 ]
 
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -73,5 +73,5 @@ build_tarballs(
     products,
     dependencies;
     julia_compat = "1.9",
-    preferred_gcc_version = v"8.1.0",
+    preferred_gcc_version = v"9.1.0",
 )
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "0513043436e4e6dddc7aac71695aea56f5a32285",
+        "23056f985b3a9e5f5acc72f881c3c8dca23ac0c7",
     ),
 ]
 
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -56,6 +56,7 @@ ${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-ar
 """
 
 platforms = supported_platforms()
+filter!(p -> triplet(p) != "aarch64-unknown-freebsd", platforms)
 platforms = expand_cxxstring_abis(platforms)
 
 products = [
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -18,12 +18,25 @@ cd $WORKSPACE/srcdir/Uno
 mkdir -p build
 cd build
 
+if [[ "${target}" == *mingw* ]]; then
+    LBT=libblastrampoline-5
+else
+    LBT=libblastrampoline
+fi
+
 cmake \
     -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
     -DBLA_VENDOR="libblastrampoline" \
     -DMUMPS_INCLUDE_DIR=${includedir} \
+    -DMETIS_INCLUDE_DIR=${includedir} \
+    -DMUMPS_LIBRARY="${libdir}/libdmumps.${dlext}" \
+    -DMUMPS_COMMON_LIBRARY="${libdir}/libmumps_common.${dlext}" \
+    -DMUMPS_PORD_LIBRARY="${libdir}/libpord.${dlext}" \
+    -DMUMPS_MPISEQ_LIBRARY="${libdir}/libmpiseq.${dlext}" \
+    -DBLAS_LIBRARIES="${libdir}/${LBT}.${dlext}" \
+    -DLAPACK_LIBRARIES="${libdir}/${LBT}.${dlext}" \
     ..
 
 make -j${nproc}
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "67b2c20ce95dc725ce5bc68a90cc1a95fe94a46a",
+        "d91ed34d5e25c3c16ebc34a7c5a67b9c35c2d33f",
     ),
 ]
 
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -22,6 +22,7 @@ cmake \
     -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
+    -DBLA_VENDOR="libblastrampoline" \
     ..
 
 make -j${nproc}
@@ -37,7 +38,14 @@ products = [
 dependencies = [
     Dependency("ASL_jll", compat="0.1.3"),
     Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"), compat="5.4.0"),
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
+    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
+
+    # We need at least 3.29 (Ygg version), or 3.30 upstream version
+    # for LBT support, so always pull the most recent CMake version.
+    HostBuildDependency(PackageSpec(; name="CMake_jll")),
 ]
 
 build_tarballs(
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -78,5 +78,5 @@ build_tarballs(
     products,
     dependencies;
     julia_compat = "1.9",
-    preferred_gcc_version = v"10.1.0",
+    preferred_gcc_version = v"10.2.0",
 )
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -40,7 +40,7 @@ cmake \
     ..
 
 make -j${nproc}
-cp libuno.a ${libdir}/libuno.a
+cp libuno.a ${prefix}/lib/libuno.a
 
 # We need -fPIC to create a shared library
 # $CXX -fPIC -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}"
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -22,7 +22,6 @@ cmake \
     -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
-    -DBLA_VENDOR="libblastrampoline" \
     -DMUMPS_INCLUDE_DIR=${includedir} \
     -DMUMPS_LIBRARY=${libdir} \
     ..
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "f723292439efa4723e1bd509c9e2253d65b49b3e",
+        "8f5a46ef9360e262caa26148ac68fcb3989a5794",
     ),
 ]
 
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -58,7 +58,6 @@ platforms = expand_cxxstring_abis(platforms)
 
 products = [
     ExecutableProduct("uno_ampl", :amplexe),
-    FileProduct("lib/libuno.a", :libuno),
 ]
 
 dependencies = [
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -73,5 +73,5 @@ build_tarballs(
     products,
     dependencies;
     julia_compat = "1.9",
-    preferred_gcc_version = v"9.1.0",
+    preferred_gcc_version = v"10.1.0",
 )
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "7f112e1d7e6cbb6c741cc25a9df76e221c2e894d",
+        "f166e60e428c8dafed6c3ded7df49fe2e41a2c03",
     ),
 ]
 
@@ -51,15 +51,15 @@ cmake \
     ..
 
 make -j${nproc}
-cp libuno.a ${libdir}/libuno.a
 cp uno_ampl${exeext} ${bindir}/uno_ampl${exeext}
-# ${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${OMP} -l${LBT} -ldmumps -lmetis
+${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${OMP} -l${LBT} -ldmumps -lmetis
 """
 
 platforms = supported_platforms()
 platforms = expand_cxxstring_abis(platforms)
 
 products = [
+    LibraryProduct("libuno", :libuno),
     ExecutableProduct("uno_ampl", :amplexe),
 ]
 
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -23,7 +23,7 @@ cmake \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
     -DMUMPS_INCLUDE_DIR=${includedir} \
-    -DMUMPS_LIBRARY=${libdir} \
+    -DMUMPS_LIBRARY=${libdir}/libdmumps.${dlext} \
     ..
 
 make -j${nproc}
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -52,7 +52,6 @@ products = [
     # ExecutableProduct("Uno", :amplexe),
     FileProduct("lib/libuno.a", :libuno),
 ]
-]
 
 dependencies = [
     Dependency(PackageSpec(name="ASL_jll", uuid="ae81ac8f-d209-56e5-92de-9978fef736f9"), compat="0.1.3"),
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -28,7 +28,7 @@ cmake \
     -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
-    -Damplsolver=${includedir} \
+    -Damplsolver=${libdir}/libasl.${dlext} \
     -DBLA_VENDOR="libblastrampoline" \
     -DMUMPS_INCLUDE_DIR=${includedir} \
     -DMETIS_INCLUDE_DIR=${includedir} \
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "f05a6f011ca548588b588ed8fd7782aeb5043def",
+        "f723292439efa4723e1bd509c9e2253d65b49b3e",
     ),
 ]
 
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -4,7 +4,7 @@ using BinaryBuilder, Pkg
 
 name = "Uno"
 
-version = "v1.0.0"
+version = v"1.0.0"
 
 sources = [
     GitSource(
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -40,7 +40,6 @@ cmake \
     ..
 
 make -j${nproc}
-make install
 """
 
 platforms = supported_platforms()
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -1,0 +1,53 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = "v1.0.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "888114ce7d5e1d208ed8865dc836e4b94017e6a8",
+    ),
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir/Uno
+mkdir -p build
+cd build
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    ..
+
+make -j${nproc}
+make install
+"""
+
+platforms = supported_platforms()
+
+products = [
+    ExecutableProduct("Uno", :amplexe),
+]
+
+dependencies = [
+    Dependency("ASL_jll", compat="0.1.3"),
+    Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"), compat="5.4.0"),
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+]
+
+build_tarballs(
+    ARGS,
+    name,
+    version,
+    sources,
+    script,
+    platforms,
+    products,
+    dependencies;
+    julia_compat = "1.9",
+)
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -23,6 +23,8 @@ cmake \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
     -DBLA_VENDOR="libblastrampoline" \
+    -DMUMPS_INCLUDE_DIR=${includedir} \
+    -DMUMPS_LIBRARY=${libdir} \
     ..
 
 make -j${nproc}
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -19,13 +19,21 @@ mkdir -p build
 cd build
 
 if [[ "${target}" == *mingw* ]]; then
-    LBT=libblastrampoline-5
+    LBT=blastrampoline-5
 else
-    LBT=libblastrampoline
+    LBT=blastrampoline
 fi
+
+if [[ "${target}" == *apple* ]] || [[ "${target}" == *freebsd* ]]; then
+    OMP=omp
+else
+    OMP=gomp
+fi
+bob
 
 cmake \
     -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_PREFIX_PATH=${libdir} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
     -Damplsolver=${libdir}/libasl.${dlext} \
@@ -36,36 +44,36 @@ cmake \
     -DMUMPS_COMMON_LIBRARY="${libdir}/libmumps_common.${dlext}" \
     -DMUMPS_PORD_LIBRARY="${libdir}/libpord.${dlext}" \
     -DMUMPS_MPISEQ_LIBRARY="${libdir}/libmpiseq.${dlext}" \
-    -DBLAS_LIBRARIES="${libdir}/${LBT}.${dlext}" \
-    -DLAPACK_LIBRARIES="${libdir}/${LBT}.${dlext}" \
+    -DBLAS_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
+    -DLAPACK_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
     ..
 
 make -j${nproc}
-install -Dvm 644 libuno.a -t ${prefix}/lib
-
-# We need -fPIC to create a shared library
-# $CXX -fPIC -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}"
+cp libuno.a ${libdir}/libuno.a
+cp uno_ampl${exeext} ${bindir}/uno_ampl${exeext}
+# ${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${OMP} -l${LBT} -ldmumps -lmetis
 """
 
 platforms = supported_platforms()
 
 products = [
-    # ExecutableProduct("Uno", :amplexe),
+    ExecutableProduct("uno_ampl", :amplexe),
     FileProduct("lib/libuno.a", :libuno),
 ]
 
 dependencies = [
+    Dependency(PackageSpec(name="METIS_jll", uuid="d00139f3-1899-568f-a2f0-47f597d42d70")),
     Dependency(PackageSpec(name="ASL_jll", uuid="ae81ac8f-d209-56e5-92de-9978fef736f9"), compat="0.1.3"),
     Dependency(PackageSpec(name="MUMPS_seq_jll", uuid="d7ed1dd3-d0ae-5e8e-bfb4-87a502085b8d")),
     Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"), compat="5.4.0"),
-    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
-    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
+    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD systems),
+    # and libgomp from `CompilerSupportLibraries_jll` everywhere else.
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
     Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
 
-    # We need at least 3.29 (Ygg version), or 3.30 upstream version
-    # for LBT support, so always pull the most recent CMake version.
-    HostBuildDependency(PackageSpec(; name="CMake_jll")),
+    # We need at least 3.29 (Ygg version), or 3.30 upstream version for LBT support,
+    # so always pull the most recent CMake version.
+    HostBuildDependency(PackageSpec(name="CMake_jll", uuid="3f4e10e2-61f2-5801-8945-23b9d642d0e6")),
 ]
 
 build_tarballs(
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "d91ed34d5e25c3c16ebc34a7c5a67b9c35c2d33f",
+        "528c2bd9025d6708f094c9bb57cffb40c13a95a7",
     ),
 ]
 
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -40,12 +40,18 @@ cmake \
     ..
 
 make -j${nproc}
+cp libuno.a ${libdir}/libuno.a
+
+# We need -fPIC to create a shared library
+# $CXX -fPIC -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}"
 """
 
 platforms = supported_platforms()
 
 products = [
-    ExecutableProduct("Uno", :amplexe),
+    # ExecutableProduct("Uno", :amplexe),
+    FileProduct("lib/libuno.a", :libuno),
+]
 ]
 
 dependencies = [
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -36,7 +36,8 @@ products = [
 ]
 
 dependencies = [
-    Dependency("ASL_jll", compat="0.1.3"),
+    Dependency(PackageSpec(name="ASL_jll", uuid="ae81ac8f-d209-56e5-92de-9978fef736f9"), compat="0.1.3"),
+    Dependency(PackageSpec(name="MUMPS_seq_jll", uuid="d7ed1dd3-d0ae-5e8e-bfb4-87a502085b8d")),
     Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"), compat="5.4.0"),
     # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
     # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "3bd224443a905c3d41b058a108db2b382a537ea4",
+        "0513043436e4e6dddc7aac71695aea56f5a32285",
     ),
 ]
 
@@ -28,6 +28,7 @@ cmake \
     -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
+    -Damplsolver=${includedir} \
     -DBLA_VENDOR="libblastrampoline" \
     -DMUMPS_INCLUDE_DIR=${includedir} \
     -DMETIS_INCLUDE_DIR=${includedir} \
@@ -40,7 +41,7 @@ cmake \
     ..
 
 make -j${nproc}
-cp libuno.a ${prefix}/lib/libuno.a
+install -Dvm 644 libuno.a -t ${prefix}/lib
 
 # We need -fPIC to create a shared library
 # $CXX -fPIC -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}"
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "539186253be74e309ede0a43579e611832353739",
+        "1f373ce25049e593088bee66b37a9057039ac19a",
     ),
 ]
 
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -54,6 +54,7 @@ cp uno_ampl${exeext} ${bindir}/uno_ampl${exeext}
 """
 
 platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
 
 products = [
     ExecutableProduct("uno_ampl", :amplexe),
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -9,7 +9,7 @@ version = v"1.1.0"
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "528c2bd9025d6708f094c9bb57cffb40c13a95a7",
+        "bf92ac7e756200c9be5adfb147b9158dd7def219",
     ),
 ]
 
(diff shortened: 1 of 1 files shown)


odow:
@amontoison here's one for you

amontoison:
@cvanaret
Charlie, do you have an option in your build system to provide the name of BLAS / LAPACK libraries?
It seems that you hardcoded `blas` and `lapack`.

I saw that you can use `ma57` in `UNO`.
Can you dynamically load the library like Ipopt or do you need the symbols of `ma57` at compilation time (like GALAHAD)?
I worked with the HSL team to easily use HSL packages in the Julia ecosystem [libHSL](https://licences.stfc.ac.uk/product/libhsl).

odow:
It looks like it expects `find_library` to work for these exact strings:
https://github.com/cvanaret/Uno/blob/5d37b660479e267a3d99c607ff69858b0443ce64/CMakeLists.txt#L82
so we likely need some changes to the CMake to let us inject `libblastrampoline` and to be able to point to `${includedir}` and `${libdir}/libasl` for ASL.

imciner2:
> so we likely need some changes to the CMake to let us inject libblastrampoline

I don't think so. This looks to just call `find_package(blas)` and `find_package(lapack)`, so you should be able to use the most recent `CMake_jll` package and the CMake variable `-DBLA_VENDOR="liblbastrampoline"` and it will use our libblastrampoline, no changes to their CMake required.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+    ..
+
+make -j${nproc}
+make install
+"""
+
+platforms = supported_platforms()
+
+products = [
+    ExecutableProduct("Uno", :amplexe),
```
imciner2, line 34: IMO, it would be better to name this similar to the actual executable name instead of calling it `amplexe`.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir/Uno
+mkdir -p build
+cd build
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    ..
```
imciner2, line 25:
```suggestion
    -DBLA_VENDOR="libblastrampoline" \
    ..
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+platforms = supported_platforms()
+
+products = [
+    ExecutableProduct("Uno", :amplexe),
+]
+
+dependencies = [
+    Dependency("ASL_jll", compat="0.1.3"),
+    Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"), compat="5.4.0"),
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
```
imciner2, line 40:
```suggestion
    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),

```

cvanaret:
Hi,
Thanks a lot for your time!
Yes, I'm calling `find_library` in my CMake for blas and lapack (these are required by the MA57 library). I need the MA57 symbols at compilation.
My compilation expertise is very limited but please do let me know if I can be of any help.
Best,
Charlie

amontoison:
@cvanaret an you give me the list of symbols of MA57 used by UNO and how we can provide a library libhsl.$dlext to UNO?
It seems that you only want a libma57.s$dlext.

cvanaret:
@amontoison the MA57 symbols that Uno needs are:
```
ma57id_
ma57ad_
ma57bd_
ma57cd_
ma57dd_
```
I didn't understand the second part of your question, I need to do some research there :)

amontoison:
> @amontoison the MA57 symbols that Uno needs are:
> 
> ```
> ma57id_
> ma57ad_
> ma57bd_
> ma57cd_
> ma57dd_
> ```
> 
> I didn't understand the second part of your question, I need to do some research there :)

I m wondering how to specify the library name thay contains these symbols. It seems that you always want libma57.so or libma57.a.
In our case we want to use libhsl.so.

U/Uno/build_tarballs.jl
odow: We've used `amplexe` as the convention for all the other AMPL-compatible solvers:
https://github.com/jump-dev/AmplNLWriter.jl/tree/master?tab=readme-ov-file#precompiled-binaries

odow:
We still get:

image

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+    ..
+
+make -j${nproc}
+make install
+"""
+
+platforms = supported_platforms()
+
+products = [
+    ExecutableProduct("Uno", :amplexe),
```
odow, line 35:
```suggestion
    ExecutableProduct("uno_ampl", :amplexe),
```

cvanaret:
@amontoison can't you simply use `cmake -Dma57=path_to_libhsl.so ..`?

imciner2:
Ah, I see what is happening with the BLAS/LAPACK CMake. Upstream is using `find_library`, which is searching for a library named `blas` or `libblas`, and not using the built-in detection logic that CMake provides to find the various BLAS/LAPACK libraries.

Ideally, it should use `find_package(BLAS)` and `find_package(LAPACK)` to find the BLAS and LAPACK libraries, since those can handle the various vendor platforms and are more configurable.

cvanaret:
Absolutely, I need to edit my CMake file.

U/Uno/build_tarballs.jl
fingolfin: Then I would suggest adding this reason as a comment in this file (on this line or next to it).

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+]
+
+dependencies = [
+    Dependency("ASL_jll", compat="0.1.3"),
```
amontoison, line 39:
```suggestion
    Dependency(PackageSpec(name="ASL_jll", uuid="ae81ac8f-d209-56e5-92de-9978fef736f9"), compat="0.1.3"),
    Dependency(PackageSpec(name="MUMPS_seq_jll", uuid="d7ed1dd3-d0ae-5e8e-bfb4-87a502085b8d")),
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,62 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir/Uno
+mkdir -p build
+cd build
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBLA_VENDOR="libblastrampoline" \
```
amontoison, line 25:
```suggestion
    -DBLA_VENDOR="libblastrampoline" \
    -DMUMPS_INCLUDE_DIR=${includedir} \
    -DMUMPS_LIBRARY=${libdir} \
```

amontoison:
@cvanaret Is it normal that `Uno` requires `MPI`?

cvanaret:
> @cvanaret Is it normal that `Uno` requires `MPI`?

Hey Alexis. Uno requires MPI only when parallel MUMPS is being linked.

amontoison:
@cvanaret How do you specify that we link Uno with a sequential version of MUMPS?
I have this error with a sequential MUMPS.

cvanaret:
@amontoison that's a very good question. The MUMPS team suggested I test the existence of the dummy library `libmpiseq.a`: if it exists, assume MUMPS was compiled in sequential mode, otherwise it was compiled in parallel mode and you should link MPI and co.

How I got mine:
- in MUMPS' Makefile.inc, the last three lines should read:
```
INCS = $(INCSEQ)
LIBS = $(LIBSEQ)
LIBSEQNEEDED = libseqneeded
```
- after compilation, `libmpiseq.a` lies in `/MUMPS_directory/libseq/`.

It should then be detected by Uno's `FindMUMPS.cmake`. Hope that helps!

amontoison:
Ok, I see but Patrick and Jean-Yves are big fan of static libraries and probably forgot to specify that we can also compile MUMPS to have shared libraries. It means that we can also have libmpiseq.$dlext where $dlext is `so` (Linux / FreeBSD) or `dll` (Windows) or `dylib` (Mac).

cvanaret:
@amontoison is there a quick fix I can use in my CMake or FindMUMPS.cmake to help?

amontoison:
An easy fix would be to detect if you if you have one these four files:
- libmpiseq.a
- libmpiseq.so
- libmpiseq.dylib
- libmpiseq.dll
We have all these libraries in the same folder when we install `MUMPS_seq` here:
Capture d’écran du 2024-10-13 15-55-46

cvanaret:
@amontoison thanks, it's [pushed](https://github.com/cvanaret/Uno/commit/d91ed34d5e25c3c16ebc34a7c5a67b9c35c2d33f) :)

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "67b2c20ce95dc725ce5bc68a90cc1a95fe94a46a",
```
odow, line 12:
```suggestion
        "d91ed34d5e25c3c16ebc34a7c5a67b9c35c2d33f",
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+]
+
+dependencies = [
+    Dependency(PackageSpec(name="ASL_jll", uuid="ae81ac8f-d209-56e5-92de-9978fef736f9"), compat="0.1.3"),
+    Dependency(PackageSpec(name="MUMPS_seq_jll", uuid="d7ed1dd3-d0ae-5e8e-bfb4-87a502085b8d")),
+    Dependency(PackageSpec(name="libblastrampoline_jll", uuid="8e850b90-86db-534c-a0d3-1478176c7d93"), compat="5.4.0"),
+    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
+    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
+    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
```
fingolfin, line 47: Does this work as intended? The final `Project.toml` can only have a single fixed list of dependencies, not different ones for BSD vs. non-BSD systems.

odow: IIUC: both get added to Project.toml, but the JLL init script will load different things on different platforms?

odow:
Failure is because we're still using `find_library` for BLAS etc:
https://github.com/cvanaret/Uno/blob/d91ed34d5e25c3c16ebc34a7c5a67b9c35c2d33f/CMakeLists.txt#L149-L155

amontoison:
Thanks Charlie! It fixed the issue. 👍 👍 
We have this error now:
```
Could not find blas using the following names: blas
```
I suppose that `Uno` wants to find a libblas.a / libblas.$dlext and a liblapack.a / liblapack.$dlext.
Is it possible to specify the name of the BLAS and LAPACK libraries?
We use `libblastrampoline.so`, `libblas libbllibblalibblastrampoline.dylib` and `libblastrampoline-5.dll` in Julia.
The simplest fix could be to add an entry in your findBLAS.cmake.

U/Uno/build_tarballs.jl
imciner2: This is the standard way of representing this. Inside the init scripts of a package, it then only loads the required ones (e.g., see the SZ_jll scripts for macOS here: https://github.com/JuliaBinaryWrappers/SZ_jll.jl/blob/main/src/wrappers/aarch64-apple-darwin.jl and for Linux here: https://github.com/JuliaBinaryWrappers/SZ_jll.jl/blob/main/src/wrappers/x86_64-linux-gnu.jl).

imciner2:
> I suppose that `Uno` wants to find a libblas.a / libblas.$dlext and a liblapack.a / liblapack.$dlext.
> Is it possible to specify the name of the BLAS and LAPACK libraries?
> We use `libblastrampoline.so`, `libblas libbllibblalibblastrampoline.dylib` and `libblastrampoline-5.dll` in Julia.
> The simplest fix could be to add an entry in your findBLAS.cmake.

@amontoison the simplest fix is to use `find_package` to use the built-in CMake logic, see above comment https://github.com/JuliaPackaging/Yggdrasil/pull/8937#issuecomment-2186594297.

amontoison:
I agree with you @imciner2 but Charlie didn't modify the build system of Uno to use yet.

I quickly proposed this temporary hotfix:
https://github.com/cvanaret/Uno/pull/16

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "d91ed34d5e25c3c16ebc34a7c5a67b9c35c2d33f",
```
amontoison, line 12:
```suggestion
        "528c2bd9025d6708f094c9bb57cffb40c13a95a7",
```

cvanaret:
The [latest commit](https://github.com/cvanaret/Uno/commit/4bdc9a4b84b5da345616aff990572d23297eb8f4) uses `find_package` for BLAS and LAPACK.
My apologies for the slow and painful iterations, CMake files are not my strong suit...

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir/Uno
+mkdir -p build
+cd build
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBLA_VENDOR="libblastrampoline" \
```
amontoison, line 25:
```suggestion
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+cd $WORKSPACE/srcdir/Uno
+mkdir -p build
+cd build
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DMUMPS_INCLUDE_DIR=${includedir} \
+    -DMUMPS_LIBRARY=${libdir} \
```
amontoison, line 26:
```suggestion
    -DMUMPS_LIBRARY=${libdir}/libdmumps.${dlext} \
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+
+script = raw"""
+cd $WORKSPACE/srcdir/Uno
+mkdir -p build
+cd build
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
```
amontoison, line 24:
```suggestion
    -DCMAKE_BUILD_TYPE=Release \
    -DBLA_VENDOR="libblastrampoline" \
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "528c2bd9025d6708f094c9bb57cffb40c13a95a7",
```
amontoison, line 12:
```suggestion
        "bf92ac7e756200c9be5adfb147b9158dd7def219",
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+cd $WORKSPACE/srcdir/Uno
+mkdir -p build
+cd build
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DMUMPS_INCLUDE_DIR=${includedir} \
+    -DMUMPS_LIBRARY=${libdir}/libdmumps.${dlext} \
```
amontoison, line 26:
```suggestion
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+cd $WORKSPACE/srcdir/Uno
+mkdir -p build
+cd build
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBLA_VENDOR="libblastrampoline" \
+    -DMUMPS_INCLUDE_DIR=${includedir} \
```
amontoison, line 26:
```suggestion
    -DMUMPS_INCLUDE_DIR=${includedir} \
    -DMETIS_INCLUDE_DIR=${includedir} \
    -DMUMPS_LIBRARY="${libdir}/libdmumps.${dlext}" \
    -DMUMPS_COMMON_LIBRARY="${libdir}/libmumps_common.${dlext}" \
    -DMUMPS_PORD_LIBRARY="${libdir}/libpord.${dlext}" \
    -DMUMPS_MPISEQ_LIBRARY="${libdir}/libmpiseq.${dlext}" \
    -DBLAS_LIBRARIES="${libdir}/${LBT}.${dlext}" \
    -DLAPACK_LIBRARIES="${libdir}/${LBT}.${dlext}" \
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "bf92ac7e756200c9be5adfb147b9158dd7def219",
+    ),
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir/Uno
+mkdir -p build
+cd build
```
amontoison, line 19:
```suggestion
cd build

if [[ "${target}" == *mingw* ]]; then
    LBT=libblastrampoline-5
else
    LBT=libblastrampoline
fi
```

amontoison:
"We're making progress; we've successfully detected all the dependencies. 🐌
The errors are now in the C++ code. Charlie, could you take a look?
https://buildkite.com/julialang/yggdrasil/builds/13883#019287f8-0cb8-4a23-8277-a7918d7403ee/678-985

The error is:
```
[22:21:54] In file included from /workspace/srcdir/Uno/uno/ingredients/globalization_mechanisms/GlobalizationMechanism.hpp:7:0,
[22:21:54]                  from /workspace/srcdir/Uno/uno/ingredients/globalization_mechanisms/GlobalizationMechanism.cpp:4:
[22:21:54] /workspace/srcdir/Uno/uno/ingredients/subproblems/Direction.hpp:8:20: fatal error: optional: No such file or directory
[22:21:54]  #include <optional>
```

cvanaret:
Indeed! You should be covered with GCC >= 8.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,76 @@
...
+build_tarballs(
+    ARGS,
+    name,
+    version,
+    sources,
+    script,
+    platforms,
+    products,
+    dependencies;
+    julia_compat = "1.9",
```
amontoison, line 75:
```suggestion
    julia_compat = "1.9",
    preferred_gcc_version = v"8.1.0",
```

giordano:
However there are loads of errors also on [aarch64-darwin](https://buildkite.com/julialang/yggdrasil/builds/13883#019287f8-0cd1-4db0-9b80-29204500b382/668-978) where we use GCC 18:
```
[22:23:03] /workspace/srcdir/Uno/uno/ingredients/globalization_strategies/GlobalizationStrategyFactory.cpp:14:25: error: invalid operands to binary expression ('const std::string' (aka 'const basic_string<char, char_traits<char>, allocator<char>>') and 'const char[9]')
[22:23:03]    14 |       if (strategy_type == "l1_merit") {
[22:23:03]       |           ~~~~~~~~~~~~~ ^  ~~~~~~~~~~
[...]
[22:23:03] /workspace/srcdir/Uno/uno/ingredients/globalization_strategies/GlobalizationStrategyFactory.cpp:26:60: error: invalid operands to binary expression ('const char[23]' and 'const std::string' (aka 'const basic_string<char, char_traits<char>, allocator<char>>'))
[22:23:03]    26 |       throw std::invalid_argument("GlobalizationStrategy " + strategy_type + " is not supported");
[22:23:03]       |                                   ~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~
[...]
[22:23:03] /workspace/srcdir/Uno/uno/ingredients/globalization_strategies/GlobalizationStrategyFactory.cpp:30:14: error: no matching constructor for initialization of 'std::vector<std::string>' (aka 'vector<basic_string<char, char_traits<char>, allocator<char>>>')
[22:23:03]    30 |       return {"l1_merit", "fletcher_filter_method", "waechter_filter_method", "funnel_method"};
[22:23:03]       |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]
[22:23:04] /workspace/srcdir/Uno/uno/optimization/Multipliers.hpp:38:26: error: call to 'abs' is ambiguous
[22:23:04]    38 |          if (tolerance < std::abs(multiplier_j)) {
[22:23:04]       |                          ^~~~~~~~
[...]
[22:23:04] /workspace/srcdir/Uno/uno/linear_algebra/Norm.hpp:110:21: error: no member named 'max' in namespace 'std'; did you mean 'fmax'?
[22:23:04]   110 |       result = std::max(result, std::abs(element));
[22:23:04]       |                ~~~~~^~~
[22:23:04]       |                     fmax
[...]
```

cvanaret:
@amontoison can you try with GCC 9 or 10?
@giordano I'm not familiar with other compilers so I did a quick search. Can you try these flags?
```
-std=c++17 -stdlib=libc++
```

giordano:
The compiler call looks like https://buildkite.com/julialang/yggdrasil/builds/13883#019287f8-0cd1-4db0-9b80-29204500b382/668-973
```
[22:23:02] /opt/bin/aarch64-apple-darwin20-libgfortran5-cxx11/aarch64-apple-darwin20-clang++ --sysroot=/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root  -I/workspace/srcdir/Uno/uno -I/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root/usr/local/lib -I/workspace/destdir/lib -Wall -Wextra -Wnon-virtual-dtor -pedantic -Wunused-value -Wconversion -Wmaybe-uninitialized -O3 -DNDEBUG -mmacosx-version-min=11.0   -D HAS_METIS -D MUMPS_SEQUENTIAL -D HAS_MUMPS_MPISEQ_LIBRARY -D HAS_MPI -D HAS_MUMPS -fopenmp=libomp -std=gnu++17 -MD -MT CMakeFiles/uno.dir/uno/ingredients/globalization_strategies/switching_methods/filter_methods/FletcherFilterMethod.cpp.o -MF CMakeFiles/uno.dir/uno/ingredients/globalization_strategies/switching_methods/filter_methods/FletcherFilterMethod.cpp.o.d -o CMakeFiles/uno.dir/uno/ingredients/globalization_strategies/switching_methods/filter_methods/FletcherFilterMethod.cpp.o -c /workspace/srcdir/Uno/uno/ingredients/globalization_strategies/switching_methods/filter_methods/FletcherFilterMethod.cpp
```
`-std=c++17` is already there, and we already link to libc++

cvanaret:
@giordano mh I see, thanks. Any luck replacing `-std=gnu++17` with `-std=c++17`?

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,77 @@
...
+    preferred_gcc_version = v"8.1.0",
```
amontoison, line 76:
```suggestion
    preferred_gcc_version = v"9.1.0",
```

giordano:
> Any luck replacing `-std=gnu++17` with `-std=c++17`?

The standard is coming from https://github.com/cvanaret/Uno/blob/3bd224443a905c3d41b058a108db2b382a537ea4/CMakeLists.txt#L13, and [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD) does the automatic "promotion" to the gnu extension if supported.  I'm not sure that's the issue here though.

Also, I'm very much not an expert in C++, but isn't [`std::abs(double)`](https://en.cppreference.com/w/cpp/numeric/math/abs) wrong and it should be [`std::fabs`](https://en.cppreference.com/w/cpp/numeric/math/fabs)?

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,77 @@
...
+    ARGS,
+    name,
+    version,
+    sources,
+    script,
+    platforms,
+    products,
+    dependencies;
+    julia_compat = "1.9",
+    preferred_gcc_version = v"9.1.0",
```
amontoison, line 76:
```suggestion
    preferred_gcc_version = v"10.1.0",
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,77 @@
...
+    -DMUMPS_LIBRARY="${libdir}/libdmumps.${dlext}" \
+    -DMUMPS_COMMON_LIBRARY="${libdir}/libmumps_common.${dlext}" \
+    -DMUMPS_PORD_LIBRARY="${libdir}/libpord.${dlext}" \
+    -DMUMPS_MPISEQ_LIBRARY="${libdir}/libmpiseq.${dlext}" \
+    -DBLAS_LIBRARIES="${libdir}/${LBT}.${dlext}" \
+    -DLAPACK_LIBRARIES="${libdir}/${LBT}.${dlext}" \
+    ..
+
+make -j${nproc}
+make install
```
amontoison, line 43:
```suggestion
```

imciner2:
> Also, I'm very much not an expert in C++, but isn't [`std::abs(double)`](https://en.cppreference.com/w/cpp/numeric/math/abs) wrong and it should be [`std::fabs`](https://en.cppreference.com/w/cpp/numeric/math/fabs)?

It's actually valid because of overloading: https://en.cppreference.com/w/cpp/numeric/math/fabs

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,76 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "bf92ac7e756200c9be5adfb147b9158dd7def219",
```
amontoison, line 12:
```suggestion
        "3bd224443a905c3d41b058a108db2b382a537ea4",
```

imciner2:
For these errors:
```
[14:27:01] /workspace/srcdir/Uno/uno/ingredients/globalization_strategies/GlobalizationStrategyFactory.cpp:14:25: error: invalid operands to binary expression ('const std::string' (aka 'const basic_string<char, char_traits<char>, allocator<char>>') and 'const char[9]')
[14:27:01]    14 |       if (strategy_type == "l1_merit") {
[14:27:01]       |           ~~~~~~~~~~~~~ ^  ~~~~~~~~~~
```

I think you need to add `#include <string>` to the file explicitly. Some compilers might transitively include it, while others don't. This type of comparison is defined in that header, so it should be included.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,76 @@
...
+    -DMETIS_INCLUDE_DIR=${includedir} \
+    -DMUMPS_LIBRARY="${libdir}/libdmumps.${dlext}" \
+    -DMUMPS_COMMON_LIBRARY="${libdir}/libmumps_common.${dlext}" \
+    -DMUMPS_PORD_LIBRARY="${libdir}/libpord.${dlext}" \
+    -DMUMPS_MPISEQ_LIBRARY="${libdir}/libmpiseq.${dlext}" \
+    -DBLAS_LIBRARIES="${libdir}/${LBT}.${dlext}" \
+    -DLAPACK_LIBRARIES="${libdir}/${LBT}.${dlext}" \
+    ..
+
+make -j${nproc}
```
amontoison, line 42:
```suggestion
make -j${nproc}
cp libuno.a ${libdir}/libuno.a

# We need -fPIC to create a shared library
# $CXX -fPIC -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}"
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,76 @@
...
+"""
+
+platforms = supported_platforms()
+
+products = [
+    ExecutableProduct("Uno", :amplexe),
```
amontoison, line 48:
```suggestion
    # ExecutableProduct("Uno", :amplexe),
    FileProduct("lib/libuno.a", :libuno),
]
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+    ARGS,
+    name,
+    version,
+    sources,
+    script,
+    platforms,
+    products,
+    dependencies;
+    julia_compat = "1.9",
+    preferred_gcc_version = v"10.1.0",
```
amontoison, line 81:
```suggestion
    preferred_gcc_version = v"10.2.0",
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+# We need -fPIC to create a shared library
+# $CXX -fPIC -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}"
+"""
+
+platforms = supported_platforms()
+
+products = [
+    # ExecutableProduct("Uno", :amplexe),
+    FileProduct("lib/libuno.a", :libuno),
+]
```
amontoison, line 54:
```suggestion
```

amontoison:
@odow @cvanaret  
The current status is that we can compile a static library `libuno.a` on all platforms where we use `g++`.  
However, it's not working on macOS or FreeBSD due to compilation errors with `clang++`.  

For the future, I have added a command to convert `libuno.a` into a shared library, so we will already have what we need if we decide to create `Uno.jl` one day (which would first require a C interface for Uno).  
Currently, this is not working because the code needs to be compiled with the `-fPIC` option.  

For now, we just want a binary to use `Uno_jll.jl` within [AMPLNLWriter.jl](https://github.com/jump-dev/AmplNLWriter.jl). To achieve this, we need to detect the AMPL library. The problem likely lies [here](https://github.com/cvanaret/Uno/blob/main/CMakeLists.txt#L169) because our library is named `libasl.${dlext}`, but even after adding an entry for `asl`, CMake is still unable to detect it, and as a result, it doesn't generate the binary we need.  

If anyone knows a CMake option to provide an additional folder where CMake should search for the library, it would be helpful.  

Regarding the support for HSL solvers (currently just MA57), I propose waiting until there is a C interface for Uno. At that point, I will add the necessary symbols to the "dummy" `HSL_jll.jl` to prevent the linker from complaining when linking `libhsl.so`. But let's take one step at a time.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,81 @@
...
+    -DMUMPS_LIBRARY="${libdir}/libdmumps.${dlext}" \
+    -DMUMPS_COMMON_LIBRARY="${libdir}/libmumps_common.${dlext}" \
+    -DMUMPS_PORD_LIBRARY="${libdir}/libpord.${dlext}" \
+    -DMUMPS_MPISEQ_LIBRARY="${libdir}/libmpiseq.${dlext}" \
+    -DBLAS_LIBRARIES="${libdir}/${LBT}.${dlext}" \
+    -DLAPACK_LIBRARIES="${libdir}/${LBT}.${dlext}" \
+    ..
+
+make -j${nproc}
+cp libuno.a ${prefix}/lib/libuno.a
```
giordano, line 43: Do you actually need to ship the static archive?  In any case, please always use verbose commands, they're always helpful in figuring out what actually happened, something like
```suggestion
install -Dvm 644 libuno.a -t ${prefix}/lib
```
but I'm not convinced we need the static library at all

amontoison: No, we don't need the static library, but it's the only `Product` that we can generate right now.  
I added it in the products so that we can see which platforms are still failing.

Uno doesn't support `make install`, so I copied the library by hand.  
@cvanaret It would be great to support `make install` so that CMake automatically installs the libraries, headers, and binaries in the prefix folder provided by the user with `-DCMAKE_INSTALL_PREFIX=${prefix}`.

cvanaret:
> @odow @cvanaret The current status is that we can compile a static library `libuno.a` on all platforms where we use `g++`. However, it's not working on macOS or FreeBSD due to compilation errors with `clang++`.

@amontoison great news! I will download clang and noodle around with it - hopefully I can figure out what happens.

> For the future, I have added a command to convert `libuno.a` into a shared library, so we will already have what we need if we decide to create `Uno.jl` one day (which would first require a C interface for Uno). Currently, this is not working because the code needs to be compiled with the `-fPIC` option.

The `-fPIC` option is now [default](https://github.com/cvanaret/Uno/commit/0513043436e4e6dddc7aac71695aea56f5a32285).

> For these errors:
> 
> ```
> [14:27:01] /workspace/srcdir/Uno/uno/ingredients/globalization_strategies/GlobalizationStrategyFactory.cpp:14:25: error: invalid operands to binary expression ('const std::string' (aka 'const basic_string<char, char_traits<char>, allocator<char>>') and 'const char[9]')
> [14:27:01]    14 |       if (strategy_type == "l1_merit") {
> [14:27:01]       |           ~~~~~~~~~~~~~ ^  ~~~~~~~~~~
> ```
> 
> I think you need to add `#include <string>` to the file explicitly. Some compilers might transitively include it, while others don't. This type of comparison is defined in that header, so it should be included.

@imciner2 thanks for the tip :)

> Uno doesn't support `make install`, so I copied the library by hand. @cvanaret It would be great to support `make install` so that CMake automatically installs the libraries, headers, and binaries in the prefix folder provided by the user with `-DCMAKE_INSTALL_PREFIX=${prefix}`.

@amontoison I'll look into it!

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,81 @@
...
+if [[ "${target}" == *mingw* ]]; then
+    LBT=libblastrampoline-5
+else
+    LBT=libblastrampoline
+fi
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
```
odow, line 30:
```suggestion
    -DCMAKE_BUILD_TYPE=Release \
    -Damplsolver=${includedir} \
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,81 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "3bd224443a905c3d41b058a108db2b382a537ea4",
```
odow, line 12:
```suggestion
        "0513043436e4e6dddc7aac71695aea56f5a32285",
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,82 @@
...
+    LBT=libblastrampoline-5
+else
+    LBT=libblastrampoline
+fi
+
+cmake \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -Damplsolver=${includedir} \
```
amontoison, line 31:
```suggestion
    -Damplsolver=${libdir}/libasl.${dlext} \
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,90 @@
...
+else
+    LBT=blastrampoline
+fi
+
+if [[ "${target}" == *apple* ]] || [[ "${target}" == *freebsd* ]]; then
+    OMP=omp
+else
+    OMP=gomp
+fi
+bob
```
odow, line 32:
```suggestion
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,89 @@
...
+    -DLAPACK_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
+    ..
+
+make -j${nproc}
+cp libuno.a ${libdir}/libuno.a
+cp uno_ampl${exeext} ${bindir}/uno_ampl${exeext}
+# ${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${OMP} -l${LBT} -ldmumps -lmetis
+"""
+
+platforms = supported_platforms()
```
odow, line 56:
```suggestion
platforms = supported_platforms()
platforms = expand_cxxstring_abis(platforms)
```

odow:
@cvanaret there are few issues related to `int` and `long int` on 32-bit:

image

cvanaret:
> @cvanaret there are few issues related to `int` and `long int` on 32-bit:

This should be fixed in commit [9f5b0849ce0925bc19f964b169e90d50fc07b6f7](https://github.com/cvanaret/Uno/commit/9f5b0849ce0925bc19f964b169e90d50fc07b6f7).

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,90 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "0513043436e4e6dddc7aac71695aea56f5a32285",
```
amontoison, line 12:
```suggestion
        "23056f985b3a9e5f5acc72f881c3c8dca23ac0c7",
```

cvanaret:
@amontoison @odow @imciner2 @giordano 
I have fixed template inference errors and have managed to compile Uno with Clang 10. The latest commit is [e5940ef2f42c8ab4a820d3f1641af820c077899a](https://github.com/cvanaret/Uno/commit/e5940ef2f42c8ab4a820d3f1641af820c077899a).

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,90 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "23056f985b3a9e5f5acc72f881c3c8dca23ac0c7",
```
odow, line 12:
```suggestion
        "e5940ef2f42c8ab4a820d3f1641af820c077899a",
```

odow:
@cvanaret there are still a couple of `fint`/`int` problems. You might want to consider adding an `x86` build to your CI?

cvanaret:
> @cvanaret there are still a couple of `fint`/`int` problems. You might want to consider adding an `x86` build to your CI?

Definitely... They were flying under the radar, my bad!
It's fixed in commit [7f112e1d7e6cbb6c741cc25a9df76e221c2e894d](https://github.com/cvanaret/Uno/commit/7f112e1d7e6cbb6c741cc25a9df76e221c2e894d).

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,90 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "e5940ef2f42c8ab4a820d3f1641af820c077899a",
```
odow, line 12:
```suggestion
        "7f112e1d7e6cbb6c741cc25a9df76e221c2e894d",
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,90 @@
...
+cp uno_ampl${exeext} ${bindir}/uno_ampl${exeext}
+# ${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${OMP} -l${LBT} -ldmumps -lmetis
+"""
+
+platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
+
+products = [
+    ExecutableProduct("uno_ampl", :amplexe),
+    FileProduct("lib/libuno.a", :libuno),
```
odow, line 61:
```suggestion
```

odow:
@cvanaret more progress. We're getting somewhere.

There are still a bunch of issues like
```
[00:14:53] In file included from /workspace/srcdir/Uno/uno/ingredients/constraint_relaxation_strategies/ConstraintRelaxationStrategy.hpp:9:
[00:14:53] /workspace/srcdir/Uno/uno/linear_algebra/Norm.hpp:16:23: warning: result of comparison against a string literal is unspecified (use an explicit string comparison function instead) [-Wstring-compare]
[00:14:53]    16 |       if (norm_string == "L1") {
[00:14:53]       |                       ^  ~~~~
```
See https://github.com/JuliaPackaging/Yggdrasil/pull/8937#issuecomment-2411469857

Then there's the `abs` `fmax` issue, which I'm not much help for. This is outside my wheelhouse.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "7f112e1d7e6cbb6c741cc25a9df76e221c2e894d",
```
amontoison, line 12:
```suggestion
        "f166e60e428c8dafed6c3ded7df49fe2e41a2c03",
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+    -DMUMPS_LIBRARY="${libdir}/libdmumps.${dlext}" \
+    -DMUMPS_COMMON_LIBRARY="${libdir}/libmumps_common.${dlext}" \
+    -DMUMPS_PORD_LIBRARY="${libdir}/libpord.${dlext}" \
+    -DMUMPS_MPISEQ_LIBRARY="${libdir}/libmpiseq.${dlext}" \
+    -DBLAS_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
+    -DLAPACK_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
+    ..
+
+make -j${nproc}
+cp libuno.a ${libdir}/libuno.a
```
amontoison, line 54:
```suggestion
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+cp uno_ampl${exeext} ${bindir}/uno_ampl${exeext}
+# ${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${OMP} -l${LBT} -ldmumps -lmetis
```
amontoison, line 56:
```suggestion
${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${OMP} -l${LBT} -ldmumps -lmetis
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+"""
+
+platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
+
+products = [
+    ExecutableProduct("uno_ampl", :amplexe),
```
amontoison, line 63:
```suggestion
    LibraryProduct("libuno", :libuno),
    ExecutableProduct("uno_ampl", :amplexe),
```

amontoison:
Ok, we can now generate a shared library `libuno`.

We should remove the platform `aarch64-unknown-freebsd`, METIS is now compiled on this new plateform yet.

Mac and FreeBSD are the last platforms that are failing because of `clang++`.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+    -DMUMPS_COMMON_LIBRARY="${libdir}/libmumps_common.${dlext}" \
+    -DMUMPS_PORD_LIBRARY="${libdir}/libpord.${dlext}" \
+    -DMUMPS_MPISEQ_LIBRARY="${libdir}/libmpiseq.${dlext}" \
+    -DBLAS_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
+    -DLAPACK_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
+    ..
+
+make -j${nproc}
+cp uno_ampl${exeext} ${bindir}/uno_ampl${exeext}
+${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${OMP} -l${LBT} -ldmumps -lmetis
```
amontoison, line 55:
```suggestion
${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${libdir} -l${OMP} -l${LBT} -ldmumps -lmetis
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "f166e60e428c8dafed6c3ded7df49fe2e41a2c03",
```
amontoison, line 12:
```suggestion
        "f05a6f011ca548588b588ed8fd7782aeb5043def",
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "f05a6f011ca548588b588ed8fd7782aeb5043def",
```
odow, line 12:
```suggestion
        "f723292439efa4723e1bd509c9e2253d65b49b3e",
```

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "f723292439efa4723e1bd509c9e2253d65b49b3e",
```
odow, line 12:
```suggestion
        "8f5a46ef9360e262caa26148ac68fcb3989a5794",
```

odow:
@cvanaret new errors, but I think we're very close.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,92 @@
...
+    -DBLAS_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
+    -DLAPACK_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
+    ..
+
+make -j${nproc}
+cp uno_ampl${exeext} ${bindir}/uno_ampl${exeext}
+${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${libdir} -l${OMP} -l${LBT} -ldmumps -lmetis
+"""
+
+platforms = supported_platforms()
```
odow, line 58:
```suggestion
platforms = supported_platforms()
filter!(p -> triplet(p) != "aarch64-unknown-freebsd", platforms)
```

cvanaret:
> @cvanaret new errors, but I think we're very close.

I wish my GCC weren't so permissive and would just crash instead of silently including missing headers :)
It's fixed in commit 539186253be74e309ede0a43579e611832353739.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,93 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "8f5a46ef9360e262caa26148ac68fcb3989a5794",
```
odow, line 12:
```suggestion
        "539186253be74e309ede0a43579e611832353739",
```

odow:
Building for Yggdrasil is a _very_ good test for how easy a project is to compile. The good news is that once it's up and running it is typically pretty easy to maintain. The hard work is getting everything working in the first place.

cvanaret:
Faulty Statistics header fixed in 1f373ce25049e593088bee66b37a9057039ac19a. Hopefully the last iteration...

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,93 @@
...
+using BinaryBuilder, Pkg
+
+name = "Uno"
+
+version = v"1.1.0"
+
+sources = [
+    GitSource(
+        "https://github.com/cvanaret/Uno.git",
+        "539186253be74e309ede0a43579e611832353739",
```
odow, line 12:
```suggestion
        "1f373ce25049e593088bee66b37a9057039ac19a",
```

odow:
Lucky last?

odow:
@amontoison I think we should remove the shared library. There's currently no need for it?

amontoison:
Yes, we can remove it for now.
At least we know we can build it for all platforms, which is good for us.
We just need to wait for a C API.

amontoison (approved):

cvanaret:
I can't thank you enough for your patience! I'll try to be more careful with the includes from now on.
I'll be in touch when the `make install` is working.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,99 @@
...
+
+# Uno does not support `make install`. Manually copy for now.
+cp uno_ampl${exeext} ${bindir}/uno_ampl${exeext}
+
+# Currently, Uno does not provide a shared library. THis may bbe useful in future once it has a C API.
+# ${CXX} -shared $(flagon -Wl,--whole-archive) libuno.a $(flagon -Wl,--no-whole-archive) -o "${libdir}/libuno.${dlext}" -L${libdir} -l${OMP} -l${LBT} -ldmumps -lmetis
+"""
+
+platforms = supported_platforms()
+filter!(p -> triplet(p) != "aarch64-unknown-freebsd", platforms)
```
giordano, line 63: Triple comparison is brittle
```suggestion
filter!(p -> !(Sys.isfreebsd(p) && arch(p} == "aarch64"), platforms)
```

odow: Oops. Okay. I'll fix in a separate PR.

U/Uno/build_tarballs.jl
```diff
@@ -0,0 +1,99 @@
...
+    -DMUMPS_PORD_LIBRARY="${libdir}/libpord.${dlext}" \
+    -DMUMPS_MPISEQ_LIBRARY="${libdir}/libmpiseq.${dlext}" \
+    -DBLAS_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
+    -DLAPACK_LIBRARIES="${libdir}/lib${LBT}.${dlext}" \
+    ..
+
+make -j${nproc}
+
+# Uno does not support `make install`. Manually copy for now.
+cp uno_ampl${exeext} ${bindir}/uno_ampl${exeext}
```
giordano, line 56: I also made this suggestion which wasn't published before
```suggestion
install -Dvm 755 "uno_ampl${exeext}" - t "${bindir}"
```

odow: Okay I'll fix in that same PR. There were so many comments on this PR that it's not surprising a few things got missed :smile:

PR #8940, opened 2024-06-21
Add builder for Wasmtime CLI and C API

ararslan:
Let's see whether this works...


ararslan:
🤔 I downloaded the macOS AArch64 artifacts to ensure things looked solid but they in fact seem not solid... `dlopen`ing the library and trying to run the CLI both bring up a message that says the file damaged and can't be opened and that I should move it to the trash.

PR #8951, opened 2024-06-23
Reactant: embarassing bugfix v2

wsmoses:


giordano:
Is the bugfix the fact that the version number wasn't changed in #8949?

wsmoses:
Yes

PR #8953, opened 2024-06-23
[Extrae] Bump to v4.2.0 + MPI support

mofeing:


E/Extrae/build_tarballs.jl
```diff
@@ -36,10 +37,14 @@ if [[ $bb_target = powerpc64le* ]]; then
 fi
 
 if [[ $bb_full_target = *cuda\+[0-9]* ]]; then
-    export ENABLE_CUDA=1
+    export FLAGS_CUDA=--with-cuda=$prefix/cuda
     export CPPFLAGS="-I${prefix}/cuda/include"
 fi
 
+if [[ $bb_full_target = *mpi* ]] && [[ $bb_full_target != *mpi\+none* ]]; then
+    export FLAGS_MPI="--with-mpi=$prefix --with-mpi-binaries=$prefix/bin --with-mpi-headers=$prefix/include --with-mpi-libs=$prefix/lib"
```
giordano, line 45:
```suggestion
    export FLAGS_MPI="--with-mpi=$prefix --with-mpi-binaries=${bindir} --with-mpi-headers=${includedir} --with-mpi-libs=${libdir}"
```

E/Extrae/build_tarballs.jl
```diff
@@ -51,39 +56,53 @@ autoreconf -fvi
...
     Platform("powerpc64le", "linux"; libc="glibc"),
 ]
-non_cuda_platforms = expand_cxxstring_abis(non_cuda_platforms)
+platforms = expand_cxxstring_abis(platforms)
 # (Almost) Same as PAPI
 cuda_platforms = CUDA.supported_platforms(; max_version=v"12.2")
 cuda_platforms = expand_cxxstring_abis(cuda_platforms)
 
+mpi_platforms, mpi_dependencies = MPI.augment_platforms(platforms)
+filter!(platform -> platform["mpi"] != "mpitrampoline", mpi_platforms)
```
giordano, line 81: Why is this removed?

E/Extrae/build_tarballs.jl
```diff
@@ -51,39 +56,53 @@ autoreconf -fvi
...
-non_cuda_platforms = [
+platforms = [
     Platform("x86_64", "linux"; libc="glibc"),
     Platform("aarch64", "linux"; libc="glibc"),
     Platform("powerpc64le", "linux"; libc="glibc"),
 ]
-non_cuda_platforms = expand_cxxstring_abis(non_cuda_platforms)
+platforms = expand_cxxstring_abis(platforms)
 # (Almost) Same as PAPI
 cuda_platforms = CUDA.supported_platforms(; max_version=v"12.2")
```
giordano, line 77: I think we can do like PAPI: https://github.com/JuliaPackaging/Yggdrasil/blob/e48091185c15cd2807bca8ad3966910ff4aadb8e/P/PAPI/build_tarballs.jl#L94

E/Extrae/build_tarballs.jl
mofeing: It's giving errors because MPITrampoline doesn't provide `libmpi.so`, which is what Extrae is looking for.

Maybe a symbolic link could work?

giordano: No, we shouldn't make up fake symlinks which are going to be broken. Also, libblastrampoline isn't alone in not using libmpi as soname (also some vendor mpi libraries have different names), it sounds like extrae should allow optionally changing the library name.

PR #8956, opened 2024-06-24
[finufft] Set Cmake build type to Release

ludvigak:
This fixes a performance problem created after the switch to Cmake building: The build type was not set to Release, thereby missing -O3 optimizations.


ludvigak:
This is ready to be merged. Also a bit critical, since FINUFFT.jl performance is currently severely degraded due the current binaries being slow (about 5x slower in some cases).

PR #8957, opened 2024-06-24
Revert to the older version of VMEC due to bug

benjaminfaber:
Need to revert to an older version of the underlying VMEC due to a bug introduced in the latest version that requires a much larger fix.

### DIFF
--- a/V/VMEC/build_tarballs.jl
+++ b/V/VMEC/build_tarballs.jl
@@ -4,11 +4,12 @@ const YGGDRASIL_DIR = "../.."
 include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
 
 name = "VMEC"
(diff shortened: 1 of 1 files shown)


giordano:
Isn't it easier to yank the broken version in the registry?

benjaminfaber:
I wasn't aware I could do that, how do I go about doing that?  Do I submit a PR to the `General` registry with that version removed?

giordano:
https://github.com/JuliaRegistries/General/blob/d36bad416efc43e6fcd66dfaf7fa597622e8a39e/README.md#when-is-yanking-a-release-appropriate

benjaminfaber:
I reverted the VMEC.jl packaged to exclude VMEC_jll v1.3.  I'll close this and update VMEC_jll once we have the bug fixed.

PR #8960, opened 2024-06-25
Wizard recipe: HeFFTe-v2.4.1

ahojukka5:

* Package name: HeFFTe
* Version: v2.4.0

### DIFF
--- a/H/HeFFTe/build_tarballs.jl
+++ b/H/HeFFTe/build_tarballs.jl
@@ -84,7 +84,7 @@ platforms = expand_cxxstring_abis(platforms)
 mpi_platforms, mpi_dependencies = MPI.augment_platforms(platforms)
 
 # CUDA+MPI platforms (x86_64 only)
-cuda_platforms = expand_cxxstring_abis(CUDA.supported_platforms(min_version=v"11.0"))
+cuda_platforms = expand_cxxstring_abis(CUDA.supported_platforms(min_version=v"12.0"))
 filter!(p -> arch(p) == "x86_64", cuda_platforms)
 cudampi_platforms, cudampi_dependencies = MPI.augment_platforms(cuda_platforms)
 
--- a/H/HeFFTe/build_tarballs.jl
+++ b/H/HeFFTe/build_tarballs.jl
@@ -1,6 +1,6 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
-using BinaryBuilder, Pkg
+using BinaryBuilder, Pkg, BinaryBuilderBase
 
 const YGGDRASIL_DIR = "../.."
 include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
@@ -16,6 +16,7 @@ sources = [
 ]
 
 # Bash recipe for building across all platforms
+# Note: HeFFTe requires MPI unconditionally (find_package(MPI REQUIRED))
 script = raw"""
 cd $WORKSPACE/srcdir/heffte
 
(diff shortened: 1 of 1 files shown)


L/libheffte/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="FFTW_jll", uuid="f5851436-0d7a-5f13-b9de-f02708fd171a"))
+    Dependency(PackageSpec(name="MPICH_jll", uuid="7cb0a576-ebde-5e09-9194-50597f1243b4"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"13.2.0")
```
imciner2, line 52: Why did you select GCC 13.2? For compatibility, we like people to use the oldest GCC version they can (https://github.com/JuliaPackaging/Yggdrasil/blob/master/RootFS.md#sources-of-incompatibility).

L/libheffte/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
```
imciner2, line 37:
```suggestion
platforms = filter(Sys.islinux, supported_platforms())
platforms = expand_cxxstring_abis(platforms)
```
It looks like you only want to support Linux, so this is a better way of showing this. Also, since you have a C++ library, we have to expand the string ABI (this is flagged in the auditor log).

L/libheffte/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd heffte
+cmake -B build -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+cmake --build build
+cmake --install build
+exit
```
imciner2, line 20:
```suggestion
```

L/libheffte/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/icl-utk-edu/heffte.git", "d2dea23b71979ddad19b58f73793fff9b1527285")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd heffte
+cmake -B build -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
```
imciner2, line 17:
```suggestion
cmake -B build \
    -DCMAKE_INSTALL_PREFIX=$prefix \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release
```
Minor style nit.

L/libheffte/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/icl-utk-edu/heffte.git", "d2dea23b71979ddad19b58f73793fff9b1527285")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd heffte
```
imciner2, line 16:
```suggestion
cd $WORKSPACE/srcdir/heffte
```

L/libheffte/build_tarballs.jl
ahojukka5: Ok, I have to try older versions. Can you suggest some versions to start with? The reason I selected 13.2 was that I was not able to compile it with the default compiler BinaryBuilder suggested.

ahojukka5: I changed the preferred version to 9 which seems to work.

ahojukka5:
After thinking a bit, HeFFTe might be a better name.

fingolfin (approved):
Ugh, it seems this one fell from the stack and we forgot about it :-/. I merged master to trigger a fresh CI run.

Codewise this seems fine to me?

ahojukka5:
Thank you for paying attention to this.

PR #8959, opened 2024-06-24
[RootFs/Go] Update to v1.23.0

lamdor:
Updates go from 1.22.0 to 1.22.4. There have been many security fixes since 1.22.0: https://go.dev/doc/devel/release#go1.22.minor

PR #8967, opened 2024-06-26
[Dex] Update to 2.40.0

lamdor:
Updates the patches to new locations in the files. The 3rd patch is now removed since it is actually contributed upstream.

Also changed to build on all supported_platforms.

It would be nice to wait until https://github.com/JuliaPackaging/Yggdrasil/pull/8959 is done before merging this.


giordano:
This is failing

tanmaykm:
Put up https://github.com/JuliaPackaging/Yggdrasil/pull/9647 which should supersede this. It doesn't yet build on all platforms though, which can be a future effort.

PR #8968, opened 2024-06-26
[GoogleCodeSearch] bump julia_compat so it can build

lamdor:
This was required so the build of GoogleCodeSearch can continue on julia 1.6. There weren't any issues with julia 1.6 and building this.

This should wait to be merged until after https://github.com/JuliaPackaging/Yggdrasil/pull/8959


giordano (changes requested):
Need to have a different version number to change Julia compat. Is there a newer version we can build?

imciner2:
> Need to have a different version number to change Julia compat. Is there a newer version we can build?

Looks like they only tagged an RC for the next 1.3.0 release (https://github.com/google/codesearch/tags), so there isn't a new version yet.

PR #3695, opened 2021-10-05
Update Openresty to v1.19.9

tanmaykm:
Update Openresty to v1.19.9 and use openssl v1.1.1l instead of v1.0.
Dropping 32bit build, it does not compile and is also not used.

PR #8970, opened 2024-06-26
[Openresty] Update to 1.25.3 and newer openssl

lamdor:
This was pretty out of date.

### DIFF
--- a/O/Openresty/build_tarballs.jl
+++ b/O/Openresty/build_tarballs.jl
@@ -3,14 +3,14 @@
 using BinaryBuilder, Pkg
 
 name = "Openresty"
-version = v"1.21.4"
+version = v"1.25.3"
 
 # Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://openresty.org/download/openresty-$(version).1.tar.gz", "0c5093b64f7821e85065c99e5d4e6cc31820cfd7f37b9a0dec84209d87a2af99"),
+    ArchiveSource("https://openresty.org/download/openresty-$(version).1.tar.gz", "32ec1a253a5a13250355a075fe65b7d63ec45c560bbe213350f0992a57cd79df"),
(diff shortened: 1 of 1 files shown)


O/Openresty/build_tarballs.jl
```diff
@@ -3,14 +3,14 @@
...
 
 # Collection of sources required to complete build
 sources = [
-    ArchiveSource("https://openresty.org/download/openresty-$(version).1.tar.gz", "0c5093b64f7821e85065c99e5d4e6cc31820cfd7f37b9a0dec84209d87a2af99"),
+    ArchiveSource("https://openresty.org/download/openresty-$(version).1.tar.gz", "32ec1a253a5a13250355a075fe65b7d63ec45c560bbe213350f0992a57cd79df"),
     ArchiveSource("https://sourceforge.net/projects/pcre/files/pcre/8.45/pcre-8.45.tar.bz2", "4dae6fdcd2bb0bb6c37b5f97c33c2be954da743985369cddac3546e3218bffb8"),
-    ArchiveSource("https://www.openssl.org/source/openssl-1.1.1p.tar.gz", "bf61b62aaa66c7c7639942a94de4c9ae8280c08f17d4eac2e44644d9fc8ace6f"),
-    ArchiveSource("https://www.zlib.net/zlib-1.2.12.tar.gz", "91844808532e5ce316b3c010929493c0244f3d37593afd6de04f71821d5136d9")
+    ArchiveSource("https://www.openssl.org/source/openssl-3.3.0.tar.gz", "53e66b043322a606abf0087e7699a0e033a37fa13feb9742df35c3a33b18fb02"),
+    ArchiveSource("https://www.zlib.net/zlib-1.3.1.tar.gz", "9a93b2b7dfdac77ceba5a558a580e74667dd6fede4585b91eefb60f03b72df23")
```
giordano, line 13: Now I'm confused: why this isn't using zib and openssl from the jlls?

lamdor: It was this way before:

https://github.com/JuliaPackaging/Yggdrasil/pull/3695#issuecomment-936090121

It looks as if openresty statically compiles them in.

giordano:
Ok, this is an executable so static linking isn't that bad.  But adding a comment to the code to explain the rationale and avoid the same question in the future would be helpful.

PR #8981, opened 2024-06-28
[open62541] specify desired gcc version; fixes segfaults on Windows + Linux.

thomvet:
Increases gcc version. When compiling with v4, v5, or v6, the library produced is not usable, due to segfaults occurring on Windows (with basic functionality). Building with v7 and above makes tests within Open62541.jl pass.

Issue described here: https://discourse.julialang.org/t/a-plea-for-help-with-open62541-opc-ua-binaries/114702
Diagnosed further here: https://github.com/martinkosch/Open62541.jl/pull/22

I know that using the lowest possible gcc version is advised for compatibility reasons, but this fix seems to do the job and resolving the issues while keeping gcc v4 might be a task that is beyond my (current) capabilities.

I hope this workaround is acceptable.

### DIFF
--- a/O/open62541/open62541@1.3/build_tarballs.jl
+++ b/O/open62541/open62541@1.3/build_tarballs.jl
@@ -11,6 +11,14 @@ sources = [
 ]
 # Bash recipe for building across all platforms
 script = raw"""
+# Deactivates stack protector under i686-linux-musl; necessary to avoid 
+# "undefined reference to `__stack_chk_fail_local`
+if [[ ${target} == i686-linux-musl ]]; then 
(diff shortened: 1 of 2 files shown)


giordano:
You can't build two different versions of the same package at the same time, this will fail in the registry

thomvet:
Also when it's just a build update? (1.3.10 and 1.4.0 are already there,so it would "only" increment the 1.3.10+x part).

How to fix? Split into two PRs?

giordano:
Ah, that may work, the problem is the merge conflict in the registry, but if there's a version in between you may not have problems.

PR #8990, opened 2024-06-28
[GLFW] Update to 3.4

JamesWrigley:
`Wayland_jll` and `xkbcommon_jll` are required for Wayland.


JamesWrigley:
FreeBSD is strange... I think this is ready to go but I'm not sure everything it's doing is sane.

PR #8992, opened 2024-06-29
Enzyme, yet again building attempt

wsmoses:


giordano:
I'll try to reduce parallelism in ghr before merging this PR.

giordano (approved):
Ok, let's hope https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1334 + #8995 are enough.  If not I wonder if we are under the [condition](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits)

> You may also encounter a secondary rate limit for undisclosed reasons.

in which case we may need to reach out to GitHub support to understand what's going on :melting_face:

PR #8993, opened 2024-06-29
[libsingular_julia] for option handling

hannes14:
@fingolfin


giordano:
This is failing on master: https://buildkite.com/julialang/yggdrasil/builds/11623

fingolfin:
```
┌ Error: Unable to find dependency!
--
  | │   name = "MPFR_jll"
  | │   version = v"4.2.1+0"
  | │   registries =

```
but why?!

fingolfin:
Aha, so MPFR 4.2.1 indeed is not in the General registry. So I guess we need to re-land PR #7376 ?

giordano:
Now I'm extremely confused: how's it possible a version of a jll is in Julia but not registered?

giordano:
This is now working after https://github.com/JuliaRegistries/General/pull/110107: https://buildkite.com/julialang/yggdrasil/builds/11623

PR #8997, opened 2024-06-30
[Extrae] Try fix `augment_platform_block` error

mofeing:
Try fix the error in https://buildkite.com/julialang/yggdrasil/builds/11585#01905fc2-8ac7-4ee3-a7f6-f475db0874c5/6-14497


giordano:
I'm a bit confused how that was considered a `Vector{String}`, it looks a `String` to me.  Having a non-conditional is the right thing to do anyway.

PR #9000, opened 2024-06-30
New Recipe: libtakum v0.1.0

hunhold:
[libtakum](https://github.com/takum-arithmetic/libtakum) is an ISO C99 implementation of [takum arithmetic](https://arxiv.org/abs/2404.18603), a proposed improvement over posit arithmetic of which Yggdrasil contains the SoftPosit implementation. This recipe provides libtakum-0.1.0, the first stable release. There are no warnings generated on any of the tested platforms.

Thanks in advance for reviewing this pull request!

### DIFF
--- a/L/libtakum/build_tarballs.jl
+++ b/L/libtakum/build_tarballs.jl
@@ -1,0 +1,34 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "libtakum"
+version = v"0.1.0"
+
+# Collection of sources required to complete build
+sources = [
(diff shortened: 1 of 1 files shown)


imciner2:
In general, we don't tend to distribute the static libraries as part of Yggdrasil unless it is really needed, since they take up space on the user's install and aren't actually used by the Julia packages (only in other Yggdrasil builds). I would suggest removing the static libraries that are being built here.

hunhold:
Thank you for your feedback, @imciner2! Please excuse that I missed out on it, which was due to a misunderstanding of how Yggdrasil works. I assumed that it autodetects dynamic library files and cherry-picks them out of the installation-prefix. I have pushed a commit that should do the trick.

One more thing, just out of interest as I couldn't find any information about this anywhere: Does the final jll-package make use of the included pkgconfig file?

Is the recipe good to go now or can I still improve it in any way? Thanks in advance for your valuable feedback! :)

PR #9001, opened 2024-06-30
[MPFR] 4.2.1 again

fingolfin:
For some reason 4.2.1 never made it into the General registry. So rebuild it in the hopes that this will resolve the situation.

See also discussion in PR #8993 

CC @benlorenz @giordano @hannes14

### DIFF
--- a/M/MPFR/build_tarballs.jl
+++ b/M/MPFR/build_tarballs.jl
@@ -46,3 +46,4 @@ dependencies = [
 ]
 
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"5", julia_compat="1.6")
+


giordano:
This is wrong, I need to understand how the version isn't registered already.

giordano:
To elaborate why this is wrong: the only way to merge this PR is to delete the existing release of MPFR_jll 4.2.1 which is used already in Julia, so that's an unacceptable option.

giordano:
I believe the right course of action is to manually register https://github.com/JuliaBinaryWrappers/MPFR_jll.jl/commit/f98f1f26f2453ea7c6ef7fff2672f4dcae9445fd

giordano:
Superseded by https://github.com/JuliaRegistries/General/pull/110107.

PR #9003, opened 2024-07-01
Upgrade object_store_ffi to v0.7.0

andrebsguedes:


giordano:
https://buildkite.com/julialang/yggdrasil/builds/11652#01906ec2-b05b-4dea-acec-bd980ed03e0d/669-1201
```
ERROR: could not load library "/cache/build/yggy-amdci7-4/julialang/yggdrasil/O/object_store_ffi/build/x86_64-linux-gnu/MaHRZDTx/x86_64-linux-gnu-libgfortran3-cxx11/destdir/lib/libobject_store_ffi.so"
/cache/build/yggy-amdci7-4/julialang/yggdrasil/O/object_store_ffi/build/x86_64-linux-gnu/MaHRZDTx/x86_64-linux-gnu-libgfortran3-cxx11/destdir/lib/libobject_store_ffi.so: undefined symbol: jl_adopt_thread
```

O/object_store_ffi/build_tarballs.jl
```diff
@@ -36,5 +36,5 @@ dependencies = Dependency[
 # Build the tarballs
 build_tarballs(
     ARGS, name, version, sources, script, platforms, products, dependencies;
-    compilers=[:c, :rust], julia_compat="1.6", preferred_gcc_version=v"5",
+    compilers=[:c, :rust], julia_compat="1.9", preferred_gcc_version=v"5",
```
giordano, line 39: Ah, you require Julia v1.9.  We run CI here with Julia v1.7 and can't upgrade that version, so I think your only option is to set `dont_dlopen=true` for the [`LibraryProduct`](https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct) above.

andrebsguedes: Done!

giordano: Aaah, we have a less "aggressive" `dont_dlopen` option to pass the `build_tarballs` instead of `LibraryProduct`

O/object_store_ffi/build_tarballs.jl
```diff
@@ -26,7 +26,7 @@ platforms = [
 
 # The products that we will ensure are always built
 products = [
-    LibraryProduct("libobject_store_ffi", :libobject_store_ffi),
+    LibraryProduct("libobject_store_ffi", :libobject_store_ffi; dont_dlopen=true),
```
giordano, line 29:
```suggestion
    LibraryProduct("libobject_store_ffi", :libobject_store_ffi),
```

O/object_store_ffi/build_tarballs.jl
giordano:
```suggestion
    compilers=[:c, :rust], julia_compat="1.9", preferred_gcc_version=v"5", dont_dlopen=true,
```

PR #9015, opened 2024-07-03
Update isa-l from 2.30 to 2.31

jakobnissen:
Let me try also enabling builds for AArch MacOS and Windows - the changelog for v2.31 says they've fixed some compilation issues. If these still fail, we can disable them again.


jakobnissen:
Okay, turns out these still fail - that might be resolved in future versions of isal (https://github.com/intel/isa-l/issues/285, https://github.com/intel/isa-l/issues/276)

PR #9022, opened 2024-07-03
[aws_c_cal] Update to version 0.7.1

quinnj2:
This PR updates aws_c_cal to version 0.7.1. cc: @quinnj @Octogonapus

### DIFF
--- a/A/aws_c_cal/build_tarballs.jl
+++ b/A/aws_c_cal/build_tarballs.jl
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "aws_c_cal"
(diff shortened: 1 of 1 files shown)


giordano:
@quinnj can you please remove the continuous force-push if nothing changes?  Thanks!

quinnj:
Yeah, sorry about that. I think I put it in initially while testing things out to make sure I force-pushed the updated-correct thing, but I think we're in a good state here now.

quinnj:
(I've taken out the force-push, so we shouldn't see that behavior any more)

giordano:
Thanks!

PR #9030, opened 2024-07-06
libunwind: Enable per thread register state cache

andrebsguedes:
https://github.com/JuliaLang/julia/pull/55049


giordano:
I think it's easier to do this only for v1.8, also because it's the only version which is actually building.

PR #9034, opened 2024-07-06
New Recipe: libdecor v0.2.2

jmert:
Create a build recipe for [libdecor](https://gitlab.freedesktop.org/libdecor/libdecor) which is used by some Wayland clients.

My motivation is to try to get GLFW working on Wayland (see [GLFW release notes](https://www.glfw.org/docs/3.4/news.html#wayland_libdecor_decorations)), though in local testing adding `libdecor_jll` as a runtime dependency to `GLFW_jll` isn't sufficient to magically solve all problems).

### DIFF
--- a/L/libdecor/build_tarballs.jl
+++ b/L/libdecor/build_tarballs.jl
@@ -37,14 +37,14 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("xkbcommon_jll"; platforms),
-    Dependency("Dbus_jll"; platforms),
-    Dependency("Libglvnd_jll"; platforms),
-    Dependency("Pango_jll"; platforms),
-    Dependency("Wayland_jll"; platforms),
-    HostBuildDependency("Wayland_jll"; platforms),
-    BuildDependency("Wayland_protocols_jll"; platforms),
-    BuildDependency("Xorg_xorgproto_jll"; platforms),
+    Dependency("xkbcommon_jll"),
+    Dependency("Dbus_jll"),
+    Dependency("Libglvnd_jll"),
+    Dependency("Pango_jll"; compat="1.52.2"),
+    Dependency("Wayland_jll"),
+    HostBuildDependency("Wayland_jll"),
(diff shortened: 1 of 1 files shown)


L/libdecor/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+products = [
+    LibraryProduct("libdecor-0", :libdecor),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("xkbcommon_jll"; platforms),
+    Dependency("Dbus_jll"; platforms),
+    Dependency("Libglvnd_jll"; platforms),
+    Dependency("Pango_jll"; platforms),
```
giordano, line 43:
```suggestion
    Dependency("Pango_jll"; compat="1.52.2", platforms),
```

L/libdecor/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+    Dependency("Wayland_jll"; platforms),
+    HostBuildDependency("Wayland_jll"; platforms),
+    BuildDependency("Wayland_protocols_jll"; platforms),
+    BuildDependency("Xorg_xorgproto_jll"; platforms),
```
giordano, line 47: Any particular reason for filtering for all platforms anyway?

jmert: Maybe no good reason — I was assuming everything should only try to load on a Wayland-compatible system, so this was my guess as to how to do that. If this isn't the right way to do it, I'm happy to learn what's the recommended practice.

giordano: With `platforms=platforms` you're making a dependency be loaded only for the listed platforms, but when these are just _all_ the platforms I miss the point of filtering in the first place, since you aren't filtering anything.

jmert: I guess I wasn't sure how the list of build platforms interacts with the generated Julia package's `using`s. If the "implicit" platform list is sufficient, I'll just get rid of these.

PR #9073, opened 2024-07-13
[CSL] Rebuild for GCC v13

staticfloat:
I'm trying to get a libstdc++ with GLIBCXX_3.4.30 or higher, which should come with GCC v13.


giordano:
Isn't this already targeting gcc 13?

giordano:
And according to https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html `GLIBCXX_3.4.30` is already in GCC 12.1

staticfloat:
Yeah, I'm confused too.  If I download the current

staticfloat:
Somehow I was downloading an older one I guess.  The libstdc++.so.6.0.32 that I had on disk had the wrong contents. Spooky.

PR #9076, opened 2024-07-13
Reactant also build debug binaries

wsmoses:


wsmoses:
@giordano offhand do you know the right way to add a custom build var here [in this case debug or not]?

R/Reactant/build_tarballs.jl
```diff
@@ -272,11 +272,12 @@ platforms = filter(p -> !(Sys.isfreebsd(p)), platforms)
 
 augment_platform_block=CUDA.augment::String
 
-for platform in platforms
+for mode in ("opt", "dbg"), platform in platforms
     augmented_platform = deepcopy(platform)
+    augmented_platform["mode"]
```
giordano, line 277: I think you want to assign a value here, like
```suggestion
    augmented_platform["mode"] = "mode"
```
?  But you also need to expand `augment_platform_block` to teach it how to set the `mode` tag.

PR #9085, opened 2024-07-16
[Osi] Upgrade to v0.108.10

amontoison:
@odow

### DIFF
--- a/C/Coin-OR/Osi/build_tarballs.jl
+++ b/C/Coin-OR/Osi/build_tarballs.jl
@@ -1,7 +1,7 @@
 # In addition to coin-or-common.jl, we need to modify this file to trigger a
 # rebuild.
 #
-# Last updated: 2022-10-26
+# Last updated: 2024-07-15
 
 include("../coin-or-common.jl")
 
@@ -33,6 +33,13 @@ elif [[ "${target}" == *linux* ]]; then
     export LDFLAGS="-ldl -lrt"
 fi
 
+# BLAS and LAPACK
+if [[ "${target}" == *mingw* ]]; then
+  LBT="-lblastrampoline-5"
+else
+  LBT="-lblastrampoline"
+fi
+
 ../configure \
     --prefix=${prefix} \
     --build=${MACHTYPE} \
@@ -43,8 +50,10 @@ fi
     --enable-shared \
     lt_cv_deplibs_check_method=pass_all \
     --with-coinutils-lib="-lCoinUtils" \
-    --with-blas-lib="-lopenblas" \
-    --with-lapack-lib="-lopenblas"
+    --with-blas \
+    --with-blas-lib="-L${libdir} ${LBT}" \
+    --with-lapack \
+    --with-lapack-lib="-L${libdir} ${LBT}"
 
 make -j${nproc}
 make install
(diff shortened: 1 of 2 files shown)


C/Coin-OR/coin-or-common.jl
```diff
@@ -42,11 +42,13 @@ Cgl_gitsha = "8952b9e737e434b730fab5967cd28180b43d7234"
...
-Osi_gitsha = "c67528fc8b2a36dc936b285fe34030439b9147c3"
+Osi_version = offset_version(v"0.108.10", v"0.0.0")
+Osi_gitsha = "2997cda8e85ccc6712c4b05404e7aa70500e422f"
 
-CoinUtils_version = offset_version(v"2.11.6", v"0.0.0")
-CoinUtils_gitsha = "26e9639ed9897e13e89169870dbe910296a9783b"
+CoinUtils_upstream_version = v"2.11.11"
+CoinUtils_version = offset_version(CoinUtils_upstream_version, v"0.0.0")
+CoinUtils_gitsha = "15a819b7e8763b2557e4bf440f8cb62ee6734f36"
+CoinUtils_hash = "27da344479f38c82112d738501643dcb229e4ee96a5f87d4f406456bdc1b2cb4"
```
odow, line 51: You might need to rebase? Why was this needed?

C/Coin-OR/Osi/build_tarballs.jl
```diff
@@ -58,9 +67,9 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("CoinUtils_jll", compat="$(CoinUtils_version)"),
-    Dependency("OpenBLAS32_jll", OpenBLAS32_version),
-    Dependency("CompilerSupportLibraries_jll"),
+    Dependency(PackageSpec(name="CoinUtils_jll", uuid="be027038-0da8-5614-b30d-e42594cb92df"), compat="$(CoinUtils_version)"),
```
odow, line 70:
```suggestion
    Dependency(PackageSpec(name="CoinUtils_jll", uuid="be027038-0da8-5614-b30d-e42594cb92df"), compat="=$(CoinUtils_version)"),
```

odow: We need to ensure strict compat `=` for all COIN-OR stuff going forward.

odow:
Looks like the failure was just that the registry hadn't propagated yet.

C/Coin-OR/Osi/build_tarballs.jl
amontoison: Ok, let's do like Ipopt.
In theory we don't need the strict compat `=` but as we observed in the past, updating one dependency could break all COIN-OR packages.

odow (approved):
LGTM, assuming CI passes

C/Coin-OR/Osi/build_tarballs.jl
```diff
@@ -74,5 +83,5 @@ build_tarballs(
     products,
     dependencies;
     preferred_gcc_version = gcc_version,
-    julia_compat = "1.6",
+    julia_compat = "1.9",
```
odow, line 86: Do we need a preferred llvm version here so that it matches the CoinUtils one?

odow, line 86:
```suggestion
    preferred_llvm_version = llvm_version,
    julia_compat = "1.9",
```

amontoison: I think that we need is `clang_use_lld = false`.
The issue is the new linker on Mac that doesn't support some options.
Mosè added this option after my recipe for `CoinUtils`.

But using an older version of `llvm` should also ensure that.

amontoison:
The next one is `Clp` (#8066).

PR #9087, opened 2024-07-17
Bump aws_c_auth cal dep version

quinnj:
@giordano, could you help me know what else I need to do to get a correct new release out this time? I know as-is, this PR isn't enough to just bump the dep version.

### DIFF
--- a/A/aws_c_auth/build_tarballs.jl
+++ b/A/aws_c_auth/build_tarballs.jl
@@ -36,7 +36,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("aws_c_cal_jll"; compat="0.6.2"),
+    Dependency("aws_c_cal_jll"; compat="0.7.1"),
     Dependency("aws_c_http_jll"; compat="0.8.1"),
     Dependency("aws_c_sdkutils_jll"; compat="0.1.12"),
     BuildDependency("aws_lc_jll"),


giordano:
Basically the version number has to be different than any other registered version.  How to do it specifically is mostly up to you, for example you could pretend this is a non-existing 0.7.23, or something like that.

quinnj:
Alternatively, I could just wait until there's an actual new release (0.7.23) and let it roll naturally, right? So I could merge this with a [no build] and then that will work next release?

giordano:
> Alternatively, I could just wait until there's an actual new release (0.7.23) and let it roll naturally, right?

Yes, that'd also work :slightly_smiling_face: 

> So I could merge this with a [no build] and then that will work next release?

I'd say `[skip ci]` because you don't want _anything_ to happen upon merging.

quinnj:
Ok, I pushed a commit with [skip ci] and enabled auto-merge. So hopefully it will do everything it needs?

giordano:
`[skip ci]` can be used only for the merge commit, otherwise we can't merge: https://github.com/JuliaPackaging/Yggdrasil/blob/eb7a7bfcb930e2751316e26c4b5ff43752443550/CONTRIBUTING.md#information-for-maintainers

quinnj:
Oops. forgot to disable auto-merge, so missed the chance to do [skip ci]

giordano:
Somehow that was already there: https://github.com/JuliaPackaging/Yggdrasil/commit/fd5c2b66c39805893eb8e951a156ba3907315f47

PR #9091, opened 2024-07-17
[Reactant] Augment platforms with Reactant compilation mode

mofeing:
CC @wsmoses

I'm not sure if we need to set the `MODE` env var in the `augment_platform_block`.


wsmoses:
@mofeing for now let's just disable linux debug since what's blocking is macos anyways

PR #9105, opened 2024-07-19
[alsa_plugins] Update to v1.2.12

giordano:


giordano:
Don't have the time to look into this now:
```
ERROR: could not load library "/cache/build/yggy-amdci7-2/julialang/yggdrasil/A/alsa_plugins/build/x86_64-linux-gnu/tJYGEpKQ/x86_64-linux-gnu-libgfortran3-cxx03/destdir/lib/alsa-lib/libasound_module_pcm_a52.so"
libcrypto.so.3: cannot open shared object file: No such file or directory
```

ViralBShah:
Is this a matter of adding a dependency on `OpenSSL_jll`? I tried it in the other PR (that I closed), and it didn't help resolve the error there, so I am doubtful if that will work here either.

ViralBShah:
@giordano Any thoughts here? If it worked on x86_64 linux, I would feel comfortable excluding the other platforms and merging.

ViralBShah:
Fixed in #10365

PR #9109, opened 2024-07-20
Add builder for spdlog

ararslan:


S/spdlog/build_tarballs.jl
```diff
@@ -0,0 +1,31 @@
...
+    -DSPDLOG_BUILD_PIC=ON \
+    -DSPDLOG_BUILD_EXAMPLE=OFF
+make -j${nproc} install
+"""
+
+platforms = expand_cxxstring_abis(supported_platforms())
+
+products = [LibraryProduct("libspdlog", :libspdlog)]
+
+dependencies = [Dependency("Fmt_jll")]
```
giordano, line 28: Based on  #9110 you have to specify the compat bound on this package (and using compat bounds is a good idea regardless)

ararslan: I didn't add one because I couldn't see anywhere in the spdlog documentation that said it only had compatibility with some version(s) of libfmt.

ararslan: Okay. How do I determine what compat bounds would be appropriate then?

giordano: According to #9110 v9 isn't compatible with v11.  So pick whatever you want.

ararslan:
```suggestion
dependencies = [Dependency("Fmt_jll"; compat="11")]
```

S/spdlog/build_tarballs.jl
```diff
@@ -0,0 +1,31 @@
...
+    -DSPDLOG_BUILD_PIC=ON \
+    -DSPDLOG_BUILD_EXAMPLE=OFF
+make -j${nproc} install
+"""
+
+platforms = expand_cxxstring_abis(supported_platforms())
+
+products = [LibraryProduct("libspdlog", :libspdlog)]
+
+dependencies = [Dependency("Fmt_jll"; compat="11")]
```
ararslan, line 28: Evidently the `compat` specification here doesn't work how I thought

giordano: https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.Dependency

> The optional keyword argument `compat` can be used to specify a string for use in the `Project.toml` of the generated Julia package. If `compat` is non-empty and `build_version` is not passed, the latter defaults to the minimum version compatible with the `compat` specifier.

First existing version of `Fmt_jll` in the v11 seres is 11.0.1:
```suggestion
dependencies = [Dependency("Fmt_jll"; compat="11.0.1")]
```

ararslan: IIUC, that will build spdlog with 11.0.1, but it will allow spdlog_jll to be used with Fmt_jll 11.0.2. Is that right?

giordano: Yes, that's standard Pkg compat specification.

ararslan: Right, I guess what I was confirming that it uses the minimum version within the declared range of compatibility for building (almost like inverse Pkg compat behavior), but maximum within the declared range otherwise (standard Pkg compat).

S/spdlog/build_tarballs.jl
```diff
@@ -0,0 +1,44 @@
...
+products = [
+    LibraryProduct("libspdlog", :libspdlog)
+]
+
+dependencies = [
+    Dependency("Fmt_jll"; compat="11.1.1")
+]
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="1.6")
```
imciner2, line 44:
```suggestion
               julia_compat="1.6", preferred_gcc_version=v"5")
```

imciner2:
Ok, everything builds on all our platforms with this version. This look goo @ararslan?

ararslan:
Yeah I think so. I opened this PR so long ago that I don't even remember why I was building this. 😅

PR #9110, opened 2024-07-20
[Exodus] Add compat entry to Fmt_jll

cmhamel:
adding compat entries to Fmt_jll and NetCDF_jll. This was necessary d…ue to the bump in Fmt_jll to 11 which is not compatable with the current version of exodusII.

### DIFF
--- a/E/EinsteinToolkit/build_tarballs.jl
+++ b/E/EinsteinToolkit/build_tarballs.jl
@@ -1,0 +1,238 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+using Base.BinaryPlatforms
+using BinaryBuilderBase: march
+const YGGDRASIL_DIR = "../.."
+include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
+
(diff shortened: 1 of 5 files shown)


giordano:
Always first fix these things in the registry.  It's useless to change anything here, if the registry is still wrong.

cmhamel:
> Always first fix these things in the registry. It's useless to change anything here, if the registry is still wrong.

Thanks for the quick feedback @giordano 

To make sure I understand correctly, should I open a PR to modify the following compat file https://github.com/JuliaRegistries/General/blob/master/jll/E/Exodus_jll/Compat.toml

giordano:
Yes, use the version of the dependencies that are actually compatible with the existing build.

giordano:
@cmhamel can you please adapt to the same changes as in? Without changing version number and introducing different changes.

cmhamel:
> @cmhamel can you please adapt to the same changes as in [JuliaRegistries/General#111403](https://github.com/JuliaRegistries/General/pull/111403)? Without changing version number and introducing different changes.

Apologies for missing your comment on adjusting the build. This should now reflect the Compat file in [https://github.com/JuliaRegistries/General/pull/111403](https://github.com/JuliaRegistries/General/pull/111403)

PR #9113, opened 2024-07-20
[file] Attempt updating to 5.45

ararslan:
Also reenable disabled platforms and install libmagic, just to see what happens

### DIFF
--- a/F/file/build_tarballs.jl
+++ b/F/file/build_tarballs.jl
@@ -15,18 +15,28 @@ script = raw"""
 cd $WORKSPACE/srcdir/file/
 
 autoreconf -i -f
-(
-    # Do native build to get the correct version of `file` locally
-    mkdir build-native && cd build-native
-    export CC=${CC_BUILD}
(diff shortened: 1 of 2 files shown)


ararslan:
Windows will need https://github.com/JuliaPackaging/Yggdrasil/pull/9124. Not sure yet what's up with x86_64-linux-musl.

ararslan:
It builds successfully on all platforms now! The remaining to do is to determine how to make libmagic look for the bundled magic.mgc. It appears to look for it on Windows but not elsewhere. Some ideas:
- Patch in a dependency on [whereami](https://github.com/gpakosz/whereami) to libmagic and always begin the search for magic.mgc relative to the filesystem location of libmagic
- Include the contents of magic.mgc in a top-level variable and default to referencing that instead of looking to the filesystem

ViralBShah:
Ready for merging?

ararslan:
Good question. @imciner2?

PR #9122, opened 2024-07-22
[Harfbuzz] Build v2.8.2

staticfloat:
This bumps us to Harfbuzz v2.8.2


imciner2:
We seem to have 8.5.0 now from https://github.com/JuliaPackaging/Yggdrasil/pull/9984, so is this still needed?

PR #9135, opened 2024-07-24
MAGEMin v1.5.0

NicolasRiel:
- Re-adds the individual solution phase minimization time
- Corrects issue when computing specific heat capacity through the Julia wrapper
- Corrects calculation of nno buffer


NicolasRiel:
@giordano Like always thanks!

PR #9156, opened 2024-07-27
[CUTEst] Upgrade to v2.2.2

amontoison:


amontoison:
I don't understand why it only doesn't work for windows 32-bit.

PR #9158, opened 2024-07-27
Update to libglvnd v1.7

jmert:
~~More importantly, setup the `__EGL_VENDOR_LIBRARY_DIRS` environment variable on load to point to locations of system-installed descriptors of the available EGL providers. Without this environment variable, compile-time paths (e.g. `/workspace/destdir/share/glvnd/egl_vendor.d`) appropriate for the Yggdrasil build environment will be used instead, which generally won't work when used by a user.~~

See also https://github.com/NVIDIA/libglvnd/blob/v1.7.0/src/EGL/icd_enumeration.md

More importantly, patch the loader to not look for the `egl_vendors.d` directory inside the BinaryBuilders path environment, which otherwise prevents the EGL library from finding the user's system EGL configuration.

### DIFF
--- a/L/Libglvnd/build_tarballs.jl
+++ b/L/Libglvnd/build_tarballs.jl
@@ -30,8 +30,8 @@ meson setup . .. \
     --cross-file="${MESON_TARGET_TOOLCHAIN}" \
     --buildtype=release \
     --prefix=$sysprefix \
-    --libdir=$prefix/lib \
-    --includedir=$prefix/include \
+    --libdir=${libdir} \
+    --includedir=${includedir} \
     "${FLAGS[@]}"
 ninja -j${nproc}
 ninja install
--- a/L/Libglvnd/build_tarballs.jl
+++ b/L/Libglvnd/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "Libglvnd"
-version = v"1.6.0"
+version = v"1.7.0"
 
 # Collection of sources required to build Libglvnd
 sources = [
     ArchiveSource("https://gitlab.freedesktop.org/glvnd/libglvnd/-/archive/v$(version)/libglvnd-v$(version).tar.gz",
-                  "efc756ffd24b24059e1c53677a9d57b4b237b00a01c54a6f1611e1e51661d70c"),
+                  "2b6e15b06aafb4c0b6e2348124808cbd9b291c647299eaaba2e3202f51ff2f3d"),
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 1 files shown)


L/Libglvnd/build_tarballs.jl
```diff
@@ -50,5 +54,14 @@ dependencies = [
     BuildDependency("Xorg_xorgproto_jll"),
 ]
 
+init_block = raw"""
+    if Sys.islinux()
+        get!(ENV, "__EGL_VENDOR_LIBRARY_DIRS", "/usr/share/glvnd/egl_vendor.d")
+    elseif Sys.isfreebsd()
+        get!(ENV, "__EGL_VENDOR_LIBRARY_DIRS", "/usr/local/share/glvnd/egl_vendor.d")
```
giordano, line 61: Isn't there anything better than hardcoding absolute paths?

jmert: I'm open to suggestions, but I haven't found a lot of useful information other than the linked icd_enumeration.md file about how the library should be configured to find various GL implementations. (It took quite a while of staring at `strace` outputs to get as far as knowing it's libglvnd which was the key link in the chain to getting GLFW to work on Wayland.)

At some level, I think there has to be an assumed absolute path somewhere — the point of Libglvnd is to use the `egl_vendor.d` directory to avoid hard-coding variations for Intel, AMD, NVIDIA, etc GL libraries. And more to the point, there's already an incorrect hard-coded absolute path built-in to the artifact:
```julia-repl
julia> using Libglvnd_jll

julia> # embedded string within the Libglvnd_jll artifact

julia> foreach(print, split(read(`strings $(Libglvnd_jll.libEGL)`, String), '\n') |> filter(contains("egl_vendor.d")))
/workspace/destdir/etc/glvnd/egl_vendor.d:/workspace/destdir/share/glvnd/egl_vendor.d

julia> # embedded string within my local system's libEGL.so

julia> foreach(print, split(read(`strings /usr/lib/libEGL.so`, String), '\n') |> filter(contains("egl_vendor.d")))
/etc/glvnd/egl_vendor.d:/usr/share/glvnd/egl_vendor.d
```

Rather than trying to split based on OS, though, I could simplify the statement to just
```julia
         get!(ENV, "__EGL_VENDOR_LIBRARY_DIRS", "/etc/glvnd/egl_vendor.d:/usr/share/glvnd/egl_vendor.d:/usr/local/share/glvnd/egl_vendor.d")
```
and cover all paths I've been able to find in one go.

jmert: Maybe as a useful comparison, I went and dug through the OpenCL loader code and found that it hard-codes a path during compile to `/etc/OpenCL/vendors/`: https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/c2cbdb33605ef9cfa02aa2a77e9cd7306c57e32a/loader/icd_platform.h#L30

The difference here is that the compile-time path makes use of the build `${prefix}` whereas the OpenCL loader ignores that and just hard-codes the normal system path.

If it'd be preferable to patch the build to similarly hard-code the destination rather than doing a runtime setup of the environment variable, I can probably do that this weekend.

jmert: I've switched to patching the build to ignore the build `$prefix` in setting the default `__EGL_VENDOR_LIBRARY_DIRS` path, as a way to avoid the `__init__()` method from being needed.

jmert:
The [previous set of builds](https://buildkite.com/julialang/yggdrasil/builds/12137) didn't include FreeBSD on aarch64, so that failure is new — looks to be because the X11 library stack is not available for FreeBSD on non-x86_64. Should this add the restriction, or should new builds for the dependencies be triggered to add builds? (I haven't yet gone looking to see if there's a known issue or not to adding those builds.)

PR #9160, opened 2024-07-27
[GLFW] Require libdecor_jll for Wayland compatibility

jmert:
GLFW 3.4 uses libdecor to draw client-side decorations when interacting with a Wayland display.


imciner2:
Are you sure this is using libdecor? I don't see anything mentioned on the command lines in the compilation about linking/including `decor`.

jmert:
> Are you sure this is using libdecor? I don't see anything mentioned on the command lines in the compilation about linking/including `decor`.

Yes, GLFW loads the library dynamically. Without the added `libdecor_jll`, I get a window without any decorations/controls:
image

But with the `libdecor_jll` dependency added, I instead get a window which has client-side decorations drawn:
image

PR #9165, opened 2024-07-29
[terminfo-db] Add a builder for the terminfo database

ararslan:
This uses the directory-based database rather than the single Berkeley database (terminfo.db). The license is from ncurses, which is close enough; see https://invisible-island.net/ncurses/terminfo.src.html.

Ideally this would be shipped as one of Julia's dependencies to ensure that the custom terminfo parser actually works on all platforms, as not all have the terminfo database.

### DIFF
--- a/T/TermInfoDB/build_tarballs.jl
+++ b/T/TermInfoDB/build_tarballs.jl
@@ -32,6 +32,7 @@ tic -sx -o "${prefix}/share/terminfo" ./terminfo-*
 pushd "${prefix}/share/terminfo/"
 for dir in *; do
     lcdir="${dir,,}"
+    mkdir -vp "${lcdir}"
     for file in ${dir}/*; do
         lcfile="${file,,}"
         if [ "${dir}/${file}" = "${lcdir}/${lcfile}" ]; then
--- a/T/TermInfoDB/build_tarballs.jl
+++ b/T/TermInfoDB/build_tarballs.jl
@@ -55,7 +55,7 @@ find "${prefix}/share/terminfo/" -type d -empty -delete
 install_license "${WORKSPACE}/srcdir/COPYING"
 """
 
-platforms = supported_platforms()
+platforms = [AnyPlatform()]
 
 # I'd rather not list out >2k individual `FileProduct`s so let's just call the entry
 # for xterm our "product" and hope the rest are there
--- a/T/TermInfoDB/build_tarballs.jl
+++ b/T/TermInfoDB/build_tarballs.jl
@@ -13,6 +13,11 @@ sources = [
 
 script = raw"""
 cd ${WORKSPACE}/srcdir/
+
+# Install `tree` to get an overview of what's going to be included in the JLL
+apk update
+apk add tree
+
 gunzip terminfo-*
 mkdir -p "${prefix}/share/terminfo"
 tic -sx -o "${prefix}/share/terminfo" ./terminfo-*
--- a/T/TermInfoDB/build_tarballs.jl
+++ b/T/TermInfoDB/build_tarballs.jl
@@ -14,6 +14,7 @@ sources = [
 script = raw"""
 cd ${WORKSPACE}/srcdir/
 gunzip terminfo-*
+mkdir -p "${prefix}/share/terminfo"
 tic -sx -o "${prefix}/share/terminfo" ./terminfo-*
 
 # When ignoring case, these files are duplicates of others. We'll remove them to ensure
--- a/T/TermInfoDB/build_tarballs.jl
+++ b/T/TermInfoDB/build_tarballs.jl
@@ -47,10 +47,10 @@ DUPS=(
     p/p9-w
 )
 for file in "${DUPS[@]}"; do
-    rm -f "${prefix}/share/terminfo/${file}"
+    rm -fv "${prefix}/share/terminfo/${file}"
 done
 # Remove empty directories
-find "${prefix}/share/terminfo/" -type d -empty -delete
+find "${prefix}/share/terminfo/" -type d -empty -print -delete
 
 install_license "${WORKSPACE}/srcdir/COPYING"
 """
--- a/T/TermInfoDB/build_tarballs.jl
+++ b/T/TermInfoDB/build_tarballs.jl
@@ -17,40 +17,27 @@ gunzip terminfo-*
 mkdir -p "${prefix}/share/terminfo"
 tic -sx -o "${prefix}/share/terminfo" ./terminfo-*
 
-# When ignoring case, these files are duplicates of others. We'll remove them to ensure
-# they don't cause trouble on case-insensitive filesystems
-DUPS=(
-    2/2621a
-    e/eterm
-    e/eterm-color
-    h/hp2621a
-    h/hp70092a
-    l/lft-pc850
-    n/ncr260vt300wpp
-    n/ncrvt100wpp
-    p/p12
-    p/p12-m
-    p/p12-m-w
-    p/p12-w
-    p/p14
-    p/p14-m
-    p/p14-m-w
-    p/p14-w
-    p/p4
-    p/p5
-    p/p7
-    p/p8
-    p/p8-w
-    p/p9
-    p/p9-8
-    p/p9-8-w
-    p/p9-w
-)
-for file in "${DUPS[@]}"; do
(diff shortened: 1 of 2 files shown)


T/TermInfoDB/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+for file in "${DUPS[@]}"; do
+    rm -f "${prefix}/share/terminfo/${file}"
+done
+# Remove empty directories
+find "${prefix}/share/terminfo/" -type d -empty -delete
+
+install_license "${WORKSPACE}/srcdir/COPYING"
+"""
+
+platforms = supported_platforms()
```
giordano, line 58: Aren't these all the same?  I sampled 2-3 platforms and the tree hash was always `df5c85191d884419d5edb6087986a5df6e642e26`.  If so, you can build a single artifact with
```suggestion
platforms = [AnyPlatform()]
```

T/TermInfoDB/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/
+gunzip terminfo-*
+mkdir -p "${prefix}/share/terminfo"
+tic -sx -o "${prefix}/share/terminfo" ./terminfo-*
+
+# When ignoring case, these files are duplicates of others. We'll remove them to ensure
+# they don't cause trouble on case-insensitive filesystems
+DUPS=(
```
giordano, line 22: There are single-letter files which are still problematic: https://buildkite.com/julialang/yggdrasil/builds/12154#0190ff34-b5bd-41c5-b1e4-3b1109eb8df8/673-903

T/TermInfoDB/build_tarballs.jl
ararslan: They are indeed the same. TIL `AnyPlatform`, thanks!

T/TermInfoDB/build_tarballs.jl
ararslan: I see the warning from the auditor but when I download the artifact and look at the contents, I don't see the things it's complaining about 🤔

T/TermInfoDB/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+    p/p7
+    p/p8
+    p/p8-w
+    p/p9
+    p/p9-8
+    p/p9-8-w
+    p/p9-w
+)
+for file in "${DUPS[@]}"; do
+    rm -f "${prefix}/share/terminfo/${file}"
```
giordano, line 50:
```suggestion
    rm -fv "${prefix}/share/terminfo/${file}"
```

T/TermInfoDB/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+done
+# Remove empty directories
+find "${prefix}/share/terminfo/" -type d -empty -delete
```
giordano, line 53:
```suggestion
find "${prefix}/share/terminfo/" -type d -empty -print -delete
```

giordano (commented):
Make log more verbose.

T/TermInfoDB/build_tarballs.jl
ararslan: I guess there aren't any empty directories after all; the log doesn't show any output for this step.

giordano:
I'm confused now: `find` doesn't seem to find (nor delete) anything, so how come the duplicate directories disappear?

ararslan:
I wish I knew...

ararslan:
Okay wtf:
```
---> mv -fv "${dir}/${file}" "${lcdir}/${lcfile}"
'2/2621A' -> '2/2621a'
```
But then
```
├── 2
│   ├── 2621
│   ├── 2621-wl
│   ├── 2621A
│   └── 2621a
```
and
```
┌ Warning: share/terminfo/2/2621a causes a case-sensitivity ambiguity!
```

ararslan:
Am I going crazy or does this make no sense? (I guess those aren't mutually exclusive...) Also very possible that how I've gone about the case handling is incorrect but it seems like it should work as written.

giordano:
For the record, it turns out the database contains lots of hardlinked files which share the same inodes (you can check it out with `ls -li`), which makes moving or copying them harder than it should be...

T/TermInfoDB/build_tarballs.jl
```diff
@@ -0,0 +1,72 @@
...
+tic -sx -o "${prefix}/share/terminfo" ./terminfo-*
+
+# The terminfo filesystem-based database contains both upper- and lowercase directory
+# names, which presents a problem for case-insensitive filesystems. Let's rename all
+# files and directories to lowercase, dealing with conflicts by overwriting and hoping
+# for the best.
+pushd "${prefix}/share/terminfo/"
+for dir in *; do
+    lcdir="${dir,,}"
+    for file in ${dir}/*; do
```
giordano, line 35: Do you need to make sure the `lcdir` exists?
```suggestion
    mkdir -vp "${lcdir}"
    for file in ${dir}/*; do
```

ararslan: I don't know, do I? I supposed I probably do

T/TermInfoDB/build_tarballs.jl
```diff
@@ -0,0 +1,73 @@
...
+
+# The terminfo filesystem-based database contains both upper- and lowercase directory
+# names, which presents a problem for case-insensitive filesystems. Let's rename all
+# files and directories to lowercase, dealing with conflicts by overwriting and hoping
+# for the best.
+pushd "${prefix}/share/terminfo/"
+for dir in *; do
+    lcdir="${dir,,}"
+    mkdir -vp "${lcdir}"
+    for file in ${dir}/*; do
```
giordano, line 36: Build works for me with this change:
```suggestion
    for file in ${dir}/*; do
    file=$(basename "${file}")
```

T/TermInfoDB/build_tarballs.jl
```diff
@@ -0,0 +1,73 @@
...
+        if [ -e "${dir}/${file}" ] || [ -e "${lcdir}/${lcfile}.TEMP" ]; then
+            echo "ERROR: '${dir}/${file}' not successfully renamed to lowercase!!!"
+            exit 1
+        fi
+    done
+    if [ -z "$(ls -A "${dir}")" ]; then
+        rm -rfv "${dir}"
+    fi
+done
+tree
```
giordano, line 54: Once build is successful, we can remove debug `tree`
```suggestion
```

T/TermInfoDB/build_tarballs.jl
```diff
@@ -0,0 +1,73 @@
...
+script = raw"""
+cd ${WORKSPACE}/srcdir/
+
+# Adopting the wisdom of https://mywiki.wooledge.org/ParsingLs (🙏 Mosè)
+shopt -s nullglob
+
+# Install `tree` to get an overview of what's going to be included in the JLL
+apk update
+apk add tree
+
```
giordano, line 23:
```suggestion
```

T/TermInfoDB/build_tarballs.jl
```diff
@@ -0,0 +1,84 @@
...
+install_license "${WORKSPACE}/srcdir/COPYING"
+"""
+
+platforms = [AnyPlatform()]
+
+# I'd rather not list out >2k individual `FileProduct`s so let's just call the entry
+# for xterm our "product" and hope the rest are there
+products = [
+    FileProduct("share/terminfo/x/xterm", :terminfo_xterm),
```
tecosaur, line 76: Can `share/terminfo` be the "file" product?

ararslan: I wondered that as well. I had assumed not but I just checked the implementation of `FileProduct` and it only requires `ispath`, not `isfile`, which suggests that it _should_ work. @giordano, is there any reason why we couldn't do what @tecosaur noted?

giordano: Yeah, I think that'd work as well

ararslan: How does versioning work if we change this?

giordano: If you don't change the version number, Pkg will only install the latest one.

PR #9170, opened 2024-07-30
Limit number of workers for LLVMExtra

pxl-th:
As mentioned here: https://github.com/JuliaPackaging/Yggdrasil/pull/9164#issuecomment-2256342740


giordano:
Thanks!

giordano:
Uhm, this package doesn't have a license

giordano:
Ok, thanks!  The problem is that we were not installing them, I added the license installation step, so now we should be good.

PR #9179, opened 2024-07-31
[HiGHS] update to v1.7.2

senhalil:
Tagging @odow @imciner2 as they were the last ones who updated HiGHS.


imciner2:
It looks like the `-DJULIA` option was actually removed in 1.7.0 (https://github.com/ERGO-Code/HiGHS/commit/90ff3fd1874e9d2b5d3c71780be53636087dbb58), but we didn't catch that. Can you modify the common build script to remove that option in the CMake command since it is now no longer present in the upstream code?

odow (approved):

senhalil:
Thank you both for the rapid turnaround 🙇

PR #9185, opened 2024-08-01
Fix the build option for LAMMPS to actually enable MPI

vchuravy:
As noted in https://github.com/cesmix-mit/LAMMPS.jl/issues/59#issuecomment-2263422347

### DIFF
--- a/L/LAMMPS/build_tarballs.jl
+++ b/L/LAMMPS/build_tarballs.jl
@@ -86,7 +86,7 @@ platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && Sys.isfreebsd(p)), plat
 platforms = filter(p -> !(Sys.isfreebsd(p) || libc(p) == "musl"), platforms)
 
 platforms = expand_cxxstring_abis(platforms)
-platforms = expand_gfortran_versions(platforms) # ABI dependent due to MPI
+# platforms = expand_gfortran_versions(platforms) # ABI dependent due to MPI
 
 # The products that we will ensure are always built
 products = [
--- a/L/LAMMPS/build_tarballs.jl
+++ b/L/LAMMPS/build_tarballs.jl
@@ -39,7 +39,7 @@ cmake -C ../cmake/presets/most.cmake -C ../cmake/presets/nolib.cmake ../cmake -D
     -DCMAKE_BUILD_TYPE=Release \
     -DBUILD_SHARED_LIBS=ON \
     -DLAMMPS_EXCEPTIONS=ON \
-    -DPKG_MPI=ON \
+    -DBUILD_MPI=ON \
     -DPKG_EXTRA-FIX=ON \
     -DPKG_ML-SNAP=ON \
     -DPKG_ML-PACE=ON \
--- a/L/LAMMPS/build_tarballs.jl
+++ b/L/LAMMPS/build_tarballs.jl
@@ -86,7 +86,7 @@ platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && Sys.isfreebsd(p)), plat
 platforms = filter(p -> !(Sys.isfreebsd(p) || libc(p) == "musl"), platforms)
 
 platforms = expand_cxxstring_abis(platforms)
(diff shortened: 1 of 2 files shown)


L/LAMMPS/build_tarballs.jl
```diff
@@ -75,18 +75,19 @@ augment_platform_block = """
 # platforms are passed in on the command line
 # platforms = supported_platforms(; experimental=true)
 platforms = supported_platforms()
-platforms = expand_cxxstring_abis(platforms)
-
 platforms, platform_dependencies = MPI.augment_platforms(platforms)
 # Avoid platforms where the MPI implementation isn't supported
 # OpenMPI
-platforms = filter(p -> !(p["mpi"] == "openmpi" && arch(p) == "armv6l" && libc(p) == "glibc"), platforms)
+platforms = filter(p -> !(p["mpi"] == "openmpi" && nbits(p) == 32), platforms)
```
giordano, line 81: You don't need to filter out 32-bit platforms if you build against OpenMPI 4: https://github.com/JuliaPackaging/Yggdrasil/blob/27db5ba5a624df4809faef80a12297ccbc570ab4/H/HDF5/build_tarballs.jl#L245.  However, one issue I never had the time to look at is #8211 (this compat doesn't seem to be propagated correctly)

L/LAMMPS/build_tarballs.jl
```diff
@@ -75,18 +75,19 @@ augment_platform_block = """
...
 # MPItrampoline
 platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && libc(p) == "musl"), platforms)
 platforms = filter(p -> !(p["mpi"] == "mpitrampoline" && Sys.isfreebsd(p)), platforms)
 
 platforms = filter(p -> !(Sys.isfreebsd(p) || libc(p) == "musl"), platforms)
 
+platforms = expand_cxxstring_abis(platforms)
+platforms = expand_gfortran_versions(platforms) # ABI dependent due to MPI
```
giordano, line 89: We don't seem to have to do it for other packages which depend on MPI, see for example https://github.com/JuliaPackaging/Yggdrasil/blob/27db5ba5a624df4809faef80a12297ccbc570ab4/A/ADIOS2/build_tarballs.jl

vchuravy, line 89:
```suggestion
# platforms = expand_gfortran_versions(platforms) # ABI dependent due to MPI
```

vchuravy:
@giordano 

```
[ Info: Checking shared library lib/liblammps.so.0
ERROR: could not load library "/cache/build/yggy-amdci7-3/julialang/yggdrasil/L/LAMMPS/build/x86_64-linux-gnu-cxx11-mpi+mpitrampoline/i3ZWjkPZ/x86_64-linux-gnu-libgfortran5-cxx11-mpi+mpitrampoline/destdir/lib/liblammps.so.0"
libgfortran.so.5: cannot open shared object file: No such file or directory
┌ Warning: lib/liblammps.so.0 cannot be dlopen()'ed
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/mzSV7/src/Auditor.jl:173
[ Info: lib/liblammps.so.0 already has SONAME "liblammps.so.0"
```

any ideas?

L/LAMMPS/build_tarballs.jl
vchuravy: How does: ` OpenMPI_compat="4.1.6, 5" ` work? Do we not need to build against OpenMPI v4 and OpenMPI v5 separately?

giordano: the compat forces the lower version as build version, for OpenMPI we don't need separate builds because the two major versions are ABI-compatible

giordano:
Uhm, maybe wrong RPATH?

vchuravy:
It's weird since it is just this one build with mpitrampoline cc: @eschnett

PR #9187, opened 2024-08-02
[MKL_Pardiso] Add a standalone library

amontoison:


M/MKL_Pardiso/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+
+sources = [
+    GitSource("https://github.com/amontoison/MKL_Pardiso.git", "cb70bc69c426a67ef9e4dd0fd526963d594600fb")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/MKL_Pardiso
+gcc -fPIC -shared -o ${libdir}/libmkl_pardiso_shim.${dlext} mkl_pardiso_shim.c -L${libdir} -lmkl_rt
+gcc -fPIC -shared -o ${libdir}/libmkl_pardiso.${dlext} mkl_pardiso.c -L${libdir} -lmkl_pardiso_shim
```
imciner2, line 14:
```suggestion
${CC} -fPIC -shared -o ${libdir}/libmkl_pardiso_shim.${dlext} mkl_pardiso_shim.c -L${libdir} -lmkl_rt
${CC} -fPIC -shared -o ${libdir}/libmkl_pardiso.${dlext} mkl_pardiso.c -L${libdir} -lmkl_pardiso_shim
```

imciner2 (commented):
Why does this need to be two shared libraries that chained?

amontoison:
> Why does this need to be two shared libraries that chained?

@imciner2
Because I didn't find a way to trick the linker to reexport the same symbols.
The first library `libmkl_pardiso_shim` just add a prefix `mkl_` to all symbols of MKL Pardiso and the second library `libmkl_pardiso` restores the original symbols without the prefix.

amontoison:
Give me a few days to test that it still works as intended for the Fortran routines.

ViralBShah:
Ok to close?

PR #9194, opened 2024-08-03
Add a builder for the AWS command line interface v2

ararslan:
This is based in part on https://github.com/aws/aws-cli/discussions/7168, which is for FreeBSD but should apply to any other platform as well (or so I would assume). It differs from the process described in the discussion in that it uses a provided installation script rather than the development build instructions in the README, which is all the process in the discussion is, except with some particular requisite system package installs. Notable among the installed packages are CMake and Rust, which is why I've added Rust to the `compilers` list and defined some CMake environment variables.

### DIFF
--- a/A/AWSCLI/build_tarballs.jl
+++ b/A/AWSCLI/build_tarballs.jl
@@ -1,0 +1,31 @@
+using BinaryBuilder
+
+name = "AWSCLI"
+version = v"2.17.22"
+
+sources = [
+    GitSource("https://github.com/aws/aws-cli.git",
+              "18d64caf588186c396dd58ff2346a8ca9657fdcb"),
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/aws-cli
+
+export CMAKE_INSTALL_PREFIX="${prefix}"
+export CMAKE_TOOLCHAIN_FILE="${CMAKE_TARGET_TOOLCHAIN}"
+
+./scripts/install -i "${bindir}"
+
+install_license ./LICENSE.txt
+"""
+
+platforms = supported_platforms()
+
+products = [
+    ExecutableProduct("aws", :awscli),
+]
+
+dependencies = Dependency[]
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               compilers=[:c, :rust], julia_compat="1.6")


A/AWSCLI/build_tarballs.jl
```diff
@@ -0,0 +1,40 @@
...
+cd ${WORKSPACE}/srcdir/aws-cli
+
+export CMAKE_INSTALL_PREFIX="${prefix}"
+export CMAKE_TOOLCHAIN_FILE="${CMAKE_TARGET_TOOLCHAIN}"
+
+# Breaking change in Clang 16 causes failure while building dependencies on macOS and FreeBSD
+# See https://discourse.llvm.org/t/clang-16-notice-of-potentially-breaking-changes/65562
+export CFLAGS="${CFLAGS} -Wno-error=incompatible-function-pointer-types"
+
+PYTHON="${bindir}/python3" ./configure --prefix=${prefix} --with-download-deps --with-install-type=portable-exe
```
giordano, line 21: If they need to run code with `PYTHON`, this isn't going to work because `${bindir}/python3` is target's python and (in general) we can't run it.  There should be a working `python3` in the system, but I don't know what they want to do with python at all.

giordano: Mixing up autotools and cmake?  what can possibly go wrong?

ararslan: I had been using the system Python but ran into an issue I didn't understand. Hadn't occurred to me that it'd need to run Python, that definitely won't work. 🤦

giordano: On the few platforms where we can actually run target code (e.g. https://buildkite.com/julialang/yggdrasil/builds/12639#01917bcf-56fa-4c1b-b4ab-d0d3f8f00321)
```
[20:39:03] configure: error: AWS CLI requires a Python interpreter compiled with sqlite3. /workspace/destdir/bin/python3 does not have sqlite3.
```

This package looks like a Python package, just with a fancy build system.  I don't really think we aim to be a Python package manager.

ararslan: The developers consider the fact that it's written in Python to be an "implementation detail," which is why they don't distribute it via PyPI; it's intended to just be a command line tool that happens to currently be written in Python rather than an actual Python package.

ararslan: That said, that implementation detail probably means we can't build it with BinaryBuilder. Oh well.

fingolfin:
Based on the discussion here, it seems this is a dead end due to the Python dependency, so I am closing this.

If this assessment is incorrect, just holler and we can re-open.

PR #9198, opened 2024-08-05
[Givaro] Add recipe

laurentbartholdi:
This the the library Givaro for C++ arithmetic, to be used in the coming Spasm PR


G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Givaro"
+version = v"4.2.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/linbox-team/givaro.git", "fc6cac7820539c900dde332326c71461ba7b910b"),
```
fingolfin, line 10: This is not [version 4.2.0](https://github.com/linbox-team/givaro/releases/tag/v4.2.0) (released Dec 14, 2021), but rather https://github.com/linbox-team/givaro/commit/fc6cac7820539c900dde332326c71461ba7b910b (a commit from March 2024), the current master branch.

Is that really necessary? If it is, then I think at the very least this should be pointed out in a comment. E.g.
```suggestion
    # The following points to a snapshot of the Givaro branch from March 2024
    # instead of 4.2.0 (the last release) because ... TODO INSERT REASON
    GitSource("https://github.com/linbox-team/givaro.git", "fc6cac7820539c900dde332326c71461ba7b910b"),
```

Of course ideally the givaro folks would be convinced to just tag a new release (seems overdue anyway) so that here a tarball can be referenced instead of a git commit. However based on the open PRs there I guess this might be hard, so I would not want to require this here.

G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p0 ${f}
+done
+
+autoreconf -i
+./configure CCNAM=${CC} CPLUS_INCLUDE_PATH=$includedir --prefix=$prefix --build=${MACHTYPE} --host=${target}
+
+# really ugly! but I see no other solution for now.
+if echo ${target} | grep -q aarch64-apple-darwin; then
```
fingolfin, line 26:
```suggestion
if [[ ${target} != aarch64-apple-darwin* ]]; then
```

G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+cd ${WORKSPACE}/srcdir/givaro
+
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p0 ${f}
+done
+
+autoreconf -i
+./configure CCNAM=${CC} CPLUS_INCLUDE_PATH=$includedir --prefix=$prefix --build=${MACHTYPE} --host=${target}
+
+# really ugly! but I see no other solution for now.
```
fingolfin, line 25: This comment is not very helpful, as it does not state what the problem is for which you see no other solution? 

Not knowing that it is also hard to suggest alternatives

G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+make -j ${nproc}
+make install
+
+install_license LICENSE
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [p for p=supported_platforms() if os(p) != "windows"] # life is too short
```
fingolfin, line 38: More idiomatic would be:
```suggestion
platforms = supported_platforms()                                                       
filter!(!Sys.iswindows, platforms)
```

G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+    FileProduct("include/givaro-config.h", :givaro_config_h)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("GMP_jll"; compat="6.2.1")
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.9")
```
fingolfin, line 52: Why 1.9 and not e.g. 1.6 (the current LTS)?

G/Givaro/bundled/patches/zap-gmp-test.patch
```diff
@@ -0,0 +1,30 @@
...
+-				return (__GNU_MP_RELEASE < $min_gmp_release);
+-			}
+-		],
+-		[ AC_MSG_RESULT(yes)
+-		],
+-		[ AC_MSG_RESULT(no)
+-		  AC_MSG_ERROR(your GMP is too old. GMP release >= $min_gmp_release needed)
+-		  exit 1]
+-	)
++	### Major hack! we need to add -lrt sometimes, let's do it here
```
fingolfin, line 23: Why do you need to add `-lrt`? And for what? Is this related to GMP at all? The comment makes it sound as if that was arbitrary?

G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Givaro"
+version = v"4.2.0"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/linbox-team/givaro/releases/download/v4.2.0/givaro-4.2.0.tar.gz", "865e228812feca971dfb6e776a7bc7ac959cf63ebd52b4f05492730a46e1f189"),
```
fingolfin, line 10:
```suggestion
    ArchiveSource("https://github.com/linbox-team/givaro/releases/download/v$(version)/givaro-$(version).tar.gz", "865e228812feca971dfb6e776a7bc7ac959cf63ebd52b4f05492730a46e1f189"),
```

laurentbartholdi:
If someone knows how to get compilation to work on x86_64-linux, that would be great! The problem is in the configure script, more precisely the check that linking with -lgmp works. Because of some incompatibility, the configure script aborts with error message "undefined symbol memcpy@GLIBC_2.14", which occurs when linking a conftest.cpp file.

G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/linbox-team/givaro/releases/download/v$version/givaro-$version.tar.gz", "865e228812feca971dfb6e776a7bc7ac959cf63ebd52b4f05492730a46e1f189"),
+    DirectorySource("./bundled")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/givaro-4.2.0
```
giordano, line 16:
```suggestion
cd ${WORKSPACE}/srcdir/givaro*
```

G/Givaro/bundled/patches/libtool-aarch64-apple-darwin.hack
giordano, line 1: Can you please rename this file with extension `.patch` to make the file more readable?  But also, I'm not convinced this is needed at all.

G/Givaro/bundled/patches/disable_bool_constant.patch
giordano, line 1: Sounds like this patch should be upstreamed?

G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+    FileProduct("include/givaro-config.h", :givaro_config_h)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("GMP_jll"; compat="6.2.1")
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
giordano, line 55: Missing newline
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")

```

G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+autoreconf
+./configure CCNAM=${CC} CPLUS_INCLUDE_PATH=$includedir --prefix=$prefix --build=${MACHTYPE} --host=${target}
+
+# really ugly! but I see no other solution for now.
+# this patch modifies libtool so that it doesn't use the not-yet-implemented "-r" flag for the linker, replacing it by -dynamiclib
+if [[ ${target} =~ aarch64-apple-darwin* ]]; then
+    echo "Patching libtool so it doesn't use ld -r"
+    patch -p0 < ${WORKSPACE}/srcdir/patches/libtool-aarch64-apple-darwin.hack
+fi
+
```
giordano, line 31: Isn't it easier not to use `lld` with `clang`?  Which means, add the option `clang_use_lld=false` to `build_tarballs` below.

G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+make -j ${nproc}
+make install
+
+install_license Licence_CeCILL-B_V1-en.txt
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [Platform("aarch64","macos"), Platform("x86_64","macos")]
+# building on x86_64-linux fails with error "undefined reference to `memcpy@GLIBC_2.14'" when linking gmp
```
giordano, line 41: When else fails, [use the search](https://github.com/search?q=repo%3AJuliaPackaging%2FYggdrasil+memcpy%40GLIBC_2.14&type=pullrequests) and discover you need a newer version of GCC (5 or 6 probably)

G/Givaro/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+
+# The products that we will ensure are always built
+products = Product[
```
giordano, line 44:
```suggestion
products = [
```

laurentbartholdi:
Hi @giordano ,
Just to know: is there something I should do to have the recipe integrated, or should I just wait?

PR #9207, opened 2024-08-06
[SuiteSparse] Update to 7.8.0 and use common build script

imciner2:
New SuiteSparse release 7.8.0: https://github.com/DrTimothyAldenDavis/SuiteSparse/releases/tag/v7.8.0.

I don't think there are major changes to anything in the core parts we build here, but there might need to be some tweaks to the wrappers (it looks like cholmod at least gained a new function).


imciner2:
Do we want to keep the 1.11 Julia compat here, or bump it to 1.12?

giordano:
I think the recent suggestion was not to touch the Julia compat, unless strictly necessary: with the hypothetical upgradable stdlibs, later versions could be installed.

PR #9216, opened 2024-08-07
[argp_standalone] Upgrade to 1.4.1 and build for all platforms

laurentbartholdi:
This is my previous PR laurentbartholdi-argp merged into the previous recipe argp_standalone


laurentbartholdi:
Ouch, I now see my PR message doesn't even say it's about the argp_standalone / Argp issue discussed with @imciner2

A/argp_standalone/build_tarballs.jl
```diff
@@ -3,30 +3,38 @@
...
+done
+
+autoreconf -i
 CFLAGS="-fPIC" ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
 make -j${nproc}
-install_license $WORKSPACE/srcdir/LICENSE
-install -D -m644 argp.h ${includedir}/argp.h
-install -D -m755 libargp.a ${libdir}/libargp.a
+install -Dvm 644 argp.h -t $prefix/include
+install -Dvm 644 libargp.a -t $prefix/lib
```
giordano, line 26:
```suggestion
install -Dvm 644 argp.h -t "${includedir}"
install -Dvm 644 libargp.a -t "${libdir}"
```

A/argp_standalone/build_tarballs.jl
```diff
@@ -3,30 +3,38 @@
...
+
+install_license ${WORKSPACE}/srcdir/argp-standalone/README.md
 """
 
-# Select Unix platforms
-platforms = filter(p->Sys.islinux(p) && libc(p) == "musl", supported_platforms(;experimental=true))
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(;experimental=true)
```
giordano, line 33: `experimental` does exactly nothing nowadays, is only kept for compatibility reasons
```suggestion
platforms = supported_platforms()
```

A/argp_standalone/build_tarballs.jl
```diff
@@ -3,30 +3,38 @@
...
 
 # The products that we will ensure are always built
-products = [
+products = Product[
```
giordano, line 36:
```suggestion
products =[
```

A/argp_standalone/bundled/patches/fix-prototype-and-random.patch
```diff
@@ -0,0 +1,53 @@
...
+diff --git a/argp-test.c b/argp-test.c
+index d9c108d..d32d509 100644
+--- a/argp-test.c
++++ b/argp-test.c
+@@ -205,7 +205,7 @@ main (int argc, char **argv)
+ {
+   struct params params;
+   params.foonly = 0;
+-  params.foonly_default = random ();
++  params.foonly_default = 0xdeadbeef;
```
giordano, line 25: Not very random.  Where are all these patches coming from?

A/argp_standalone/build_tarballs.jl
laurentbartholdi: ${libdir} is wrong on windows builds, they would go in $prefix/bin rather than $prefix/lib. I guess that ${libdir} is meant for dynamic libraries.

A/argp_standalone/bundled/patches/fix-prototype-and-random.patch
laurentbartholdi: Well, random() isn't really needed at all, but it's not in the standard library so doesn't compile on windows. No randomness is needed anyways, and there's a single call to random() in the whole system, so this is just obfuscation. I could replace random() by the portable rand() / srand() but I don't see the point.

A/argp_standalone/build_tarballs.jl
```diff
@@ -3,30 +3,38 @@
...
+autoreconf -i
 CFLAGS="-fPIC" ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
 make -j${nproc}
-install_license $WORKSPACE/srcdir/LICENSE
-install -D -m644 argp.h ${includedir}/argp.h
-install -D -m755 libargp.a ${libdir}/libargp.a
+install -Dvm 644 argp.h -t $prefix/include
+install -Dvm 644 libargp.a -t $prefix/lib
+
+install_license ${WORKSPACE}/srcdir/argp-standalone/README.md
```
fingolfin, line 28: That's not a valid license file. It does state what the license is, though, and its SPDX identifier `LGPL-2.1-or-later`, so perhaps this is enough, @giordano ?

A/argp_standalone/bundled/patches/add-prototypes.patch
giordano, line 1: If this is needed for musl platforms, using gcc 6 should be sufficient instead.

A/argp_standalone/build_tarballs.jl
giordano: While a proper license file would be better, a document stating what's the license would be OK-ish, if that's the only thing upstream is doing anyway. But also, we ship the text of that licence under `/usr/share/licenses` so you can also just use that and solve all issues.

laurentbartholdi:
I don't see why the checks were unsuccessful: looking at the logs on buildkite, the last lines are
```
2024-10-16 13:04:41 INFO   Artifact uploads completed successfully
--
  | 🚨 Error: The plugin coppermind post-command hook exited with status 1
```

However, there doesn't seem to have been a post-command hook, and the artifacts are properly compiled (with lots of warnings, though).

Any ideas where this comes from?

PR #9219, opened 2024-08-08
[libiconv] Disable build on glibc platforms

staticfloat:
Glibc contains its own `iconv` implementation, and `libiconv` is meant to be used only on non-glibc platforms.  Therefore, let us disable building libiconv for glibc platforms, to avoid issues where dependencies accidentally build against the wrong implementation.

### DIFF
--- a/L/Libiconv/build_tarballs.jl
+++ b/L/Libiconv/build_tarballs.jl
@@ -37,9 +37,8 @@ Cflags: -I\${includedir}
 EOF
 """
 
-# These are the platforms we will build for by default, unless further
-# platforms are passed in on the command line
-platforms = supported_platforms(; experimental=true)
+# Build for non-glibc platforms, as `glibc` contains its own `iconv` implementation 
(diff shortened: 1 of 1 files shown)


L/Libiconv/build_tarballs.jl
```diff
@@ -37,9 +37,8 @@ Cflags: -I\${includedir}
 EOF
 """
 
-# These are the platforms we will build for by default, unless further
-# platforms are passed in on the command line
-platforms = supported_platforms(; experimental=true)
+# Build for non-glibc platforms, as `glibc` contains its own `iconv` implementation 
+platforms = filter(p -> libc(p) != "glibc", supported_platforms(; experimental=true))
```
giordano, line 41:
```suggestion
platforms = filter(p -> libc(p) != "glibc", supported_platforms())
```

staticfloat:
Converting to draft until I do some testing to see if this breaks everything that depends on it

ViralBShah:
Let's reopen this if still needs to be done. I have generally seen libiconv being used in packages, and I don't think we have seen complaints.

PR #9220, opened 2024-08-09
Add support for FreeBSD AArch64

ararslan:
This uses FreeBSD 13.2 to match the setup for x86_64. When adding support for FreeBSD AArch64 to Julia, I was using FreeBSD 14.1, so this may turn out to be a different beast altogether. I guess we'll find out.

Trying to remember what all it is I need to do in order to add a new platform... I think there was something I needed to do that made a tag with release assets for all of the platforms. 🤔

### DIFF
--- a/0_RootFS/PlatformSupport/build_tarballs.jl
+++ b/0_RootFS/PlatformSupport/build_tarballs.jl
@@ -51,11 +51,19 @@ sources = [
                   "b5de28fd594a01edacd06e53491ad0890293e5fbf98329346426cf6030ef1ea6"),
     ArchiveSource("https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v7.0.0.tar.bz2",
                   "aa20dfff3596f08a7f427aab74315a6cb80c2b086b4a107ed35af02f9496b628"),
-    ArchiveSource("http://ftp-archive.freebsd.org/pub/FreeBSD-Archive/old-releases/amd64/13.2-RELEASE/base.txz",
-                  "3a9250f7afd730bbe274691859756948b3c57a99bcda30d65d46ae30025906f0"),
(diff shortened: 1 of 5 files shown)


giordano (approved):
Looks good to me, but you'll have to deploy PlatformSupport (and all the compilers), if you want them, and then open the corresponding PR to BinaryBuilderBase 🙂

ararslan:
Accompanying BinaryBuilderBase PR: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/387

Things to build:
- [x] PlatformSupport
- [ ] GCCBootstrap
    - [ ] 4
    - [ ] 5
    - [ ] 6
    - [x] 7-13
- [ ] Rust
- [ ] (Anything else?)

Current issues:
- It appears gcc 4 doesn't support FreeBSD AArch64 (and vice versa); the FreeBSD port for gcc 4.8 is (/was, as it's been deleted) only installable on x86-64 and i686
- I forget what the problem was with gcc 5 was but that at least had a FreeBSD port for AArch64
- gcc 6 complained about not being able to find gmp, mpfr, and mpc 🤔
- FreeBSD AArch64 is tier 3 for Rust and doesn't have an official precompiled standard library, not sure what to do about that

giordano:
For GCC, I think it's fine to stick to limited versions, we do the same with aarch64-darwin, but you need to add some logic in BinaryBuilderBase to choose the right version. For Rust, while we have a build for i686 mingw, in practice we can't use it so we always have to skip that platform. I guess we'll have to skip aarch64 freebsd, too

ararslan:
I wonder if we can use the compiled FreeBSD AArch64 Rust binaries that FreeBSD's package manager would install on that platform...

ararslan:
Let's stick with the compilers we've got for now. Perhaps someday I'll try to build Rust and/or earlier GCC versions but in the meantime, ¯\\\_(ツ)\_/¯

PR #9227, opened 2024-08-09
[libssh] Update to 0.11.0

JamesWrigley:
Note that setting the doxygen tag has been merged upstream.

### DIFF
--- a/L/libssh/build_tarballs.jl
+++ b/L/libssh/build_tarballs.jl
@@ -65,4 +65,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"5", julia_compat="1.6")
--- a/L/libssh/build_tarballs.jl
+++ b/L/libssh/build_tarballs.jl
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "libssh"
-version = v"0.10.6"
+version = v"0.11.0"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


giordano:
```
[20:46:25] [ 71%] Building C object src/CMakeFiles/ssh.dir/pki_ed25519_common.c.obj
[20:46:25] cd /workspace/srcdir/libssh-0.11.0/build/src && /opt/bin/i686-w64-mingw32-libgfortran3-cxx03/i686-w64-mingw32-gcc --sysroot=/opt/i686-w64-mingw32/i686-w64-mingw32/sys-root/ -DLIBSSH_EXPORTS -D_POSIX_SOURCE @CMakeFiles/ssh.dir/includes_C.rsp -O3 -DNDEBUG -MD -MT src/CMakeFiles/ssh.dir/pki_ed25519_common.c.obj -MF CMakeFiles/ssh.dir/pki_ed25519_common.c.obj.d -o CMakeFiles/ssh.dir/pki_ed25519_common.c.obj -c /workspace/srcdir/libssh-0.11.0/src/pki_ed25519_common.c
[20:46:25] /workspace/srcdir/libssh-0.11.0/src/misc.c: In function ‘ssh_check_username_syntax’:
[20:46:25] /workspace/srcdir/libssh-0.11.0/src/misc.c:2195:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
[20:46:25]      for (size_t i = 0; i < username_len; i++) {
[20:46:25]      ^
[20:46:25] /workspace/srcdir/libssh-0.11.0/src/misc.c:2195:5: note: use option -std=c99 or -std=gnu99 to compile your code
```
A couple of comments:

* the day developers will start setting standards in their build systems as appropriate life will be much better (to be clear, this is an upstream bug, they should be setting the standard when calling the compiler, it'd be nice if you opened an issue there to report it)
* you can work around this issue by compiling with gcc 5, which defaults to `-std=gnu11`

JamesWrigley:
Ah hah, thanks I was getting very confused by that... They actually do set standards, but it looks like it's not done in a very portable way: https://gitlab.com/libssh/libssh-mirror/-/blob/master/CompilerChecks.cmake?ref_type=heads#L19
I'll make a PR to fix it.

giordano:
Yes, using CMake own C standard options is a better way to do it, thanks.

PR #9228, opened 2024-08-09
[Rmath] Manually set `USEGCC`/`USECLANG`

giordano:
Ref: https://github.com/JuliaStats/Rmath-julia/issues/47, fix https://github.com/JuliaStats/Rmath-julia/issues/44.  CC: @ViralBShah.

### DIFF
--- a/R/Rmath/build_tarballs.jl
+++ b/R/Rmath/build_tarballs.jl
@@ -10,12 +10,22 @@ sources = [
 
 script = raw"""
 cd $WORKSPACE/srcdir/Rmath-julia*
-make -j${nproc}
+# The whole `USEGCC`/`USECLANG` business is wrong and backword. Until this this fixed upstream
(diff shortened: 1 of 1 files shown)


giordano:
The x86_64-darwin build now has:
```console
% otool -L libRmath-julia.dylib
libRmath-julia.dylib:
	@rpath/libRmath-julia.dylib (compatibility version 0.0.0, current version 0.0.0)
```

ViralBShah:
What if we just directly call `make CC=${CC}` here and ignore the whole USECLANG / USEGCC bit?

giordano:
That wouldn't work, because https://github.com/JuliaStats/Rmath-julia/blob/18dcd7c259b031c3ce9c275b7dd136585d126017/Make.inc#L21-L28 overrides any outer value of `CC`.

PR #9231, opened 2024-08-09
[Rmath] Update to 0.4.3

ViralBShah:


R/Rmath/build_tarballs.jl
```diff
@@ -1,28 +1,17 @@
...
-   USEGCC=0
-else
-   USECLANG=0
-   USEGCC=1
-fi
-make -j${nproc} USECLANG=${USECLANG} USEGCC=${USEGCC}
-mkdir -p "${libdir}"
-mv src/libRmath-julia.* "${libdir}"
+cd $WORKSPACE/srcdir/Rmath-julia
+make -j${nproc} CC=${CC}
```
giordano, line 13: I think
```suggestion
make -j${nproc}
```
should work, too

R/Rmath/build_tarballs.jl
```diff
@@ -1,28 +1,17 @@
...
+cp -p libRmath-julia.${dlext} ${libdir}
```
giordano, line 14:
```suggestion
install -Dvm 755 libRmath-julia.${dlext} -t ${libdir}
```

R/Rmath/build_tarballs.jl
ViralBShah: I thought that just ends up picking `gcc` everywhere. Is that ok?

giordano: With https://github.com/JuliaStats/Rmath-julia/blob/d560159af0a388d8afaf19f7bf2efc51afcf53d9/src/Makefile#L3-L5 `CC` can also be set as an environment variable, which we do here.  Let's see how it goes without `CC=${CC}`, if gcc is used instead then we can restore it.

R/Rmath/build_tarballs.jl
```diff
@@ -11,7 +11,7 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/Rmath-julia
```
giordano, line 12: I think we now need to enter `src`?
```suggestion
cd $WORKSPACE/srcdir/Rmath-julia/src
```

R/Rmath/build_tarballs.jl
giordano: It's working well, `cc` is used everywhere: https://buildkite.com/julialang/yggdrasil/builds/12397#0191393f-c4b6-4089-835e-97106b705dd8/675-683.  And I can confirm for the x86_64 build:
```console

PR #9235, opened 2024-08-10
[XDiag] fixing OpenBLAS 64 bit interface problem

awietek:
After the build worked correctly and is linked to the 64 bit OpenBLAS library, the code seems to be still compiled with a 32 bit interface. At the same time I am getting segmentation faults, which are likely due to this mismatch. With this PR I am trying to fix this.


awietek:
I am having issues with segmentation faults in the wrapping julia library which come from calls to OpenBLAS. I am trying to link to the 64 bit library (at least that's what I think I am doing) and try to use the ILP64 interface. BLAS / LAPACK is used through armadillo. However, I can check at runtime what the size of the BLAS integer is and it turns out to be only 32 bit when I use the nor registered XDiag_jll package.

ViralBShah:
Good to merge?

awietek:
Yes

PR #9250, opened 2024-08-13
Add in DuckDB shell to match SQLite

frankier:
This was removed, but it could be quite useful to have it accessible whenever the DuckDB library is, e.g. for quick database admin. This is also the same as SQLite.


giordano:
Thanks!

PR #9263, opened 2024-08-15
[OpenBLAS] Backport new threading buffer patch

giordano:
Follow up to #9262, backport https://github.com/OpenMathLib/OpenBLAS/pull/4879.


giordano:
This now seems to work on the machines where I observed segfaults in https://github.com/JuliaLang/julia/pull/55496.  Let's merge and see if the full julia build works as well.

PR #9265, opened 2024-08-15
[XDiag] trying to fix a problem with OpenMP on apple-aarch64

awietek:
I face a problem with OpenMP on apple-aarch64. OpenMP seems to be flawed with the current setup, but it was working with a previous setup which I am hereby trying to fall back to.


giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/6258bc4c81855c8221279fce5d62b409de4ed694/CONTRIBUTING.md#branch-naming

giordano:
Also, it'd be helpful to elaborate what's the issue you faced, since basically all packages using openmp use llvmopenmp on macos, and I'm not aware of anyone reporting issues lately.

awietek:
Yes, of course. Everything works if I set the environment variable OMP_NUM_THREADS=1 , but I get segmentation faults e.g. for OMP_NUM_THREADS=2. This issue is not present on a machine I have with osx and Intel x86 chip.

awietek:
Also I did not have these segmentation faults with the other options before, which I now included again. I guess that would then mean I am linking to another OpenMP not LLVMOpenMP, if I understand this correctly.

giordano:
Have you investigated the issue with a debugger?  Do you have a stacktrace?

X/XDiag/build_tarballs.jl
```diff
@@ -42,7 +42,9 @@ if [[ "${target}" == x86_64-apple-* ]]; then
     export MACOSX_DEPLOYMENT_TARGET=10.14
 fi
 
-cmake -S . \
+
+if [[ "${target}" == *-apple-* ]]; then
```
giordano, line 46: Instead of duplicating build configurations, please do something like
```sh
    FLAGS=(...)
fi
cmake ... \
    ... \
    "${FLAGS[@]}"
```

giordano (dismissed):
If you really want to do this (and I believe it'd be better to investigate the issue you're facing rather than using libgomp on macOS which is known to cause other issues), you have to change the dependencies filters for CompilerSupportLibraries and LLVMOpenMP

awietek:
Okay, to be more precise about the issue, the code currently gives wrong results without segfaults when routed through the XDiag.jl wrapper library only on osx with aarch64 only if OMP_NUM_THREADS != 1. It works on Linux and OSX with x86. Then I tried starting to debug it. The way I went about it, was that I compiled the library locally and replaced the downloaded shared object with the newly compiled library in the Overrides.toml. This then gave me a segfault, which I then narrowed down to be happening to be happening when invoking an OpenMP parallel section using "#pragma omp parallel". Previously I did not have these problems which is why I wanted to resort the previous solution.

I'd be happy though to try and figure out if there is some other issue. Would you have a suggestion what I could do to get to the bottom of this (I might do it tomorrow then, today is a bit late)?

giordano:
After https://github.com/JuliaRegistries/General/pull/113206 is merged you can try updating LLVMOpenMP to v17 in your environment, and see if that helps anything, or if you're in a rush and can't wait for the PR to be merged you can do
```
]add LLVMOpenMP_jll#main
```
right now.

awietek:
I can wait a bit for LLVMOpenMP v17, I'm not totally in a rush. Does https://github.com/JuliaRegistries/General/pull/113206 have the 3 day waiting time as well?

awietek:
So that would just require updating my package dependencies in XDiag.jl and no changes to XDIag_jll.jl, right?

awietek:
Okay, that did change something. The result when running the tests in XDiag.jl on two threads is now not wrong all the time but roughly in half of the runs.

awietek:
Also it's worth noting that I get the same faulty behavior on two different osx aarch64 computers.

awietek:
I think I would for now prefer to resort to the other OpenMP and give the LLVMOpenMP a try if a newer version comes up again. I am not sure how to debug this issue at the moment. I now made the requested changes in the build_tarballs.jl and added a few things in the wrapper. With this I would like to have a new patch version 0.2.1. Thanks for your help

X/XDiag/build_tarballs.jl
awietek: I added this now in the new PR. Thanks for the suggestion

X/XDiag/build_tarballs.jl
```diff
@@ -52,11 +56,10 @@ cmake -S . \
     -DXDIAG_JULIA_WRAPPER=On \
     -DJlCxx_DIR=$prefix/lib/cmake \
     -DBLAS_LIBRARIES=${libdir}/libopenblas64_.${dlext} \
-    -DLAPACK_LIBRARIES=${libdir}/libopenblas64_.${dlext}
+    -DLAPACK_LIBRARIES=${libdir}/libopenblas64_.${dlext} "${OMP_DEFINES[@]}"
```
giordano, line 59:
```suggestion
    -DLAPACK_LIBRARIES=${libdir}/libopenblas64_.${dlext} \
    "${OMP_DEFINES[@]}"
```

giordano:
Please next time open a PR from a branch out of current `master`, always working on a diverging `master` is unsustainable for you, and makes review harder for us, just don't do it.

awietek:
Okay, will do. Thanks again for your help!

PR #9271, opened 2024-08-16
Add ELFIO v3.12

Mellich:


vchuravy:
Why do we need a header only library in Yggdrasil?

Mellich:
As a build dependency. Are there other ways of using header-only libraries without polluting the resulting artifacts?

vchuravy:
Sure you can add it as a second source and not install this header file?

giordano:
We have other header-only packages, to be used as BuildDependency'ies

E/ELFIO/build_tarballs.jl
```diff
@@ -0,0 +1,26 @@
...
+platforms = [AnyPlatform()]
+
+products = [
+    FileProduct("include/elfio/elfio.hpp", :elfio_hpp)
+]
+
+
+dependencies = []
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
giordano, line 26:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")

```

E/ELFIO/build_tarballs.jl
```diff
@@ -0,0 +1,26 @@
...
+"""
+
+platforms = [AnyPlatform()]
+
+products = [
+    FileProduct("include/elfio/elfio.hpp", :elfio_hpp)
+]
+
+
+dependencies = []
```
giordano, line 24:
```suggestion
dependencies = Dependency[]
```

PR #9272, opened 2024-08-16
Add rapidjson v1.1.0

Mellich:


R/rapidjson/build_tarballs.jl
```diff
@@ -0,0 +1,29 @@
...
+name = "rapidjson"
+version = v"1.1.0"
+sources = [
+    GitSource("https://github.com/Tencent/rapidjson.git", "f54b0e47a08782a6131cc3d60f94d038fa6e0a51")
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/rapidjson
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release \
+    -DRAPIDJSON_BUILD_DOC=No -DRAPIDJSON_BUILD_EXAMPLES=No -D RAPIDJSON_BUILD_TESTS=No -DRAPIDJSON_BUILD_CXX17=Yes -DRAPIDJSON_BUILD_CXX11=No
```
imciner2, line 14:
```suggestion
cmake -S . -B build \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DRAPIDJSON_BUILD_DOC=No \
    -DRAPIDJSON_BUILD_EXAMPLES=No \
    -DRAPIDJSON_BUILD_TESTS=No \
    -DRAPIDJSON_BUILD_CXX17=Yes \
    -DRAPIDJSON_BUILD_CXX11=No 
```
Style nit, this makes it easier to track changes and read.

imciner2 (commented):
This release is from 2016 and there have been 500+ commits to the repository since then, several issues asking for a new release, and no release ever made.

Also, it turns out that a recent CVE reported against rapidjson (https://nvd.nist.gov/vuln/detail/CVE-2024-38517) was actually fixed in 2018 through PR https://github.com/Tencent/rapidjson/pull/1261, but that hasn't been released since there haven't been any releases since then.

So, what exactly do you need rapidjson for, and would anything break if we just package the most recent commits from their master branch?

Mellich:
> So, what exactly do you need rapidjson for, and would anything break if we just package the most recent commits from their master branch?

I need it as a dependency for another C++ project.
This is a good idea, but what about the version numbering in that case? BinaryBuilder seems to complain about pre-releases in the version number.

Mellich:
> We usually make it up. Maybe something like v1.1.1?

Okay, I changed the version number to 1.1.1. With that, from my side the PR is ready to be merged.

PR #9274, opened 2024-08-16
[open62541] v1.4.4

thomvet:
Updates open62541 to version 1.4.4.


thomvet:
openssh change is unintentional; mixed up branches.

PR #9283, opened 2024-08-20
[Wasmtime] Update to v24.0.2

ararslan:


ararslan:
> error: package `wasmtime-cli v24.0.0 (/workspace/srcdir/wasmtime)` cannot be built because it requires rustc 1.78.0 or newer, while the currently active rustc version is 1.77.1

Huh. Alright then.

giordano:
We can update Rust, but maybe not to 1.80, which seems to break a few packages.

ararslan:
I wonder if it would be worth tracking multiple Rust versions like we do with GCC and LLVM then adding a `preferred_rust_version` (or whatever argument) to `build_tarballs`?

giordano:
Yes, we keep multiple versions of the Rust and Go toolchains, we only need to pass the appropriate argument for the version numbers around. We're only missing the pumbling basically

W/Wasmtime/build_tarballs.jl
```diff
@@ -1,10 +1,10 @@
 using BinaryBuilder
```
imciner2, line 2:
```suggestion
# Build counter: 1

```

imciner2:
We have Rust 1.83 now, so let's see if this works. And I see upstream release 28.0.0 already... so if this works we can also try bumping to that new version - unless there is a specific need for 24.0.0 @ararslan?

imciner2:
Turns out there is a 24.0.2 patch release that mentions some security advisories, so let's bump to that instead of just 24.0.0.

ararslan:
> And I see upstream release 28.0.0 already... so if this works we can also try bumping to that new version - unless there is a specific need for 24.0.0

Apologies for the delayed response. No specific need for v24 that I know of. I would have been fine to do v28, but the patch version is good too. Thanks for getting this over the finish line!

PR #9289, opened 2024-08-21
[XDiag] bump 0.2.2

awietek:
Just a new patch release, see https://github.com/awietek/xdiag/releases

### DIFF
--- a/X/XDiag/build_tarballs.jl
+++ b/X/XDiag/build_tarballs.jl
@@ -9,13 +9,13 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 
 name = "XDiag"
-version = v"0.2.1"
+version = v"0.2.2"
 
 include("../../L/libjulia/common.jl")
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/awietek/xdiag.git", "f2c77f8575f06a6d2614c22440676afd76f53fbc")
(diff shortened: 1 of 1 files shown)


giordano:
The history of this PR is again messed up.  The problem is that the history of your [`master`](https://github.com/awietek/Yggdrasil/tree/master) branch diverged from ours, you should hard-reset it.  If you aren't familiar with git you should do
```sh
git checkout master
# if you don't have the JuliaPackaging remote already run the following line
git remote add upstream https://github.com/JuliaPackaging/Yggdrasil.git
# In all the following commands replace `upstream` with the name of the JuliaPackaging remote, if different from `upstream`
git fetch --prune upstream
git reset --hard upstream/master
```
At this point your `master` branch will be in-sync with ours.  Now you can update `xdiag` to be on top of current `master`.  You have two options:

1. this one I'm not actually sure would work perfectly, try it out and check the history
   ```sh
   git checkout xdiag
   git rebase master xdiag
   ```
   Check with `git log` that in the `xdiag` branch you have only one commit on top of `upstream/master`
2. if previous option didn't work, then you should be able to do
   ```sh
   git checkout xdiag
   git reset --hard master
   git cherry-pick 5da71218cfeff946b37cdf7d6f0ca5f099ba73cd
   ```

In any case, after having ensure you have ***only a single commit*** on top of our `master`, you can push to your branch again:
```sh
git push --force-with-lease
```
This will automatically update this PR.

PR #9298, opened 2024-08-22
Reactant: try hermetic cuda build

wsmoses:


R/Reactant/build_tarballs.jl
```diff
@@ -7,15 +7,18 @@ include(joinpath(YGGDRASIL_DIR, "platforms", "cuda.jl"))
 
 name = "Reactant"
 repo = "https://github.com/EnzymeAD/Reactant.jl.git"
-version = v"0.0.15"
+version = v"0.0.16"
```
mofeing, line 10: I'm releasing a new build in #9321 with the new API functions so this should be v0.0.17

```suggestion
version = v"0.0.17"
```

wsmoses: er, because of the bazel requirement update [which this PR is trying to work around, this is going to be needed first].

R/Reactant/build_tarballs.jl
```diff
@@ -27,32 +30,58 @@ if [[ "${bb_full_target}" == x86_64-apple-darwin* ]]; then
     popd
 fi
 
-apk add bazel --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing/
 apk add py3-numpy py3-numpy-dev
```
mofeing, line 33: Current error seems to be that `zlib` is not installed.

```suggestion
apk add py3-numpy py3-numpy-dev zlib
```

giordano:
Was it intentional to delete the file?

wsmoses:
It was not xD

PR #9299, opened 2024-08-22
[Pango] update to 1.54.0

jwahlstrand:
This will probably be a fix for https://github.com/JuliaGraphics/Cairo.jl/issues/363 , based on discussions in https://gitlab.gnome.org/GNOME/pango/-/merge_requests/720

When building with the latest FriBidi_jll, the following error happens on MacOS:

```
../pango/pango-bidi-type.c:83:10: error: use of undeclared identifier '_FRIBIDI_TYPE_SENTINEL'
[10:08:01]     case _FRIBIDI_TYPE_SENTINEL:
[10:08:01]          ^
[10:08:01] 1 error generated.
```

I think the fix is easy, but using the previous version of FriBidi_jll was easier for me than generating a patch. I'll do that if preferred, and either way I'll create an issue upstream.

### DIFF
--- a/P/Pango/build_tarballs.jl
+++ b/P/Pango/build_tarballs.jl
@@ -71,7 +71,7 @@ dependencies = [
     Dependency("Cairo_jll"; compat="1.18.0"),
     Dependency("Fontconfig_jll"),
     Dependency("FreeType2_jll"; compat="2.13.1"),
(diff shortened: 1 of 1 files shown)


giordano:
Setting a compat is a good idea anyway, but please do open an issue/merge request upstream!

PR #9302, opened 2024-08-23
[Pango] rebuild for HarfBuzz update

jwahlstrand:


giordano (changes requested):
You must change the version number to be able to change compat bound (it's a requirement from Pkg)

jwahlstrand:
Ah, unfortunately this is the latest. I think I've seen sometimes people use a fake version in this situation?

giordano:
Yeah

giordano (approved):

PR #9309, opened 2024-08-24
[x264] Upgrade to a newer commit and change versioning scheme

giordano:


giordano:
This also breaks the ABI (soversion 164 vs 163), sigh.

giordano:
For the benefit of future readers (including myself), the soversion [comes](https://code.videolan.org/videolan/x264/-/blob/4613ac3c15fd75cebc4b9f65b7fb95e70a3acce1/version.sh#L28) from the [`X264_BUILD`](https://code.videolan.org/videolan/x264/-/blob/4613ac3c15fd75cebc4b9f65b7fb95e70a3acce1/x264.h#L48) macro

PR #9330, opened 2024-08-29
Reactant: hermetic build, part 2

wsmoses:


R/Reactant/build_tarballs.jl
```diff
@@ -0,0 +1,422 @@
...
+		"libcudnn",
+		"libnvJitLink",
+		"libcublas",
+		"libcudnn_ops",
+		"libnvrtc-builtins",
+		"libcudnn_graph",
+		"libcusolver",
+		"libcuda",
+		"libcudnn_engines_runtime_compiled",
+		"libcusparse",
```
giordano, line 395: Are these libraries from CUDA, or are part of Reactant?

wsmoses:
Reactant builds them (or rather it is part of Bazel hermetic cuda)

PR #9331, opened 2024-08-30
Update libjulia 1.12.0-DEV to latest master

fingolfin:
There were several breaking kernel API changes

Will need this to update GAP related JLLs.

Perhaps we should also update 1.11 to rc3? Dunno if there were relevant changes.

CC @benlorenz


fingolfin:
Thank you for the analysis @benlorenz, very helpful as always!

I've updated to 1.11.0-rc3. I don't have time to look into the rest right now, maybe later or monday (of course everyone is welcome to push additional fixes in the meantime).

fingolfin:
Seems to build now with my latest patch. @benlorenz could you please double check what I did?

L/libjulia/common.jl
```diff
@@ -266,6 +262,24 @@ function build_julia(ARGS, version::VersionNumber; jllversion=version)
         LIBUNWIND:=-losxunwind
         JCPPFLAGS+=-DLIBOSXUNWIND
     EOM
+    else
+        # LLVMLIBUNWIND is currently only defined for USE_SYSTEM_UNWIND=0
+        # but we always need that for apple and julia > 1.6
+        cat << EOM >>Make.user
+        JCPPFLAGS+=-DLLVMLIBUNWIND
+    EOM
```
fingolfin, line 270: @benlorenz this now contains your PR https://github.com/JuliaLang/julia/pull/55639 but I left the workaround for that issue in as you wrote earlier:
> It might make sense to keep the extra flag here for older julia versions even if the PR is merged

We may wish to at least adjust this comment though, e.g.
```suggestion
    else
        # Ensure LLVMLIBUNWIND is always defined for apple and julia > 1.6
        # to work around issues in old versions of the Julia build system
        # which did not always add this flag when needed
        cat << EOM >>Make.user
        JCPPFLAGS+=-DLLVMLIBUNWIND
    EOM
```

fingolfin: I've restricted it to 1.6-1.11 on Apple and tweaked the comment

fingolfin:
Indeed (I could have sworn that I already updated `jllversion` but clearly I didn't).

Hmm, do we have any rules as to when we update the patch level vs. the minor version? I've now changed it to `1.10.11` but could just as well use `1.11.0` ...

PR #9336, opened 2024-08-30
MAGEMin v1.5.2

NicolasRiel:
- Corrected issue when adding anti-ordered pseudocompound that was leading to site fraction constraint being not respected


NicolasRiel:
@giordano Thanks a lot!

PR #9365, opened 2024-09-03
[proxTV] Recompile with -O3 and include headers in the JLL

amontoison:
@nathanemac
I also added some
```c
#ifdef __cplusplus
extern "C" {
#endif
...
#ifdef __cplusplus
}
#endif
```
such that we can easlily generate wrappers with Clang.jl.

### DIFF
--- a/P/proxTV/build_tarballs.jl
+++ b/P/proxTV/build_tarballs.jl
@@ -2,11 +2,11 @@ using BinaryBuilder, Pkg
 using Base.BinaryPlatforms
 
 name = "proxTV"
-version = v"3.3.0"
+version = v"3.3.1"
 
 # Collection of sources required to build proxTV.
 sources = [
-    GitSource("https://github.com/albarji/proxTV", "0b8973cd7e1a2348c842393f689949eeebc9f654"),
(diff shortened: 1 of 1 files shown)


P/proxTV/build_tarballs.jl
```diff
@@ -2,26 +2,27 @@ using BinaryBuilder, Pkg
 using Base.BinaryPlatforms
 
 name = "proxTV"
-version = v"3.3.0"
+version = v"3.3.1"
 
 # Collection of sources required to build proxTV.
 sources = [
-    GitSource("https://github.com/albarji/proxTV", "0b8973cd7e1a2348c842393f689949eeebc9f654"),
+    GitSource("https://github.com/amontoison/proxTV", "6f7ad7e1048a6d97aa1e92ec35937ae4f155b1a0"),
```
dpo, line 9: @amontoison This is no longer the official repository. Did you submit your changes to the original proxTV?

amontoison: I forgot to submit the changes.
I opened this PR: https://github.com/albarji/proxTV/pull/62

Otherwise, we can't use proxTV in Julia.

PR #9367, opened 2024-09-04
nauty: bump version to 2.8.9, add TLSlibs

mxhbl:
This PR bumps nauty to the latest version and adds build targets for shared libraries with thread local storage enabled. This makes it possible to call into nauty from different threads simultaneously.

Changes are summarized here:
- I found that the autoconf tools that come with nauty pretty much just work with minimal patching, so I removed the extensive autotooling patch. I am not sure if this is because the 2.8.9 version of the nauty configure script is actually much better, or because my changes break things on systems I didn't test. It would be great if somebody can confirm that this change does not ruin everything.
- I was pretty conservative with respect to hardware assumptions. I disabled the `popcnt` instruction and added `--enable-generic`. Maybe it would make sense to add additional targets with e.g. avx enabled.
- The source file for the `blisstog` program is missing since at least nauty 2.8.8, so I removed this build target. (This seems like a bug.)
- Some shared libraries have slightly different names in this version of nauty, e.g. `libnautyL0` is now `libnautyL`.
- I added build targets for shared libraries with TLS enabled. All TLS libraries have a T in their name, so that e.g. the TLS version of `libnautyS1` is `libnautyTS1`.
- There is a new graph utility `uniqg` which is now also a target.


fingolfin:
The old version of nauty did not use autoconf (the patch added it). They had no good way to support e.g. cross compilation or anything else. That the new system is much better is not a surprise.


The two other JLLs using `nauty_jll` both have a compat bound that should ensure this update doesn't cause their existing release to regress. And I'd be happy to update `normaliz_jll` to use the newer `nauty_jll`.

N/nauty/build_tarballs.jl
```diff
@@ -3,44 +3,41 @@
...
 export CPPFLAGS="${CPPFLAGS} -I${prefix}/include"
 export LDFLAGS="${LDFLAGS} -L${prefix}/lib"
 
-./configure \
-    --prefix=$prefix \
-    --build=${MACHTYPE} \
-    --host=$target \
-    --enable-shared \
-    --disable-static
+atomic_patch -p1 ../patches/configure.patch
```
fingolfin, line 25: I would strongly recommend to annotate all patches with a comment explaining them, so that people who update this recipe have less work in the future ("hmm this patch doesn't apply anymore. What was it about anyway? Do we still need? What would be an appropriate replacement?" -- all of that is hard to answer otherwise).

Two options for that:
1. add a comment before this line
2. change the .patch to use the format produced by `git format-patch` (i.e. put it into a commit then use `git format-patch HEAD^` to produce a patch file with the commit message in it).

N/nauty/build_tarballs.jl
```diff
@@ -3,44 +3,41 @@
...
+	    --host=${target} \
+	    --enable-generic \
+	    --enable-shared \
+	    --disable-popcnt \
+	    --libdir=${libdir} \
+	    --bindir=${bindir}
 
 make -j${nproc}
-make install
+make install TLSinstall checks6 sumlines
```
fingolfin, line 37: What is `checks6`, `sumlines` and why do they need to be listed here? Perhaps add another comment explaining that?

N/nauty/build_tarballs.jl
mxhbl: I added a comment explaining the patch.

mxhbl:
I added some comments, I hope everything is clear enough. (Two commits because I did not realize that the shell cant handle comments in a multiline command, sorry about that).

fingolfin:
@thofma @benlorenz any concerns about merging this? It seems reasonable to me

N/nauty/build_tarballs.jl
fingolfin: Thank you, great comment!

thofma:
No concerns, but I also don't have a good overview about nauty and its dependencies. If you and @benlorenz are happy, then I am happy.

N/nauty/build_tarballs.jl
giordano: Isn't it better to patch `configure.ac`?  `configure` can change a lot from version to version.

mxhbl: You're right. The patch is only needed because a runtime check in `configure.ac` is missing a default argument, causing the check to fail during cross-compilation. I will add a patch for `configure.ac` and run `autoconf` in the build script.

mxhbl:
Great! I took @giordano's suggestion and simplified the patch a little bit. It now patches `configure.ac` directly.

PR #9375, opened 2024-09-05
add arm64 to quarto

bdklahn:
That same macos ArchiveSource does, in fact, seem to be able to support both amd64 and arm64. It looks like the shell wrapper chooses the appropriate deno JS runtime executable dynamically (I.e. the artifacts in that upstream tarball do seem to also run on my M1 MacBook Pro).

I'm not positive if it makes any difference to have the "unpack_target" be " . . . darwin" or ". . . darwin14". I assumed the former, reasoning that amd Macs are becloming "legacy", and maybe it required pinning to an older Darwin version to work.
We'll see . . .


bdklahn:
Looks like adding the "20" fixed the build issue.
I wonder if it ends up being cleaner/easier to just omit specifying the "unpack_target", taking the default(?).
The API docs don't seem to address the purpost/advantage of specifying that. E.g., I'm not sure if the "darwin14", for the amd64 target makes sure it is signaled to be backwards compatible to Dawin14, or . . .
By setting the arm64 one, will that prohibit the artifact from being download on systems with anything but Darwin 20 or newer?

PR #9398, opened 2024-09-09
Wizard recipe: libbensolve-v2.1.0

kofgokhan:
This pull request contains a new build recipe I built using the BinaryBuilder.jl wizard:

* Package name: libbensolve
* Version: v2.1.0

@staticfloat please review and merge.

PR #9401, opened 2024-09-09
MAGEMin v1.5.3

NicolasRiel:
- Added an extended version of the ultramafic database (ume), accounting for hb, pl and aug from metabasite database
- Added atom per formula unit (apfu) output for the Julia wrapper
- Corrected inaccurate reaction lines in case of liquid solvus 
- Added support for reduced compositional system for ultramafic database (O and S can be set to 0.0)


giordano:
https://github.com/JuliaPackaging/Yggdrasil/pull/9398#discussion_r1750666199

PR #9404, opened 2024-09-10
[Zlib] Trigger a new build

ararslan:
To get FreeBSD AArch64 support

### DIFF
--- a/Z/Zlib/build_tarballs.jl
+++ b/Z/Zlib/build_tarballs.jl
@@ -21,7 +21,7 @@ if [[ ${bb_full_target} == *-sanitize+memory* ]]; then
     # Install msan runtime (for clang)
     cp -rL ${libdir}/linux/* /opt/x86_64-linux-musl/lib/clang/*/lib/linux/
 fi
-# We use `-DUNIX=true` to ensure that it is always named `libz` instead of `libzlib` or something ridiculous like that.
+# We use `-DUNIX=true` to ensure that it is always named `libz` instead of `libzlib` or something absolutely absurd like that.
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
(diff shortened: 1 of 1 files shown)


giordano:
https://buildkite.com/julialang/yggdrasil/builds/13148#0191d976-0012-41b3-b57e-608bdf6baeac/666-963
```
┌ Warning: libz.so.1.3.1 has an ELF header OS/ABI value that is not set to FreeBSD (9), this may be an issue at link time
```
Sigh.

ararslan:
> Warning: libz.so.1.3.1 has an ELF header OS/ABI value that is not set to FreeBSD (9), this may be an issue at link time

😕

ararslan:
https://buildkite.com/julialang/yggdrasil/builds/13148#0191d976-0012-41b3-b57e-608bdf6baeac/666-674

This is weird, the other platforms don't have this in their logs 🤔

giordano:
You mean
```
Downloading artifact: PlatformSupport-aarch64-unknown-freebsd13.2.v2024.8.10.x86_64-linux-musl.unpacked
```
? Their platform support artifacts are probably cached: they're always needed and they're always the same (unlike for example gcc: it's also always needed, but we often use different versions, so caching isn't very effective for that)

ararslan:
> unlike for example gcc: it's also always needed, but we often use different versions, so caching isn't very effective for that

The other platforms aren't downloading GCC either, but this is:

```
Downloading artifact: GCCBootstrap-aarch64-unknown-freebsd13.2.v7.1.0.x86_64-linux-musl.unpacked
```
https://buildkite.com/julialang/yggdrasil/builds/13148#0191d976-0012-41b3-b57e-608bdf6baeac/666-676

Why would this need GCC? FreeBSD uses Clang as its compiler.

giordano:
Ah, when we ask for the "C" compilers we always have both gcc and clang, it's only the default (what `cc` points to) which is platform-specific (clang on bsd systems, gcc elsewhere)

PR #9405, opened 2024-09-10
[OpenBLAS@0.3.28] Trigger a new build

ararslan:
To get FreeBSD AArch64 support. This will be interesting since it uses CSL but the CSL libraries have their OS/ABI set incorrectly on FreeBSD AArch64. Will it break? Let's find out!


giordano:
There's also openblas32

giordano:
I think it isn't entirely clear what toolchain we should use for that platform (I think someone suggested clang with mingw, not sure gcc is supported at all), while it was much more obvious what to do for aarch64-freebsd

ararslan:
> There's also openblas32

Do you mean I should trigger a new build for that as part of this PR?

PR #9407, opened 2024-09-10
[DuckDB] Bump to v1.1.0

Mytherin:


Mytherin:
Not so sure what to do about this error message:

```
In file included from /workspace/srcdir/duckdb/third_party/fmt/include/fmt/format.h:37:0,
                 from /workspace/srcdir/duckdb/third_party/fmt/include/fmt/format-inl.h:11,
                 from /workspace/srcdir/duckdb/third_party/fmt/format.cc:8:
/workspace/srcdir/duckdb/third_party/fmt/include/fmt/core.h: In member function ‘void duckdb_fmt::v6::internal::arg_map<Context>::push_back(duckdb_fmt::v6::internal::value<Context>) [with Context = duckdb_fmt::v6::basic_format_context<std::back_insert_iterator<duckdb_fmt::v6::internal::buffer<char> >, char>]’:
/workspace/srcdir/duckdb/third_party/fmt/include/fmt/core.h:1054:8: note: the ABI of passing aggregates with 16-byte alignment has changed in GCC 5
   void push_back(value<Context> val) {
        ^~~~~~~~~
```

giordano:
Maybe we can skip powerpc for this package for the time being, I can do the edit later

PR #9412, opened 2024-09-11
Update LLVMOpenMP to v18.1.8

jkrimmer:


L/LLVMOpenMP/build_tarballs.jl
```diff
@@ -3,18 +3,18 @@
 using BinaryBuilder, Pkg
 
 name = "LLVMOpenMP"
-version = v"17.0.6"
+version = v"18.1.8"
```
giordano, line 6: Can you please use 18.1.7 to match the llvm version we have in Julia https://github.com/JuliaPackaging/Yggdrasil/blob/4f143bdef1bd8e51e4eff41f2dedaeeeb4cc785d/L/LLVM/common.jl#L22? It shouldn't matter much for compatibility, but it'd be nicer to have homogeneous versions. Also, I seem to remember there were only two commits related to libcxx between 18.1.7 and 18.1.8, so the two versions of openmp should be identical

jkrimmer: Sure!

PR #9414, opened 2024-09-11
[FLEXPART] change maxspec to 10

PaulPerezRMA:
Fix the error we get when trying to model multiple species in a single run.


barche:
To add this changed parameter, the patch needs to be regenerated with the diff command, the line you are modifying is in the context, not in the changed lines.

PR #9420, opened 2024-09-12
[LibUnwind@1.8.1] Rebuild for FreeBSD AArch64 support

ararslan:


ararslan:
Whoops, this needs https://github.com/JuliaPackaging/Yggdrasil/pull/9421; I missed that xz was a build dependency

giordano:
You need to use xz 5.2.5, that's what you just built.

ararslan:
Wait did I build the wrong version? Looks like Linux AArch64 installed xz 5.4.6 😕

ararslan:
Oh I see, the builder for xz was downgraded to 5.2.5 since it's the last version before the malicious actor got involved.

ararslan:
Looks like i686 Linux has a problem with the older xz 😩

ararslan:
Same issue on x86_64 and i686 Linux, both glibc and musl: it can't find liblzma. But the recently rebuilt xz v5.2.5 was successful on all of those platforms and the liblzma product was found. I don't understand... 🤔 😕

giordano:
https://docs.binarybuilder.org/stable/troubleshooting/#General-comments

ararslan:
> I don't understand...

> you will need to understand

lol, welp. I was hoping somebody might have some wisdom like "oh yeah I've seen that before, you need to change X" but it's into the ~~abyss~~ shell I go.

giordano:
"libx cannot be found" doesn't mean anything, you want to know what's the check they're doing and why it's failing.  There's nothing useful written to screen.

ararslan:
Okay, I think the issue is that xz was built with gcc 7 and this builder uses gcc 5. AFAICT, there's no reason to still need gcc 7 given that we reverted to an old version of xz, so #9426 should fix this. This is based on this bit of `config.log` plus some googling:

```
configure:18480: cc -o conftest -DPI -fPIC  -D_GNU_SOURCE -DNDEBUG  conftest.c -L/workspace/destdir/lib -l:liblzma.a  >&5
/opt/x86_64-linux-gnu/bin/../lib/gcc/x86_64-linux-gnu/5.2.0/../../../../x86_64-linux-gnu/bin/ld: /workspace/destdir/lib/liblzma.a(liblzma_la-lz_encoder.o): unrecognized relocation (0x2a) in section `.text'
```

giordano:
Ah, good find! In that case how about using gcc 7 here? For C code the version of gcc is less problematic (but still, it's always better to avoid versions more recent than ~10 if possible), the main issue is sometimes the bundled version of binutils, which is the issue you're having here.

giordano:
BTW, that error was also mentioned at https://docs.binarybuilder.org/stable/troubleshooting/#Building-with-an-old-GCC-version-a-library-that-has-dependencies-built-with-newer-GCC-versions 😁

ararslan:
Would updating the GCC version here have any downstream consequences for building Julia, since this is a direct dependency? Like would we lose support for older glibcs or something? (I don't really know how these things work tbh so maybe that doesn't make sense.)

ararslan:
> BTW, that error was also mentioned at [...]

image

giordano:
> Would updating the GCC version here have any downstream consequences for building Julia, since this is a direct dependency? Like would we lose support for older glibcs or something? (I don't really know how these things work tbh so maybe that doesn't make sense.)

Good question. Glibc not really since we use the same glibc for all the builds (it [varies by architecture](https://github.com/JuliaPackaging/Yggdrasil/blob/1ec20823d1e27a380d74623bc345879f36e8c2ee/RootFS.md#glibc-versions), but within an architecture it's homogenous), but we have other Julia deps already using newer gcc versions (only because of specific needs), like openblas:
https://github.com/JuliaPackaging/Yggdrasil/blob/1ec20823d1e27a380d74623bc345879f36e8c2ee/O/OpenBLAS/OpenBLAS%400.3.28/build_tarballs.jl#L19
Llvm: https://github.com/JuliaPackaging/Yggdrasil/blob/1ec20823d1e27a380d74623bc345879f36e8c2ee/L/LLVM/LLVM_full%4018/build_tarballs.jl#L6
As I mentioned above, the main issue for C packages one could have is being forced to use a newer binutils during the build, but for the rest we aren't aware of issues at runtime. For C++ code, like llvm, there's indeed a compatibility issue in the version of libstdc++ required, but we ship that one in CompilerSuppotLibraries

ararslan:
I appreciate the explanation! I bumped to GCC 7 here and things are working now. Could still be worth doing #9426 since XZ is likely near the bottom of the dependency tree for a number of things that would all need to bump their preferred GCC version as well. (Although I suppose most probably have anyway given that XZ has used GCC 7 for like 6 months at this point.)

giordano:
Ok, yes

PR #9422, opened 2024-09-13
Wizard recipe: libgmime-v3.2.15

AlexKlo:

* Package name: libgmime
* Version: v3.2.15


AlexKlo:
I have an interesting issue with this script. It fails with an error in its current form, but if I change the sting here 
`products = [
    LibraryProduct("libgmime", :libgmime)
]`
from `"libgmime"` to `'libgmime-3.0"`, it works. 
Also, I was not able to build other platforms using this script.

PR #9433, opened 2024-09-15
Add Shared Object to atomsk Build

jmeziere:
Currently, the atomsk build only produces executables. To write a wrapper around atomsk, I would like to compile the shared library.


A/Atomsk/build_tarballs.jl
```diff
@@ -7,25 +7,35 @@ version = v"0.11.2"
...
-cd $WORKSPACE/srcdir/atomsk/src/
-mkdir -p ${bindir}
+cd $WORKSPACE/srcdir/atomsk/src
+sed -i '172s/$/ -I $(OBJ) $(LAPACK)/' Makefile
+sed -i '172s/atomsk.so/\$(libdir)\/libatomsk.$(dlext)/' Makefile
+if [[ "${exext}" == "" ]]; then
+    sed -i '161s/\$(BIN)/\$(bindir)\/atomsk/' Makefile
+else
+    sed -i '161s/\$(BIN)/\$(bindir)\/atomsk.$(exeext)/' Makefile
+fi
```
giordano, line 22: Can you please add a proper patch?  `sed` patching-on-the-fly is fragile and error-prone when changing version.

A/Atomsk/build_tarballs.jl
```diff
@@ -7,25 +7,35 @@ version = v"0.11.2"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/pierrehirel/atomsk.git", "3333858281e0ebd6279825b83ab871cf4d050a8d")
+    GitSource("https://github.com/pierrehirel/atomsk.git", "84f60a20c5b814fec03bffe1bccc3daaed0fc65d"),
```
giordano, line 10: Changed the source without changing the version number

A/Atomsk/build_tarballs.jl
jmeziere: I was having difficulty with that, but let me try again, and I'll see if I can get it to work. It was telling me that the patch failed.

A/Atomsk/build_tarballs.jl
jmeziere: This is fixed in the latest commit.

A/Atomsk/build_tarballs.jl
jmeziere: The last commit has the patch, but running `git diff -p` produces a broken patch. Do you know why this is?

giordano:
What's the line ending of the file?

jmeziere:
I actually fixed the patch. It had to do with tabs vs spaces. But I am concerned about the macOS builds. It seems like it's using a static library that is compiled with iOS instead?

jmeziere:
The exact warning is 

```
[15:29:15] f951: Warning: Nonexistent include directory ‘../OBJ/’ [-Wmissing-include-dirs]
[15:29:15] f951: Warning: Nonexistent include directory ‘../OBJ/’ [-Wmissing-include-dirs]
[15:29:16] ld: warning: building for macOS, but linking in object file (/opt/aarch64-apple-darwin20/lib/gcc/aarch64-apple-darwin20/12.0.1/libgcc.a(cas_8_1.o)) built for iOS
[15:29:16] ld: warning: building for macOS, but linking in object file (/opt/aarch64-apple-darwin20/lib/gcc/aarch64-apple-darwin20/12.0.1/libgcc.a(ldadd_4_1.o)) built for iOS
[15:29:16] ld: warning: building for macOS, but linking in object file (/opt/aarch64-apple-darwin20/lib/gcc/aarch64-apple-darwin20/12.0.1/libgcc.a(_powidf2.o)) built for iOS
[15:29:16] ld: warning: building for macOS, but linking in object file (/opt/aarch64-apple-darwin20/lib/gcc/aarch64-apple-darwin20/12.0.1/libgcc.a(lse-init.o)) built for iOS
```

jmeziere:
@giordano All the builds are passing now. However, the previously mentioned warnings still persist for the  aarch64-apple-darwin build. Is this a problem?

jmeziere:
I just removed the aarch64-apple-darwin build, so everything should be good to merge, I believe.

giordano:
> Is this a problem?

No

jmeziere:
Ok, I reverted back to add back in the Mac build. So everything should be good now.

giordano:
```
┌ Warning: /cache/build/yggy-amdci7-4/julialang/yggdrasil/A/Atomsk/build/x86_64-apple-darwin/mM4nN5aa/x86_64-apple-darwin14-libgfortran5-cxx11/destdir/lib/libatomsk.dylib links to libgfortran!  This causes incompatibilities across major versions of GCC.  To remedy this, you must build a tarball for each major version of GCC.  To do this, immediately after your `platforms` definition in your `build_tarballs.jl` file, add the line:
│ 
│     platforms = expand_gfortran_versions(platforms)
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/HTdhp/src/auditor/compiler_abi.jl:60
```

A/Atomsk/build_tarballs.jl
```diff
@@ -3,29 +3,43 @@
...
+fi
+cd src
+
 # The makefile doesn't handle parallel builds
-make atomsk BIN="atomsk${exeext}"
-make install INSTPATH=${prefix} BIN="atomsk${exeext}"
+if [[ ! -d "$bindir" ]]; then
+    mkdir ${bindir}
+fi
+make atomsk
```
giordano, line 31: Let's make this build less slow
```suggestion
make -j${nproc} atomsk
```

A/Atomsk/build_tarballs.jl
```diff
@@ -3,29 +3,43 @@
...
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(; experimental=true)
-platforms = expand_gfortran_versions(platforms)
+platforms = supported_platforms()
+platforms = expand_cxxstring_abis(platforms)
```
giordano, line 37: Why?

A/Atomsk/bundled/patches/atomsk_64.patch
```diff
@@ -0,0 +1,101 @@
...
+@@ -169,7 +171,7 @@ atomsk: message modes compute general
+ # Compile the program as a library
+ # Note that the preprocessor must be called (-cpp)
+ lib: message compute modes
+-	$(FORTRAN) $(FFLAGS) -cpp -shared -o atomsk.so $(OBJ)$(SEP)*.o
++	$(FORTRAN) $(EXEFFLAGS) -cpp -shared -o $(libdir)/libatomsk.$(dlext) $(OBJ)$(SEP)*.o $(LAPACK)
+ 	@echo ""
+ 	@echo "    \o/ Atomsk was compiled as a shared library file, 'atomsk.so'."
+ 	@echo ""
+diff --git a/src/compute/compute_G.f90 b/src/compute/compute_G.f90
```
giordano, line 38: It'd be easier to do this with some macros, but the build system doesn't seem to be sensible enough (they ask you to _edit_ the makefile....) to allow that.

A/Atomsk/build_tarballs.jl
jmeziere: I just saw it in the LAMMPS build_tarballs.jl. Is one preferred over the other?

A/Atomsk/bundled/patches/atomsk_64.patch
jmeziere: Yea, the package is just maintained by one person, and I think he's good with the way it is now.

giordano: It isn't about preference, it's about one thing is needed or not: https://docs.binarybuilder.org/stable/building/#Expanding-C-string-ABIs-or-libgfortran-versions.  Here you have to expand the libgfortran version because we're linking to libgfortran, this code seems to be written entirely in Fortran so expanding the C++ string ABI makes no sense.

A/Atomsk/build_tarballs.jl
giordano: It isn't about preference, it's about one thing is needed or not: https://docs.binarybuilder.org/stable/building/#Expanding-C-string-ABIs-or-libgfortran-versions.  Here you have to expand the libgfortran version because we're linking to libgfortran, this code seems to be written entirely in Fortran so expanding the C++ string ABI makes no sense.

A/Atomsk/build_tarballs.jl
```diff
@@ -3,29 +3,43 @@
...
-mkdir -p ${bindir}
+cd $WORKSPACE/srcdir/atomsk
+if [[ "$nbits" == 32 || "$target" == *apple* ]]; then
+    atomic_patch -p1 ../patches/atomsk_32.patch
+else
+    atomic_patch -p1 ../patches/atomsk_64.patch
+fi
+if [[ "$target" == *mingw* ]]; then
+    ln -s $prefix/bin/libatomsk.dll $prefix/lib/libatomsk.dll
+fi
```
giordano, line 24: Why?

A/Atomsk/build_tarballs.jl
giordano: Ah, I see the build system is also broken and can't deal with parallel build.  Nice.

A/Atomsk/build_tarballs.jl
jmeziere: Ah, I see. That makes sense.

A/Atomsk/build_tarballs.jl
jmeziere: Haha, I'll change it back.

A/Atomsk/build_tarballs.jl
jmeziere: I link the executable off of the lib that was just built and saved to `libdir`. But `libdir` is not in the `LD_LIBRARY_PATH` for `mingw` builds, so I link it there.

giordano: > the lib that was just built and saved to `libdir`

That's where it's supposed to be, yes.

> But `libdir` is not in the `LD_LIBRARY_PATH`

What does that mean?  `LD_LIBRARY_PATH` has no role here (and is not a thing on Windows at all)

jmeziere: I guess I mean that when I try to link it with `-latomsk`, it doesn't find the library unless I move it over to the `$prefix/lib` directory as well.

giordano: There are other options besides creating useless links.

jmeziere: Is there any option that you would prefer? Is there an environment variable that BinaryBuilder uses that I can set?

jmeziere: I set LDFLAGS to include that directory in the search for libraries.

jmeziere:
@giordano I believe I've fixed everything now.

vchuravy:
Also probably good to propose your patch upstream as well.

jmeziere:
I noticed that for 64-bit vs 32-bit, the `LAPACK` function names are different. For example, the function `DGETRF` in 32-bit corresponds to `DGETRF_64` in 64-bit. Is there something wrong in the way I am loading the library? A good portion of the patch converts these function calls over, but it would be nice not to have to do this.

giordano:
https://docs.binarybuilder.org/stable/build_tips/#link-blas

A/Atomsk/build_tarballs.jl
```diff
@@ -3,29 +3,45 @@
 using BinaryBuilder, Pkg
 
 name = "Atomsk"
-version = v"0.11.2"
+version = v"0.12.0"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/pierrehirel/atomsk.git", "3333858281e0ebd6279825b83ab871cf4d050a8d")
+    GitSource("https://github.com/byu-cxi/atomsk.git", "adeb5d5c059a16bf649ab6657cc550b5868fac0b"),
```
giordano, line 10: Why using a fork?

A/Atomsk/build_tarballs.jl
```diff
@@ -3,29 +3,45 @@
...
-cd $WORKSPACE/srcdir/atomsk/src/
-mkdir -p ${bindir}
+cd $WORKSPACE/srcdir/atomsk
+if [[ "$nbits" == 32 || "$target" == *apple* ]]; then
+    atomic_patch -p1 ../patches/atomsk_32.patch
+else
+    atomic_patch -p1 ../patches/atomsk_64.patch
+fi
+if [[ "$target" == *mingw* ]]; then
+    export LDFLAGS="-L$bindir"
```
giordano, line 23:
```suggestion
    export LDFLAGS="-L${libdir}"
```

A/Atomsk/build_tarballs.jl
```diff
@@ -3,29 +3,45 @@
...
+fi
+cd src
+
 # The makefile doesn't handle parallel builds
-make atomsk BIN="atomsk${exeext}"
-make install INSTPATH=${prefix} BIN="atomsk${exeext}"
+if [[ ! -d "$bindir" ]]; then
+    mkdir ${bindir}
+fi
```
giordano, line 30:
```suggestion
mkdir -p ${bindir}
```
but I find it bizarre the build system isn't able to deal with this

A/Atomsk/build_tarballs.jl
```diff
@@ -3,29 +3,45 @@
...
+make atomsk
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = supported_platforms(; experimental=true)
```
giordano, line 36:
```suggestion
platforms = supported_platforms()
```

A/Atomsk/build_tarballs.jl
```diff
@@ -3,29 +3,45 @@
...
 platforms = expand_gfortran_versions(platforms)
+platforms = filter(p -> !(Sys.isfreebsd(p) || libc(p) == "musl"), platforms)
+# Atomsk is not supported for libgfortran versions less than 5
+platforms = filter(p -> p["libgfortran_version"] == "5.0.0", platforms)
```
giordano, line 40: String comparison for version numbers is fragile, at best
```suggestion
platforms = filter(p -> libgfortran_version(p) >= v"5", platforms)
```

A/Atomsk/build_tarballs.jl
jmeziere: The maintainer is currently busy and hasn't had time to look over my pull request. So I put mine temporarily until he has time to check things out.

A/Atomsk/build_tarballs.jl
jmeziere: Made this change.

A/Atomsk/build_tarballs.jl
jmeziere: This is fixed now.

A/Atomsk/build_tarballs.jl
jmeziere: This is fixed now

A/Atomsk/build_tarballs.jl
jmeziere: I don't know why, but it seems that the `bindir` directory isn't there for all but the windows os.

A/Atomsk/build_tarballs.jl
jmeziere: Everything is pushed up now.

jmeziere:
@giordano I believe everything should be good to go now.

A/Atomsk/build_tarballs.jl
```diff
@@ -3,37 +3,49 @@
...
+    export LDFLAGS="-L${libdir}"
+fi
+cd src
+
 # The makefile doesn't handle parallel builds
-make atomsk BIN="atomsk${exeext}"
-make install INSTPATH=${prefix} BIN="atomsk${exeext}"
+mkdir -p $bindir
+make shared=yes clib
+cp atomsk$exeext $bindir
```
giordano, line 27:
```suggestion
make shared=yes clib
install -Dvm 755 "atomsk${exeext}" -t "${bindir}"
```

giordano:
Not a blocker for this PR, but for the future: [avoid opening PRs from the same branch as the target one](https://github.com/JuliaPackaging/Yggdrasil/blob/2e2ebd06b660e9c5342442adbf53c5cb766cec5e/CONTRIBUTING.md#branch-naming)

jmeziere:
> Not a blocker for this PR, but for the future: [avoid opening PRs from the same branch as the target one](https://github.com/JuliaPackaging/Yggdrasil/blob/2e2ebd06b660e9c5342442adbf53c5cb766cec5e/CONTRIBUTING.md#branch-naming)

Sorry about that. I'll make sure for any future PRs.

PR #9434, opened 2024-09-15
[MPFR] Rebuild for FreeBSD AArch64 support

ararslan:


ararslan:
Ah, this is using GMP v6.2.1 but I only rebuilt v6.3.0. @giordano, would it be better to use GMP v6.3.0 here or rebuild v6.2.1? (Both is also an option, I suppose.)

giordano:
Changing compat bound of a dependency without changing the version number of the package isn't an option.

ararslan:
I've seen that error for the msan build before but I'm not sure what fixed it at the time. 🤔

ararslan:
Um, this seems like an issue, why did this not cause the build to fail?
```
[ Info: Checking shared library lib/libmpfr.so.6.2.1
ERROR: could not load library "/cache/build/yggy-amdci7-14/julialang/yggdrasil/M/MPFR/build/x86_64-linux-gnu-sanitize+memory/IS0ST4la/x86_64-linux-gnu-libgfortran3-cxx11-sanitize+memory/destdir/lib/libmpfr.so.6.2.1"
/cache/build/yggy-amdci7-14/julialang/yggdrasil/M/MPFR/build/x86_64-linux-gnu-sanitize+memory/IS0ST4la/x86_64-linux-gnu-libgfortran3-cxx11-sanitize+memory/destdir/lib/libmpfr.so.6.2.1: undefined symbol: __msan_va_arg_overflow_size_tls
┌ Warning: lib/libmpfr.so.6.2.1 cannot be dlopen()'ed
└ @ BinaryBuilder.Auditor /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/HTdhp/src/Auditor.jl:173
```

giordano:
> Um, this seems like an issue

That's expected on msan builds: you need an msan julia to be able to load that.

> why did this not cause the build to fail?

Because the target platform (msan julia) doesn't match the host platform (non-msan julia)

PR #9435, opened 2024-09-15
[SuiteSparse@7] Rebuild for FreeBSD AArch64 support

ararslan:


ararslan:
Needs https://github.com/JuliaRegistries/General/pull/115226; the libblastrampoline registration failed initially

ararslan:
Ugh, SuiteSparse sets its libblastrampoline compat to 5.8.0 build I rebuilt 5.11.0. We don't have separated recipes for different LBT versions.

ararslan:
Based on a conversation in Slack with @rayegun, it sounds like the approach taken here (setting the build version to 5.11 while keeping the compat at 5.8) should be just fine, which @staticfloat appears to have agreed with, at least based on Slack emoji reactions.

PR #9441, opened 2024-09-16
Initial build recipe for OWENSOpenFAST custom built libraries

kevmoor:
This pull request contains a new build recipe for custom OpenFAST Libraries I built manually:

Package name: OWENSOpenFAST
Version: v1.0.0

There is an OpenFAST recipe already, but these library versions are not slated to be merged into OpenFAST main for another 6ish months.  When that happens, updating the OpenFAST build recipe to include these is planned.  These libraries were built specifically to enable c-binding compatible interfacing, which simplifies/bypasses the complex fortran types currently used in OpenFAST that prohibits current usage of ccall.  Julia side package has already been written and this would replace the build.jl script here https://github.com/sandialabs/OWENSOpenFASTWrappers.jl/blob/master/deps/build.jl


fingolfin:
@kevmoor is this still something you work on (since it is marked as draft, nobody here reviewed or otherwise commented)

kevmoor:
Yes, and no.  It can be closed for now.  Thank you.

PR #9443, opened 2024-09-16
[SuiteSparse@7] Update to v7.8.2

ararslan:


ararslan:
cc @rayegun

imciner2:
This looks good to me. Note, after this new version gets registered, we need an equivalent version bump in SuiteSparse_GPU (we can't do them at the same time because the GPU package needs some of the libraries from this package, and they need to be at the correct version).

imciner2 (approved):

PR #9448, opened 2024-09-18
[S2Geometry] Bump version to 0.11.1 and bump OpenSSL compat

asinghvi17:
From their README it seems you can build s2geometry with OpenSSL v3.x, so I've updated that as well.


giordano:
Looks like the patch can be dropped and you need to add `-DBUILD_TESTS=OFF` to the cmake options.

evetion:
@asinghvi17 Can you also include `Platform("aarch64", "macos")` in this build? That would help me with actually wrapping it 😅

S/S2Geometry/build_tarballs.jl
```diff
@@ -3,23 +3,23 @@
...
 
 name = "S2Geometry"
-version = v"0.9.0"
+version = v"0.11.1"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/google/s2geometry.git", "a4dddf40647c68cd0104eafc31e9c8fb247a6308"),
+    GitSource("https://github.com/google/s2geometry.git", "5b5eccd54a08ae03b4467e79ffbb076d0b5f221e"),
     DirectorySource("./bundled"),
```
giordano, line 11:
```suggestion
```

S/S2Geometry/build_tarballs.jl
```diff
@@ -34,6 +34,7 @@ platforms = [
     Platform("x86_64", "linux"; libc="musl"),
     Platform("aarch64", "linux"; libc="musl"),
     Platform("x86_64", "macos"),
+    Platform("aarch64", "macos"),
```
giordano, line 37: Something like (I haven't counted the number of parentheses, adjust as needed)
```julia
platforms = supported_platforms(; exclude=p -> !((Sys.islinux(p) || Sys.isapple(p)) || nbits(p) == 64))
```
looks more automatic than manually listing the platforms

asinghvi17:
```suggestion
platforms = supported_platforms(; exclude=p -> !((Sys.islinux(p) || Sys.isapple(p)) && nbits(p) == 64))
```

S/S2Geometry/build_tarballs.jl
```diff
@@ -3,38 +3,30 @@
...
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/s2geometry
-atomic_patch -p1 ../patches/disable-s2testing.patch
 mkdir build && cd build
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
     -DBUILD_EXAMPLES=OFF \
+    -DBUILD_TESTS=OFF \
```
giordano, line 21:
```suggestion
    -DBUILD_TESTS=OFF \
```

asinghvi17:
This now works on Mac but I'm getting this error in all other builds:
```
This package requires GCC 7 or higher
C++ versions less than C++ 14 are not supported
```
from Abseil.

giordano:
> This package requires GCC 7 or higher

https://github.com/JuliaPackaging/Yggdrasil/blob/31fae03f806f176a5e085b2cadbcc4006ef1882d/A/abseil_cpp/build_tarballs.jl#L136

S/S2Geometry/build_tarballs.jl
```diff
@@ -3,38 +3,31 @@
...
+    -DCMAKE_CXX_STANDARD=17 \
```
giordano, line 22: Besides the fact there's another extra space, you shouldn't set the C++ standard manually (if you have to do that it's an upstream bug):
```suggestion
```
Instead, you should use the requested compiler.

asinghvi17:
> `error: ‘(1.2e+1l / 7.20575940379279305358983848622456e+16l)’ is not a constant expression`

but it is?  This only happens on PowerPC so we could also just drop that.

giordano:
> so we could also just drop that.

Yeah, that's fine.

aarch64-freebsd is a new platform and we don't have many packages built for that yet, it's fine to skip it for the time being.  For Windows, if the only issue is `M_PI` not being defined is the only issue (but it happens very early on in the compilation, there may be much more afterwards) then that's easy to fix, there's a macro to define to enable that definition, but if they aren't doing that already makes me think they don't really support Windows much.

S/S2Geometry/build_tarballs.jl
```diff
@@ -3,38 +3,31 @@
...
 # platforms are passed in on the command line
-platforms = [
-    Platform("x86_64", "linux"; libc="glibc"),
-    Platform("aarch64", "linux"; libc="glibc"),
-    Platform("powerpc64le", "linux"; libc="glibc"),
-    Platform("x86_64", "linux"; libc="musl"),
-    Platform("aarch64", "linux"; libc="musl"),
-    Platform("x86_64", "macos"),
-]
+platforms = supported_platforms(; exclude=p -> !(nbits(p) == 64) || Sys.isfreebsd(p))
```
giordano, line 30: x86_64 FreeBSD worked fine 🙂
```suggestion
platforms = supported_platforms()
# Only 64-bit platforms supported
filter!(p -> nbits(p) == 64, platforms)
# We are missing some dependencies for aarch64-freebsd, can be re-enabled in the future when we have them
filter!(p -> !(Sys.isfreebsd(p) && arch(p) == "aarch64", platforms)
```

S/S2Geometry/build_tarballs.jl
```diff
@@ -3,38 +3,31 @@
...
 script = raw"""
 cd $WORKSPACE/srcdir/s2geometry
-atomic_patch -p1 ../patches/disable-s2testing.patch
 mkdir build && cd build
 cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
     -DBUILD_EXAMPLES=OFF \
+    -DBUILD_TESTS=OFF \
+    -D_USE_MATH_DEFINES \
```
giordano, line 22: This is not how you define macros for the C/C++ compilers, that's only a CMake option.

asinghvi17: Ah darn, I thought it might pass through.

giordano:
> S2 does seem to set that if it detects MSVC:

Of course that's inaccurate, they should check WIN32 target instead.

S/S2Geometry/build_tarballs.jl
asinghvi17: Would that get passed to `make` then?

S/S2Geometry/bundled/patches/msvc_to_win32_target.patch
giordano, line 1: Can you please upstream this patch?

S/S2Geometry/build_tarballs.jl
```diff
@@ -3,38 +3,38 @@
...
-    Platform("aarch64", "linux"; libc="musl"),
-    Platform("x86_64", "macos"),
-]
+platforms = supported_platforms()
+# Only 64-bit platforms supported
+filter!(p -> nbits(p) == 64, platforms)
+# We are missing some dependencies (Abseil) for aarch64-freebsd and powerpc, 
+# can be re-enabled in the future when we have them
+filter!(p -> !((Sys.isfreebsd(p) && arch(p) == "aarch64") || arch(p) == "powerpc64le"), platforms)
+#
```
giordano, line 37:
```suggestion
# We are missing some dependencies (Abseil) for aarch64-freebsd, 
# can be re-enabled in the future when we have them
filter!(p -> !(Sys.isfreebsd(p) && arch(p) == "aarch64"), platforms)
# Compilation fails for powerpc:
#     /workspace/srcdir/s2geometry/src/s2/s2edge_crossings.cc:120:31: error: ‘(6.15348059642740421245081038903225e-15l / 5.40431955284459475358983848622456e+16l)’ is not a constant expression
filter!(p -> arch(p) != "powerpc64le", platforms)
```

S/S2Geometry/build_tarballs.jl
```diff
@@ -3,38 +3,40 @@
...
-    GitSource("https://github.com/google/s2geometry.git", "a4dddf40647c68cd0104eafc31e9c8fb247a6308"),
+    GitSource("https://github.com/google/s2geometry.git", "5b5eccd54a08ae03b4467e79ffbb076d0b5f221e"),
     DirectorySource("./bundled"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/s2geometry
-atomic_patch -p1 ../patches/disable-s2testing.patch
+atomic_patch -p1 ../patches/msvc_to_win32_target.patch
```
giordano, line 17:
```suggestion
# Upstream PR: https://github.com/google/s2geometry/pull/379
atomic_patch -p1 ../patches/msvc_to_win32_target.patch
```

giordano:
It's looking much better now 🙂

asinghvi17:
Thanks for all the help @giordano!!

PR #9453, opened 2024-09-19
New recipe: MEOS

evetion:
> MEOS (Mobility Engine, Open Source) is a C library and its associated API for manipulating temporal and spatiotemporal data. It is the core component of [MobilityDB](https://mobilitydb.com/), an open source geospatial trajectory data management & analysis platform built on top of [PostgreSQL](https://www.postgresql.org/) and [PostGIS](https://postgis.net/).

See https://libmeos.org/

### DIFF
--- a/M/MEOS/build_tarballs.jl
+++ b/M/MEOS/build_tarballs.jl
@@ -1,0 +1,41 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "MEOS"
+version = v"1.1.1"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/MobilityDB/MobilityDB.git", "4273bdd4f70ee6fb8b28cd269385d83da6eeba31")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/MobilityDB
+
+cmake -S -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DMEOS=ON
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DMEOS=ON
+cmake --build build --parallel ${nproc}
+cmake --install build
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; experimental=true)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libmeos", :libmeos)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="GEOS_jll", uuid="d604d12d-fa86-5845-992e-78dc15976526"))
+    Dependency(PackageSpec(name="JSON_C_jll", uuid="9cdfc4e7-e793-5089-b6f7-569a57a60f0a"))
+    Dependency(PackageSpec(name="Proj_jll", uuid="58948b4f-47e0-5654-a9ad-f609743f8632"))
+    Dependency(PackageSpec(name="GSL_jll", uuid="1b77fbbe-d8ee-58f0-85f9-836ddc23a7a4"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")


evetion:
This fails on PROJ not being found, because:
```┌ Warning: Dependency Proj_jll does not have a mapping for artifact Proj for platform x86_64-linux-gnu-libgfortran3-cxx03```.

However, I'm not sure how/why the libgfortran(3) gets injected, and why PROJ_jll has no such mapping, whereas GEOS_jll with a similar mapping does.

File: M/MEOS/build_tarballs.jl [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libmeos", :libmeos)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="GEOS_jll", uuid="d604d12d-fa86-5845-992e-78dc15976526"))
+    Dependency(PackageSpec(name="JSON_C_jll", uuid="9cdfc4e7-e793-5089-b6f7-569a57a60f0a"))
+    Dependency(PackageSpec(name="Proj_jll", uuid="58948b4f-47e0-5654-a9ad-f609743f8632"))
```
eschnett, line 35: `PROJ_jll`, not `Proj_jll`.

evetion: Wow, that's a simple fix, thanks for spotting it. 😅

imciner2:
> edit: This release now points to the release candidate 1 release (as it requires no patches) to build. Is it ok to set the version number to v1.2.0 already and later at the actual release just rerunning it with the new commits (producing a jll with the +1 postfix)?

I'd suggest not doing that. The postfix is not really liked by the general registry, so while we use it for the individual builds here, it would be better to not rely on it for this.

How soon will upstream release 1.2.0?

M/MEOS/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/MobilityDB/MobilityDB.git", "b5dfee7083beb85de772767971d52745f9b200ad")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/MobilityDB
+
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DMEOS=ON -DHAVE_X86_64_POPCNTQ_EXITCODE="FAILED_TO_RUN" -DHAVE_X86_64_POPCNTQ_EXITCODE__TRYRUN_OUTPUT=
```
imciner2, line 17:
```suggestion
cmake -B build \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DMEOS=ON \
    -DHAVE_X86_64_POPCNTQ_EXITCODE="FAILED_TO_RUN" \
    -DHAVE_X86_64_POPCNTQ_EXITCODE__TRYRUN_OUTPUT=
```
Style nit. Also, the last argument appears to be missing its assigned value.

evetion:
> > edit: This release now points to the release candidate 1 release (as it requires no patches) to build. Is it ok to set the version number to v1.2.0 already and later at the actual release just rerunning it with the new commits (producing a jll with the +1 postfix)?
> 
> I'd suggest not doing that. The postfix is not really liked by the general registry, so while we use it for the individual builds here, it would be better to not rely on it for this.
> 
> How soon will upstream release 1.2.0?

Fair enough, let's wait for the 1.2.0 release, but I'm not sure how soon that will be. For now, this is also still blocked on Windows (`Error: incorrect register '%eax' used with 'q' suffix`), for which I made an upstream issue.

mschoema:
We just produced the [v1.2.0 release](https://github.com/MobilityDB/MobilityDB/releases/tag/v1.2.0) of MEOS. Feel free to let us know if you encounter any issues with it.

M/MEOS/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libmeos", :libmeos)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="GEOS_jll", uuid="d604d12d-fa86-5845-992e-78dc15976526"))
+    Dependency(PackageSpec(name="JSON_C_jll", uuid="9cdfc4e7-e793-5089-b6f7-569a57a60f0a"))
+    Dependency(PackageSpec(name="PROJ_jll", uuid="58948b4f-47e0-5654-a9ad-f609743f8632"))
```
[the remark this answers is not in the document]
evetion: Thanks, good catch. I added these packages, to https://github.com/JuliaPackaging/Yggdrasil/issues/3024#issuecomment-2460984329.

M/MEOS/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "MEOS"
+version = v"1.2.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/MobilityDB/MobilityDB.git", "60048b5b4b7ce2f7560c024d1af024db73b3bd5b")
```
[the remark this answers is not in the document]
stemann: "Mobility Engine Open Source" also does not seem to be the obvious search hit for "MeOS" on Google.

Recommendation: Choose a longer, more explicit name, e.g., `MobilityEngineOpenSource`.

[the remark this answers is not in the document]
evetion: Agreed that Google doesn't help (right now), but I'm not sure if that's a good requirement? Eigen, much more famous, doesn't show up in my search results either. I suppose we tend to follow how the external library calls itself, which is MEOS. Is there a naming guide for when we deviate from the library name itself?

FWIW, it is supposed to mirror similarly named software (also recipes here), namely GEOS (but GDAL and PROJ also come to mind).

M/MEOS/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+script = raw"""
+cd $WORKSPACE/srcdir/MobilityDB
+
+cmake -B build \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DMEOS=ON \
+    -DHAVE_X86_64_POPCNTQ_EXITCODE="FAILED_TO_RUN" \
+    -DHAVE_X86_64_POPCNTQ_EXITCODE__TRYRUN_OUTPUT=
```
giordano, line 24: Why?  This is such an old instruction which should be moderately safe to assume to be available.

evetion: Apparently it's not the instruction, but the TRY_RUN statement is not allowed in cross-compile mode:
```
[14:57:59] CMake Error: TRY_RUN() invoked in cross-compiling mode, please set the following cache variables appropriately:
[14:57:59]    HAVE_X86_64_POPCNTQ_EXITCODE (advanced)
[14:57:59]    HAVE_X86_64_POPCNTQ_EXITCODE__TRYRUN_OUTPUT (advanced)
[14:57:59] For details see /workspace/srcdir/MobilityDB/build/TryRunResults.cmake
```

giordano: > but the TRY_RUN statement is not allowed in cross-compile mode:

I know.  But I expect that try-run is running a check whether the instruction is available.  Which of course could also be done with a simple compiler check instead of a try-run.

evetion: I just workarounded the error with the instruction CMake gave me (I'm not affiliated with MobilityDB/MEOS so I can't answer the why, but I'm happy to make an upstream issue). AFAIK there's no code in MEOS actually making use of this check (anymore https://github.com/MobilityDB/MobilityDB/pull/610).

imciner2: We use a cross-compilation environment, so any try_run call in a CMake script will basically be pointless. From the CMake docs (https://cmake.org/cmake/help/latest/command/try_run.html#behavior-when-cross-compiling):

> When cross compiling, the executable compiled in the first step usually cannot be run on the build host. The try_run command checks the [CMAKE_CROSSCOMPILING](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html#variable:CMAKE_CROSSCOMPILING) variable to detect whether CMake is in cross-compiling mode. If that is the case, it will still try to compile the executable, but it will not try to run the executable unless the [CMAKE_CROSSCOMPILING_EMULATOR](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING_EMULATOR.html#variable:CMAKE_CROSSCOMPILING_EMULATOR) variable is set. Instead it will create cache variables which must be filled by the user or by presetting them in some CMake script file to the values the executable would have produced if it had been run on its actual target platform. These cache entries are:

PR #9456, opened 2024-09-19
Update mlpack to latest stable version.

rcurtin:
Just a bump from 4.4.0 to 4.5.0.  Some functionality improvements, but nothing new from the side of Julia packaging.  Fingers crossed that everything builds right...


rcurtin:
Thanks @giordano!  You are awesome as always. :100:

PR #9467, opened 2024-09-20
Bump rr to 5.8

Keno:
Updates rr to current master, which in particular has https://github.com/rr-debugger/rr/commit/07415f58c423053ad024c69918a014275ef6f73f which fixes rr asserts seen on CI. Also picks up https://github.com/rr-debugger/rr/commit/c1ca64585a923f75b12df2eba259ab64bc5646a1, which was a patch we'd been carrying locally.


Keno:
cc @JeffBezanson

PR #9470, opened 2024-09-21
[LLVM_full@18] Rebuild for FreeBSD AArch64 support

ararslan:


giordano:
You also need to build the corresponding assert version

ararslan:
I had planned to do that in a subsequent PR, do you mean it should be part of this one?

giordano:
Yes, we usually do them both in a single PR.

giordano:
@ararslan can you please follow up with all the other related packages? It's quite important to have consistent version numbers among all of them

ararslan:
By related packages do you mean like Clang, LLD, etc.? Basically just all of the other things that are in `L/LLVM/`?

giordano:
Yes, also libllvm and such.  See the history of the directory to see what's the chain (I don't know it myself)

ararslan:
It looks like it's LLVM_full then libLLVM (just opened a PR) then Clang, LLD, MLIR, and LLVM in any order.

PR #9473, opened 2024-09-21
Add builder for whereami

ararslan:
Cross-platform library for identifying where on the filesystem the current executable lives.

### DIFF
--- a/W/WhereAmI/build_tarballs.jl
+++ b/W/WhereAmI/build_tarballs.jl
@@ -1,7 +1,7 @@
 using BinaryBuilder
 
 name = "WhereAmI"
-version = v"2024.8.6"  # no proper versions/releases, this is the date of the commit used
+version = v"1.0.0"  # no proper versions/releases, this is made up
 
 sources = [
     GitSource("https://github.com/gpakosz/whereami",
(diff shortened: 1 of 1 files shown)


giordano:
I think `-ldl` is needed only with glibc?  And not even with [glibc 2.34 anymore](https://lists.gnu.org/archive/html/info-gnu/2021-08/msg00001.html), but we still use 2.17.

ararslan:
Nice, it appears you are correct, thank you!

W/WhereAmI/build_tarballs.jl
```diff
@@ -0,0 +1,32 @@
+using BinaryBuilder
+
+name = "WhereAmI"
+version = v"2024.8.6"  # no proper versions/releases, this is the date of the commit used
```
giordano, line 4: For lack of proper veesioning, how about something like 0.1.0 or 1.0.0? Calendar-based veesioning plays very badly with semver

ararslan: Sure, makes sense
```suggestion
version = v"1.0.0"  # no proper versions/releases, this is made up
```

PR #9476, opened 2024-09-23
[ECharts] Add recipe for v5.5.1

mofeing:


E/ECharts/build_tarballs.jl
```diff
@@ -0,0 +1,23 @@
...
+
+name = "ECharts"
+version = v"5.5.1"
+sources = GitSource[]
+
+script = "version=$(version)\n" * raw"""
+apk add --update npm
+cd ${prefix}
+npm install echarts@${version}
+install_license ${prefix}/node_modules/echarts/licenses/LICENSE-d3
```
giordano, line 11: Is this right?

mofeing: I based my recipe on https://github.com/JuliaPackaging/Yggdrasil/blob/master/A/azurite/build_tarballs.jl

giordano: I meant to say the path of the file, it refers to d3.  Looking at https://github.com/apache/echarts/blob/5cb2a72cefd684a326b83a9343a69c2367083b89/LICENSE, that's an apache license, the file you're installing looks like BSD 3-clause

mofeing: Ah, that seems weird. Anyway, I've found the correct LICENSE file in `node_modules/echarts/LICENSE`.

PR #9479, opened 2024-09-23
[SuiteSparse] Rebuild SuiteSparse extras that weren't registered

rayegun:
Previous builds didn't get registered. These should be fine to register as the SuiteSparse in 1.11 will be the right one.

### DIFF
--- a/S/SuiteSparse/CXSparse/build_tarballs.jl
+++ b/S/SuiteSparse/CXSparse/build_tarballs.jl
@@ -33,7 +33,5 @@ dependencies = append!(dependencies, [
 products = [
     LibraryProduct("libcxsparse", :libcxsparse)
 ]
-build_tarballs(ARGS, name, version, sources, script,
-               # REMOVE THIS FILTER for 7.8+
-               filter!(p -> !(Sys.isfreebsd(p) && arch(p) == "aarch64"), platforms), 
+build_tarballs(ARGS, name, version, sources, script, platforms,  
                products, dependencies; julia_compat="1.11", preferred_llvm_version=LLVM_version)
--- a/S/SuiteSparse/CXSparse/build_tarballs.jl
+++ b/S/SuiteSparse/CXSparse/build_tarballs.jl
@@ -20,7 +20,7 @@ if [[ "${target}" == aarch64-apple-* ]]; then
         -DCMAKE_SHARED_LINKER_FLAGS="-L${libdir}/darwin -lclang_rt.osx"
     )
 fi
-""" * build_script(true)
+""" * build_script(; use_omp=true)
 
 # Add dependency on SuiteSparse_jll
 dependencies = append!(dependencies, [
@@ -33,5 +33,7 @@ dependencies = append!(dependencies, [
 products = [
     LibraryProduct("libcxsparse", :libcxsparse)
 ]
-build_tarballs(ARGS, name, version, sources, script, platforms, products, 
-               dependencies; julia_compat="1.11", preferred_llvm_version=LLVM_version)
+build_tarballs(ARGS, name, version, sources, script,
+               # REMOVE THIS FILTER for 7.8+
+               filter!(p -> !(Sys.isfreebsd(p) && arch(p) == "aarch64"), platforms), 
+               products, dependencies; julia_compat="1.11", preferred_llvm_version=LLVM_version)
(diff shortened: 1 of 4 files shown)


rayegun:
Huh these are totally new errors, this exact same PR worked previously, it just wasn't registered due to 1.11 compat issues at the time.

rayegun:
@giordano looks like SuiteSparse_jll 7.7 skipped FreeBSD. But now I'm confused. Julia 1.11 is releasing with 7.7 as far as I can tell.

Do we leave these red on aarch64-unknown-freebsd and 1.12 will have SuiteSparse 7.8+ on all platforms? Is aarch64-freebsd new?

So I suppose we can merge these for 7.7 and then I can rebuild for 7.8+

giordano:
Is this good to go now?

ararslan:
If https://github.com/JuliaPackaging/Yggdrasil/pull/9483 seems reasonable then we could do that first, which seems like it would make this PR's changes easier? We'd at least have FreeBSD AArch64 for 7.7 so it wouldn't require a follow-up.

S/SuiteSparse/CXSparse/build_tarballs.jl
```diff
@@ -33,5 +33,7 @@ dependencies = append!(dependencies, [
 products = [
     LibraryProduct("libcxsparse", :libcxsparse)
 ]
-build_tarballs(ARGS, name, version, sources, script, platforms, products, 
-               dependencies; julia_compat="1.11", preferred_llvm_version=LLVM_version)
+build_tarballs(ARGS, name, version, sources, script,
+               # REMOVE THIS FILTER for 7.8+
+               filter!(p -> !(Sys.isfreebsd(p) && arch(p) == "aarch64"), platforms),
```
ararslan, line 38: The filters can now be removed since 7.7.0 has been built for FreeBSD AArch64

rayegun:
This should be ready @giordano

S/SuiteSparse/SPEX/build_tarballs.jl
```diff
@@ -29,11 +28,11 @@ dependencies = append!(dependencies, [
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
     Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
     Dependency("GMP_jll"; compat="6.2.1"),
-    Dependency("MPFR_jll"; compat="4.1.1"),
+    Dependency("MPFR_jll"; compat="4.2.1"),
```
ararslan, line 31: Does changing the compat version here require also bumping the version of SPEX?

giordano: Usual reminder that changing these compat bounds (and I'm ***not*** suggesting removing them) will make the package uninstallable in julia v1.11-

rayegun: I'm sorry does that mean only in 1.12 and onwards? Or does it include 1.11? 

I changed the compat bounds again because of FreeBSD. We'd need older MPFR built for freeBSD or to filter on older Julia otherwise

giordano: > I'm sorry does that mean only in 1.12 and onwards? Or does it include 1.11?

1.11- includes 1.11, so this would be installable only in Julia v1.12+.

> I changed the compat bounds again because of FreeBSD. We'd need older MPFR built for freeBSD or to filter on older Julia otherwise

There's also the option to leave aarch64-freebsd out 🙂 I appreciate building packages for the new platform when it's feasible, but I'm not really sure it's worth wrecking the ecosystem for a platform with 0.5±0.5 users.

ararslan: I can build older versions of things as needed to facilitate making more minimal changes. If you need an older MPFR I can do that, or you can filter out aarch64-unknown-freebsd (but leave x86_64-unknown-freebsd).

rayegun: If you want to build older mpfr that would be great, saves us time going back in the future :)

ararslan: https://github.com/JuliaPackaging/Yggdrasil/pull/9543

ararslan:
```suggestion
    Dependency("MPFR_jll"; compat="4.1.1"),
```
Should be all set now (https://github.com/JuliaRegistries/General/pull/116169)

ararslan (approved):

rayegun:
> Is this good to go now?

Yes!

rayegun:
@giordano can we merge this?

PR #9480, opened 2024-09-23
Update OpenSSH and add `sftp` as a product

JamesWrigley:
Updated to 9.5.0 on Windows, which seems to be the latest available.


JamesWrigley:
The `aarch64-unknown-freebsd` build was failing because libcrypto wasn't found, I'm guessing that's because OpenSSH_jll was only rebuilt recently in #9471. I bumped the compat bound to 3.0.15 in a1994801, with luck that'll work :crossed_fingers:

PR #9483, opened 2024-09-23
[SuiteSparse@7] Temporarily set version back to 7.7.0

ararslan:
Julia 1.11 uses SuiteSparse 7.7 and Julia 1.12 uses SuiteSparse 7.8, so separating the builders permits easier per-Julia-version maintenance.

cc @rayegun @giordano

### DIFF
--- a/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
+++ b/S/SuiteSparse/SuiteSparse@7/build_tarballs.jl
@@ -1,8 +1,8 @@
-# Build counter: 0
+# Build counter: 1
 include("../common.jl")
 
 name = "SuiteSparse"
-version = v"7.8.2"
+version = v"7.7.0"
 
 sources = suitesparse_sources(version)
 


ararslan:
Hm, but then do we keep making 7.x directories for each new minor version of SuiteSparse we build? 🤔 That seems suboptimal. Maybe we just add `SuiteSparse@7.7` and continue to use `SuiteSparse@7` instead of making a new `SuiteSparse@7.8`?

ViralBShah:
You can kick off a build for 7.7 by bumping down, and once merged, get back to 7.8. I assume this is a one time thing for freebsd aarch64.

ViralBShah:
In the past we have not really kept old configurations around to match Julia versions, although one perhaps should do that going forward, especially since 1.10 will be the new LTS.

giordano:
> Relatedly, when setting the version back to 7.8, do I use [skip build] or something to avoid registering a new build number for 7.8?

Yes please:

ararslan:
I've updated this PR to set the version for SuiteSparse@7 back to 7.7.0. Once merged, I'll submit a follow-up PR to set it back to 7.8.2 and include `[skip build]` in the commit message to avoid making a new build.

giordano:
Looks good to me, but to be clear, do you want to rename the directory to something like `SuiteSparse@7.7`?

ararslan:
No, I figured it's not worth keeping around the separate directory for 7.7. We can leave it as 7 and I'll change the version back.

PR #5904, opened 2022-11-21
[MPFR] 4.1.1

simonbyrne:
This is an actual 4.1.1 release

PR #9543, opened 2024-09-27
[MPFR] Temporarily set version back to 4.1.1 for rebuild

ararslan:
There were no functional changes between #5904 and now; all that should be needed is the version change. Once merged, I'll revert this with `[skip build]`.


giordano:
Correct

PR #9565, opened 2024-10-01
[PETSc] version 3.22.0

boriskaus:
PETSc version 3.22.0


ViralBShah:
I think you just need to filter out the failing platforms.

boriskaus:
Feel free to push that; I don’t have my computer for the coming few days

PR #9573, opened 2024-10-04
Reactant: fix bazel build

wsmoses:
@mofeing


mofeing:
JAJAJA this is so horrible, but let's go with it

PR #9578, opened 2024-10-08
wolfSSL: Update to 5.7.2

dantaras:


giordano:
Thanks for your contribution!

giordano:
[SecTrustEvaluateWithError](https://developer.apple.com/documentation/security/sectrustevaluatewitherror(_:_:)) requires macOS 10.14, you should use that SDK instead of skipping macOS entirely :slightly_smiling_face:

dantaras:
How can I get it to use that version? Do you have an example?

giordano:
> How can I get it to use that version? Do you have an example?

https://github.com/JuliaPackaging/Yggdrasil/blob/46da732ee03cd19542a62a137b05893179f0d073/A/AMReX/build_tarballs.jl#L16-L17
https://github.com/JuliaPackaging/Yggdrasil/blob/46da732ee03cd19542a62a137b05893179f0d073/A/AMReX/build_tarballs.jl#L24-L31

dantaras:
Thanks, tried something different, still waiting to see if it worked.

W/wolfSSL/build_tarballs.jl
```diff
@@ -16,13 +16,18 @@ cd $WORKSPACE/srcdir/wolfssl*
 
 mkdir build && cd build
 
+if [[ "${target}" == *-apple-darwin* ]]; then
+    CFLAGS="-mmacosx-version-min=10.14"
+fi
```
giordano, line 21: 1. You're matching also aarch64-apple-darwin, which doesn't have the same problem (we use the 11.0 SDK there)
2. it isn't sufficient to pretend to have the 10.14 SDK, you actually need to use it as suggested in https://github.com/JuliaPackaging/Yggdrasil/pull/9578#issuecomment-2399871955 :slightly_smiling_face:

Side node, `export MACOSX_DEPLOYMENT_TARGET=10.14` is a nicer way to accomplish this, since our compiler wrappers read the `MACOSX_DEPLOYMENT_TARGET` env var, the way you do it the compiler gets duplicate and conflicting `-mmacosx-version-min` options.

dantaras: Thanks! Did that and it worked. Any chance we can merge this in without the Windows build?

dantaras: Trying something for the windows build now.

W/wolfSSL/bundled/patches/mingw32-link-with-ws2_32.patch
giordano, line 1: Can you please upstream this patch so that we'll be able to drop it in the future?  Thanks!

dantaras: Opened a PR here: https://github.com/wolfSSL/wolfssl/pull/8068. It'll be a while until that's merged and released though.

ViralBShah:
Off-topic:

@giordano It seems that we now require "Update to latest master" button to be hit before merging is allowed. Is that by design - since it probably increases the CI load a fair bit. Sorry to ask here, but I saw it on a couple of PRs (and it may have already been discussed somewhere I didn't see it).

giordano:
Thanks for pointing it out, that was not intended, my mistake while I was reviewing branch protection settings, I disabled that.

PR #9580, opened 2024-10-08
[Libxc] version 6.2.2 + CUDA binaries for aarch64

abussy:
Update Libxc version from 6.1.0 to 6.2.2.

Additionally, added recipes for the combination of `aarch64` and `CUDA`, for Libxc_GPU_jll. Because BinaryBuilder only runs on `x86_64`, and `nvcc` is not a cross compiler, this is a non-trivial operation.

Pre-compiled binaries are available on a separate GitHub [repository](https://github.com/abussy/libxc_aarch64_gpu_binaries), specifically created for this purpose. The `aarch64` BinaryBuilder recipe downloads the binary  for the given `CUDA` version (from 11.4 to 12.6), and repackages it in the correct format.

The binaries were all generated in Docker images, for total control of the environment and reproducibility. All scripts and recipes are available in https://github.com/abussy/libxc_aarch64_gpu_binaries as well. After a discussion on the Julia [slack](https://julialang.slack.com/archives/C674ELDNX/p1726747412673179), special care was given to the `glibc` version and `RPATH`. All `glibc` symbols used by the `libxc.so` shared library come from version 2.17, as per the BinaryBuilder requirements for `aarch64`.

Both `x86_64` and `aarch64` GPU binaries were succesfully tested using the Libxc.jl and DFTK.jl packages.

### DIFF
--- a/L/Libxc/Libxc/build_tarballs.jl
+++ b/L/Libxc/Libxc/build_tarballs.jl
@@ -1,7 +1,7 @@
 using BinaryBuilder, Pkg
 
 name = "Libxc"
-version = v"6.1.0"
+version = v"6.2.2"
 include("../sources.jl")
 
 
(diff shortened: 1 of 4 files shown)


abussy:
Closed since Libxc v7.0.0 came out in the mean time.

PR #9587, opened 2024-10-09
[GDAL] Add LERC v4 and use latest PROJ

visr:
Needs:
- [x] #9586
- [x] #9589

The recent LERC_jll v4 release broke GDAL_jll since it was an implicit dependency. See https://github.com/JuliaGeo/GDAL.jl/issues/179.

This tries to make a new release compatible with LERC_jll v4, and the latest breaking PROJ release. Since PROJ is also a dependency of libgeotiff this needs #9586 first.

It also adds compats for dependencies that did not yet have them.


ViralBShah:
Good to merge?

visr:
Yes, thanks.

PR #9588, opened 2024-10-09
[GDAL] Add LERC v3 only

visr:
A more minimal predecessor to #9587 to hopefully be able to fix https://github.com/JuliaGeo/GDAL.jl/issues/179 more quickly.

### DIFF
--- a/G/GDAL/build_tarballs.jl
+++ b/G/GDAL/build_tarballs.jl
@@ -112,6 +112,8 @@ install_license ../LICENSE.TXT
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = expand_cxxstring_abis(supported_platforms())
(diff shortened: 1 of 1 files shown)


ViralBShah:
@ararslan I think we need LibCURL_jll built for freebsd aarch64 for 8.6.1, which ships in Julia 1.11. (https://github.com/JuliaBinaryWrappers/LibCURL_jll.jl/releases/tag/LibCURL-v8.6.0%2B0)

giordano:
Just ignore aarch64-freebsd for the time being.

I strongly recommend leaving curl as is at the moment and focusing on finishing https://github.com/JuliaLang/julia/pull/53891, I ran out of solutions and no one else seems to be caring, but we must get that done since mbedtls 2 is out of support now.

visr:
Done, this should be good to go.

PR #9596, opened 2024-10-11
[algoim] patches to compile with clang 16

ericneiva:
Thank you.


A/algoim/build_tarballs.jl
```diff
@@ -3,8 +3,9 @@ using BinaryBuilder
 name = "algoim"
 version = v"0.1.0"
 sources = [
-    GitSource("https://github.com/algoim/algoim.git", "979cb3b7860b53751291352a929754d525b7fde1"),
-]
+    GitSource("https://github.com/ericneiva/algoim.git", "e80020a5182597a2392194651f62fb6f4c631082"),
+] # Required to compile with Apple's clang 16. 
+# See https://github.com/algoim/algoim/pull/8 and https://github.com/algoim/algoim/pull/6
```
giordano, line 8: Can you please apply those as patches, rather than using a fork?  Thanks!

ericneiva: Hi, @giordano, I believe I apply them now as patches. Let me know if you meant something else. Thanks!

giordano: Yes, but the patch doesn't seem to apply.

ericneiva: Hi, @giordano, it should be ok now, one of the patches required CRLF line endings, apparently.

Let me know if I need to apply any other changes.

Thanks a lot.

A/algoim/bundled/patches/.gitattributes
```diff
@@ -0,0 +1 @@
+0002-std-result_of-deprecated-in-C-17-remove-in-C-20-std-.patch -text diff
```
giordano, line 1:
```suggestion
0002-std-result_of-deprecated-in-C-17-remove-in-C-20-std-.patch -text diff

```

PR #9614, opened 2024-10-15
New: LVGL - Light and Versatile Graphics Library build_tarballs.jl

notinaboat:
LVGL - Light and Versatile Graphics Library
https://github.com/lvgl/lvgl

### DIFF
--- a/L/LVGL_SDL/build_tarballs.jl
+++ b/L/LVGL_SDL/build_tarballs.jl
@@ -1,0 +1,106 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "LVGL_SDL"
+version = v"9.2.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/lvgl/lv_port_pc_eclipse.git", "ba45efb68b125aa93237e7c75c51d4cd65fc10cf")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd lv_port_pc_eclipse/
+git submodule update --init --recursive
+cd lvgl/
+patch -p1 << "EOF"
+diff --git a/env_support/cmake/custom.cmake b/env_support/cmake/custom.cmake
+index e4cd36376..0e1d4b9d0 100644
+--- a/env_support/cmake/custom.cmake
++++ b/env_support/cmake/custom.cmake
+@@ -26,6 +26,7 @@ file(GLOB_RECURSE THORVG_SOURCES ${LVGL_ROOT_DIR}/src/libs/thorvg/*.cpp ${LVGL_R
+ # Build LVGL library
+ add_library(lvgl ${SOURCES})
+ add_library(lvgl::lvgl ALIAS lvgl)
++target_link_libraries(lvgl lvgl_thorvg ${SDL2_LIBRARIES})
+ 
+ target_compile_definitions(
+   lvgl PUBLIC $<$<BOOL:${LV_LVGL_H_INCLUDE_SIMPLE}>:LV_LVGL_H_INCLUDE_SIMPLE>
+@@ -49,7 +50,7 @@ if(NOT LV_CONF_BUILD_DISABLE_THORVG_INTERNAL)
+     add_library(lvgl_thorvg ${THORVG_SOURCES})
+     add_library(lvgl::thorvg ALIAS lvgl_thorvg)
+     target_include_directories(lvgl_thorvg SYSTEM PUBLIC ${LVGL_ROOT_DIR}/src/libs/thorvg)
+-    target_link_libraries(lvgl_thorvg PUBLIC lvgl)
++    #    target_link_libraries(lvgl_thorvg PUBLIC lvgl)
+ endif()
+ 
+ if(NOT (CMAKE_C_COMPILER_ID STREQUAL "MSVC"))
+EOF
+
+patch -p1 << "EOF"
+diff --git a/src/drivers/sdl/lv_sdl_window.c b/src/drivers/sdl/lv_sdl_window.c
+index df8a28e4e..40c5c0bcd 100644
+--- a/src/drivers/sdl/lv_sdl_window.c
++++ b/src/drivers/sdl/lv_sdl_window.c
+@@ -438,7 +438,12 @@ static void * sdl_draw_buf_realloc_aligned(void * ptr, size_t new_size)
+     /* Size must be multiple of align, See: https://en.cppreference.com/w/c/memory/aligned_alloc */
+ 
+ #define BUF_ALIGN (LV_DRAW_BUF_ALIGN < sizeof(void *) ? sizeof(void *) : LV_DRAW_BUF_ALIGN)
+-    return aligned_alloc(BUF_ALIGN, LV_ALIGN_UP(new_size, BUF_ALIGN));
++    void* result;
++    if (posix_memalign(&result, BUF_ALIGN, LV_ALIGN_UP(new_size, BUF_ALIGN)) == 0)
++    {
++        return result;
++    }
++    return NULL;
+ #else
+     return _aligned_malloc(LV_ALIGN_UP(new_size, LV_DRAW_BUF_ALIGN), LV_DRAW_BUF_ALIGN);
+ #endif /* _WIN32 */
+EOF
+
+cd ..
+patch -p1 << "EOF"
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 78eaf02..1753f9f 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1,6 +1,8 @@
+ cmake_minimum_required(VERSION 3.10)
+ project(lvgl)
+ 
++option(BUILD_SHARED_LIBS "Build shared libraries" ON)
++
+ option(LV_USE_DRAW_SDL "Use SDL draw unit" OFF)
+ option(LV_USE_LIBPNG "Use libpng to decode PNG" OFF)
+ option(LV_USE_LIBJPEG_TURBO "Use libjpeg turbo to decode JPEG" OFF)
+EOF
+
+cmake -B build -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+cmake --build build --parallel ${nproc}
+cmake --install build
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "linux"; libc="glibc")
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("liblvgl", :liblvgl),
+    LibraryProduct("liblvgl_thorvg", :liblvgl_thorvg)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency(PackageSpec(name="SDL2_jll", uuid="ab825dc5-c88e-5901-9575-1e5e20358fcf"))
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"13.2.0")


L/LVGL_SDL/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+      -DCMAKE_BUILD_TYPE=Release
+cmake --build build --parallel ${nproc}
+cmake --install build
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "linux"; libc="glibc")
```
imciner2, line 42: This is a very odd list of platforms. Why does it only build on these two specific ones?

L/LVGL_SDL/bundled/patches/lvgl_sdl.patch
imciner2, line 1: This patch has a lot of changes, and they aren't documented. Why is this patch needed here in Yggdrasil, and will you send it upstream so we don't have to try keeping it updated when a new version is introduced?

L/LVGL_SDL/bundled/patches/x86_64-apple-darwin14/alloc_aligned.patch
imciner2, line 1: Has this been sent upstream? I would guess we aren't the only ones seeing this issue.

L/LVGL_SDL/build_tarballs.jl
notinaboat: These are the only two platforms that I am able to test (my dev laptop and my embedded target board).

L/LVGL_SDL/bundled/patches/lvgl_sdl.patch
notinaboat: That's a fair question. I probably should have split the patch up to make the purpose of the different parts more obvious.

Here is a summary:

`CMakeLists.txt` and `custom.cmake`  -- This is cmake configuration to build LVGL as shared libraries and using the SDL2 backend  (based on the example [here](https://github.com/lvgl/lv_port_pc_eclipse/blob/master/CMakeLists.txt)).

`lv_conf_template.h` -- This is build configuration to enable SDL2 integration (based on the example [here](https://github.com/lvgl/lv_port_pc_eclipse/blob/master/lv_conf.h))

`lv_font.[ch]` -- This is an access function to allow access to `const struct` font definitions.  As far as I know, [Libdl](https://docs.julialang.org/en/v1/stdlib/Libdl/) does not provide a way to access a `const struct` defined in a dynamic library (`dlsym` only returns function pointers). This access function is probably not interesting to the upstream project. I assume they are focused on supporting users who are accessing the library from C.

L/LVGL_SDL/bundled/patches/x86_64-apple-darwin14/alloc_aligned.patch
notinaboat: I doubt anyone else would be using the [very old (2016) version of the MacOS system headers](https://github.com/JuliaPackaging/Yggdrasil/blob/a30c6b78ceb73811653f4fa2f9839629400d5ce8/0_RootFS/PlatformSupport/build_tarballs.jl#L72) that are used by BinaryBuilder.jl
So, this isn't suitable to send upstream. It's just a hack to build using Yggdrasil's outdated cross compiler setup.

L/LVGL_SDL/build_tarballs.jl
fingolfin: That's not a great reason to limit the list: it's not expected that you personally test all platforms. But by opening up, you vastly expand who can benefit from this JLL; and if anyone uses them, and there are problems, then they may give you feedback or may even help out.

Of course if other platforms cause build errors that are hard to track down, it's OK to remove them and worry about those you need first. But looking at this list I'd be surprised if not at least Linux x86 and mac ARM would work, too.

So how about starting with the full list, and see if any build jobs fail here on Yggdrasil, and then we can see how to proceed :-)

L/LVGL_SDL/bundled/patches/lvgl_sdl.patch
fingolfin: Yeah ideally each of those should be a separate patch. Bonus points if they use the "git commit patch format" ([such as here](https://github.com/JuliaPackaging/Yggdrasil/blob/master/L/libjulia/bundled/patches/1.11.1/bf16-workaround.patch)) which includes spaces for a "commit message" that can contain explanations such as yours above.

My experience is that you yourself will greatly benefit from this when you try to update this recipe in the future :-).

L/LVGL_SDL/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "LVGL_SDL"
+version = v"9.3.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/lvgl/lvgl.git", "4f086111a107718b719a45e2cc422f5b756d7edd"),
```
fingolfin, line 10: Why is this using a git snapshot and not a released version? In particular it seems the latest released version is 9.2.0, so why do you call this snapshot "9.3.0" ?

fingolfin: It's true they don't offer stable release tarballs, just the the autogenerated ones (my guess would be that as usual they are not even aware that those archives are not stable). But they still have release tags...

And by the discussion in https://github.com/lvgl/lvgl/issues/7098 and https://github.com/lvgl/lvgl/issues/6763 it seems 9.3.0 is only planned for early 2025, so it seems unlikely to match what is being used here as 9.3.0

kisvegabor:
Hey,

I've just saw in [this LVGL issue](https://github.com/lvgl/lvgl/issues/6763) that your integrating LVGL into Julia. It's very exciting and I'd love to 
- hear more about the motivation/use cases
- hear any technical questions/issues
- spread the word of this integration

L/LVGL_SDL/bundled/patches/x86_64-apple-darwin14/alloc_aligned.patch
stemann: It is also possible use a more recent macOS SDK, cf. e.g. `Qt6Base` or `cmt` recipes.

fingolfin:
@notinaboat hi there, are you still interested in completing this? We left a bunch of change suggestions but there was no activity in the past 12 months :-(

notinaboat:
Hi @fingolfin, as I recall I reached an impasse when I looked into the MacOS SDK issue.
@stemann pointed to a more recent SDK being used by Qt6Base.
It looks like the Qt6Base recipe [downloads the SDK](https://github.com/JuliaPackaging/Yggdrasil/blob/master/Q/Qt6Base/build_tarballs.jl#L17) from a [3rd party git repository](https://github.com/roblabla/MacOSX-SDKs) that has Apple's license agreement removed.
The Apple agreement includes things like "you agree not to use ... on any non-Apple-branded computer, or to enable others to do so", "you may not make the Developer Software available over a network", "You may not ... redistribute".
IANAL, but it does not seem proper to use a copy of the SDK that comes from a source that is clearly in violation of the licence.
Given that the SDK can only be used on apple hardware it seems like it would be better to just use the SDKs installed on the apple hardware. I assume that BinaryBuilder runs Docker on top of MacOS, Maybe the SDK folders could be exported in to the Docker image somehow?

fingolfin:
Actually there's a `require_macos_sdk` helper these days which makes it easy to install alternative macOS SDKs.

L/LVGL_SDL/build_tarballs.jl
fingolfin:
```suggestion
    GitSource("https://github.com/lvgl/lvgl.git", "c033a98afddd65aaafeebea625382a94020fe4a7"),
```
This is the actual 9.3.0 release commit.

Though by now they are at 9.4.0

L/LVGL_SDL/bundled/patches/x86_64-apple-darwin14/alloc_aligned.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/src/drivers/sdl/lv_sdl_window.c b/src/drivers/sdl/lv_sdl_window.c
+index 9a89271c5..4fdf52760 100644
+--- a/src/drivers/sdl/lv_sdl_window.c
++++ b/src/drivers/sdl/lv_sdl_window.c
+@@ -410,7 +410,11 @@ static void * sdl_draw_buf_realloc_aligned(void * ptr, size_t new_size)
+     /* Size must be multiple of align, See: https://en.cppreference.com/w/c/memory/aligned_alloc */
+ 
+ #define BUF_ALIGN (LV_DRAW_BUF_ALIGN < sizeof(void *) ? sizeof(void *) : LV_DRAW_BUF_ALIGN)
+-    return aligned_alloc(BUF_ALIGN, LV_ALIGN_UP(new_size, BUF_ALIGN));
++    // MacOS SDK version 14 does not support aligned_malloc()
```
fingolfin, line 10:
```suggestion
+    // MacOS SDK version 14 does not support aligned_alloc()
```

fingolfin: Also, do you really mean SDK 14? Because this PR was still building against SDK 11 (on ARM) resp. 10.14 (on Intel).

L/LVGL_SDL/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
```
fingolfin, line 4: Step 1 for using `require_macos_sdk`:
```suggestion

const YGGDRASIL_DIR = "../.."
include(joinpath(YGGDRASIL_DIR, "platforms", "macos_sdks.jl"))

```

L/LVGL_SDL/build_tarballs.jl
```diff
@@ -0,0 +1,58 @@
...
+      -DLV_CONF_BUILD_DISABLE_DEMOS=1 \
+      -DLV_CONF_BUILD_DISABLE_EXAMPLES=1 \
+      -DLV_CONF_INCLUDE_SIMPLE=OFF \
+      -DCMAKE_INSTALL_PREFIX=${prefix} \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+      -DCMAKE_BUILD_TYPE=Release
+cmake --build build --parallel ${nproc}
+cmake --install build
+"""
+
```
fingolfin, line 37: Step 2 for using `require_macos_sdk`:
```suggestion

sources, script = require_macos_sdk("11.1", sources, script)

```
Here 11.1 is just an example. You can also use something newer (ideally the older the better because then more people can use it, but of course it must be new enough for the package to build. Look at the file `platforms/macos_sdks.jl` to see which SDKs we currently have there; more can be added as needed)

notinaboat:
> Actually there's a require_macos_sdk helper these days which makes it easy to install alternative macOS SDKs.

Hi @fingolfin, I've looked at require_macos_sdk.  It seems to still download SDKs from various github user accounts. 

I've been trying to understand a bit more about how the buildkite setup works. The [buildkite docs](https://buildkite.com/docs/pipelines/hosted-agents/macos#macos-instance-software-support) say that the MacOS SDKs are pre-installed.

[This FAQ](https://docs.binarybuilder.org/stable/FAQ/#What-is-this-I-hear-about-the-macOS-SDK-license-agreement?) led be to believe that BinaryBuilder runs Docker on Apple hardware in order to comply with the SDK agreement ("The Docker runner implements the virtualization approach on macOS machines").

It looks like the buildkite [config used by JuliaCI](https://github.com/JuliaCI/julia-buildkite/blob/7bc1cd3a7c90c04d998dc9bce59ff5ac7c70863d/pipelines/main/platforms/build_macos.yml#L15) is using Apple hardware, but the [config used by Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil/blob/1186d237e7e9811e09f691ee3d8c66d1e3ecb428/.buildkite/pipeline.yml#L18-L19) uses non-Apple hardware and [bypasses the SDK agreement check](https://github.com/JuliaPackaging/Yggdrasil/blob/1186d237e7e9811e09f691ee3d8c66d1e3ecb428/.buildkite/utils.jl#L80).

Am I missing some detail about how this works? (Buildkite is new to me).

fingolfin:
The FAQ mentions what happens when a user tries to run BinaryBuilder on macOS: namely, it uses Docker; and Docker always runs Linux images (not macOS images). So no matter where you run BB, be it on a macOS or a Linux host, in the end all compilation is performed in a Linux environment by cross compilers.

notinaboat:
> in the end all compilation is performed in a Linux environment by cross compilers.

OK. Understood.

The remaining issue is that of the underlying hardware on buildkite.
Is it intended that Yggdrasil uses non-apple buildkite hardware to create macOS binaries?
Or is that just a configuration mistake?

fingolfin:
I don't understand why you think that it matters on which hardware the Linux VM performing the build runs?

notinaboat:
> I don't understand why you think that it matters on which hardware the Linux VM performing the build runs?

As the [FAQ](https://docs.binarybuilder.org/stable/FAQ/#What-is-this-I-hear-about-the-macOS-SDK-license-agreement?) says, the [Apple and Xcode SDK agreement](https://images.apple.com/legal/sla/docs/xcode.pdf) restricts distribution and usage of the macOS SDK.
The agreement is unambiguous with respect to use of non-apple hardware:

> You are **expressly prohibited** from separately **using the Apple SDKs** or attempting to run any
part of the Apple Software on **non-Apple-branded hardware**.

> [Permitted Uses:] "Install a reasonable number of copies of the Apple Software on **Apple-branded computers** that
are owned or controlled by You to be used internally by You or Your Authorized Developers"

> The grants set forth in this Agreement **do not permit** You to, and You agree not to, install, use or run the
Apple Software or Apple Services on any **non-Apple-branded computer** or device, or to **enable others to
do so**.

PR #9626, opened 2024-10-17
MAGEMin v1.5.5

NicolasRiel:
- Added elements names to structure output -> for MAGEMinApp calculator
- Corrected calculation of apfu
- Slight discrepancy in density calculation has been solver


NicolasRiel:
@ViralBShah Thank you !

PR #9641, opened 2024-10-18
[libblastrampoline] Separate LBT builds to coincide with Julia minor versions and build freebsd aarch64 for older versions

ViralBShah:
Try to build LBT 5.4 with freebsd aarch64 support as a result.

cc @ararslan @giordano


giordano (commented):
I'm ok with splitting by version, but I'd prefer reducing duplication by having a `common.jl` script which the other packages include.

L/libblastrampoline/libblastrampoline@5.11/build_tarballs.jl
```diff
@@ -0,0 +1,15 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+using BinaryBuilderBase: sanitize
+
```
giordano, line 5: I think we can remove these lines, as they are already in the common file
```suggestion
```

L/libblastrampoline/libblastrampoline@5.4/build_tarballs.jl
```diff
@@ -0,0 +1,15 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+using BinaryBuilderBase: sanitize
+
```
giordano, line 5:
```suggestion
```

giordano (approved):

L/libblastrampoline/common.jl
```diff
@@ -7,11 +7,18 @@ name = "libblastrampoline"
 version = v"5.11.1"
```
giordano, line 7: Uhm, this is wrong

ararslan:
Thanks Viral!

PR #9644, opened 2024-10-18
[CI] Reduce parallelism for all packages with large number of artifacts

giordano:
I haven't tested this, but it sounds like it should be a better approach than manually listing the packages one-by-one (and discover this is needed only after you merge the PR).

I can't sign the new pipeline right now, so this is going to fail as is.


giordano:
Let's see how this goes.

giordano:
[It worked](https://buildkite.com/julialang/yggdrasil/builds/14082#0192a545-d2b3-4850-83df-138f3e7f2d71/568-726):
```
┌ Info: Reducing ghr concurrency
│   NAME = "Enzyme"
│   NUM_PLATFORMS = 152
└   concurrency = 4
```
and in the [environment tab of the register job](https://buildkite.com/julialang/yggdrasil/builds/14082#0192a546-52ce-4ea9-9f72-7a78b7f5f93b):
```
BINARYBUILDER_GHR_CONCURRENCY="4"
```
:partying_face:

PR #9647, opened 2024-10-19
[Dex] Update Dex to v2.41.1

tanmaykm:
Updates Dex to v2.41.1.

Also updates the patch made earlier to oidc plugin, to remove parts of the changes that got upstreamed via https://github.com/dexidp/dex/pull/3074.

Retains (with updates to match source), the other patches made earlier to support additional scopes for github adn gitlab logins and to enable multiple sessions.


giordano:
> Retains (with updates to match source), the other patches made earlier to support additional scopes for github adn gitlab logins and to enable multiple sessions.

Any chance to upstream them?

tanmaykm:
We are trying. Hoping that these or some variations of these will get accepted. But it's not there yet.

PR #9661, opened 2024-10-21
[FFLAS_FFPACK] New recipe

laurentbartholdi:
This is a recipe for the FFLAS_FFPACK - Finite Field Linear Algebra Subroutines package.

It is based on Givaro (see Givaro_jll), and is part of the LinBox series of libraries. It is purely C++.


laurentbartholdi:
Hi @giordano , this is a dependency that I created out of my original attempt to have a binary package for the library Spasm. In particular, I'm working on a recipe for Spasm with a Dependency("FFLAS_FFPACK_jll"; compat="2.5.0") in it. For now, of course, this recipe fails because FFLAS_FFPACK is not yet merged; however, is there any way I can test my new recipe, to make sure that I cooked FFLAS_FFPACK correctly?

F/FFLAS_FFPACK/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+install_license COPYING
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(exclude=Sys.iswindows) |> expand_cxxstring_abis
+
+# Strangely, libblastrampoline does not seem to be there on aarch64-unknown-freebsd
+filter!(p->!(os(p)=="freebsd" && arch(p)=="aarch64"), platforms)
```
giordano, line 36: Should be fixed by #9671

F/FFLAS_FFPACK/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+make -j ${nproc}
+make install
+
+install_license COPYING
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(exclude=Sys.iswindows) |> expand_cxxstring_abis
```
giordano, line 33: What's the issue with Windows?

F/FFLAS_FFPACK/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+
+# Strangely, libblastrampoline does not seem to be there on aarch64-unknown-freebsd
+filter!(p->!(os(p)=="freebsd" && arch(p)=="aarch64"), platforms)
+
+# The products that we will ensure are always built
+products = [
+    FileProduct("include/fflas-ffpack/fflas-ffpack-config.h", :fflas_ffpack_config_h)
+]
```
giordano, line 41: Is the product only a header file?

F/FFLAS_FFPACK/bundled/patches/fflas-ffpack-config-without-run.patch
giordano, line 1: What's this large patch for?

F/FFLAS_FFPACK/build_tarballs.jl
laurentbartholdi: There's no Givaro_jll for Windows, it was too much of a hassle to get it to compile.

F/FFLAS_FFPACK/build_tarballs.jl
laurentbartholdi: No, this is almost-pure C++ library, which produces lots of files, all in include/ except for a bin/fflas-ffpack-config. Should I include more?

F/FFLAS_FFPACK/bundled/patches/fflas-ffpack-config-without-run.patch
laurentbartholdi: It patches lots of the macros/*.m4 files to remove calls to AC_TRY_RUN which are no-no when cross-compiling. I put them all together since they really amount to zapping some useless lines

laurentbartholdi:
Dear @giordano , I believe I addressed all the points you raised. Is the correct procedure to wait for the formula to be merged, and hope that it works as expected for Spasm?

F/FFLAS_FFPACK/bundled/patches/fflas-ffpack-config-without-run.patch
giordano: I think patching the config.ac file is much better, as the diff is much smaller and easier to parse and maintain in the future.  Patching configure directly is always a bad idea, since it's an autogenerated file.

giordano: Ah, sorry, this is m4 files.  Nevermind then.

F/FFLAS_FFPACK/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+
+make -j ${nproc}
+make install
+
+install_license COPYING
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(exclude=Sys.iswindows) |> expand_cxxstring_abis
```
giordano, line 33: Why `expand_cxxstring_abis`?

F/FFLAS_FFPACK/build_tarballs.jl
giordano: After opening the archive and not seeing any binary product I believe the correct answer should have been: yes, this is a header-only package.

F/FFLAS_FFPACK/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("fflas-ffpack-config", :fflas_ffpack_config)
```
giordano, line 37: This is fairly useless, it's just the config script (which in any case shouldn't be an `ExecutableProduct` since it isn't a binary executable).
```suggestion
```

F/FFLAS_FFPACK/build_tarballs.jl
```diff
@@ -30,10 +30,11 @@ install_license COPYING
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(exclude=Sys.iswindows)
+platforms = [AnyPlatform()]
```
giordano, line 33: I don't think this is correct, apparently there were some diffences between 32-bit and 64-bit platforms

PR #9662, opened 2024-10-21
MAGEMin v1.5.6

NicolasRiel:
- corrected mass conservation issue when buffer is active
- corrected outputing of bulk_wt that was introduced in version 1.5.5


NicolasRiel:
Thanks @giordano !

PR #9667, opened 2024-10-23
[libjulia] enable FreeBSD aarch64

fingolfin:
This is mainly an experiment to see what is missing. I am also not sure which version of Julia is supposed to be the first with support for FreeBSD aarch64

### DIFF
--- a/L/libjulia/build_tarballs.jl
+++ b/L/libjulia/build_tarballs.jl
@@ -31,7 +31,7 @@ function Pkg.Types.is_stdlib(uuid::Base.UUID, julia_version::VersionNumber)
     return false
 end
 
-jllversion=v"1.10.11"
+jllversion=v"1.10.12"
 for ver in julia_full_versions
     build_julia(ARGS, ver; jllversion)
 end
--- a/L/libjulia/common.jl
+++ b/L/libjulia/common.jl
@@ -30,10 +30,10 @@ function libjulia_platforms(julia_version)
         filter!(p -> arch(p) != "armv7l", platforms)
     end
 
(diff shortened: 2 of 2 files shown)


giordano:
We don't have all deps for this platform.

fingolfin:
sure, but I was wondering what exactly is not there. For 1.12, the "does not have a mapping" warning shows for "only" three JLLs, which is better than I expected:
- `Objconv_jll`
- `LibUnwind_jll`  (actually there but I failed to update it from 1.8.1+1 to 1.8.1+2, will fix)
- `LibCURL_jll`

For 1.11, also `LLVM_full_jll` so I guess that answers the question "which is the first which *may* have supported" by "1.12".

fingolfin:
Huh, so it built despite still two missing JLLs (the build took place before #9668 was merged). So are Objconv_jll and LibCURL_jll not actually needed for the build then? Perhaps they are runtime-only deps?

giordano:
> Perhaps they are runtime-only deps?

Yeah, it's possible libraries like libcurl are dlopened at runtime and aren't needed as load-time dynamic dependencies

fingolfin:
That makes me wonder if we should just remove them from the recipe. Ah well, for now they don't hurt.

So now I wonder: should we merge this PR (modulo removing the hacks in it of course), i.e., merge the FreeBSD aarch64 support for libjulia >= 1.12? @ararslan may have a thought on that as well

ararslan:
That works for me. The only dependencies that still need to be compiled from source when building Julia are those that currently link to MbedTLS in Julia but OpenSSL in Yggdrasil, so we're close (maybe?) on that front as well.

PR #9673, opened 2024-10-24
[libcellml] Update to version 0.6.0

hsorby:


fingolfin:
Build fails because new libxml2 unfortunately has a breaking change in the callback API. To fix it you need to apply a patch to `src/xmldoc.cpp`  in libcellml, changing `void structuredErrorCallback(void *userData, xmlErrorPtr error)` to `void structuredErrorCallback(void *userData, const xmlError *error)`

hsorby:
Oh, great thanks.  I'll look into how this works with our other testing.

hsorby:
Can I get some help in understanding why CMake cannot find `LibXml2` on `aarch64-unknown-freebsd` please.
Thank you.

giordano:
> Can I get some help in understanding why CMake cannot find LibXml2 on aarch64-unknown-freebsd please.
> Thank you.

That's fixed now.  Is this good to go now?

hsorby:
Sure.

I am still confused as to how setting the LibXml2 compat option fixed finding the LibXml2 library on that one platform.  Every other platform was able to find the LibXml2 libraries.  The compilation problem stemming from the change to the function arguments appears much later than finding LibXml2 so that issue wasn't in play.

giordano:
> I am still confused as to how setting the LibXml2 compat option fixed finding the LibXml2 library on that one platform. Every other platform was able to find the LibXml2 libraries. The compilation problem stemming from the change to the function arguments appears much later than finding LibXml2 so that issue wasn't in play.

There were two separate issues:

* apparently there was a breaking change in xml2, I added a compat specification to make sure we only use a (hopefully) compatible version of xml2 together with libcellml
* aarch64-freebsd is a platform we introduced only recently, and many packages haven't been built for it yet, including xml2 until this morning I opened #9685 (and #9684 for a dependency of xml2 itself).  By upgrading to xml2 v2.3.14 we also got a build of the package for the new platform

hsorby:
Great. Thanks for the explanation.

PR #9674, opened 2024-10-24
[HSL] Upgrade to v4.0.1

amontoison:
The dummy library `libhsl.$dlext` now includes the symbols of HSL solvers needed by Uno. 
-> https://github.com/ralna/libHSL/pull/11

cc @odow

### DIFF
--- a/H/HSL/build_tarballs.jl
+++ b/H/HSL/build_tarballs.jl
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "HSL"
-version = v"4.0.0"
+version = v"4.0.1"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


odow:
So we could now build Uno_jll with HSL_jll, and then hot-swap in the real MA57?

amontoison:
Yes, you're right. I prepared a PR for `Uno_jll.jl`, but I’m waiting for the general registry to be updated.
I quickly discussed with Charlie, and he plans to support MA27 in the future, so I added the symbols of both solvers (`MA27` and `MA57`) in the dummy library.

amontoison:
-> https://github.com/JuliaPackaging/Yggdrasil/compare/master...amontoison:Yggdrasil:uno_hsl

PR #9675, opened 2024-10-24
Compile Uno with HSL_jll.jl

amontoison:
I need to bump the version number because we add a new dependency (`HSL_jll.jl`).
cc @odow @cvanaret

Related PR #9674

### DIFF
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -4,12 +4,12 @@ using BinaryBuilder, Pkg
 
 name = "Uno"
 
-version = v"1.1.0"
+version = v"1.1.1"
 
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
-        "d8571b6f8d68808707e71a361109b8713f300b4c",
+        "df18261f32991cd74a2d6f49b975145591ce38d3",
     ),
 ]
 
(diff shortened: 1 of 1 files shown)


U/Uno/build_tarballs.jl
```diff
@@ -55,22 +56,25 @@ make -j${nproc}
...
 products = [
-    # This LibraryProduct may be useful once Uno provides a C API. We omit it for nnow.
+    # This LibraryProduct may be useful once Uno provides a C API. We omit it for now.
     # LibraryProduct("libuno", :libuno),
     # We call this amplexe to match the convention of other JLL packages (like Ipopt_jll) that provide AMPL wrappers
     ExecutableProduct("uno_ampl", :amplexe),
 ]
 
 dependencies = [
+    Dependency(PackageSpec(name="HSL_jll", uuid="017b0a0e-03f4-516a-9b91-836bbd1904dd")),
```
odow, line 77: We really need compat bounds one stuff like this

amontoison: We can't do that on `HSL_jll.jl`. We don't have the official release number in the Julia registry.

odow: I don't know the right answer for this particular case. But I also mean it as a general comment for all JLL builds.

amontoison: The issue is how the upstream packages are versioned.
Sometimes they break things with a patch release, and in other cases, nothing is broken with a major one.

For the COIN-OR packages, everything is doomed by the developers, and the only solution is to use strict equality in the compat bounds to ensure that a new release of a dependency doesn't break everything. But it's not viable for all JLL, because we can't upgrade one package without recompiling all packages that depend on it if we want to use it.

odow: > But it's not viable for all JLL, because we can't upgrade one package without recompiling all packages that depend on it if we want to use it.

I kinda think this is what we should do

PR #9689, opened 2024-10-25
[Uno] bump gitsha without changing version

odow:
cc @amontoison


amontoison:
@odow You also need to use -DHSL and -DAMPLSOLVER now instead of -Dma57 and -Damplsolver.

amontoison (approved):

PR #9690, opened 2024-10-25
[sshpass] Add recipe

JamesWrigley:
Some oddities:
- This required a configure.ac patch to compile on a lot of platforms.
- I think it needs to be built under cygwin to work on windows, so windows is disabled.

### DIFF
--- a/S/sshpass/build_tarballs.jl
+++ b/S/sshpass/build_tarballs.jl
@@ -20,6 +20,7 @@ cd sshpass-*
 # ARM/OSX/musl: https://github.com/maxmind/libmaxminddb/pull/152
 # Submitted upstream: https://sourceforge.net/p/sshpass/patches/16/
 atomic_patch -p1 ../patches/configure-ac.patch
+autoreconf -i
 
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
 make -j${nproc}
@@ -27,7 +28,7 @@ make install
 """
 
 # sshpass support on Windows requires Cygwin so we don't support it
-platforms = filter(p -> !Sys.iswindows(p), supported_platforms())
+platforms = filter(!Sys.iswindows, supported_platforms())
 
 # The products that we will ensure are always built
 products = [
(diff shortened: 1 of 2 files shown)


S/sshpass/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+# Submitted upstream: https://sourceforge.net/p/sshpass/patches/16/
+atomic_patch -p1 ../patches/configure-ac.patch
+
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+make -j${nproc}
+make install
+"""
+
+# sshpass support on Windows requires Cygwin so we don't support it
+platforms = filter(p -> !Sys.iswindows(p), supported_platforms())
```
giordano, line 30:
```suggestion
platforms = filter(!Sys.iswindows, supported_platforms())
```

S/sshpass/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd sshpass-*
+
+# The default configure.ac includes a malloc check that causes failures on
+# ARM/OSX/musl: https://github.com/maxmind/libmaxminddb/pull/152
+# Submitted upstream: https://sourceforge.net/p/sshpass/patches/16/
+atomic_patch -p1 ../patches/configure-ac.patch
+
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
```
giordano, line 24: I'm confused: you patch configure.ac and then immediately run configure without regenerating it? I'm missing how this works.

S/sshpass/build_tarballs.jl
JamesWrigley: Fixed in 0c8223c8.

S/sshpass/build_tarballs.jl
JamesWrigley: Right I see what was happening, `make` saw the changed `configure.ac` file and was regenerating `configure` and running it again: https://buildkite.com/julialang/yggdrasil/builds/14255#0192c5d5-a395-4527-974b-c31ea530086c/660-851

I added a call to `autoreconf -i` in 0c8223c8 so that should be fixed now.

PR #9692, opened 2024-10-26
[ADOLC] Update to 2.0.0

TimSiebert1:
New version
- is based on ADOL-C's C Interface
- remove CxxWrap dependency


giordano:
The change looks overall ok to me, but just to understand: this doesn't build the Julia wrapper with libcxxwrap anymore?

TimSiebert1:
Hi @giordano 

yes right. I have built a C Interface for ADOL-C that now allows me to use the built-in Julia tools to call C code. The CxxWrap dependency has been removed.

PR #9696, opened 2024-10-27
[ldid] Build against OpenSSL v3

staticfloat:


L/ldid/build_tarballs.jl
```diff
@@ -30,7 +30,7 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("libplist_jll"),
-    Dependency("OpenSSL_jll"; compat="1.1.13"),
+    Dependency("OpenSSL_jll"; compat="3"),
```
giordano, line 33:
```suggestion
    Dependency("OpenSSL_jll"; compat="3.0.8"),
```

L/ldid/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder
 
 name = "ldid"
-version = v"2.1.3" # <-- Fake version to build for new platforms
+version = v"2.1.4" # <-- Fake version to build with new OpenSSL compat bounds
```
giordano, line 6: The upstream repo has newer versions: https://git.saurik.com/ldid.git. BTW, why do we fork it?

staticfloat: It’s actually forked from https://github.com/tpoechtrager/ldid which made a bunch of build system changes

L/ldid/build_tarballs.jl
```diff
@@ -30,7 +30,7 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("libplist_jll"),
-    Dependency("OpenSSL_jll"; compat="1.1.13"),
+    Dependency("OpenSSL_jll"; compat="3.0.8"),
```
giordano, line 33: Ok, we need a newer version for aarch64-freebsd:
```suggestion
    Dependency("OpenSSL_jll"; compat="3.0.15"),
```

L/ldid/build_tarballs.jl
staticfloat: To actually answer your question, the `tpoechtrager` fork doesn't have any new commits, and glancing over the changes made in Saurik's source, I'm not sure it's worth my time to try and integrate them.  They don't seem that applicable to us.

PR #9703, opened 2024-10-29
[libcellml] Update to version 0.6.2

hsorby:


L/libcellml/build_tarballs.jl
```diff
@@ -49,7 +49,7 @@ products = [
 dependencies = [
     # XML2 apparently had a breaking change, so it's important to specify the compat bound:
     # https://github.com/JuliaPackaging/Yggdrasil/pull/9673#issuecomment-2434514026
-    Dependency("XML2_jll"; compat="2.13.4"),
+    Dependency("XML2_jll"),
```
giordano, line 52: I presume the compat bound should remain, no?  You still need to have at runtime a compatible build of xml2 than what was used at compile-time, right?  The compat here is to ensure this.

hsorby: No, I have fixed up the code to deal with this situation now.

hsorby: Well probably not actually is this a command for other package linking? Not a fixing of the build environment?

hsorby: If this compatibility is for the build environment the code has been fixed to deal with this. If this compatibility is required for package version management it will be needed.  I didn't want to fix the version for the build environment which is what I thought this part of the code was doing.

giordano: This is both for build and runtime basically: at build time we use the lowest version in the allowed range, at runtime we can use any version in the allowed range (by default the package manager would prefer the latest version, but it needs to check if it's compatible with other packages in the environment)

hsorby: So this is a minimum setting, not pinning the version to this specific version?

giordano: Yes, this is the syntax of the compat bounds: https://pkgdocs.julialang.org/v1/compatibility/#Version-specifier-format.  I believe it's similar to npm syntax.  Pinning to a single version would be with a leading `=`.

hsorby: Okay, thanks so much.

PR #9704, opened 2024-10-29
[PETSc] version 3.22.1

boriskaus:
This compiles PETSc version 3.22.1. It is also tries to do away with applying the `mpi-constants` patch.


ViralBShah:
#12339 provides 3.24.0

PR #9708, opened 2024-10-29
[FFLAS_FFPACK] fix a breaking typo in a header file

laurentbartholdi:
There is a silly typo in a header file of the blas interface library (they wrote ssryk_ instead of ssyrk_).
I should, and will, upstream it, but meanwhile here is a simple fix.


laurentbartholdi:
Uh, @giordano , sorry for my silly questions, but should I submit a pull request relative to the previous FFLAS_FFPACK (which you merged, thank you!), or as here continuing on my previous branch? The point is that the codespaces environment doesn't have the most recent binarybuilder data.

laurentbartholdi:
OK, never mind, I found out that the merge anyways was forbidden, so I'll submit a new PR wrt the head.

PR #9709, opened 2024-10-29
[FFLAS_FFPACK] Patched package's config-blas.h to fix silly typo

laurentbartholdi:
This fixes a silly typo in a header file of the FFLAS_FFPACK library (BLAS call ssryk_ instead of ssyrk_). I should and will upstream it, but meanwhile this fixes compilation problems for libraries using FFLAS_FFPACK_jll


laurentbartholdi:
Uh, no need to patch upstream, it's corrected in the master branch on github, just hasn't been released.

PR #9712, opened 2024-10-30
[Spasm] New recipe

laurentbartholdi:
This is a fresh start at a recipe for the SpaSM direct linear algebra solver modulo p, by Charles Bouillaguet (see https://github.com/cbouilla/spasm).

### DIFF
--- a/S/Spasm/build_tarballs.jl
+++ b/S/Spasm/build_tarballs.jl
@@ -28,7 +28,7 @@ for t in bitmap check_cert dm echelonize kernel rank solve stack transpose verti
 done
 
 cp build/src/libspasm* $prefix/lib/
-cp src/spasm.h $prefix/include/
+install -Dvm 644 "src/spasm.h" "${includedir}/spasm.h"
 
 install_license ${WORKSPACE}/srcdir/spasm/COPYING
 """
--- a/S/Spasm/build_tarballs.jl
+++ b/S/Spasm/build_tarballs.jl
@@ -63,8 +63,10 @@ dependencies = [
     Dependency("argp_standalone_jll"),
     Dependency("libblastrampoline_jll"; compat="5.4.0"),
     Dependency("GMP_jll"; compat="6.2.1"),
-    Dependency("LLVMOpenMP_jll"),
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+            # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
+            # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
+            Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
+            Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/S/Spasm/build_tarballs.jl
+++ b/S/Spasm/build_tarballs.jl
@@ -1,0 +1,70 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Spasm"
+version = v"1.3.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/laurentbartholdi/spasm.git", "072719a40c837e447dfe4ae9e4941c60d9a28eda"),
+    DirectorySource("./bundled")
+]
+
+# Bash recipe for building across all platforms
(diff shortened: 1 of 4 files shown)


S/Spasm/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+cmake --build build
+cmake --install build
+
+for t in bitmap check_cert dm echelonize kernel rank solve stack transpose vertical_swap; do
+    cp build/tools/$t $prefix/bin/
+done
+
+cp build/src/libspasm* $prefix/lib/
```
imciner2, line 30: Is this just copying one library called `libspasm` (with the appropriate extension) into the directory? If so, our standard way of doing that is the command
```
install -Dvm 755 "build/src/libspasm.${dlext}" "${libdir}/libspasm.${dlext}"
```

laurentbartholdi: OK, done. The build will fail on some architectures because it depends on a PR for FFLAS_FFPACK

S/Spasm/build_tarballs.jl
```diff
@@ -0,0 +1,72 @@
...
+
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+
+cmake -B build \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release
+cmake --build build
```
imciner2, line 26:
```suggestion
cmake --build build --parallel ${nproc}
```

S/Spasm/build_tarballs.jl
```diff
@@ -0,0 +1,72 @@
...
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    Dependency("Givaro_jll"; compat="4.2.0"),
+    Dependency("FFLAS_FFPACK_jll"; compat="2.5.0"),
+    Dependency("argp_standalone_jll"),
+    Dependency("libblastrampoline_jll"; compat="5.4.0"),
+    Dependency("GMP_jll"; compat="6.2.1"),
+    Dependency("LLVMOpenMP_jll"),
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
```
imciner2, line 67:
```suggestion
            # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
            # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
            Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
            Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
```
Are you using these to get openMP support on all platforms? If so, we need to change these to only add one openMP library on the platforms (we only want the LLVM OpenMP library on BSD systems, not the other systems).

laurentbartholdi: OK, thanks!
(but it did build correctly the tarballs before... what was wrong with the previous?)

imciner2: It will build with it, but it might not get the right OpenMP on the platforms, and that could cause problems then if we try to load two different OpenMP libraries at the same time.

imciner2:
This looks good now. Thanks for your work on it.

S/Spasm/build_tarballs.jl
```diff
@@ -0,0 +1,73 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Spasm"
+version = v"1.3.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/laurentbartholdi/spasm.git", "072719a40c837e447dfe4ae9e4941c60d9a28eda"),
```
stemann, line 10: Why was a fork used?

S/Spasm/build_tarballs.jl
```diff
@@ -0,0 +1,73 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Spasm"
```
stemann, line 5: Why name it "Spasm" and not "SpaSM" as the source specifies, https://github.com/cbouilla/spasm ?

S/Spasm/build_tarballs.jl
laurentbartholdi: Some local changes, which I'll push in a new version when the author has integrated the changes. Thanks for reminding me!

S/Spasm/build_tarballs.jl
laurentbartholdi: Good suggestion, will do.

PR #9713, opened 2024-10-30
[FFLAS_FFPACK] Fixed return types for cblas_dsyrk and cblas_ssyrk

laurentbartholdi:
Another small error in the original library, which breaks some compilers.


F/FFLAS_FFPACK/bundled/patches/fflas-ffpack-typo-in-config-blas.patch
giordano, line 1: Should this patch be upstreamed?

laurentbartholdi: It's already fixed on their master.

PR #9715, opened 2024-10-30
[argp_standalone] A fresh attempt at having it compile for macos and freebsd

laurentbartholdi:
The current recipe is broken (the licence sha changed) and it doesn't compile for macos and freebsd, which are precisely the platforms on which argp is not bundled in libc.


laurentbartholdi:
Somehow .ci/{Manifest,Project}.toml were included in the PR. I reset them.

A/argp_standalone/bundled/patches/remove-throw.patch
giordano, line 1: Should this patch be upstreamed?

laurentbartholdi: I don't think so. this is 2004 code, so the project should be considered frozen. Almost all compilers accept the original code, only a few request __THROW to be only in the declaration. There's a "1.4" version available on the same website, but it's also from 2004, and marked as "test" so I'd rather not go there.

There are lots of other argp_standalone implementations, https://github.com/ericonr/argp-standalone or https://github.com/tom42/argp-standalone which are more actively maintained; I made the minimal amount of changes that make the recipe compile.

PR #9717, opened 2024-10-31
[libjulia] update julia 1.12.0-DEV version

lgoettgens:
such that https://github.com/JuliaLang/julia/pull/56373 is included

cc @fingolfin


fingolfin (approved):
Thanks!

PR #9718, opened 2024-10-31
T8code 3.0.0

jmark:
This PR upgrades `t8code_jll` to the most recent version of t8code: 3.0.0.

Additionally, the build recipe switched from autotools to CMake for configuring and building t8code.


jmark:
All green! Please merge if you agree.

T/t8code/bundled/patches/p4est.patch
giordano, line 1: Should this patch be upstreamed?

jmark: I opened issues upstream.
https://github.com/cburstedde/libsc/issues/206
https://github.com/cburstedde/p4est/issues/322

I am not sure what exactly the problem is. The patches are results via trial-and-error and specifically tailored for getting t8code compiled on MinGW in the Yggdrasil resp. BinaryBuilder context.

From past experience, merging patches to the aforementioned libraries takes a very long time.

I appreciate if this workaround is accepted for now, so I can move on with providing a much needed update to T8code.jl.

PR #9722, opened 2024-11-01
[Uno] update to v1.2.0

cvanaret:
cc @odow

### DIFF
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -4,12 +4,12 @@ using BinaryBuilder, Pkg
 
 name = "Uno"
 
-version = v"1.1.2"
+version = v"1.2.0"
 
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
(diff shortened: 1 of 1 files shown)


imciner2:
In general, you should do a version change if you change the gitsha - especially in this case since it is changing some actual code inside the solver.

(while we do technically support different builds of the same version, it isn't really supported by the Pkg system, so it is generally not suggested to use them)

cvanaret:
Got it! So v"1.1.2" would be ok?

odow (changes requested):
I was only bumping the gitsha without the version change because previous versions of Uno did not work, and I want to avoid the Yggdrasil version and the upstream version getting out of sync.

@cvanaret perhaps it's time for a new v1.2.0 upstream release?

cvanaret:
I learn along the way :)
[Uno 1.2.0](https://github.com/cvanaret/Uno/releases/tag/v1.2.0) was released and the build_tarballs.jl was updated accordingly.

odow (approved):

PR #9737, opened 2024-11-05
Add special case for windows

KronosTheLate:
Attempting to fix [this issue](https://github.com/duckdb/duckdb/issues/13911) by adding a special case for windows. Tried to make that special case a mixture of the generic build script, and the contents [here](https://duckdb.org/docs/dev/building/build_instructions.html#windows)


D/DuckDB/build_tarballs.jl
```diff
@@ -10,8 +10,7 @@ sources = [
     GitSource("https://github.com/duckdb/duckdb.git", "f680b7d08f56183391b581077d4baf589e1cc8bd"),
 ]
 
-# Bash recipe for building across all platforms
-script = raw"""
+script_nonwindows = raw"""
```
giordano, line 13: Why not using conditionals inside the script? Having different scripts for different platforms is a needless complication.

KronosTheLate: I figured that the cases inside the script would have be decided based on the target, which seemed to be defined to this julia array:
```julia
julia> using BinaryBuilder; expand_cxxstring_abis(supported_platforms())
┌ Warning: Running Registrator on a Julia version that does not support weak dependencies. Weak dependencies will not be registered.
└ @ RegistryTools C:\Users\dennis.bal\.julia\packages\RegistryTools\tHkoa\src\RegistryTools.jl:18
29-element Vector{Platform}:
 Linux i686 {cxxstring_abi=cxx03, libc=glibc}
 Linux i686 {cxxstring_abi=cxx11, libc=glibc}
 Linux x86_64 {cxxstring_abi=cxx03, libc=glibc}
 Linux x86_64 {cxxstring_abi=cxx11, libc=glibc}
 Linux aarch64 {cxxstring_abi=cxx03, libc=glibc}
 Linux aarch64 {cxxstring_abi=cxx11, libc=glibc}
 Linux armv6l {call_abi=eabihf, cxxstring_abi=cxx03, libc=glibc}
 Linux armv6l {call_abi=eabihf, cxxstring_abi=cxx11, libc=glibc}
 Linux armv7l {call_abi=eabihf, cxxstring_abi=cxx03, libc=glibc}
 Linux armv7l {call_abi=eabihf, cxxstring_abi=cxx11, libc=glibc}
 Linux powerpc64le {cxxstring_abi=cxx03, libc=glibc}
 Linux powerpc64le {cxxstring_abi=cxx11, libc=glibc}
 Linux i686 {cxxstring_abi=cxx03, libc=musl}
 Linux i686 {cxxstring_abi=cxx11, libc=musl}
 Linux x86_64 {cxxstring_abi=cxx03, libc=musl}
 Linux x86_64 {cxxstring_abi=cxx11, libc=musl}
 Linux aarch64 {cxxstring_abi=cxx03, libc=musl}
 Linux aarch64 {cxxstring_abi=cxx11, libc=musl}
 Linux armv6l {call_abi=eabihf, cxxstring_abi=cxx03, libc=musl}
 Linux armv6l {call_abi=eabihf, cxxstring_abi=cxx11, libc=musl}
 Linux armv7l {call_abi=eabihf, cxxstring_abi=cxx03, libc=musl}
 Linux armv7l {call_abi=eabihf, cxxstring_abi=cxx11, libc=musl}
 macOS x86_64
 macOS aarch64
 FreeBSD x86_64
 Windows i686 {cxxstring_abi=cxx03}
 Windows i686 {cxxstring_abi=cxx11}
 Windows x86_64 {cxxstring_abi=cxx03}
 Windows x86_64 {cxxstring_abi=cxx11}
 ```
 
 Where there are e.g. multiple names for windows that have to be concidered. Furthermore, it is unclear how those `Platform` objects are passed to bash.
 
 I figured it was actually extremely neat and tidy to move that logic into julia, using clearly named and easy-to-use function is Sys.iswindows, instead of adding string matching for multiple cases, which to me felt hacky. So we seem to have different intuitions from looking at the same stuff.

That said, you are the desciding vote, and I am happy to implement it whatever way you deem best. I am however possibly not capable of doing it the way you want, so you might have to make the change yourself.

**Comment by KronosTheLate** [comment]
The current error from the build is [here](https://buildkite.com/julialang/yggdrasil/builds/14465#0192fb85-098e-4daa-ac83-df3081af0443/6-2074), and reads
```bash
[09:13:37] ../../third_party/mbedtls/libduckdb_mbedtls.a(mbedtls_wrapper.cpp.o): In function `duckdb_mbedtls::MbedTlsWrapper::AESGCMStateMBEDTLS::~AESGCMStateMBEDTLS()':
[09:13:37] mbedtls_wrapper.cpp:(.text+0x4c): undefined reference to `duckdb::EncryptionState::~EncryptionState()'
[09:13:37] ../../third_party/mbedtls/libduckdb_mbedtls.a(mbedtls_wrapper.cpp.o): In function `duckdb_mbedtls::MbedTlsWrapper::AESGCMStateMBEDTLS::~AESGCMStateMBEDTLS()':
[09:13:37] mbedtls_wrapper.cpp:(.text+0x3f4): undefined reference to `duckdb::EncryptionState::~EncryptionState()'
[09:13:37] ../../third_party/mbedtls/libduckdb_mbedtls.a(mbedtls_wrapper.cpp.o): In function `duckdb_mbedtls::MbedTlsWrapper::AESGCMStateMBEDTLS::AESGCMStateMBEDTLS()':
[09:13:37] mbedtls_wrapper.cpp:(.text+0xd80): undefined reference to `duckdb::EncryptionState::EncryptionState()'
[09:13:37] mbedtls_wrapper.cpp:(.text+0xdcc): undefined reference to `duckdb::EncryptionState::~EncryptionState()'
[09:13:37] ../../third_party/mbedtls/libduckdb_mbedtls.a(mbedtls_wrapper.cpp.o):(.data.rel.ro._ZTIN14duckdb_mbedtls14MbedTlsWrapper18AESGCMStateMBEDTLSE[_ZTIN14duckdb_mbedtls14MbedTlsWrapper18AESGCMStateMBEDTLSE]+0x10): undefined reference to `typeinfo for duckdb::EncryptionState'
```

Since the [official build instructions](https://duckdb.org/docs/dev/building/build_instructions.html#windows):
```bash
python scripts/windows_ci.py
        cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_GENERATOR_PLATFORM=x64 -DENABLE_EXTENSION_AUTOLOADING=1 -DENABLE_EXTENSION_AUTOINSTALL=1 -DDUCKDB_EXTENSION_CONFIGS="${GITHUB_WORKSPACE}/.github/config/bundled_extensions.cmake" -DDISABLE_UNITY=1 -DOVERRIDE_GIT_DESCRIBE="$OVERRIDE_GIT_DESCRIBE"
        cmake --build . --config Release --parallel
```
must work, something in my merging of the existing cmake command and that one has gone wrong. My version looks like this:
```bash
cd $WORKSPACE/srcdir/duckdb/

python scripts/windows_ci.py

cmake -B build \
      -DCMAKE_BUILD_TYPE=Release \
      -DENABLE_SANITIZER=FALSE \
      -DBUILD_EXTENSIONS='autocomplete;icu;parquet;json;fts;tpcds;tpch' \
      -DENABLE_EXTENSION_AUTOLOADING=1 \
      -DENABLE_EXTENSION_AUTOINSTALL=1 \
      -DBUILD_UNITTESTS=FALSE \
      -DBUILD_SHELL=TRUE \
      -DDUCKDB_EXPLICIT_PLATFORM="${target}" \
      -DUCKDB_EXTENSION_CONFIGS="$WORKSPACE/srcdir/duckdb/.github/config/bundled_extensions.cmake" \
      -DDISABLE_UNITY=1
cmake --install build

install -Dvm 755 "build/src/libduckdb.${dlext}" -t "${libdir}"
```

We are already past my paygrade, so help is appreciated.

D/DuckDB/build_tarballs.jl
stemann: @KronosTheLate The platform being built is available in `$target` (and in `$bb_full_target`).

stemann:
@KronosTheLate You should (very) likely use the MinGW64 instructions and not the MSVC instructions: https://duckdb.org/docs/dev/building/build_instructions.html#msys2-and-mingw64

BinaryBuilder/Yggdrasil is using GCC, not MSVC.

D/DuckDB/build_tarballs.jl
giordano: This is also explained in the docs:
* https://docs.binarybuilder.org/stable/build_tips/#Initiating-different-shell-commands-based-on-target
* https://docs.binarybuilder.org/stable/building/#Platforms

KronosTheLate:
I am getting a build error, which appears related to not having the right dependencies. Microsoft copilot suggests fixing this in the workflow:
image

I need some help :/

stemann:
You should not copy/paste the DuckDB instructions verbatim.

You should not install stuff with pacman - JLL dependencies is the way to get dependencies.

From the DuckDB documentation you should only take away the cmake args:
```sh
cmake -G "Ninja" -DCMAKE_BUILD_TYPE=Release -DBUILD_EXTENSIONS="icu;parquet;json"
cmake --build . --config Release
```

D/DuckDB/build_tarballs.jl
giordano: > In any case, both options are far less understandable to me than what I have done.

What you've done is wrong because you'll register the package twice. Also, what you've done has duplicate code, which is error prone and less maintainable.  Please use the conditionals inside the script instead.

D/DuckDB/build_tarballs.jl
```diff
@@ -10,32 +10,48 @@ sources = [
...
+        -DCMAKE_INSTALL_PREFIX=${prefix} \
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DENABLE_SANITIZER=FALSE \
+        -DENABLE_EXTENSION_AUTOLOADING=1 \
+        -DENABLE_EXTENSION_AUTOINSTALL=1 \
+        -DBUILD_UNITTESTS=FALSE \
+        -DBUILD_SHELL=TRUE \
+        -DDUCKDB_EXPLICIT_PLATFORM="${target}" \
+        -DBUILD_EXTENSIONS="icu;parquet;json"
```
stemann, line 29: You should keep `-B build` to keep build output in `build` (and that would match you `install` call).

I would recommend that you simplify to something like this (a single `cmake` call - with sorted args):
```suggestion
if [[ "$target" == *-w64-mingw32 ]]; then
  build_extensions="icu;parquet;json"
else
  build_extensions="autocomplete;icu;parquet;json;fts;tpcds;tpch"
fi

cmake -B build \
       -DBUILD_EXTENSIONS="$build_extensions" \
       -DBUILD_SHELL=TRUE \
       -DBUILD_UNITTESTS=FALSE \
       -DCMAKE_BUILD_TYPE=Release \
       -DCMAKE_INSTALL_PREFIX=$prefix \
       -DCMAKE_TOOLCHAIN_FILE=$CMAKE_TARGET_TOOLCHAIN \
       -DDUCKDB_EXPLICIT_PLATFORM="$target" \
       -DENABLE_EXTENSION_AUTOINSTALL=1 \
       -DENABLE_EXTENSION_AUTOLOADING=1 \
       -DENABLE_SANITIZER=FALSE
 cmake --build build --parallel $nproc
 cmake --install build

if [[ "$target" == *-w64-mingw32 ]]; then
  install -Dvm 755 build/src/libduckdb.$dlext -t $libdir
fi
```

D/DuckDB/build_tarballs.jl
```diff
@@ -10,32 +10,40 @@ sources = [
...
-      -DBUILD_UNITTESTS=FALSE \
-      -DBUILD_SHELL=TRUE \
-      -DDUCKDB_EXPLICIT_PLATFORM="${target}"
-cmake --build build --parallel ${nproc}
-cmake --install build
-
-if [[ "${target}" == *-mingw32 ]]; then
-    install -Dvm 755 "build/src/libduckdb.${dlext}" -t "${libdir}"
+if [[ "$target" == *-ming* ]]; then
+  build_command="cmake -B build -G \"Ninja\""
```
giordano, line 17: Why is the build command different at all across platforms?  Either of the two should be just fine everywhere.

D/DuckDB/build_tarballs.jl
```diff
@@ -10,32 +10,40 @@ sources = [
...
+ cmake --install build
+
+if [[ "$target" == *-w64-mingw32 ]]; then
+  install -Dvm 755 "build/src/libduckdb.${dlext}" -t "${libdir}"
 fi
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = expand_cxxstring_abis(supported_platforms())
```
giordano, line 38: Why was this removed?

D/DuckDB/build_tarballs.jl
```diff
@@ -1,5 +1,5 @@
 # Note that this script can accept some limited command-line arguments, run
-# `julia build_tarballs.jl --help` to see a usage message.
+# `julia build_tarballs.jl --help` to see a usage message.
```
giordano, line 2:
```suggestion
# `julia build_tarballs.jl --help` to see a usage message.
```

D/DuckDB/build_tarballs.jl
```diff
@@ -50,6 +58,11 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = Dependency[
+    #Dependency("Mingw_jll"),
+    #Dependency("PlatformSupport-x86_64-w64-mingw32_jll"),
+    #Dependency("PlatformSupport-i686-w64-mingw32_jll"),
+    #Dependency("Ninja_jll"),
+    #Dependency("CMake_jll"),
```
giordano, line 65:
```suggestion
```

D/DuckDB/build_tarballs.jl
KronosTheLate: Just because the official build instructions for windows use `-G "Ninja`. The library failed to precompile on windows, so I was just guessing that if I use the windows-specific instructions more closely, then that might fix the issue.

As you can see by the commit-history, I am just spitballing, and have no idea what I am doing. Help appreciated.

D/DuckDB/build_tarballs.jl
KronosTheLate: Whoops - I mistakenly forgot the expand_cxxstring_abis part when undoing the separation between the windows command. Fixed in separate commit.

KronosTheLate:
It now builds without error, but that was the case before this PR. I am not sure how to test the build other than attempting to install DuckDB on a windows machine after this is merged, which I am happy to test. I am also not sure if merging this will result in DuckDB_jll.jl to use the new binaries, or if that only happens when a new version is tagged. Perhaps one of the maintainers know if this will be effective after merge?

Because the precompilation error occurs for DuckDB_jll.jl, which is built by Yggdrasil, it seems fair enough to make a test that installs the package as part of the build process. That might add a lot of overhead though, so I am not sure that it is overall worth it.

jhmenke:
Are there any updates here? The PR does not seem to be merged. At least the automatic install still fails with the same error.

giordano:
https://github.com/JuliaPackaging/Yggdrasil/pull/9775 (unaddressed review comments)

PR #9738, opened 2024-11-05
[Pythia8_cxxwrap] Regenerate with new CxxWrap

peremato:


barche:
You are hitting the second bullet here I think: https://github.com/JuliaInterop/CxxWrap.jl?tab=readme-ov-file#breaking-changes-in-v015

peremato:
> You are hitting the second bullet here I think: https://github.com/JuliaInterop/CxxWrap.jl?tab=readme-ov-file#breaking-changes-in-v015

Thanks. I forgot to delete a file from previous version

PR #9742, opened 2024-11-06
[SuiteSparse@7] Update to v7.8.3

mzy2240:
ParU 1.0.0 is bundled in this version, if anyone could shed some light on how to update the ParU_jll I might give a try


giordano:
Looks good to me as tests are passing, but I'd like some more eyes on it since this is a Julia dependency.  CC: @ViralBShah @rayegun

ViralBShah:
For ParU, just make a PR to https://github.com/JuliaPackaging/Yggdrasil/tree/master/S/SuiteSparse/ParU.

PR #9744, opened 2024-11-06
[GDAL] Bump to 3.10.0

jeremiahpslewis:


ViralBShah:
Is it worth trying FreeBSD aarch64?

ViralBShah:
Seems like it is breaking on ODBC - for which I have a PR up for unixODBC (but it may not be what is actually needed).

jeremiahpslewis:
Probably best to think about this as two separate PRs, one to bump the version, the other to add new platforms?

PR #9745, opened 2024-11-06
[ParU] Update to v1.0.0

mzy2240:


mzy2240:
@ViralBShah is it possible to approve and merge?

ericphanson:
fyi, the package is not loadable: https://github.com/JuliaRegistries/General/pull/118860

ViralBShah:
It's always challenging to pull in other SuiteSparse libraries and get the dependency ordering right.

imciner2:
> fyi, the package is not loadable: https://github.com/JuliaRegistries/General/pull/118860

I believe this is to be expected for Julia 1.11. This version of ParU requires 7.8.3, but that currently only has 7.7.0 in it. Since SuiteSparse is a stdlib, we are limited to compatibility with the version that is included in the Julia release - so this version of ParU won't install there. I actually don't think this will even install on Julia master, since that only has SuiteSparse 7.8.3, and this will have a dependency that is strict equality with SuiteSparse 7.8.3.

ViralBShah:
We can go bump master to 7.8.3.

rayegun:
I specifically wanted to avoid updating ParU yet for this reason, since the PR that works on 1.11 previously failed... I wish we could have a codeowners on this haha

I'll have to do some fiddling here, although Tim said we can probably attempt to build ParU v1 against older versions of SuiteSparse_config and the other packages

ViralBShah:
I didn't realize - will ping you on future ones!

PR #9746, opened 2024-11-06
[SCIP_PaPILO] up scip 920

matbesancon:


matbesancon:
For OSX, since the TBB version hasn't changed since the last build, no clue what is going on:
```
[16:15:25] /opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/local/include/tbb/../oneapi/tbb/detail/_task_handle.h:56:5: error: aligned deallocation function of type 'void (void *, std::align_val_t) noexcept' is only available on macOS 10.13 or newer
[16:15:25]    56 |     ~task_handle_task() override {
[16:15:25]       |     ^
[16:15:25] /opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/local/include/tbb/../oneapi/tbb/detail/_task_handle.h:56:5: note: if you supply your own aligned allocation functions, use -faligned-allocation to silence this diagnostic
[16:15:25] In file included from /workspace/srcdir/scipoptsuite-9.2.0/scip/src/lpi/lpi_spx2.cpp:98:
[16:15:25] In file included from /workspace/srcdir/scipoptsuite-9.2.0/soplex/src/soplex.h:75:
[16:15:25] In file included from /workspace/srcdir/scipoptsuite-9.2.0/soplex/src/soplex/spxpapilo.h:156:
[16:15:25] In file included from /workspace/srcdir/scipoptsuite-9.2.0/papilo/src/papilo/core/Presolve.hpp:33:
[16:15:25] In file included from /workspace/srcdir/scipoptsuite-9.2.0/papilo/src/papilo/core/PresolveMethod.hpp:27:
[16:15:25] In file included from /workspace/srcdir/scipoptsuite-9.2.0/papilo/src/papilo/core/ConstraintMatrix.hpp:33:
[16:15:25] In file included from /workspace/srcdir/scipoptsuite-9.2.0/papilo/src/papilo/core/SingleRow.hpp:27:
[16:15:25] In file included from /workspace/srcdir/scipoptsuite-9.2.0/papilo/src/papilo/core/VariableDomains.hpp:31:
[16:15:25] In file included from /workspace/srcdir/scipoptsuite-9.2.0/papilo/src/papilo/misc/tbb.hpp:57:
[16:15:25] In file included from /opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/local/include/tbb/parallel_for.h:17:
[16:15:25] /opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/local/include/tbb/../oneapi/tbb/parallel_for.h:66:8: error: aligned deallocation function of type 'void (void *, std::align_val_t) noexcept' is only available on macOS 10.13 or newer
```

imciner2:
It looks like oneTBB isn't on the aarch64 FreeBSD build yet. While I am working on an update that would include that arch hopefully (https://github.com/JuliaPackaging/Yggdrasil/pull/10082), that wouldn't be this version specified in the dependency - and if I recall, oneTBB changed their API a lot at some point after 2021.9.0, so updating to the new version might not be easy without upstream actually doing it.

To make this build, we need to filter out that platform
```
```

We probably should also filter out the RISC-V architecture for now (since again, oneTBB won't be present).
```
filter!(p -> !(Sys.islinux(p) && arch(p) == "riscv64"), platforms)
```

imciner2:
> If someone knows what can be going on with linking fortran on windows? It just seems not to recognize the fortran symbols at all:

@matbesancon it is basically just the comment from the last SCIP_PaPILO update here: https://github.com/JuliaPackaging/Yggdrasil/pull/8224#issuecomment-2034941690. Basically, it looks like we still need to include that patch to enable the Fortran language for that one executable - so it looks like upstream hasn't fixed that issue yet.

imciner2:
So, I'm not sure why TBB is now giving us this error on macOS, but one possible solution would be to just bump the macOS SDK version to be 10.13 so that we then get the required function. I don't think there would be any problems with that, @matbesancon are you fine with that?

matbesancon:
yes this would work for me :)

imciner2:
Ok, everything works. I did notice that the `make install` step seems to actually be doing quite a bit of compilation still, so it would be good to add a `-j${nproc}` to that step as well on the next update.

PR #9757, opened 2024-11-08
MAGEMin v1.5.8

NicolasRiel:
- Corrected unforeseen segfault problem affecting MAGEMin_C (julia wrapper)
- New tests have been added to MAGEMin_C to avoid that in the future


NicolasRiel:
@giordano  Thank you very much!

PR #9769, opened 2024-11-12
[OpenFHE_int128] Initializing

ArseniyKholod:
This package is a copy of OpenFHE with another building options, namely NATIVESIZE=128


ArseniyKholod:
I also disabled armv6l and armv7l architectures, as they do not support 128 bit integer

ArseniyKholod:
@sloede, could you please take a look?

O/OpenFHE_128/build_tarballs.jl
sloede, line 1: Is there any difference to the original file, except for the name, the build arguments, and the platforms? If no, maybe you could create a file, say `common.jl` in the OpenFHE directory that can be included and that contains a function such as
```julia
sources, script, platforms, products, dependencies = prepare_openfhe_build(...)
```
to avoid duplication?

Anyways, you should add a README in both directories explaining why there are two separate JLL packages for this. Furthermore, it should be clarified in each `build_tarballs.jl` that if one is updated, the other one should likely be updated as well.

sloede (changes requested):
I think we need to at least try to remove some of the redundancies here - I left two comments pertaining to that.

@giordano Alternatively, do you see a cleaner way to achieve what we want to do, i.e., having a separate 128-bit build available within a single `build_tarballs.jl` that is somehow selectable by the user in a sane way (e.g., through preferences)? Sorry to bother you again with this...

O/OpenFHE_128/bundled/patches/windows-fix-cmake-libdir.patch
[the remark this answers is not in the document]
ArseniyKholod: Yes, it is the same. I could not make it work with symlink, because the target path does not exist. So I just copied it internally on runners.

O/OpenFHE_128/build_tarballs.jl
ArseniyKholod: I created the common.jl in #9891. And also added readme and comments.

ArseniyKholod:
This PR should be merged after #9891. @sloede could you take a look?

O/OpenFHE_128/README.md
```diff
@@ -0,0 +1,8 @@
+The OpenFHE library works with both 64-bit and 128-bit integers.
```
sloede, line 1: Same comments as in the other README.md apply

O/OpenFHE_128/build_tarballs.jl
```diff
@@ -0,0 +1,18 @@
...
+# If you make changes to this file, be sure to update ../OpenFHE/build_tarballs.jl as well.
+include(joinpath(pwd(), "..", "OpenFHE", "common.jl"))
+
+name = "OpenFHE_128"
+version = v"1.2.3"
+
+# symlink patch file from OpenFHE
+mkpath(joinpath(pwd(), "bundled", "patches"))
+cp(joinpath(pwd(), "..", "OpenFHE", "bundled", "patches", "windows-fix-cmake-libdir.patch"),
+   joinpath(pwd(), "bundled", "patches", "windows-fix-cmake-libdir.patch"))
```
sloede, line 13: Can't you just symlink the files directly in git? You might need to do this from a linux machine, but copying the files here seems weird to me...

O/OpenFHE_128/build_tarballs.jl
```diff
@@ -0,0 +1,18 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+
+# If you make changes to this file, be sure to update ../OpenFHE/build_tarballs.jl as well.
```
sloede, line 4: Same comment as in the other `build_tarballs.jl` applies here as well

sloede (changes requested):

O/OpenFHE/common.jl
```diff
@@ -0,0 +1,84 @@
+using BinaryBuilder, Pkg
+
+function prepare_openfhe_build(name::String, git_hash::String)
+    # Collection of sources required to complete build
+    sources = [
+        GitSource("https://github.com/openfheorg/openfhe-development.git",
+                  git_hash),
+        DirectorySource("./bundled")
+    ]
```
sloede, line 9: Instead of creating a symlink to the `bundled` directory, could you just modify the directory source here? That is, use something like
```suggestion
    # Collection of sources required to complete build
    # Note: The directory source is made to work for both OpenFHE and OpenFHE_128 builds
    sources = [
        GitSource("https://github.com/openfheorg/openfhe-development.git",
                  git_hash),
        DirectorySource("../OpenFHE/bundled")
    ]
```
If I'm not mistaken, this should work for either library build?

ArseniyKholod: That is really a cool idea! It works! Thank you.

sloede:
Another thought - what is the best name? `OpenFHE_128` is rather nondescript. Would `OpenFHE_128bit` be better? Or something else?

However, in the end, it's not such a big deal since a user should ideally never see its name anyways.

ArseniyKholod:
> Another thought - what is the best name? `OpenFHE_128` is rather nondescript. Would `OpenFHE_128bit` be better? Or something else?

128 bit has dual meaning in OpenFHE, native int size and security level. So I have no preference between 128 or 128bit. If you think 128bit is more descriptive, I have nothing against it;)

sloede:
> > Another thought - what is the best name? `OpenFHE_128` is rather nondescript. Would `OpenFHE_128bit` be better? Or something else?
> 
> 128 bit has also dual meaning in OpenFHE, native int size and security level. So I have no preference between 128 or 128bit. If you think 128bit is more descriptive, I have nothing against it;)

What about `OpenFHE_int128`? Or would that be too long (or lead to confusion with Julia's `Int128` type)?

ArseniyKholod:
> What about `OpenFHE_int128`? Or would that be too long (or lead to confusion with Julia's `Int128` type)?
Yeah, it is better, I have renamed.

sloede (approved):

sloede:
@giordano This is ready to be merged from our side 😊

sloede:
Thanks 🙏

PR #9774, opened 2024-11-13
[libcxxwrap-julia] Version 0.14.0

barche:
This fixes some issues with the STL wrappings and allows simplification of the OpenCV patch.

Breaks binary compatibility, hence the bump to 0.14.


giordano:
Side question: any interest in dropping support for Julia v1.9- at some point, now that v1.10 is the new LTS?  That'd reduce the number of artifacts a bit 🥲

barche:
I'd be happy to drop support for Julia <= 1.9 already this release, but maybe it's best to do this through a change in `julia_versions` defined in libjulia_jll? CC @fingolfin

PR #9777, opened 2024-11-14
[gismo] New Recipe: Geometry + Simulation Modules v24.08.0

hverhelst:
This is a new recipe for the Geometry + Simulation Modules (_gismo_), a C++ library for Isogeometric Analysis.
See [https://github.com/gismo/gismo](https://github.com/gismo/gismo)

### DIFF
--- a/G/gismo/build_tarballs.jl
+++ b/G/gismo/build_tarballs.jl
@@ -11,7 +11,6 @@ sources = [
 # NOTE: to control nproc, use the environment variable BINARYBUILDER_NPROC=<number of processors>
 script = raw"""
 cmake -B build                                      \
-  `# cmake specific`                                \
   -DCMAKE_INSTALL_PREFIX=${prefix}                  \
   -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}  \
   -DCMAKE_BUILD_TYPE=Release                        \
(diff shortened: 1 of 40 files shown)


G/gismo/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+version = v"24.08.0"
+sources = [
+    GitSource("https://github.com/gismo/gismo.git",       # The URL of the git repository
+              "fac4a95e60825bbc34014a524759e1178c9be646") # The commit hash to checkout
+]
+
+# NOTE: to control nproc, use the environment variable BINARYBUILDER_NPROC=<number of processors>
+script = raw"""
+cmake -B build                                      \
+  `# cmake specific`                                \
```
giordano, line 14:
```suggestion
```

G/gismo/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+  -DCMAKE_INSTALL_PREFIX=${prefix}                  \
+  -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}  \
+  -DCMAKE_BUILD_TYPE=Release                        \
+  -DGISMO_OPTIONAL="gsCInterface"                   \
+  -DGISMO_gsCInterface_HEAD=""                      \
+  -DGISMO_WITH_OPENMP=ON                            \
+  -DTARGET_ARCHITECTURE=none                        \
+  -DCMAKE_CXX_STANDARD=11                           \
```
giordano, line 22: You shouldn't need to specify the C++ standard: if they package fails to do so then it's in their build system that should be reported and fixed upstream, not something we should deal with.

hverhelst:
@giordano thanks for having a look! I fixed both your points. Let me know if there is anything else to fix!

giordano:
Looking better now, thanks for your contribution!

PR #9788, opened 2024-11-15
[libjulia] Remove versions older than 1.10

barche:
As pointed out by @giordano (https://github.com/JuliaPackaging/Yggdrasil/pull/9774#issuecomment-2474091059): because Julia 1.10 is now the official LTS, older versions of libjulia no longer need to be maintained here, and the list of versions can start at 1.10, greatly reducing the number of artifacts produced by libcxxwrap-julia and its dependents.


fingolfin:
Hrm, in Oscar.jl we would like to keep supporting older Julia versions a bit longer. yes, 1.10 is LTS, but it is also barely a year old. While I will quickly adopt any new release, I still know plenty people who have 1.8 or 1.9 installed.

Is there any concrete problem this solves?

L/libjulia/build_tarballs.jl
```diff
@@ -31,7 +31,7 @@ function Pkg.Types.is_stdlib(uuid::Base.UUID, julia_version::VersionNumber)
     return false
 end
 
-jllversion=v"1.10.13"
+jllversion=v"1.10.14"
```
lgoettgens, line 34:
```suggestion
jllversion=v"1.11.0"
```
restricting the compatible julia versions should include a minor version bump

barche:
Well, if we don't limit the versions at some point the list of artifacts will keep growing. For me it's OK to close this for now and postpone the decision.

barche:
I propose we revisit this after Julia 1.12 is released, or when and if there is a real pressing reason to do this. Closing for now.

PR #9802, opened 2024-11-17
[QRUpdate_ng] Rebuild 1.1.5 with LP64 rather than ILP64.

ViralBShah:
It appears that the upstream package does not handle ILP64 BLAS and LAPACK. The produced symbols are not mangled. Rebuild it to link for LP64 and use libblastrampoline.

Also pull in all the latest platforms.


ViralBShah:
@giordano Would be good to have your eyes on this diff here, especially since I added Wine for the build. I am not sure if Wine is required at runtime or not. It appears it is just a build time dependency: https://github.com/mpimd-csc/qrupdate-ng/blob/master/INSTALL.md

ViralBShah:
cc @JuhaHeiskala

PR #9803, opened 2024-11-17
[LAPACK32] Build against OpenBLAS32

ViralBShah:
Build LAPACK32 directly against OpenBLAS32. 

@amontoison This is a bit of a step backwards from LBT. The reason is that when I build other executables, like Octave, where one would like to run the JLL binaries directly, it is unclear where the LBT forwarding should be done. Unless of course, we require having an Octave.jl that does all the forwarding and such before binaries using LBT for 32-bit BLAS forwarding can be used.

Any thoughts?


amontoison:
Sorry @ViralBShah for not answering earlier.
You can use some environment variable to specify to which library LBT should forward the symbols.

We encountered this issue with the binaries of optimization solvers compiled with LBT (like Ipopt).

PR #9810, opened 2024-11-19
[CImGuiPack] Update for ImGui v1.91.6

JamesWrigley:
Building for ~~https://github.com/JuliaImGui/cimgui-pack/pull/8~~ https://github.com/JuliaImGui/cimgui-pack/pull/9.


JamesWrigley:
After yet more build failures I give up on aarch64-freebsd for now... One of the x86_64-linux-musl builds failed for an entirely different reason though:
```
🚨 Error: The plugin coppermind post-command hook exited with status 22
```

All the previous attempts on that platform succeeded so maybe it's just a fluke :thinking: Is it possible to redo that specific job or should I trigger CI again for all platforms?

JamesWrigley:
(thanks, looks like that did it :+1: )

PR #9813, opened 2024-11-19
[SHTns] up to v3.7 and add CUDA support

fgerick:


fgerick:
Does someone know what has to be done to keep the cuda builds from failing? When I run this locally it works fine and it seems it can build the library just fine, but then the dlopen checks fail. When using `dont_dlopen=true` the builds end up in an infinite loop.

fgerick:
Okay, after quite some painful trial and error I believe this is finally working. I've removed the micro-architecture expansion, that I think is not necessary with the removal of the `-mtune=skylake` flag that was still around in the build.

I hope this is correct how the CUDA and non-CUDA versions are build now and that the `x86_64-linux-gnu-cuda+none` is the correct way to do it. I do not manage to build everything without this flag for the CPU-only version.

fgerick:
Who should I get in touch with to supportive this? Thank you.

S/SHTns/build_tarballs.jl
```diff
@@ -2,53 +2,66 @@
...
+YGGDRASILPATH = joinpath(@__DIR__, "..", "..")
+
+include(joinpath(YGGDRASILPATH, "fancy_toys.jl"))
+include(joinpath(YGGDRASILPATH, "platforms", "microarchitectures.jl"))
+include(joinpath(YGGDRASILPATH, "platforms", "cuda.jl"))
 
 name = "SHTns"
-version = v"3.6.6"
+version = v"3.7"
+version_string = version.patch == 0 ? string(version.major)*"."*string(version.minor) : string(version)
```
giordano, line 13: This seems a lot simpler:
```suggestion
version_string = "3.7"
version = VersionNumber(version_string)
```

S/SHTns/build_tarballs.jl
```diff
@@ -57,15 +70,24 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="FFTW_jll", uuid="f5851436-0d7a-5f13-b9de-f02708fd171a")),
+    Dependency(PackageSpec(name="FFTW_jll")),
     # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
-    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
-    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
-    Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
+    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
```
giordano, line 75:
```suggestion
    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
```

S/SHTns/build_tarballs.jl
fgerick: true 👍

fgerick:
Is there anything else to do @giordano ?

PR #9817, opened 2024-11-19
New Recipe: libposit v1.0.0

hunhold:
[libposit](https://github.com/takum-arithmetic/libposit) is an ISO C99 implementation of [posit arithmetic](https://posithub.org/docs/posit_standard-2.pdf), a proposed improvement over floating-point arithmetic that is currently in the process of standardisation. This recipe provides libposit-1.0.0, the first stable release. There are no warnings generated on any of the tested platforms.

Thanks in advance for reviewing this pull request!


giordano:
Are you familiar with https://github.com/milankl/SoftPosit.jl? CC @milankl

hunhold:
> Are you familiar with https://github.com/milankl/SoftPosit.jl? CC @milankl

Thanks for your quick response, @giordano (CC @milankl)! Yes, I am familiar with this implementation. The main motivation behind libposit is two-fold: Firstly, supporting 64-bit posits, and secondly logarithmic posits in general (which have become a field of research in the last 1-2 years). In both aspects SoftPosit is lacking, and it's severely limited in regard to its architecture.

I just saw that the builds are failing due to __builtin_clz not being available on some platforms. I will adress this and push a change in this PR.

milankl:
>     1. supporting 64-bit posits to test high-precision arithmetic,

You could use any of the existing higher precision formats like `Float128` or `Double64` if you want to formulate posit arithmetic via back & forth conversion?

> In both aspects the pure Julia SoftPosit is lacking, partly due to some limitations imposed by the Julia language. This is no criticism of Julia, as I very well understand and even support these choices. However, in the posit context, it is currently not possible to write a pure Julia implementation of 64-bit posits, as they have a higher precision than Float64.

I don't understand your point. There are higher precision formats you can use, and I bet they'd work perfectly fine -- it's just that no one has done that yet.

> It is also not possible to support the so-called posit quires in pure Julia for the same reason, which is why they were dropped in SoftPosit.jl v0.5 even though they are also an active field of research.

No that was just me being lazy. I do see the appeal of quires, but my research back then was just not focussed on them. I wanted to drop the SoftPosit-C dependency as my pure Julia implementation was waaay faster and so instead of writing quires again, I decided to take them out as people were mostly excited about posits not about quires. You can put them back in if you like.

> There are a few other issues with SoftPosit.jl (rounding/truncation interfaces, negative integer powers, type promotion rules, etc.) that make it a bit fragile in use with numerical libraries; these issues have surfaced during work on a paper. The long-term plan is to cross-pollinate with SoftPosit and cross-validate implementations.

Why didn't you flag them? I know that SoftPosit could be better maintained but I welcome anyone to step in for this, happy to move the package to an organisation for example and have several people with maintainer status.

Having said this, I am not against the idea of a libposit C library, quite the opposite, especially one following the standard. It seems that you want to register this to build a Julia package on top of this? I'd be happy to make SoftPosit.jl do that. We could start by taking non-breaking features from libposit and get them in, replace functionality that's better there and do some performance comparisons too. Also Posits.jl isn't registered yet, so we could also merge both efforts towards such a package.

hunhold:
Thanks for your elaborate reply! It's nice to hear directly from you as the author of SoftPosit.jl.

> You could use any of the existing higher precision formats like `Float128` or `Double64` if you want to formulate posit arithmetic via back & forth conversion?

You are probably much more familiar with the landscape of extended precision libraries in Julia. From what I found out, e.g. looking at DoubleFloats.jl or Quadmath.jl, is that Float128 datatypes are realised with structs or tuples of two 64-bit primitives. 

> I don't understand your point. There are higher precision formats you can use, and I bet they'd work perfectly fine -- it's just that no one has done that yet.

Yes, you are totally right, I was wrong there. The thing is that there is no full-fledged C posit implementation, and the closest you get is C++ implementations that make heavy use of templates. This is the gap I want to fill with libposit. Given only C has a stable ABI (C++ does not), a solid C implementation is the only way to get proper FFI-libraries in other languages, unless you go through the trouble of doing it natively in the language as you did.

> > It is also not possible to support the so-called posit quires in pure Julia for the same reason, which is why they were dropped in SoftPosit.jl v0.5 even though they are also an active field of research.
> 
> No that was just me being lazy. I do see the appeal of quires, but my research back then was just not focussed on them. I wanted to drop the SoftPosit-C dependency as my pure Julia implementation was waaay faster and so instead of writing quires again, I decided to take them out as people were mostly excited about posits not about quires. You can put them back in if you like.

Ah now I see! I thought it was due to Julia's limitation for such large datatypes. Performance is one further point of criticism I have against the SoftPosit C-library.

> Why didn't you flag them? I know that SoftPosit could be better maintained but I welcome anyone to step in for this, happy to move the package to an organisation for example and have several people with maintainer status.

Sadly had no time yet, but will address that soon.
 
> Having said this, I am not against the idea of a libposit C library, quite the opposite, especially one following the standard. It seems that you want to register this to build a Julia package on top of this? I'd be happy to make SoftPosit.jl do that. We could start by taking non-breaking features from libposit and get them in, replace functionality that's better there and do some performance comparisons too. Also Posits.jl isn't registered yet, so we could also merge both efforts towards such a package.

This is a very cool idea! I actually already have a [Posits.jl, a thin wrapper around libposit, ready to go here](https://github.com/takum-arithmetic/Posits.jl). My plan was, as soon as libposit_jll gets (hopefully) accepted, to test it and register a Posits.jl package. I'm sure you as a Julia expert see lots of ways the code can get refactored. :)

hunhold:
Oh yeah, and thanks, @milankl, for this interesting exchange! I hope I didn't come across too abrasive regarding SoftPosit.jl, as I had a lot of trouble with the C library SoftPosit, which however since v0.5 is no longer a factor for SoftPosit.jl.

Maintenance takes a lot of time, and fortunately I have a few years until I finish my PhD to polish the code and maintain it, and it will most likely be stable enough by then that only little maintenance is necessary. When talking to those interested in trying next-generation arithmetic like posits, it boiled down to two things: 64-bit support for e.g. iterative refinement and FFT benchmarks, and logarithmic posits to play around with NNs that benefit from it.

So while I didn't have only Julia in mind when writing libposit, and more striving for a general easy-to-FFI C library for any language (Haskell, Ada, Fortran, you name it), I hope that I can provide a small benefit to the Julia ecosystem this way.

hunhold:
Hey @giordano! I fixed the issue with __builtin_clz and now all build tests pass on all platforms.

Can I address any other issues that are preventing a merge? Thanks in advance for your time and feedback! :)

PR #9819, opened 2024-11-20
Rebuild METIS for new architectures

amontoison:


amontoison:
It's the first time that I encounter this error:
https://buildkite.com/julialang/yggdrasil/builds/14850#01934a13-6988-45f8-80a9-a05160d8fae0/659-962

@staticfloat Do you have idea of what is the issue on `x64-freebsd`?

amontoison:
You were right.
I restarted the build and it solved the issue.

PR #9820, opened 2024-11-20
[DACE] bump to v0.6.0

afossa:


afossa:
It seems I hit the same issue as #9819 but with Apple clang:
https://buildkite.com/julialang/yggdrasil/builds/14854#01934bb2-9654-498b-8f2f-fcffce33a13a

D/DACE/build_tarballs.jl
```diff
@@ -30,9 +30,10 @@ cmake . -B build \
     -DCMAKE_BUILD_TYPE=Release \
     -DWITH_PTHREAD=ON \
     -DWITH_ALGEBRAICMATRIX=ON \
-    -DCMAKE_CXX_STANDARD=17 \
+    -DCMAKE_CXX_STANDARD=20 \
```
giordano, line 33: The upstream package should really define the standard needed, it shouldn't be up to downstream packagers to do it. Can you please fix it upstream?

afossa: It is fixed now

D/DACE/build_tarballs.jl
```diff
@@ -53,7 +53,8 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     BuildDependency(PackageSpec(name="libjulia_jll")),
-    Dependency("libcxxwrap_julia_jll"; compat = "~0.13.2"),
+    Dependency("libcxxwrap_julia_jll"; compat = "~0.14.0"),
+    Dependency("Eigen_jll"; compat = "~3.4.0")
```
giordano, line 57: Isn't Eigen headers-only?  If so, it should be a build-only dependency, since it isn't needed at runtime.  Also, style request: please keep the trailing comma.

afossa: Yes it is. I updated it to build dependency only.

PR #9851, opened 2024-11-23
[Singular] ABI extended

hannes14:
@fingolfin


fingolfin:
@hannes14 so is this supposed to restore the *current* Sibgular.jl to working again? Or will it require an updated libsingular & Singular.jl?

in the latter case we should adjust the version offset here to increment the minor version.

in either case we should also yank the bad versions of this JLL from the registry 


CC @thofma @benlorenz

hannes14:
it works with current libsingulsr and singular.jl

PR #9858, opened 2024-11-23
[XSLT] Update dependency for freebsd-aarch64 support

jmert:
This takes inspiration from #9680 to raise the compatibility bound of the Libgpg_error_jll dependency to include a version that has the necessary FreeBSD support

### DIFF
--- a/X/XSLT/build_tarballs.jl
+++ b/X/XSLT/build_tarballs.jl
@@ -1,12 +1,12 @@
 using BinaryBuilder
 
 name = "XSLT"
-version = v"1.1.41"
+version = v"1.1.42"
 
 # Collection of sources required to build XSLT
 sources = [
     ArchiveSource("https://download.gnome.org/sources/libxslt/$(version.major).$(version.minor)/libxslt-$(version).tar.xz",
-                  "3ad392af91115b7740f7b50d228cc1c5fc13afc1da7f16cb0213917a37f71bda"),
+                  "85ca62cac0d41fc77d3f6033da9df6fd73d20ea2fc18b0a3609ffb4110e1baeb"),
 ]
 
 # Bash recipe for building across all platforms
--- a/X/XSLT/build_tarballs.jl
+++ b/X/XSLT/build_tarballs.jl
@@ -44,4 +44,4 @@ dependencies = [
 # Build the tarballs, and possibly a `build.jl` as well.
 # XML2_jll builds with GCC 8 and we need to do the same to avoid linker errors
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat="1.6", preferred_gcc_version=v"8", )
+               julia_compat="1.6", preferred_gcc_version=v"11")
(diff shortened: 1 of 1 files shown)


X/XSLT/build_tarballs.jl
```diff
@@ -34,7 +34,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Libgpg_error_jll", v"1.42.0"; compat="1.42.0"),
+    Dependency("Libgpg_error_jll"; compat="1.50"),
```
giordano, line 37: Is there a newer version of xslt to build? We can't change an existing compat without changing the version number of the package itself (it's a Pkg limitation)

jmert: Thankfully there is. Commit added to bump to v1.1.42.

jmert:
We'll, looks like no easy fix for the win32 build error — requesting an earlier version of XML2 breaks the aarch-freebsd build we're trying to add...

jmert:
Switching to GCC 11 fixes the build error, but I'll need some guidance on whether that's an acceptable change or not.

I found the clue to updating the compiler version in the `config.log` during its check for `snprintf` (which fails and then causes the search for the trio library, that the reverted patch was attempting to make use of):
```
configure:14762: checking for snprintf
configure:14762: cc -o conftest.exe -g -O2   conftest.c  >&5
conftest.c:65:6: warning: conflicting types for built-in function 'snprintf' [-Wbuiltin-declaration-mismatch]
 char snprintf ();
      ^~~~~~~~
/opt/i686-w64-mingw32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld: /opt/i686-w64-mingw32/bin/../i686-w64-mingw32/sys-root/lib/../lib/libmsvcrt.a(lib32_libmsvcrt_extra_a-_vscprintf.o): in function `_mingw_get_msvcrt_handle':
/workspace/srcdir/mingw-w64-v11.0.1/mingw-w64-crt/include/msvcrt.h:9: undefined reference to `_imp__GetModuleHandleW@4'
/opt/i686-w64-mingw32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld: /opt/i686-w64-mingw32/bin/../i686-w64-mingw32/sys-root/lib/../lib/libmsvcrt.a(lib32_libmsvcrt_extra_a-_vscprintf.o): in function `init_vscprintf':
/workspace/srcdir/mingw-w64-v11.0.1/mingw-w64-crt/stdio/_vscprintf.c:73: undefined reference to `_imp__GetProcAddress@8'
configure:14762: $? = 1
configure: failed program was:
```

Note that the `libmsvcrt.a` library apparently has objects compiled from sources in the suggestively-named `mingw-w64-v11.0.1` directory.

I think I'll need help here as this is stretching beyond my understanding of how the root FS + GCC + Mingw/Windows combination is all supposed to work.

PR #9863, opened 2024-11-24
[CXSparse 4.4.0] Build for SuiteSparse 7.7.0 for Julia 1.11

ViralBShah:
We need to build CXSparse for every release SuiteSparse that is in Julia. Julia 1.10 is not straightforward because it uses SuiteSparse 7.2 which had the old build system.

### DIFF
--- a/S/SuiteSparse/CXSparse/build_tarballs.jl
+++ b/S/SuiteSparse/CXSparse/build_tarballs.jl
@@ -1,8 +1,8 @@
 include("../common.jl")
 
 name = "CXSparse"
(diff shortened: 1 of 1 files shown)


ViralBShah:
@rayegun If we don't use `SUITESPARSE_USE_SYSTEM_SUITESPARSE_CONFIG` - will that remove the dependency that CXSparse has for libsuitesparseconfig? In which case it can be completely separated from the SuiteSparse version we ship.

rayegun:
Let me test. I think that's fairly likely but I have to check.

PR #9864, opened 2024-11-25
[Singular] Revert recent changes due to downstream breakage

lgoettgens:
This reverts PRs #9811, #9840, #9846, #9851. (i.e. everything since last Friday).

Therefore I expect it to resolve https://github.com/oscar-system/Oscar.jl/issues/4341 that started happening after #9811 (but no idea how to test that locally). This would be the fastest way to get green OSCAR CI again that I can think of. The "real fixes" can be worked on afterwards, maybe in a minor version bump to not have it automatically downstream.

ping @hannes14 @fingolfin

### DIFF
--- a/S/Singular/build_tarballs.jl
+++ b/S/Singular/build_tarballs.jl
@@ -28,8 +28,8 @@ import Pkg.Types: VersionSpec
 #
 name = "Singular"
 
-upstream_version = v"4.4.0-7" # 4.4.0p7
-version_offset = v"0.0.8"
+upstream_version = v"4.4.0-7" # 4.4.0p6
+version_offset = v"0.0.7"
 
 version = VersionNumber(upstream_version.major * 100 + upstream_version.minor + version_offset.major,
                         upstream_version.patch * 100 + version_offset.minor,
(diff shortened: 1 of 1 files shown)


fingolfin:
Fine by me of @thofma and @benlorenz agree. Perhaps we should also yank the bad versions from the registry.

Then afterwards @hannes14 can resubmit this but with a `version_offset` of `0.1.0` to avoid existing Singular.jl using it.

S/Singular/build_tarballs.jl
```diff
@@ -28,16 +28,16 @@ import Pkg.Types: VersionSpec
 #
 name = "Singular"
 
-upstream_version = v"4.4.0-7" # 4.4.0p7
-version_offset = v"0.0.10"
+upstream_version = v"4.4.0-6" # 4.4.0p6
+version_offset = v"0.0.11"
```
lgoettgens, line 32: I only just saw that the VersionNumber scheme here includes the pre-release data, so we would need one of the following two to get a higher version than `404.0.710`
```suggestion
upstream_version = v"4.4.0-6" # 4.4.0p6
version_offset = v"0.0.111"
```
or
```suggestion
upstream_version = v"4.4.0-7" # 4.4.0p6 (the difference here is due to a retroactive re-release of 404.0.606 as 404.0.711)
version_offset = v"0.0.11"
```

PR #9865, opened 2024-11-25
[Singular] restore 4.4.0p7 but with minor version +1

fingolfin:
For reference, `Singular.jl` currently has this compat entry `Singular_jll = "~404.000.606"` which this PR should produce version `404.001.700`.

This way the current Singular.jl release will not pick it up giving us time to adjust and release libsingular_julia_jll and then Singular.jl and then finally Oscar.jl (via  PR that changes the compat entry in Project.toml also adjusts the doctests).

ping @hannes14 @thofma @benlorenz @lgoettgens 

Let's wait with merging this until *at least* https://github.com/JuliaRegistries/General/pull/120134 is merged. And it'd be great if someone double checks my logic above so we don't run into a brick wall right again.

### DIFF
--- a/S/Singular/build_tarballs.jl
+++ b/S/Singular/build_tarballs.jl
@@ -28,8 +28,8 @@ import Pkg.Types: VersionSpec
 #
 name = "Singular"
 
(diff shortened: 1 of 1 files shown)


lgoettgens (approved):
I can follow your version number reasoning and see no problems

PR #9870, opened 2024-11-26
[Xorg_libxcb] Rebuild for freebsd-aarch64

jmert:
In checking that the dependencies were properly built, it seems that the dependecy on XSLT_jll is unnecessary. The in-source `ChangeLog` file (which has more detail than the git-repo NEWS [1]) says that XSLT has been made optional since v1.10 in 2013, and the Arch Linux package [2] agrees in making it a build-only dependecy. See also [3] where it is mentioned that it's only used for generating HTML checks results.

[1] https://gitlab.freedesktop.org/xorg/lib/libxcb/-/blob/master/NEWS
[2] https://archlinux.org/packages/extra/x86_64/libxcb/
[3] https://bugs.freedesktop.org/show_bug.cgi?id=23863

### DIFF
--- a/X/Xorg_libxcb/build_tarballs.jl
+++ b/X/Xorg_libxcb/build_tarballs.jl
@@ -59,7 +59,6 @@ dependencies = [
     BuildDependency("Xorg_util_macros_jll"),
     BuildDependency("Xorg_xproto_jll"),
     BuildDependency("Xorg_xcb_proto_jll"),
-    Dependency("XSLT_jll"),
     Dependency("Xorg_libXau_jll"),
     Dependency("Xorg_libXdmcp_jll"),
     Dependency("Xorg_libpthread_stubs_jll"),


X/Xorg_libxcb/build_tarballs.jl
```diff
@@ -59,7 +59,7 @@ dependencies = [
     BuildDependency("Xorg_util_macros_jll"),
     BuildDependency("Xorg_xproto_jll"),
     BuildDependency("Xorg_xcb_proto_jll"),
-    Dependency("XSLT_jll"),
+    Dependency("XSLT_jll"),  # remove in next release --- only used for docs check: https://bugs.freedesktop.org/show_bug.cgi?id=23863
```
giordano, line 62: You can already remove it now.  It'll be still a dependency of the package, but won't be loaded.

jmert:
@giordano You mentioned in the [XSLT PR](https://github.com/JuliaPackaging/Yggdrasil/pull/9858#discussion_r1855272183) that Pkg is limited to changing dependencies only when a new version is released — I assume that also applies here in wanting to remove a dependency?

X/Xorg_libxcb/build_tarballs.jl
jmert: Made a comment faster than I could ask my clarifying question :-D I'll remove and push again.

PR #9891, opened 2024-11-28
[OpenFHE] introduce common.jl to share source code with OpenFHE_int128

ArseniyKholod:
The build recipes for OpenFHE and OpenFHE_int128(to be introduced) are almost identical. We collect all of the common source code in common.jl


ArseniyKholod:
@sloede, could you approve, please?

O/OpenFHE/README.md
```diff
@@ -0,0 +1,8 @@
+The OpenFHE library works with both 64-bit and 128-bit integers.
+
+This package builds the default 64-bit version.
+
+To build the 128-bit version, use OpenFHE_128.
+
+Most of the build steps for OpenFHE and OpenFHE_128 are the same, so the shared parts are in the
+common.jl script.
```
sloede, line 8: Can you maybe add links to the `OpenFHE_128` directory and the `common.jl` script for easy access?

O/OpenFHE/common.jl
```diff
@@ -0,0 +1,75 @@
...
+
+    cmake .. \
+      -DCMAKE_INSTALL_PREFIX=$prefix \
+      -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DWITH_BE2=ON \
+      -DWITH_BE4=ON \
+      -DBUILD_UNITTESTS=OFF \
+      -DBUILD_BENCHMARKS=OFF \
+    """ * (name == "OpenFHE_128" ? "-DNATIVE_SIZE=128" : "") *
```
sloede, line 30: IMHO it would be clearer if you always include the `-DNATIVE_SIZE=` argument and set it to either `64` or `128` (maybe by setting a variable above, such that there is less logic here)

O/OpenFHE/common.jl
```diff
@@ -0,0 +1,75 @@
+using BinaryBuilder, Pkg
+
+function prepare_openfhe_build(name::String)
+    # Collection of sources required to complete build
+    sources = [
+        GitSource("https://github.com/openfheorg/openfhe-development.git",
+                  "7b8346f4eac27121543e36c17237b919e03ec058"),
```
sloede, line 7: IMHO the hash needs to be an input variable as well, since it changes with each version. You could, e.g., add a second argument `git_hash` to `prepare_openfhe_build`

O/OpenFHE/build_tarballs.jl
```diff
@@ -1,72 +1,13 @@
 # Note that this script can accept some limited command-line arguments, run
 # `julia build_tarballs.jl --help` to see a usage message.
-using BinaryBuilder, Pkg
+
+# If you make changes to this file, be sure to update ../OpenFHE_128/build_tarballs.jl as well.
```
sloede, line 4: I would put the comment next to the `name = ...` line, since the `common.jl` file should not be changed, right? Also, maybe make the motivation more explicit:
```julia
# If you make changes in this file, e.g., to release a new version,
# be sure to also release a new version of `OpenFHE_128` as well (see `../OpenFHE_128/build_tarballs.jl`)
```

sloede (changes requested):

sloede (approved):
LGTM, thanks!

sloede (approved):
LGTM, thanks a lot!

O/OpenFHE/common.jl
```diff
@@ -0,0 +1,78 @@
...
+        # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
+        # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
+        Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae");
+                   platforms=filter(!Sys.isbsd, platforms)),
+        Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e");
+                   platforms=filter(Sys.isbsd, platforms)),
+    ]
+
+    return sources, script, platforms, products, dependencies
+end
```
giordano, line 78: Missing newline
```suggestion
end

```

PR #9892, opened 2024-11-28
MAGEMin v1.6.0

NicolasRiel:
- Added extended metapelite database to account for CO2 and S
- Added option to select combination of solution phase for metapelite and metabasite database
- Added “light” option to save minimum output from MAGEMin_C calculations
- Added several pseudocompounds for metapelite to improve phase diagrams quality 
- Corrected mistake when saving endmember wt fraction and end member wt composition


NicolasRiel:
@giordano Thanks!

PR #9899, opened 2024-11-30
[GSL] Update to 2.8

inkydragon:
Not sure if we want to remove this:

https://github.com/JuliaPackaging/Yggdrasil/blob/e504e3ece997c2977e71d1004749cbd2bb4f42ab/G/GSL/GSL%402/build_tarballs.jl#L20-L25

_GSL-2.8/configure_
```sh
    case $host_os in
    rhapsody* | darwin1.[012])
      _lt_dar_allow_undefined='$wl-undefined ${wl}suppress' ;;
    darwin1.*)
      _lt_dar_allow_undefined='$wl-flat_namespace $wl-undefined ${wl}suppress' ;;
    darwin*)
      case $MACOSX_DEPLOYMENT_TARGET,$host in
        10.[012],*|,*powerpc*-darwin[5-8]*)
          _lt_dar_allow_undefined='$wl-flat_namespace $wl-undefined ${wl}suppress' ;;
        *)
          _lt_dar_allow_undefined='$wl-undefined ${wl}dynamic_lookup $wl-no_fixup_chains' ;;
      esac
    ;;
```

### DIFF
--- a/G/GSL/GSL@2/bundled/patches/0001-remove-unknown-ld-option.patch
+++ b/G/GSL/GSL@2/bundled/patches/0001-remove-unknown-ld-option.patch
@@ -1,7 +1,7 @@
 Remove unknown linker option: -no_fixup_chains
 
---- configure   2024-05-25 21:26:49.000000000 +0800
-+++ configure   2024-11-30 23:33:16.848212900 +0800
+--- a/configure   2024-05-25 21:26:49.000000000 +0800
++++ b/configure   2024-11-30 23:33:16.848212900 +0800
 @@ -8764,7 +8764,7 @@
          10.[012],*|,*powerpc*-darwin[5-8]*)
            _lt_dar_allow_undefined='$wl-flat_namespace $wl-undefined ${wl}suppress' ;;
--- a/G/GSL/GSL@2/bundled/patches/0001-remove-unknown-ld-option.patch
+++ b/G/GSL/GSL@2/bundled/patches/0001-remove-unknown-ld-option.patch
@@ -1,0 +1,13 @@
+Remove unknown linker option: -no_fixup_chains
+
+--- configure   2024-05-25 21:26:49.000000000 +0800
++++ configure   2024-11-30 23:33:16.848212900 +0800
+@@ -8764,7 +8764,7 @@
+         10.[012],*|,*powerpc*-darwin[5-8]*)
+           _lt_dar_allow_undefined='$wl-flat_namespace $wl-undefined ${wl}suppress' ;;
+         *)
+-          _lt_dar_allow_undefined='$wl-undefined ${wl}dynamic_lookup $wl-no_fixup_chains' ;;
++          _lt_dar_allow_undefined='$wl-undefined ${wl}dynamic_lookup' ;;
+       esac
+     ;;
+   esac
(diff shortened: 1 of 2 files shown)


inkydragon:
aarch64-unknown-freebsd
```
[14:35:53] In file included from fp.c:26:
[14:35:53] ./fp-freebsd.c:27:3: error: unknown type name 'fp_prec_t'; did you mean 'fp_rnd_t'?
[14:35:53]    27 |   fp_prec_t prec = 0 ;
[14:35:53]       |   ^~~~~~~~~
[14:35:53]       |   fp_rnd_t
[14:35:53] /opt/aarch64-unknown-freebsd13.2/aarch64-unknown-freebsd13.2/sys-root/usr/include/machine/ieeefp.h:34:3: note: 'fp_rnd_t' declared here
[14:35:53]    34 | } fp_rnd_t;
[14:35:53]       |   ^
[14:35:53] In file included from fp.c:26:
[14:35:53] ./fp-freebsd.c:34:14: error: use of undeclared identifier 'FP_PS'
[14:35:53]    34 |       prec = FP_PS;
[14:35:53]       |              ^
[14:35:53] ./fp-freebsd.c:35:7: error: call to undeclared function 'fpsetprec'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
[14:35:53]    35 |       fpsetprec(prec);      
[14:35:53]       |       ^
[14:35:53] ./fp-freebsd.c:38:14: error: use of undeclared identifier 'FP_PD'
[14:35:53]    38 |       prec = FP_PD;
[14:35:53]       |              ^
[14:35:53] ./fp-freebsd.c:42:14: error: use of undeclared identifier 'FP_PE'
[14:35:53]    42 |       prec = FP_PE;
[14:35:53]       |  
```

https://github.com/freebsd/freebsd-src/blob/release/13.2.0/sys/amd64/include/ieeefp.h#L64-L72

I wonder if this is an error in our toolchain.

G/GSL/GSL@2/bundled/patches/0001-remove-unknown-ld-option.patch
```diff
@@ -0,0 +1,13 @@
+Remove unknown linker option: -no_fixup_chains
```
inkydragon, line 1: ref:

- https://github.com/haskell/zlib/issues/53
- https://github.com/macports/macports-ports/commit/90be777d2ce451d3c23783cb2be0efab9732e4d0

giordano:
> https://github.com/freebsd/freebsd-src/blob/release/13.2.0/sys/amd64/include/ieeefp.h#L64-L72
> 
> I wonder if this is an error in our toolchain.

I believe you're looking at the wrong file, since the failure is in the aarch64-freebsd platform you should look at
https://github.com/freebsd/freebsd-src/blob/d4dd9e22c13896e6b5e2a6fc78dad4f8496cc14d/sys/arm64/include/ieeefp.h and here I don't see `fp_prec_t` being defined.  Looks like an upstream bug to me.  Easier is to skip the platform for the time being.

inkydragon:
It seems that 2.7 can't be built on aarch64-unknown-freebsd either, so let's skip that platform for now.

PR #9912, opened 2024-12-02
New Version: FuzzifiED v0.10.5

mankai-chow:
I notice that there are some issue with Arpack_jll v3.9 --- it cannot find link to the BLAS library, so I am forcing compatibility with only the lower versions.


mankai-chow:
Can you help me exclude the platform aarch64-unknown-freebsd that failed the pipeline test ? The versions <3.8 of Arpack_jll do not support these platforms.

mankai-chow:
> Can you help me exclude the platform aarch64-unknown-freebsd that failed the pipeline test ? The versions <3.8 of Arpack_jll do not support these platforms.

Solved.

PR #9913, opened 2024-12-02
MAGEMin v1.6.1

NicolasRiel:
- Added frac_M_vol, frac_F_vol and frac_S_vol output for MAGEMin_C
- Corrected wrong set of W's for spinel in the metapelite and metapelite extended database
- Added molar mass of CO2 for MAGEMin_C (which was leading to failed mpe bulk loading in the App)


NicolasRiel:
Thank you very much! @giordano

PR #9917, opened 2024-12-03
[GR] bump version

jheinen:
bump GR run-time version


jheinen:
@giordano : Could you please merge #9917? Thanks in advance.

PR #9927, opened 2024-12-05
Bump mlpack, ensmallen, and Armadillo to latest stable versions.

rcurtin:
No serious changes, just a version bump... let's see if CI is happy :smile:


rcurtin:
Thanks @giordano as always! :rocket:

PR #9929, opened 2024-12-05
[Clp] Rebuild with latest LLVM toolchain

giordano:
We previously restricted compiling this package with LLVM 13 to work around an issue with the `lld` linker, but a better approach is to use the latest LLVM toolchain available but not use `lld` as linker with `clang_use_lld=false`.

### DIFF
--- a/C/Coin-OR/Clp/build_tarballs.jl
+++ b/C/Coin-OR/Clp/build_tarballs.jl
@@ -96,6 +96,6 @@ build_tarballs(
     products,
     dependencies;
     preferred_gcc_version = gcc_version,
-    preferred_llvm_version = llvm_version,
+    clang_use_lld = false,
     julia_compat = "1.9",
 )


odow:
Do we also need to rebuild the dependencies?

Osi_jll: https://github.com/JuliaPackaging/Yggdrasil/blob/aa7691d9cc9dce09e56f45a30dd829af6055e9b3/C/Coin-OR/Osi/build_tarballs.jl#L86
CoinUtils_jll: https://github.com/JuliaPackaging/Yggdrasil/blob/aa7691d9cc9dce09e56f45a30dd829af6055e9b3/C/Coin-OR/CoinUtils/build_tarballs.jl#L89

giordano:
The version of llvm shouldn't really matter much, we aren't aware of runtime incompatibilities unlike for libstdc++/libgfortran with gcc.

giordano:
This doesn't seem to help with https://github.com/jump-dev/Clp.jl/issues/131

PR #9932, opened 2024-12-06
[Qt6Charts] Add Qt6Charts

barche:
For issue https://github.com/JuliaGraphics/QML.jl/issues/210

### DIFF
--- a/Q/Qt6Charts/build_tarballs.jl
+++ b/Q/Qt6Charts/build_tarballs.jl
@@ -1,0 +1,88 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Qt6Charts"
+version = v"6.7.1"
+
+# Collection of sources required to build qt6
+sources = [
+    ArchiveSource("https://download.qt.io/official_releases/qt/$(version.major).$(version.minor)/$version/submodules/qtcharts-everywhere-src-$version.tar.xz",
+                  "c236875ab3c812f0d59bad81a783095ae50932c7bea23cb81cd55437083d5d8c"),
+    ArchiveSource("https://github.com/roblabla/MacOSX-SDKs/releases/download/13.3/MacOSX13.3.sdk.tar.xz",
+                  "e5d0f958a079106234b3a840f93653308a76d3dcea02d3aa8f2841f8df33050c"),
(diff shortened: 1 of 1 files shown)


Q/Qt6Charts/build_tarballs.jl
```diff
@@ -0,0 +1,88 @@
...
+platforms_macos = [ Platform("x86_64", "macos"), Platform("aarch64", "macos") ]
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct(["Qt6Charts", "libQt6Charts", "QtCharts"], :libqt6charts),
+    LibraryProduct(["Qt6ChartsQml", "libQt6ChartsQml", "QtChartsQml"], :libqt6chartsqml),
+]
+
+products_macos = [
+    FrameworkProduct("QtCharts", :libqt6svg),
```
imciner2, line 69:
```suggestion
    FrameworkProduct("QtCharts", :libqt6charts),
```
I assume you meant for this product to have the same name as the other platforms.

Q/Qt6Charts/build_tarballs.jl
```diff
@@ -0,0 +1,88 @@
...
+    FrameworkProduct("QtChartsQml", :libqt6chartsqml),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+    HostBuildDependency("Qt6Base_jll"),
+    HostBuildDependency("Qt6Declarative_jll"),
```
imciner2, line 76: Is there a reason these don't have to be version constrained during the build but you need an exact version constraint at runtime?

Q/Qt6Charts/build_tarballs.jl
barche: Missed that, thanks!

Q/Qt6Charts/build_tarballs.jl
barche: It is not possible to set a version for build dependencies I think. The runtime dependencies are what really matters because they determine the compat settings of the generated jll package.

imciner2: You can use a `PackageSpec` to build the dependency, see e.g. https://github.com/JuliaPackaging/Yggdrasil/blob/7b1e1f040e02079f300b78f06817c9db3c8273af/S/SuiteSparse/common.jl#L94

PR #9939, opened 2024-12-07
[GAP_pkg_juliainterface] update for GAP 4.14.0

fingolfin:


G/GAP_pkg/update.jl
```diff
@@ -62,7 +62,7 @@ function update_gap_pkg_recipe(dir)
 
     # new metadata from the GAP package registry
     if pkgname == "juliainterface"
-        upstream_version = "0.9.0"
+        upstream_version = "0.12.0"
```
lgoettgens, line 65:
```suggestion
        upstream_version = "0.13.0"
```
shouldn't this match the version from above?

PR #9961, opened 2024-12-09
[openfhe_julia_int128] initialize and [openfhe_julia] update

ArseniyKholod:


ArseniyKholod:
I have opened an issue to llvm about the typeinfo for __int128 problem: https://github.com/llvm/llvm-project/issues/119608.
For now, I disable Apple compilations for 128-bit version.

O/openfhe_julia/common.jl
```diff
@@ -0,0 +1,107 @@
...
+    
+    # For MacOS additional linker flag is required to link typeid(__int128)
+    if [[ "$target" == *-apple-darwin* ]]; then
+        export LDFLAGS="-lc++abi"
+    fi
+    export LDFLAGS="-lc++abi"
+    # For MacOS additional linker flag is required to link typeid(__int128)
+    if [[ "$target" != *-apple-darwin* ]]; then
+        export LDFLAGS=""
+    fi
```
sloede, line 51: @ArseniyKholod this seems wrong?

ArseniyKholod: Now it is actually fails only on x86_64, but aarch64 is OK

ArseniyKholod: And I tested only aarch86 in our repo

sloede (approved):
LGTM! And thanks a lot for keeping at it until you found a solution for the macOS issues 💪

PR #9962, opened 2024-12-09
Enabled GMock and exported libraries

el-oso:
Not sure why GMock was disabled in GoogleTest. I just reenabled it and increased the version number.

### DIFF
--- a/G/GoogleTest/build_tarballs.jl
+++ b/G/GoogleTest/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "GoogleTest"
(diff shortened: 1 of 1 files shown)


G/GoogleTest/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "GoogleTest"
-version = v"1.11.0"
+version = v"1.11.1"
```
giordano, line 6: No need to change the version number for this change.  You could also update to a newer version, up to you

el-oso:
ok, then let's roll it back to the previous version

PR #10000, opened 2024-12-13
[GAP_pkg_zeromqinterface] new package

fingolfin:


fingolfin:
I get a build failure x86_64-linux-musl because it doesn't find zmq.h -- it looks as if issue #3949 is back? The compiler here is once again clang, but this time not on FreeBSD but on musl C:
```
# cc --version
clang version 18.1.7 (/home/tim/.cache/BinaryBuilder/downloads/clones/llvm-project.git-1df819a03ecf6890e3787b27bfd4f160aeeeeacd50a98d003be8b0893f11a9be 768118d1ad38bf13c545828f67bd6b474d61fc55)
Target: x86_64-unknown-freebsd13.2
Thread model: posix
InstalledDir: /opt/x86_64-linux-musl/bin
```

Perhaps we ought to apply `add_system_includedir` in `BinaryBuilderBase` also in this situation? @giordano ?

(I can work around it here locally for now)

PR #10003, opened 2024-12-13
[xkbcommon] Bump to build for aarch64-freebsd

JamesWrigley:


JamesWrigley:
There's some weird pkgconf/meson stuff going on, attempted to fix it in 3f01012b. TL;DR I don't think this issue is fully fixed: https://github.com/pkgconf/pkgconf/issues/213

There's an open PR which will apparently fix it for realsies but I didn't really grok it: https://github.com/pkgconf/pkgconf/pull/280

PR #10006, opened 2024-12-13
[Wayland] Bump to build for aarch64-freebsd

JamesWrigley:


JamesWrigley:
Removing the compat bounds to allow using the latest libffi and expat JLLs fixed the build on aarch64-freebsd, but I see some warnings:
```julia
┌ Warning: wayland-scanner has an ELF header OS/ABI value that is not set to FreeBSD (9), this may be an issue at link time
```

Not sure how serious that is. They do seem to be specific to aarch64, there's no warnings on x86-64.

JamesWrigley:
Ok, I think the warning is a false positive: https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1344

W/Wayland/build_tarballs.jl
```diff
@@ -44,12 +44,13 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Expat_jll"; compat="2.2.10"),
-    Dependency("Libffi_jll"; compat="~3.2.2"),
+    Dependency("Expat_jll"),
+    Dependency("Libffi_jll"),
```
giordano, line 48: Please don't remove compat bounds, they're there for a reason: #3048.

giordano (changes requested):

W/Wayland/build_tarballs.jl
JamesWrigley: Ah, I missed that those require bounds... The problem is that only the newer versions of the JLLs have been built for aarch64-freebsd, so if the compat bounds are kept then downstream packages won't be able to support aarch64-freebsd (which would be fine with me given what a PITA rebuilding all the packages is). Is it possible to rebuild JLLs at a specific version? If we could rebuild Expat/Libffi at those versions to get aarch64-freebsd builds then I think everything would work.

giordano: Also, like in all other PRs, _changing_ compat bounds means that you must change the package number.

JamesWrigley: Ok, but my question remains: is it possible to rebuild the JLLs at those compat versions? Because otherwise I don't know how to get Wayland building on aarch64-freebsd.

giordano: Just revert to the old version, and then revert the revert.

JamesWrigley: Alrighty, with the new JLLs the build succeeds on aarch64-freebsd.

giordano (approved):

PR #10007, opened 2024-12-13
[Xorg_xkbcomp] Bump to build for aarch64-freebsd

JamesWrigley:


JamesWrigley:
The build failed because the new libxkbfile version hadn't been registered yet, now it should be: https://github.com/JuliaRegistries/General/pull/121359

PR #10008, opened 2024-12-13
[Reactant] refactor toolchains to better use Yggdrasil's cross-compilers

mofeing:
cc @wsmoses

### DIFF
--- a/R/Reactant/build_tarballs.jl
+++ b/R/Reactant/build_tarballs.jl
@@ -1,6 +1,8 @@
 using BinaryBuilder, Pkg
 using Base.BinaryPlatforms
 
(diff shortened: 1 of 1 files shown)


mofeing:
@giordano if this works, we know how to then create a generic bazel toolchain for yggdrasil

giordano:
It'd be great to put that in BinaryBuilderBase, to save the pain in the future.

mofeing:
> It'd be great to put that in BinaryBuilderBase, to save the pain in the future.

yes, we're first trying it here with Reactant as test case. and will later push to BinaryBuilderBase once we figure it out that everything works.

ViralBShah:
Closing since we have lots of new releases since. Please reopen if we still want it open.

PR #10010, opened 2024-12-13
[Xorg_libXinerama] Bump to build for aarch64-freebsd

JamesWrigley:


X/Xorg_libXinerama/build_tarballs.jl
```diff
@@ -38,4 +38,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
giordano, line 41: Changing compat bounds ***requires*** changing the version number of the package itself (Pkg won't be able to deal with that otherwise)

giordano (changes requested):

X/Xorg_libXinerama/build_tarballs.jl
JamesWrigley: Bumped the version in 870e0f32.

giordano (approved):

PR #10011, opened 2024-12-13
[Xorg_libXfixes] Bump to build for aarch64-freebsd

JamesWrigley:


X/Xorg_libXfixes/build_tarballs.jl
```diff
@@ -38,4 +38,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
```
giordano, line 41: Changing compat bounds ***requires*** changing the version number of the package itself (Pkg won't be able to deal with that otherwise)

giordano (changes requested):

X/Xorg_libXfixes/build_tarballs.jl
JamesWrigley: Bumped the version in 24444989.

X/Xorg_libXfixes/build_tarballs.jl
```diff
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "Xorg_libXfixes"
-version = v"5.0.3"
+version = v"6.0.1"
```
giordano, line 6: Did the API/ABI change in a breaking way?

JamesWrigley: From looking at the commits I think the ABI did break in 6.0.1 with this change to function argument types: https://gitlab.freedesktop.org/xorg/lib/libxfixes/-/commit/aba73d85ce4a2fc1a5fb0ec27b15415eed21ebae
Full changelog seems to be here: https://gitlab.freedesktop.org/xorg/lib/libxfixes/-/merge_requests/5

In 6.0.0 there's some minor build system changes and this addition to the API: https://gitlab.freedesktop.org/xorg/lib/libxfixes/-/commit/883f89dd39dcdae9a38e46f74a43b316180c8a19
So my best guess is that 6.0.1 is breaking but not 6.0.0 :thinking: I'll bump it down to 6.0.0.

giordano: > I think the ABI did break in 6.0.1 with this change to function argument types: [gitlab.freedesktop.org/xorg/lib/libxfixes/-/commit/aba73d85ce4a2fc1a5fb0ec27b15415eed21ebae](https://gitlab.freedesktop.org/xorg/lib/libxfixes/-/commit/aba73d85ce4a2fc1a5fb0ec27b15415eed21ebae)

I'm not convinced that affects the ABI: the type of the arguments of C functions isn't part of the ABI (unlike C++), and that's only changing the constantness of the type.  What's more interesting is whether the soname changed, you should compare the old one in https://github.com/JuliaBinaryWrappers/Xorg_libXfixes_jll.jl/blob/f31dffebcf47362ae0d1e1ceefe46527a1a03b59/src/wrappers/aarch64-linux-gnu.jl#L6 with the new one in the auditor log

JamesWrigley: Looks like the 6.0.1 build does have a matching soname: https://buildkite.com/julialang/yggdrasil/builds/15579#0193c4cd-a0ea-4192-8540-2362d0a1978e/669-1084
```julia
[ Info: lib/libXfixes.so.3.1.0 already has SONAME "libXfixes.so.3"
```

I thought that `const` mattered to the ABI for pointer argument types, but I'm not too familiar with these things. From looking at the commits I think that and the addition of `ClientDisconnectMode` are the only potential changes to the API/ABI. Do you reckon it's safe to go with 6.0.1 in that case?

JamesWrigley: FWIW I would still feel more comfortable sticking with 6.0.0 in the interest of changing as little as possible.

giordano (approved):

PR #10021, opened 2024-12-16
Restore LDAP support

maxfadson:
Hi! 

How about restoring LDAP support in libcurl_jll? Without it, working with LDAP servers in LibCURL is not possible. 

To ensure support, I added the OpenLDAPClient dependency, which provides the necessary headers.

### DIFF
--- a/L/LibCURL/common.jl
+++ b/L/LibCURL/common.jl
@@ -49,7 +49,7 @@ function build_libcurl(ARGS, name::String, version::VersionNumber)
         --without-libidn2 --without-librtmp
         --without-nss --without-libpsl
         --disable-ares --disable-manual
-        --disable-ldap --disable-ldaps --without-zsh-functions-dir
+        --without-zsh-functions-dir
         --disable-static --without-libgsasl
         --without-brotli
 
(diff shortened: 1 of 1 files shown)


giordano (changes requested):
Curl is a Julia dependency and adding dependencies to it has a cost. Someone tried to add kerberos in the past but we had to revert it quickly because everything broke down.

Furthermore, curl is part of a major update in Julia to move openssl and on top of everything I said above I'm strongly against making major changes to the package.

maxfadson:
@giordano what do you think about creating a separate package? For example, libcurl2_jll or another name, but without tying it to Julia’s core. This way, it could be modified and used as the foundation for an improved LibCURL.jl package (perhaps named LibCURL2.jl, for instance).

giordano:
You can only load a single shared library with a specific soname.  That's a limitation of how dynamic libraries work (and no, before someone suggests it, a platform-specific solution like `dlmopen` isn't viable, since it won't work anyway on all the other platforms).

maxfadson:
I think it’s better to move the discussion to [Slack](https://julialang.slack.com/archives/C674ELDNX/p1734364716961399) :)

ViralBShah:
I would imagine that it is ok to have a `LibCURL_extras` or some such thing, with the files within appropriately renamed.

PR #10041, opened 2024-12-19
[snappy] Remove unused dependencies

nhz2:
libsnappy does not need any dependencies. The dependencies are only used for the `snappy_test_tool` binary to compare performance with other compression libraries, but this binary is not being built.

### DIFF
--- a/S/snappy/build_tarballs.jl
+++ b/S/snappy/build_tarballs.jl
@@ -36,7 +36,7 @@ products = [
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = [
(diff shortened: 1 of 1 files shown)


nhz2:
This does not change the Windows and Linux artifact hashes.

S/snappy/build_tarballs.jl
```diff
@@ -37,9 +37,6 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
```
giordano, line 39:
```suggestion
dependencies = Dependency[
```

PR #10042, opened 2024-12-19
[ORTools] Bump version

dourouc05:


O/ORTools/build_tarballs.jl
```diff
@@ -164,4 +164,4 @@ products = [
 dependencies = Dependency[]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"11", julia_compat="1.9")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version=v"11", julia_compat="1.10")
```
giordano, line 167: Any specific reason for changing this?  It looks unnecessary to me.

dourouc05: It's an attempt at making more platforms compile. If it does not bring any benefit, I'll revert.

giordano: I don't see how changing this compat affects anything here.

PR #10068, opened 2024-12-22
MAGEMin v1.6.2

NicolasRiel:
- Added Igneous Alkaline Dry database (igad) from Weller et al., 2024
- Added Igneous Green et al., 2025 corrected and updated from Holland et al., 2018


NicolasRiel:
Thank you very much @giordano 
Merry Christmas!

PR #10158, opened 2024-12-27
[CI] Update Manifest

imciner2:
Update Manifest to get new versions of BinaryBuilder and BinaryBuilderBase to get the new Rust/Go versioning features.


giordano (approved):
Note that you can use https://github.com/JuliaPackaging/Yggdrasil/actions/workflows/update_manifest.yml, which also generates links to the diffs of BB and BBB, for future reference.  See for example #10084

PR #10178, opened 2024-12-30
[Qhull] Upgrade to v8.1.alpha4

giordano:


Q/Qhull/build_tarballs.jl
```diff
@@ -3,12 +3,12 @@
 using BinaryBuilder
 
 name = "Qhull"
-version = v"8.0.1003" # alpha-3 release of 8.1, but Julia Pkg doesn't support prerelease versions
+version = v"8.0.1004" # alpha.4 release of 8.1, but Julia Pkg doesn't support prerelease versions
```
giordano, line 6: I'm a bit confused by why we don't use 8.1.x instead of 8.0.100x, but ok.

PR #10186, opened 2025-01-01
[GraphicsMagick] Build without profiler support

ViralBShah:


ViralBShah:
This package does not get registered
```
ERROR: Overlapping ranges for JLLWrappers for version 1.3.45+0 in registry.
```
Any idea why? https://github.com/JuliaRegistries/General/pull/122252#issuecomment-2567067693

giordano:
All new builds must change the version number, because JLLWrappers changes the compat bound

ViralBShah:
Don’t we do build triggers that just rebuild the binaries and add +1 and such to the name? Why is this case different?

PR #10198, opened 2025-01-02
[HiGHS] update to v1.9.0

odow:


imciner2:
The RISC-V error that appeared in the first attempt is the same as this one: https://github.com/JuliaPackaging/Yggdrasil/pull/9967#issuecomment-2563936131 (missing the atomics symbols). There's a patch in that PR, or also it might be fixed if we get a newer GCC version (which is currently in progress).

PR #10206, opened 2025-01-05
MAGEMin v1.6.3

NicolasRiel:
- Added solution phase full name
- Updated mineral naming to fit THERMOCALC convention
- Corrected issue when deactivating solution phase model


NicolasRiel:
@giordano Thank you, like always!

PR #10210, opened 2025-01-07
New Recipe: mmtk_julia v0.30.2

udesou:
Adding a recipe to build the [mmtk-julia](https://github.com/mmtk/mmtk-julia/tree/master) binding, which should enable building Julia using [MMTk](https://www.mmtk.io).

### DIFF
--- a/M/mmtk_julia/build_tarballs.jl
+++ b/M/mmtk_julia/build_tarballs.jl
@@ -1,0 +1,38 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "mmtk_julia"
+version = v"0.30.2"
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://github.com/mmtk/mmtk-julia/archive/refs/tags/v$(version).tar.gz", "d0db08487adadb24c3ac4a8a6243ce895617bfc429edb008bc69db7ee7df45b4")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd mmtk-julia-0.30.2/
+make release 
+mv mmtk/target/x86_64-unknown-linux-gnu/release/libmmtk_julia.so ${libdir}
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc")
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("lib", :libmmtk_julia)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", compilers = [:rust, :c], preferred_gcc_version = v"13.2.0")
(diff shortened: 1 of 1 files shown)


M/mmtk_julia/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+sources = [
+    ArchiveSource("https://github.com/mmtk/mmtk-julia/archive/refs/tags/v$(version).tar.gz", "d0db08487adadb24c3ac4a8a6243ce895617bfc429edb008bc69db7ee7df45b4")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd mmtk-julia-0.30.2/
+make release 
+mv mmtk/target/x86_64-unknown-linux-gnu/release/libmmtk_julia.so ${libdir}
```
giordano, line 18:
```suggestion
install  -Dvm 755 "mmtk/target/${rust_target}/release/libmmtk_julia.${dlext}" -t "${libdir}"
```

M/mmtk_julia/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+products = [
+    LibraryProduct("lib", :libmmtk_julia)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", compilers = [:rust, :c], preferred_gcc_version = v"13.2.0")
```
giordano, line 38:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", compilers = [:rust, :c], preferred_gcc_version = v"10")
```

M/mmtk_julia/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+cd $WORKSPACE/srcdir
+cd mmtk-julia/
+make release 
+install  -Dvm 755 "mmtk/target/${rust_target}/release/libmmtk_julia.${dlext}" -t "${libdir}"
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc")
```
imciner2, line 24: Is this really only for this one platform?

giordano:
One concern I have is that by (implicitly) depending on libjulia this package isn't quite compatible with different Julia versions (ABI always changes across minor versions), on the other hand my understanding is that this package is only meant to be a (conditional/optional) dependency of Julia itself, and at the moment this is all a bit experimental still. 

Is the above an accurate description of this package?

udesou:
That’s correct. The big picture is that there has been some recent work on supporting third party GCs on Julia, by exposing a [GC interface](https://docs.google.com/document/u/0/d/1v0jtSrIpdEDNOxj5S9g1jPqSpuAkNWhr_T8ToFC9RLI/mobilebasic#heading=h.9bwdm63xoj0y) and MMTk is [being added](https://github.com/JuliaLang/julia/pull/56288) as one of the supporting GCs. The idea is to use this package instead of building the Rust part of the code when building Julia+MMTk which would require users (or the Julia CI) to have the Rust toolchain installed. 
But indeed, this package is only meant to be a dependency of Julia itself and the plan is continuously improve the support over time.

Keno:
Honestly, it seems a bit weird to me to run this through Yggdrasil. MMTK itself sure, but the julia bindings are very tightly tied to the runtime. In a world where it's the default GC, I would imagine it would even sit in the julia repo so that new GC-dependent features can be co-developed. I understand that that would mean the julia repo picks up a dependency on a rust toolchain, but we may just have to do that. Not to stand in the way of any near-term experimentation, just commenting on the longer-term path.

udesou:
@Keno That's a great point. I'll raise it in the next Julia GC meeting (feel free to join that meeting as well, btw!)

PR #10215, opened 2025-01-07
[SuiteSparse] Fix build for 64-bit BLAS

imciner2:
The new builds were not properly mangling the symbol names to call 64-bit BLAS functions, as noticed in https://github.com/JuliaLang/julia/pull/56959), this fixes that.

Also, remove hard-coded CMake version to let it get the newest version.


ViralBShah:
I see what I did wrong. Thanks.

PR #10218, opened 2025-01-08
[Geant4] Bump to latest version

peremato:


G/Geant4/bundled/patches/windows.patch
giordano, line 1: sounds like this patch should be upstreamed

peremato: Yes of course.

PR #10221, opened 2025-01-08
FlowCutterPACE17: Apple silicon support

samuelsonric:
Set `julia_compat = 1.6` so that binaries will be built for Apple silicon.

### DIFF
--- a/F/FlowCutterPACE17/build_tarballs.jl
+++ b/F/FlowCutterPACE17/build_tarballs.jl
@@ -19,8 +19,7 @@ c++ -o "${bindir}/flow_cutter_pace17${exeext}" -Wall -std=c++11 -O3 -DNDEBUG src
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = filter(p -> arch(p) != "riscv64", supported_platforms())
-platforms = expand_cxxstring_abis(platforms)
+platforms = expand_cxxstring_abis(supported_platforms())
 
 # The products that we will ensure are always built
 products = [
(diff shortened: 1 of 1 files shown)


giordano (dismissed):
Only changing compat bounds requires changing the version number (it's a Pkg limitation). Is there a newer version of the package?

samuelsonric:
> Only changing compat bounds requires changing the version number (it's a Pkg limitation). Is there a newer version of the package?

No! https://github.com/kit-algo/flow-cutter-pace17

samuelsonric:
Should I do what you did here? https://github.com/JuliaPackaging/Yggdrasil/pull/5833/files
With a `version` / `upstream_version` ?

giordano:
Why removing riscv64? Looks like it's something easy to patch, no?
```
[17:17:44] src/list_graph.h:7:1: note: ‘std::string’ is defined in header ‘<string>’; did you forget to ‘#include <string>’?
```

samuelsonric:
> Why removing riscv64? Looks like it's something easy to patch, no?
> 
> ```
> [17:17:44] src/list_graph.h:7:1: note: ‘std::string’ is defined in header ‘<string>’; did you forget to ‘#include <string>’?
> ```

It was not apparent to me how to fix the problem (I don't know C++). Thanks for pointing it out!

imciner2:
I can't make a suggestion on the right line, but you can add the following lines after you `cd` into the source directory to fix the compilation error:
```
# Fix missing include needed for newer GCC version
sed -i '/#include <tuple>/a #include <string>' src/list_graph.h
```

samuelsonric:
> I can't make a suggestion on the right line, but you can add the following lines after you `cd` into the source directory to fix the compilation error:
> 
> ```
> # Fix missing include needed for newer GCC version
> sed -i '/#include <tuple>/a #include <string>' src/list_graph.h
> ```

That's incredible! I was about to submit a patch to the original repository...

imciner2:
If you have a patch, please still submit it upstream. This error is actually because of the new GCC version we use, so other people could also be affected by it.

imciner2 (approved):

samuelsonric:
> If you have a patch, please still submit it upstream. This error is actually because of the new GCC version we use, so other people could also be affected by it.

Ok done: https://github.com/kit-algo/flow-cutter-pace17/pull/1

PR #10250, opened 2025-01-12
libCppInterOp: fix symbol visibility issues on Windows

Gnimuc:
- update patches
- retrigger build

### DIFF
--- a/L/libCppInterOp/bundled/patches/windows.patch
+++ b/L/libCppInterOp/bundled/patches/windows.patch
@@ -1,7 +1,7 @@
 From 21805e1bdbdda55f36f8e7d76a99032e86b0c750 Mon Sep 17 00:00:00 2001
 From: Gnimuc <qiyupei@gmail.com>
 Date: Sat, 11 Jan 2025 20:56:57 +0900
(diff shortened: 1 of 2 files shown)


L/libCppInterOp/bundled/patches/windows.patch
```diff
@@ -47,3 +47,286 @@ index e25ee9a..6064952 100644
 -- 
 2.39.5 (Apple Git-154)
 
+
+From ce769c6e91730c753e30dfd072a3d991d87e4688 Mon Sep 17 00:00:00 2001
+From: Gnimuc <qiyupei@gmail.com>
+Date: Sun, 12 Jan 2025 19:01:50 +0900
+Subject: [PATCH 2/2] Fix symbol visibility issues for C API
```
giordano, line 54: I presume this second patch should be upstreamed too?

Gnimuc: Yes. https://github.com/compiler-research/CppInterOp/pull/440

Gnimuc: Ideally all patches except the cmake patch should be upstreamed.

PR #10262, opened 2025-01-13
[snappy] Change version number to remove dependencies

nhz2:
Pkg still installs the dependencies removed in https://github.com/JuliaPackaging/Yggdrasil/pull/10041 when using `snappy_jll`.

Creating a new version of `snappy_jll` should fix this.


giordano:
Technically they shouldn't be loaded.  Are you worried they still appear in the dependency graph?

PR #10263, opened 2025-01-13
MAGEMin v1.6.4

NicolasRiel:
- Corrected issue that was making the ultramafic extended database crash
- Corrected issue when computing graphite mol fraction for metapelite extended database


giordano:
Powerpc64 with old libgfortran versions could be made work, but it's such a niche (and basically unsupported by Julia nowadays) that it's probably not worth it. Also, there's still libgfortran5 available (and most users would likely be using that one anyway)

NicolasRiel:
@giordano Thank you for the explanation! I think this is fine, we don't need it. Also the failed built happened on Julia 1.7, and I don't think we have users on such old version anyway.

PR #10265, opened 2025-01-13
[Reactant] Check CUDA version number in platform augmentation

giordano:
Ref: https://github.com/EnzymeAD/Reactant.jl/issues/526.

### DIFF
--- a/R/Reactant/build_tarballs.jl
+++ b/R/Reactant/build_tarballs.jl
@@ -9,7 +9,7 @@ repo = "https://github.com/EnzymeAD/Reactant.jl.git"
 version = v"0.0.40"
 
 sources = [
-  GitSource(repo, "6c98d9ace9043bd797d767377706874f4772468d"),
+  GitSource(repo, "ab946703ee6935887b7b441fbe16a3dac260fbc0"),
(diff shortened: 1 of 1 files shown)


R/Reactant/build_tarballs.jl
```diff
@@ -348,12 +368,19 @@ augment_platform_block="""
...
                 handle = Libdl.dlopen(cuname)
+                cuda_version = cuDriverGetVersion(handle)
                 path = Libdl.dlpath(handle)
                 Libdl.dlclose(handle)
 
-                @debug "Adding include dependency on \$path"
-                Base.include_dependency(path)
-                gpu = "cuda"
+                if cuda_version isa VersionNumber
+                    if cuda_version == v"12.6"
```
giordano, line 376: I don't know what check you exactly want, but you can tweak it here.

wsmoses: I think probably >= 12.6

We can also build different reactant versions for different cuda versions too

giordano:
Uhm, sounds like https://github.com/EnzymeAD/Reactant.jl/commit/eef65ce615a95f02b3b3c3c507f4f856161b7208 was a commit in https://github.com/EnzymeAD/Reactant.jl/pull/497 which got squashed upon merging? 🫠

PR #10279, opened 2025-01-16
Add BVPInterface

ErikQQY:
Add BinaryBuilder recipe for BVPInterface


B/BVPInterface/build_tarballs.jl
```diff
@@ -0,0 +1,236 @@
...
+  LibraryProduct("twpbvp_i32",   :libtwpbvp_i32),
+  LibraryProduct("bvp_m_proxy",  :libbvp_m_proxy),
+]
+
+dependencies = [
+  Dependency("CompilerSupportLibraries_jll")
+]
+
+build_tarballs(ARGS, name, version, sources, script,
+               platforms, products, dependencies; preferred_gcc_version=v"9.1.0")
```
giordano, line 236:
```suggestion
               platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"9.1.0")
```

imciner2:
A few points:

1) I am not thrilled with having the entire build logic be in the Yggdrasil build script. It is usually better and more maintainable to have a build system in the upstream code that is then just called from the build script here.
2) Ideally it should be using `${FC}` to call the Fortran compiler, instead of invoking `gfortran` directly. That ensures the correct compiler executable is always called.
3) Is the only difference between the various cases just the extension of the final library? If so, you can generalize the script by using the `${dlext}` variable to find out the proper extension for the platform. Having all these branches will make maintenance and future changes very tedious/difficult.
4) Use `install -Dvm 755` to move the compiled libraries into their final location.

ErikQQY:
> 1. I am not thrilled with having the entire build logic be in the Yggdrasil build script. It is usually better and more maintainable to have a build system in the upstream code that is then just called from the build script here.

What do you mean by upstream build system in the repo here? The BVPInterface repo has a script to build all the Fortran code there https://github.com/ErikQQY/BVPInterface.jl/blob/master/src/build.jl, and could be used during development.

> 2. Ideally it should be using ${FC} to call the Fortran compiler, instead of invoking gfortran directly. That ensures the correct compiler executable is always called.

Done

> 3. Is the only difference between the various cases just the extension of the final library? If so, you can generalize the script by using the ${dlext} variable to find out the proper extension for the platform. Having all these branches will make maintenance and future changes very tedious/difficult.

Sorry this may be a dumb question, so instead of having different branches for different OS, we can put them together by using the `${dlext}` to automatically find the appropriate OS here?

> 4. Use install -Dvm 755 to move the compiled libraries into their final location.

Done

giordano:
> Sorry this may be a dumb question, so instead of having different branches for different OS, we can put them together by using the ${dlext} to automatically find the appropriate OS here?

Yes: https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables

ErikQQY:
I tink this PR is ready now

ErikQQY:
@giordano This PR should be good to go🤠

B/BVPInterface/build_tarballs.jl
```diff
@@ -0,0 +1,119 @@
...
+    ${FC} -c -fPIC -fdefault-integer-8 -fdefault-real-8 -fdefault-double-8 -w -std=legacy -o ./bvpsol.o ./bvpsol.f
+    ${FC} -c -fPIC -fdefault-real-8 -fdefault-double-8 -w -std=legacy -o ./bvpsol_i32.o ./bvpsol.f
+    ${FC} -c -fPIC -fdefault-integer-8 -fdefault-real-8 -fdefault-double-8 -w -std=legacy -o ./linalg_bvpsol.o ./linalg_bvpsol.f
+    ${FC} -c -fPIC -fdefault-real-8 -fdefault-double-8 -w -std=legacy -o ./linalg_bvpsol_i32.o ./linalg_bvpsol.f
+    ${FC} -c -fPIC -fdefault-integer-8 -fdefault-real-8 -fdefault-double-8 -w -std=legacy -o ./zibconst.o ./zibconst.f
+    ${FC} -c -fPIC -fdefault-real-8 -fdefault-double-8 -w -std=legacy -o ./zibconst_i32.o ./zibconst.f
+    ${FC} -c -fPIC -fdefault-integer-8 -fdefault-real-8 -fdefault-double-8 -w -std=legacy -o ./ma28_bvpsol.o ./ma28_bvpsol.f
+    ${FC} -c -fPIC -fdefault-real-8 -fdefault-double-8 -w -std=legacy -o ./ma28_bvpsol_i32.o ./ma28_bvpsol.f
+    ${FC} -shared -fPIC -o ./bvpsol.${dlext} ./bvpsol.o ./linalg_bvpsol.o ./zibconst.o ./ma28_bvpsol.o
+    ${FC} -shared -fPIC -o ./bvpsol_i32.${dlext} ./bvpsol_i32.o ./linalg_bvpsol_i32.o ./zibconst_i32.o ./ma28_bvpsol_i32.o
```
giordano, line 32: Many tools, like linkers, expect shared libraries to have names of the form `lib...`

B/BVPInterface/build_tarballs.jl
```diff
@@ -0,0 +1,119 @@
+using BinaryBuilder
+
+name = "BVPInterface"
+version = v"0.0.1"
```
giordano, line 4: IF you have to make up a version number, for your own sanity I'd suggest using something like 0.1.0 or 1.0.0.  All versions in the 0.0.x are considered breaking, which is a pain to deal with

B/BVPInterface/build_tarballs.jl
ErikQQY: Agree, changed them to `lib...`

giordano:
Thanks for your contribution!

PR #10280, opened 2025-01-16
[GAP_pkg_semigroups] new package

fingolfin:


fingolfin:
Hrm, fails on FreeBSD (I don't care right now) but also macOS. There, while linking the `libsemigroups` shared library, using GNU libtool.

```
[08:44:29] ld64.lld: warning: Option `-keep_private_externs' is not yet implemented. Stay tuned...
[08:44:29] ld64.lld: warning: Option `-r' is not yet implemented. Stay tuned...
[08:44:29] ld64.lld: warning: Option `-r' is not yet implemented. Stay tuned...
[08:44:29] ld64.lld: error: undefined symbol: main
[08:44:29] >>> referenced by the entry point
[08:44:29] 
[08:44:29] ld64.lld: error: undefined symbol: operator delete(void*)
[08:44:29] >>> referenced by src/.libs/libsemigroups_la-bipart.o:(symbol libsemigroups::Blocks::~Blocks()+0x28)
[08:44:29] >>> referenced by src/.libs/libsemigroups_la-bipart.o:(symbol libsemigroups::Blocks::~Blocks()+0x18)
[08:44:29] >>> referenced by src/.libs/libsemigroups_la-bipart.o:(symbol libsemigroups::Blocks::~Blocks()+0x28)
[08:44:29] >>> referenced 5765 more times
[08:44:29] 
[08:44:29] ld64.lld: error: undefined symbol: _Unwind_Resume

...
```
As far as I can tell, the offending `-keep_private_externs` and `-r` options are introduced by GNU libtool.

Note that this builds fine directly on macOS.


Does the above perhaps ring a bell and you have a quick idea what might be causing this, @giordano @benlorenz ?

giordano:
`clang_use_lld=false`

G/GAP_pkg/GAP_pkg_semigroups/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
...
+platforms = expand_cxxstring_abis(platforms)
+
+# The products that we will ensure are always built
+products = [
+    FileProduct("lib/gap/semigroups.so", :semigroups),
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="1.6", preferred_gcc_version=v"7")
```
giordano, line 47:
```suggestion
               julia_compat="1.6", preferred_gcc_version=v"7", clang_use_lld=false)
```

giordano:
I think I think something (autotools?) misdetects the linker and passes the wrong options but honestly I never went deep down to investigate the issue, it'd be great if someone did.

giordano:
So the issue is the version of libtool we install at https://github.com/JuliaPackaging/Yggdrasil/blob/16f33e945bb8a8d0c03b3e35a8441ee631bab1df/0_RootFS/Rootfs/build_tarballs.jl#L175? Can you please open a ticket with the details of the issue so we can think about it in a dedicated place?

PR #10284, opened 2025-01-16
[BLIS] Update to 2.0

imciner2:
This is a bug fix release of various things since 1.0 last year. They already have a 2.0 RC out actually (which looks pretty cool...)

I have enabled aarch64 freebsd to see what happens. I haven't looked into how RISCV is implemented in the BLIS targets though, so I haven't added that arch.


B/blis/blis/build_tarballs.jl
```diff
@@ -7,6 +7,9 @@ script = blis_script(blis32=false)
 products = [
     LibraryProduct("libblis", :blis)
 ]
- 
+
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
                preferred_gcc_version=v"11", lock_microarchitecture=false, julia_compat="1.6")
+
+
+# Build counter: 1
```
giordano, line 15:
```suggestion
# Build counter: 1

```

imciner2: That seems to be the least of the problems with this...

imciner2:
Well, this is interesting. Apparently what we have here as 1.0 from commit https://github.com/flame/blis/commit/6d0ab74f6975fdf4d19cee06d946b09b6ca89656 is actually just a random commit from after 1.0 was released, since 1.0 is commit https://github.com/flame/blis/commit/49af2243c2a60ed8fedb44f237f4ec100465cd89, and so our current version is closer to 2.0 than 1.0 since they had started to make major architectural changes already (hence why the patches are failing).

giordano:
@jd-foster

jd-foster:
> Well, this is interesting. Apparently what we have here as 1.0 from commit https://github.com/flame/blis/commit/6d0ab74f6975fdf4d19cee06d946b09b6ca89656 is actually just a random commit from after 1.0 was released, since 1.0 is commit https://github.com/flame/blis/commit/49af2243c2a60ed8fedb44f237f4ec100465cd89, and so our current version is closer to 2.0 than 1.0 since they had started to make major architectural changes already (hence why the patches are failing).

Yes, this was noted at the time in #8626
Maintenance of the patches is a little cumbersome, so any ideas to avoid this are welcome.

ViralBShah:
Yeah these patches are annoying to maintain - not hard but just annoying. Perhaps this release can even be skipped and the right thing is to wait for 2.0.

giordano:
> Maintenance of the patches is a little cumbersome, so any ideas to avoid this are welcome.

CC @xrq-phys

B/blis/common.jl
```diff
@@ -80,18 +84,18 @@ function blis_script(;blis32::Bool=false)
...
-        atomic_patch -p1 ${WORKSPACE}/srcdir/patches/armsve_kernels_unscreen_armsve512_int_12xk.patch
-        atomic_patch -p1 ${WORKSPACE}/srcdir/patches/armsve_kernels_unscreen_armsve256_int_8x10.patch
+        #atomic_patch -p1 ${WORKSPACE}/srcdir/patches/armsve_kernels_unscreen_arm_sve_h.patch
+        #atomic_patch -p1 ${WORKSPACE}/srcdir/patches/armsve_kernels_unscreen_armsve512_int_12xk.patch
+        #atomic_patch -p1 ${WORKSPACE}/srcdir/patches/armsve_kernels_unscreen_armsve256_int_8x10.patch
 
         # Screen out A64FX sector cache.
-        patch config/a64fx/bli_cntx_init_a64fx.c ${WORKSPACE}/srcdir/patches/a64fx_config_screen_sector_cache.patch
-    fi
+        #patch config/a64fx/bli_cntx_init_a64fx.c ${WORKSPACE}/srcdir/patches/a64fx_config_screen_sector_cache.patch
```
giordano, line 97: My understanding is that removing these is going to make performance worse

ViralBShah: Yeah I was just seeing if the build goes through.

ViralBShah:
@imciner2 Perhaps it would be best to update this PR to 2.0-RC0 and see how we do in building it. Maybe we can even get our patches upstreamed, if we run into serious issues.

imciner2:
I guess we can try building the 2.0 RC, as long as we don't merge it, since I would prefer if we weren't misaligned on the version.

giordano:
[v2.0](https://github.com/flame/blis/releases/tag/2.0) is out

ViralBShah:
@imciner2 I enabled all the `supported_platforms()`, imagining that it is doing something reasonable on the ones that don't have special treatment.

B/blis/common.jl
```diff
@@ -56,19 +61,19 @@ function blis_script(;blis32::Bool=false)
             export BLI_CONFIG=firestorm
```
imciner2:
```suggestion
           export BLI_CONFIG=arm64
```
There is an armsve config upstream actually, so I wonder if it would be better for us to do arm64 here as well (yes, I know we would be compiling a lot of kernels we wouldn't need), but I think it could be good to have the kernel autodetection to use SVE when available.

B/blis/common.jl
```diff
@@ -56,19 +61,19 @@ function blis_script(;blis32::Bool=false)
...
             export BLI_THREAD=openmp
             ;;
-        *"arm"*"linux"*) 
-            export BLI_CONFIG=arm32
-            export BLI_THREAD=none
+        *"aarch64"*"freebsd"*)
+            export BLI_CONFIG=arm64
+            export BLI_THREAD=openmp
             ;;
         *)
```
imciner2:
```suggestion
           ;;
       *"powerpc64le"*)
           export BLI_CONFIG=power
           export BLI_THREAD=openmp
           ;;
       *)
```
Let's try using the Power target for the powerpc build to at least get Power10 and Power9 kernels instead of just the generic ones.

imciner2 (commented):
I think we can tweak the aarch64 Apple config a bit to use SVE as well, and try the power kernels for powerpc since it is enabled.

There are some riscv configs upstream, but they don't appear to have a meta target, so let's hold off on using them for now and just use the generic there.

B/blis/common.jl
```diff
@@ -56,19 +61,19 @@ function blis_script(;blis32::Bool=false)
...
             # Default (Generic) configuration without optimized kernel.
```
imciner2, line 75:
```suggestion
            # Default (Generic) configuration without optimized kernel.
            # For now, RISC-V uses the generic kernels here until upstream implements a meta target: https://github.com/flame/blis/issues/902
```

ViralBShah:
@imciner2 Maybe push directly to the branch? I just thought I would lend a helping hand to revive your original PR.

imciner2:
The builds are all passing, and it is building the arm64 and powerpc kernels properly on those platforms, so lets merge.

PR #10287, opened 2025-01-17
[mmtk_julia] Bump version to v0.30.3

udesou:


M/mmtk_julia/build_tarballs.jl
```diff
@@ -31,7 +30,7 @@ platforms = [
 
 # The products that we will ensure are always built
 products = [
-    LibraryProduct("libmmtk_julia", :libmmtk_julia)
+    LibraryProduct("libmmtk_julia", :libmmtk_julia; dont_dlopen=true)
```
giordano, line 33: I added this because otherwise we need to always manually merge PRs to `master`.  As far as I understand you don't plan to use this package standalone but only within julia itself, right?

udesou: That's right. Awesome, thanks for fixing it!

PR #10289, opened 2025-01-17
Reactant 0.0.42

wsmoses:


wsmoses:
@giordano assuming you awake before me, feel free to add the extra cuda builds here too if it's not done by the time you awake (this JLL currently adds profiling apis)

wsmoses:
for now probably a lower 12 such that cyclops works, and maybe a cuda 11? I'm fine for now just picking based off of "making sure machines we access work"

PR #10293, opened 2025-01-18
MAGEMin v1.6.5

NicolasRiel:
- Deactivated pure endmembers swap for epidote as this was leading to mistakenly stabilizing epidote outside the feasible domain. This results in significant improvement of the phase diagram for hydrous systems at low pressure and low temperature conditions for metapelite, metabasite and igneous database.
- Corrected phase_info() function that was displaying wrong phase names for pure phases


NicolasRiel:
@giordano thank you! :D

PR #10295, opened 2025-01-18
[ITKIOWrapper-v2.0.0] Added support for mac OS

divital-coder:
-Upstream package, wrapping functions for I/O functionality from NIFTI and DICOM datasets from the Insight Toolkit library.


divital-coder:
Hello apologies for holding this PR.
This cold kinda knocked me out for a bit too long.

I/ITKIOWrapper/build_tarballs.jl
```diff
@@ -20,40 +16,53 @@ script = raw"""
...
+    BuildDependency(PackageSpec(name="Eigen_jll", uuid="bc6bbf8a-a594-5541-9c57-10b0d0312c70")),
+    Dependency(PackageSpec(name="ITK_jll", uuid="3324d3a8-621a-5aaa-97fa-c3bc8dfc0481"); compat="5.3.1"),
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"); compat = "0.13.2"),
+    BuildDependency(PackageSpec(name="libjulia_jll", uuid="5ad3ddd2-0711-543a-b040-befd59781bbf"))
 ]
 
-# Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"8.1.0")
+# Build the tarballs
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.7", preferred_gcc_version=v"8.1.0",
```
giordano, line 67: You're still building for julia v1.6 though.

I/ITKIOWrapper/build_tarballs.jl
```diff
@@ -20,40 +16,53 @@ script = raw"""
...
+preferred_llvm_version=v"13.0.1")
```
giordano, line 68: why llvm 13?

I/ITKIOWrapper/src/CMakeLists.txt
```diff
@@ -36,8 +49,28 @@ message(STATUS "Found ITK at ${ITK_USE_FILE}")
...
+        INSTALL_RPATH "@loader_path"
+        BUILD_WITH_INSTALL_RPATH TRUE
+        INSTALL_NAME_DIR "@rpath"
+    )
+endif()
+
+install(TARGETS ITKIOWrapper 
+    LIBRARY DESTINATION lib
+    RUNTIME DESTINATION lib
+)
```
giordano, line 76: Missing newline
```suggestion
)

```

I/ITKIOWrapper/build_tarballs.jl
divital-coder: oh based on this : 
```
```

I/ITKIOWrapper/build_tarballs.jl
divital-coder: macos llvm version 13 built successfully on my intel Mac locally so provided that as a default here.

I/ITKIOWrapper/build_tarballs.jl
giordano: That doesn't do whatever you may think it does. Just look at the list of platforms queued up at https://buildkite.com/julialang/yggdrasil/builds/17430.  The list of julia versions you're building for comes from the line below:
https://github.com/JuliaPackaging/Yggdrasil/blob/a3b2ad83c19b25270c6f6dd79b6a77714c6b63db/I/ITKIOWrapper/build_tarballs.jl#L11-L12 You need to filter `julia_versions` to take only Julia >= 1.7.  Also, any specific reason for still supporting v1.7 and not, say, require Julia v1.10 (the current LTS) at this point?  Consider that having lots of Julia versions supported means loads of more platforms to build for, much longer build times, increased probability of failures, etc.

I/ITKIOWrapper/build_tarballs.jl
giordano: Let's change the question: what's the problem with LLVM 18?

I/ITKIOWrapper/build_tarballs.jl
divital-coder: Julia 1.10 LTS sounds plausible, I was manually testing each iteration of version bumped for julia_compat to eliminate UUID errors related to OpenSSL_jll, I will skip to that as a default.

I/ITKIOWrapper/build_tarballs.jl
divital-coder: I was following sort of like a boilerplate from available package recipes, now that you mention it I will consider that.

divital-coder:
Any directions to remedy this?
```
ERROR: LoadError: KeyError: key UUID("458c3c95-2e84-50aa-8efc-19380b2a3a95") not found
Stacktrace:
```

I/ITKIOWrapper/build_tarballs.jl
```diff
@@ -1,59 +1,75 @@
...
+        -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+        -DCMAKE_OSX_DEPLOYMENT_TARGET=10.10
+else
+    cmake -B build -S . \
+        -DCMAKE_INSTALL_PREFIX=${prefix} \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+        -DBUILD_SHARED_LIBS:BOOL=ON \
+        -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+fi
```
giordano, line 41:
```suggestion
if [[ "${target}" == x86_64-apple-* ]]; then
    MACOSX_DEPLOYMENT_TARGET=10.10
fi
cmake -B build -S . \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DBUILD_SHARED_LIBS:BOOL=ON \
    -DCMAKE_POSITION_INDEPENDENT_CODE=ON
```
but I'm missing why you need to set the SDK to 10.10 at all, since we're using 10.12, an explanation would be most most welcome.

divital-coder: Setting it to 10.10 ensures broader compatibility with older macOS  and a lower deployment target allows the binary to run on older versions

giordano: [Julia itself requires macOS 10.14](https://julialang.org/downloads/#supported_platforms), so I'm not sure what you gain by doing that.

divital-coder: interesting, so removing the condition at all will certainly set the default to 10.12 I believe, I will try that locally, and commit changes if i succeed.

giordano: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/bc1c293e367714537136d22ece9d81a82731b009/src/Runner.jl#L116-L145 Actually it may be 10.10 _already_.  Either way, I don't think you need to set anything older than whatever is already the (old) default.

divital-coder:
done :)

I/ITKIOWrapper/build_tarballs.jl
```diff
@@ -24,36 +25,39 @@ cmake -B build -S . \
...
+    Dependency(PackageSpec(name="ITK_jll", uuid="3324d3a8-621a-5aaa-97fa-c3bc8dfc0481"); compat="5.3.1"),
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"); compat = "0.13.4"),
+    BuildDependency(PackageSpec(name="libjulia_jll", uuid="5ad3ddd2-0711-543a-b040-befd59781bbf"))
 ]
 
-# Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"8.1.0")
+# Build the tarballs
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.10", preferred_gcc_version=v"8.1.0",
+preferred_llvm_version=v"18")
```
giordano, line 62:
```suggestion
)
```

I/ITKIOWrapper/src/CMakeLists.txt
```diff
@@ -2,8 +2,19 @@ cmake_policy(SET CMP0025 NEW)
 project(ITKIOWrapper)
 
 cmake_minimum_required(VERSION 3.14)
-set(CMAKE_CXX_STANDARD 17)
+
+# Add macOS specific settings
+if(APPLE)
+    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.10" CACHE STRING "Minimum OS X deployment version")
```
giordano, line 8: Ah, you were also setting this here for good measure.

I/ITKIOWrapper/src/CMakeLists.txt
giordano, line 76:
```suggestion
)

```

divital-coder:
done :)

PR #10299, opened 2025-01-19
[ITK-v5.3.1] ommision of itk minc lib copying resulted in upstream errors

divital-coder:
While working on the upstream packages, specifically for the windows platforms, 
I carefully anaylsed the successful PR from last year, and realised the following lines were ommitted for a crucial library : 
```
    cp $prefix/lib/libitkminc2-5.3.dll $prefix/bin
    cp $prefix/lib/libitkminc2-5.3.dll.a $prefix/bin
```

Which are hereby added in this PR .

### DIFF
--- a/I/ITK/build_tarballs.jl
+++ b/I/ITK/build_tarballs.jl
@@ -5,7 +5,7 @@ version = v"5.3.1"
 sources = [
     GitSource("https://github.com/InsightSoftwareConsortium/ITK.git", "1fc47c7bec4ee133318c1892b7b745763a17d411")
 ]
-# Bash recipe for building across all platforms
+# Bash recipe for building across all Platforms
 script = raw"""
 if [[ "${target}" == *x86_64-w64-mingw32* ]]; then
     CONFIG=msys2-64
--- a/I/ITK/build_tarballs.jl
+++ b/I/ITK/build_tarballs.jl
@@ -55,6 +55,8 @@ install_license ${WORKSPACE}/srcdir/ITK/LICENSE
 if [[ "${target}" == *x86_64-w64-mingw32* ]]; then
 mkdir -pv ${libdir}
 find "${prefix}/lib" -name "*.${dlext}" -exec mv -v {} ${libdir} \;
+cp $prefix/lib/libitkminc2-5.3.dll $prefix/bin
+cp $prefix/lib/libitkminc2-5.3.dll.a $prefix/bin
 fi
 """
 # These are the platforms we will build for by default, unless further
--- a/I/ITK/build_tarballs.jl
+++ b/I/ITK/build_tarballs.jl
@@ -53,7 +53,6 @@ cmake --install build
 install_license ${WORKSPACE}/srcdir/ITK/LICENSE
 
 if [[ "${target}" == *x86_64-w64-mingw32* ]]; then
-mkdir -p ${prefix}/bin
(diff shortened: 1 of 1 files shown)


divital-coder:
This is in support for the continuation of this paricular library related error on mingw based systems : 
log : 
```
[02:10:11] CMake Error at /workspace/destdir/lib/cmake/ITK-5.3/ITKTargets.cmake:815 (message):
[02:10:11]   The imported target "itkminc2" references the file
[02:10:11] 
[02:10:11]      "/workspace/destdir/lib/libitkminc2-5.3.dll"
[02:10:11] 
[02:10:11]   but this file does not exist.  Possible reasons include:
[02:10:11] 
[02:10:11]   * The file was deleted, renamed, or moved to another location.
[02:10:11] 
[02:10:11]   * An install or uninstall procedure did not complete successfully.
[02:10:11] 
[02:10:11]   * The installation package was faulty and contained
[02:10:11] 
[02:10:11]      "/workspace/destdir/lib/cmake/ITK-5.3/ITKTargets.cmake"
[02:10:11] 
[02:10:11]   but not all the files it references.
```

I/ITK/build_tarballs.jl
```diff
@@ -53,6 +53,8 @@ cmake --install build
 install_license ${WORKSPACE}/srcdir/ITK/LICENSE
 
 if [[ "${target}" == *x86_64-w64-mingw32* ]]; then
+cp $prefix/lib/libitkminc2-5.3.dll $prefix/bin
+cp $prefix/lib/libitkminc2-5.3.dll.a $prefix/bin
```
giordano, line 57: Why?  Import libraries should be in `${prefix}/lib`, not `${libdir}`.

divital-coder: Yes indeed, its the same review on the last pr as well i believe, in this case copying libitkminc is the most optimal solution I could find, and also this was present in the previous original version of itk merged pr as well.

divital-coder: Also, I manually checked the availability of itkminc-5.3 and it is present in lib on mingw systems but, the error still persisted, to which i added the above lines.

divital-coder: https://discourse.itk.org/t/error-in-itk-compilation-with-wingw32-make/1122/2

giordano: I still don't believe this is right, but I also don't have the energy/will to argue further.

divital-coder: """"
MinGW is neither officially supported, nor tested, so it is not surprising that it doesn't work. Contributions to add support are welcome.

Microsoft offers just the command-line compiler, without needing Visual Studio. You might also have better luck with Clang for Windows than MinGW.

""'"
https://stackoverflow.com/questions/74576002/unable-to-build-itk-5-version-on-gcc-mingw-10-pthread-linking-error-win

Just citing some stuff here 🙃

giordano: I have no clue of how any of this is related to the fact you're moving files in the wrong directory, but as I said above I don't have the energy to keep arguing.

divital-coder: Sorry, I was citing stuff and in between chat updated and your reviews appeared, pretty hectic day for you I suppose , 
ok, I will perhaps dig in more in see if I can get it to work on libdir.
Thanks.

PR #10304, opened 2025-01-20
[GR] bump version

jheinen:
bump run-time version to 0.73.12


jheinen:
@giordano : Thanks a lot!

PR #10306, opened 2025-01-21
[GAP_lib] Backport new package directories features

lgoettgens:
cf. https://github.com/oscar-system/GAP.jl/pull/1095#issuecomment-2598465768

cc @fingolfin

### DIFF
--- a/G/GAP_lib/build_tarballs.jl
+++ b/G/GAP_lib/build_tarballs.jl
@@ -22,7 +22,7 @@ using BinaryBuilder, Pkg
 
 name = "GAP_lib"
 upstream_version = v"4.14.0"
-version = v"400.1400.002"
+version = v"400.1401.000"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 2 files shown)


G/GAP_lib/build_tarballs.jl
```diff
@@ -22,7 +22,7 @@ using BinaryBuilder, Pkg
 
 name = "GAP_lib"
 upstream_version = v"4.14.0"
-version = v"400.1400.002"
+version = v"400.1401.000"
```
lgoettgens, line 25:
```suggestion
version = v"400.1400.003"
```
I am not sure which of the two versions better matches the GAP_lib_jll versioning scheme. @fingolfin please decide and either apply the patch or don't.

fingolfin: Either is fine by me, as this change is fully backwards compatible. For more risky changes I'd make sure to increment the minor version to ensure nobody gets the changes without also updating GAP.jl. Here either should be fine...

That said I guess the slightly safer approach is to use 1401 as you do, again to ensure nobody out there gets these changes before we make a new GAP.jl release

fingolfin (approved):

PR #5660, opened 2022-10-12
[libosmium] Add project

jeremiahpslewis:

PR #10307, opened 2025-01-21
[libosmium] Bump boost to v1.87 to pick up aarch64-unknown-freebsd platform

jeremiahpslewis:


L/libosmium/build_tarballs.jl
```diff
@@ -39,7 +39,7 @@ dependencies = [
     BuildDependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a")),
     Dependency("Expat_jll"; compat="2.2.10"),
     Dependency("Bzip2_jll"; compat="1.0.8"),
-    Dependency("boost_jll"; compat="=1.76.0"),
+    Dependency("boost_jll"; compat="=1.79.0"),
```
giordano, line 42: May be worth building a new boost to support riscv64 too

jeremiahpslewis: @giordano looks like that's already ✅ https://github.com/JuliaPackaging/Yggdrasil/pull/10119

giordano: No, there's not: https://github.com/JuliaRegistries/General/pull/122129

L/libosmium/build_tarballs.jl
```diff
@@ -39,7 +39,7 @@ dependencies = [
     BuildDependency(PackageSpec(name="Zlib_jll", uuid="83775a58-1f1d-513f-b197-d71354ab007a")),
     Dependency("Expat_jll"; compat="2.2.10"),
     Dependency("Bzip2_jll"; compat="1.0.8"),
-    Dependency("boost_jll"; compat="=1.76.0"),
+    Dependency("boost_jll"; compat="=1.87.0"),
```
giordano, line 42: Now that I'm looking at this more carefully, this is a header-only file, so I have the same comment as this one: https://github.com/JuliaPackaging/Yggdrasil/pull/5660#discussion_r993955189

PR #10309, opened 2025-01-21
[GAP_pkg_*] Rebuild for GAP_lib_jll 400.1401.0

lgoettgens:
Since we bumped the minor version in https://github.com/JuliaPackaging/Yggdrasil/pull/10306, all of its dependents need to get rebuilt for this new version...
We should've decided to instead release https://github.com/JuliaPackaging/Yggdrasil/pull/10306 as a patch...

cc @fingolfin

### DIFF
--- a/G/GAP_pkg/GAP_pkg_browse/build_tarballs.jl
+++ b/G/GAP_pkg/GAP_pkg_browse/build_tarballs.jl
@@ -3,10 +3,10 @@
 include("../common.jl")
 
 gap_version = v"400.1400.0"
-gap_lib_version = v"400.1400.0"
+gap_lib_version = v"400.1401.0"
 name = "Browse"
 upstream_version = "1.8.21" # when you increment this, reset offset to v"0.0.0"
-offset = v"0.0.1" # increment this when rebuilding with unchanged upstream_version, e.g. gap_version changes
+offset = v"0.0.2" # increment this when rebuilding with unchanged upstream_version, e.g. gap_version changes
 version = offset_version(upstream_version, offset)
 
 # Collection of sources required to build this JLL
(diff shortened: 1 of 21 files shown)


fingolfin:
Alternatively edit GAP_lib_jll to release 400.1400.3  (or what the number is). The two releases can coexist, we just don't use GAP_lib_jll 400.1401.0

lgoettgens:
> Alternatively edit GAP_lib_jll to release 400.1400.3 (or what the number is). The two releases can coexist, we just don't use GAP_lib_jll 400.1401.0

As there are some build failures here, I will proceed as suggested.

G/GAP_pkg/GAP_pkg_browse/build_tarballs.jl
```diff
@@ -3,10 +3,10 @@
 include("../common.jl")
 
 gap_version = v"400.1400.0"
-gap_lib_version = v"400.1400.0"
+gap_lib_version = v"400.1401.0"
```
fingolfin, line 6: So `setup_gap_package` does this right now:
```julia    dependencies = BinaryBuilder.AbstractDependency[
        Dependency("GAP_jll", gap_version; compat="~$(gap_version)"),
        Dependency("GAP_lib_jll", gap_lib_version; compat="~$(gap_lib_version)"),
    ]
```
I now wonder why I used `~` instead of `^` here. Maybe had a good reason but I should have written it in there as a comment :-(. As it is, it makes some sense to depend on the `GAP_jll` somewhat closely (to avoid ABI change issue), but a change to `GAP_lib_jll` should be fine in most cases? To the extent that I wonder why we specify a dependency on `GAP_lib_jll` at all, a `BuildDependency` should be enough (if we even need that) ...

lgoettgens:
> Alternatively edit GAP_lib_jll to release 400.1400.3 (or what the number is). The two releases can coexist, we just don't use GAP_lib_jll 400.1401.0

See https://github.com/JuliaPackaging/Yggdrasil/pull/10310/

G/GAP_pkg/GAP_pkg_browse/build_tarballs.jl
lgoettgens: These are indeed good questions, but I am afraid that I don't know enough about the gap build system / gap package building

PR #10311, opened 2025-01-21
[GAP_pkg_*] change GAP_lib_jll to a BuildDependency

fingolfin:
Logically the code in these packages only depends on the GAP kernel ABI, and hence GAP_jll, but not on the library code in GAP_lib_jll. Decoupling these make it easier to perform various upgrades. See also #10309 for some background.

This changes has no effect yet, but the next time we update any of the GAP_pkg_* JLLs we will see the benefit.

CC @lgoettgens @ThomasBreuer


lgoettgens (approved):
sounds good

PR #10315, opened 2025-01-22
[Reactant] Build for CUDA 12.3

giordano:


R/Reactant/build_tarballs.jl
```diff
@@ -63,6 +63,9 @@ BAZEL_FLAGS+=(--server_javabase=$JAVA_HOME)
 
 BAZEL_BUILD_FLAGS+=(--jobs ${nproc})
 
+# Use ccache to speedup re-builds
+BAZEL_BUILD_FLAGS+=(--action_env=USE_CCACHE=${USE_CCACHE})
```
giordano, line 67: This should finally make `ccache` work and hopefully save time during rebuilds with hot cache.

R/Reactant/build_tarballs.jl
```diff
@@ -312,7 +315,7 @@ augment_platform_block="""
     """
 
 # for gpu in ("none", "cuda", "rocm"), mode in ("opt", "dbg"), platform in platforms
-for gpu in ("none", "cuda"), mode in ("opt", "dbg"), cuda_version in ("none", "12.1", "12.6"), platform in platforms
+for gpu in ("none", "cuda"), mode in ("opt", "dbg"), cuda_version in ("none", "12.1", "12.3", "12.6"), platform in platforms
```
giordano, line 318: Is this what you suggested in https://github.com/LuxDL/Lux.jl/issues/1225#issuecomment-2606017095?

wsmoses: mostly, but I think we can drop 12.1 even

R/Reactant/build_tarballs.jl
wsmoses: ooh nice

R/Reactant/platform_augmentation.jl
```diff
@@ -93,8 +93,10 @@ function augment_platform!(platform::Platform)
             Libdl.dlclose(handle)
 
             if cuda_version_tag == "none" && current_cuda_version isa VersionNumber
-                if v"12.1" <= current_cuda_version < v"12.6"
+                if v"12.1" <= current_cuda_version < v"12.3"
                     cuda_version_tag = "12.1"
+                elseif v"12.3" <= current_cuda_version < v"12.6"
```
wsmoses, line 98: I was kind of suggesting to do

if VERSION > v12.3
  cuda_version_tag = "12.6"
elseif VERSION > v12
  cuda_version_tag = "12.1"
else..

giordano: > if VERSION > v12.3
> cuda_version_tag = "12.6"

Wait, I find that that confusing. Are you suggesting that if system has CUDA later than 12.3 (so 12.4+) then we use our 12.6 build? I'm not sure that's going to work, but maybe I'm misunderstanding something.

wsmoses: it's possible yeah, but as we saw elsewhere the insides of XLA actually does check dynamically the cuda version. However, if ver >= 12.3then it links against something that makes loading impossible. So we should put something at that boundary point

giordano: Billy approved offline

PR #10319, opened 2025-01-22
[MKL] Restore the symlinks

amontoison:


M/MKL/build_tarballs.jl
```diff
@@ -21,8 +21,13 @@ if [[ ${target} == *x86_64-w64-mingw* ]]; then
...
 if [[ ${target} == *x86_64-linux-gnu* ]]; then
     install -Dvm 755 mkl-${target}/mkl-*.data/data/lib/* -t "${libdir}"
+    cd mkl-${target}/mkl-*.data/data/lib
+    for lib in *.so.2; do
+        target="${lib%.2}"
+        ln -s "${libdir}/$lib" "${libdir}/$target"
+    done
 fi
-install_license mkl-${target}/mkl-*.dist-info/LICENSE.txt
+install_license $WORKSPACE/srcdir/mkl-${target}/mkl-*.dist-info/LICENSE.txt
```
amontoison, line 30:
```suggestion
cd $WORKSPACE/srcdir
install_license mkl-${target}/mkl-*.dist-info/LICENSE.txt
```

M/MKL/build_tarballs.jl
```diff
@@ -21,7 +21,13 @@ if [[ ${target} == *x86_64-w64-mingw* ]]; then
...
 if [[ ${target} == *x86_64-linux-gnu* ]]; then
     install -Dvm 755 mkl-${target}/mkl-*.data/data/lib/* -t "${libdir}"
+    cd mkl-${target}/mkl-*.data/data/lib
+    for lib in *.so.2; do
+        target="${lib%.2}"
+        ln -s "${libdir}/$lib" "${libdir}/$target"
+    done
 fi
+cd $WORKSPACE/srcdir
 install_license mkl-${target}/mkl-*.dist-info/LICENSE.txt
```
amontoison, line 31:
```suggestion
install_license mkl-${target}/mkl-2025.0.1.dist-info/LICENSE.txt
```

imciner2:
Huh, we also apparently lost the `locale` folder in the `lib` directory when we upgraded. This must be because we switched from using the Intel conda channel (which no longer exists) to the conda-forge repackaging, which makes editorial decisions about what to include. I don't know if that will make a difference though.

imciner2:
This is probably fine. We are actually creating more symbolic links than there were originally (older versions didn't link the `libmkl_vml*` and the arch-specific MKL libraries), but those links shouldn't pose a problem.

imciner2 (approved):

amontoison:
Thanks Ian for the comments!

PR #10325, opened 2025-01-23
[EDFlib] Update to v1.26.0

ararslan:
Also include the header as a file product, just in case.

### DIFF
--- a/E/EDFlib/build_tarballs.jl
+++ b/E/EDFlib/build_tarballs.jl
@@ -1,26 +1,28 @@
 using BinaryBuilder
 
 name = "EDFlib"
-version = v"1.16.0"
+version = v"1.26.0"
 
 sources = [
-    ArchiveSource("https://www.teuniz.net/edflib/edflib_116.tar.gz",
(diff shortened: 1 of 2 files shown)


E/EDFlib/build_tarballs.jl
```diff
@@ -1,29 +1,32 @@
...
+                  "e9e37aa561fa094cb759b4da6d4741f0092d7851a375ee877c18f993150443a8";
                   unpack_target="edflib")
 ]
 
 script = raw"""
 cd ${WORKSPACE}/srcdir/edflib
-mkdir ${libdir}
+mkdir -p ${libdir}
+mkdir -p ${includedir}
 ${CC} edflib.c -shared -fPIC -D_LARGEFILE64_SOURCE -D_LARGEFILE_SOURCE -o ${libdir}/libedflib.${dlext}
```
ararslan, line 16: Looks like there's almost sort of a Makefile now: https://gitlab.com/Teuniz/EDFlib/-/blob/master/lib/makefile?ref_type=heads. Seems to be effectively this, but one notable difference is that the library there is called `libedf` rather than `libedflib`. Perhaps I should change that here too.

PR #10327, opened 2025-01-23
[OpenFHE] and [OpenFHE_int128] rebuild for riscv64

ArseniyKholod:


giordano:
Unfortunately you need to change the version number of these package to do this rebuild, because due to recent changes in BinaryBuilder some compat bounds will change.  My suggestion is to close this PR and for the time being skip riscv64 in #10320

ArseniyKholod:
> Unfortunately you need to change the version number of these package to do this rebuild, because due to recent changes in BinaryBuilder some compat bounds will change. My suggestion is to close this PR and for the time being skip riscv64 in #10320. Next time there's an upstream version of these package you should hopefully get riscv64 builds in

Thanks! I will do so)

PR #10329, opened 2025-01-23
[openfhe_julia] and [openfhe_julia_int128] update to version v0.3.8

ArseniyKholod:


giordano:
Now this build is gone, but I see that [`OpenFHE.jl` requires Julia v1.8](https://github.com/hpsc-lab/OpenFHE.jl/blob/2911a328c990b13cbbe0c7fbe20021434c2a4f30/Project.toml#L17), so for the next time it'd be good to stop working these libraries for Julia v1.6-1.7, which would reduce the number of binary blobs we produce, and shorten the waiting time for the other PRs.

Also, you want to consider requiring Julia v1.10 in `OpenFHE.jl` now that that's the new LTS version.

PR #10337, opened 2025-01-24
[Aeron] Update to v1.47.0

DarrylGamroth:
Update Aeron to v1.47.0

- Include support for macOS Intel
- Add Aeron Archive client library
- Add Media Driver library


A/Aeron/build_tarballs.jl
```diff
@@ -3,79 +3,88 @@
...
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(
     ARGS, name, version, sources, script, platforms, products, dependencies;
     julia_compat="1.6",
-    preferred_gcc_version = v"11.1.0",
-)
+    preferred_gcc_version=v"11",
+)
```
giordano, line 90:
```suggestion
)

```

A/Aeron/bundled/patches/win32/mingw.patch
giordano, line 1: These patches should probably be upstreamed

DarrylGamroth: Agreed. I haven't tested that it still works under MSVC that's supported by upstream, I can just verify that it functions on mingw. I can remove it for this pull request.

PR #10338, opened 2025-01-24
[SuiteSparse Extras] Re-run builders for SPEX, ParU, GraphBLAS

rayegun:
Re-run these so that they can appear on General and then I'll get them merged. Will only function on v1.12

### DIFF
--- a/S/SuiteSparse/ParU/build_tarballs.jl
+++ b/S/SuiteSparse/ParU/build_tarballs.jl
@@ -33,4 +33,4 @@ products = [
 
 
 build_tarballs(ARGS, name, version, sources, script, platforms, 
-               products, dependencies; julia_compat="1.11",preferred_gcc_version=v"9")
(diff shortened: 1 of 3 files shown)


S/SuiteSparse/ParU/build_tarballs.jl
```diff
@@ -33,4 +33,4 @@ products = [
 
 
 build_tarballs(ARGS, name, version, sources, script, platforms, 
-               products, dependencies; julia_compat="1.11",preferred_gcc_version=v"9")
+               products, dependencies; julia_compat="1.12",preferred_gcc_version=v"9")
```
giordano, line 36: You can't change compat bound without changing version number.  You aren't changing anything here, why doing this?

giordano: Discussed this offline: https://github.com/JuliaRegistries/General/pull/118860 wasn't actually merged, so changing this compat bound is safe.

PR #10352, opened 2025-01-27
[Faiss] Replaced MKL and OpenBLAS with libblastrampoline

stemann:
Due to issue with linking with MKL: https://github.com/JuliaPackaging/Yggdrasil/pull/8572#issuecomment-2614600730

Also, removes the need to build with a newer GCC (than v7) wrt. OpenBLAS on PowerPC64LE.

### DIFF
--- a/F/Faiss/bundled/patches/faiss-cmake-mkl-optional.patch
+++ b/F/Faiss/bundled/patches/faiss-cmake-mkl-optional.patch
@@ -1,0 +1,27 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index c4fea46a..12f8aef5 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -59,6 +59,7 @@ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+ # Valid values are "generic", "avx2", "avx512", "sve".
+ option(FAISS_OPT_LEVEL "" "generic")
+ option(FAISS_ENABLE_GPU "Enable support for GPU indexes." ON)
++option(FAISS_ENABLE_MKL "Enable MKL." ON)
+ option(FAISS_ENABLE_RAFT "Enable RAFT for GPU indexes." OFF)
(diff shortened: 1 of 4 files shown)


F/Faiss/common.jl
```diff
@@ -11,12 +11,21 @@ apk del cmake
...
 
 cmake_extra_args=()
 
+if [[ "$bb_full_target" != armv6l-linux-* ]]; then
+    libblastrampoline=$(find $libdir -name libblastrampoline'*'.$dlext) # Enable Windows builds to find libblastrampoline
+    cmake_extra_args+=(
+        "-DBLAS_LIBRARIES=$libblastrampoline"
+        "-DLAPACK_LIBRARIES=$libblastrampoline"
+    )
+fi
```
stemann, line 27: @giordano Is there a nicer way to make CMake recognize lbt as BLAS and LAPACK libraries?

imciner2: If upstream actually uses the proper CMake modules for BLAS/LAPACK, then you can just copy what SuiteSparse does and use the newest CMake and pass a single variable: https://github.com/JuliaPackaging/Yggdrasil/pull/9207/files

F/Faiss/bundled/patches/faiss-cmake-mkl-optional.patch
```diff
@@ -0,0 +1,27 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index c4fea46a..12f8aef5 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -59,6 +59,7 @@ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+ # Valid values are "generic", "avx2", "avx512", "sve".
+ option(FAISS_OPT_LEVEL "" "generic")
+ option(FAISS_ENABLE_GPU "Enable support for GPU indexes." ON)
++option(FAISS_ENABLE_MKL "Enable MKL." ON)
```
stemann, line 9: Upstreamed: https://github.com/facebookresearch/faiss/pull/4147

PR #10354, opened 2025-01-27
MAGEMin v1.6.6

NicolasRiel:
- Added option to deactivate pure phase
- Added information about end-member and solution phases abbreviations and compositions
- Corrected unit output for entropy and enthalpy for pure and solution phases


NicolasRiel:
@giordano Thank you very much! :))

PR #10355, opened 2025-01-27
[oxigraph_server] Bump to v0.4.7

jeremiahpslewis:


O/oxigraph_server/build_tarballs.jl
```diff
@@ -35,6 +35,8 @@ platforms = filter(x -> cxxstring_abi(x) != "cxx03", platforms)
 # Rust toolchain for i686 Windows is unusable
 filter!(p -> !Sys.iswindows(p) || arch(p) != "i686", platforms)
 
+filter!(p -> arch(p) != "riscv64", platforms)
```
giordano, line 38: Why?

jeremiahpslewis: We use prebuilt binaries and they don't produce one for riscv

PR #10313, opened 2025-01-21
[Reactant] Build for aarch64-linux with CUDA

giordano:
This is only to save my progress along the way, nothing works at the moment.

PR #10356, opened 2025-01-28
Downdeploy

wsmoses:
idk why my local build hates me


giordano:
Let's continue in #10313, I removed the upgrade of the SDK, not necessary with https://github.com/EnzymeAD/Reactant.jl/pull/643

PR #10359, opened 2025-01-28
[open62541] update to version 1.4.10

thomvet:
Should be a straightforward patch version bump.

### DIFF
--- a/O/open62541/open62541@1.4/build_tarballs.jl
+++ b/O/open62541/open62541@1.4/build_tarballs.jl
@@ -3,12 +3,12 @@
 using BinaryBuilder, Pkg
 
 name = "open62541"
-version = v"1.4.9"
+version = v"1.4.10"
 
 # Collection of sources required to complete build
 sources = [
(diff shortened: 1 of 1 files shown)


thomvet:
Will require a patch, see here: https://github.com/open62541/open62541/issues/7034

thomvet:
Reading in the related issues some more, I am going to wait for v 1.4.10 and then update this PR.

thomvet:
PR updated to 1.4.10 that came out yesterday. Let's see whether builds work.

thomvet:
This seems unrelated to the build itself? The errors happen before the buildscript even starts?

O/open62541/open62541@1.4/build_tarballs.jl
```diff
@@ -62,5 +66,5 @@ dependencies = [
     Dependency("OpenSSL_jll"; compat="3.0.8")
```
giordano, line 66:
```suggestion
    Dependency("OpenSSL_jll"; compat="3.0.15")
```

PR #10373, opened 2025-01-30
[Reactant] Enable ccache

giordano:
I want to see how effective it is now that we set `CCACHE_DIR` correctly, with that locally I get nearly 100% hits on rebuilds, but not all compiler calls seem to go through ccache, definitely not `nvcc` ones, but I suspect there's also something else


giordano:
There are still way too many misses here, I don't understand it 😞 I can't tell what would explain the difference with what I see locally.  The hit rate is so bad here that this is probably not worth it.

giordano:
This is really not working well enough.  I tried to cleanup the cache on the build machine, in case there were some problems, but we're still missing the cache way more times than expected (locally cache hit is extremely good instead), and I don't want to spend more time investigating the issue here.

PR #10374, opened 2025-01-30
[MLX] Updated to v0.22.1

stemann:


stemann:
Currently just validating changes to MLX upstream to see if stuff still builds.

ararslan:
It looks like the FreeBSD compilers are still 13.2:
```
[02:18:53] -- Check for working C compiler: /opt/bin/x86_64-unknown-freebsd13.2-libgfortran5-cxx11/x86_64-unknown-freebsd13.2-clang - skipped
```
That would likely explain why the same error is still occurring.

giordano:
The paths don't magically change, they come from BinaryBuilderBase.jl which believes we use 13.2, it doesn't know what's going on inside the recipe.

stemann:
Just checked as well, the error that is left now is not the "C++-17 on FreeBSD 13.2 issue", cf. #10402, but just the one being addressed in https://github.com/ml-explore/mlx/issues/1823 .

stemann:
@giordano This one should be good to go :-)

M/MLX/bundled/patches/freebsd-backend-cpu-quantized.patch
giordano, line 1: Upstream?

stemann: It already is :-) https://github.com/ml-explore/mlx/pull/1890

stemann: OK - I'll try to remember it next time.

giordano: Making it obvious that you already did something makes review easier, quicker, saves question and back and forth

stemann: Yes, I know.

PR #10378, opened 2025-01-30
[OCaml] Upgrade to v5.3.0

topolarity:


giordano:
We recently changed some compat bounds which means the version number of this package must change.  I see our version (4.10.0) is quite old, there's [4.10.1](https://github.com/ocaml/ocaml/releases/tag/4.10.1) if you don't want to jump too much, otherwise latest version is [5.3.0](https://github.com/ocaml/ocaml/releases/tag/5.3.0)

topolarity:
Looks like Ocaml is dropping 32-bit support, so we're down to just `x86_64-linux-gnu/musl`

Keno:
The other issue is that if we want to build ocaml libraries, we need an ocaml cross compiler and unfortunately it has a GCC style with separate cross-compiler binaries for each architecture. I don't know if we have a good way to do that with current Yggdrasil.

Keno:
@staticfloat I'm assuming BB2 would make that kind of thing easier?

topolarity:
FWIW @giordano this is ready to go from my side

PR #10380, opened 2025-01-30
[opam] Add OCaml package manager 2.3.0

mlechu:


giordano (commented):
Thanks for your contribution!  A couple of changes to make future updates smoother by removing hard-coded version numbers

O/opam/build_tarballs.jl
```diff
@@ -0,0 +1,33 @@
...
+    Platform("x86_64", "linux"; libc="glibc"),
+    Platform("x86_64", "linux"; libc="musl")
+]
+
+products = [
+    ExecutableProduct("opam", :opam),
+]
+
+dependencies = [
+    BuildDependency(PackageSpec(name="OCaml_jll", version=v"5.3.0"))
```
Keno, line 29: Leaving a note here that the reason this was required to be a `BuildDependency` not `HostBuildDependency`, because our OCaml packages hardcode the library search path to `/destdir`.

mlechu:
Thanks for the help! I realize that opam installs fine, but once you try to `opam init` it asks for another dependency (bwrap). Would it make the most sense to add that here, or package it on its own?

Keno:
Separate package, I think.

O/opam/build_tarballs.jl
topolarity: `OCAMLLIB` is apparently a workaround for that: https://discuss.ocaml.org/t/stdlib-directory-relative-to-the-compiler-bins/2940

Keno: We can also patch OCaml to add a relative path option.

topolarity: https://github.com/JuliaPackaging/Yggdrasil/pull/10389

topolarity:
> once you try to `opam init` it asks for another dependency (bwrap)

Works OK for me if I use `opam init --disable-sandboxing` to bypass the bubblewrap dependency.

However, running from within Julia has problems (when invoking `curl`):
```julia
julia> run(`$(opam()) init --disable-sandboxing`)
No configuration file found, using built-in defaults.
Checking for available remotes: rsync and local, git.
  - you won't be able to use mercurial repositories unless you install the hg command on your system.
  - you won't be able to use darcs repositories unless you install the darcs command on your system.


<><> Fetching repository information ><><><><><><><><><><><><><><><><><><><><><>
[ERROR] Could not update repository "default": OpamDownload.Download_fail(_, "curl failed: \"/usr/bin/curl --write-out
        %{http_code}\\\\n --retry 3 --retry-delay 2 --user-agent opam/2.3.0 -L -o /tmp/opam-727479-ccdb4c/index.tar.gz.part --
        https://opam.ocaml.org/index.tar.gz\" exited with code 1 \"/usr/bin/curl:
        /home/topolarity/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/bin/../lib/julia/libcurl.so.4: version `CURL_OPENSSL_4' not found
        (required by /usr/bin/curl)\"")
 ```

**Comment by giordano** [comment]
Should work on nightly.

**Comment by giordano** [comment]
Uhm, I wonder if it's due to a mismatch with libcurl version between system curl and our libcurl, Elliot had similar problems lately: https://github.com/JuliaRegistries/General/pull/123382.  How are you running this program from Julia?

**Comment by topolarity** [comment]
I am doing
```
using opam_jll
run(`$(opam()) init --disable-sandboxing`)
```

**Comment by giordano** [comment]
Not super standard, but
```julia
using opam_jll
run(`$(opam_jll.opam_path) init --disable-sandboxing`)
```
works for me.  Alternatively, if `only` is the only external program called by `opam` you could do
```julia
using opam_jll, CURL_jll
opam_exe = addenv(opam(), "PATH"=>dirname(CURL_jll.curl_path))
run(`$(opam_exe) init --disable-sandboxing`)
```

topolarity:
Hmm the first one works for me (not surprising, since I guess that's mostly equivalent to executing it directly)

The second one seems to shave off too much of the PATH, breaking OCaml's OS detection:
```
Fatal error: Undefined boolean filter value: os != "win32"
```

If I extend the `PATH`, it behaves better but still fails:
```julia
using opam_jll, CURL_jll
opam_exe = addenv(opam(), "PATH"=>"$(ENV["PATH"]):$(dirname(CURL_jll.curl_path))")
run(`$(opam_exe) init --disable-sandboxing`)
```

This gives:
```julia
<><> Fetching repository information <><><><><><><><><><><><><><><><>
[ERROR] Could not update repository "default":
        OpamDownload.Download_fail(_, "curl failed: \"/usr/bin/curl
        --write-out %{http_code}\\\\n --retry 3 --retry-delay 2
        --user-agent opam/2.3.0 -L -o
        /tmp/opam-816965-3e35d9/index.tar.gz.part --
        https://opam.ocaml.org/index.tar.gz\" exited with code 60")
[ERROR] Initial download of repository failed.
```
which is what happens with nightly, or if I add `CURL_jll` as a Dependency and deploy to a local JLL.

giordano:
> The second one seems to shave off too much of the PATH, breaking OCaml's OS detection:

Interesting, it worked for me (well, I only tried to start the program and that's it, didn't proceed with anything). But yeah, the suggestion was a bit brutal. You can prepend curl's PATH to system's PATH if that helps.

topolarity:
Most likely `opam` was happy from your previous run because it had already setup `~/.opam` (you have to clear it out between tests)

My takeaway is that opam just doesn't like Julia's CURL for whatever reason, so we'll have to force it to use the system CURL instead. Anyway I appreciate the help!

PR #10381, opened 2025-01-31
Add sokol

VarLad:
Currently only compiles on Linux.
Not sure whats the issue on Windows, on Apple I believe the `lld-macos: error: undefined symbol: ___isPlatformVersionAtLeast` issue can be fixed by linking to `libclang_rt.osx` but I'm not sure how to do that... any help would be appreciated!
Also, Apple and Windows are having the same issue of undefined reference to `sokol_main`, not sure what that means....


giordano:
> on Apple I believe the lld-macos: error: undefined symbol: `___isPlatformVersionAtLeast` issue can be fixed by linking to `libclang_rt.osx` but I'm not sure how to do that... any help would be appreciated!

When else fails, use the [search functionality](https://github.com/search?q=repo%3AJuliaPackaging%2FYggdrasil%20libclang_rt&type=code)

S/Sokol/build_tarballs.jl
```diff
@@ -0,0 +1,133 @@
...
+    if (CMAKE_SYSTEM_NAME STREQUAL Linux)
+        target_link_libraries(sokol INTERFACE X11 Xi Xcursor GL asound dl m)
+        target_link_libraries(sokol PUBLIC Threads::Threads)
+    endif()
+    if(CMAKE_SYSTEM_NAME STREQUAL Windows)
+        set(exe_type WIN32)
+    endif()
+endif()
+target_include_directories(sokol INTERFACE sokol)
+EOF
```
giordano, line 91: Can you please put this stuff in separate files?  It'd make reading them much easier.

VarLad: Will patching like the latest commit do?

VarLad:
> When else fails, use the search functionality

That issue is solved, thanks! :)
All thats left is the undefined reference to sokol_main issue in both Windows and Apple...

S/Sokol/build_tarballs.jl
giordano: Not really, just have them as separate files.

S/Sokol/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+
+install -Dm 755 "build/libsokol.${dlext}" "${libdir}/libsokol.${dlext}"
+install -Dm 755 "sokol_gfx.h" "${includedir}/sokol_gfx.h"
+install -Dm 755 "sokol_app.h" "${includedir}/sokol_app.h"
+install -Dm 755 "sokol_audio.h" "${includedir}/sokol_audio.h"
+install -Dm 755 "sokol_time.h" "${includedir}/sokol_time.h"
+install -Dm 755 "sokol_glue.h" "${includedir}/sokol_glue.h"
+install -Dm 755 "sokol_log.h" "${includedir}/sokol_log.h"
+install -Dm 755 "sokol_args.h" "${includedir}/sokol_args.h"
+install -Dm 755 "sokol_fetch.h" "${includedir}/sokol_fetch.h"
```
giordano, line 44: 1. header files aren't executable
2. can't just iterate over them?
```suggestion
for file in sokol_*.h; do
    install -Dvm 644 "${file}" -t "${includedir}"
done
```

S/Sokol/build_tarballs.jl
```diff
@@ -0,0 +1,71 @@
...
+
+cmake -B build \
+    -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    "${FLAGS[@]}"
+
+cmake --build build
+
+install -Dm 755 "build/libsokol.${dlext}" "${libdir}/libsokol.${dlext}"
```
giordano, line 36:
```suggestion
install -Dvm 755 "build/libsokol.${dlext}" "${libdir}/libsokol.${dlext}"
```

VarLad:
@giordano Any idea on the issue that is occurring for Windows and Apple builds? Looks like its the same issue for both :upside_down_face:

VarLad:
All required builds pass, except rv64 since it lacks a dependency, which I disabled! Should be ready for review now.

imciner2:
I'm confused by exactly what you are trying to package here. According to the upstream git repo (https://github.com/floooh/sokol), sokol is a header library, meaning it should just be packaging headers that the dependencies include. You seem to be adding a C file that has an empty main function and compiling a shared library - why is this needed? If this is truly just a header only library, then this library just needs to install the headers and have an AllPlatforms tag.

VarLad:
@imciner2 I share similar concerns so I looked previous instances of something similar
- I found https://github.com/Gnimuc/STBImage.jl which was doing something similar with https://github.com/Gnimuc/STBImageBuilder/tree/master/src
- I believe the official bindings like https://github.com/floooh/sokol-nim/ do something similar

So I thought building it as a library and wrapping it with Clang.jl is the best way forward.

VarLad:
> Maybe the question is: what do you want to do exactly with this?

Generate bindings with clang and use its functions

VarLad:
@giordano Given that I want to use the functions from this library in Julia, if not this, what is the best way to do this in your opinion.
The first example I've posted is doing exactly this I believe. 
The second one I suspect is doing something similar too, and those are the official bindings from the developer.

Although, I looked for something similar in JuliaBinaryWrappers and couldn't find anything similar to my case. There's stb but its simply packaged as headers and I don't see a julia package which uses them or expose them to someone using it to write a Julia library.

imciner2:
Ok, if you want to actually use these functions in Julia, then you probably do need to compile this into a library. Usually we see header only libraries used as dependencies for other packages instead of from Julia itself.

As for the wrapper, it looks more complicated and confusing than it really needs to be. It references emscripten in the CMake, you don't need to give the headers to the target, and also the main function isn't needed (libraries don't have main functions).

VarLad:
@imciner2 Thanks for the comment! I made some changes accordingly :)

ViralBShah:
Merge?

PR #10385, opened 2025-01-31
[snappy] Pick up riscv platform

jeremiahpslewis:


ViralBShah:
What do we do when upstream releases the next version? I suppose we have to wait two releases to sync up?

jeremiahpslewis:
Next minor release we can catch up, with the point releases we are stuck being one ahead

PR #10389, opened 2025-01-31
[OCaml] Add fallback for re-located libdir / bindir

topolarity:
The existing `OCAMLLIB` environment variable already allows you to use a standard library installed elsewhere, but it doesn't let you re-locate an existing installation. As one might expect, the problem is that the stdlib itself contains absolute references to its own paths.

This patch tries to hook into both places where that happens and do an on-the-fly replacement of the standard library path, if it turns out this is the standard library that OCaml that was built with (just re-located).

This kicks in if the default hard-coded `%%LIBDIR%%`/`%%BINDIR%%` cannot be found.


topolarity:
Just as a note to my future self, this is tested by:
1. Checking the paths in `ocamlrun -config` and `ocamlc -config` 
2. Compiling a hello-world via `ocamlc hello.ml -o hello` and `ocamlopt hello.ml -o hello` and running it

Notes:
(1) All absolute paths except `standard_library_default` should track the installation prefix if you move it
(2) `ocamlc` hard-codes a path to the OCaml interpreter in its output, so it's expected to break if you move `ocamlc` but it should work as long as `ocamlc` was moved before you byte-compile the test program. The problem fixed here is that `ocamlc` would write files that referenced its original install location, even if it had already moved.

Keno:
I don't know enough about OCaml to evaluate the actual logic, but the concept is right.

PR #10399, opened 2025-02-01
[CImGuiPack] Bump version of wrappers

JamesWrigley:
Building for https://github.com/JuliaImGui/cimgui-pack/pull/10.

### DIFF
--- a/C/CImGuiPack/build_tarballs.jl
+++ b/C/CImGuiPack/build_tarballs.jl
@@ -50,8 +50,8 @@ platforms = vcat(libjulia_platforms.(julia_versions)...)
 platforms = expand_cxxstring_abis(platforms)
 
 # We don't build for armv6l because GLFW_jll doesn't support it, and we don't
-# build for aarch64-freebsd because it's GLFW-related pain.
-platforms = filter(p -> arch(p) != "armv6l" && !(arch(p) == "aarch64" && os(p) == "freebsd"), platforms)
+# build for aarch64-freebsd or riscv64 because they're dependency pain.
+platforms = filter(p -> arch(p) != "armv6l" && arch(p) != "riscv64" && !(arch(p) == "aarch64" && os(p) == "freebsd"), platforms)
 
 # The products that we will ensure are always built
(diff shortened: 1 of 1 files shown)


JamesWrigley:
I am far too lazy to add support for riscv for all the dependencies so I just disabled it in 3a6122c7.

giordano:
Why are you building for all versions of Julia down to v1.6 if `CImGui.jl` supports v1.9+: https://github.com/JuliaImGui/CImGui.jl/blob/e020273525ad5e9c95f953630fbc9d075a59af00/Project.toml#L35?

JamesWrigley:
No reason, I just never noticed it. Bumped the compat bounds in da4472fd.

C/CImGuiPack/build_tarballs.jl
```diff
@@ -80,4 +80,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version=v"8")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.9", preferred_gcc_version=v"8")
```
giordano, line 83: This doesn't do anything, the problem is `julia_versions`

JamesWrigley: Mmmm I see.

JamesWrigley: Should be filtered out in 6ab08098.

giordano: To be clear, I don't mind _too_ much about this compat, what I meant is that it doesn't affect the "platforms" for which you actually build, but perhaps now that you're building only for julia v1.9+ keeping the compat to v1.9 would be more consistent.

JamesWrigley: Ok, added it back in b1b9f3a0.

PR #10408, opened 2025-02-02
[OpenSSL] bump to v3.4.0 and add riscv

jeremiahpslewis:


giordano:
We're sticking to the LTS for the time being

PR #10423, opened 2025-02-03
[libjulia] update 1.12-DEV to most recent master

fingolfin:
There were some ABI breaking changes e.g. to struct layouts.

CC @benlorenz @barche


fingolfin:
Will disable RISC-V but first need to see what the 1.12 job does

giordano:
We're still missing a couple of deps for riscv64 here in Yggdrasil. We have a build of llvm, but that's for version 19, Julia master is still on 18 at the moment

PR #10424, opened 2025-02-03
[goaws] Update to 0.5.2

omus:


omus:
I've used this locally with some GoAWS.jl tests and am going to proceed

PR #10429, opened 2025-02-04
[libjulia] Rebuild for julia 1.13.0-DEV

lgoettgens:
cc @benlorenz @fingolfin @barche

### DIFF
--- a/L/libjulia/common.jl
+++ b/L/libjulia/common.jl
@@ -2,7 +2,7 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder, Pkg
 
-include("../../fancy_toys.jl") # for get_addable_spec
+include("../../fancy_toys.jl") # for get_addable_spec and should_build_platform
 
 # list of supported Julia versions
 julia_full_versions = [v"1.6.3", v"1.7.0", v"1.8.2", v"1.9.0", v"1.10.0", v"1.11.1", v"1.12.0-DEV", v"1.13.0-DEV"]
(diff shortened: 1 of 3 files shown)


L/libjulia/common.jl
```diff
@@ -444,6 +449,12 @@ function build_julia(ARGS, version::VersionNumber; jllversion=version)
         push!(dependencies, Dependency(get_addable_spec("LibUnwind_jll", v"1.8.1+2"); platforms=filter(!Sys.isapple, platforms)))
         push!(dependencies, Dependency(get_addable_spec("LLVMLibUnwind_jll", v"19.1.4+0"); platforms=filter(Sys.isapple, platforms)))
         push!(dependencies, BuildDependency(get_addable_spec("LLVM_full_jll", v"18.1.7+3")))
+    elseif version.major == 1 && version.minor == 13
```
giordano, line 452: Uhm, there's something wrong here, both 1.12 and 1.13 will use openssl in place of mbedtls (completely removed).  I guess #10423 was already wrong.

lgoettgens: Thanks for the catch! Does the updated handling now look right to you?

giordano: At a quick glance yes but I haven't double checked the versions numbers.

fingolfin: While I agree we should get this right, AFAICT this does not affect the produced libjulia because we don't like against OpenSSL or MBedTLS on the C/C++ level, only the stdlib does. Or am I mistaken there?

The same probably also holds for e.g. SuiteSparse_jll and a bunch more

If this is right, then *ideally* the Julia build system would allow us to build the kernel w/o downloading these JLLs first...

fingolfin:
CI failed in a strange way. @lgoettgens perhaps `git commit --amend . && git push -f` to get it to re-run CI (unfortunately I can't restart buildkite jobs and IIRC closing & reopening PRs doesn't seem to trigger a fresh fresh run there)

giordano:
Yesterday there was an agent which was misbehaving (see also #10431), I had deleted its directory (already yesterday) and now I restarted the failed jobs here.

PR #10433, opened 2025-02-04
added recipe for blockSQP

chplate:
I wrote a wrapper for [blockSQP](https://github.com/djanka2/blockSQP), an NLP solver using a SQP method which builds upon the open-source QP solver [QPOASES](https://github.com/coin-or/qpOASES), for which I've recently added [qpOASES_jll](https://github.com/JuliaBinaryWrappers/qpOASES_jll.jl).


B/blockSQP/build_tarballs.jl
```diff
@@ -0,0 +1,97 @@
...
+    -DBLAS_LIBRARIES=${libdir}/${BLASLIBNAME}.${dlext} \
+    ..
+
+make
+make install
+install_license ${WORKSPACE}/srcdir/blockSQP/LICENSE
+"
+
+#platforms = vcat(libjulia_platforms.(julia_versions[julia_versions .>= v"1.7.0"])...) |> expand_cxxstring_abis
+platforms = vcat(libjulia_platforms.(julia_versions)...) |> expand_cxxstring_abis
```
giordano, line 75: Below you put compat for julia to v1.7, you likely want to filter only Julia v1.7+ versions in `julia_versions`.  Also, any specific reason for supporting v1.7 at this point?  Would you consider v1.10 instead (current LTS).  That'd further reduce the number of builds here.

B/blockSQP/build_tarballs.jl
```diff
@@ -0,0 +1,97 @@
...
+filter!(p -> !(arch(p) == "riscv64"), platforms)
+filter!(p -> !(os(p) == "freebsd"), platforms)
+
+
+
+dependencies = [
+    Dependency("qpOASES_jll"),
+    Dependency("libblastrampoline_jll"),#; compat="5.4.0"),
+    Dependency("CompilerSupportLibraries_jll"),
+    Dependency("libcxxwrap_julia_jll"),
```
giordano, line 87: You must specify compat bounds for lbt and libcxxwra

giordano (changes requested):

B/blockSQP/build_tarballs.jl
```diff
@@ -0,0 +1,97 @@
...
+
+
+include("../../L/libjulia/common.jl")
+
+# This is for anyone who is on macos
+preamble = if os() === "macos"
+    raw"cd /usr/share/cmake/Modules/Compiler/; find  . -name '._*' -exec rm {} \;"
+else
+    raw""
+end
```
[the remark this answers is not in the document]
chplate: This was a suggested fix to a problem in BinaryBuilder for people on MacOS, see [this issue](https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1348). I removed it now as we can do without it.

imciner2:
I would suggest modifying the recipe to use the new way of linking against LBT:

1) Add the line `apk remove cmake` to the top of the script
2) Remove all lines setting the blas library file name
3) Add the CMake variable `"-DBLA_VENDOR=libblastrampoline"`, and remove the other BLAS/LAPACK variables from the Cmake line.

B/blockSQP/build_tarballs.jl
chplate: I can't get it working with 5.4.0 on lbt. Now, I went back to 3.0.4, and locally it builds for me.

B/blockSQP/build_tarballs.jl
```diff
@@ -0,0 +1,78 @@
...
+filter!(p -> !(arch(p) == "i686"), platforms)
+filter!(p -> !(libc(p) == "musl"), platforms)
+filter!(p -> !(arch(p) == "riscv64"), platforms)
+filter!(p -> !(os(p) == "freebsd"), platforms)
+
+
+
+dependencies = [
+    Dependency("qpOASES_jll"; compat="3.2.1"),
+    Dependency("libblastrampoline_jll"; compat="3.0.4"),
```
imciner2, line 66:
```suggestion
    Dependency("libblastrampoline_jll"; compat="5.4"),
```
No, you need this compatibility the way it was, and we just need to figure out why it didn't work before.

LBT 3 is very old and is in Julia 1.7, but all Julia versions 1.8+ use LBT 5 of some version, so if you constrain it to LBT 3, it will make this uninstallable on Julia 1.8+. By choosing LBT 5.4, you support versions of Julia 1.9+.

B/blockSQP/build_tarballs.jl
```diff
@@ -0,0 +1,78 @@
...
+
+    SYMB_DEFS=()
+    if [[ \"${target}\" != *-apple-* ]]; then
+        for sym in cblas_dgemm; do
+            SYMB_DEFS+=(\"-D${sym}=${sym}64_\")
+        done
+    fi
+    if [[ \"${target}\" == *-apple-* ]] || [[ \"${target}\" == *-mingw* ]]; then
+        FLAGS+=(-DALLOW_OPENBLAS_MACOS=ON)
+    fi
```
imciner2, line 32:
```suggestion
```
I don't think you need this when using LBT.

imciner2 (changes requested):

B/blockSQP/build_tarballs.jl
giordano: > I can't get it working with 5.4.0 on lbt

Care to elaborate?

chplate:
So, I had a look at all other build scripts in Yggdrasil, and this is the only one that uses libcxxwrap, libjulia and libblastrampoline together. There were others that used the first two and then OpenBLAS_jll. I tried to link against OpenBLAS_jll, but then again I was required to use a compat=0.3.13 for it, a version which is from August 2021. 
However if I leave away the compat for OpenBLAS or lbt altogether, everything builds and I can use the resulting jll-package just fine. 
All in all, I don't really know how to proceed here and would be grateful for any further hints.

chplate:
For testing purposes, I deployed the jll-package without the compat bound on libblastrampoline to [this repository](https://github.com/chplate/blockSQP_jll.jl), and successfully tested the wrapper code under MacOS and Linux with Julia version 1.10. Do we really need the compat bound on libblastrampoline?

chplate:
I've seen that LAPACK/BLAS was only used in functions that were never called in the library, so we removed the part of the code and the dependency on lbt. However, now there are two builds failing with C++ compiler errors. Could you please have a look at this, and maybe restart? Thank you.

giordano (approved):

imciner2 (approved):

PR #10439, opened 2025-02-05
[GAP] rebuild with latest libjulia (for 1.13-DEV)

lgoettgens:
cc @fingolfin

PR #10441, opened 2025-02-05
[libcxxwrap] Rebuild using latest libjulia (for 1.13-DEV)

lgoettgens:
cc @barche @fingolfin @benlorenz

### DIFF
--- a/L/libcxxwrap_julia/build_tarballs.jl
+++ b/L/libcxxwrap_julia/build_tarballs.jl
@@ -11,7 +11,7 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 include("../../L/libjulia/common.jl")
 
 name = "libcxxwrap_julia"
-version = v"0.13.3"
+version = v"0.13.4"
 
 git_repo = "https://github.com/JuliaInterop/libcxxwrap-julia.git"
 
--- a/L/libcxxwrap_julia/build_tarballs.jl
+++ b/L/libcxxwrap_julia/build_tarballs.jl
@@ -17,7 +17,7 @@ git_repo = "https://github.com/JuliaInterop/libcxxwrap-julia.git"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource(git_repo, "f7bb672a0e72437d7ab044e1950a62ae7dd3399b"),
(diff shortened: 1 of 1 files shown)


fingolfin (approved):

L/libcxxwrap_julia/build_tarballs.jl
```diff
@@ -59,3 +59,5 @@ dependencies = [
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
     preferred_gcc_version = v"10", julia_compat = "1.6")
+
+# rebuild trigger: 1
```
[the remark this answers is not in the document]
lgoettgens: The version of the jll is tightly coupled with https://github.com/JuliaInterop/libcxxwrap-julia, so I don't know if I can just bump the version here. Maybe @barche could comment on this.

But as just mentioned in https://github.com/JuliaPackaging/Yggdrasil/pull/10439#discussion_r1943329995, I don't think we need a version bump at all here. (But I could be wrong)

L/libcxxwrap_julia/build_tarballs.jl
```diff
@@ -11,7 +11,7 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 include("../../L/libjulia/common.jl")
 
 name = "libcxxwrap_julia"
-version = v"0.13.3"
+version = v"0.13.4"
```
lgoettgens, line 14: I now bumped the version here. However, this introduces a mismatch between the libcxxwrap version and the jll one.

I would leave it up to @barche to decide on how to deal with this.

fingolfin: yes. and in the meantime we can see whether it works for GAP_jll

barche: I will make a release for v0.13.4, identical to v0.13.3 except for the version number, so this PR will need to be updated with the new SHA once this is done.

barche: Updated now, should be good to go after the rebuild.

lgoettgens: Thanks @barche! Looks good to me

PR #10442, opened 2025-02-05
[PROJ] Bump to v9.5.1

jeremiahpslewis:


P/PROJ/build_tarballs.jl
```diff
@@ -96,8 +96,10 @@ dependencies = [
     # Host SQLite needed to build proj.db
     HostBuildDependency("SQLite_jll")
     Dependency("SQLite_jll")
-    Dependency("Libtiff_jll"; compat="4.5.1")
-    Dependency("LibCURL_jll"; compat="7.73,8")
+    Dependency("Libtiff_jll"; compat="4.7.1")
+    Dependency("LibCURL_jll"; compat="8.11.1")
```
giordano, line 100: This won't work in any version of julia, don't change libcurl dependency randomly
```suggestion
    Dependency("LibCURL_jll"; compat="7.73,8")
```

PR #10444, opened 2025-02-05
[GAP] rebuild with latest libjulia (for 1.13-DEV), now with version bump

lgoettgens:
cc @fingolfin


lgoettgens:
It seems that we indeed need this.
The problem is that in https://github.com/JuliaPackaging/Yggdrasil/pull/10444, we need to specify against which GAP_jll version to build juliainterface. A simple rebuild without version bump makes it unable to specify the new version.

PR #10445, opened 2025-02-05
[Arpack] Rebuild to include riscv

ViralBShah:
Is this the right way to pull in new architectures? Just rebuild without changing anything else?


imciner2:
> Is this the right way to pull in new architectures? Just rebuild without changing anything else?

Not in this case, because now it would have bumped the compat version for JLLWrappers and that causes issues in the registry. We had to back out prior changes because of this: https://github.com/JuliaRegistries/General/pull/122129. In order to do a rebuild in Ygg to get riscv now, we also need to bump the version.

ViralBShah:
It seems that the Arpack32 build went through and even got merged on General - since it is a new package and probably used the latest JLLWrappers. Good luck I suppose that the Arpack build itself failed. 

We can just wait for a new upstream release.

PR #10446, opened 2025-02-05
[Heptagon] Add v1.05

topolarity:
This adds a recipe for heptagon (https://gitlab.inria.fr/synchrone/heptagon) a synchronous dataflow language and compiler.

Almost entirely thanks to @mlechu and @xal-0

### DIFF
--- a/H/Heptagon/build_tarballs.jl
+++ b/H/Heptagon/build_tarballs.jl
@@ -1,0 +1,85 @@
+using BinaryBuilder, Pkg
+
+name = "Heptagon"
+version = v"1.05.00"
+
+sources = [
+    DirectorySource("./bundled"),
+
+    ArchiveSource("http://download.camlcity.org/download/findlib-1.9.6.tar.gz",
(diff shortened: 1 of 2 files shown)


H/heptagon/build_tarballs.jl
giordano, line 1: Any reason for having a single megapackage instead of individual packages?

topolarity: Not particularly.. it was just the simplest thing to make sure we could get heptagon working

I can break it up if you'd prefer

giordano: In general I'd prefer it to maximise reusability, but in this case I'm not entirely sure it'd be worth it, these packages feel like a bit niche

topolarity: Maybe we can start with `dune` and `ocamlbuild` since those are broadly useful for OCaml projects, and leave the rest for later

giordano: Ok, let's do that.

topolarity:
@giordano Any lingering issues here?

PR #10453, opened 2025-02-06
[GAP_pkg_*] Rebuild for julia 1.13-DEV (group 1)

lgoettgens:
I have split all the GAP_pkg_*_jlls up into four groups with approximately the same number of build jobs, with the goal to not have to rebuild everything if there is some random failure in one of the builds.

The groups are:
- `browse` - `deepthought` (alphabetically): here
- `digraphs` - `io`: https://github.com/JuliaPackaging/Yggdrasil/pull/10454
- `juliainterface`: https://github.com/JuliaPackaging/Yggdrasil/pull/10440
- `json` - `zeromqinterface`: https://github.com/JuliaPackaging/Yggdrasil/pull/10455

cc @fingolfin

### DIFF
--- a/G/GAP_pkg/GAP_pkg_browse/build_tarballs.jl
+++ b/G/GAP_pkg/GAP_pkg_browse/build_tarballs.jl
@@ -2,10 +2,10 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 include("../common.jl")
 
-gap_version = v"400.1400.0"
+gap_version = v"400.1400.3"
 name = "Browse"
 upstream_version = "1.8.21" # when you increment this, reset offset to v"0.0.0"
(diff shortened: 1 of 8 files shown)


lgoettgens:
Hm, after looking at the generated jobs, it seems that none of the jlls expands on the julia version. Thus, it may be that this rebuild of all GAP_pkg_*_jlls is unnecessary. I guess we will see that in https://github.com/oscar-system/GAP.jl/pull/1139 once GAP_pkg_juliainterface_jll is released

fingolfin:
juliainterface is the only which accesses libjulia. At least from an ABI point of view, none of the other JLLs "need" an update. I also don't see other technical reasons that would require one. But let's see and try w/o these three JLLs. If they work we can close these PRs.

lgoettgens:
CI in https://github.com/oscar-system/GAP.jl/pull/1139 is happy, even without this. Thus closing...

PR #10456, opened 2025-02-06
[CImGuiPack] Bump to build for Julia 1.13

JamesWrigley:
After #10437 was merged CI failed on Julia 1.13 and that caused registration to be blocked: https://buildkite.com/julialang/yggdrasil/builds/17275#_

But libcxxwrap is now built for 1.13 so the build should succeed: #10441.


giordano (changes requested):
Can't change compat bounds without changing the version number.

JamesWrigley:
(bump)

giordano (approved):

PR #10458, opened 2025-02-06
[libsingular_julia@0.46] rebuild for julia 1.13-DEV

lgoettgens:
cc @fingolfin @hannes14 

This is a step in making Oscar work with 1.13-nightly, even without the breaking Singular.jl release you are currently working on. And in any case, this rebuild will become available for Oscar 1.2.2 automatically, so the latest Oscar release then works with julia nightly as well.

### DIFF
--- a/L/libsingular_julia/libsingular_julia@0.46/build_tarballs.jl
+++ b/L/libsingular_julia/libsingular_julia@0.46/build_tarballs.jl
@@ -1,0 +1,72 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+using Base.BinaryPlatforms
+
+# See https://github.com/JuliaLang/Pkg.jl/issues/2942
+# Once this Pkg issue is resolved, this must be removed
+uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
+delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
+
+name = "libsingular_julia"
+version = v"0.46.1"
+
+# Collection of sources required to build libsingular-julia
+sources = [
+    GitSource("https://github.com/oscar-system/Singular.jl.git", "9d9b9756549fd6594f1fd555039650d19c11652f"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd Singular.jl/deps/src
+cmake . -B build \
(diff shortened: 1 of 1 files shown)


fingolfin:
I don't understand the motivation behind this. We never promised that Oscar 1.2.2 will run on Julia 1.12 or 1.13. And surely we can wait a few more days before Oscar master passes on Julia nightly again? It hasn't been passing for a few weeks now anyway, hasn't it?

I am very wary of introducing duplicates for recipes like this.

lgoettgens:
> I don't understand the motivation behind this. We never promised that Oscar 1.2.2 will run on Julia 1.12 or 1.13.

You are right, we never promised that. But not doing this would imply that we need to do a Oscar 1.3 release before julia 1.12 gets released, or otherwise we wouldn't have a release working with the latest julia.
Rebuilding this version for a while (e.g. until Oscar 1.3 is released) would ease the transition for our users since then it does not matter if they first upgrade julia to 1.12 or Oscar to 1.3.
 
> I am very wary of introducing duplicates for recipes like this.

The alternative would be to overwrite the main recipe. But since this would mess up the git history if we are changing back and forth between 0.46 and 0.47, I opted for this way (which btw 51 other folders in yggdrasil). And we can remove this folder at any time, as soon as we don't need it anymore.

If you are still not convinced, that is more than fine for me as well. In this case, please just close this PR.

fingolfin:
We also would need to update Singular.jl on a branch to pick up this JLL update. And it is not clear to me this alone is enough to make Oscar 1.2.2 work (don't we also need a GAP.jl update, for GAP.jl 0.12, to deal with the worldage issues?) So yeah, I am still not convinced. But I am not the only person in charge here. Perhaps something to discuss during next triage *shrug*.

Regardless of everything else, Oscar 1.3 should be released soon, independent of this. Waiting too long between releases is bad.

lgoettgens:
> We also would need to update Singular.jl on a branch to pick up this JLL update. And it is not clear to me this alone is enough to make Oscar 1.2.2 work (don't we also need a GAP.jl update, for GAP.jl 0.12, to deal with the worldage issues?) So yeah, I am still not convinced. But I am not the only person in charge here. Perhaps something to discuss during next triage _shrug_.

True. I didn't think of everything that's needed to get this here to users. And you are right with the worldage stuff.
So, overall this seems to be much more overhead than I originally thought. Closing...


> Regardless of everything else, Oscar 1.3 should be released soon, independent of this. Waiting too long between releases is bad.

I fully agree.

PR #10475, opened 2025-02-07
[SCIP_Papilo] up scip papilo to 9.2.1

matbesancon:


giordano:
This package is hitting ICEs with any version of GCC.

imciner2:
And not related to the problem at hand, but I see there is an option for a HiGHS interface in papilo, would it be worth turning that on now?

matbesancon:
> would it be worth turning that on now?

Yes I think it will be worth creating a separate build for this

imciner2:
> Yes I think it will be worth creating a separate build for this

Oh, it has to be separate from the normal build and isn't just an option the user can choose at runtime?

matbesancon:
Indeed, SCIP works with an abstract linear programming interface, only one corresponding to the current LP solver is compiled. Otherwise this would also bloat the binary size

imciner2:
> Indeed, SCIP works with an abstract linear programming interface, only one corresponding to the current LP solver is compiled. Otherwise this would also bloat the binary size

Makes sense. If you do make a new build for this, it would be good to do it with the common build script structure so that they can share as much of the code as possible.

matbesancon:
yes for sure, for now I went the lazy version with SCIP{_PaPILO}, it will be good to put things in common

PR #10500, opened 2025-02-10
[ITK-v5.3.1] Fixing Windows minc error

divital-coder:


I/ITK/build_tarballs.jl
```diff
@@ -10,13 +10,39 @@ script = raw"""
...
+    -DQNANHIBIT_VALUE__TRYRUN_OUTPUT:STRING=0 \
+    -DVXL_HAS_SSE2_HARDWARE_SUPPORT:STRING=1 \
+    -DVCL_HAS_LFS:STRING=1 \
+    -DDOUBLE_CONVERSION_CORRECT_DOUBLE_OPERATIONS:STRING=1 \
+    -DHAVE_CLOCK_GETTIME_RUN:STRING=0
+cmake --build build --parallel ${nproc}
+cmake --install build
+install_license ${WORKSPACE}/srcdir/ITK/LICENSE
+cp $prefix/lib/libitkminc2-5.3.dll $prefix/bin
+cp $prefix/lib/libitkminc2-5.3.dll.a $prefix/bin
```
giordano, line 39: 1. shared libraries go to `${libdir}`, import libraries are usually in `${prefix}/lib`
2. move files, don't duplicate them with a copy
3. you're already moving `.dll` files to `${libdir}` below, no?
```suggestion
```

I/ITK/build_tarballs.jl
```diff
@@ -10,13 +10,39 @@ script = raw"""
...
+    -DITK_USE_SYSTEM_ZLIB:BOOL=ON \
+    -DQNANHIBIT_VALUE:BOOL=0 \
+    -DQNANHIBIT_VALUE__TRYRUN_OUTPUT:STRING=0 \
+    -DVXL_HAS_SSE2_HARDWARE_SUPPORT:STRING=1 \
+    -DVCL_HAS_LFS:STRING=1 \
+    -DDOUBLE_CONVERSION_CORRECT_DOUBLE_OPERATIONS:STRING=1 \
+    -DHAVE_CLOCK_GETTIME_RUN:STRING=0
+cmake --build build --parallel ${nproc}
+cmake --install build
+install_license ${WORKSPACE}/srcdir/ITK/LICENSE
```
giordano, line 37: Avoid duplication, you're running pretty much the same commands in the other branch too.

I/ITK/build_tarballs.jl
divital-coder: Yes moving dll files to the newly created ${libdir} below this

divital-coder:
Hello, I tried simplifying the script but it appears I have some flags that need to run additionally on non-windows platforms, and the simplification resulted in other platforms breaking, is it possible i can have 2 different 
Cmake build commands with flags for the windows and non-windows platforms?

giordano:
>  I have some flags that need to run additionally on non-windows platforms

Sure, but add those flags conditionally only on Windows.

divital-coder:
done :)

PR #10506, opened 2025-02-11
[Reactant_jll] - Fix the reading of "cuda_version_tag" in ENV for platform augmentation.

alexiscltrn:
@avik-pal, @wsmoses 
I believe this might be a typo. I need to temporarily set `ENV['REACTANT_GPU']` and `ENV['REACTANT_CUDA_VERSION']` in order to fetch the correct tarball of `Reactant_jll` before sending it to my cluster (the platform is different from my local one).


giordano:
I'll take over this later, but please wait for https://buildkite.com/julialang/yggdrasil/builds/17546 to finish and the following PR to the merged. Also, this change doesn't require a full rebuild of Reactant (which takes 3 hours) but ony to regenerate the wrapper: https://github.com/JuliaPackaging/Yggdrasil/blob/master/CONTRIBUTING.md#special-keywords-in-commit-messages

giordano:
Thanks!

alexiscltrn:
Thank you very much for your help!

PR #10511, opened 2025-02-11
XDiag: new version 0.3.1

awietek:


giordano:
For [Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170)

> The math constants aren't defined in Standard C/C++. To use them, you must first define `_USE_MATH_DEFINES`, and then include `<cmath>` or `<math.h>`.

awietek:
Thanks, I figured this out and now some windows compilation is working. I have another question for OSX on the arm chips. I am currently compiling with GNU OpenMP, but you mentions LLVM OpenMP is preferable. At the time, LLVM OpenMP did not work, there were some segfaults if I remember. With GNU OpenMP it compiles and works, it's just not parallel. I realized there was this discussion, I was mentioned:

https://github.com/JuliaPackaging/Yggdrasil/pull/9423

 which seemed to have fixed this issue. Is it clear now what to do to use OpenMP on apple arm? Should I maybe try adapting this solution for the build script?

PR #10515, opened 2025-02-11
[ITK-v5.3.2] Finally fixed windows minc error

divital-coder:
Hello, So after weeks of head butting, I finally realised the minc error is only and only resolved if we standalone copy the libitkminc2 dll files in the bin directory for windows platforms.

Hoping to get this merged, this is the last time , since i tested it and it works like a charm with the upstream package.

Also, changed the package version to ensure compat bounds.

I tried everything but the only solution for libitkminc2 error in Cmake configuration on windows was copying both the dll files within the 
$prefix/bin directory.

### DIFF
--- a/I/ITK/build_tarballs.jl
+++ b/I/ITK/build_tarballs.jl
@@ -1,6 +1,6 @@
 using BinaryBuilder, Pkg
 name = "ITK"
-version = v"5.3.1"
+version = v"5.3.2"
 # Collection of sources required to complete build
 sources = [
     GitSource("https://github.com/InsightSoftwareConsortium/ITK.git", "1fc47c7bec4ee133318c1892b7b745763a17d411")
@@ -49,8 +49,8 @@ cmake --install build
 install_license ${WORKSPACE}/srcdir/ITK/LICENSE
 
 if [[ "${target}" == *x86_64-w64-mingw32* ]]; then
-mkdir -pv ${libdir}
-find "${prefix}/lib" -name "*.${dlext}" -exec mv -v {} ${libdir} \;
+    cp $prefix/lib/libitkminc2-5.3.dll $prefix/bin
+    cp $prefix/lib/libitkminc2-5.3.dll.a $prefix/bin
 fi
 """
 # These are the platforms we will build for by default, unless further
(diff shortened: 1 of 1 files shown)


I/ITK/build_tarballs.jl
```diff
@@ -49,8 +49,8 @@ cmake --install build
 install_license ${WORKSPACE}/srcdir/ITK/LICENSE
 
 if [[ "${target}" == *x86_64-w64-mingw32* ]]; then
-mkdir -pv ${libdir}
-find "${prefix}/lib" -name "*.${dlext}" -exec mv -v {} ${libdir} \;
+    cp $prefix/lib/libitkminc2-5.3.dll $prefix/bin
+    cp $prefix/lib/libitkminc2-5.3.dll.a $prefix/bin
```
giordano, line 53: I don't know how many times I need to repeat that these files should be moved, not copied, you just create useless duplicates.  And I still haven't understood what's the problem with the import library being in `${prefix}/lib` which is where it is for all the other thousands of packages we have here.

divital-coder: Yes i tried moving them and whatnot, I acknowledge your  requests, the CMake config failed at each point.
So this libitkminc is a third party library that is used in ITK, and the Cmake config error, is doing some weird checks.

giordano: > So this libitkminc is a third party library that is used in ITK, and the Cmake config error, is doing some weird checks.

Maybe that should be fixed instead of moving files around without a clear strategy?

divital-coder: This is only for windows, and I believe this sort of little workaround is really the sufficient thing. I have previously opened issues on the ITK forum, where they mentioned that minc is notorious for errors.

giordano: General suggestion (not just specific to this case): add a comments to explain why you're doing things, especially if they're surprising, hard to understand, and easy to forget for reviewers.  I'm 100% sure that next time you touch this code I'll be making the same remarks all over again.

divital-coder: Sorry for that, get carried away when things work right and just want to get over them, I will precisely add related comments, Thank you so much for raising this concern, and you are right. next time  when this thing need to be upgraded, I would have probably gone over the stuff once again.

PR #10519, opened 2025-02-12
[oneTBB] Bump to 2022.0.0

jeremiahpslewis:


O/oneTBB/build_tarballs.jl
```diff
@@ -30,6 +30,11 @@ if [[ ${target} == i686-linux-musl* ]]; then
     atomic_patch -p1 "${WORKSPACE}/srcdir/patches/i686-musl.patch"
 fi
 
+if [[ "${target}" == *-freebsd* ]]; then
+    # Follow oneTBB makefile fix for missing symbols error: https://cgit.freebsd.org/ports/commit/?id=3677983542cc09a0e5f085e463a895e9e4dce9aa
+    export LDFLAGS="-Wl,--undefined-version"
```
ararslan, line 35: I don't know whether there's an existing `LDFLAGS` defined (I strongly suspect there isn't) but you could do this just in case:
```suggestion
    export LDFLAGS="${LDFLAGS} -Wl,--undefined-version"
```

O/oneTBB/build_tarballs.jl
```diff
@@ -41,7 +46,8 @@ cmake --build build --parallel ${nproc}
 cmake --install build
 """
 
-platforms = expand_cxxstring_abis(supported_platforms())
+platforms = supported_platforms(exclude= x -> !Sys.iswindows(x))
```
ararslan, line 49: This builds _only_ for Windows. Is that intentional?

ararslan (approved):

PR #10523, opened 2025-02-13
[GR] bump version

jheinen:
Bump GR run-time version


jheinen:
I have no idea why `buildkite/yggdrasil/build-g-slash-gr-x86-64-unknown-freebsd` fails.

PR #10525, opened 2025-02-13
[Lua] Bump to pick up riscv

jeremiahpslewis:


L/Lua/build_tarballs.jl
```diff
@@ -2,6 +2,7 @@ using BinaryBuilder
 
 name = "Lua"
 version = v"5.4.7"
+ygg_version = v"5.4.8" # version increment to pick up riscv
```
ararslan, line 5: It isn't necessary to introduce alternate universe versioning when adding a new platform; you can just make a trivial change to the file to retrigger a build of the same version. I'd suggest just adding `# build trigger: 1` at the top or bottom of the file, and that can be edited the next time someone needs to rebuild an existing version.

ararslan (changes requested):

L/Lua/build_tarballs.jl
jeremiahpslewis: In this case I think it is necessary, because of a change to the jll wrapper version in the meantime. @giordano can you confirm?

giordano: Yes, https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1357 changed an implicit compat bound for all recipes, so that all rebuilds since 2024-12-24 must change the compat bound (only for the first time of course).

imciner2:
Note, this requires https://github.com/JuliaPackaging/Yggdrasil/pull/10526, so it should wait until that is merged. (although I don't understand why the build is succeeding on risc-v without readline, is it an optional dependency?)

ViralBShah:
I thought closing and reopening would rebuild.

giordano:
> I thought closing and reopening would rebuild.

No, buildkite react to draft/ready for review change, not closing/opening because it isn't triggered by merge commit (which changes on close/reopen) but by push commit.

ViralBShah:
Kicked off a rebuild from buildkite

ViralBShah:
Ok to nerge?

PR #10528, opened 2025-02-14
Try to add GAMS.

bdklahn:
https://www.gams.com/download


imciner2:
Are you affiliated with GAMS? They already package a wrapper for their software here https://github.com/GAMS-dev/gams.jl, and explicitly tell people to download it normally through their website, so that seems to be what they would prefer. Additionally, I don't think we would have a license to actually distribute the GAMS software through here, since that doesn't seem to be provided for in their license agreement (https://www.gams.com/GAMS_EULA.pdf).

bdklahn:
@imciner2 
Ah! Thanks so much for pointing that out! The last information I had found was a thread asking about support for Julia, which wasn't then implemented. Looks like maybe my information is old.

If this is already in-progress (and the license doesn't really permit it), I'll stop this. :-)

I was also partly testing my hand at being able to do this sort of thing for more things (thinking Julia Binary Builder and Artifacts might serve as simpler, more portable, "containers" in many cases).

bdklahn:
Closing until upstream confirms this is ok. . . and also so further development doesn't trigger any CI.

odow:
Just chiming in to say that we should not use Yggdrasil to distribute GAMS in any form.

PR #10530, opened 2025-02-14
MAGEMin v1.6.9

NicolasRiel:
- Changed bound values for compositional variables from 1e-10 to 1e-6, which largely improves accuracy and performances (15% faster)
- Improved calculation results for extended metapelite database
- Increased maximum size of possible considered phases (max value was hit in some cases)
- Fixed NLopt version to 2.8.0 as more recent version leads to large variation in calculation results between platforms

### DIFF
--- a/M/MAGEMin/build_tarballs.jl
+++ b/M/MAGEMin/build_tarballs.jl
@@ -85,7 +85,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="NLopt_jll", uuid="079eb43e-fd8e-5478-9966-2cf3e3edb778", compat="2.8.0"))
+    Dependency(PackageSpec(name="NLopt_jll", uuid="079eb43e-fd8e-5478-9966-2cf3e3edb778"); compat="2.8.0")
     Dependency("OpenBLAS32_jll"; platforms=filter(!Sys.isapple, platforms))
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
 ]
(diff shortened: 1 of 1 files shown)


M/MAGEMin/build_tarballs.jl
```diff
@@ -85,7 +85,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="NLopt_jll", uuid="079eb43e-fd8e-5478-9966-2cf3e3edb778"))
+    Dependency(PackageSpec(name="NLopt_jll", uuid="079eb43e-fd8e-5478-9966-2cf3e3edb778"); compat="2.8.0")
```
giordano, line 88: > Fixed NLopt version to 2.8.0 as more recent version leads to large variation in calculation results between platforms

This compat follows [Pkg compat bounds syntax](https://pkgdocs.julialang.org/v1/compatibility/#Version-specifier-format).  `2.8.0` means `[2.8.0, 3)`, which I'm not sure is what you want, based on your comment.

NicolasRiel: Ah sorry. You are right. I want to fix the specific version not put a lower bound. Thank you for providing the documentation. I fix this after lunch

giordano: Might also be beneficial for you to work with upstream to get whatever issue you see fixed, not to get stuck with an outdated version forever.

NicolasRiel: Yes, absolutely. This is also what I want to do. I am teaching with my code next week, and my plan was to contact NLopt author afterward.

NicolasRiel:
@giordano Sorry to annoy you. But I am not sure what to do here.

It seems "aarch64-apple-darwin-libgfortran5-mpi+mpich" has a problem with clang:

> ccache: error: execute_noreturn of /opt/x86_64-linux-musl/bin/clang failed: No such file or directory

It compiled correctly with the previous version 1.6.8 :( Do you think this is because I fixed NLopt version?

NicolasRiel:
@giordano Well thank you... my apologies for the hassle.

PR #10534, opened 2025-02-14
[GDAL] Bump v3.10.2

jeremiahpslewis:
Binaries to build to get gdal on riscv
- [x] Hdf5
- [x] lerc
- [x] libpq
- [x] NetCDF_jll
- [x] OpenJpeg_jll
- [x] PCRE2_jll
- [x] libgeotiff_jll


eschnett:
You can try (if you want) to build for `aarch64-*-freebsd` and/or `riscv64-*-linux` as well. This may or may not work, depending on whether the dependencies are available.

eschnett:
Almost there: https://github.com/JuliaPackaging/Yggdrasil/pull/10711

G/GDAL/build_tarballs.jl
```diff
@@ -200,30 +195,30 @@ hdf5_platforms = expand_cxxstring_abis(hdf5_platforms)
...
     # Dependency("HDF4_jll"; compat="4.3.0"),
     Dependency("HDF5_jll"; compat="~1.14.3", platforms=hdf5_platforms),
     Dependency("LERC_jll"; compat="4"),
     Dependency("LibCURL_jll"; compat="7.73,8"),
     Dependency("LibPQ_jll"; compat="16"),
-    Dependency("Libtiff_jll"; compat="4.7"),
-    Dependency("Lz4_jll"; compat="1.9.3"),
+    Dependency("Libtiff_jll"; compat="4.7.1"),
+    Dependency("Lz4_jll"; compat="1.10.1"),
     Dependency("NetCDF_jll"; compat="400.902.210", platforms=hdf5_platforms),
```
giordano, line 209: Probably want to use the 401 version

eschnett:
After https://github.com/JuliaPackaging/Yggdrasil/pull/10712 we can also enable HDF4 support in GDAL. (That's probably not very important at all.)

eschnett:
@jeremiahpslewis `SQLite_jll @3.49` does not exist; it was incompatible with `SQLite @3.48`. Try using `@3.48` instead.

eschnett:
We also need Blosc for riscv64: https://github.com/JuliaPackaging/Yggdrasil/pull/10715 .

G/GDAL/build_tarballs.jl
```diff
@@ -200,30 +195,30 @@ hdf5_platforms = expand_cxxstring_abis(hdf5_platforms)
...
 dependencies = [
     BuildDependency(PackageSpec(; name="OpenMPI_jll", version=v"4.1.6"); platforms=filter(p -> nbits(p)==32, platforms)),
-    Dependency("Arrow_jll"; compat="18.1.0"),
-    Dependency("Blosc_jll"; compat="1.21.1"),
-    Dependency("Expat_jll"; compat="2.2.10"),
-    Dependency("GEOS_jll"; compat="3.11.2"),
+    Dependency("Arrow_jll"; compat="19.0.0"),
+    Dependency("Blosc_jll"; compat="1.21.6"),
+    Dependency("Expat_jll"; compat="2.6.5"),
+    Dependency("GEOS_jll"; compat="3.13.1"),
```
giordano, line 201: GEOS devs believe it's a good idea to break the ABI in every single version: https://github.com/JuliaBinaryWrappers/GEOS_jll.jl/commit/e2e83661fbd85faedbd29485ac0a328338321f83#diff-0ed6f04e79778dda22f5565f89d1f6e95329a8d15c57c4efa586fbe7b7374bcd
```suggestion
    Dependency("GEOS_jll"; compat="=3.13.1"),
```
But also, hold on this on https://github.com/JuliaRegistries/General/pull/126663

giordano (changes requested):

eschnett:
This https://github.com/JuliaPackaging/Yggdrasil/pull/10712 is important. Currently, HDF4 and NetCDF conflict, and you can't use them together.

giordano:
Someone must also take care of https://github.com/JuliaRegistries/General/pull/126663#issuecomment-2712088284 (I don't have the time)

eschnett:
The breaking changes that @giordano mentions means that we have to do either of these two things:
- Increase the major version number of `GEOS_jll`, probably introducing a version number offset as we're doing for `GDAL_jll`. If `GEOS_jll` has been released as Julia package then we need to un-release ("yank") the broken version.
- Investigate why the shared libraries are marked as incompatible, and confirm that this was only by accident, and confirm that there is no actual incompatibility, and then manually change the soversion of the shared libraries (this is a long shot and is somewhat dangerous)

giordano:
As usual, need also to add appropriate compat bounds to all downstream JLL packages which directly depend on GEOS_jll

visr:
Looking back at the last years GEOS always had shared libraries with patch versions in the name, but it never caused issues that I know of. I just ran GDAL.jl tests switching between GEOS_jll 3.12.0, 3.12.1 and 3.12.0 for the same GDAL_jll, and it always passed. So it doesn't seem to be an issue in practice?

I checked the build logs and artifacts, and it shows that there are also `libgeos_c.1.dylib` and `libgeos_c.dylib` which are both symlinks pointing to `libgeos_c.1.19.2.dylib`

```
/opt/bin/aarch64-apple-darwin20-libgfortran5-cxx11/aarch64-apple-darwin20-clang++ --sysroot=/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root -O3 -DNDEBUG -mmacosx-version-min=11.0 -dynamiclib -Wl,-headerpad_max_install_names -compatibility_version 1.0.0 -current_version 1.19.2 -o lib/libgeos_c.1.19.2.dylib -install_name @rpath/libgeos_c.1.dylib CMakeFiles/geos_c.dir/capi/geos_c.cpp.o CMakeFiles/geos_c.dir/capi/geos_ts_c.cpp.o  -Wl,-rpath,/workspace/srcdir/geos-3.13.1/lib lib/libgeos.3.13.1.dylib 
make[2]: Leaving directory '/workspace/srcdir/geos-3.13.1'
/usr/bin/cmake -E cmake_symlink_library lib/libgeos_c.1.19.2.dylib lib/libgeos_c.1.dylib lib/libgeos_c.dylib

-- Installing: /workspace/destdir/lib/libgeos_c.1.19.2.dylib
-- Installing: /workspace/destdir/lib/libgeos_c.1.dylib
[cctools-port]: generating fake signature for '/workspace/destdir/lib/libgeos_c.1.19.2.dylib'
-- Installing: /workspace/destdir/lib/libgeos_c.dylib
```

The GDAL build logs themselves mention the full name but there is also this update linkage log:

```
install_name_tool -change libgeos_c.1.dylib @rpath/libgeos_c.1.dylib lib/libgdal.36.3.10.1.dylib
```

Is that why this isn't causing issues?

And note that these symlinks are only made on macOS and Linux, the Windows builds just have `libgeos_c.dll`. So it seems https://github.com/JuliaPackaging/Yggdrasil/issues/3024#issuecomment-2460984329 is not quite accurate.

giordano:
The soname that changed is `libgeos_cpp`, not `libgeos`

giordano:
Also, why the heck they do it right for one library but not the other? 🤦

visr:
In the [diff you linked](https://github.com/JuliaBinaryWrappers/GEOS_jll.jl/commit/e2e83661fbd85faedbd29485ac0a328338321f83#diff-0ed6f04e79778dda22f5565f89d1f6e95329a8d15c57c4efa586fbe7b7374bcd) both the C and C++ sonames change right?

And both have symlinks:
```
libgeos.dylib
libgeos.3.13.1.dylib
libgeos_c.dylib
libgeos_c.1.dylib
libgeos_c.1.19.2.dylib
```

The https://libgeos.org/usage/c_api/ and https://libgeos.org/usage/cpp_api/ docs really make it clear that folks that need any stability at all need to use the C API. And in the GDAL logs I see only linkage to the C API. I'd be very surprised if any of the others do use the C++ API.

giordano:
No, the soname is the argument to `declare_library_product`

visr:
Oh wait I see now that for `libgeos_c` the `declare_library_product` is `libgeos_c.1.dylib` and therefore stays the same and the `init_library_product` changes to `libgeos_c.1.19.2.dylib`. Odd that these two use different sonames?

giordano:
>  and the init_library_product changes to libgeos_c.1.19.2.dylib. Odd that these two use different sonames?

That's not odd, they simply use different arguments: `declare_library_product` uses the soname, `init_library_product` uses the real file name.  They need not to match, in fact quite often they don't.

giordano:
> The question is: does any of the [dependents of geos](https://juliahub.com/ui/Packages/General/GEOS_jll#dependents) link to the C++ library?

an answer to this question would be very helpful, I don't have the time to investigate this (beside zero personal interest in this stuff anyway), so if you do care about it please help out and you can move things more quickly, I won't do do anything myself

visr:
I already checked that GDAL only uses the C API. Libspatialite also does (only `-lgeos_c` in build logs) and Valhalla has no mention of geos in the build logs.

visr:
@jeremiahpslewis do you want to try enabling HDF4 support in this PR as well? See https://github.com/JuliaPackaging/Yggdrasil/pull/10534#issuecomment-2710628621. Contrary to that comment it is probably the most requested driver to add to our GDAL build.

eschnett:
As I mentioned, you'll have to use the newest release of HDF4 4.3.1 to avoid a conflict with NetCDF. Best use a compat entry for this; `compat="4.3.1"` should work.

giordano:
Yeah, ok, fine, at this point I'm wondering why we have the C++ library in the first place though 🫠

visr:
Yeah fair point, it would probably be fine to remove the C++ API from the JLL.

eschnett:
@jeremiahpslewis Do you want to close this PR now?

PR #10557, opened 2025-02-18
[GraphicsMagick] Rebuild 1.3.45 but bump version for Ygg.

ViralBShah:
Try to fix the segfault on `using GraphicsMagick_jll` by using fewer deps.

### DIFF
--- a/G/GraphicsMagick/build_tarballs.jl
+++ b/G/GraphicsMagick/build_tarballs.jl
@@ -3,6 +3,7 @@
 using BinaryBuilder
 name = "GraphicsMagick"
 version = v"1.3.45"
+ygg_version = v"1.3.46"
 
 # Collection of sources required to build GraphicsMagick
 sources = [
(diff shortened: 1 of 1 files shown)


G/GraphicsMagick/build_tarballs.jl
```diff
@@ -70,8 +74,9 @@ dependencies = [
     # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM
     # as compiler (BSD systems), and libgomp from
     # `CompilerSupportLibraries_jll` everywhere else.
-    Dependency("CompilerSupportLibraries_jll"; platforms=filter(!Sys.isbsd, platforms)),
-    Dependency("LLVMOpenMP_jll"; platforms=filter(Sys.isbsd, platforms)),
+    # Dependency("CompilerSupportLibraries_jll"; platforms=filter(!Sys.isbsd, platforms)),
+    # Dependency("LLVMOpenMP_jll"; platforms=filter(Sys.isbsd, platforms)),
+    Dependency("CompilerSupportLibraries_jll"),
```
giordano, line 79: Is CSL needed at all if you disable openmp?

ViralBShah: I usually always add it as a safe thing! I can remove it.

giordano: If the auditor doesn't say you should add then there's no need to add it.

PR #10564, opened 2025-02-19
[Octave] Update to 9.4.0

ViralBShah:


O/Octave/build_tarballs.jl
```diff
@@ -1,18 +1,16 @@
...
 name = "Octave"
-version = v"9.3.0" 
-bb_ver = v"9.3.1" # Bump patch level to have a new version for Ygg compat
+version = v"9.4.0" 
 
 # Collection of sources required to build Octave
 sources = [
-   ArchiveSource("https://ftpmirror.gnu.org/octave/octave-$(version).tar.gz",
-                  "809fa39a7acc84815bf4dc4d2d7e6b228ce75a07f3b2413f3313aa8e0aaa3287"),
+   GitSource("https://github.com/gnu-octave/octave/", "155645d388b7f50b4081a4b769dc9dd4a757f72b"),
```
giordano, line 8:
```suggestion
   GitSource("https://github.com/gnu-octave/octave.git", "155645d388b7f50b4081a4b769dc9dd4a757f72b"),
```

ViralBShah:
Needs to run `autoconf` etc. The tarballs do not need that, but the mirror sites are often unreliable. Safer to use github but needs a few more steps.

PR #10402, opened 2025-02-02
0_RootFS: Updated FreeBSD from 13.2 to 13.4 / 14.1

stemann:
Enables better C++-17 support.

E.g. for MLX, which ran into `error: constexpr variable cannot have non-literal type 'const std::array<uint32_t, 8>' (aka 'const array<unsigned int, 8>')` with the FreeBSD 13.2 libc++ SDK in #10374 - which could be remedied by using FreeBSD 13.4 instead of 13.2: https://github.com/JuliaPackaging/Yggdrasil/pull/10374/commits/cea8b97bfb5a11db2e8a8d03664aa06943dfadcc

How to use another FreeBSD SDK for a CMake-based build: 
```julia
sources = [
    ArchiveSource("http://ftp-archive.freebsd.org/pub/FreeBSD-Archive/old-releases/arm64/13.4-RELEASE/base.txz",
                  "730f4d4d850fa96263a874340d84a2f0575cb917fc035359ae34823cae56c809";
                  unpack_target="freebsd-base-aarch64"),
    ArchiveSource("http://ftp-archive.freebsd.org/pub/FreeBSD-Archive/old-releases/amd64/13.4-RELEASE/base.txz",
                  "8e13b0a93daba349b8d28ad246d7beb327659b2ef4fe44d89f447392daec5a7c";
                  unpack_target="freebsd-base-x86_64"),
]

script = raw"""
if [[ "$target" == *-unknown-freebsd* ]]; then
    sdk_root=$WORKSPACE/srcdir/freebsd-base-$(echo $target | cut -d - -f1)
    sed -i "s#/opt/$bb_target/$bb_target/sys-root#$sdk_root#" $CMAKE_TARGET_TOOLCHAIN
    sed -i "s#/opt/$bb_target/$bb_target/sys-root#$sdk_root#" /opt/bin/$bb_full_target/$target-clang*
    sed -i "s#/opt/$bb_target/$bb_target/lib#$sdk_root/usr/lib#" /opt/bin/$bb_full_target/$target-clang*
fi
"""
```

PR #10570, opened 2025-02-20
[MLX] Updated to v0.23.1

stemann:


stemann:
Awaiting #10374

stemann:
I'm hesitant to upstream the patch to not enable COMPILE_WARNING_AS_ERROR, as upstream is likely happy with warnings as errors.

... and so are we - happy that upstream is not allowing errors on the platforms being built and tested upstream, `aarch64-apple-darwin`, and `x86_64-linux-gnu`.

stemann:
[Removed customized FreeBSD 13.4/14.1 set-up](https://github.com/JuliaPackaging/Yggdrasil/pull/10570/commits/37dd29938768aaa519e0d62aba5280f00d674dd1) following the merge of #10402

M/MLX/bundled/patches/cmake-warning-as-error.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index d3df54878..e716e925a 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -81,7 +81,6 @@ include(FetchContent)
+ cmake_policy(SET CMP0135 NEW)
+ 
+ add_library(mlx)
+-set_target_properties(mlx PROPERTIES COMPILE_WARNING_AS_ERROR ON)
```
imciner2, line 9: According to this https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_WARNING_AS_ERROR.html, we can instead pass `--compile-no-warning-as-error ` on the CMake command line to ignore this property, so then we won't need to patch it away.

imciner2 (approved):

PR #10580, opened 2025-02-21
Wizard recipe: very_nauty-v1.1.0

Krastanov:

* Package name: very_nauty
* Version: v1.1.0

### DIFF
--- a/V/very_nauty/build_tarballs.jl
+++ b/V/very_nauty/build_tarballs.jl
@@ -15,7 +15,7 @@ script = raw"""
 cd $WORKSPACE/srcdir/very_nauty/
 mkdir -p ${libdir}
 cc -shared -fPIC -Wall -O3 -o "${libdir}/libvn_graph.${dlext}" vn_graph.c
-cp vn_graph.h ${includedir}
+install -Dvm 644 vn_graph.h -t ${includedir}
 """
 
 # These are the platforms we will build for by default, unless further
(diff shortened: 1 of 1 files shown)


Krastanov:
There are a few platforms that do not work for me yet:
```
 - FreeBSD aarch64
 - FreeBSD x86_64
 - Windows i686
 - Windows x86_64
 - macOS x86_64
 - macOS aarch64
```

For one of these two reasons:
- missing `<sys/times.h>`
- their compilers do not permit inline functions defined inside the body of other functions (something outside the C standard, but supported by gcc)

Is there a standard simple fix for these two issues?

V/very_nauty/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
...
+sources = [
+    GitSource("https://github.com/JuliaGraphs/very_nauty.git", "d6d28d3ad92f86953e7523749e1468ac4a936603")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/very_nauty/
+mkdir -p ${libdir}
+cc -shared -fPIC -Wall -O3 -o "${libdir}/libvn_graph.${dlext}" vn_graph.c
+cp vn_graph.h ${includedir}
```
giordano, line 18: You're creating a file named `${includedir}` which is most likely not what you meant
```suggestion
install -Dvm 644 vn_graph.h -t ${includedir}
```

Krastanov:
Thanks, @giordano , and thanks for the explanations you gave on slack.

I believe this PR is ready. The platforms that are not included are due to them not being supported upstream either. I will plan on working with the original author to resolve these, but they would need to be in a separate future release.

Krastanov:
It seems I am wrong about this being ready. On slack there were a few more comments about how this can crash on modern systems because of the inline functions. Marking as draft for now.

Krastanov:
closing in favor of https://github.com/JuliaPackaging/Yggdrasil/pull/10582

PR #10590, opened 2025-02-22
[Pixi] build 0.41.4

cjdoris:


giordano:
```
[12:01:07]   Downloaded clap v4.5.30
[12:01:07] error: failed to unpack package `clap v4.5.30`
[12:01:07] 
[12:01:07] Caused by:
[12:01:07]   failed to unpack entry at `clap-4.5.30/examples/tutorial_derive/03_05_default_values.rs`
[12:01:07] 
[12:01:07] Caused by:
[12:01:07]   failed to unpack `/opt/x86_64-linux-musl/registry/src/index.crates.io-6f17d22bba15001f/clap-4.5.30/examples/tutorial_derive/03_05_default_values.rs`
[12:01:07] 
[12:01:07] Caused by:
[12:01:07]   failed to unpack `clap-4.5.30/examples/tutorial_derive/03_05_default_values.rs` into `/opt/x86_64-linux-musl/registry/src/index.crates.io-6f17d22bba15001f/clap-4.5.30/examples/tutorial_derive/03_05_default_values.rs`
[12:01:07] 
[12:01:07] Caused by:
[12:01:07]   No space left on device (os error 28)
```
mumble mumble mumble, maybe we should revise where we put `CARGO_HOME`.

cjdoris:
I'm not sure what to do here, what's the fix? Can we give the runner more storage or something?

Anyway, what changed, the previous version of Pixi built no problem last week? If be surprised if the Pixi versions were so different as to cause this.

giordano:
> I'm not sure what to do here, what's the fix?

As I said above, change where [`CARGO_HOME`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads) points to.

> Can we give the runner more storage or something?

There's plenty (well, somewhat) of storage, but there are also many different partitions.  The one where `CARGO_HOME` is mounted out-of-the-box at the moment is small:

```console
sandbox:${WORKSPACE} # df -h ${CARGO_HOME}
Filesystem                Size      Used Available Use% Mounted on
overlay                   1.0G         0      1.0G   0% /opt/x86_64-linux-musl
```

> Anyway, what changed, the previous version of Pixi built no problem last week?

Dependencies may have changed?  If this was already close to filling the 1 GB partition you don't need many more binaries being built in `CARGO_HOME` to overflow the partition.

> If be surprised if the Pixi versions were so different as to cause this.

Again, this is clearly not a problem with Pixi itself since the error is happening in a dependency.

giordano:
If someone wanted to fix this issue, I think the solution should be along these lines:

* change where [`CARGO_HOME`](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/be6a14c6dd412549bae717e0f8d038720e88ea9a/src/Runner.jl#L1234) points to (I haven't thought of a good place though)
* around https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/be6a14c6dd412549bae717e0f8d038720e88ea9a/src/Runner.jl#L1424-L1431, `cargo_dir = mktempdir()` is fine, but I believe it should be mounted in `envs["CARGO_HOME"]` or its parent dir, not sure, needs testing.  The call to `cargo_config_file!` should also be fine, I think
* I have the vague recollection that `${CARGO_HOME}/bin` should be in `PATH`, which may be the motivation why `${CARGO_HOME}` currently points to `/opt/$(host_target)`: `/opt/$(host_target)/bin` is already in `${PATH}` so you get that for free.  I'm not entirely sure of this, needs some testing, but if a cargo build freaks out because it can't find executables that are in `${CARGO_HOME}/bin` then the other step is definitely to add that directory to `PATH`

ViralBShah:
Superseded by #13154

PR #10604, opened 2025-02-23
Add Jinja2Cpp wrapper

AlexKlo:
This PR introduces a minimal wrapper for Jinja2Cpp with only the necessary files:  

- Added `/include` with required headers  
- Added `/src` with C++ implementation  
- Added `CMakeLists.txt` for building the wrapper  
- Added `build_tarballs.jl` for BinaryBuilder


imciner2:
I would suggest that you maintain a separate git repository with the wrapper C++ code instead of keeping it here in Yggdrasil. That will allow you to better develop/maintain it, and keep the history here related to only version updates.

AlexKlo:
@imciner2 
I would prefer to keep this code here for now because it’s not a complete wrapper yet, but rather just a part of the functionality. At the moment, I need to register a JLL with this functionality.

giordano:
I'm also not a big fan of hosting external code here, especially not this much, let's merge this for now, but I ***warmly*** recommend setting up an external repo with the code and use that as source for the next time.

AlexKlo:
Thanks, I will do this in the future!

AlexKlo:
#10643

PR #10612, opened 2025-02-24
[CUDSS] Upgrade to v0.5

amontoison:


amontoison:
@maleadt @giordano 
The last release of `cuDSS` has a new library `libcudss_mtlayer_*` and on Windows, it is linking to `vcomp140.dll`.
Should we download this library and add it in the tarball for Windows or can we expect the user to find it at runtime?

giordano:
I don't know this library, but if it's a windows system library it's fair to expect users to have it, **_especially_** if it's a proprietary library we can't redistribute.

amontoison:
I checked online and it's not a windows system library.
It is a library of `Microsoft Visual C++ Redistributable` and we are not allowed to redistribute only one library.

amontoison:
Good news, we can do our own version with `libgomp` on Windows:
https://docs.nvidia.com/cuda/cudss/advanced_features.html#multi-threaded-mt-mode

I will add it later if some users request it.

Franc-Z:
Hope that we could use 0.5.0 on linux as soon as possible, since the hybrid execution mode is useful.

amontoison:
Check the opened PR in CUDSS.jl.
I have some errors with refactorization.

PR #10616, opened 2025-02-25
MAGEMin v1.7.0

NicolasRiel:
- Added option to use initial guess when computing phase diagrams
- Corrected recently introduced mistake that was leading to wrong values for Vp_S and Vs_S
- Improved computational time for latent heat of reaction calculation
- Slightly shifting compositional variables bounds to smaller values (1e-8)


NicolasRiel:
@giordano Thank you very much!

PR #10623, opened 2025-02-26
[CSL@1.3] Increment patch version

ararslan:
This rebuilds CSL 1.3 to use the updated FreeBSD versions (13.4 on x86-64 and 14.1 on AArch64). Let me know if previous CSL versions should also be rebuilt, or if this requires a new version altogether rather than a rebuild.


giordano:
We can change the version number here (e.g. bump patch), it's all made up anyway

ararslan:
Are there (dis)advantages to that vs. bumping the build number? Happy to do whatever makes sense. Also, your comment suggests to me that it's not necessary to rebuild prior CSL versions, is that accurate?

giordano:
Pkg doesn't like build numbers, it's nice to be able to actually distinguish between different versions when possible

ararslan:
Okie doke, I've changed it to bump the patch version instead of the build number

PR #10625, opened 2025-02-26
[LibCURL] Add back OpenSSL platform filter which was accidentally dro…

jeremiahpslewis:
…pped


giordano:
You need to touch build_tarballs.jl to actually build something

L/LibCURL/LibCURL@8/build_tarballs.jl
```diff
@@ -1,3 +1,5 @@
 include("../common.jl")
 
 build_libcurl(ARGS, "LibCURL", v"8.12.1")
+
+# trigger
```
giordano, line 5: Missing newline
```suggestion
# trigger

```

PR #10627, opened 2025-02-26
[Zlib@1.3.1] Trigger a rebuild with updated PlatformSupport

ararslan:
FreeBSD 13.2 is now 13.4 on x86-64 and 14.1 on AArch64

### DIFF
--- a/Z/Zlib/Zlib@1.3.1/build_tarballs.jl
+++ b/Z/Zlib/Zlib@1.3.1/build_tarballs.jl
@@ -5,4 +5,4 @@ include("../common.jl")
 build_tarballs(ARGS, configure_zlib_build(version)...;
                julia_compat="1.9", preferred_llvm_version=llvm_version)
 
-# build trigger: 1
+# build trigger: 2


giordano:
Is this really _needed_?  Aren't the existing binaries installable anyway?

ararslan:
> Aren't the existing binaries installable anyway?

Yes, they should be installable regardless of whether this is rebuilt.

> Is this really _needed_?

Maybe, maybe not. For reference, binaries built for distribution with FreeBSD's package manager are built separately for each supported FreeBSD release series, i.e. everything has separate binaries for 13.x and 14.x, including all linked dependencies. While binaries are generally forward-compatible across major releases, there have been cases where things built on an older release just don't work correctly on current releases, which has generally been the motivator for each RootFS change we've made (11.1 -> 12.2 -> 13.2 -> [13.4, 14.1]). I've rebuilt the entirety of Julia's binary dependency tree for each of those prior changes to ensure there are no weird issues that arise due to mismatched major OS versions. A rebuilt MPFR for FreeBSD 14.1 might (hopefully...) resolve the `exp10` issue you and I had observed, or maybe it needs all of its dependencies rebuilt as well in order to resolve the issue. There are also a number of other test failures unrelated to MPFR; perhaps those could also be fixed by having dependencies rebuilt. Having a common base state for dependencies at least lets you rule out one source of potential error when issues arise.

giordano:
> Maybe, maybe not. For reference, binaries built for distribution with FreeBSD's package manager are built separately for each supported FreeBSD release series, i.e. everything has separate binaries for 13.x and 14.x, including all linked dependencies.

I'm not surprised, also Linux distributions mass rebuild everything for every release.  But for us rebuilding _everything_, every time, without a clear _need_ to do it, is a major churn.

> A rebuilt MPFR for FreeBSD 14.1 might (hopefully...) resolve the `exp10` issue you and I had observed, or maybe it needs all of its dependencies rebuilt as well in order to resolve the issue.

Maybe start from MPFR?

ararslan:
Actually, it looks like MPFR's only dependency is GMP (which I seem to recall I also had some weird issue with on FreeBSD AArch64 but it magically resolved itself, maybe) and GMP has no dependencies, so I'll start at GMP and then do MPFR.

PR #10634, opened 2025-02-27
[XRT] Add versioning and recipe for v2.16 to increase compatibility

DO6LTM:
Another problem with XRT: We need to maintain multiple versions because some cards are not supported since XRT 2.17 (2024.1) (see [Release Notes](https://docs.amd.com/r/2024.1-English/ug1451-xrt-release-notes/New-Features-and-Changes)).
I have moved the 2.17 recipe and patches into an own directory and adapted them to build working 2.16 artifacts.

Relevant changes for the build script in my opinion:

- Add Xilinx DMA IP driver sources to build
- Include OpenCL C++ headers
- boost is required when loading the package
- We can build with GCC 8
- I set the --prefix flag at the install command, so that the builder is able to find the products


fingolfin (commented):
Sorry for overlooking this since February :-(. I've now merged `master` into this to force a fresh run of CI.

X/xrt/xrt@2.17/bundled/patches/windows/aligned_malloc.patch
```diff
@@ -0,0 +1,21 @@
+--- a/src/runtime_src/core/common/memalign.h
++++ b/src/runtime_src/core/common/memalign.h
+@@ -21,6 +21,18 @@
+ #include <memory>
+ #include <stdexcept>
+ 
++// Add memalgin definitions here since
```
fingolfin, line 7: Do you mean
```suggestion
+// Add memalign definitions here since
```
? Though the code then defines `_aligned_malloc` so it's still confusing,,,

DO6LTM: Thank you for the review. I've adopted this change from our previous version 2.17 (commit 62d98d6) since there were no changes to `memalign.h` in this update. So I think we'll be safe when we keep those definitions.

imciner2:
@DO6LTM, sorry this seems to have slipped off our radar. there are some conflicts with the XRT build_tarballs. Can you please rebase this to the latest commit (it needs to pick up the new CI configuration)and resolve the conflicts?

PR #10635, opened 2025-02-27
[GAP_pkg_juliainterface] Update to avoid on-the-fly recompilation

lgoettgens:
ping @fingolfin


fingolfin:
There is a build failure on aarch that looks like a timeout. Could someone (@giordano, @staticfloat, ..?) please restart that job?

Alternatively @lgoettgens you could force push an update I guess

lgoettgens:
Thanks! CI is green now

PR #10637, opened 2025-02-27
[MPFR] Trigger a rebuild for updated PlatformSupport

ararslan:
I've observed odd behavior with GMP and MPFR (the latter more recently) on FreeBSD 14.1 AArch64 with binaries built for 13.2. Now that the FreeBSD version for AArch64 has been updated to 14.1 and GMP 6.3.0 has been rebuilt atop that, let's rebuild MPFR to see whether oddities are resolved.

### DIFF
--- a/M/MPFR/build_tarballs.jl
+++ b/M/MPFR/build_tarballs.jl
@@ -55,4 +55,4 @@ build_tarballs(ARGS, name, version, sources, script, platforms, products, depend
                preferred_gcc_version=v"5", preferred_llvm_version=llvm_version,
                julia_compat="1.6")
 
-# Build trigger: 1
+# Build trigger: 2


giordano:
We can probably test this before merging?

ararslan:
> Now that [...] GMP 6.3.0 has been rebuilt

Except MPFR is using GMP 6.2.1, which I did not rebuild. 🤦 I rebuilt 6.3.0 because it's what Julia is using.

> We can probably test this before merging?

Yes, I had intended to, though it's not like it would hurt anything to merge it regardless.

giordano:
> Except MPFR is using GMP 6.2.1, which I did not rebuild. 🤦 I rebuilt 6.3.0 because it's what Julia is using.

The ABI of the dep library at build time shouldn't matter too much in principle.  On macOS you don't even use real libraries but stub files (they're plain text files with list of symbols, end), what matters at runtime is the library you load...at runtime.

fingolfin (approved):
This seems OK to me?

giordano:
It's useless though.

giordano:
Superseded by #10805

PR #10639, opened 2025-02-27
[MbedTLS] Enable release builds

xal-0:
I payed for 1000 mbps and I want to use it


giordano (commented):
I don't think we need to rebuild all versions, maybe only 2.28.2 (used in Julia v1.10) and 2.28.6 (used in Julia v1.11), all other versions are associated to unsupported versions of Julia.

M/MbedTLS/MbedTLS@2.28.2/build_tarballs.jl
```diff
@@ -3,4 +3,4 @@ include("../common.jl")
 
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.9")
```
giordano, line 4: Same as the 2.28.6 (this is needed in particular for the sanitize build)
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
               julia_compat="1.9", preferred_llvm_version=llvm_version)
```

PR #10661, opened 2025-03-01
[MozillaCACerts] Update 2025-02-25

IanButterworth:
Latest from https://curl.se/docs/caextract.html


giordano:
The hash is also available at the page you linked https://curl.se/docs/caextract.html

PR #10662, opened 2025-03-01
[GTK4] update to 4.16.7

jwahlstrand:


G/GTK4/build_tarballs.jl
```diff
@@ -32,6 +34,13 @@ rm ${prefix}/lib/pkgconfig/gio-2.0.pc
 FLAGS=()
 if [[ "${target}" == *-apple-* ]]; then
     FLAGS+=(-Dx11-backend=false -Dwayland-backend=false)
+
+    # Install a newer SDK to work around compilation failures
```
giordano, line 38: Newer? For aarch64-darwin this is older, default is v11.0: https://github.com/JuliaPackaging/Yggdrasil/blob/d3c0b805a0874c2f2fea8b1d77deeaf7ce0da97c/0_RootFS/gcc_sources.jl#L290-L299

giordano: Also, for future reference can add a comment to explain what's the issue you're trying to fix?

PR #10664, opened 2025-03-02
MAGEMin v1.7.1

NicolasRiel:
- Corrected trace-elements output
- Corrected legacy solver that was leading to unsatisfied bulk-rock composition in some cases
- Largely improved use of initial guess to compute phase diagrams


NicolasRiel:
Thank you! like always @giordano

PR #10668, opened 2025-03-04
Update ORTools_jll to 9.12

dourouc05:
Release notes: https://github.com/google/or-tools/releases/tag/v9.12


dourouc05:
Question for maintainers: can I skip https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/be6a14c6dd412549bae717e0f8d038720e88ea9a/src/Runner.jl#L251-L259? Abseil requires some ARM extensions for cryptography, which leads to the use of `-march=armv8-a+crypto` (https://github.com/abseil/abseil-cpp/blob/bdbaf376e297978a8d4206a5a860c88e5b2b2428/absl/random/internal/BUILD.bazel#L99). Our internal build process generates binaries for these ARM platforms and users don't complain about them (https://github.com/google/or-tools/releases/tag/v9.12).

Thanks!

giordano:
> Abseil requires some ARM extensions for cryptography

Do I understand correctly that this package wouldn't work anyway on baseline armv8-a?

dourouc05:
For Apple machines, it will work. For other targets, well, the binaries run without trouble within Qemu and an aarch64 Bootlin (https://toolchains.bootlin.com/).

giordano:
Add `lock_microarchitecture=false` to the arguments of `build_tarballs(...)`

dourouc05:
Once that check is lifted, compilation scripts shine by mixing host and target tool chains. Would you be open to the idea of simply installing prebuilt packages instead of working around issues in undocumented scripts?

O/ORTools/build_tarballs.jl
```diff
@@ -1,18 +1,25 @@
...
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
+# OR-Tools 9.12 starts requiring CMake 3.24+.
+wget -q "https://cmake.org/files/v3.31/cmake-3.31.0-linux-x86_64.sh"
+chmod a+x cmake-3.31.0-linux-x86_64.sh
+./cmake-3.31.0-linux-x86_64.sh --prefix=/usr/local/ --skip-license
+rm cmake-3.31.0-linux-x86_64.sh
+export PATH=/usr/local/bin/:$PATH
```
giordano, line 21: Why not using `CMake_jll` as host depdency (which contrary to this installation method can be cached)?

dourouc05: Apparently, this installation of CMake doesn't work, I can't do `cmake --version` on FreeBSD (https://buildkite.com/julialang/yggdrasil/builds/18478#01956e72-c43b-46f7-bd4a-441fe26f4e19). The change seems similar to what others do to have a recent version of CMake (like https://github.com/dourouc05/Yggdrasil/blob/5ca97400bc9d37ef30ad4e5b58cc711cb1f01cf7/C/Catalyst/build_tarballs.jl, among the simple scripts that update CMake).

O/ORTools/build_tarballs.jl
```diff
@@ -161,7 +164,15 @@ products = [
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[]
+dependencies = [
+    # OR-Tools 9.12 starts depending on CMake 3.28.
+    # OR-Tools 9.12 starts depending on CMake 3.28.
+    BuildDependency(PackageSpec(; name="CMake_jll", version = v"3.28.1"))
```
giordano, line 170: https://docs.binarybuilder.org/stable/build_tips/#Dependencies-for-the-target-system-vs-host-system

giordano: And to be extra explicit: https://github.com/JuliaPackaging/Yggdrasil/blob/ddf863dea81a134cba44f5005fce32d985c7ada3/S/SuiteSparse/common.jl#L94

PR #10682, opened 2025-03-05
[ITKIOWrapper] updated dependency version

divital-coder:
Updated upstream Wrapper package to use ITK_jll with new HDF5 version.


giordano (dismissed):
I feel like I may have repeated this multiple times before, but here we go again: changing compat bound ***always*** requires changing the version number.

divital-coder:
Sorry...

divital-coder:
Does this look good?

PR #10690, opened 2025-03-06
[Singular] default to sequential algorithms

hannes14:
@fingolfin


fingolfin:
has this been tested against the Oscar master resp. 1.3.0 test suites to make sure nothing is broken?

hannes14:
tested against Oscar/runtests and doctests

PR #10701, opened 2025-03-07
[FLEXPART] update to v11.0.0

tcarion:
Update to [FLEXPART v11](https://www.flexpart.eu/) release


F/FLEXPART/bundled/patches/flexpart-makefile.patch
```diff
@@ -1,39 +1,31 @@
...
--           -fconvert=little-endian -frecord-marker=4 -fmessage-length=0 \
--           -flto=jobserver -O$(O_LEV) $(NCOPT) $(FUSER)  #-Warray-bounds -fcheck=all # -march=native
-+FFLAGS   = -I$(INCPATH1) -O$(O_LEV) -cpp -mcmodel=large -fconvert=little-endian -frecord-marker=4 -fmessage-length=0 -O$(O_LEV) $(NCOPT) $(FUSER)
+ #F90 = /usr/bin/gfortran
+@@ -87,13 +89,13 @@ O_LEV_DBG = g # [0,g]
+ # (defined by the machine where you are compiling it)
+ #
+ ifeq ($(arch), generic)
+-	FUSER = -g -fopenmp -march=core-avx2
++	FUSER = -g -fopenmp
```
barche, line 19: I would also remove the `-g` flag here, adding debug info isn't that useful in the release build here.

barche:
Thanks for this PR! To make the aarch64 build work the `-m64` flag needs to be removed from `FFLAGS` in the makefile you are already patching, if I remember correctly this is set automatically on x86-64 anyway and so not needed.

tcarion:
Thanks for the help @barche!
I now have this `Array specification at (1) has more than 7 dimensions` for glibfortran4 platforms. I'm tempted to drop these builds, unless you have a better idea?

barche:
Yes, the only way to solve this is to drop libgfortran4, since support for higher dimensions was only added in gfortran 8, which already uses libgfortran5.

PR #10706, opened 2025-03-08
Reactant 0.0.84

wsmoses:
x/ref https://github.com/EnzymeAD/Reactant.jl/pull/856

essentially we want future precompiled libs to find XLAExecuteSharded and friends, but not use our LLVM symbols [which conflict].

We are explicitly okay here for macos because of the filter. Linux does the right thing already

@avik-pal

### DIFF
--- a/R/Reactant/build_tarballs.jl
+++ b/R/Reactant/build_tarballs.jl
@@ -6,7 +6,7 @@ include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
 
 name = "Reactant"
 repo = "https://github.com/EnzymeAD/Reactant.jl.git"
-version = v"0.0.83"
+version = v"0.0.84"
 
 sources = [
   GitSource(repo, "e191da45854a8f103af2a858805ecbe983aed2fd"),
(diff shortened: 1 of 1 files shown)


R/Reactant/build_tarballs.jl
```diff
@@ -439,7 +434,15 @@ for gpu in ("none", "cuda"), mode in ("opt", "dbg"), cuda_version in ("none", "1
...
 
     should_build_platform(triplet(augmented_platform)) || continue
-    products2 = copy(products)
+	
+    # The products that we will ensure are always built
+    products2 = Product[]
+    if Sys.isapple(platform)
+	push!(products2, LibraryProduct(["libReactantExtra", "libReactantExtra"], :libReactantExtra))
+    else
+	push!(products2, LibraryProduct(["libReactantExtra", "libReactantExtra"], :libReactantExtra, dlopen_flags=[:RTLD_GLOBAL,:RTLD_DEEPBIND]))
```
giordano, line 443: Can you elaborate on why `RTLD_GLOBAL` helps?  Also, `RTLD_GLOBAL` is a bit problematic because it makes symbols available to everyone, and this library ships loads of symbols.

wsmoses: It’s needed for Linux with precompiled, since the precompiled so can’t find the symbol otherwise in my PR. I confirmed loading like this fixes locally

PR #10707, opened 2025-03-08
[aws_c_http_jq] Add aws_c_http_jq fork of aws_c_http recipe

quinnj:
I have functionality implemented in a fork of aws-c-http that is [stalled upstream](https://github.com/awslabs/aws-c-http/pull/500). The maintainers haven't given a [strong signal](https://github.com/awslabs/aws-c-http/discussions/493#discussioncomment-11378308) that they'll even be open to accepting the functionality or continue to maintain the overall server-side support, so I'm proactively forking the library with the idea that one of two scenarios plays out:
1. My contribution remains stalled or even rejected as the maintainers want to remove support for server-side. In this case, HTTP.jl would want to depend on a fork that _keeps_ server-side support, so this fork would be used long-term (we'd continue to merge contributions from the main project in as it applied to client-side).
2. They review, accept, and merge my functionality, and continue to be open to contributions to maintain server-side code; HTTP.jl would then be able to change the planned dependency from aws_c_http_jq_jll to aws_c_http_jll with minimal disruption and this proposed fork would be stale unless needed for other forked functionality in the future.

Release-wise, I'm proposing to start at the 0.9.3 release, which is the latest aws-c-http release, but that also includes my forked contribution. In that way, we could maintain compatibility with upstream releases when they come. If new forked contributions were needed, we could possibly wait until a new upstream release was made, or do an immediate release and maintain an N+M release schedule where this fork would always be M versions (# of fork-only releases) ahead of hte upstream version (N).

I'm open to concerns about this approach and willing to discuss. I'm also just trying to unblock the new HTTP.jl internals-rewrite I've been working on for a while and want to make some progress moving forward. It needs this forked functionality I worked on in an officially published binary, and trying to swap out the existing aws-c-http binary recipe with my fork would seem to have its own set of concerns.

### DIFF
--- a/A/aws_c_http_jq/build_tarballs.jl
+++ b/A/aws_c_http_jq/build_tarballs.jl
@@ -1,0 +1,47 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "aws_c_http_jq"
+version = v"0.9.3"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/quinnj/aws-c-http.git", "048eb767e87b4b5dd94e12ea50281e4a942ca708"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/aws-c-http
+
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_PREFIX_PATH=${prefix} \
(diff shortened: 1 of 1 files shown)


giordano:
I'm in principle ok with this, my concern is only on whether it's sustainable for you to maintain a fork of a project you aren't (I presume?) exceedingly familiar with but only an end user

quinnj:
I'm no unconfident I could maintain the fork; I've already fixed a fairly critical memory bug and made a non-trivial contribution on my fork to provide server-side support for websockets. I'm fairly familiar with a lot of the code, but C isn't my strongest language, so it's certainly more effort to make changes there. I anticipate most of the effort required will be bringing in upstream fixes/changes and the occasional HTTP.jl-required change, both of which I feel comfortable with.

PR #10713, opened 2025-03-10
[GEOS] Update to v3.13.1

jaakkor2:
Release notes https://github.com/libgeos/geos/releases/tag/3.13.1

### DIFF
--- a/G/GEOS/build_tarballs.jl
+++ b/G/GEOS/build_tarballs.jl
@@ -3,13 +3,13 @@
 using BinaryBuilder
 
 name = "GEOS"
-version = v"3.13.0"
+version = v"3.13.1"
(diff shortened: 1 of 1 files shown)


G/GEOS/build_tarballs.jl
```diff
@@ -3,13 +3,13 @@
 using BinaryBuilder
 
 name = "GEOS"
-version = v"3.13.0"
+version = v"3.13.1"
 ygg_version = v"3.13.1" # bump ygg version to pick up new platforms
```
giordano, line 7:
```suggestion
```
and adapt below

giordano: > and adapt below

:point_up:

PR #10721, opened 2025-03-11
neper: Set NLopt compat bounds

KnutAM:
Running ``run(`$(neper()) -T -n 10`)`` fails with 
`dyld[7307]: Library not loaded: @rpath/libnlopt.0.dylib` using default installation, but solved by `add NLopt_jll@2.9` before loading `neper_jll`.

I'm not sure if this is the correct way to fix it, but from what I could find, 
the update from NLopt_jll from v2.9 to v2.10 renames, e.g., `libnlopt.so.0` to `libnlopt.so.1`: https://github.com/JuliaBinaryWrappers/NLopt_jll.jl/commit/28cc85df871ea6eaa5a8de90dfb7ee9858b322a5

Since I'm not sure how I should verify that this change is the correct, this PR needs to be verified by someone who knows - tagging those who have worked on this file: @fredrikekre @amontoison @kimauth

Also not sure why the compat bounds says no versions left, shouldn't this correspond to `[2, 2.10)`?

### DIFF
--- a/N/neper/build_tarballs.jl
+++ b/N/neper/build_tarballs.jl
@@ -45,7 +45,7 @@ dependencies = [
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
     Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
     Dependency(PackageSpec(name="GSL_jll", uuid="1b77fbbe-d8ee-58f0-85f9-836ddc23a7a4"); compat="~2.7.2"),
-    Dependency(PackageSpec(name="NLopt_jll", uuid="079eb43e-fd8e-5478-9966-2cf3e3edb778"); compat="2 - 2.10"),
+    Dependency(PackageSpec(name="NLopt_jll", uuid="079eb43e-fd8e-5478-9966-2cf3e3edb778"); compat="2.6.2 - 2.9"),
     Dependency(PackageSpec(name="SCOTCH_jll", uuid="a8d0f55d-b80e-548d-aff6-1a04c175f0f9"); compat="6.1.3"),
     RuntimeDependency(PackageSpec(name="gmsh_jll", uuid="630162c2-fc9b-58b3-9910-8442a8a132e6")),
 ]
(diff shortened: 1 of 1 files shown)


giordano:
You need to fix this in the registry, not here

KnutAM:
Thanks for the fast response! 

> You need to fix this in the registry, not here

Sorry, I don't think I understand: Do you mean directly editing https://github.com/JuliaRegistries/General/blob/master/jll/N/neper_jll/Compat.toml? 

I thought that was updated based on the compat entries `Project.toml` file generated by `BinaryBuilder.jl` when releasing a patch from this repo?

giordano:
> Do you mean directly editing [JuliaRegistries/General@master/jll/N/neper_jll/Compat.toml](https://github.com/JuliaRegistries/General/blob/master/jll/N/neper_jll/Compat.toml?rgh-link-date=2025-03-11T10%3A27%3A44Z)?

Yes.

> I thought that was updated based on the compat entries in the Project.toml file generated by BinaryBuilder.jl when releasing a patch from this repo?

Only for newer version, but BinaryBuilder doesn't retroactively fix previous versions, there's no point in releasing a new version and having old versions with inconsistent broken compat bounds anyway, allowing the package manager to install broken combinations (it's actually possible the package manager could favour that combination anyway, so that by releasing a new version here doesn't achieve anything).

N/neper/build_tarballs.jl
```diff
@@ -45,7 +45,7 @@ dependencies = [
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
     Dependency(PackageSpec(name="LLVMOpenMP_jll", uuid="1d63c593-3942-5779-bab2-d838dc0a180e"); platforms=filter(Sys.isbsd, platforms)),
     Dependency(PackageSpec(name="GSL_jll", uuid="1b77fbbe-d8ee-58f0-85f9-836ddc23a7a4"); compat="~2.7.2"),
-    Dependency(PackageSpec(name="NLopt_jll", uuid="079eb43e-fd8e-5478-9966-2cf3e3edb778")),
+    Dependency(PackageSpec(name="NLopt_jll", uuid="079eb43e-fd8e-5478-9966-2cf3e3edb778"); compat="2 - 2.10"),
```
giordano, line 48:
```suggestion
    Dependency(PackageSpec(name="NLopt_jll", uuid="079eb43e-fd8e-5478-9966-2cf3e3edb778"); compat="2.6.2 - 2.9"),
```

giordano:
I'm going to merge this but _without_ triggering a new version because, as I said above, that's useless.  This is merely for reflecting here the change in https://github.com/JuliaRegistries/General/pull/126694 for consistency.

KnutAM:
Thanks for solving this @giordano!

PR #10725, opened 2025-03-11
[nghttp2] Bump to 1.65.0

fxcoudert:
New version: https://github.com/nghttp2/nghttp2/releases/tag/v1.65.0

### DIFF
--- a/N/nghttp2/nghttp2@1.65/build_tarballs.jl
+++ b/N/nghttp2/nghttp2@1.65/build_tarballs.jl
@@ -1,0 +1,8 @@
+version = v"1.65.0"
+
+include("../common.jl")
+
+build_tarballs(ARGS, configure_nghttp2_build(version)...;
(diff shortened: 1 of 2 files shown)


N/nghttp2/nghttp2@1.65/build_tarballs.jl
giordano, line 1: Uhm, I think for julia's dependencies like this one we're duplicating the recipes rather then moving them, occasionally we need to go back and rebuild older versions.

fxcoudert: OK, that's new right? I don't remember that before (only for major versions). I'll update the PR.

giordano: Yeah, recently we had to do these rebuilds a lot and we started this more frequently for julia's deps.

PR #10733, opened 2025-03-12
[GTPSA] bump version

mattsignorelli:


ViralBShah:
Is freebsd still a problem, or should we try it again?

mattsignorelli:
@ViralBShah I have not made any changes to the build script/cmake that would fix it, nor have checked... but unless anything with BinaryBuilder/FreeBSD workspace has changed then I expect it would still be a problem

PR #10737, opened 2025-03-13
Add PlantUML 1.2025.2

tz-lom:


tz-lom:
@ViralBShah Can you review it? 😁 

I need it for 
https://github.com/tz-lom/DocumenterPlantUML.jl
and
https://github.com/tz-lom/PlantUML.jl

imciner2:
Hmm, we don't seem to have any other Java-based things in Ygg currently, so I am not sure how well the JLL system is setup to work with these. We don't have any sort of Java distribution in here to ensure dependencies are met to run it, so this seems somewhat difficult to distribute here in Ygg vs. as an artifact downloaded by packages that need it.

@giordano thoughts?

giordano:
Yeah, I don't see any benefit in using BinaryBuilder.

fingolfin:
There are no build steps, so the result will be identical on all platforms -- so it should be using `AnyPlatform()` if it was an Ygg package.

But as was stated here, there seems to be no point in doing so, and none was provided in the past 8 months... so I am going to close this (but feel free to appeal with an explanation)

PR #10745, opened 2025-03-14
[Singular] Rebuild for FLINT 3.2.0

lgoettgens:
cc @benlorenz @fingolfin


lgoettgens:
According to https://github.com/oscar-system/Oscar.jl/issues/4732, this should wait until the rebuilds for the polymake dependency tree are a lot more progressed.

PR #10746, opened 2025-03-14
[normaliz] Rebuild with FLINT 3.2.0

lgoettgens:
cc @benlorenz

### DIFF
--- a/N/normaliz/build_tarballs.jl
+++ b/N/normaliz/build_tarballs.jl
@@ -69,12 +69,13 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("GMP_jll", v"6.2.0"),
+    Dependency("GMP_jll", v"6.2.1"),
     Dependency("MPFR_jll", v"4.1.1"),
+    Dependency("OpenBLAS32_jll", v"0.3.28"),
     Dependency("FLINT_jll"; compat = "~300.200.0"),
-    Dependency("nauty_jll"; compat = "~2.6.13"),
-    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD
-    # systems), and libgomp from `CompilerSupportLibraries_jll` everywhere else.
+    Dependency("nauty_jll"; compat = "~2.8.10"),
+    # For OpenMP we use libomp from `LLVMOpenMP_jll` where we use LLVM as compiler (BSD systems),
+    # and libgomp from `CompilerSupportLibraries_jll` everywhere else.
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"); platforms=filter(!Sys.isbsd, platforms)),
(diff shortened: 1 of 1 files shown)


lgoettgens:
The powerpc failure looks identical to the one in https://github.com/JuliaPackaging/Yggdrasil/pull/10642. Let's wait for them there to find a solution.

N/normaliz/build_tarballs.jl
```diff
@@ -71,8 +71,9 @@ products = [
 dependencies = [
     Dependency("GMP_jll", v"6.2.0"),
     Dependency("MPFR_jll", v"4.1.1"),
-    Dependency("FLINT_jll"; compat = "~300.100.300"),
-    Dependency("nauty_jll"; compat = "~2.6.13"),
+    Dependency("OpenBLAS32_jll", v"0.3.28"),
+    Dependency("FLINT_jll"; compat = "~300.200.0"),
+    Dependency("nauty_jll"; compat = "~2.8.9"),
```
lgoettgens, line 76:
```suggestion
    Dependency("nauty_jll"; compat = "~2.8.10"),
```
Bump nauty again to fix aarch64-freebsd

N/normaliz/build_tarballs.jl
```diff
@@ -71,8 +71,9 @@ products = [
 dependencies = [
     Dependency("GMP_jll", v"6.2.0"),
```
lgoettgens, line 72:
```suggestion
    Dependency("GMP_jll", v"6.2.1"),
```

lgoettgens:
According to https://mailman.anu.edu.au/pipermail/nauty/2024-August/000985.html, there have been some changes in the nauty build system between 2.8.8 and 2.8.9. Normaliz states that is compatible to 2.8.8 (https://github.com/Normaliz/Normaliz/blob/f2f100fae3f30ceffc2bc010454bca984faa281a/INSTALL#L15).

The handful failures are all due to
```
checking whether nauty headers and library are available (nauty/nauty.h)... checking for nauty/nauty.h... no
checking whether nauty headers and library are available (nauty.h)... checking for nauty.h... no
configure: error: nauty is not available but was requested
```
which hints at me that normaliz hasn't adapted to the build system changes in nauty.

Any ideas on how to proceed / fix / workaround are very welcome.

fingolfin:
As a normaliz maintainer I guess I should look into creating a patch for it to work with newer nauty (as a PR for normaliz but ideally we could backport it...) I'll give it a quick spin but I also have other things to do today...

fingolfin:
Indeed, normaliz builds just fine for me with nauty 2.8.9.

(BTW Normaliz 3.10.4 is the current version)

Anyway so the mailing list discussion @benlorenz linked suggests a fix for the nauty headers. So I guess we next need to update `nauty_jll` with that (or a similar) patch

lgoettgens:
> Anyway so the mailing list discussion @benlorenz linked suggests a fix for the nauty headers. So I guess we next need to update `nauty_jll` with that (or a similar) patch

I am trying to build a nauty_jll locally with that patch and then try to build normalz_jll. I'll come back (and open a PR for natuy_jll) once that succeeded.

PR #10753, opened 2025-03-15
[nauty] Build for aarch64-freebsd

lgoettgens:
See https://github.com/JuliaPackaging/Yggdrasil/pull/10746#pullrequestreview-2687753820.

cc @benlorenz

### DIFF
--- a/N/nauty/build_tarballs.jl
+++ b/N/nauty/build_tarballs.jl
@@ -129,3 +129,4 @@ dependencies = [
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
 
+# Build trigger: 1
--- a/N/nauty/build_tarballs.jl
+++ b/N/nauty/build_tarballs.jl
@@ -3,11 +3,12 @@
 using BinaryBuilder
 
 name = "nauty"
-version = v"2.8.9" 
+upstream_version = v"2.8.9"
+version = v"2.8.10"
 
 # Collection of sources required to build nauty
 sources = [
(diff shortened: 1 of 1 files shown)


giordano (changes requested):
must change the version number

lgoettgens:
You guys are crazy fast :joy: 

I was still looking into how other recipes handle the bump (without the 100 times version scaling)

lgoettgens:
Everything should be good now

N/nauty/build_tarballs.jl
```diff
@@ -57,6 +58,9 @@ install_license COPYRIGHT
 platforms = supported_platforms(;experimental=true)
 filter!(!Sys.iswindows, platforms)
 
+# Currently skipped due to the following error: `sumlines.c:152:10: fatal error: gmp.h: No such file or directory`
+filter!(p -> arch(p) != "riscv64", platforms)
```
giordano, line 62: Uhm, we do have GMP for riscv64

giordano: Use v6.2.1

N/nauty/build_tarballs.jl
```diff
@@ -123,9 +124,12 @@ products = [
...
+    Dependency("GMP_jll", v"6.2.1"), # for sumlines
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="1.6",
+               preferred_gcc_version=v"6")
 
+# Build trigger: 1
```
lgoettgens, line 135:
```suggestion
```

N/nauty/build_tarballs.jl
lgoettgens: Works, thanks!

giordano (approved):

PR #10755, opened 2025-03-17
MAGEMin v1.7.2

NicolasRiel:
- Re-added the use of pure-endmember swapping during levelling stage to improve accuracy of minimizations
- Further improves initial guess to compute phase diagrams


NicolasRiel:
@giordano Thank you very much!

PR #10771, opened 2025-03-18
MAGEMin 1.7.3

NicolasRiel:
- Dropping out support for MPI, which is never used with the Julia wrapper and was creating conflict with LaMEM.
- This new version will allow to easily call LaMEM and MAGEMin_C at the same time to compute stable phase equilibrium from passives tracers

### DIFF
--- a/M/MAGEMin/build_tarballs.jl
+++ b/M/MAGEMin/build_tarballs.jl
@@ -42,7 +42,7 @@ install_license LICENSE
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = supported_platforms((; exclude=p->Sys.isfreebsd(p) && arch(p) == "aarch64"))
 platforms = expand_gfortran_versions(platforms)
 platforms = filter(p -> !(arch(p)  == "riscv64"), platforms)
 platforms = filter(p -> !( (libgfortran_version(p) == v"3" || libgfortran_version(p) == v"4") && arch(p)=="powerpc64le"), platforms)
--- a/M/MAGEMin/build_tarballs.jl
+++ b/M/MAGEMin/build_tarballs.jl
@@ -16,9 +16,7 @@ sources = [GitSource("https://github.com/ComputationalThermodynamics/MAGEMin",
 script = raw"""
 cd MAGEMin*
 
-USE_MPI ?= 0
-
-CCFLAGS="-DUSE_MPI -O3 -g -fPIC -std=c99"
+CCFLAGS="-DUSE_MPI=0 -O3 -g -fPIC -std=c99"
 
 if [[ "${target}" == *-apple* ]]; then 
     # Use Accelerate for Lapack dependencies
--- a/M/MAGEMin/build_tarballs.jl
+++ b/M/MAGEMin/build_tarballs.jl
@@ -16,7 +16,9 @@ sources = [GitSource("https://github.com/ComputationalThermodynamics/MAGEMin",
 script = raw"""
 cd MAGEMin*
 
-CCFLAGS="-DUSE_MPI=0 -O3 -g -fPIC -std=c99"
+USE_MPI ?= 0
+
+CCFLAGS="-DUSE_MPI -O3 -g -fPIC -std=c99"
 
 if [[ "${target}" == *-apple* ]]; then 
     # Use Accelerate for Lapack dependencies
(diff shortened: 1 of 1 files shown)


M/MAGEMin/build_tarballs.jl
```diff
@@ -3,47 +3,35 @@
 using BinaryBuilder, Pkg
 using Base.BinaryPlatforms
 const YGGDRASIL_DIR = "../.."
-include(joinpath(YGGDRASIL_DIR, "platforms", "mpi.jl"))
+include(joinpath(YGGDRASIL_DIR, "platforms"))
```
giordano, line 6: You're including a directory?  Maybe you wanted to remove this line entirely?

NicolasRiel: Ah yes ok I see. Sorry I am not fully understanding everything with the binaryBuilder.
I removed the line, thank you very much for your help and patience!

M/MAGEMin/build_tarballs.jl
```diff
@@ -3,47 +3,34 @@
...
     INC="-I${includedir}"
 fi
 
 # Compile library:
-make -j${nproc} CC="${CC}" CCFLAGS="${CCFLAGS}" LIBS="${LIBS}" INC="${INC}" lib
+make CC="${CC}" CCFLAGS="${CCFLAGS}" LIBS="${LIBS}" INC="${INC}" lib
 
 # Compile binary
-make -j${nproc} EXE_NAME="MAGEMin${exeext}" CC="${CC}" CCFLAGS="${CCFLAGS}" LIBS="${LIBS}" INC="${INC}" all
+make EXE_NAME="MAGEMin${exeext}" CC="${CC}" CCFLAGS="${CCFLAGS}" LIBS="${LIBS}" INC="${INC}" all
```
giordano, line 33: Why removing `-j${nproc}`?  That'd tend to make compilation a lot slower.

M/MAGEMin/build_tarballs.jl
```diff
@@ -56,25 +43,14 @@ install_license LICENSE
 
 augment_platform_block = """
     using Base.BinaryPlatforms
-    $(MPI.augment)
-    augment_platform!(platform::Platform) = augment_mpi!(platform)
 """
```
giordano, line 46: I don't think you need this at all if you don't want to link to MPI.  Also, remove from it from the last line too.

M/MAGEMin/build_tarballs.jl
```diff
@@ -56,25 +43,14 @@ install_license LICENSE
...
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = supported_platforms(; exclude=p->Sys.isfreebsd(p) && arch(p) == "aarch64")
```
giordano, line 50: If you're happy to use NLopt 2.9 or 2.10 you have also aarch64-freebsd and riscv64-linux builds:
```suggestion
platforms = supported_platforms()
```

M/MAGEMin/build_tarballs.jl
```diff
@@ -56,25 +43,14 @@ install_license LICENSE
...
 platforms = expand_gfortran_versions(platforms)
-
-platforms, platform_dependencies = MPI.augment_platforms(platforms; MPItrampoline_compat="5.3.1", OpenMPI_compat="4.1.6, 5")
-
-# Avoid platforms where the MPI implementation isn't supported
-# OpenMPI
-platforms = filter(p -> !(p["mpi"] == "openmpi" && arch(p) == "armv6l" && libc(p) == "glibc"), platforms)
 platforms = filter(p -> !(arch(p)  == "riscv64"), platforms)
```
giordano, line 52:
```suggestion
```

M/MAGEMin/build_tarballs.jl
NicolasRiel: I see, I thought this had to do something with MPI here

M/MAGEMin/build_tarballs.jl
NicolasRiel: ok thank you very much

M/MAGEMin/build_tarballs.jl
NicolasRiel: Ah yes, I forgot about NLopt. I was fixing to 2.8.0 because the newer version were yielding different results for different platform. I still need to open an issue for this. For now I am happy to still fix NLopt to 2.8.0, but will remove the constraint in the future

NicolasRiel:
I forgot about NLopt. I was fixing to 2.8.0 because the newest version were yielding different results for different platform. I still need to open an issue on NLopt for this. For now I am happy to still fix NLopt to 2.8.0, but will remove the constraint in the future. Is that fine with you Mose?

giordano:
Ok, yes.

NicolasRiel:
thank you for your help!

PR #10786, opened 2025-03-19
TDLib: try building for macos

aplavin:


giordano:
https://docs.binarybuilder.org/stable/troubleshooting/#General-comments

PR #10790, opened 2025-03-20
Update for new OptimPack 3.3.1 release

emmt:


imciner2:
You really shouldn't be using `-Werror` in code that is being released for others to compile - https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/, because you never know what warnings will be in specific compilers, and some compilers can add new warnings that would then trigger it.

giordano:
Agreed with what Ian said, but it seems like your code does have some problems, there are warnings about use of possibly uninitialised variables.

emmt:
Each version of compilers found different possibly unitialized variables. These warnings are mostly false positive but the compilers make such errors because the code is full of `goto` statements (notably when code wwas proted from FORTRAN to C). Anyway, I want to fix these warnings so I will correct the code as soon as possible. Sorry for not having testing the builds on all supported platforms...

emmt:
I investigated the error and there were indeed a potentially unititialized variables and also a small bug. I have fixed all these problems for all platforms supported by Julia and I am going to make a PR. I do not know if this will update this PR...

emmt:
Well, now I do not understand the errors. It does not seem to be related to a complilation error:

> ERROR: LoadError: IOError: could not spawn `/cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/artifacts/73b10e7f9f6b94c9754056030a3adbcf64bf5b90/sandbox --rootfs /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/artifacts/73b10e7f9f6b94c9754056030a3adbcf64bf5b90 -- /bin/sh -c 'echo hello julia'`: no such file or directory (ENOENT)

PR #10794, opened 2025-03-21
New Recipe: gecko v1.0.0

termi-official:
This PR adds the build script to get a gecko artifact. Gecko is a graph ordering library developed at LLNL.

### DIFF
--- a/G/gecko/build_tarballs.jl
+++ b/G/gecko/build_tarballs.jl
@@ -14,30 +14,20 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/gecko/
 mkdir build
-cd build
-cmake .. -DCMAKE_INSTALL_PREFIX=$WORKSPACE/destdir/
-cmake --build . --config Releaese
-make install
+cmake -S ./ -B build .. \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_INSTALL_PREFIX=${prefix} \
+    -DCMAKE_BUILD_TYPE=Release
+
+cmake --build build --parallel ${nproc}
+cmake --install build
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = [
-    Platform("i686", "linux"; libc = "glibc"),
-    Platform("x86_64", "linux"; libc = "glibc"),
-    Platform("aarch64", "linux"; libc = "glibc"),
-    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
-    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
-    Platform("powerpc64le", "linux"; libc = "glibc"),
-    Platform("riscv64", "linux"; libc = "glibc"),
-    Platform("i686", "linux"; libc = "musl"),
-    Platform("x86_64", "linux"; libc = "musl"),
-    Platform("aarch64", "linux"; libc = "musl"),
-    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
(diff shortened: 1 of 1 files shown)


G/gecko/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("riscv64", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "freebsd"; ),
+    Platform("aarch64", "freebsd"; )
+]
```
termi-official, line 40: Should this be

```suggestion
platforms = expand_cxxstring_abis([
    Platform("i686", "linux"; libc = "glibc"),
    Platform("x86_64", "linux"; libc = "glibc"),
    Platform("aarch64", "linux"; libc = "glibc"),
    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
    Platform("powerpc64le", "linux"; libc = "glibc"),
    Platform("riscv64", "linux"; libc = "glibc"),
    Platform("i686", "linux"; libc = "musl"),
    Platform("x86_64", "linux"; libc = "musl"),
    Platform("aarch64", "linux"; libc = "musl"),
    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
    Platform("x86_64", "freebsd"; ),
    Platform("aarch64", "freebsd"; )
])
```
?

imciner2, line 40: You just want to include Linux and FreeBSD, correct? If so, this is a better way to do the filtering.
```suggestion
# Only support Linux and FreeBSD
platforms = expand_cxxstring_abis(supported_platforms())
filter!(p -> (Sys.islinux(p) || Sys.isfreebsd(p)), platforms)
```

G/gecko/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/gecko/
+mkdir build
+cd build
+cmake .. -DCMAKE_INSTALL_PREFIX=$WORKSPACE/destdir/
+cmake --build . --config Releaese
+make install
```
imciner2, line 20:
```suggestion
cmake -S ./ -B build .. \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_BUILD_TYPE=Release

cmake --build build --parallel ${nproc}
cmake --install build
```
You need to follow the guidelines here for the CMake builds to make sure the compilers are correct: https://docs.binarybuilder.org/stable/build_tips/#CMake-builds.

This is also slightly cleaner when done this way.

G/gecko/build_tarballs.jl
termi-official: Thanks for the pointer! Included now.

G/gecko/build_tarballs.jl
termi-official: Good point. Included now.

imciner2:
The build failure was a transient issue on the build agent, so I just restarted it and it now all builds correctly. Looks good, so now this is merged and it should be available in the General Registry soon-ish.

PR #10799, opened 2025-03-21
[Modflow6] New version 6.6.1

visr:
Adding netcdf-fortran dependency to support NetCDF IO. Bumping GCC version for [compatibility](https://modflow6.readthedocs.io/en/stable/_dev/developer.html#compile).

### DIFF
--- a/M/Modflow6/build_tarballs.jl
+++ b/M/Modflow6/build_tarballs.jl
@@ -15,7 +15,7 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/modflow6*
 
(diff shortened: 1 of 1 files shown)


visr:
Putting this on hold for now until the new [netcdf-fortran](https://github.com/Unidata/netcdf-fortran) v4.6.2 release is available which can build on the latest netcdf-c and thus should have more platforms available.

Also somewhat considering going for the [extended build](https://github.com/MODFLOW-ORG/modflow6/blob/develop/EXTENDED.md) with MPI and PETSc if that is workable.

PR #10802, opened 2025-03-22
[MLX] Updated to v0.23.2

stemann:
Also,
* Added patches for nbits32 and FreeBSD.
* Using OpenMPI 4 where OpenMPI 5 is not available.
* Updated mpi patch.


stemann:
Patches have been upstreamed:
* https://github.com/ml-explore/mlx/pull/1996
  * https://github.com/ml-explore/mlx/commit/885e10a06d92910deb7b0ee666519c94e9630a51.patch
* https://github.com/ml-explore/mlx/pull/1997
  * https://github.com/ml-explore/mlx/commit/89112bf23fc76d3adf1262f9590d5c0569fb051f.patch

M/MLX/build_tarballs.jl
```diff
@@ -103,7 +102,11 @@ products = Product[
 dependencies = [
     Dependency("libblastrampoline_jll"; compat="5.4", platforms = libblastrampoline_platforms),
     Dependency("OpenBLAS32_jll"; platforms = openblas_platforms),
-    Dependency("OpenMPI_jll"),
+
+    # OpenMPI 5 is ABI compatible with OpenMPI 4, but OpenMPI 5 does not support 32-bit platforms, or FreeBSD
+    Dependency("OpenMPI_jll", v"4.1.8"; compat="4, 5", platforms = filter(p -> nbits(p) == 32 || Sys.isfreebsd(p), platforms)),
+    Dependency("OpenMPI_jll", v"5.0.7"; compat="4, 5", platforms = filter(p -> !(nbits(p) == 32 || Sys.isfreebsd(p)), platforms)),
```
giordano, line 108:
```suggestion
    Dependency("OpenMPI_jll"; compat="4.1.8, 5"),
```

stemann: I've updated the recipe to use
```
    Dependency("OpenMPI_jll", v"4.1.8"; compat="4, 5"), # OpenMPI 5 is ABI compatible with OpenMPI 4
```

To include compat with earlier OpenMPI 4 versions (4.1.8 (and 5.0.7) is/was just a randomly selected (most recent) v4 version to be used for build).

stemann:
@giordano Would it be better to build for avx (and perhaps not for x86_64)? Cf. 
```
[13:09:29] /workspace/srcdir/mlx/mlx/backend/cpu/simd/accelerate_simd.h:112:25: warning: AVX vector argument of type 'short const __attribute__((ext_vector_type(16)))' (vector of 16 'short' values) without 'avx' enabled changes the ABI [-Wpsabi]
[13:09:29]   112 | SIMD_DEFAULT_UNARY(abs, asd::abs)
```
Cf. https://buildkite.com/julialang/yggdrasil/builds/18941#0195d7ae-93b7-41cf-8275-9b6f3863e237/6-15933

M/MLX/build_tarballs.jl
```diff
@@ -103,8 +102,8 @@ products = Product[
 dependencies = [
     Dependency("libblastrampoline_jll"; compat="5.4", platforms = libblastrampoline_platforms),
     Dependency("OpenBLAS32_jll"; platforms = openblas_platforms),
-    Dependency("OpenMPI_jll"),
-    HostBuildDependency(PackageSpec(name="CMake_jll")),  # Need CMake >= 3.30 for BLA_VENDOR=libblastrampoline
+    Dependency("OpenMPI_jll", v"4.1.8"; compat="4, 5"), # OpenMPI 5 is ABI compatible with OpenMPI 4
```
giordano, line 105:
```suggestion
    Dependency("OpenMPI_jll"; compat="4.1.8, 5"), # OpenMPI 5 is ABI compatible with OpenMPI 4
```

M/MLX/build_tarballs.jl
giordano: > To include compat with earlier OpenMPI 4 versions

But why if it won't work?

stemann: The only issue is that there is no OpenMPI v4.0.0 - there are plenty of versions prior to v4.1.8, e.g. v4.0.2 I assume should be fine as well (and I guess v4.0.0 would be OK too if someone updated the OpenMPI recipe to push a v4.0.0).

giordano: > I assume should be fine as well

I always find it extremely dodgy having minimum compat bound != build version (reason why I made the former the default build version): if you have to tell that a package is compatible with ***minimum*** a specific version, what's the point of using an ***earlier*** build version?  On point here, it won't work on riscv64 where there's no build before 4.1.8, so no, it won't be fine as well.

stemann: OK, I get your point regarding ensuring minimum compat bound == build_version - to have some form of check that the minimum compat actually works.

I had not checked that earlier versions were missing platforms (that are now default).

But on the other hand, compat 4.1.8 is an artificial compat bound - limiting users to using only 4.1.8+ - at the benefit of having a BB build/audit pass.

I have no upstream info on which OpenMPI version is required - the build system is only saying the OpenMPI version is version 3.1 - the only info I had on compat was the comment in the OpenMPI@4 recipe saying that v4 and v5 are ABI compatible.

But I’m fine with limiting to 4.1.8+.

stemann: I've committed the suggested change.

PR #10806, opened 2025-03-23
[aws_c_http_jq] Update to version 0.9.6

quinnj2:
This PR updates aws_c_http_jq to version 0.9.6. cc: @quinnj @Octogonapus

### DIFF
--- a/A/aws_c_http_jq/build_tarballs.jl
+++ b/A/aws_c_http_jq/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "aws_c_http_jq"
-version = v"0.9.5"
+version = v"0.9.6"
 
 # Collection of sources required to complete build
 sources = [


ViralBShah:
Doesn't the GIT sha need updating?

giordano:
@quinnj

quinnj:
Yeah, sorry, I was trying something out here. In the last aws_c_http_jq build, it seems to have actually built an older commit than I expected? I think there may be a build cache thing going on, because there was an order of operations like the following:
1. I set up the quinnj/aws-c-http fork and put up an initial recipe PR; my fork was aws-c-http main + 1 commit
2. There were changes in awslabs/aws-c-http that weren't in my fork, so I rebased my fork on new main + my 1 commit
3. I remade the 0.9.5 release in my fork, which updated the Yggdrasil PR which was then merged
4. The 0.9.5 release, however, seems to not include the rebased changes from awslabs/aws-c-http main, but instead is based off hte original fork.

So I was trying to just make a synthentic new release here to see if it would rebuild off the full set of changes, but I'm not sure that's right. Maybe we just need to do a new build +1 release?

quinnj:
@giordano, can you remind me how we do a build-only increment release? and if this might be a good usecase for it?

PR #10811, opened 2025-03-24
libCppInterOp: Upgrade to v0.1.6

Gnimuc:
Also removed the old window patch as it's in upstream now.


Gnimuc:
https://buildkite.com/julialang/yggdrasil/builds/18897#0195cb53-c471-46db-b107-374ec8d871d7/683-845

It failed to config CMake on aarch-freebsd. Is it a new platform that should be filtered out?

Gnimuc:
I think it's ok to just disable aarch freebsd for now.

PR #10818, opened 2025-03-24
gecko wrapper lib

termi-official:
Followup for #10794 @imciner2 . Trying to expose the C++ API via CxxWrap here, analogue to https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SDPA/build_tarballs.jl .

### DIFF
--- a/G/gecko/build_tarballs.jl
+++ b/G/gecko/build_tarballs.jl
@@ -64,4 +64,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"8")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"8", julia_compat="1.6")
--- a/G/gecko/build_tarballs.jl
+++ b/G/gecko/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "gecko"
-version = v"1.0.0"
+version = v"1.0.1"
 
 # Collection of sources required to complete build
 sources = [
@@ -58,10 +58,10 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("libcxxwrap_julia_jll"),
+    Dependency("libcxxwrap_julia_jll"; compat="~0.13.4"),
     Dependency("CompilerSupportLibraries_jll"),
     BuildDependency("libjulia_jll"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"8", julia_compat="1.11")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"8")
--- a/G/gecko/build_tarballs.jl
+++ b/G/gecko/build_tarballs.jl
@@ -7,11 +7,13 @@ version = v"1.0.0"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/LLNL/gecko.git", "490ab7d9b7b4e0f007e10d2af2b022b96d427fee")
+    GitSource("https://github.com/LLNL/gecko.git", "490ab7d9b7b4e0f007e10d2af2b022b96d427fee"),
+    DirectorySource("./bundled"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
+# First build the gecko library
 cd $WORKSPACE/srcdir/gecko/
 mkdir build
 cmake -S ./ -B build .. \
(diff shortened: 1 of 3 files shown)


barche:
For the error on Julia 1.6 you need this workaround:

https://github.com/JuliaPackaging/Yggdrasil/blob/2e30859417cbef5ff2eb2e916a9078f4e86b7d27/L/libcxxwrap_julia/build_tarballs.jl#L5-L8

barche:
I just tried with this build command from within the G/gecko subdir:

```
julia +1.7 --project=../../.ci build_tarballs.jl --verbose --debug --deploy=local x86_64-linux-gnu-cxx11-julia_version+1.11.1
```

You should then be able to `dev_jll` the package (see https://docs.binarybuilder.org/stable/building/#Building-and-testing-JLL-packages-locally) and overwrite the libraries by extracting the tarball produced by the binarybuilder command. In my case I also had to manually trigger a new precompile of gecko_jll, not sure why.

termi-official:
Thanks again. That was very helpful. I could manage to get at minimal working prototype https://github.com/termi-official/Gecko.jl/blob/main/test/runtests.jl . If CI passes this should be good to go.

I am stuck at the last step tho. Essentially I want to allow users to provide a custom sub-class of `Progress`. My first conceptial idea here is to provide an additional class with settable callbacks, i.e.
```cpp
struct JLProgress : public Progress {
private:
   jl_function_t* callback1; // or c-pointer type
   //...
   jl_function_t* quitcallback; // or c-pointer type

public:
    JLProgress() {...}
    void setbeginordercallback(jl_function_t*) {...}
    void beginorder(const Graph* /*graph*/, Float /*cost*/) const {
        // call callback1(graph, cost);
    }
    void setquitcallback(jl_function_t*) {...}
};
```
Is this how it is intended to extend C++ class hierarchies for Julia user packages? If so, how do I pass state around, e.g. in form of a struct which users can define and pass?

Conceptually I am trying to do the following

```julia
mutable struct CustomProgressMonitor
    should_abort::Bool
end

my_monitor = CustomProgressMonitor(false)

function my_gecko_quit_callback(monitor)
    return 
end

gecko_jl_progress_wrapper = Gecko.JLProgress(my_monitor)
Gecko.setquickcallback(my_gecko_quit_callback);

# setup more stuff ....
Gecko.order(..., Gecko.CxxPtr(gecko_jl_progress_wrapper));
```

Does this make sense?

termi-official:
So this is now the concrete example here 
https://github.com/JuliaPackaging/Yggdrasil/blob/443da2a7970bee5272974da290a6c45de17a7a4c/G/gecko/bundled/geckowrapper/geckowrapper.cxx#L8-L82

I think we can auto-generate these with WrapIt. Three questions on this tho, as this does not look fully correct.

1. The CxxWrap type are not conserved. They all now have an additional `CxxRef`. Is this intended? And if not, what is the correct way to remove these?
2. Is there a way to minimize overhead in this redirection layer for cases where performance matters?
3. I get segfaults from time to time but cannot find a minimal reproducer. Is there something obvious missing?

The library side looks like this for now

https://github.com/termi-official/Gecko.jl/blob/ccfb9c5d43b4930b0be3238f654d8d4c093e3f43/test/runtests.jl#L5-L43

where I pass the instance via https://github.com/termi-official/Gecko.jl/blob/ccfb9c5d43b4930b0be3238f654d8d4c093e3f43/test/runtests.jl#L88-L90

termi-official:
https://github.com/termi-official/Gecko.jl?tab=readme-ov-file#how-to-use-the-library has now an example, a proper user-facing API and test coverage for the most important functionality.

barche:
I have added an example for this use case to CxxWrap in:

- https://github.com/JuliaInterop/CxxWrap.jl/pull/480
- https://github.com/JuliaInterop/libcxxwrap-julia/pull/186

It shows how to get better performance using a `cfunction`, gets around the reference problem with a `static_cast` (only needed for regular Julia functions) and makes an attempt to make the class extension a little more intuitive.

I haven't gotten any segfaults from this, first thing to try would be to see if disabling the GC makes them go away, in that case it means something is not properly rooted.

termi-official:
Thanks, this is already very helpful! However, I am still missing one thing here. In C++ we also have access to object data via `this`. Can we emulate this behavior in an efficient way by passing around a struct with additional data on the Julia side? Right now I have https://github.com/JuliaPackaging/Yggdrasil/blob/443da2a7970bee5272974da290a6c45de17a7a4c/G/gecko/bundled/geckowrapper/geckowrapper.cxx#L11 with https://github.com/JuliaPackaging/Yggdrasil/blob/443da2a7970bee5272974da290a6c45de17a7a4c/G/gecko/bundled/geckowrapper/geckowrapper.cxx#L42 , but from my understanding this is not the most efficient way.

barche:
I have updated the example with a way to get internal data from the base class, but since you can design the callback function interface, I would recommend to pass required data as callback function parameters. The updated example uses a callback which is a cfunction wrapping a Julia closure, I'm actually surprised it works (kudos to the Julia developers for that).

termi-official:
I think I am done. Thanks again for the help here! I decided for now to use the slow route for this package, as the callbacks are really just the logger which is only occasionally called. However, I have also already a wrapping package for a bigger software package called MFEM (https://github.com/termi-official/MFEM.jl/) where I want users to add have callbacks which we really need to be as fast as possible, as they are close to the inner loops.

The callback is really some kind of GC-related issue. I hotfix this for now by forcing the GC to not delete the objects by https://github.com/termi-official/Gecko.jl/blob/99c65f7326152f19654684082a5cc0d0818d4678/src/optimization.jl#L19 . Is that a viable option? If so, then I think this can be merged and I can register the package.

barche:
Sorry for the slow response, with `GC.@preserve` the arguments are only protected during the execution of the function call, is that sufficient here? If you need more persistent GC protection you can also use the `CxxWrap.gcprotect` function.

termi-official:
No problem and thanks for taking the time for the detailed review here!

Bypassing GC is doing the trick here. However, I am not sure yet why the objects are GCed in first place, as they are merely passed around. From my understanding they shouldn't bet added to void the GC completely, as they need to be collected later.

barche:
OK, I don't see the reason here either. At any rate, I have no further comments on this. Note that in the mean time CxxWrap 0.17 based on libcxxwrap-julia 0.14 was released, so optionally you could upgrade.

termi-official:
Great. Thanks again for taking time here! I updated the compat and tested it downstream.

barche:
Should we merge this then?

barche:
Actually I just checked and it seems the compat change didn't move to the new version:
https://buildkite.com/julialang/yggdrasil/builds/20317#0196d8a0-b06e-429b-8535-dbcf146e362f/686-700

G/gecko/build_tarballs.jl
```diff
@@ -2,42 +2,72 @@
...
 # The products that we will ensure are always built
 products = [
     LibraryProduct("libgecko", :libgecko)
+    LibraryProduct("libgeckowrapper", :libgeckowrapper)
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[
+dependencies = [
+    Dependency("libcxxwrap_julia_jll"; compat=">=0.13.4"),
```
giordano, line 67: `>=` is not acceptable.

barche: To clarify, `"0.14.2"`should do the trick here, that will take anything in the [0.14,0.15) interval (i.e. excluding 0.15).

giordano: And to answer

> Actually I just checked and it seems the compat change didn't move to the new version:

https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.Dependency

> The optional keyword argument compat can be used to specify a string for use in the Project.toml of the generated Julia package. If compat is non-empty and build_version is not passed, the latter defaults to the minimum version compatible with the compat specifier.

termi-official: Sorry, I misunderstood the change to make here.

> >= is not acceptable.

Just for my understanding here: `>=` shall not be used, because potential breaking changes in newer versions might break the build or because it actually has no effect on the build, as we enforce via `Dependency` anyway that the minimum version is used?

termi-official:
> Should we merge this then?

From my side this PR is fine. If I have not messed up anything else regarding the guidelines, then this is good to go.

termi-official:
Thanks again for taking time to give me detailed feedback here @barche and for elaborating on my mistake @giordano !

PR #10838, opened 2025-03-27
[GALAHAD] Upgrade to v5.2.0

amontoison:


giordano:
Ugh, I cancelled one job because it took much longer than the others and thought that was stalling, but looking at the timestamps I'm not so sure it was stalling, was probably just slow (well, slower than all the other builds) :disappointed:

amontoison:
Can you just restart this build if all the other passed?
On some platforms, the Fortran compiler is very slow!

giordano:
> Can you just restart this build if all the other passed?

I wish, but looks like for cancelled jobs that's not possible

giordano:
>  On some platforms, the Fortran compiler is very slow!

Also because we can't use cache with it.

giordano:
But if we only retrigger the build without changing the source code most builds will be simply cached (because the source code didn't change)

PR #10847, opened 2025-03-28
[aws_c_io] Update to version 0.18.0

quinnj2:
This PR updates aws_c_io to version 0.18.0. cc: @quinnj @Octogonapus


A/aws_c_io/build_tarballs.jl
```diff
@@ -52,7 +52,7 @@ products = [
 dependencies = [
     Dependency("s2n_tls_jll"; compat="1.5.13"),
     Dependency("aws_c_cal_jll"; compat="0.8.3"),
-    Dependency("aws_c_common_jll"; compat="0.11.1"),
+    Dependency("aws_c_common_jll"; compat="0.12.2"),
```
quinnj, line 55:
```suggestion
    Dependency("aws_c_common_jll"; compat="0.12.0"),
```

A/aws_c_io/build_tarballs.jl
```diff
@@ -52,7 +52,7 @@ products = [
 dependencies = [
     Dependency("s2n_tls_jll"; compat="1.5.13"),
     Dependency("aws_c_cal_jll"; compat="0.8.3"),
-    Dependency("aws_c_common_jll"; compat="0.11.1"),
+    Dependency("aws_c_common_jll"; compat="0.12.0"),
```
quinnj, line 55:
```suggestion
    Dependency("aws_c_common_jll"; compat="0.11.1"),
```

quinnj:
The build failure here is real, but I don't know how to fix. The aws-c-io library added support for native apple framework network sockets for iOS and apple TV devices (separate from default posix sockets used on normal macos) and it looks like the build for apple-darwin is failing because it's expecting some new system headers? But that's confusing because it seems like apple-darwin should still just use the system posix sockets and ignore the new stuff, but maybe it's trying to build those anyway because it's detecting some kind of intel + apple configuration where it thinks it can build it?

So a potentially naive idea I have is to try and disable the new functionality altogether via compiler flag, but I'd need to dig in and try that out.

If anyone else has any ideas; I'd appreciate the help.

ararslan:
Maybe it needs a newer macOS SDK as a dependency?

PR #10849, opened 2025-03-28
[aws_c_http] Update to version 0.9.6

quinnj2:
This PR updates aws_c_http to version 0.9.6. cc: @quinnj @Octogonapus


quinnj:
This update should require https://github.com/JuliaPackaging/Yggdrasil/pull/10847, if I understand the upstream library changes correctly (aws-c-io added support for new kinds of apple-native sockets, and this aws-c-http release added ability to use those sockets in http connections).

PR #10862, opened 2025-03-30
[Reactant] Dlopen `libnccl` before `libReactantExtra`

giordano:
@wsmoses this two two things:

* dlopening `libnccl`
* renaming `libReactantExtra` to `libreactantextra`, so that it's dlopened after `libnccl` (sadly at the moment this is the only way with which we can control dlopen order).  This is a breaking change, but we can easily adapt it in Reactant

I need this change to be able to use system nccl, the one which comes from XLA can be much worse.


wsmoses:
is there another way this can be done, e.g. LD_PRELOAD nccl?

PR #10872, opened 2025-03-31
Update AcousticsToolbox to 2024_12_25

mchitre:
Upgrade to the latest version of AcousticsToolbox.


mchitre:
Managed to patch the `Makefile`. Seems to work now!

staticfloat:
Note that the executable names all end in `.exe` on all platforms.  That's weird, but not fatal.

PR #10925, opened 2025-04-03
[Vim] Add 9.1

### LINKED ISSUE #10860: Builder request: vim_jll
From @staticfloat on slack:

If anyone would like a minor binary building challenge, I’d really love to have a vim_jll, as then I could more easily include it in [my BB2 host tools toolchain](https://github.com/JuliaPackaging/BinaryBuilder2.jl/blob/bf8af9c2505ee693aea2a9074d52cd3b70acccc9/BinaryBuilderToolchains.jl/src/toolchains/HostToolsToolchain.jl#L26-L77).  This isn’t something critical which is why I haven’t done it yet, but it would make things slightly more ergonomic for me, so I figured I’d mention it here in case anyone wants to try their hand at building all the dependencies and getting it pushed through.  :)

ViralBShah:
Fix #10860 

cc @staticfloat

### DIFF
--- a/V/Vim/build_tarballs.jl
+++ b/V/Vim/build_tarballs.jl
@@ -58,4 +58,4 @@ dependencies = [
 
 # Build the tarballs
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat="1.6", preferred_gcc_version=v"6", clang_use_lld=false)
+               julia_compat="1.6")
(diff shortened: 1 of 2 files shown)


ViralBShah:
@staticfloat @giordano I can't figure out what to do on windows. Otherwise this seems good to go.

ViralBShah:
The issue seems to be that os_win32.h and os_unix.h are both getting included - the cross compilation is not working out on windows. Need to find a way to make it do the right thing.

PR #10927, opened 2025-04-04
Update mlpack to version 4.6.0

rcurtin:
Hopefully this should be just a simple version bump... a minor patch is needed to keep everything building correctly.


rcurtin:
Does Armadillo need to be built for RISC-V first?  I'm not sure what the process for that is, please let me know.

imciner2:
> Does Armadillo need to be built for RISC-V first? I'm not sure what the process for that is, please let me know.

The best way is to just submit an update to the newest version, and it should get added automatically.

rcurtin:
Hey @giordano, thanks for the help with this one.  But it seems like even a few days on, the `mlpack_jll.jl` package is not updated: https://github.com/JuliaBinaryWrappers/mlpack_jll.jl

Is there anything else I need to do to make that happen?  I thought it was automatic.  Thanks!

giordano:
Registration failed because a job timed out: https://buildkite.com/julialang/yggdrasil/builds/19269#0196102c-3b83-48fa-a8c4-386607a07007.  Let me try to restart it.

PR #10929, opened 2025-04-05
[satsuma] new package

laurentbartholdi:
Satsuma is a preprocessor for SAT problems, and implements symmetry breaking through computation of the automorphism group.
It is an executable-only package.

### DIFF
--- a/S/satsuma/build_tarballs.jl
+++ b/S/satsuma/build_tarballs.jl
@@ -8,11 +8,22 @@ version = v"0.0.0"
 # Collection of sources required to complete build
 sources = [
     GitSource("https://github.com/markusa4/satsuma", "be6beeb6d2538aa133b1f6b7cad84655cda950bb"),
-    DirectorySource("./bundled")
+    DirectorySource("./bundled"),
+    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
+                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
+if [[ "${target}" == x86_64-apple-darwin* ]]; then
+    pushd ${WORKSPACE}/srcdir/MacOSX10.*.sdk
+    rm -rf /opt/${target}/${target}/sys-root/System
+    cp -ra usr/* "/opt/${target}/${target}/sys-root/usr/."
+    cp -ra System "/opt/${target}/${target}/sys-root/."
+    export MACOSX_DEPLOYMENT_TARGET=10.15
+    popd
+fi
+
 cd $WORKSPACE/srcdir/satsuma
 
 for f in $WORKSPACE/srcdir/patches/*.patch; do
@@ -32,7 +43,7 @@ install_license LICENSE
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = filter(!Sys.iswindows, supported_platforms(; experimental=true)) # windows has no boost
-platforms = filter(≠(Platform("x86_64","macOS")),platforms) # ld64.lld: error: undefined symbol: std::__1::__itoa::__u32toa(unsigned int, char*)
+#platforms = filter(≠(Platform("x86_64","macOS")),platforms) # ld64.lld: error: undefined symbol: std::__1::__itoa::__u32toa(unsigned int, char*)
 
 # The products that we will ensure are always built
 products = [
--- a/S/satsuma/build_tarballs.jl
+++ b/S/satsuma/build_tarballs.jl
@@ -48,7 +48,7 @@ products = [
 ]
 
 # Dependencies that must be installed before this package can be built
-dependencies = Dependency[
+dependencies = [
     Dependency("boost_jll"),
 ]
 
--- a/S/satsuma/build_tarballs.jl
+++ b/S/satsuma/build_tarballs.jl
@@ -9,8 +9,8 @@ version = v"1.2.0"
 sources = [
     GitSource("https://github.com/markusa4/satsuma", "be6beeb6d2538aa133b1f6b7cad84655cda950bb"),
     DirectorySource("./bundled"),
-    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
-                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62")
+#    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz",
+#                  "2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62")
 ]
 
 # Bash recipe for building across all platforms
@@ -35,13 +35,13 @@ cp $WORKSPACE/srcdir/tsl/* tsl/
 cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
 
 make -C build satsuma
-install satsuma $bindir
+install build/satsuma $bindir
 install_license LICENSE
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms(; exclude=Sys.iswindows) # windows has no boost
+platforms = supported_platforms(; exclude=Sys.iswindows) |> expand_cxxstring_abis # windows has no boost
 # The products that we will ensure are always built
 products = [
     ExecutableProduct("satsuma", :satsuma)
--- a/S/satsuma/build_tarballs.jl
+++ b/S/satsuma/build_tarballs.jl
@@ -23,7 +23,8 @@ cp $WORKSPACE/srcdir/tsl/* tsl/
 
 cmake .
 make satsuma
-install satsuma $WORKSPACE/bin
+install satsuma $bindir
+install_license LICENSE
 """
 
 # These are the platforms we will build for by default, unless further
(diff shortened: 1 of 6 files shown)


S/satsuma/bundled/tsl/robin_growth_policy.h
imciner2, line 1: I'm not a large fan of keeping whole source files as bundled in Yggdrasil. Patches are fine, but we try to avoid having this be a dev-like repo. Are these files going to be included in upstream at some point, or are they going to always need to be separate?

laurentbartholdi: Yes, there's actually a missing file in the distro, I contacted the author but he may take some time to reply.

laurentbartholdi:
BTW, there's a weird error when building apple_x86-64,
```
[10:04:00] ld64.lld: error: undefined symbol: std::__1::__itoa::__u32toa(unsigned int, char*)
```
for which google gives no suggestion. Should I just disable the platform? It seems to be an issue of broken glibc++.

S/satsuma/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+
+for f in $WORKSPACE/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+
+cp $WORKSPACE/srcdir/tsl/* tsl/
+
+cmake .
+make satsuma
+install satsuma $bindir
```
giordano, line 26: https://docs.binarybuilder.org/stable/build_tips/#CMake-builds

laurentbartholdi: I'm following the instructions from the package.

giordano: You must pass the toolchain file (and also all the other options specified there if you want to do the right thing), whatever the instructions of the package may say because they ignore the fact people may need toolchain files.

laurentbartholdi:
> > for which google gives no suggestion.
> 
> Interesting, the very first result of the search for "std::__1::__itoa::__u32toa" gives me https://stackoverflow.com/a/59672017 which looks to me very much relevant.

Well, yes, I did find that link, and the advice to recompile libc++. Since there's no recipe for a different libc++ in BB, I don't see how to fix the issue.

giordano:
You need a newer SDK, 10.5 based on https://github.com/search?q=repo%3Aphracker%2FMacOSX-SDKs%20__ZNSt3__16__itoa8__u32toaEjPc&type=code

S/satsuma/build_tarballs.jl
giordano:
```suggestion
cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
cmake --build build --parallel ${nproc}
cmake --install build
```

laurentbartholdi: Well, the build worked before your suggested changes, but with the suggestion from #CMake-builds it doesn't anymore :(
First, the cmake --build builds the wrong target (dejavu) and not satsuma; second, the cmake --install doesn't install anything.
I kept the first incantation, but replaced the next two by a direct make and install call.

laurentbartholdi:
> You need a newer SDK, 10.5 based on https://github.com/search?q=repo%3Aphracker%2FMacOSX-SDKs%20__ZNSt3__16__itoa8__u32toaEjPc&type=code

OK, sorry for being thick, but how do I specify that? I searched for kwargs of build_tarballs without success

ararslan:
> how do I specify that?

You have to add the appropriate macOS SDK version as a source. There are a handful of builders here that do that; you can search this repository (e.g. by `grep`ing locally or using the GitHub search) for "MacOSX-SDK" and find one that specifically uses 10.15. An example of one that uses 10.14 is [`aws_c_io`](https://github.com/JuliaPackaging/Yggdrasil/blob/master/A/aws_c_io/build_tarballs.jl).

S/satsuma/robin-map
ararslan, line 1: Did you mean to add git submodules?

laurentbartholdi: I'd rather not go there: the distribution had just forgot to include a file, this will certainly be fixed in the next satsuma release so the current fix is temporary.

laurentbartholdi:
OK, I added the SDK as a source, and build now works for all platforms.

S/satsuma/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "satsuma"
+version = v"0.0.0"
```
ararslan, line 6: I don't think the JLL package will be able to be registered with an all-zero version number. It looks like Satsuma has versions (https://github.com/markusa4/satsuma/releases); many build recipes align the JLL version with that of the software they're building.

Relatedly, based on the hash in `GitSource` below, it appears this builds from a commit on `main` rather than building the most recent release. Is that intentional?

S/satsuma/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
...
+
+for f in $WORKSPACE/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+
+cp $WORKSPACE/srcdir/tsl/* tsl/
+
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+# cmake --build build --parallel ${nproc} # this builds the wrong targets, namely dejavu
+# cmake --install build # this doesn't install anything
```
ararslan, line 37:
```suggestion
```
No reason to keep commented out code if it's known not to work.

S/satsuma/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
...
+make -B build satsuma
```
ararslan, line 38: Did you mean `make -C` instead of `make -B`? I'd never seen `make -B` before but it does appear to exist.

S/satsuma/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
...
+install satsuma $bindir
+install_license LICENSE
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = filter(!Sys.iswindows, supported_platforms(; experimental=true)) # windows has no boost
+#platforms = filter(≠(Platform("x86_64","macOS")),platforms) # ld64.lld: error: undefined symbol: std::__1::__itoa::__u32toa(unsigned int, char*)
```
ararslan, line 46:
```suggestion
platforms = supported_platforms(; exclude=Sys.iswindows) # windows has no boost
```

S/satsuma/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
...
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+    Dependency("boost_jll"),
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+  julia_compat="1.6", preferred_gcc_version=v"13")
```
ararslan, line 60:
```suggestion
               julia_compat="1.6", preferred_gcc_version=v"13")
```
Just making the indentation consistent

S/satsuma/satsuma
ararslan, line 1: AFAICT, these git submodules aren't actually being used, you should be able to just delete them.

ViralBShah:
Merge?

S/satsuma/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+for f in $WORKSPACE/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+
+cp $WORKSPACE/srcdir/tsl/* tsl/
+
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+
+make -C build satsuma
+install build/satsuma $bindir
```
giordano, line 38:
```suggestion
install -Dvm 755 build/satsuma -t "${bindir}"
```

S/satsuma/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; exclude=Sys.iswindows) |> expand_cxxstring_abis # windows has no boost
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("satsuma", :satsuma)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
```
giordano, line 51:
```suggestion
dependencies = [
```

S/satsuma/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+
+make -C build satsuma
+install build/satsuma $bindir
+install_license LICENSE
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; exclude=Sys.iswindows) |> expand_cxxstring_abis # windows has no boost
```
giordano, line 44: > windows has no boost

That's false: https://github.com/JuliaBinaryWrappers/boost_jll.jl/blob/b16816584af63a82f2116b53fb131f3448b92f33/README.md#platforms.

laurentbartholdi:
@ViralBShah ready to merge!

PR #10944, opened 2025-04-07
[Makemsix] add builder

JanisErdmanis:
Makemsix is a Microsoft-created open-sourced tool for creating `.msix` installer archives. It can be used in place of MakeAppx.exe, which comes with the Microsoft Visual Code SDK. This library is needed for the AppBundler project, which aims to integrate postprocessing steps in the `bundle` function, making it much easier to use.

The tool is poorly maintained, and the signing functionality never made it into the master branch. However, according to https://github.com/microsoft/msix-packaging/issues/340#issuecomment-2628599604, Mozilla is using the tool to make its MSIX installers for Windows and also maintains/keeps the fork. 

The Linux build for x86_64 is relatively easy. I just needed to update the C++ std flag for the ICU dependency. I also managed to get other Linux platforms running, and MacOS builds to compile.

I do need Windows support for the AppBundler project. As Julia does not currently cross(pre)compile, I need to do the bundling on the platform it compiles. However, I have given up on the Windows cross-compilation build. WIN32 assumes it is compiled on Windows, and using Linux code paths also did not work for me. Any help is appreciated.

I also tried to use Yggdrasil’s already-built dependencies. Since the CMake project does not support this, I tried to modify `lib/CMakeLists.txt` and `src/msix/CMakeLists.txt`. I managed to build it with OpenSSL, but the executable produced was unusable and had a linking error. So, for now, it builds the vendored dependencies. 

As this is my first Yggdrasil recipe and compiled CMake project, I would appreciate guidance on better practices.


imciner2:
Please ensure you are following the guidance here about how to setup the CMake system: https://docs.binarybuilder.org/stable/build_tips/#CMake-builds. A lot of what you are doing is probably already done for you if you use the toolchain file.

JanisErdmanis:
Once I add the Cmake toolchain file by replacing:
```
cmake ${CMAKE_OPTIONS} ..
```
with 
```
cmake ${CMAKE_OPTIONS} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} ..
```
My build errors. It seems that `CMAKE_CXX_STANDARD` that I had set up at `CMakeLists.txt` with:
```
    sed -i 's/set(CMAKE_CXX_STANDARD 14)/set(CMAKE_CXX_STANDARD 17)/' CMakeLists.txt
```
does not propagate when I use the toolchain file. Why could it be so?

JanisErdmanis:
I also tried:
```
cmake ${CMAKE_OPTIONS} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_STANDARD=17 ..
```
as I need to set the compiler to clang for the project to compile, whereas the toolchain file sets it to gcc.

giordano:
https://docs.binarybuilder.org/stable/build_tips/#Using-GCC-on-macOS-and-FreeBSD

M/Makemsix/build_tarballs.jl
```diff
@@ -0,0 +1,159 @@
...
+]
+
+
+# Script that will adapt to each platform
+script = raw"""
+    cd $WORKSPACE/srcdir/msix-packaging
+
+    # Update C++ standard to 17
+    sed -i 's/set(CMAKE_CXX_STANDARD 14)/set(CMAKE_CXX_STANDARD 17)/' CMakeLists.txt
+    find . -name "CMakeLists.txt" -type f -exec sed -i 's/set(CMAKE_CXX_STANDARD 14)/set(CMAKE_CXX_STANDARD 17)/' {} \;
```
giordano, line 22: Sounds like this should be fixed upstream, please submit a PR, we don't want to maintainer other people's code.

M/Makemsix/build_tarballs.jl
```diff
@@ -0,0 +1,159 @@
...
+    find . -name "CMakeLists.txt" -type f -exec sed -i 's/cmake_minimum_required(VERSION 3.29.0 FATAL_ERROR)/cmake_minimum_required(VERSION 3.21.7)/' {} \;
```
giordano, line 23: If you need CMake 3.29 you can just install it via the jll

M/Makemsix/build_tarballs.jl
```diff
@@ -0,0 +1,159 @@
...
+
+    cp ../lib/CMakeLists.txt lib/CMakeLists.txt
+    cp ../src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp
```
giordano, line 26: Any chance for this code not to live here?

M/Makemsix/build_tarballs.jl
```diff
@@ -0,0 +1,159 @@
...
+          -DCMAKE_OSX_SYSROOT=${SYSROOT} \
+          -DCOREFOUNDATION_LIBRARY=${CF_FRAMEWORK} \
+          -DFOUNDATION_LIBRARY=${FOUNDATION_FRAMEWORK} \
+          -DCMAKE_SHARED_LINKER_FLAGS="-Wl,-install_name,@rpath/libmsix.dylib" \
+          -DCMAKE_SHARED_LIBRARY_SUFFIX=".dylib" \
+          -DXML_PARSER=xerces \
+          -DCRYPTO_LIB=openssl \
+          -DCMAKE_PREFIX_PATH=${prefix} \
+          -DBUILD_SHARED_LIBS=ON \
+          ..
```
giordano, line 79: It'd be better to rationalise this in a way such that all common code is set once.  Something like
```sh
CMAKE_FLAGS=(
...
)
if [[ something ]]; then
    CMAKE_FLAGS+=(...)
elif [[ something else ]];
    CMAKE_FLAGS+=(...)
fi
cmake "${CMAKE_FLAGS[@]}" ..
```
Code duplication is bad and error-prone.

M/Makemsix/build_tarballs.jl
JanisErdmanis: Thanks. I will try that out later.

M/Makemsix/build_tarballs.jl
JanisErdmanis: It may be possible to depreciate those lines, as they were added due to external dependencies.

M/Makemsix/build_tarballs.jl
JanisErdmanis: Eventually, I will condense it into a patch. The changes are necessary to eliminate the use of OpenSSL internal API functions. 

The original `lib/CMakeLists.txt` was written with vendored dependencies in mind and required significant work to compile with external dependencies.

I am currently trying to port from MSVC to the Mingw compiler for Windows compilation. If successful, this will also involve significant patching.

M/Makemsix/build_tarballs.jl
JanisErdmanis: I will take a look at how I can improve it.

M/Makemsix/build_tarballs.jl
```diff
@@ -0,0 +1,124 @@
...
+
+    # Install license and binary
+    install_license LICENSE 
+    install -Dvm 755 "build/bin/makemsix${exeext}" "${bindir}/makemsix${exeext}"
+
+    # Platform-specific library installation
+    if [[ "${target}" == "x86_64-w64-mingw32" ]]; then
+        install -Dvm 755 build/lib/libmsix.* "${libdir}/"
+        install -Dvm 755 build/bin/libmsix.* "${libdir}/"
+        install -Dvm 755 build/bin/libmsix.* "${bindir}/"
```
giordano, line 93: One of these lines is redundant (the first one maybe?)

JanisErdmanis: I kept the first one.

giordano:
The `build_tarballs.jl` file looks ok, but same comment as https://github.com/JuliaPackaging/Yggdrasil/pull/11242#issuecomment-2903927801: can the source code be moved elsewhere (in an ideal world upstreamed)?

JanisErdmanis:
The upstream is not actively maintained, judging by the backlog of pull requests and the duration it has taken to accept my simple PR for header changes. There is a high likelihood of getting OpenSSL API fixes into the upstream over time. However, I do not see a way to incorporate MinGW fixes into it due to the low maintenance and my limited expertise in ensuring code quality.

Forking could be an option. Nonetheless, there remains a chance that PRs will eventually be merged, and the project will be updated. It would also be beyond my control regarding the order in which PRs are merged, which would significantly increase the effort required to keep my fork in sync.

Hence, I prefer to keep the patching way to deal with this project. I could host the patched files on a separate GitHub repository. However, I would prefer to keep it within the build script, as it consolidates the concerns of building the project in one place.

JanisErdmanis:
@giordano, I have significantly reduced the amount of code I need for patches. Can you review whether it's fine as it is now, or do I need to host the patched files elsewhere?

giordano:
Patches are fine, but the fact is that these aren't patches, it's some source code with its build system.  I'm going to merge this now to move this forward now, but hosting that code somewhere else in the future would be much welcome.

PR #10950, opened 2025-04-07
[osslsigncode] add builder

JanisErdmanis:
Osslsigncode allows one to sign Windows binaries like `.exe` and `.dll`. It also contains functionality for signing MSI installers, and according to https://github.com/mtrojnar/osslsigncode/pull/277#issue-1771451905, it can also sign MSIX installers, which is relevant for the AppBundler project.

### DIFF
--- a/O/osslsigncode/build_tarballs.jl
+++ b/O/osslsigncode/build_tarballs.jl
@@ -12,6 +12,11 @@ sources = [
 ]
 
 script = raw"""
+    if [[ "${target}" == *-mingw32 ]]; then
+        cd /opt/x86_64-w64-mingw32/x86_64-w64-mingw32/lib/
+        ln -s libws2_32.a libWs2_32.a
+    fi
+
     cd $WORKSPACE/srcdir/osslsigncode
 
     cmake -B build \
@@ -27,15 +32,7 @@ script = raw"""
     install -Dvm 755 "build/osslsigncode${exeext}" "${bindir}/osslsigncode${exeext}"
 """
 
-platforms = [
-    Platform("x86_64", "linux"; libc = "glibc"),
-    Platform("aarch64", "linux"; libc = "glibc"),
-    Platform("x86_64", "linux"; libc = "musl"),
-    Platform("aarch64", "linux"; libc = "musl"),
-    Platform("x86_64", "macos"),
-    Platform("aarch64", "macos"),
-    Platform("x86_64", "windows"),
-]
+platforms = supported_platforms()
 
 # The products that we will ensure are always built
 products = [
@@ -44,7 +41,7 @@ products = [
 
 dependencies = [
     Dependency("OpenSSL_jll", compat="3.0.16"),
-    Dependency("Zlib_jll", compat="1.3.0"),
+    Dependency("Zlib_jll", compat="1.3.1"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/O/osslsigncode/build_tarballs.jl
+++ b/O/osslsigncode/build_tarballs.jl
@@ -21,7 +21,7 @@ script = raw"""
         -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
         -DCMAKE_BUILD_TYPE=Release \
         -DZLIB_INCLUDE_DIR=${includedir} \
-        -DZLIB_LIBRARY=/workspace/destdir/lib/libz.${dlext}
+        -DZLIB_LIBRARY="${libdir}/libz.${dlext}"
 
     cmake --build build --parallel ${nproc}
     
--- a/O/osslsigncode/build_tarballs.jl
+++ b/O/osslsigncode/build_tarballs.jl
@@ -39,7 +39,7 @@ products = [
 
 dependencies = [
     Dependency("OpenSSL_jll", compat="3.0.16"),
-    Dependency("Zlib_jll", compat="1.3.1"),
+    Dependency("Zlib_jll"),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
--- a/O/osslsigncode/build_tarballs.jl
+++ b/O/osslsigncode/build_tarballs.jl
@@ -33,6 +33,7 @@ script = raw"""
 """
 
 platforms = supported_platforms()
+filter(!(==)(Platform("i686", "windows")), platforms)
 
 # The products that we will ensure are always built
 products = [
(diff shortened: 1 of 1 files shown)


JanisErdmanis:
The Windows build works on my machine locally. What could cause it to fail on the buildkite?

O/osslsigncode/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/mtrojnar/osslsigncode",
+                  "4568c890cc1538ca80be3ee36775ba42223dea04"),
+]
+
+script = raw"""
+    cd $WORKSPACE/srcdir/osslsigncode
+
+    sed -i 's/Ws2_32/ws2_32/g' CMakeLists.txt
```
giordano, line 17: Please submit a PR to fix this upstream, it's an upstream bug

O/osslsigncode/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+filter(!(==)(Platform("i686", "windows")), platforms)
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("osslsigncode", :osslsigncode)
+]
+
+dependencies = [
+    Dependency("OpenSSL_jll", compat="3.0.16"),
+    Dependency("Zlib_jll", compat="1.3.1"),
```
giordano, line 42: This won't work in any released version of Julia, either do
```suggestion
    Dependency("Zlib_jll"; compat="1.2.12"),
```
or
```suggestion
    Dependency("Zlib_jll"),
```

O/osslsigncode/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+        -DZLIB_LIBRARY=/workspace/destdir/lib/libz.a
+
+    cmake --build build --parallel ${nproc}
+    
+    install_license LICENSE.txt
+    install -Dvm 755 "build/osslsigncode${exeext}" "${bindir}/osslsigncode${exeext}"
+"""
+
+platforms = supported_platforms()
+filter(!(==)(Platform("i686", "windows")), platforms)
```
giordano, line 33: Why?

O/osslsigncode/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+script = raw"""
+    cd $WORKSPACE/srcdir/osslsigncode
+
+    sed -i 's/Ws2_32/ws2_32/g' CMakeLists.txt
+
+    cmake -B build \
+        -DCMAKE_INSTALL_PREFIX=${prefix} \
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DZLIB_INCLUDE_DIR=/workspace/destdir/include \
```
giordano, line 23:
```suggestion
        -DZLIB_INCLUDE_DIR=${includedir} \
```

O/osslsigncode/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+        -DZLIB_LIBRARY=/workspace/destdir/lib/libz.a
```
giordano, line 24: I'm confused: if you're linking the static library of zlib, why is that a runtime dependency?  Why not using the shared library instead?

O/osslsigncode/build_tarballs.jl
JanisErdmanis: I opened a PR: https://github.com/mtrojnar/osslsigncode/pull/426

O/osslsigncode/build_tarballs.jl
JanisErdmanis: Doesn't static linking solve this issue?

giordano: You either do static linking ***or*** use zlib as a runtime dependency, doing both doesn't make sense: https://github.com/JuliaPackaging/Yggdrasil/pull/10950#discussion_r2041104429.  We usually try to do dynamic linking as much as possible though (although this is an executable, so the argument is a bit moot)

O/osslsigncode/build_tarballs.jl
JanisErdmanis: Simply me being risk-averse, I guess. I changed to a shared library instead.

O/osslsigncode/build_tarballs.jl
JanisErdmanis: Added it back.

O/osslsigncode/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "osslsigncode"
+version = v"2.10.0"
```
giordano, line 6: Where's v2.10.0 coming from?  In https://github.com/mtrojnar/osslsigncode/tags I only see up to v2.9

JanisErdmanis: https://github.com/mtrojnar/osslsigncode/blob/4568c890cc1538ca80be3ee36775ba42223dea04/CMakeLists.txt#L13

JanisErdmanis: I changed to a V2.9 tagged commit.

O/osslsigncode/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+    cd $WORKSPACE/srcdir/osslsigncode
+
+    sed -i 's/Ws2_32/ws2_32/g' CMakeLists.txt
+
+    cmake -B build \
+        -DCMAKE_INSTALL_PREFIX=${prefix} \
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DZLIB_INCLUDE_DIR=${includedir} \
+        -DZLIB_LIBRARY=/workspace/destdir/lib/libz.${dlext}
```
giordano, line 24: Please ***always*** use the most specific [environment variable](https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables):
```suggestion
        -DZLIB_LIBRARY="${libdir}/libz.${dlext}"
```
Don't use absolute paths, absolute paths are a bad idea.

JanisErdmanis: Ok, I will take a note.

PR #10967, opened 2025-04-11
[nghttp3] Add builder

ararslan:


N/nghttp3/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+              "96ad17fd71d599b78a11e0ff635eccb7d2f6d649"),
+]
+
+script = raw"""
+cd ${WORKSPACE}/srcdir/nghttp3
+
+if [[ ${bb_full_target} == *-sanitize+memory* ]]; then
+    # Install msan runtime (for clang)
+    cp -rL ${libdir}/linux/* /opt/x86_64-linux-musl/lib/clang/*/lib/linux/
+fi
```
giordano, line 19: Are you planning to use it as julia dep?

ararslan: Not today, at least. But I figured since nghttp2 does an msan build, it wouldn't hurt for nghttp3 to do one as well. It does seem within the realm of possibility that one could at some point swap Julia's use of nghttp2 for nghttp3.

PR #10955, opened 2025-04-08
[libdmg-hfsplus] new package

JanisErdmanis:
Libdmg-hfsplus offers open-source utilities for creating compressed dmg files, a de facto standard for MacOS application distribution. It can be used as a cross-platform alternative to `hdiutil`, which, to my understanding, [according to the thread](https://bugzilla.mozilla.org/show_bug.cgi?id=935237), is used by Mozilla to package Firefox in a Linux build environment something like:
```
mkisofs -D -V "Firefox test" -no-pad -r -apple -o firefox-uncompressed.dmg /tmp/bundle/
dmg dmg /build/obj-firefox-darwin/dist/firefox-uncompressed.dmg /build/obj-firefox-darwin/dist/firefox.dmg
```
Alternatively, one can preallocate a disk image and add files to it:
```
   dd if=/dev/zero of=uncompressed.dmg bs=1M count=256
    mkfs.hfsplus -v "Firefox 40.x" uncompressed.dmg
    hfsplus attract.dmg addall stagedir
    dmg dmg uncompressed.dmg compressed.dmg
```

Although the software is used in the wild, it has never been released. Hence, I had to set its version to `0.0.0`.

PR #10971, opened 2025-04-12
[LZFSE] add builder

JanisErdmanis:
LZFSE is an Apple compression utility. Its decompression performance is one of its significant benefits over alternatives while maintaining a decent compression ratio. 

It was integrated into `libdmg_hfsplus` as an optional dependency. I hope to integrate it in https://github.com/JuliaPackaging/Yggdrasil/pull/10955 so that DMG installers can be made with LZFSE compression, improving the user experience. 

Freebsd failed to build, which seems easy to fix with some fiddling. I also get a warning about using the AVX2 instruction set when compiling for the `x86_64` Linux platform.

### DIFF
--- a/L/LZFSE/build_tarballs.jl
+++ b/L/LZFSE/build_tarballs.jl
@@ -14,14 +14,16 @@ script = raw"""
     cd $WORKSPACE/srcdir/lzfse
 
     if [[ "${target}" == *"freebsd"* ]]; then
-        export CMAKE_TARGET_TOOLCHAIN=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake
+        # Add the flag for FreeBSD targets
+        export CFLAGS="${CFLAGS} -D_XOPEN_SOURCE=700"
     fi
 
     # Build with CMake
     cmake -B build \
         -DCMAKE_INSTALL_PREFIX=${prefix} \
         -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
-        -DCMAKE_BUILD_TYPE=Release 
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_C_FLAGS="${CFLAGS}"
 
     cmake --build build --parallel ${nproc}
     cmake --install build
--- a/L/LZFSE/build_tarballs.jl
+++ b/L/LZFSE/build_tarballs.jl
@@ -13,10 +13,14 @@ sources = [
 script = raw"""
     cd $WORKSPACE/srcdir/lzfse
 
+    if [[ "${target}" == *"freebsd"* ]]; then
+        export CMAKE_TARGET_TOOLCHAIN=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake
+    fi
+
     # Build with CMake
     cmake -B build \
         -DCMAKE_INSTALL_PREFIX=${prefix} \
-        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%} \
         -DCMAKE_BUILD_TYPE=Release 
 
     cmake --build build --parallel ${nproc}
@@ -35,8 +39,8 @@ platforms = [
     Platform("x86_64", "macos"),
     Platform("aarch64", "macos"),
     Platform("x86_64", "windows"),
-    #Platform("x86_64", "freebsd"), 
-    #Platform("aarch64", "freebsd")
+    Platform("x86_64", "freebsd"), 
+    Platform("aarch64", "freebsd")
 ]
 
 # The products that we will ensure are always built
--- a/L/LZFSE/build_tarballs.jl
+++ b/L/LZFSE/build_tarballs.jl
@@ -20,7 +20,7 @@ script = raw"""
     # Build with CMake
     cmake -B build \
         -DCMAKE_INSTALL_PREFIX=${prefix} \
-        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN%} \
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
         -DCMAKE_BUILD_TYPE=Release 
 
     cmake --build build --parallel ${nproc}
--- a/L/LZFSE/build_tarballs.jl
+++ b/L/LZFSE/build_tarballs.jl
@@ -1,0 +1,53 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "LZFSE"
+version = v"1.0.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/lzfse/lzfse.git", "e634ca58b4821d9f3d560cdc6df5dec02ffc93fd")
+]
+
+script = raw"""
+    cd $WORKSPACE/srcdir/lzfse
+
+    # Build with CMake
+    cmake -B build \
+        -DCMAKE_INSTALL_PREFIX=${prefix} \
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+        -DCMAKE_BUILD_TYPE=Release 
+
+    cmake --build build --parallel ${nproc}
+    cmake --install build
+
+    install_license LICENSE
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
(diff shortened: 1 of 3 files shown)


L/LZFSE/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "LZFSE"
+version = v"1.0.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/lzfse/lzfse.git", "e634ca58b4821d9f3d560cdc6df5dec02ffc93fd")
```
giordano, line 10: [v1.0](https://github.com/lzfse/lzfse/releases/tag/lzfse-1.0) seems to be https://github.com/lzfse/lzfse/commit/88e2d2788b4021d0b2eb9fe2d97352ae9190f128?

L/LZFSE/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("lzfse", :lzfse),
+    LibraryProduct("liblzfse", :liblzfse),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
```
giordano, line 43:
```suggestion
dependencies = Dependency[
```

L/LZFSE/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+sources = [
+    GitSource("https://github.com/lzfse/lzfse.git", "e634ca58b4821d9f3d560cdc6df5dec02ffc93fd")
+]
+
+script = raw"""
+    cd $WORKSPACE/srcdir/lzfse
+
+    if [[ "${target}" == *"freebsd"* ]]; then
+        export CMAKE_TARGET_TOOLCHAIN=${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake
+    fi
```
giordano, line 18: Why is this necessary?

JanisErdmanis: Clang give a strange error:
```
[13:27:30] [ 90%] Building C object CMakeFiles/lzfse_cli.dir/src/lzfse_main.c.o
[13:27:30] /opt/bin/x86_64-unknown-freebsd13.4-libgfortran5-cxx11/x86_64-unknown-freebsd13.4-clang --sysroot=/opt/x86_64-unknown-freebsd13.4/x86_64-unknown-freebsd13.4/sys-root/ -DLZFSE_DLL  -O3 -DNDEBUG  -Wall -Wno-unknown-pragmas -Wno-unused-variable -std=gnu99 -MD -MT CMakeFiles/lzfse_cli.dir/src/lzfse_main.c.o -MF CMakeFiles/lzfse_cli.dir/src/lzfse_main.c.o.d -o CMakeFiles/lzfse_cli.dir/src/lzfse_main.c.o -c /workspace/srcdir/lzfse/src/lzfse_main.c
[13:27:30] /workspace/srcdir/lzfse/src/lzfse_main.c:76:7: error: call to undeclared function 'gettimeofday'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
[13:27:30]    76 |   if (gettimeofday(&tv, 0) != 0) {
[13:27:30]       |       ^
[13:27:30] 1 error generated.
[13:27:30] make[2]: *** [CMakeFiles/lzfse_cli.dir/build.make:76: CMakeFiles/lzfse_cli.dir/src/lzfse_main.c.o] Error 1
[13:27:30] make[2]: Leaving directory '/workspace/srcdir/lzfse/build'
[13:27:30] make[1]: *** [CMakeFiles/Makefile2:529: CMakeFiles/lzfse_cli.dir/all] Error 2
[13:27:30] make[1]: Leaving directory '/workspace/srcdir/lzfse/build'
[13:27:30] make: *** [Makefile:146: all] Error 2
[13:27:30]  ---> cmake --build build --parallel ${nproc}
[13:27:30]  ---> cmake --build build --parallel ${nproc}
[13:27:30] Previous command exited with 2
```

L/LZFSE/build_tarballs.jl
JanisErdmanis: Good catch. I updated to the proper commit.

L/LZFSE/build_tarballs.jl
giordano: That doesn't sound like a good reason for using GCC.

JanisErdmanis: Claude pointed out that the `#include <sys/time.h>` header not being present. I did try to modify the source and explicitly add it to the source file, which gave no result.

giordano: > Claude pointed out that the `#include <sys/time.h>` header not being present.

I do see it at https://github.com/lzfse/lzfse/blob/88e2d2788b4021d0b2eb9fe2d97352ae9190f128/src/lzfse_main.c#L53.

JanisErdmanis: I can confirm that there is a definition of `gettimeofday` in `<sys/time.h>` header:
```
sandbox:${WORKSPACE}/srcdir/lzfse # cat /opt/aarch64-unknown-freebsd14.1/aarch64-unknown-freebsd14.1/sys-root/include/sys/time.h | grep gettimeofday
int	gettimeofday(struct timeval *, struct timezone *);
```

giordano: You usually want to read the file entirely, it's possible the function is prototyped conditionally, and that condition may not be met here.

giordano: Unsurprisingly https://github.com/freebsd/freebsd-src/blob/58066db597befb899b2fe59031b2a32fb9183f0f/sys/sys/time.h#L630-L633
```c
#if __XSI_VISIBLE
int	getitimer(int, struct itimerval *);
int	gettimeofday(struct timeval *, struct timezone *);
#endif
```

I'm mildly confident the problem is https://github.com/lzfse/lzfse/blob/88e2d2788b4021d0b2eb9fe2d97352ae9190f128/src/lzfse_main.c#L24-L27:
```console
sandbox:${WORKSPACE} # echo "#include <sys/time.h>" | cc -x c -dM -E - -D_POSIX_C_SOURCE=200112L | grep __XSI_VISIBLE
sandbox:${WORKSPACE} #
```
Defining `_XOPEN_SOURCE` should help:
```console
sandbox:${WORKSPACE} # echo "#include <sys/time.h>" | cc -x c -dM -E - -D_POSIX_C_SOURCE=200112L -D_XOPEN_SOURCE=700 | grep __XSI_VISIBLE
#define __XSI_VISIBLE 700
```
Without having to change compiler

L/LZFSE/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/lzfse/lzfse.git", "88e2d2788b4021d0b2eb9fe2d97352ae9190f128")
+]
+
+script = raw"""
+    cd $WORKSPACE/srcdir/lzfse
+
+    if [[ "${target}" == *"freebsd"* ]]; then
+        # Add the flag for FreeBSD targets
```
giordano, line 17: Might be useful for future readers to elaborate what this is needed for

L/LZFSE/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
...
+        export CFLAGS="${CFLAGS} -D_XOPEN_SOURCE=700"
+    fi
+
+    # Build with CMake
+    cmake -B build \
+        -DCMAKE_INSTALL_PREFIX=${prefix} \
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_C_FLAGS="${CFLAGS}"
```
giordano, line 26: Technically you don't need this, [`CMAKE_C_FLAGS`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html) is already initialised based on the value of `CFLAGS`.

L/LZFSE/build_tarballs.jl
JanisErdmanis: This is good to know. The issue here was my workflow, where I just happened to have three copies of the `build_tarballs.jl` files and replaced one with another. This workflow partially developed because I need to do a `git checkout` frequently and test the combination of recipes that each sits in its own branch.

L/LZFSE/build_tarballs.jl
JanisErdmanis: Added a sentence:
```
# FreeBSD requires _XOPEN_SOURCE=700 to make gettimeofday() visible in <sys/time.h>
```

L/LZFSE/build_tarballs.jl
JanisErdmanis: Removed `CMAKE_C_FLAGS` from the list of options

PR #10984, opened 2025-04-15
Update Vulkan_Headers to 1.4.312

serenity4:


serenity4:
I am having the following error:
> CMake 3.22.1 or higher is required.  You are running version 3.21.7

Is there a way to require a newer CMake version for the build, or update the current one?

giordano:
https://discourse.julialang.org/t/how-to-update-cmake-version-in-binarybuilders-wizard/109461/2

PR #10991, opened 2025-04-15
blockSQP: adapt towards sparse version of qpOASES

chplate:
I recently adapted the build script of qpOASES_jll #10843  such that the full functionality of the solver could be used with the sparse LP solver MUMPS.
This PR now takes the change in qpOASES_jll into account and builds the blockSQP library linking against the sparse version of qpOASES.

However, there is still an error that we already encountered in #10433, namely that there are dependency errors for OpenBLAS_jll, perhaps stemming from the fact that we are running the build script with Julia version 1.7 but with compat bounds on OpenBLAS_jll for intended use with newer Julia versions.

I was wondering if we could somehow fix this issue? If it would be better I can also open an issue over at BinaryBuilder.

### DIFF
--- a/B/blockSQP/build_tarballs.jl
+++ b/B/blockSQP/build_tarballs.jl
@@ -4,7 +4,7 @@ name = "blockSQP"
 version = v"0.0.1"
 sources = [
     GitSource("https://github.com/djanka2/blockSQP.git", "da05d3957b93bd55b9a7f08c859d302851056a6d"),
-    GitSource("https://github.com/mathopt/blockSQPWrapper.git", "8d796eda6ceb986c2e04888176c6cb292b94011e"),
+    GitSource("https://github.com/mathopt/blockSQPWrapper.git", "df33d191f1eec687d9efb0116ab515bb7b354dcc"),
     DirectorySource("./bundled")
 ]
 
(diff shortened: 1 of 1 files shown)


fingolfin:
FYI `libpolymake_julia` uses `libblastrampoline_jll`, `libcxxwrap_julia_jll` and `libjulia_jll` together, so perhaps you can take inspiration from there. And perhaps @benlorenz (who maintains it) can answer your questions if you have any.

fingolfin:
@chplate are you still interested in finishing this up?

chplate:
Hey @fingolfin, thanks for reaching out and sorry for the delay. Yes, I am still interested in finishing this up, but after a brief look at ```libpolymake_julia``` I don't know where BLAS enters in the build process there, as it is not explicitly linked against. Also I don't know why [this](https://github.com/JuliaPackaging/Yggdrasil/blob/bff7297ed47f6b6d4fe3f4f7bc7b5ba3e3a44772/L/libpolymake_julia/build_tarballs.jl#L14) block might be needed and whether this could help in our case.

benlorenz:
> Hey @fingolfin, thanks for reaching out and sorry for the delay. Yes, I am still interested in finishing this up, but after a brief look at `libpolymake_julia` I don't know where BLAS enters in the build process there, as it is not explicitly linked against.

libpolymake-julia does not directly use libbastrampoline, but it depends on [`Ipopt_jll`](https://github.com/JuliaPackaging/Yggdrasil/blob/master/C/Coin-OR/Ipopt/build_tarballs.jl)
 which does. We only needed it as a build-dependency for libpolymake-julia to pin the version.

> Also I don't know why [this](https://github.com/JuliaPackaging/Yggdrasil/blob/bff7297ed47f6b6d4fe3f4f7bc7b5ba3e3a44772/L/libpolymake_julia/build_tarballs.jl#L14) block might be needed and whether this could help in our case.

I think this block is not relevant anymore and I will try to remove it for the next rebuild. It was necessary when Yggdrasil was still using julia 1.7. But this has been changed to 1.12 very recently and this block should be obsolete.

fingolfin:
You might also wish to ask for help on the #binarybuilder channel on the Julia slack

PR #11038, opened 2025-04-18
[GettextRuntime] Make gettext-runtime its own package

### LINKED ISSUE #11030: A new GettextRuntime_jll package
Usually, the full Gettext package is not needed at runtime.

For example, glib requires using the `gettext()` function provided by [`libintl` or the system](https://github.com/GNOME/glib/blob/e237ec6c81fbb393c4a26b8ee14d4553ff65ee51/INSTALL.md?plain=1#L63-L66).

Quoting the ["PACKAGING" file in Gettext](https://github.com/autotools-mirror/gettext/blob/master/PACKAGING):

> Packaging hints for binary package distributors
> ===============================================
> 
> Although the source of the gettext package comes as a single package,
> I recommend that in distributions of binary packages the installed files
> be split into three packages:
> 
>    gettext-runtime
>         Contents: Runtime libraries and programs.
>         Audience: Anyone who wants to run internationalized programs.
> 
>    libtextstyle
>         Contents: Text styling library.
>         Audience: Anyone who wants to run or develop programs that produce
>                   styled text, to be displayed in a terminal emulator.
> 
>    gettext-tools
>         Contents: Tools and documentation for developers and translators.
>         Audience: Anyone who wants to develop or localize internationalized
>                   programs, i.e. developers and translators.
>         Dependencies: requires gettext-runtime.
> 
> The 'gettext-runtime' binary package is much smaller than the 'gettext-tools'
> binary package (420 KB versus 5.5 MB). I recommend to include it in the default
> installation list, like GNU 'coreutils'; in fact, the runtime programs are
> utilities for sh programs.

Gettext_jll.jl currently only provides: `LibraryProduct`: `libgettext`

This is part of `gettext-tools`.

Can I create a new package `GettextRuntime_jll.jl` with only the runtime?

The runtime includes `lib/libintl`, `lib/libasprintf`, and a few executables listed in ["PACKAGING" file in Gettext](https://github.com/autotools-mirror/gettext/blob/master/PACKAGING).

xref: https://github.com/JuliaPackaging/Yggdrasil/issues/792

nhz2:
Fixes #11030

These are the parts of gettext that are useful at runtime. Unlike the current `Gettext_jll`, this package doesn't depend on `XML2_jll`.

### DIFF
--- a/G/GettextRuntime/build_tarballs.jl
+++ b/G/GettextRuntime/build_tarballs.jl
@@ -12,8 +12,7 @@ sources = [
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd $WORKSPACE/srcdir/gettext-*/
-cd gettext-runtime
+cd $WORKSPACE/srcdir/gettext-*/gettext-runtime
 
 ./configure --prefix=${prefix} \
     --build=${MACHTYPE} \
--- a/G/GettextRuntime/build_tarballs.jl
+++ b/G/GettextRuntime/build_tarballs.jl
@@ -2,10 +2,11 @@ using BinaryBuilder
 
 # Collection of sources required to build GettextRuntime
 name = "GettextRuntime"
-version = v"0.22.4"
+version_string = "0.22.4"
+version = VersionNumber(version_string)
 
 sources = [
-    ArchiveSource("https://ftp.gnu.org/pub/gnu/gettext/gettext-$(version.major).$(version.minor).$(version.patch).tar.xz",
+    ArchiveSource("https://ftp.gnu.org/pub/gnu/gettext/gettext-$(version_string).tar.xz",
                   "29217f1816ee2e777fa9a01f9956a14139c0c23cc1b20368f06b2888e8a34116"),
 ]
 
--- a/G/GettextRuntime/build_tarballs.jl
+++ b/G/GettextRuntime/build_tarballs.jl
@@ -24,6 +24,11 @@ cd gettext-runtime
 
 make -j${nproc}
 make install
+
+# Multiple licenses are required
+cp ../COPYING toplevel-COPYING
+cp intl/COPYING.LIB intl-COPYING.LIB
+install_license COPYING toplevel-COPYING intl-COPYING.LIB
 """
 
 # These are the platforms we will build for by default, unless further
--- a/G/GettextRuntime/build_tarballs.jl
+++ b/G/GettextRuntime/build_tarballs.jl
@@ -1,0 +1,48 @@
+using BinaryBuilder
+
+# Collection of sources required to build GettextRuntime
+name = "GettextRuntime"
+version = v"0.24.0"
+
+sources = [
+    ArchiveSource("https://ftp.gnu.org/pub/gnu/gettext/gettext-$(version.major).$(version.minor).tar.xz",
+                  "e1620d518b26d7d3b16ac570e5018206e8b0d725fb65c02d048397718b5cf318"),
+]
+
+# Bash recipe for building across all platforms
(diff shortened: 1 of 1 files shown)


ViralBShah:
Perhaps @giordano should chime in given the original issue about using Gettext_jll as a BuildDependency. 

In the meanwhile, might as well update this to 0.25.

nhz2:
I need to test that Makie still works with this instead of the current `Gettext_jll`. I think I can do that using <https://docs.binarybuilder.org/stable/FAQ/#Can-I-publish-a-JLL-package-locally-without-going-through-Yggdrasil?>

nhz2:
Makie works on Linux using Glib built with GettextRuntime_jll, but on Windows I get

image

This is when I run:

```julia
using Pkg
Pkg.activate(;temp=true)
Pkg.add([
    (;url="https://github.com/nhz2/GettextRuntime_jll.jl", rev="02e0a8c"),
    (;url="https://github.com/nhz2/Glib_jll.jl", rev="be9f964"),
    (;name="CairoMakie"),
])
```

nhz2:
The DllMain error goes away if I use version 0.22.4.

```julia
using Pkg
Pkg.activate(;temp=true)
Pkg.add([
    (;url="https://github.com/nhz2/GettextRuntime_jll.jl", rev="85fa1b0"),
    (;url="https://github.com/nhz2/Glib_jll.jl", rev="3e617f9"),
    (;name="CairoMakie"),
])
```

nhz2:
This doesn't work on MacOS, but it seems unrelated to the change. Ref: #11498

JanisErdmanis:
IMO, it could be merged now, as there is still an option to fix the recipe in the future, which seems unlikely to need to be done from the #11498 discussion. Also, later we can think about upgrading the `GettextRuntime_jll` once we have it as a dependency in `Glib_jll`, making debugging a little bit easier.

eschnett:
What kind of dependency would this be? `Dependency` or `Runtime Dependency`?

eschnett:
https://github.com/JuliaPackaging/Yggdrasil/pull/11518 is now blocked on this PR.

nhz2:
I used `Dependency` to make https://github.com/nhz2/Glib_jll.jl but I didn't try `RuntimeDependency`. I think `libintl` is needed to compile Glib.

G/GettextRuntime/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
+using BinaryBuilder
+
+# Collection of sources required to build GettextRuntime
+name = "GettextRuntime"
+version = v"0.22.4"
+
+sources = [
+    ArchiveSource("https://ftp.gnu.org/pub/gnu/gettext/gettext-$(version.major).$(version.minor).$(version.patch).tar.xz",
```
giordano:
```suggestion
version_string = "0.22.4"
version = VersionNumber(version_string)

sources = [
    ArchiveSource("https://ftp.gnu.org/pub/gnu/gettext/gettext-$(version_string).tar.xz",
```

G/GettextRuntime/build_tarballs.jl
```diff
@@ -0,0 +1,55 @@
...
+                  "29217f1816ee2e777fa9a01f9956a14139c0c23cc1b20368f06b2888e8a34116"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/gettext-*/
+cd gettext-runtime
```
giordano:
```suggestion
cd $WORKSPACE/srcdir/gettext-*/gettext-runtime
```

PR #11054, opened 2025-04-21
CoolProp: update to 6.8.0

longemen3000:


longemen3000:
Fails mainly on linux, windows and intel macs, the Aarch64 architecture seems fine, gonna check what i can do, but if you want to take a stab at it, it would help me a lot 😅

barche:
Good to go I think, all ok for you @longemen3000 ?

PR #11072, opened 2025-04-24
[Bun] new package

disberd:
This is a PR to add the [Bun](https://bun.sh/) library

### DIFF
--- a/B/Bun/build_tarballs.jl
+++ b/B/Bun/build_tarballs.jl
@@ -7,29 +7,17 @@ version = v"1.2.10"
 
 release_url = "https://github.com/oven-sh/bun/releases/download/bun-v$version"
 
-# We download the SHA file to extract the SHAs of the archives, taken from https://github.com/MichaelHatherly/BundledWebResources.jl
-sha256sums = let 
-    d = Dict{String,String}()
-    sha256sums_url = "$(release_url)/SHASUMS256.txt"
-    for line in eachline(download(sha256sums_url))
-        sha, file = strip.(split(line, ' '; limit = 2))
-        d[file] = sha
-    end
-    d
-end
-
-repo_SHA1 = let
-    response = read(download("https://api.github.com/repos/oven-sh/bun/git/refs/tags/bun-v$(version)"), String)
-    match(r"\"sha\": \"([0-9a-f]{40})\"", response).captures[1]
-end
+# The code for generating the data and source arrays below can be automatically computed based on bun version from the colab notebook at this link: https://colab.research.google.com/drive/1bicSGoDOkd7u1SeUrDjJIe2yAIR83vWu?usp=sharing
 
 # This will store filename, arch and platform for each supported platform
 data = [
-    (; filename = "bun-linux-x64.zip", sha = sha256sums["bun-linux-x64.zip"], target = "x86_64-linux-gnu", platform = Platform("x86_64", "linux"; libc = "glibc")),
-    (; filename = "bun-linux-aarch64.zip", sha = sha256sums["bun-linux-aarch64.zip"], target = "aarch64-linux-gnu", platform = Platform("aarch64", "linux"; libc = "glibc")),
-    (; filename = "bun-darwin-x64.zip", sha = sha256sums["bun-darwin-x64.zip"], target = "x86_64-apple-darwin14", platform = Platform("x86_64", "macos")),
-    (; filename = "bun-darwin-aarch64.zip", sha = sha256sums["bun-darwin-aarch64.zip"], target = "aarch64-apple-darwin20", platform = Platform("aarch64", "macos")),
-    (; filename = "bun-windows-x64.zip", sha = sha256sums["bun-windows-x64.zip"], target = "x86_64-w64-mingw32", platform = Platform("x86_64", "windows")),
+    (; filename = "bun-linux-x64.zip", sha = "68a154ff1be96851b4d1a87cc5197f027ef80ab79afa3d4587150fae5c34c36e", target = "x86_64-linux-gnu", platform = Platform("x86_64", "linux"; libc = "glibc")),
+    (; filename = "bun-linux-aarch64.zip", sha = "54592fd0237e3ec91a2933dff015e15a78989d97234042e7d5334a4c0ad50603", target = "aarch64-linux-gnu", platform = Platform("aarch64", "linux"; libc = "glibc")),
+    (; filename = "bun-linux-x64-musl.zip", sha = "68a154ff1be96851b4d1a87cc5197f027ef80ab79afa3d4587150fae5c34c36e", target = "x86_64-linux-musl", platform = Platform("x86_64", "linux"; libc = "musl")),
+    (; filename = "bun-linux-aarch64-musl.zip", sha = "54592fd0237e3ec91a2933dff015e15a78989d97234042e7d5334a4c0ad50603", target = "aarch64-linux-musl", platform = Platform("aarch64", "linux"; libc = "musl")),
+    (; filename = "bun-darwin-x64.zip", sha = "3443df70c763665db70267c6e883d312c922ccb54a085a40645e7d5603ba9b59", target = "x86_64-apple-darwin14", platform = Platform("x86_64", "macos")),
+    (; filename = "bun-darwin-aarch64.zip", sha = "07895ef0fb661249b86b0b723b65b1cc4790c7f3685b63cda90122b00299972c", target = "aarch64-apple-darwin20", platform = Platform("aarch64", "macos")),
+    (; filename = "bun-windows-x64.zip", sha = "e936fcfacb3cdd823f4fd60402920f52219cfe1dd970eb5650656f67a5ee9109", target = "x86_64-w64-mingw32", platform = Platform("x86_64", "windows")),
 ]
 
 
@@ -37,7 +25,7 @@ data = [
 # We start by putting the per-arch archive with the executable
 sources = [
     [ArchiveSource("$release_url/$(d.filename)", d.sha; unpack_target = d.target) for d in data]...,
-    GitSource("https://github.com/oven-sh/bun.git", String(repo_SHA1))
+    GitSource("https://github.com/oven-sh/bun.git", "db2e7d7f748dd3951ac0c983de73e75df51bb735")
 ]
 
 # Bash recipe for building across all platforms
@@ -45,7 +33,7 @@ script = raw"""
 cd ${WORKSPACE}/srcdir/
 install_license bun/LICENSE.md
 mkdir "${bindir}"
-install -m 755 "${target}/"*"/bun${exeext}" "${bindir}"
+install -m 755 $target/*/bun$exeext $bindir
 """
 
 # These are the platforms we will build for by default, unless further
--- a/B/Bun/build_tarballs.jl
+++ b/B/Bun/build_tarballs.jl
@@ -6,15 +6,40 @@ name = "Bun"
 version = v"1.2.10"
 
 release_url = "https://github.com/oven-sh/bun/releases/download/bun-v$version"
-# Collection of sources required to complete build
-sources = [
-    ArchiveSource("$release_url/bun-linux-x64.zip", "68a154ff1be96851b4d1a87cc5197f027ef80ab79afa3d4587150fae5c34c36e"; unpack_target = "x86_64-linux-gnu"),
-    ArchiveSource("$release_url/bun-linux-aarch64.zip", "54592fd0237e3ec91a2933dff015e15a78989d97234042e7d5334a4c0ad50603"; unpack_target = "aarch64-linux-gnu"),
-    ArchiveSource("$release_url/bun-darwin-x64.zip", "3443df70c763665db70267c6e883d312c922ccb54a085a40645e7d5603ba9b59"; unpack_target = "x86_64-apple-darwin14"),
-    ArchiveSource("$release_url/bun-darwin-aarch64.zip", "07895ef0fb661249b86b0b723b65b1cc4790c7f3685b63cda90122b00299972c"; unpack_target="aarch64-apple-darwin20"),
-    ArchiveSource("$release_url/bun-windows-x64.zip", "e936fcfacb3cdd823f4fd60402920f52219cfe1dd970eb5650656f67a5ee9109"; unpack_target = "x86_64-w64-mingw32"),
-    GitSource("https://github.com/oven-sh/bun.git", "db2e7d7f748dd3951ac0c983de73e75df51bb735"),
+
+# We download the SHA file to extract the SHAs of the archives
+sha256sums = let 
+    d = Dict{String,String}()
+    sha256sums_url = "$(release_url)/SHASUMS256.txt"
+    for line in eachline(download(sha256sums_url))
+        sha, file = strip.(split(line, ' '; limit = 2))
+        d[file] = sha
+    end
+    d
+end
+
+repo_SHA1 = let
+    response = read(download("https://api.github.com/repos/oven-sh/bun/git/refs/tags/bun-v$(version)"), String)
+    match(r"\"sha\": \"([0-9a-f]{40})\"", response).captures[1]
+end
+
+# This will store filename, arch and platform for each supported platform
+data = [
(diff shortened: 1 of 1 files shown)


B/Bun/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+    GitSource("https://github.com/oven-sh/bun.git", "db2e7d7f748dd3951ac0c983de73e75df51bb735"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/
+install_license bun/LICENSE.md
+mkdir "${bindir}"
+install -m 755 "${target}/"*"/bun${exeext}" "${bindir}"
+"""
```
stemann, line 25: Not possible to build from source?

disberd: It think it is possible, though it's not even listed in the [docs](https://bun.sh/docs/installation) or in the [README](https://github.com/oven-sh/bun) as an option. I also created this recipe based on the one from [Deno](https://github.com/JuliaPackaging/Yggdrasil/blob/master/D/Deno/build_tarballs.jl) which is incidentally also a very similar package, so I thought it would be OK to do so.

Looking at the [Contributing](https://github.com/oven-sh/bun/blob/main/CONTRIBUTING.md) file in the github repo it would like building from source is much more complex and as I am not really experienced in this it would take a lot of effort for me to fully understand how to set this up.

stemann: It might be a challenge - in particular wrt. zig (though we do have `zig_jll`).

Ideally though you should only have to do a standard cmake invocation - e.g. by following the instructions from the BB wizard - could be interesting to know how far you could get.

stemann: You should be able to disregard all of their contributing doc and only use it as a reference.

stemann: https://docs.binarybuilder.org/stable/build_tips/#CMake-builds

disberd: The biggest problem at the moment is that I only have access to a windows machine without admin rights (so can't install WSL) and when I tried using a bit the wizard from BinaryBuilder or directly calling `build_tarballs.jl` on this file I started getting errors much earlier than when I was getting errors in this PR's CI. 
It also seems building on windows is more convoluted?

I unfortunately won't have the time to navigate through this myself in the near future.
I wanted to use Bun for another package and I though it would be useful also for others to have access to the binary as a jll.

But if this recipe/approach is not an option I guess I'll just go with handmade artifacts pointing to the release files like done in https://github.com/MichaelHatherly/BundledWebResources.jl in my new package where I need Bun and leave thie PR to the next person that might want to add Bun and has more experience than me with cmake and maybe access to less restricted developing enviornment

disberd: Maybe worth asking @MichaelHatherly.
I saw you made the Deno recipe for Yggdrasil, did you go for handmade artifacts for Bun in https://github.com/MichaelHatherly/BundledWebResources.jl
Was this just because of time/convenience or you tried building locally Bun with BinaryBuilder and faced challenges?

MichaelHatherly: > Was this just because of time/convenience or you tried building locally Bun with BinaryBuilder and faced challenges?

It was just convenience at the time. Ideally having `bun_jll` available would be much nicer than my hand-rolled artefacts.

disberd:
@stemann, I took inspiration from he build script from @MichaelHatherly in https://github.com/MichaelHatherly/BundledWebResources.jl to extract all paths/SHAs from the package version itself.

This way it should be much easier to update the version without having to manually copy and paste all the SHAs.

I hope this can be enough to have this recipe being acceptable (without having to go through building directly from source).

I also checked the artifact created in buildkite on windows and the produced binary seems to be working fine

stemann:
I took the liberty to try and add a commit which showcase an attempt at building from source

stemann:
OK, it seems Bun is one of the more complicated things to build from source:

1. The Bun build system relies on the `bun` executable itself - likely that can be remedied with some seemingly undocumented bootstrapping, perhaps https://github.com/oven-sh/bun/blob/bun-v1.2.10/scripts/bootstrap.sh - or worked around by just downloading `bun` for x86_64-linux-gnu and putting it in `$PATH`
2. The Bun build system downloads and builds zig and a lot of other dependencies as part of the build, e.g.
```
  BoringSSL
  Brotli
  Cares
  LibDeflate
  LolHtml
  Lshpack
  Mimalloc
  TinyCC
  Zlib
  LibArchive
  Zstd
```
https://github.com/oven-sh/bun/blob/bun-v1.2.10/cmake/targets/BuildBun.cmake#L1088-L1100

The Bun build system should be modified to be able to use system packages: https://github.com/oven-sh/bun/issues/6299

3. Some parts are passed "forbidden" flags like `-march`

So in conclusion - sounds fine to drop building from source in this case.

B/Bun/build_tarballs.jl
stemann: @disberd Just to answer your question, yes, making BB run on Windows (even with the Docker runner) is quite likely a bit of an endeavour: https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/25

stemann:
> @stemann, I took inspiration from the build script from @MichaelHatherly in https://github.com/MichaelHatherly/BundledWebResources.jl to extract all paths/SHAs from the package version itself.
> 
> This way it should be much easier to update the version without having to manually copy and paste all the SHAs, ...

It is actually preferable to have the checksums/signatures in the build recipes in Yggdrasil - it is preferred for reproducibility reasons.

B/Bun/build_tarballs.jl
```diff
@@ -0,0 +1,68 @@
...
+    [ArchiveSource("$release_url/$(d.filename)", d.sha; unpack_target = d.target) for d in data]...,
+    GitSource("https://github.com/oven-sh/bun.git", String(repo_SHA1))
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/
+install_license bun/LICENSE.md
+mkdir "${bindir}"
+install -m 755 "${target}/"*"/bun${exeext}" "${bindir}"
```
stemann, line 48:
```suggestion
install -D -m 755 -v $target/*/bun$exeext $bindir
```

B/Bun/build_tarballs.jl
```diff
@@ -0,0 +1,68 @@
...
+end
+
+repo_SHA1 = let
+    response = read(download("https://api.github.com/repos/oven-sh/bun/git/refs/tags/bun-v$(version)"), String)
+    match(r"\"sha\": \"([0-9a-f]{40})\"", response).captures[1]
+end
+
+# This will store filename, arch and platform for each supported platform
+data = [
+    (; filename = "bun-linux-x64.zip", sha = sha256sums["bun-linux-x64.zip"], target = "x86_64-linux-gnu", platform = Platform("x86_64", "linux"; libc = "glibc")),
```
stemann, line 28: Any reason for leaving out https://github.com/oven-sh/bun/releases/download/bun-v1.2.10/bun-linux-x64-musl.zip ?

disberd: not really, didn't realize this was available. Updated the recipe to add this

B/Bun/build_tarballs.jl
disberd: Thanks, modified this

disberd:
> > @stemann, I took inspiration from the build script from @MichaelHatherly in [MichaelHatherly/BundledWebResources.jl](https://github.com/MichaelHatherly/BundledWebResources.jl?rgh-link-date=2025-04-26T05%3A28%3A42Z) to extract all paths/SHAs from the package version itself.
> > This way it should be much easier to update the version without having to manually copy and paste all the SHAs, ...
> 
> It is actually preferable to have the checksums/signatures in the build recipes in Yggdrasil - it is preferred for reproducibility reasons.

OK, I moved the code to automatically extract the SHAs inside a colab notebook and put just the plain values in the build_tarballs.jl file

disberd:
@stemann is there something else missing for merging this?

B/Bun/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("bun", :bun)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
```
stemann, line 51:
```suggestion
dependencies = Dependency[]
```

B/Bun/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+    [ArchiveSource("$release_url/$(d.filename)", d.sha; unpack_target = d.target) for d in data]...,
+    GitSource("https://github.com/oven-sh/bun.git", "db2e7d7f748dd3951ac0c983de73e75df51bb735")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/
+install_license bun/LICENSE.md
+mkdir "${bindir}"
+install -m 755 $target/*/bun$exeext $bindir
```
stemann, line 36: I would suggest to remove the last of the unnecessary curly braces, and the following should make `install` do the right thing without needing `mkdir`:
```suggestion
cd $WORKSPACE/srcdir
install_license bun/LICENSE.md
install -D -m 755 -v $target/*/bun$exeext $bindir/bun$exeext
```

B/Bun/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+    (; filename = "bun-linux-x64-musl.zip", sha = "353e3a3d6fc4576592bf14a6fc3ea783a8a2811ff68e8d22f6f92fa6141c52a5", target = "x86_64-linux-musl", platform = Platform("x86_64", "linux"; libc = "musl")),
+    (; filename = "bun-linux-aarch64-musl.zip", sha = "3201318206827a993e84d45d5b523a2aecd607057a623a5c12fc11e4a9b231fe", target = "aarch64-linux-musl", platform = Platform("aarch64", "linux"; libc = "musl")),
+    (; filename = "bun-darwin-x64.zip", sha = "3443df70c763665db70267c6e883d312c922ccb54a085a40645e7d5603ba9b59", target = "x86_64-apple-darwin14", platform = Platform("x86_64", "macos")),
+    (; filename = "bun-darwin-aarch64.zip", sha = "07895ef0fb661249b86b0b723b65b1cc4790c7f3685b63cda90122b00299972c", target = "aarch64-apple-darwin20", platform = Platform("aarch64", "macos")),
+    (; filename = "bun-windows-x64.zip", sha = "e936fcfacb3cdd823f4fd60402920f52219cfe1dd970eb5650656f67a5ee9109", target = "x86_64-w64-mingw32", platform = Platform("x86_64", "windows")),
+]
```
stemann, line 21: Might be possible to something like the following (and use `unpack_target = triplet(platform)` in `sources`) to avoid the repetition of platform/triplet:
```suggestion
sources_and_platforms = Dict(
    Platform("aarch64", "linux"; libc = "glibc") => (; filename = "bun-linux-aarch64.zip", sha = "54592fd0237e3ec91a2933dff015e15a78989d97234042e7d5334a4c0ad50603"),
    Platform("aarch64", "linux"; libc = "musl") => (; filename = "bun-linux-aarch64-musl.zip", sha = "3201318206827a993e84d45d5b523a2aecd607057a623a5c12fc11e4a9b231fe"),
    Platform("aarch64", "macos") => (; filename = "bun-darwin-aarch64.zip", sha = "07895ef0fb661249b86b0b723b65b1cc4790c7f3685b63cda90122b00299972c"),
    Platform("x86_64", "linux"; libc = "glibc") => (; filename = "bun-linux-x64.zip", sha = "68a154ff1be96851b4d1a87cc5197f027ef80ab79afa3d4587150fae5c34c36e"),
    Platform("x86_64", "linux"; libc = "musl") => (; filename = "bun-linux-x64-musl.zip", sha = "353e3a3d6fc4576592bf14a6fc3ea783a8a2811ff68e8d22f6f92fa6141c52a5"),
    Platform("x86_64", "macos") => (; filename = "bun-darwin-x64.zip", sha = "3443df70c763665db70267c6e883d312c922ccb54a085a40645e7d5603ba9b59"),
    Platform("x86_64", "windows") => (; filename = "bun-windows-x64.zip", sha = "e936fcfacb3cdd823f4fd60402920f52219cfe1dd970eb5650656f67a5ee9109"),
]
```

B/Bun/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Bun"
+version = v"1.2.10"
```
stemann, line 6: Might want to bump to v1.2.11

B/Bun/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+    (; filename = "bun-darwin-aarch64.zip", sha = "07895ef0fb661249b86b0b723b65b1cc4790c7f3685b63cda90122b00299972c", target = "aarch64-apple-darwin20", platform = Platform("aarch64", "macos")),
+    (; filename = "bun-windows-x64.zip", sha = "e936fcfacb3cdd823f4fd60402920f52219cfe1dd970eb5650656f67a5ee9109", target = "x86_64-w64-mingw32", platform = Platform("x86_64", "windows")),
+]
+
+
+# Collection of sources required to complete build
+# We start by putting the per-arch archive with the executable
+sources = [
+    [ArchiveSource("$release_url/$(d.filename)", d.sha; unpack_target = d.target) for d in data]...,
+    GitSource("https://github.com/oven-sh/bun.git", "db2e7d7f748dd3951ac0c983de73e75df51bb735")
```
stemann, line 28: Was a bit puzzled why the git source was needed, but I see that the license is not included in the binary releases.

stemann (changes requested):
A few suggestions, but otherwise good-to-go.

disberd:
@stemann I applied your suggestions except the one on the triplet due to issue mentioned in the comment

B/Bun/build_tarballs.jl
stemann: You should be able to add something like the following workaround to the script:
```sh
unpack_target=$target
if [[ $target =~ *-apple-darwin* ]]; then
  unpack_target=$rust_target
fi
```

Based on the following info:
```
sandbox:${WORKSPACE}/srcdir # env | grep -i target | grep darwin | sort
CMAKE_TARGET_TOOLCHAIN=/opt/toolchains/x86_64-apple-darwin14-libgfortran5-cxx11/target_x86_64-apple-darwin14.cmake
LLVM_TARGET=x86_64-apple-darwin14
MESON_TARGET_TOOLCHAIN=/opt/toolchains/x86_64-apple-darwin14-libgfortran5-cxx11/target_x86_64-apple-darwin14.meson
bb_full_target=x86_64-apple-darwin14-libgfortran5-cxx11
bb_target=x86_64-apple-darwin14
rust_target=x86_64-apple-darwin
target=x86_64-apple-darwin14
```

stemann: Annoying, but perhaps a workaround...

disberd:
@stemann the automatic extraction of unpack_target should now also be in.

The code you gave as example did not work so I tried a different approach instead of just using `rust_target`.

I eventually realized that the issue was in the `if` clause which is now fixed, but I kept the current strategy of just removing everything after `apple-darwin` instead of just using `rust_target` as I think it's more robust as it doesn't rely on rust_target not actually changing at some point?

Hope this is acceptable.

Would you take care of the merge yourself or shall I ping someone else?

B/Bun/build_tarballs.jl
disberd: Done with slightly different workaround

giordano: Technically what you're looking for is [`BinaryBuilderBase.aatriplet(::Platform)`](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/45acaa5c78a091ef75e5daa5de7a8676600509a5/src/Runner.jl#L45-L50), but sadly we don't seem to provide an env var with that value.  The small hack you used here is probably fine.

stemann:
I think it looks fine - and I pinged @giordano to keep my back clear, as this is actually my first PR reviewed after getting the auth. to click merge :-) Thanks for going through all the paces with this!

stemann (approved):

PR #11075, opened 2025-04-25
Create build_tarballs.jl for norMD

diegozea:
**Description:**  
This PR adds a build recipe for norMD (normalized Mean Distance) version 1.3, a tool from the AQUA suite used to assess the quality of multiple sequence alignments (MSAs).

The build process compiles several executables; however, only the `normd` binary is installed.

**References:**
- Thompson, J. D., Plewniak, F., Ripp, R., Thierry, J. C., & Poch, O. (2001). *Towards a reliable objective function for multiple sequence alignments*. Journal of Molecular Biology, 314(4), 937–951.
- Muller, J., Creevey, C. J., Thompson, J. D., Arendt, D., & Bork, P. (2010). *AQUA: automated quality improvement for multiple sequence alignments*. Bioinformatics, 26(2), 263–265.

### DIFF
--- a/N/norMD/build_tarballs.jl
+++ b/N/norMD/build_tarballs.jl
@@ -1,0 +1,68 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "norMD"
+version = v"1.3.0"
+
+# This script installs norMD (normalized Mean Distance) version 1.3, provided by the AQUA suite.
+# norMD is a statistical metric used to assess the quality of multiple sequence alignments (MSAs).
+# The `normd` program calculates the overall norMD score for an entire multiple sequence alignment.
+#
+# Usage example:
+# - `normd aln_file`: Calculates the norMD score for the specified alignment file.
+#
+# If you use this tool, please cite the following references:
+# - Thompson, J. D., Plewniak, F., Ripp, R., Thierry, J. C., & Poch, O. (2001). Towards a reliable objective function for multiple sequence alignments. Journal of molecular biology, 314(4), 937-951.
+# - Muller, J., Creevey, C. J., Thompson, J. D., Arendt, D., & Bork, P. (2010). AQUA: automated quality improvement for multiple sequence alignments. Bioinformatics, 26(2), 263-265.
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://www.bork.embl.de/Docu/AQUA/latest/norMD1_3.tar.gz", "24ba32425640ae6288d59ca2bf5820dd85616132fe6a05337d849035184c660d")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/normd_noexpat
+
+sed -i '/#include "score.h"/a#include <string.h>' init.c
+
+make
+
+# Only the normd executable is installed to the prefix.
+# The programs normd_subaln, normd_range, normd_sw, normd_aln, and normd_aln1 are built but not installed.
+cp normd $prefix/
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("i686", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("i686", "windows"; ),
+    Platform("x86_64", "windows"; )
+]
+
+
+# The products that we will ensure are always built
+products = [
+    ExecutableProduct("normd", :normd)
+]
+
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_llvm_version = v"6.0.1")


N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "norMD"
```
stemann, line 5: I would suggest to use a less concise name as the tool is for a specific field (bioinformatics, MSA), e.g.,
```suggestion
name = "AquaMSANorMD"
```
Even adding MSA to disambiguate among all the "Aqua"'s out there (Aqua.jl, Aqua macOS GUI, JetBrains Aqua, ... https://github.com/search?q=aqua&type=repositories&s=stars&o=desc ).

diegozea: Hi! Thanks, I understand :) Can it be `NormalizedMeanDistance` ?

stemann: From my perspective, with a background in Computer Science, and not in Bioinformatics, I would be more comfortable with a name completely without field-specific abbreviations, i.e., I would really be most comfortable with a name like
```
AquaMultipleSequenceAlignmentNormalizedMeanDistance
```
But I'm OK with AquaMSA as the prefix if that is preferred, but I think there should be a prefix to disambiguate among other interpretations of "NorMD" or "NormalizedMeanDistance".

I do not think NormalizedMeanDistance only makes sense for MSA (with sequence referring to DNA/RNA sequences, I presume...). E.g., in Computer Vision, it is common to look at the IoU (intersection over union) distance between object detections, and then compute `mAP` (mean average precision) based on an IoU threshold. I would not expect everyone in the whole world (Julia world) to understand, or even accept, computer vision-specific terms as universal terms.

diegozea: In the last commit, I changed the name from norMD to AQUAnorMD to better align with the upstream project and improve clarity.
Should the file also be renamed and moved into the A/ folder to reflect this new naming?

giordano: > I think that the main concern with naming Yggdrasil recipes should be that of the Julia General registry

That's not a concern, JLL packages don't necessarily need to follow regular packages naming.

stemann: Not necessarily, I guess so, but perhaps in e.g. niche cases like this it's a nice default?

See also https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1379

diegozea: Thanks for the feedback and the thoughtful discussion.

I'd like to explain why I still think `norMD` is the most appropriate name for the JLL package:

1. **There is no other standalone software named `norMD`**. So, currently there are no naming clashes, and such conflicts are unlikely in the future — especially since the casing (`norMD`) isn’t standard in most Julia packages.

2. **I really like the idea of JLL packages being named after their actual software or suite**. It helps a lot with discoverability. For example, that’s how I found useful tools like `MAFFT`. I wouldn’t have found them easily if the name had been something like `MultipleAlignmentUsingFastFourierTransform_jll`.

3. **While I’ve taken the implementation from AQUA because it’s the only publicly accessible one — available via `wget` — the `norMD` tool was published independently before AQUA**. In fact, to get other versions, you'd need to contact the paper authors directly. So AQUA is just bundling `norMD`, not the origin of it. Future versions of the tool could appear outside AQUA too. So, I would prefer not to tie the package name to AQUA.

Let me know what you think — I’m open to discussing further, but I believe keeping the `norMD` name fits the conventions and benefits users the most.

N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd normd_noexpat/
+sed -i '/#include "score.h"/a#include <string.h>' init.c
+make
+mkdir ${bindir}
+cp normd ${bindir}/
```
giordano, line 32:
```suggestion
install -Dvm 755 normd -to "${bindir}"
```

N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+    ArchiveSource("https://www.bork.embl.de/Docu/AQUA/latest/norMD1_3.tar.gz", "24ba32425640ae6288d59ca2bf5820dd85616132fe6a05337d849035184c660d"),
+    FileSource("https://www.bork.embl.de/Docu/AQUA/latest/License.txt", "ddb9db7630752f8fdc6898f7c99a99eaeeac5213627ecb093df9c82f56175dc7")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd normd_noexpat/
+sed -i '/#include "score.h"/a#include <string.h>' init.c
+make
```
giordano, line 30:
```suggestion
make -j${nproc}
```

N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
```
giordano, line 51:
```suggestion
platforms = supported_platforms()
```

N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,64 @@
...
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://www.bork.embl.de/Docu/AQUA/latest/norMD1_3.tar.gz", "24ba32425640ae6288d59ca2bf5820dd85616132fe6a05337d849035184c660d"),
+    FileSource("https://www.bork.embl.de/Docu/AQUA/latest/License.txt", "ddb9db7630752f8fdc6898f7c99a99eaeeac5213627ecb093df9c82f56175dc7")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd normd_noexpat/
```
giordano, line 28:
```suggestion
cd $WORKSPACE/srcdir/normd_noexpat/
```

N/norMD/build_tarballs.jl
diegozea: Note that the mentioned `normD` and `normd` do not clash because of the different casing ;) Also, `normD` is an R subroutine, so it would not be a JLL package.

stemann: I would say that casing is too brittle to rely on for disambiguation.

And yes, I don't see the R package command becoming a JLL, it was more to show that there were (more common?) interpretations of what "normd" might be.

N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,62 @@
...
+sources = [
+    ArchiveSource("https://www.bork.embl.de/Docu/AQUA/latest/norMD1_3.tar.gz", "24ba32425640ae6288d59ca2bf5820dd85616132fe6a05337d849035184c660d"),
+    FileSource("https://www.bork.embl.de/Docu/AQUA/latest/License.txt", "ddb9db7630752f8fdc6898f7c99a99eaeeac5213627ecb093df9c82f56175dc7")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/normd_noexpat/
+sed -i '/#include "score.h"/a#include <string.h>' init.c
+make -j${nproc}
```
giordano, line 29:
```suggestion
make -j${nproc} CFLAGS="-c -O2 -std=c99"
```

N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,62 @@
...
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+] # norMD uses an old C standard and cannot be compiled for all the supported platforms
```
giordano, line 49: To be clear, the makefile doesn't even specify what C standard they use, which of course breaks when compilers change their defaults.
```suggestion
platforms = supported_platforms(; exclude=!Sys.islinux)
```

giordano:
This code is so bad, it's a miracle it was ever possible to compile it.

N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+sources = [
+    ArchiveSource("https://www.bork.embl.de/Docu/AQUA/latest/norMD1_3.tar.gz", "24ba32425640ae6288d59ca2bf5820dd85616132fe6a05337d849035184c660d"),
+    FileSource("https://www.bork.embl.de/Docu/AQUA/latest/License.txt", "ddb9db7630752f8fdc6898f7c99a99eaeeac5213627ecb093df9c82f56175dc7")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir/normd_noexpat/
+sed -i '/#include "score.h"/a#include <string.h>' init.c
+make -j${nproc} CFLAGS="-c -O2 -std=c99"
```
giordano, line 29:
```suggestion
make -j${nproc} CFLAGS="-c -O2 -std=c99 -Wno-implicit-function-declaration"
```

giordano: BTW, this should also solve the issue for macOS and FreeBSD

N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+install -Dvm 755 normd -t "${bindir}"
```
giordano, line 30: And this just makes it work on Windows too
```suggestion
install -Dvm 755 normd -t "${bindir}/normd${exeext}"
```

N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+"""
+# NOTE: Only the normd executable is installed.
+# The programs normd_subaln, normd_range, normd_sw, normd_aln, and normd_aln1 are built but not installed.
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; exclude=!Sys.islinux)
+
```
giordano, line 38:
```suggestion
platforms = supported_platforms()
```

diegozea:
Yes! I tried a similar set-up early unsuccessfully; so at some moment desisted and chose to allow only the list of platforms for which I was able to compile it :/

giordano:
Also the Makefile is complete rubbish and doesn't consistently use the `CFLAGS` variable.

fingolfin:
Maybe disable Apple/macOS, FreeBSD and RISC-V support, it seems to work otherwise?

giordano:
> it seems to work otherwise? 

If the error is implicit declaration of functions, as it is, then it fails only on bsd systems and new gcc versions (such as what's used for riscv64) just because those toolchains care to fail during compilation rather than at runtime. It's not useful to ignore those errors.

*Note*: I fat fingered the close button, didn't mean to close

fingolfin:
I nerd-sniped myself into fixing this code... now I feel dirty *shudder*

There is no license indicated anywhere, it's not even clear who authored this code. That's problematic...

fingolfin:
Seems this builds now.

@giordano what do you think?

giordano:
> There is no license indicated anywhere, it's not even clear who authored this code. That's problematic...

The license is present, no?  It's one of the sources, and it seems to be installed automatically.

N/norMD/build_tarballs.jl
```diff
@@ -0,0 +1,54 @@
...
+# - `normd aln_file`: Calculates the norMD score for the specified alignment file.
+#
+# If you use this tool, please cite the following references:
+# - Thompson, J. D., Plewniak, F., Ripp, R., Thierry, J. C., & Poch, O. (2001). Towards a reliable objective function for multiple sequence alignments. Journal of molecular biology, 314(4), 937-951.
+# - Muller, J., Creevey, C. J., Thompson, J. D., Arendt, D., & Bork, P. (2010). AQUA: automated quality improvement for multiple sequence alignments. Bioinformatics, 26(2), 263-265.
+
+# Collection of sources required to complete build
+sources = [
+    ArchiveSource("https://www.bork.embl.de/Docu/AQUA/latest/norMD1_3.tar.gz", "24ba32425640ae6288d59ca2bf5820dd85616132fe6a05337d849035184c660d"),
+    FileSource("https://www.bork.embl.de/Docu/AQUA/latest/License.txt", "ddb9db7630752f8fdc6898f7c99a99eaeeac5213627ecb093df9c82f56175dc7"),
```
fingolfin, line 22: Ah the license is a separate file -- I was working with the tar.gz and there is none in there. Ah well

diegozea:
Thanks a lot, @fingolfin and @giordano !!!

PR #11096, opened 2025-04-30
MAGEMin v1.7.4

NicolasRiel:
- Added chlorite to igneous database
- Updated heat capacity unit for phases to J/kg/K
- Updated light output for geodynamic coupling
- Removed old Matlab apps


NicolasRiel:
Thank you very much @giordano :))

PR #11098, opened 2025-04-30
[Geant4_julia] Use latest Geant4_jll 11.3

peremato:


peremato:
@giordano perhaps you can help me here. I'm facing two issues:

- For apple-darwin, when the package Geant4_jll was built, it found ZLIB version 1.3.1 in the "system". However, now when building this package, it finds version 1.2.11 in the "system", which causes a conflict. A test build for `x86_64-apple-darwin` works, but the one for `x86_64-apple-darwin-julia_version+1.11` fails. Any idea why this is happening?
- It also tries to build for `julia_version+1.13.0`, but fails—presumably because CxxWrap is not available for that version.

Any suggestions on how to resolve these issues? Thanks!

G/Geant4_julia/build_tarballs.jl
```diff
@@ -56,14 +57,12 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     BuildDependency("libjulia_jll"),
+    BuildDependency("Xorg_xorgproto_jll"),
     Dependency("libcxxwrap_julia_jll"; compat="0.13.2"),
```
giordano, line 61: You may want to try version 0.14

G/Geant4_julia/build_tarballs.jl
```diff
@@ -24,7 +24,8 @@ cmake ${WORKSPACE}/srcdir/Geant4_cxxwrap -B build \
     -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
-    -DCMAKE_CXX_STANDARD=17
+    -DCMAKE_CXX_STANDARD=17 \
+    -DZLIB_ROOT="${prefix}"
```
giordano, line 28: It doesn't look like this option is being used or effective.  One option is to delete the tbd file in the sysroot, but I think this is a cmake issue.

barche:
Versions `0.13.4` and `0.14.2` of `libcxxwrap_julia_jll` support Julia 1.13. I recommend using the 0.14 series since the new CxxWrap (0.17) using it is released, unless you need to interoperate with other CxxWrap-based packages that haven't upgraded yet.

G/Geant4_julia/build_tarballs.jl
peremato: yes, sure I can try to use version 0.14. However if I do this, then I would have to upgrade all "binding" packages using CxxWrap. If I want to use, for example, `Geant4` together with `PYTHIA8`, which I do in one of the examples, I would need to update `PYTHIA8`.

peremato:
> Versions `0.13.4` and `0.14.2` of `libcxxwrap_julia_jll` support Julia 1.13. I recommend using the 0.14 series since the new CxxWrap (0.17) using it is released, unless you need to interoperate with other CxxWrap-based packages that haven't upgraded yet.

yes, this what I mentioned in the previous comment. I need to upgrade to new CxxWrap all packages at the same time.

peremato:
@giordano I am not making progress in finding ZLIB. I noticed that if I run locally the following command:
```
julia +1.7 build_tarballs.jl --verbose x86_64-apple-darwin
```
It finds the expected version of ZLIB and succeeds. But if I specify the Julia version with:
```
julia +1.7 build_tarballs.jl --verbose x86_64-apple-darwin-julia_version+1.7.0
```
It fails as is shown in the CI. Do you have an explanation for this? Why in one case the `Zlib_jll v1.3.1+2` binaries get installed in `/opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/local/lib/libz.dylib` and in the other case it appears `/opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/lib/libz.tbd` with the old version?

giordano:
I don't have an explanation for why you observe different behaviour when having different triples, but my suggestion is to delete the tbd file as [I had already suggested](https://github.com/JuliaPackaging/Yggdrasil/pull/11098#discussion_r2105946968).

`/opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/local/lib/libz.dylib` is simply `${libdir}/libz.dylib`: `/opt/${target}/${target}/sys-root/usr/local/` is a symlink to `${prefix}` https://github.com/JuliaPackaging/Yggdrasil/blob/7894537190dd35b010b47fd825c7ca06787560f1/0_RootFS/PlatformSupport/build_tarballs.jl#L181

peremato:
If I delete by adding these lines:
```
if [[ "${target}" == *-apple-darwin* ]]; then
   rm /opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/lib/libz.*
fi
```
It simply does not find the ZLIB at all
```
[17:59:11] CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
[17:59:11]   Could NOT find ZLIB: Found unsuitable version "1.2.8", but required is at
[17:59:11]   least "1.3.1" (found ZLIB_LIBRARY-NOTFOUND)
```
The problem is that `${prefix}/lib` does not have ZLIB.
```
sandbox:${WORKSPACE}/srcdir # ls ${prefix}/lib
Geant4-11.3.0                         libG4graphics_reps.dylib              libG4visHepRep.dylib                  libjulia-codegen-debug.1.10.dylib     libparametric.dylib
cmake                                 libG4intercoms.dylib                  libG4vis_management.dylib             libjulia-codegen-debug.dylib          libpointer_modification.dylib
libG4FR.dylib                         libG4interfaces.dylib                 libbasic_types.dylib                  libjulia-codegen.1.10.0.dylib         libtypes.dylib
libG4GMocren.dylib                    libG4materials.dylib                  libcxxwrap_julia.0.13.2.dylib         libjulia-codegen.1.10.dylib           libunwind.1.0.dylib
libG4RayTracer.dylib                  libG4mctruth.dylib                    libcxxwrap_julia.0.dylib              libjulia-codegen.dylib                libunwind.1.dylib
libG4ToolsSG.dylib                    libG4modeling.dylib                   libcxxwrap_julia.dylib                libjulia-debug.1.10.0.dylib           libunwind.a
libG4Tree.dylib                       libG4parmodels.dylib                  libcxxwrap_julia_stl.dylib            libjulia-debug.1.10.dylib             libunwind.dylib
libG4VRML.dylib                       libG4particles.dylib                  libexcept.dylib                       libjulia-debug.dylib                  libuv.2.dylib
libG4analysis.dylib                   libG4physicslists.dylib               libexpat.1.10.0.dylib                 libjulia-internal-debug.1.10.0.dylib  libuv.a
libG4clhep.dylib                      libG4processes.dylib                  libexpat.1.dylib                      libjulia-internal-debug.1.10.dylib    libuv.dylib
libG4digits_hits.dylib                libG4ptl.3.0.0.dylib                  libexpat.dylib                        libjulia-internal-debug.dylib         libxerces-c-3.2.dylib
libG4error_propagation.dylib          libG4ptl.3.dylib                      libextended.dylib                     libjulia-internal.1.10.0.dylib        libxerces-c.dylib
libG4event.dylib                      libG4ptl.dylib                        libfunctions.dylib                    libjulia-internal.1.10.dylib          pkgconfig
libG4gdml.dylib                       libG4readout.dylib                    libhello.dylib                        libjulia-internal.dylib
libG4geometry.dylib                   libG4run.dylib                        libinheritance.dylib                  libjulia.1.10.0.dylib
libG4geomtext.dylib                   libG4track.dylib                      libjlcxx_containers.dylib             libjulia.1.10.dylib
libG4global.dylib                     libG4tracking.dylib                   libjulia-codegen-debug.1.10.0.dylib   libjulia.dylib
```

giordano:
What if you add zlib as an explicit dependency, since that's needed anyway?

peremato:
How I do it only for `apple-darwin`?

giordano:
https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.Dependency
```julia
Dependency("Zlib_jll"; platforms=filter(Sys.isapple, platforms))
```
but isn't that always needed?

peremato:
If I add "Zlib_jll" with compat "1.3.1" I get:
```
ERROR: LoadError: Unsatisfiable requirements detected for package Zlib_jll [83775a58]:
 Zlib_jll [83775a58] log:
 ├─possible versions are: 1.2.12 or uninstalled
 └─restricted to versions 1.3.1 by an explicit requirement — no versions left
```
If I do not specify the version then it loads "Zlib_jll v1.2.12" in ${prefix}

giordano:
> If I do not specify the version then it loads "Zlib_jll v1.2.12" in ${prefix}

Is that a problem?

peremato:
Yes, It is a problem because the binaries Geant4_jll (the C++ libraries) were built with version 1.3.1 and when building the wapper it requires to ave exactly the same version. The artifacts contain: ` find_dependency(ZLIB 1.3.1)`

peremato:
Everything boils down to the fact the if I specify the version of Julia in the platform triplet, which I have two do if I am using CxxWrap, it loads 1.2.12, but if I do not specify it loads 1.3.1, which is the way the Geant4_jll was built.

giordano:
> Everything boils down to the fact the if I specify the version of Julia in the platform triplet, which I have two do if I am using CxxWrap, it loads 1.2.12

Ah, that's probably because if you specify the julia version in the triple then the package is resolved for that julia version for consistency.

>  but if I do not specify it loads 1.3.1, which is the way the Geant4_jll was built.

I don't see zlib in https://github.com/JuliaPackaging/Yggdrasil/blob/7894537190dd35b010b47fd825c7ca06787560f1/G/Geant4/build_tarballs.jl#L108-L114 If that's a dependency of Geant4 then it should be listed explicitly as a dependency.

giordano:
Time to drop old julia versions? :grin:

peremato:
> Time to drop old julia versions? 😁

Yes, I think so. Should I drop all versions prior current LTS (1.10.9)?

giordano:
That'd be good, yes, which would also reduce the number of artifacts here.

PR #11110, opened 2025-05-02
MAGEMin v1.7.5

NicolasRiel:
- Corrected wrong memory access for ume database (leading to crash during memory deallocation)
- Added test for ume database to avoid such problem in the future


NicolasRiel:
Thank you @giordano !

PR #11111, opened 2025-05-02
add N2QN1 in reverse mode

aitzkora:


N/N2QN1/build_tarballs.jl
```diff
@@ -7,7 +7,7 @@ version = v"3.0"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://gitlab.inria.fr/fuentes/n2qn1.git", "7c74437b3439ea32e88df2f7e69e166a070e8568")
+    GitSource("https://gitlab.inria.fr/fuentes/n2qn1.git", "a3b6c2cf8d429591976f5e33c0c9766b1e8fea05")
```
giordano, line 10: I presume you need to change version number

aitzkora: Hi @giordano , I was thinking that the number version was the version of the software. Technically I did not change the version of N2QN1 because I just added another previous version of the software which works in a total different way.  It is a very old software, and i does not aim to change.
If you agree, I could artificiallly propose v3.0.1 ?

giordano: Yeah, having an artificial version number may be a good case here, but maybe you want a more radical version number change if that's completely different as you say (e.g. 2.0.0)

aitzkora: I updated the version  to 4.0  and I updated also my master branch rebasing on modifications. I did not want to open a new pull request. How to update my pull request ?

N/N2QN1/build_tarballs.jl
```diff
@@ -3,11 +3,11 @@
 using BinaryBuilder, Pkg
 
 name = "N2QN1"
-version = v"3.0"
+version = v"4.0"
```
giordano, line 6: Why 4 and [not 2](https://github.com/JuliaPackaging/Yggdrasil/pull/11111#discussion_r2072448122) if you said this is an _older_ version?  When they'll release v4 then what do you do here?

aitzkora: I was not sure if it was possible to update a package with a decreasing  number version. I decided to make it monotonic increasing. :smile: In fact, we do not care, it is a arbitrary number.  I wanted to say that the original software is old and will not be updated. So the number version put on my mirror does not reflect anything. Along the direct version of N2QN1 I add a reverse communication version for which I have the authorization of the authors to release under GPL license as the direct version. So compare to 3.0 there is no regression, the direct version is still the same, and we have a new feature enabling us to use reverse communication. So 4.0 seems to be coherent. Hoping that it will clarify some points and the software could be merge. Thanks , by advance.

PR #11115, opened 2025-05-03
[GASNet] Add GASNet smp conduit recipe

mofeing:
there are still some things to properly configure, like:

- ~the `CROSS_HAVE_MMAP`, `CROSS_PAGESIZE`, `CROSS_STACK_GROWS_UP`, `CROSS_HAVE_X86_CMPXCHG16B`, `CROSS_HAVE_SHM_OPEN` env vars~ (needs better tuning but can be done later)
- ~~flags for debug mode~~ (not a priority)
- ~~fix for macOS~~ (fixed)

if this works correctly, i will add more conduits like the mpi and the ucx ones

### DIFF
--- a/G/GASNet/GASNet_conduit_smp/build_tarballs.jl
+++ b/G/GASNet/GASNet_conduit_smp/build_tarballs.jl
@@ -1,0 +1,69 @@
+using BinaryBuilder, Pkg
+
+include(joinpath(@__DIR__, "../common.jl"))
+
+name = gasnet_name("smp")
+version = v"2024.5.0"
+
+# Bash recipe for building across all platforms
+script = raw"""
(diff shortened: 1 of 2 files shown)


mofeing:
@giordano is there a reason why it has converted back to draft? If it's about the env vars, they work for the current configuration and would only need some tuning for some performance reasons (which should already be good)

giordano:
I may have misunderstood the TODO list in the first message, the first item wasn't crossed-out, but maybe "can be done later" was sufficient?

mofeing:
> I may have misunderstood the TODO list in the first message, the first item wasn't crossed-out, but maybe "can be done later" was sufficient?

my bad, i crossed it out now. i just wanted to clarify that they might need better platform-dependent tuning for some performance increase, but that the current config works.

i'm marking it as ready for review again.

PR #11118, opened 2025-05-04
new: IRBEM

Beforerr:
add International Radiation Belt Environment Modeling (IRBEM) library

### DIFF
--- a/I/IRBEM/build_tarballs.jl
+++ b/I/IRBEM/build_tarballs.jl
@@ -1,0 +1,38 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder
+
+name = "IRBEM"
+version = v"5.0.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/PRBEM/IRBEM.git", "e7cecb00caf97bb6357f063d2ba1aa76d71a3705")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd IRBEM/
+make -j${nproc} all
+make install
+mkdir -p ${libdir}
+cp -L ./libirbem.so ${libdir}/
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(; )
+platforms = expand_gfortran_versions(platforms)
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libirbem", :libirbem)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6")
--- a/I/IRBEM/build_tarballs.jl
+++ b/I/IRBEM/build_tarballs.jl
@@ -23,7 +23,7 @@ else
     make -j${nproc} OS=linux64 ENV=gfortran64 all
     make OS=linux64 ENV=gfortran64 install
 fi
-install -Dvm 755 libirbem.so "${libdir}/libirbem.${dlext}"
+install -Dvm 755 libirbem.* "${libdir}/libirbem.${dlext}"
 """
 
 # These are the platforms we will build for by default, unless further
--- a/I/IRBEM/build_tarballs.jl
+++ b/I/IRBEM/build_tarballs.jl
@@ -23,12 +23,7 @@ else
     make -j${nproc} OS=linux64 ENV=gfortran64 all
     make OS=linux64 ENV=gfortran64 install
 fi
-mkdir -p ${libdir}
-if [[ ${target} == *apple* ]]; then
-    cp -L ./libirbem.so ${libdir}/libirbem.dylib
-else
-    cp -L ./libirbem.* ${libdir}/
-fi
+install -Dvm 755 libirbem.so ${libdir}/libirbem.dylib
 """
 
 # These are the platforms we will build for by default, unless further
--- a/I/IRBEM/build_tarballs.jl
+++ b/I/IRBEM/build_tarballs.jl
@@ -4,7 +4,6 @@ using BinaryBuilder, Pkg
 
 name = "IRBEM"
 version = v"5.0.0"
-
 # Collection of sources required to complete build
 sources = [
     GitSource("https://github.com/PRBEM/IRBEM.git", "e7cecb00caf97bb6357f063d2ba1aa76d71a3705")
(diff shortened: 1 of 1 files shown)


I/IRBEM/build_tarballs.jl
```diff
@@ -0,0 +1,50 @@
...
+else
+    make -j${nproc} OS=linux64 ENV=gfortran64 all
+    make OS=linux64 ENV=gfortran64 install
+fi
+mkdir -p ${libdir}
+if [[ ${target} == *apple* ]]; then
+    cp -L ./libirbem.so ${libdir}/libirbem.dylib
+else
+    cp -L ./libirbem.* ${libdir}/
+fi
```
giordano, line 31:
```suggestion
install -Dvm 755 libirbem.so ${libdir}/libirbem.dylib
```

I/IRBEM/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+if [[ ${target} == *mingw* ]]; then
+    make -j${nproc} OS=win64 ENV=gfortran64 all
+    make OS=win64 ENV=gfortran64 install
+elif [[ ${target} == *apple* ]]; then
+    make -j${nproc} OS=osx64 ENV=gfortran64 all
+    make OS=osx64 ENV=gfortran64 install
+else
+    make -j${nproc} OS=linux64 ENV=gfortran64 all
+    make OS=linux64 ENV=gfortran64 install
+fi
```
giordano, line 25:
```suggestion
make -j${nproc} OS=linux64 ENV=gfortran64 CC=cc all
make OS=linux64 ENV=gfortran64 install
```

I/IRBEM/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+install -Dvm 755 libirbem.* "${libdir}/libirbem.${dlext}"
```
giordano, line 26:
```suggestion
install -Dvm 755 libirbem.so "${libdir}/libirbem.${dlext}"
```

I/IRBEM/build_tarballs.jl
```diff
@@ -0,0 +1,45 @@
...
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = expand_gfortran_versions(supported_platforms())
+filter!(p -> !(Sys.iswindows(p) && arch(p) == "i686"), platforms)
```
giordano, line 32:
```suggestion
```

I/IRBEM/build_tarballs.jl
giordano: Actually:
```suggestion
make -j${nproc} OS=linux64 ENV=gfortran64 CC=cc COMPILE_LIB_NAME="libirbem.${dlext}" INSTALL_LIB_NAME="libirbem.${dlext}" all
make OS=linux64 ENV=gfortran64 COMPILE_LIB_NAME="libirbem.${dlext}" INSTALL_LIB_NAME="libirbem.${dlext}" install
```

I/IRBEM/build_tarballs.jl
giordano, line 26:
```suggestion
install -Dvm 755 "libirbem.${dlext}" "${libdir}/libirbem.${dlext}"
```

I/IRBEM/build_tarballs.jl
Beforerr: Thanks for reviewing. May I ask why make such change? It seems that makefile uses `OS` info (`FFLAGS` is different)

I/IRBEM/build_tarballs.jl
Beforerr: It seems that the CI could compile for "i686".

giordano: You need https://github.com/JuliaPackaging/Yggdrasil/pull/11118#discussion_r2072751193

Beforerr: Thanks for reviewing. But it seems that makefile uses OS info (FFLAGS is not the same)

giordano: Yes, the only thing that the windows Makefile you selected was doing differently than the Linux one is to prevent compilation for 32-bit target, hence why I suggested not to use that Makefile _and_ allow compilation for the 32-bit platform.

Beforerr: Ohhh, got you! Thanks for the explanation

Beforerr:
Thanks @giordano. It works finally. Could we merge it so I can test against it?

I/IRBEM/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "IRBEM"
+version = v"5.0.0"
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/PRBEM/IRBEM.git", "e7cecb00caf97bb6357f063d2ba1aa76d71a3705")
```
stemann, line 9: v5.0.0 seems to be https://github.com/PRBEM/IRBEM/commit/46b834627361ff4af40c974c337c3c77e4dab07f

Beforerr: Right, but there are few bugs corrected afterwards (no breaking changes). Do we have to stick to the exact commmit? (if so how to track `main` branch)

stemann: Not sure what would be best here. @giordano ?

I would be OK with calling it v5.0.0, and just add a comment stating what you just wrote.

Optionally, I would suggest v0.0.YYYYMMDD for untagged/unreleased commits as it is almost certain to not interfere with upstream versioning.

stemann: … but I’m also not in the target audience for this, so my attitude towards calling it v5.0.0 and just commenting might be too sloppy. It could easily annoy me if it was something I was going to use.

stemann: Likely better to go for the second option mentioned by @giordano - adding the necessary patches on top of v5.0.0 - and although tempting my mind in this case, perhaps better not to do the needless dance I did 😁: https://github.com/JuliaPackaging/Yggdrasil/pull/9898#discussion_r1873733652

stemann: ... might also be possible to get upstream to release a v5.0.1 (they merged something five days ago...)

Beforerr: Sorry, but I really do not have the experience of patching C codes...
 
P.S.: For this case, I do not think using the latest version would cause any problem though sloppy indeed. I would really like to make relevant Julia package work just fine and then it would put us in a better position to ask upstream to do a patch release.

stemann: I'm no expert at creating patches from git commits, but you might be able to create the patches by simply checking out the commit you would like, and then run `git format-patch -o bundled/patches 46b834627361ff4af40c974c337c3c77e4dab07f..HEAD` to get the patches from the v5.0.0 commit and to the commit you would like to package.

... and add a couple of `atomic_patch -p1 < filename` lines to the build script.

Beforerr: Thanks for the advice! I just think for this specific case (more than one bugfix), patching does not seem better compared to a sloppy commit id (both approaches are very specific but better than using the exact 5.0.0 commit id).

PR #11123, opened 2025-05-06
[FLINT] Bump to 3.2.2 (+ epsilon)

thofma:
Need the fix for https://github.com/flintlib/flint/issues/2296 and thought we might as well bump to 3.2.2.

@lgoettgens it was correct to dump the other patch, right?

CC: @fingolfin @benlorenz (@Joel-Dahne)


lgoettgens:
> @lgoettgens it was correct to dump the other patch, right?

Indeed, the patch notes for https://github.com/flintlib/flint/releases/tag/v3.2.2 mention https://github.com/flintlib/flint/pull/2264 which is also contained in the patch.

F/FLINT/build_tarballs.jl
```diff
@@ -25,7 +25,7 @@ using BinaryBuilder, Pkg
 # coordinated with corresponding changes to Singular_jll.jl, Nemo.jl and polymake_jll.jl
 # and possibly other packages.
 name = "FLINT"
-upstream_version = v"3.2.1"
+upstream_version = v"3.2.2"
 version_offset = v"0.0.0"
```
lgoettgens, line 29:
```suggestion
version_offset = v"0.0.1"
```
maybe bump the offset to mark that this is not *pure* 3.2.2, but includes some further change (i.e. the patch you included)

F/FLINT/bundled/patches/Fix-fmpz_mat_minpoly_modular-with-large-coefficients.patch
```diff
@@ -0,0 +1,162 @@
+From f8ef228807ba245538492eaf7abc47e804154eef Mon Sep 17 00:00:00 2001
```
lgoettgens, line 1: Have you checked locally that this patch cleanly applies to FLINT v3.2.x, and that it still passes the tests? Iirc, there have been some major changes considering matrices on the FLINT side between 3.2.x and master, so it would be good to check that

PR #11131, opened 2025-05-07
Update TDLib to 1.8.48

AlexKlo:


T/TDLib/build_tarballs.jl
```diff
@@ -29,17 +36,19 @@ cmake \
...
+cd ../
 mkdir build-cross && cd build-cross
 cmake \
     -DCMAKE_BUILD_TYPE=Release \
-    -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_C_COMPILER=clang \
+    -DCMAKE_CXX_COMPILER=clang++ \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DNATIVE_BUILD_DIR=${WORKSPACE}/srcdir/td/build_native \
     ..
```
giordano, line 47: Why do you need specifically clang?  Also, there's an easier way: https://docs.binarybuilder.org/stable/build_tips/#Using-GCC-on-macOS-and-FreeBSD

T/TDLib/build_tarballs.jl
```diff
@@ -16,9 +16,16 @@ cd ${WORKSPACE}/srcdir/td/
 
 install_license LICENSE_1_0.txt
 
+export CC=clang
+export CXX=clang++
+export HOSTCC=clang
+export HOSTCXX=clang++
```
giordano, line 22: This is wrong, these aren't host compilers.

T/TDLib/build_tarballs.jl
```diff
@@ -16,9 +16,16 @@ cd ${WORKSPACE}/srcdir/td/
 
 install_license LICENSE_1_0.txt
 
+export CC=clang
+export CXX=clang++
```
giordano, line 20: This is useless for CMake

T/TDLib/build_tarballs.jl
```diff
@@ -29,17 +36,19 @@ cmake \
...
-    -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_C_COMPILER=clang \
+    -DCMAKE_CXX_COMPILER=clang++ \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DNATIVE_BUILD_DIR=${WORKSPACE}/srcdir/td/build_native \
     ..
 cmake --build . --target tdjson -j${nproc}
 cmake --build . --target tdjson_static -j${nproc}
 cmake --install .
+cp libtdjson.${dlext} $libdir/libtdjson.${dlext}
```
giordano, line 51: Why?  Isn't it installed already?

T/TDLib/build_tarballs.jl
AlexKlo: According to the documentation, Clang is the recommended compiler. 
The earlier version of TDLib exhibited instability on Linux.

T/TDLib/build_tarballs.jl
AlexKlo: Oops, I just forgot to include -DCMAKE_INSTALL_PREFIX=$prefix .

T/TDLib/build_tarballs.jl
```diff
@@ -19,6 +19,8 @@ install_license LICENSE_1_0.txt
 mkdir build_native && cd build_native
 cmake \
     -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_COMPILER=clang \
+    -DCMAKE_CXX_COMPILER=clang++ \
     -DCMAKE_INSTALL_PREFIX=$host_prefix \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_HOST_TOOLCHAIN} \
```
giordano, line 25: Again, you only need to change the toolchain file if you really want to use clang: https://docs.binarybuilder.org/stable/build_tips/#Using-GCC-on-macOS-and-FreeBSD

T/TDLib/build_tarballs.jl
```diff
@@ -62,4 +66,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"9", julia_compat="1.8")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"10", julia_compat="1.8")
```
giordano, line 69: What's the point of changing the GCC version if you don't use GCC?

T/TDLib/build_tarballs.jl
AlexKlo: thanks, I will try!

T/TDLib/build_tarballs.jl
```diff
@@ -62,4 +62,4 @@ dependencies = [
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"9", julia_compat="1.8")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.8")
```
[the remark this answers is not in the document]
AlexKlo: I assumed that with an older version, the builds would be unstable
With v"10" build works stable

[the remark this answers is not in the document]
giordano: I don't really know what that means 🙂

AlexKlo:
@giordano 
It looks like the new version is working stably.
Let’s merge it!

PR #11135, opened 2025-05-07
[LLVM] enable Zstd and build-id for all versions

vtjnash:


L/LLVM/common.jl
```diff
@@ -793,7 +800,8 @@ function configure_extraction(ARGS, LLVM_full_version, name, libLLVM_version=not
     end
 
     dependencies = BinaryBuilder.AbstractDependency[
-        Dependency("Zlib_jll"), # for LLD&LTO
+        Dependency("Zlib_jll"), # for LLD&LTO&debuginfo
+        Dependency("Zstd_jll"), # for debuginfo
```
giordano, line 804: This means also Zstd_jll will have to become an stdlib.

vtjnash: You're in luck haha: https://github.com/JuliaLang/julia/pull/58344

giordano:
Other changes necessary are:

* bump the number in https://github.com/JuliaPackaging/Yggdrasil/blob/12862efcc82a57192e7021e5c6d7ac16810f2ae8/L/LLVM/LLVM_full%4019/build_tarballs.jl#L7
* bump the number in https://github.com/JuliaPackaging/Yggdrasil/blob/12862efcc82a57192e7021e5c6d7ac16810f2ae8/L/LLVM/LLVM_full_assert%4019/build_tarballs.jl#L7
* enabling zstd should be guarded by version checks.  For example https://github.com/JuliaPackaging/Yggdrasil/blob/fb459a50ae779792a3fa99b8899fc2815c78cdef/L/LLVM/common.jl#L208-L209 should be guarded by `if [[ "${LLVM_MAJ_VER}" -ge "19" ]]; then` and https://github.com/JuliaPackaging/Yggdrasil/blob/fb459a50ae779792a3fa99b8899fc2815c78cdef/L/LLVM/common.jl#L687 and https://github.com/JuliaPackaging/Yggdrasil/blob/fb459a50ae779792a3fa99b8899fc2815c78cdef/L/LLVM/common.jl#L804 should only be pushed to `dependencies` `if version >= v"19"

L/LLVM/common.jl
```diff
@@ -162,6 +170,10 @@ CMAKE_C_FLAGS=()
 
 CMAKE_FLAGS=()
 
+if [[ "${target}" != *-apple-darwin* ]]; then
```
Zentrik, line 173: we should disable this on windows as well right?
```suggestion
if [[ "${target}" != *-apple-darwin* || ${target} == *mingw* ]]; then
```

vtjnash: No, it's fine there too. `lld` already supports this (in fact, it likely cannot turn it off), I'm just a fool who cannot copy-paste correctly, and apparently only lld cares enough to warn me.

vtjnash:
It looks like it did not rebuild those binaries, since it thinks they were in the cache. How do we drop those bad builds from the caches?

giordano:
Yeah, as I mentioned in https://github.com/JuliaPackaging/Yggdrasil/pull/11152#discussion_r2083120335 I think there's been a race condition with the other PR because our cache indexing doesn't capture changes in directories above the `build_tarballs.jl` file, but they're relevant in this case, I don't think we ever had this situation before. You can make a change like https://github.com/JuliaPackaging/Yggdrasil/compare/0a518d858d89ec06880b5c01114a552159e60b42..184adfda7640461f67125e1192f47153fd36d058, the two second PR of the two to be merged will have to resolve a merge conflict, but that's fine, I guess.

PR #11142, opened 2025-05-08
MAGEMin v1.7.6

NicolasRiel:
- Corrected mistake when outputting phase in mol fraction.
- Added benchmark for mol, wt, vol and densities against Theriak.
- Corrected call to wrong function leading to segfault error in some special circumstances


NicolasRiel:
Thank you very much @giordano !

PR #11143, opened 2025-05-08
libwirePi

stensmo:
Added WiringPi as JLL. Built only for aarch64 so far.


W/WiringPi/build_tarballs.jl
```diff
@@ -0,0 +1,57 @@
...
+cd WiringPi/
+
+cd wiringPi
+
+cp COPYING.LESSER LICENSE
+
+make
+make V=1 install||true
+
+install -Dvm 755 "libwiringPi.so.3.14" "${libdir}/libwiringPi.$dlext"
```
imciner2, line 30:
```suggestion
```
Shouldn't be needed with above make changes.

imciner2 (changes requested):

stensmo:
Should be fixed now

imciner2 (approved):

imciner2:
Looks good now. Thanks for your contribution.

(p.s., in the future, it is better if you create a new branch for your changes instead of master. Using master can sometimes lead to problems in some cases).

stensmo:
> Looks good now. Thanks for your contribution.
> 
> (p.s., in the future, it is better if you create a new branch for your changes instead of master. Using master can sometimes lead to problems in some cases).

Thanks.. It was my first JLL, and it was difficult to find a "cookbook" style tutorial. One would be greatly appreciated.

PR #11157, opened 2025-05-10
MAGEMin v1.7.7

NicolasRiel:
- Fixes shear modulus calculation for slb database
- Fixes issue introduced in version 1.7.6 that was wrongfully outputting phases in mol% on a 1atom basis instead of mol%.

While double checking the output of MAGEMin with a reference software (Theriak-Domino), I realized that the unit displayed in the reference software were not the one used in the calculation. This discrepancy led to breaking MAGEMin in v1.7.6 and version v1.7.7 rolls back the changes...


NicolasRiel:
Thank you @giordano 
My apologies for pushing several new versions in such short amount of time.

PR #11167, opened 2025-05-12
New recipe: NodeJS_22_jll

asinghvi17:
Adds a new recipe (with updated checksums) based on NodeJS_20_jll for Node 22.

### DIFF
--- a/N/NodeJS/NodeJS_22/build_tarballs.jl
+++ b/N/NodeJS/NodeJS_22/build_tarballs.jl
@@ -1,0 +1,69 @@
+using BinaryBuilder
+
+name = "NodeJS_22"
+version = v"22.15.0"
+
+url_prefix = "https://nodejs.org/dist/v$version/node-v$version"
+sources = [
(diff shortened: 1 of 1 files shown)


ViralBShah (approved):

N/NodeJS/NodeJS_22/build_tarballs.jl
```diff
@@ -0,0 +1,69 @@
...
+        install -Dvm 0755 "${file}" "${bindir}/${file}"
+    done
+    mv node_modules "${bindir}/."
+fi
+install_license LICENSE
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
```
giordano, line 40: Why not supported_platforms?

ViralBShah: I thought this is just unpacking upstream binaries into our JLLs. Seems like just follows the pattern from earlier builds.

PR #11201, opened 2025-05-15
[aws_c_sdkutils] Update to version 0.2.4

quinnj2:
This PR updates aws_c_sdkutils to version 0.2.4. cc: @quinnj @Octogonapus

### DIFF
--- a/A/aws_c_sdkutils/build_tarballs.jl
+++ b/A/aws_c_sdkutils/build_tarballs.jl
@@ -7,7 +7,7 @@ version = v"0.2.4"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/awslabs/aws-c-sdkutils.git", "ba6a28fab7ed5d7f1b3b1d12eb672088be093824"),
+    GitSource("https://github.com/awslabs/aws-c-sdkutils.git", "f678bda9e21f7217e4bbf35e0d1ea59540687933"),
 ]
 
 # Bash recipe for building across all platforms


A/aws_c_auth/build_tarballs.jl
```diff
@@ -38,7 +38,7 @@ products = [
 dependencies = [
     Dependency("aws_c_cal_jll"; compat="0.8.3"),
     Dependency("aws_c_http_jll"; compat="0.9.3"),
-    Dependency("aws_c_sdkutils_jll"; compat="0.2.3"),
+    Dependency("aws_c_sdkutils_jll"; compat="0.2.4"),
```
giordano, line 41: This would need to update the version number of the package.  But also, changing the compat 0.2.3 -> 0.2.4 is likely unnecessary, it's more important. when changing across breaking version (e.g. if jumping to 0.3)

quinnj: Here's my thought process:
* I'm not super worried in this case about doing a new build/release of aws_c_auth, because it hasn't actually had a new release
* But the next time it _does_ have a release, I'd ideally use the latest version of all dependencies (that's how they have the aws crt libraries set up to work in practice, each essentially relying on the latest version of deps)
* But instead of having to go out and parse all the latest versions of each dependency on each new release, I'll just update any _downstream_ dependencies when an upstream has a new release.

Does that sound like an ok plan?

quinnj: Ah, but now I see that it's still trying to build the aws_c_auth library; is there a way I can add a comment to only build a single designated library? or to skip the others?

giordano: No, it's all or nothing

PR #11219, opened 2025-05-17
[NCCL] Update to 2.26 + arm64 support

ejmeitz:
Bumped version of library (allowed me to remove the patch) and also add aarch64. 

Edit: Welp seems ARM is not gonna be as easy as I thought

### DIFF
--- a/N/NCCL/build_tarballs.jl
+++ b/N/NCCL/build_tarballs.jl
@@ -10,7 +10,7 @@ include(joinpath(YGGDRASIL_DIR, "platforms", "cuda.jl"))
 name = "NCCL"
 version = v"2.26.5"
 
-MIN_CUDA_VERSION = v"111.8" # doesnt quite match NCCL actual support
+MIN_CUDA_VERSION = v"11.8" # doesnt quite match NCCL actual support
 
 sources = [
     GitSource("https://github.com/NVIDIA/nccl.git", "3000e3c797b4b236221188c07aa09c1f3a0170d4"),


N/NCCL/build_tarballs.jl
```diff
@@ -7,43 +7,48 @@ include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
 include(joinpath(YGGDRASIL_DIR, "platforms", "cuda.jl"))
 
 name = "NCCL"
-version = v"2.19.4"
+version = v"2.26.5"
+
+MIN_CUDA_VERSION = v"11.4" # previous min was 11.4
```
simonbyrne, line 12: It looks like NCCL 2.26.5 only supports CUDA 12.2, 12.4 and 12.9: https://docs.nvidia.com/deeplearning/nccl/release-notes/rel_2-26-5.html#rel_2-26-5

N/NCCL/build_tarballs.jl
```diff
@@ -7,43 +7,48 @@ include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
...
-mkdir -p ${TMPDIR}
+export CUDA_HOME=${prefix}/cuda;
+export PATH=$PATH:$CUDA_HOME/bin
+export CUDACXX=$CUDA_HOME/bin/nvcc
+
+# nvcc/nccl thinks the libraries are located inside lib64, but the SDK actually has them in lib
+ln -s ${CUDA_HOME}/lib ${CUDA_HOME}/lib64
 
 cd nccl
+make -j ${nproc} src.build CUDA_HOME=${CUDA_HOME} PREFIX=${prefix}
```
simonbyrne, line 36: Looking at my build history in https://github.com/JuliaPackaging/Yggdrasil/pull/7794, I tried this originally, but for some reason I decided the `pkg.txz.build` path was easier.

N/NCCL/build_tarballs.jl
```diff
@@ -60,9 +65,13 @@ for platform in platforms
...
 
+    cuda_ver = VersionNumber(tags(platform)["cuda"])
+
+    gcc_ver = cuda_ver >= v"11.6" ? v"10" : v"8"
+
     build_tarballs(ARGS, name, version, sources, script, [platform],
                    products, [dependencies; cuda_deps]; 
-                   lazy_artifacts=true,
-                   julia_compat="1.6", 
+                   lazy_artifacts=true, julia_compat="1.6",
```
simonbyrne, line 74: Not sure if it makes much difference, but you could safely bump this to 1.10

ejmeitz: I'll do that, I was wary of breaking whatever library you used this for.

N/NCCL/build_tarballs.jl
ejmeitz: Hm yeah, that's mildly annoying. Is best practice to just release a bunch of versions of NCCL then to get full converge of CUDA 12.x?

The original version you had build only supports a couple as well. The builds all work though? Like I was able to build NCCL 2.26.5 for CUDA 11.6 - 12.8. Doesn't mean it would work I guess.

ejmeitz:
Seems to work. Two questions I still have:

- I never had to set a target arch, is nvcc just inferring that from the compiler we specify

- NCCL docs specify pretty limited compat with CUDA. Typically only 3 versions, but the builds work fine past 11.8. Should we ship multiple NCCL versions to cover all off 12.x or just to with this since the build seems fine?

giordano:
Nvcc itself only compiles code for the GPU, which doesn't depend on the host CPU. For compiling host code it uses the host compiler, which goes through our compiler wrappers where we force the right target (CPU) architecture.

ViralBShah:
Good to merge?

ejmeitz:
Yep. Asked some other folks about the versioning thing. Shouldn't be a problem.

giordano:
Thanks for this!

PR #11228, opened 2025-05-19
[DACE] use native isinf and isnan methods

afossa:
Bump libcxxwrap to `v0.14.2`


D/DACE/build_tarballs.jl
```diff
@@ -51,7 +51,7 @@ products = [
 dependencies = [
     BuildDependency(PackageSpec(name="libjulia_jll")),
     BuildDependency("Eigen_jll"),
-    Dependency("libcxxwrap_julia_jll"; compat="~0.14.1"),
+    Dependency("libcxxwrap_julia_jll"; compat="~0.14.2"),
```
giordano, line 54: @barche is there a significant change in libcxxwrap_julia which warrants this?

PR #11242, opened 2025-05-22
[slope] update to 2.5.1 and update wrapper

jolars:
Update to libslope 2.5.1, and also update the wrapper to include support fort cross-validation. Also, use `Eigen::Map` to avoid copying and refactor the wrapper into separate source and header files.


giordano:
Any chance of moving the C++ files elsewhere?  We shouldn't really be hosting source code.

jolars:
> Any chance of moving the C++ files elsewhere? We shouldn't really be hosting source code.

Sorry, yes absolutely. Done now!

PR #11249, opened 2025-05-23
[GR] bump version

jheinen:
Bump GR version to 0.73.16
- add compatibility constraints for Bzip2, Cairo and Liftiff

### DIFF
--- a/G/GR/build_tarballs.jl
+++ b/G/GR/build_tarballs.jl
@@ -90,15 +90,15 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Bzip2_jll"),
-    Dependency("Cairo_jll"),
+    Dependency("Bzip2_jll"; compat="1.0.8"),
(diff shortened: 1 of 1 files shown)


G/GR/build_tarballs.jl
```diff
@@ -90,15 +90,15 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("Bzip2_jll"),
-    Dependency("Cairo_jll"),
+    Dependency("Bzip2_jll"; compat="1.0.8"),
```
barche, line 93: This should be `1.0.9` to satisfy the version needed by freetype.

G/GR/build_tarballs.jl
```diff
@@ -90,15 +90,15 @@ products = [
...
+    Dependency("Cairo_jll"; compat="1.16.1"),
```
barche, line 94: When i install just Qt6Base_jll it pulls in `Cairo_jll v1.18.4`, so possibly this needs to be bumped too.

PR #11256, opened 2025-05-24
update pmix to v6, fix freebsd

rayegun:


giordano:
Uhm, I think there's an issue with our pthreads mingw header file, maybe that explains the issue?

PR #11293, opened 2025-05-27
Add ROCmDeviceLibs 6.4.1

sheevy:
This is based on https://github.com/JuliaPackaging/Yggdrasil/pull/11084/files
I'm not sure how to test if this works.
I basically found an RPM and changed the URL and SHA-256, updated version number, and change the paths to to `mv` command.

Any guidance would be aprecciated.

### DIFF
--- a/R/ROCmDeviceLibs/ROCmDeviceLibs@6.4.1/build_tarballs.jl
+++ b/R/ROCmDeviceLibs/ROCmDeviceLibs@6.4.1/build_tarballs.jl
@@ -1,0 +1,40 @@
+using Pkg
+using BinaryBuilder
+
+const NAME = "ROCmDeviceLibs"
+const ROCM_PLATFORMS = [AnyPlatform()]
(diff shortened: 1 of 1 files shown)


sheevy:
+ @amontoison for visibility, hence tests now passed

amontoison (approved):

pxl-th:
This broke Julia 1.12 support, but can be fixed by not using this version of device libraries:
https://buildkite.com/julialang/amdgpu-dot-jl/builds/3139#019737eb-04b5-4599-b8a3-2671f26e6227

ROCm 6.4.1 device libraries use LLVM 19, which is not compatible with any released Julia version so far.
In future please make sure that AMDGPU.jl passes tests with this version of device libraries.

pxl-th:
Alternatively, Julia 1.13 compat should've been specified as now every 1.12 will try to pull this version

sheevy:
Created a PR to bump compat version: https://github.com/JuliaPackaging/Yggdrasil/pull/11352

PR #11296, opened 2025-05-27
MAGEMin v1.7.9

NicolasRiel:
- Fixes major source of artifacts and inaccurate phase equilibrium calculation
- Fixes mistake on DQF for orthoamphibole
- Fixes recently introduced mistake when outputting apfu


NicolasRiel:
As always thank you @giordano !

PR #11307, opened 2025-05-29
[GALAHAD] Upgrade to v5.2.1

amontoison:


amontoison:
`RANDOM_INIT` is a Fortran 2018 standard and requires at least GCC v9.1, I will close the PR to stop the remaining builds.
I will reopen it with a workaround later.

PR #11308, opened 2025-05-29
Update PCL_jll to v1.15

halleysfifthinc:
I tested that this built locally for the x86_64-linux-gnu target. I'm not sure whether the exact version requirement for boost is still required, but I followed the Arrow_jll build script which was just updated and still uses an exact compat for boost.


halleysfifthinc:
Several platforms[^1] failed because their existence is newer than the last FLANN build. #11321 should fix those failures.

Other failures due to bad (too high) CXX standard. Once #11321 has been merged, I will push the remaining updates to build.

[^1]: `aarch64-unknown-freebsd`, `aarch64-apple-darwin`, `armv6l-*`, and `riscv64-*`

PR #11313, opened 2025-05-29
[OSQP] Update to 1.0 and include CUDA and MKL builders

imciner2:
Update to latest upstream OSQP release - OSQP 1.0. This release includes first-class support for MKL and CUDA libraries, so introduce those as separate JLLs to separate out the library dependencies. (OSQP supports this by making new shared libraries for every backend that all expose the same API layer).


fingolfin:
```
ERROR: LoadError: Unsatisfiable requirements detected for package CUDA_SDK_static_jll [8d8e3561]:
--
  | CUDA_SDK_static_jll [8d8e3561] log:
  | ├─possible versions are: 10.2.89-12.8.1 or uninstalled
  | └─restricted to versions 12.9.0 by an explicit requirement — no versions left
```

imciner2:
Yea, the CUDA 12.9 update apparently exceeded the GitHub artifact size limit for the static libraries, so they weren't built. A while ago CMake wouldn't detect CUDA properly without the static CUDA runtime library, so that is why I have it here. I need to revisit that to see if maybe they relaxed CMake to work without it, or if there is another way to work around it.

Technici4n:
Hey, I managed to make cmake work with the dynamic CUDA sdk, if you want to have a look: https://github.com/JuliaPackaging/Yggdrasil/pull/11418/commits/b15a8035625c93e625782ac7323c8b4d2d6191e2.

imciner2:
Any comments @giordano? I could just merge this, but thought a second set of eyes would be good 😁.

giordano:
Looks overall good to me.  About the CUDA builds, do we still need to build for CUDA 11 nowadays?  Also, if you felt brave you could also try building for aarch64, but that'd likely complicate the recipe a bit, but it's up to you.

Technici4n:
Just a curiosity: did the size of the static CUDA artifacts eventually get reduced to <2Gb?

imciner2:
> Just a curiosity: did the size of the static CUDA artifacts eventually get reduced to <2Gb?

Yes, we introduced a new compression strategy for the artifacts that made the compressed size <2GB again, so we have static SDKs for 12.9 now.

imciner2:
> Looks overall good to me. About the CUDA builds, do we still need to build for CUDA 11 nowadays? Also, if you felt brave you could also try building for aarch64, but that'd likely complicate the recipe a bit, but it's up to you.

Is it possible for us to see how many people might be still downloading CUDA 11 artifacts? If we see a general decline in it, then I guess we could do something globally in Ygg to push up the version. As it is, we only build 11.4-11.8 (not the older ones)

I have never quite understood how we do platform versioning for CUDA. I was under the impression CUDA had "Minor version compatibility" (https://docs.nvidia.com/deploy/cuda-compatibility/minor-version-compatibility.html), so I thought we could just compile with one version and use it on others in the same major version, but we build for every minor here. @maleadt does this compatibility only apply to the driver/runtime and not the libraries like cuBLAS, cuSPARSE, etc? Basically could we have a single CUDA 11 build if we wanted?

maleadt:
> I have never quite understood how we do platform versioning for CUDA. I was under the impression CUDA had "Minor version compatibility" (https://docs.nvidia.com/deploy/cuda-compatibility/minor-version-compatibility.html), so I thought we could just compile with one version and use it on others in the same major version, but we build for every minor here. @maleadt does this compatibility only apply to the driver/runtime and not the libraries like cuBLAS, cuSPARSE, etc? Basically could we have a single CUDA 11 build if we wanted?

It should work like that: https://github.com/JuliaPackaging/Yggdrasil/blob/ac4c358fbfaf9109f04f7edc11438ae9378a45e2/platforms/cuda.jl#L51-L53
However, it's sometimes advantageous to build against newer versions of CUDA, because the code is then allowed to e.g. use newer APIs. So it can make sense to build for every minor version, but it depends on the software.

platforms/cuda.jl
```diff
@@ -316,4 +316,54 @@ function cuda_nvcc_redist_source(cuda_ver, arch)
...
+    ver = VersionNumber(tags(platform)["cuda"])
+    cuda_gpu_archs(ver)
+end
+
+"""
+    cuda_gpu_archs(cuda_ver)
+
+Returns a list of the GPU architectures supported by the given CUDA version.
+"""
+function cuda_gpu_archs(cuda_ver)
```
imciner2, line 334: @maleadt 

I added this function to help CMake get an architecture list to build for the various CUDA versions (just building with `all` gives errors sometimes, and I don't actually want everything). Do you see any issues with this listing, or are there any that I am missing?

(I know I don't have the Jetson ones in here).

PR #11317, opened 2025-05-29
[NCCL] Remove Symlink to lib64

ejmeitz:
Removes `ln -s ${CUDA_HOME}/lib ${CUDA_HOME}/lib64` which was causing issues in local builds that depend on NCCL


imciner2:
> Removes ln -s ${CUDA_HOME}/lib ${CUDA_HOME}/lib64 which was causing issues in local builds that depend on NCCL

What are the errors it is throwing with that symlink in place? In my past experiences, CMake only looked in the lib64 directory for the libraries, so that is why it needs the symlink.

PR #11322, opened 2025-05-30
Wizard recipe: OptFrame-v5.1.0

pedromxavier:

* Package name: OptFrame
* Version: v5.1.0

### DIFF
--- a/O/OptFrame/build_tarballs.jl
+++ b/O/OptFrame/build_tarballs.jl
@@ -1,0 +1,48 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "OptFrame"
+version = v"5.1.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/OptFrame/OptFrame.git", "d30011695eac4532f9592a17f91b39e9d44f0069")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/OptFrame/
+
+make optframe_lib
+
+cp ${WORKSPACE}/srcdir/OptFrame/build/optframe_lib.so ${bindir}
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = [
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("riscv64", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("x86_64", "freebsd"; ),
+    Platform("aarch64", "freebsd"; ),
+    Platform("i686", "windows"; ),
+    Platform("x86_64", "windows"; )
+]
+
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("bin", :binary)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"10.2.0")


O/OptFrame/build_tarballs.jl
```diff
@@ -0,0 +1,48 @@
...
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/OptFrame/
+
+make optframe_lib
+
+mkdir -p ${libdir}
+
+cp ${WORKSPACE}/srcdir/OptFrame/build/optframe_lib.so ${libdir}/optframe_lib.${dlext}
```
giordano:
```suggestion
install -Dvm755 build/optframe_lib.so "${libdir}/optframe_lib.${dlext}"
```

O/OptFrame/build_tarballs.jl
```diff
@@ -0,0 +1,48 @@
...
+    Platform("x86_64", "linux"; libc = "glibc"),
+    Platform("aarch64", "linux"; libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("riscv64", "linux"; libc = "glibc"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("x86_64", "freebsd"; ),
+    Platform("aarch64", "freebsd"; ),
+    Platform("i686", "windows"; ),
+    Platform("x86_64", "windows"; )
```
giordano: What's the problem with the other platforms?

pedromxavier: Hi @giordano, thanks for your help :) There were some compilation issues with the other plats, but they are pretty much fixed and I will be adding them back with the next tag.

pedromxavier:
x86 macos is the only missing platform to fix!

O/OptFrame/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("optframe_lib", :optframe)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"10.2.0", debug = true)
```
imciner2:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"10.2.0")
```
We don't want this in the version kept in Yggdrasil.

O/OptFrame/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name    = "OptFrame"
+version = v"6.0.0"
```
imciner2, line 6: Why this version number? The latest release is 5.1.0, and you are building from a more recent commit on master. If you use this version number now, then you can't use it when 6.0 is actually released.

O/OptFrame/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+    )
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/OptFrame/
+
+make optframe_lib
+
+install -Dvm755 build/optframe_lib.so "${libdir}/optframe_lib.${dlext}"
```
fingolfin, line 22: It seems this only installs this one file?

But that means it does not install a license file, which is a requirement...

Otherwise, the tests here seem to have passed last time, and we just missed this PR was ready :-(. Please don't hesitate to ping someone next time this happens.

imciner2: > But that means it does not install a license file, which is a requirement...

There is an automatic license install step if there isn't one installed normally and the upstream tarball includes a license named appropriately. See in the logfile:

> [20:23:06] Installing license file "/workspace/srcdir/OptFrame/LICENSE" to "/workspace/destdir/share/licenses/OptFrame/LICENSE"...

fingolfin: OK, good, I wasn't aware :-)

Anyway: I'd merge this, but the CI run was in June. Can you perhaps rebase this on latest master so that we get a fresh run?

PR #11324, opened 2025-05-30
Added skia library

stensmo:
Added Skia library. The 2D library in use by Chrome and other web browsers

### DIFF
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -16,7 +16,7 @@ sources = [
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = supported_platforms()
-filter!(p -> Sys.islinux(p) && libc(p) == "glibc" && call_abi(p) != "eabihf", platforms)
+filter!(p -> Sys.islinux(p) && libc(p) == "glibc" && arch(p) ∉ ("armv6l", "armv7l"), platforms)
 
 
 
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -10,13 +10,33 @@ sources = [
    GitSource("https://github.com/google/skia.git", "482de011c920d85fdbe21a81c45852655df6a809"),
    GitSource("https://github.com/stensmo/cskia.git", "3438e6efd3a4f27f43457db675ceb33da30c60cf"),
    DirectorySource("./bundled"),
+   GitSource("https://chromium.googlesource.com/chromium/src/buildtools.git","1760ff6d7267dd97ae1968c7bee9ce04a2a8489d"),
+   GitSource("https://android.googlesource.com/platform/external/dng_sdk.git","dbe0a676450d9b8c71bf00688bb306409b779e90"),
+   GitSource("https://chromium.googlesource.com/external/github.com/libexpat/libexpat.git","624da0f593bb8d7e146b9f42b06d8e6c80d032a3"),
+   GitSource("https://chromium.googlesource.com/chromium/src/third_party/freetype2.git","5d4e649f740c675426fbe4cdaffc53ee2a4cb954"),
+   GitSource("https://chromium.googlesource.com/external/github.com/harfbuzz/harfbuzz.git","ca3cd48fa3e06fa81d7c8a3f716cca44ed2de26a"),
+   GitSource("https://chromium.googlesource.com/chromium/deps/icu.git","364118a1d9da24bb5b770ac3d762ac144d6da5a4"),
+   GitSource("https://chromium.googlesource.com/codecs/libgav1.git","5cf722e659014ebaf2f573a6dd935116d36eadf1"),
+   GitSource("https://chromium.googlesource.com/chromium/deps/libjpeg_turbo.git","e14cbfaa85529d47f9f55b0f104a579c1061f9ad"),
+   GitSource("https://chromium.googlesource.com/external/gitlab.com/wg1/jpeg-xl.git","a205468bc5d3a353fb15dae2398a101dff52f2d3"),
+   GitSource("https://skia.googlesource.com/third_party/libpng.git","ed217e3e601d8e462f7fd1e04bed43ac42212429"),
+   GitSource("https://chromium.googlesource.com/webm/libwebp.git","845d5476a866141ba35ac133f856fa62f0b7445f"),
+   GitSource("https://chromium.googlesource.com/libyuv/libyuv.git","d248929c059ff7629a85333699717d7a677d8d96"),
+   GitSource("https://android.googlesource.com/platform/external/piex.git","bb217acdca1cc0c16b704669dd6f91a1b509c406"),
+   GitSource("https://chromium.googlesource.com/external/github.com/unicode-org/unicodetools","66a3fa9dbdca3b67053a483d130564eabc5fe095"),
+   GitSource("https://skia.googlesource.com/external/github.com/google/wuffs-mirror-release-c.git","e3f919ccfe3ef542cfc983a82146070258fb57f8"),
+   GitSource("https://chromium.googlesource.com/chromium/src/third_party/zlib","646b7f569718921d7d4b5b8e22572ff6c76f2596"),
+
 ]
+
+
 
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = supported_platforms()
 filter!(p -> Sys.islinux(p) || (Sys.isapple(p) && arch(p) == "aarch64"), platforms)
+
 
 
 
@@ -39,20 +59,32 @@ dependencies = [
 # Bash recipe for building across all platforms
 script = raw"""
 
-echo $target
-
 cd $WORKSPACE/srcdir/
 for f in ${WORKSPACE}/srcdir/patches/*.patch; do
     atomic_patch -p1 ${f}
 done
+
+mkdir skia/third_party/externals/
+shopt -s extglob
+
+# Rename libraries to the names used in Skia
+mv wuffs-mirror-release-c wuffs
+mv freetype2 freetype
+mv libexpat expat
+mv libjpeg_turbo libjpeg-turbo
+
+# Move dependencies to the correct location
+mv !(cskia|buildtools|skia|patches) skia/third_party/externals/
+
+mv buildtools/ skia/
+
+
 cd skia
 
 install_license LICENSE 
 
 bin/fetch-gn 
 
-# The || true is necessary, since the script fails for some Android deps. || true should be removed for future versions
-python3 tools/git-sync-deps || true
 
 cp ../cskia/capi/sk_capi.cpp src/base/
 cp ../cskia/capi/sk_capi.h src/base/
@@ -74,6 +106,9 @@ elif [[ "${target}" == armv6l-* ]]; then
     target_cpu=arm
 fi
 
+
+
+
 if [[ "${target}" == aarch64-apple-* ]]; then
 PLATFORM_ARGS="
 skia_use_x11=false \
@@ -85,7 +120,7 @@ skia_use_fonthost_mac=true
 else
 PLATFORM_ARGS="
 skia_use_fontconfig=true \
-skia_use_vulkan=true
+skia_use_vulkan=false
 "
 fi
 
@@ -98,6 +133,7 @@ is_official_build=true
 skia_enable_pdf=true
 skia_use_gl=true
 skia_use_harfbuzz=false
+skia_use_dng_sdk=true
 skia_use_system_expat=false
 skia_use_system_freetype2=false
 skia_use_system_icu=false
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -70,9 +70,7 @@ cp ../cskia/capi/sk_capi.h src/base/
 
 if [[ "${target}" == x86_64-* ]]; then
     target_cpu=x64
-fi
-
-if [[ "${target}" == aarch64-* ]]; then
+elif [[ "${target}" == aarch64-* ]]; then
     target_cpu=arm64
 fi
 
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -10,19 +10,21 @@ sources = [
    GitSource("https://github.com/google/skia.git", "482de011c920d85fdbe21a81c45852655df6a809"),
    GitSource("https://github.com/stensmo/cskia.git", "3438e6efd3a4f27f43457db675ceb33da30c60cf"),
    DirectorySource("./bundled"),
-   #Missing headers for freetype2
+   # Missing header ft2build.h for freetype2
    GitSource("https://chromium.googlesource.com/chromium/src/third_party/freetype2.git","5d4e649f740c675426fbe4cdaffc53ee2a4cb954"),
    GitSource("https://chromium.googlesource.com/libyuv/libyuv.git","d248929c059ff7629a85333699717d7a677d8d96"),
    # These two have some kind of source dependency. 
    GitSource("https://skia.googlesource.com/external/github.com/google/wuffs-mirror-release-c.git","e3f919ccfe3ef542cfc983a82146070258fb57f8"),
    GitSource("https://chromium.googlesource.com/chromium/src/third_party/zlib","646b7f569718921d7d4b5b8e22572ff6c76f2596"),
+   # Need 10.15 SDK for Mac
+   FileSource("https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/MacOSX10.15.sdk.tar.xz","2408d07df7f324d3beea818585a6d990ba99587c218a3969f924dfcc4de93b62"),
 ]
 
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = supported_platforms()
-filter!(p -> !Sys.iswindows(p)  && !(Sys.isapple(p)  && arch(p) ∈ ("x86_64",)), platforms)
+filter!(p -> !Sys.iswindows(p), platforms)
 platforms = expand_cxxstring_abis(platforms)
 
 # Remove musl && cxx03, since there is a bug preventing Skia to build
@@ -41,10 +43,7 @@ dependencies = [
     Dependency("libpng_jll"; compat="1.6.49")
     Dependency("libwebp_jll"; compat="1.5.0")
     Dependency("ICU_jll"; compat="76.1")
-    Dependency("Zlib_jll"; compat="1.3.1")
-    #Dependency("FreeType2_jll"; compat="2.13.4")
     Dependency("Expat_jll"; compat="2.6.5")
-    Dependency("wuffs_jll"; compat="0.3.4")
     Dependency(PackageSpec(name="Xorg_libX11_jll", uuid="4f6342f7-b3d2-589e-9d20-edeb45f2b2bc");)
     Dependency(PackageSpec(name="xkbcommon_jll", uuid="d8fb68d0-12a3-5cfd-a85a-d49703b185fd"); )
     Dependency(PackageSpec(name="Libglvnd_jll", uuid="7e76a0d4-f3c7-5321-8279-8d96eeed0f29"); )
@@ -100,8 +99,22 @@ elif [[ "${target}" == armv6l-* ]]; then
 fi
 
 
+if [[ "${target}" == x86_64-apple-darwin* ]]; then
+    # Work around the issue
+    export MACOSX_DEPLOYMENT_TARGET=10.15
+    # ...and install a newer SDK
+    rm -rf /opt/${target}/${target}/sys-root/System
+    tar --extract --file=${WORKSPACE}/srcdir/MacOSX10.15.sdk.tar.xz --directory="/opt/${target}/${target}/sys-root/." --strip-components=1 MacOSX10.15.sdk/System MacOSX10.15.sdk/usr
 
-if [[ "${target}" == *-apple-* ]]; then
+PLATFORM_ARGS="
+skia_use_x11=false \
+target_os=\\"mac\\" 
+skia_use_metal=true  
+skia_enable_fontmgr_fontconfig=false
+skia_use_fonthost_mac=true
+skia_use_dng_sdk=true
+"   
+elif [[ "${target}" == *-apple-* ]]; then
 PLATFORM_ARGS="
 skia_use_x11=false \
 target_os=\\"mac\\" 
@@ -118,7 +131,6 @@ skia_use_dng_sdk=false
 else
 PLATFORM_ARGS="
 skia_use_fontconfig=true \
-skia_use_dng_sdk=true
 "
 fi
 
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -102,4 +102,4 @@ install -Dvm 755 "libskia.${dlext}" "${libdir}/libskia.${dlext}"
 """
 
 
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.10")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version= v"11.1.0", julia_compat="1.10")
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -90,7 +90,7 @@ skia_use_system_zlib=false
 skia_use_vulkan=true
 extra_cflags=[\\"-fpic\\", \\"-fvisibility=default\\"]
 "
-bin/gn gen out/Static --args="$ARGS"
+bin/gn gen out/Dynamic --args="$ARGS"
 
 ninja -j${nproc} -C out/Dynamic
 
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -68,7 +68,7 @@ mv wuffs-mirror-release-c wuffs
 mv freetype2 freetype
 
 # Move dependencies to the correct location
-mv !(cskia|buildtools|skia|patches) skia/third_party/externals/
+mv !(cskia|skia|patches|MacOSX10.15.sdk.tar.xz) skia/third_party/externals/
 
 
 cd skia
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -8,7 +8,7 @@ version = v"0.40.0"
 # Collection of sources required to complete build
 sources = [
    GitSource("https://github.com/google/skia.git", "482de011c920d85fdbe21a81c45852655df6a809"),
-   GitSource("https://github.com/stensmo/cskia.git", "9a7c1f23ff12fbc94eed577ea366b0e51fedb119"),
+   GitSource("https://github.com/stensmo/cskia.git", "3438e6efd3a4f27f43457db675ceb33da30c60cf"),
    DirectorySource("./bundled"),
 ]
 
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -102,4 +102,4 @@ install -Dvm 755 "libskia.${dlext}" "${libdir}/libskia.${dlext}"
 """
 
 
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version= v"11.1.0", julia_compat="1.10")
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version= v"11.1.0", preferred_llvm_version = v"15.0.7", julia_compat="1.10")
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -88,7 +88,6 @@ skia_use_system_libpng=false
 skia_use_system_libwebp=false
 skia_use_system_zlib=false
 skia_use_vulkan=true
-extra_cflags=[\\"-fpic\\", \\"-fvisibility=default\\"]
 "
 bin/gn gen out/Dynamic --args="$ARGS"
 
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -10,23 +10,15 @@ sources = [
    GitSource("https://github.com/google/skia.git", "482de011c920d85fdbe21a81c45852655df6a809"),
    GitSource("https://github.com/stensmo/cskia.git", "3438e6efd3a4f27f43457db675ceb33da30c60cf"),
    DirectorySource("./bundled"),
-   GitSource("https://chromium.googlesource.com/chromium/src/buildtools.git","1760ff6d7267dd97ae1968c7bee9ce04a2a8489d"),
    GitSource("https://android.googlesource.com/platform/external/dng_sdk.git","dbe0a676450d9b8c71bf00688bb306409b779e90"),
-   GitSource("https://chromium.googlesource.com/external/github.com/libexpat/libexpat.git","624da0f593bb8d7e146b9f42b06d8e6c80d032a3"),
+   #Missing headers for freetype2
    GitSource("https://chromium.googlesource.com/chromium/src/third_party/freetype2.git","5d4e649f740c675426fbe4cdaffc53ee2a4cb954"),
-   GitSource("https://chromium.googlesource.com/external/github.com/harfbuzz/harfbuzz.git","ca3cd48fa3e06fa81d7c8a3f716cca44ed2de26a"),
-   GitSource("https://chromium.googlesource.com/chromium/deps/icu.git","364118a1d9da24bb5b770ac3d762ac144d6da5a4"),
-   GitSource("https://chromium.googlesource.com/codecs/libgav1.git","5cf722e659014ebaf2f573a6dd935116d36eadf1"),
-   GitSource("https://chromium.googlesource.com/chromium/deps/libjpeg_turbo.git","e14cbfaa85529d47f9f55b0f104a579c1061f9ad"),
-   GitSource("https://chromium.googlesource.com/external/gitlab.com/wg1/jpeg-xl.git","a205468bc5d3a353fb15dae2398a101dff52f2d3"),
-   GitSource("https://skia.googlesource.com/third_party/libpng.git","ed217e3e601d8e462f7fd1e04bed43ac42212429"),
-   GitSource("https://chromium.googlesource.com/webm/libwebp.git","845d5476a866141ba35ac133f856fa62f0b7445f"),
    GitSource("https://chromium.googlesource.com/libyuv/libyuv.git","d248929c059ff7629a85333699717d7a677d8d96"),
    GitSource("https://android.googlesource.com/platform/external/piex.git","bb217acdca1cc0c16b704669dd6f91a1b509c406"),
    GitSource("https://chromium.googlesource.com/external/github.com/unicode-org/unicodetools","66a3fa9dbdca3b67053a483d130564eabc5fe095"),
+   # These two have some kind of source dependency. 
    GitSource("https://skia.googlesource.com/external/github.com/google/wuffs-mirror-release-c.git","e3f919ccfe3ef542cfc983a82146070258fb57f8"),
    GitSource("https://chromium.googlesource.com/chromium/src/third_party/zlib","646b7f569718921d7d4b5b8e22572ff6c76f2596"),
-
 ]
 
 
@@ -49,6 +41,14 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("Fontconfig_jll"; compat="2.16.0")
+    Dependency("JpegTurbo_jll"; compat="3.1.1")
+    Dependency("libpng_jll"; compat="1.6.49")
+    Dependency("libwebp_jll"; compat="1.5.0")
+    Dependency("ICU_jll"; compat="76.1")
+    Dependency("Zlib_jll"; compat="1.3.1")
+    #Dependency("FreeType2_jll"; compat="2.13.4")
+    Dependency("Expat_jll"; compat="2.6.5")
+    Dependency("wuffs_jll"; compat="0.3.4")
     Dependency(PackageSpec(name="Xorg_libX11_jll", uuid="4f6342f7-b3d2-589e-9d20-edeb45f2b2bc");)
     Dependency(PackageSpec(name="xkbcommon_jll", uuid="d8fb68d0-12a3-5cfd-a85a-d49703b185fd"); )
     Dependency(PackageSpec(name="Libglvnd_jll", uuid="7e76a0d4-f3c7-5321-8279-8d96eeed0f29"); )
@@ -71,13 +71,12 @@ shopt -s extglob
 # Rename libraries to the names used in Skia
 mv wuffs-mirror-release-c wuffs
 mv freetype2 freetype
-mv libexpat expat
-mv libjpeg_turbo libjpeg-turbo
+
 
 # Move dependencies to the correct location
 mv !(cskia|buildtools|skia|patches) skia/third_party/externals/
 
-mv buildtools/ skia/
+
 
 
 cd skia
@@ -139,12 +138,12 @@ is_official_build=true
 skia_enable_pdf=true
 skia_use_gl=true
 skia_use_harfbuzz=false
-skia_use_system_expat=false
+skia_use_system_expat=true
 skia_use_system_freetype2=false
-skia_use_system_icu=false
-skia_use_system_libjpeg_turbo=false
-skia_use_system_libpng=false
-skia_use_system_libwebp=false
+skia_use_system_icu=true
+skia_use_system_libjpeg_turbo=true
+skia_use_system_libpng=true
+skia_use_system_libwebp=true
 extra_cflags=[\\"-fpic\\", \\"-fvisibility=default\\"]
 $PLATFORM_ARGS
 "
--- a/S/Skia/build_tarballs.jl
+++ b/S/Skia/build_tarballs.jl
@@ -30,6 +30,8 @@ platforms = expand_cxxstring_abis(platforms)
 # Remove musl && cxx03, since there is a bug preventing Skia to build
 filter!(p -> !(cxxstring_abi(p) == "cxx03" && libc(p) == "musl"), platforms)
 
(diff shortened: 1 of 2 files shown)


giordano:
I don't know which of #11323 or #11324 we should look at.

stensmo:
Sorry, I made some mistakes. You can look at 11324. It was difficult to make this build, since the build system was completely different to the normal ones in Julia. I will try to bump the Fontconfig_jll and other dependencies as well so it builds for RISC-V. Skia can be built for all supported platforms, but I have not figured out how to to the Windows/Mac builds yet. 

Some expert help would be greatly appreciated.

stensmo:
Do I need to do anything to get this first version merged?

imciner2:
This build script is very confusing to follow and understand. For instance, why do you have each platform being built in its own call to `build_tarballs`? I can't see any differences other than CPU type in the build scripts. Please clean this up a bit, unify parts that are the same/similar between platforms, and ideally write the build script in a way you can just do a single `build_tarballs` call.

stensmo:
@imciner2 I updated the build script now

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,101 @@
...
+version = v"0.0.1"
+
+# Collection of sources required to complete build
+sources = [
+
+   GitSource("https://github.com/google/skia.git", "482de011c920d85fdbe21a81c45852655df6a809"),
+    GitSource("https://github.com/stensmo/cskia.git", "500cdca61e2105555f339fa363f55e30696b009f"),
+    DirectorySource("./bundled")
+
+]
```
imciner2:
```suggestion
sources = [
    GitSource("https://github.com/google/skia.git", "482de011c920d85fdbe21a81c45852655df6a809"),
    GitSource("https://github.com/stensmo/cskia.git", "500cdca61e2105555f339fa363f55e30696b009f"),
    DirectorySource("./bundled")
]
```
Just minor style.

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,101 @@
...
+
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+
+
+platforms = [
+    
+
+     Platform("x86_64", "linux"; libc = "glibc"),
```
imciner2:
```suggestion
platforms = [
     Platform("x86_64", "linux"; libc = "glibc"),
```

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,101 @@
...
+
+
+ninja -j${nproc} -C out/Static
+
+cd out/Static/
+
+clang++ -shared -o libskia.so   -Wl,--whole-archive libskia.a -Wl,--no-whole-archive libfreetype2.a libjpeg.a libbentleyottmann.a libcompression_utils_portable.a libdng_sdk.a libjsonreader.a libpathkit.a  libpiex.a libpng.a libskcms.a libsksg.a libskshaper.a libskunicode_core.a libskunicode_icu.a libsvg.a -fpic -fvisibility=default -lstdc++ -dl -lfontconfig -lGL 
+
+mkdir -p ${prefix}/lib/
+cp libskia.so ${prefix}/lib/
```
imciner2:
```suggestion
install -Dvm 755 "libskia.${dlext}" "${libdir}/libskia.${dlext}"
```
Cleaner command, and it ensures permissions are correct.

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,101 @@
...
+if [[ "${target}" == powerpc64le-* ]]; then
+bin/gn gen out/Static --args='target_cpu="powerpc64le" cc="clang" cxx="clang++" is_official_build=true skia_use_system_libjpeg_turbo=false skia_use_system_libpng=false skia_use_system_libwebp=false skia_use_system_zlib=false skia_use_vulkan=true skia_use_system_freetype2=false skia_use_fontconfig=true skia_enable_pdf=true  skia_use_system_icu=false skia_use_system_expat=false skia_use_harfbuzz=false skia_use_vulkan=true skia_use_gl=true extra_cflags=["-fpic", "-fvisibility=default"]'
+fi
+
+
+ninja -j${nproc} -C out/Static
+
+cd out/Static/
+
+clang++ -shared -o libskia.so   -Wl,--whole-archive libskia.a -Wl,--no-whole-archive libfreetype2.a libjpeg.a libbentleyottmann.a libcompression_utils_portable.a libdng_sdk.a libjsonreader.a libpathkit.a  libpiex.a libpng.a libskcms.a libsksg.a libskshaper.a libskunicode_core.a libskunicode_icu.a libsvg.a -fpic -fvisibility=default -lstdc++ -dl -lfontconfig -lGL
```
imciner2: Writing out the entire compilation command like this is a bit cumbersome and isn't always the most maintainable. Is there not a build command inside the package that will make a shared library for us?

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,101 @@
...
+
+fi
+
+if [[ "${target}" == riscv64-* ]]; then
+bin/gn gen out/Static --args='target_cpu="riscv" cc="clang" cxx="clang++" is_official_build=true skia_use_system_libjpeg_turbo=false skia_use_system_libpng=false skia_use_system_libwebp=false skia_use_system_zlib=false skia_use_vulkan=true skia_use_system_freetype2=false skia_use_fontconfig=true skia_enable_pdf=true  skia_use_system_icu=false skia_use_system_expat=false skia_use_harfbuzz=false skia_use_vulkan=true skia_use_gl=true extra_cflags=["-fpic", "-fvisibility=default"]'
+fi
+
+if [[ "${target}" == powerpc64le-* ]]; then
+bin/gn gen out/Static --args='target_cpu="powerpc64le" cc="clang" cxx="clang++" is_official_build=true skia_use_system_libjpeg_turbo=false skia_use_system_libpng=false skia_use_system_libwebp=false skia_use_system_zlib=false skia_use_vulkan=true skia_use_system_freetype2=false skia_use_fontconfig=true skia_enable_pdf=true  skia_use_system_icu=false skia_use_system_expat=false skia_use_harfbuzz=false skia_use_vulkan=true skia_use_gl=true extra_cflags=["-fpic", "-fvisibility=default"]'
+fi
```
imciner2: Is the only difference between these commands the `target_cpu` argument that is being passed? If so, I think this is much better done by doing an if/elseif to set one variable with the CPU name, and then used that variable in a command after it to actually do the call to the build system. (and even if there is another command or two that change value, I think it is still better to do it this way).

Also, please break each build option into its own line. It is very hard to keep track of the options now, and in the future if changes have to be made, having them on their own lines allows better tracking of changes in the git history.

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,101 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Skia"
+version = v"0.0.1"
```
imciner2: Customarily we have used the upstream version here, but if you want to iterate on the C API part for a bit, then this would be fine (so you don't mess up upstream version numbering).

My suggestion then is to add a comment in the sources field to say which version of skia is actually being built here.

imciner2 (changes requested):

S/Skia/build_tarballs.jl
stensmo: There is a flag is_component_build=true which in theory should do the job, but I could not make that work.

S/Skia/build_tarballs.jl
stensmo: There is some C-api work left, but not that much. When the init code for Vulkan and Metal is working (OpenGL already works) , the Skia library could be set to the Skia version.

stensmo:
@imciner2 Can you please tell me how to set a variable? There is no documentation on how to do that.

giordano:
What variable?

stensmo:
It is requested that " I think this is much better done by doing an if/elseif to set one variable with the CPU name, and then used that variable in a command after it to actually do the call to the build system." I tried various ways to set variables (as you would in bash) and tried to look for documentation, but no sucess. The variable that imciner2 requested me to use is target_cpu.

imciner2:
The variable is just a bash variable in the build script. See this for instance: https://github.com/JuliaPackaging/Yggdrasil/blob/9b2ba497e1f5d84dd8c0f34ebdb7eb9393cb73ee/S/SCIP/build_tarballs.jl#L19. In there it sets a variable called `LDFLAGS` based on the target and then uses that in the build command.

stensmo:
Now a variable is introduced. It needed some escaping...

stensmo:
What do I need to do next?

stensmo:
@imciner2 I updated bild_tarballs with a variable. Can you please state what additional things need to be done.

stensmo:
@imciner2  Can you please respond to this? I have tried for several hours to fix one of your comments, and there is no documentation what so ever on how julia and bash character escaping interacts. If you feel that multiline is absolutely necessary, please show me how this is achived.

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,118 @@
...
+
+cp ../cskia/capi/sk_capi.cpp src/base/
+cp ../cskia/capi/sk_capi.h src/base/
+
+
+if [[ "${target}" == x86_64-* ]]; then
+    target_cpu=x64
+fi
+
+if [[ "${target}" == aarch64-* ]]; then
```
giordano:
```suggestion
elif [[ "${target}" == aarch64-* ]]; then
```

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,118 @@
...
+    target_cpu=arm64
+fi
+
+if [[ "${target}" == riscv64-* ]]; then
```
giordano:
```suggestion
elif [[ "${target}" == riscv64-* ]]; then
```

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,118 @@
...
+    target_cpu=riscv
+fi
+
+if [[ "${target}" == powerpc64le-* ]]; then
```
giordano:
```suggestion
elif [[ "${target}" == powerpc64le-* ]]; then
```

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,118 @@
...
+    target_cpu=powerpc64le
+fi
+
+if [[ "${target}" == i686-* ]]; then
```
giordano:
```suggestion
elif [[ "${target}" == i686-* ]]; then
```

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,118 @@
...
+else
+    clang++ -shared -o libskia.$dlext   -Wl,--whole-archive libskia.a -Wl,--no-whole-archive libfreetype2.a libjpeg.a libbentleyottmann.a libcompression_utils_portable.a libdng_sdk.a libjsonreader.a libpathkit.a  libpiex.a libpng.a libskcms.a libsksg.a libskshaper.a libskunicode_core.a libskunicode_icu.a libsvg.a -fpic -fvisibility=default -lstdc++ -dl -lfontconfig -lGL
+fi
+
+
+install -Dvm 755 "libskia.${dlext}" "${libdir}/libskia.${dlext}"
+"""
+
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.10", preferred_gcc_version = v"11.1.0", preferred_llvm_version = v"15.0.7")
```
giordano: Why llvm 15?  That looks like an oddly specific version, unless there's a strong reason for it.

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,118 @@
...
+cd $WORKSPACE/srcdir/
+for f in ${WORKSPACE}/srcdir/patches/*.patch; do
+    atomic_patch -p1 ${f}
+done
+cd skia
+
+install_license LICENSE 
+
+bin/fetch-gn 
+python3 tools/git-sync-deps || true
```
giordano: Why `|| true`?

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,118 @@
...
+ninja -j${nproc} -C out/Static
+
+cd out/Static/
+
+
+if [[ "${target}" == *-apple-* ]]; then
+    clang++ -shared -o libskia.$dlext  libjpeg.a libbentleyottmann.a libcompression_utils_portable.a libdng_sdk.a libjsonreader.a libpathkit.a  libpiex.a libpng.a libskcms.a libsksg.a libskshaper.a libskunicode_core.a libskunicode_icu.a libsvg.a -fpic -fvisibility=default -lstdc++ -dl 
+else
+    clang++ -shared -o libskia.$dlext   -Wl,--whole-archive libskia.a -Wl,--no-whole-archive libfreetype2.a libjpeg.a libbentleyottmann.a libcompression_utils_portable.a libdng_sdk.a libjsonreader.a libpathkit.a  libpiex.a libpng.a libskcms.a libsksg.a libskshaper.a libskunicode_core.a libskunicode_icu.a libsvg.a -fpic -fvisibility=default -lstdc++ -dl -lfontconfig -lGL
+fi
```
giordano: How about?
```suggestion
clang++ -shared -o libskia.${dlext} $(flagon -Wl,--whole-archive) libskia.a $(flagon -Wl,--no-whole-archive) libfreetype2.a libjpeg.a libbentleyottmann.a libcompression_utils_portable.a libdng_sdk.a libjsonreader.a libpathkit.a  libpiex.a libpng.a libskcms.a libsksg.a libskshaper.a libskunicode_core.a libskunicode_icu.a libsvg.a -fpic -dl -lfontconfig -lGL
```
?  Why using `clang++` instead of c++?

But also, I agree with @imciner2 that writing out an explicit compiler invocation like that isn't sustainable: if the upstream source code changes what they link to, this won't work anymore.

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,118 @@
...
+
+if [[ "${target}" == aarch64-apple-* ]]; then
+    bin/gn gen out/Static --args='target_cpu="arm64" target_os="mac" skia_use_metal=true skia_use_x11=false skia_enable_fontmgr_fontconfig=false  skia_use_fonthost_mac=true cc="clang" cxx="clang++" is_official_build=true  skia_use_freetype=false  skia_use_system_libjpeg_turbo=false skia_use_system_libpng=false skia_use_system_libwebp=false skia_use_system_zlib=false skia_use_vulkan=true skia_use_fontconfig=false skia_enable_pdf=true  skia_use_system_icu=false skia_use_system_expat=false skia_use_harfbuzz=false skia_use_vulkan=true skia_use_gl=true extra_cflags=["-fpic", "-fvisibility=default"]'
+fi
+
+if [[ "${target}" == *-linux-* ]]; then
+    bin/gn gen out/Static --args='target_cpu="'$target_cpu'" cc="clang" cxx="clang++" is_official_build=true skia_use_system_libjpeg_turbo=false skia_use_system_libpng=false skia_use_system_libwebp=false skia_use_system_zlib=false skia_use_vulkan=true skia_use_system_freetype2=false skia_use_fontconfig=true skia_enable_pdf=true  skia_use_system_icu=false skia_use_system_expat=false skia_use_harfbuzz=false skia_use_vulkan=true skia_use_gl=true extra_cflags=["-fpic", "-fvisibility=default"]'
+fi
```
giordano: Why not using `${target_cpu}` which you defined above for the macOS branch?  What are the differences between the two branches?  Would it be possible to consolidate everything into a single invocation to `bin/gn` with the different arguments coming from variables (as for the case of `${target_cpu}`)?

As we're trying to say, having so much code to handle compilation of third-party code is very cumbersome for downstream projects.

S/Skia/build_tarballs.jl
stensmo: It's because the git-sync-deps fails for some Android stuff, but that's not important in this case. I should probably write a comment about that.

S/Skia/build_tarballs.jl
stensmo: I can fix the macOS branch. Currently it only works for aarch 64, since the Darwin version used by default is too old for the x64 build, and I don't know how to change that in Julia (yet).

I understand it's very cumbersome. Skia is one of the few libraries which is actually still developed, so in my opinion it is worth it, even if it's a bit cumbersome. Skia is supported for all relvant platforms (since all platforms need a browser).

S/Skia/build_tarballs.jl
stensmo: > Why llvm 15? That looks like an oddly specific version, unless there's a strong reason for it.

Skia needs a new version of CLANG, and I decided Julia 1.10 and above would be a good target for the Skia library. As I understand it, Julia 1.10 uses that version.

stensmo:
@giordano Thanks a lot for your suggestion, but Skia did not build on my machine with all of your suggestions. Since the Mac version creates so much pollution in the build script, I decided to remove it. This is pretty much how clean it's going to get, so I hope you can accept it. I really did my very best.

S/Skia/build_tarballs.jl
giordano: > Skia needs a new version of CLANG

In BinaryBuilder the default clang version is the latest available, which currently is llvm 19, so I don't understand why you want to use a version which is 2 years older.

> and I decided Julia 1.10 and above would be a good target for the Skia library

Why? There's _no_ need to use as compiler toolchain the same llvm version used somewhere in Julia.

stensmo:
@giordano Thanks a lot for your suggestions. I assume you are now blocking this because it has the wrong LLVM version. If you would run the build, you would notice it will not build with different LLVM versions. I tried them all 16, 17, 18 (19 is not default on my machine, but 18.1.7). In LLVM 17 there is a bug. In version 16 and 18 it won't link against fontconfig and OpenGL. Possibly due to some miconfiguration in these packages. The version I have provided works and I have tested it.

S/Skia/build_tarballs.jl
fingolfin: This comment still needs to be addressed

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,105 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "Skia"
+version = v"0.0.1"
+
+# Collection of sources required to complete build
+sources = [
+
```
fingolfin:
```suggestion
```

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,105 @@
...
+   GitSource("https://github.com/google/skia.git", "482de011c920d85fdbe21a81c45852655df6a809"),
+   GitSource("https://github.com/stensmo/cskia.git", "500cdca61e2105555f339fa363f55e30696b009f"),
+   DirectorySource("./bundled")
+
```
fingolfin:
```suggestion
   DirectorySource("./bundled"),
```

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,105 @@
...
+]
+
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+
+
```
fingolfin:
```suggestion
```

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,105 @@
...
+platforms = [
+     Platform("x86_64", "linux"; libc = "glibc"),
+     Platform("aarch64", "linux"; libc = "glibc"),
+     Platform("riscv64", "linux"; libc = "glibc"), 
+     Platform("powerpc64le", "linux"; libc = "glibc"), 
+     Platform("i686", "linux"; libc = "glibc"),
+]
```
fingolfin: So basically all Linux variants with glibc? But without 32bit ARM (is that intentional?) How about
```suggestion
platforms = supported_platforms()
filter!(p -> Sys.islinux(p) && libc(p) == "glibc", platforms)
```
which has the advantage of automatically getting support for further archs if needed.

S/Skia/build_tarballs.jl
fingolfin: You added a comment which is good, but this is still fishy. If this step is not needed at all, it should be removed. If it is needed, but only parts of it, and the other parts which are not needed are OK to fail, then perhaps those should not be downloaded *at all*. Then:
- we avoid overhead for downloading stuff we don't even need,
- we can remove the `|| true`
- we can detect future problems in this line that *do* matter, instead of silencing them all unconditionally.

Looking at https://github.com/google/skia/blob/main/tools/git-sync-deps it seems to support options to control it?



But really, what does it download? Aren't dependencies supposed to come from JLLs?
What happens if you drop this line?

S/Skia/build_tarballs.jl
fingolfin: still needs to be addressed

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,105 @@
...
+    target_cpu=powerpc64le
+fi
+
+if [[ "${target}" == i686-* ]]; then
+    target_cpu=x86
+fi
+
+
+
+bin/gn gen out/Static --args='target_cpu="'$target_cpu'" cc="clang" cxx="clang++" is_official_build=true skia_use_system_libjpeg_turbo=false skia_use_system_libpng=false skia_use_system_libwebp=false skia_use_system_zlib=false skia_use_vulkan=true skia_use_system_freetype2=false skia_use_fontconfig=true skia_enable_pdf=true  skia_use_system_icu=false skia_use_system_expat=false skia_use_harfbuzz=false skia_use_vulkan=true skia_use_gl=true extra_cflags=["-fpic", "-fvisibility=default"]'
```
fingolfin: This will be difficult to maintain and tweak in the future. I recommend using line continuations to make this readable (and in the future, "diff'able", i.e. make adding, modifying or removing options easier to see), i.e. like so:
```suggestion
bin/gn gen out/Static --args='target_cpu="'$target_cpu'" \
  cc="clang" cxx="clang++" \
  is_official_build=true \
  skia_use_system_libjpeg_turbo=false \
  skia_use_system_libpng=false \
  skia_use_system_libwebp=false \
  skia_use_system_zlib=false \
  skia_use_vulkan=true \
  skia_use_system_freetype2=false \
  skia_use_fontconfig=true \
  skia_enable_pdf=true \
  skia_use_system_icu=false \
  skia_use_system_expat=false \
  skia_use_harfbuzz=false \
  skia_use_vulkan=true \
  skia_use_gl=true \
  extra_cflags=["-fpic", "-fvisibility=default"]'
```
I would also suggest to sort those options. That will make it obvious when you do things like specifying an option twice (here: `skia_use_vulkan`).


Oh, and: are you sure it is OK to specify `is_official_build=true` ?

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,105 @@
...
+
+
+clang++ -shared -o libskia.${dlext} $(flagon -Wl,--whole-archive) libskia.a $(flagon -Wl,--no-whole-archive) libfreetype2.a libjpeg.a libbentleyottmann.a libcompression_utils_portable.a libdng_sdk.a libjsonreader.a libpathkit.a  libpiex.a libpng.a libskcms.a libsksg.a libskshaper.a libskunicode_core.a libskunicode_icu.a libsvg.a -fpic -dl -lfontconfig -lGL
+
+
+install -Dvm 755 "libskia.${dlext}" "${libdir}/libskia.${dlext}"
+"""
+
+
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version= v"11.1.0", preferred_llvm_version = v"15.0.7", julia_compat="1.10")
```
fingolfin, line 172: why "15.0.7" and not "15" ?

S/Skia/build_tarballs.jl
fingolfin: OK, so `is_official_build=true` is explained [here](https://skia.org/docs/user/build/#is_official_build-and-third-party-dependencies) and it sounds to me as if we indeed want that, so I withdraw my question about it :-).

However I am worried about your use of `skia_use_system_libpng=false` etc. -- but why? Shouldn't we take those from our JLLs for them?

S/Skia/build_tarballs.jl
stensmo: 15.0.7 is what you get if you run the Wizard. But I can change that.

stensmo:
python3 tools/git-sync-deps is absolutely necessary to complete the build. This script fails for some reason, but still it syncs enough deps to complete the build (except for Android). I hope Google will fix the error in the next version of Skia, so the || true can be removed.

S/Skia/build_tarballs.jl
stensmo: > This will be difficult to maintain and tweak in the future. I recommend using line continuations to make this readable (and in the future, "diff'able", i.e. make adding, modifying or removing options easier to see), i.e. like so:
> 
> I would also suggest to sort those options. That will make it obvious when you do things like specifying an option twice (here: `skia_use_vulkan`).
> 
> Oh, and: are you sure it is OK to specify `is_official_build=true` ?

I built a version showing you that it is not so simple. I have tried your exact approach to make it multiline. It's because the line is started with ' (and ended with '), so your "normal" approach won't work. I spent several hours on the request to split it into multiple lines, but I did not find a solution. Therefore I will revert again to a single line which works (albeit could be cumbersome to read).

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,87 @@
...
+elif [[ "${target}" == riscv64-* ]]; then
+    target_cpu=riscv
+elif [[ "${target}" == powerpc64le-* ]]; then
+    target_cpu=powerpc64le
+elif [[ "${target}" == i686-* ]]; then
+    target_cpu=x86
+fi
+
+
+bin/gn gen out/Static --args='target_cpu="'$target_cpu'" cc="clang" cxx="clang++" is_official_build=true skia_use_system_libjpeg_turbo=false skia_use_system_libpng=false skia_use_system_libwebp=false skia_use_system_zlib=false skia_use_system_freetype2=false skia_use_fontconfig=true skia_enable_pdf=true skia_use_system_icu=false skia_use_system_expat=false skia_use_harfbuzz=false skia_use_vulkan=true skia_use_gl=true extra_cflags=["-fpic", "-fvisibility=default"]'
```
fingolfin: I am sorry to hear you spent hours on make multiline string arguments work, but it really isn't that hard -- I did in fact miss that the argument was in a string, but that actually makes it easier, not harder:
```suggestion
bin/gn gen out/Static --args="
target_cpu=$target_cpu
cc=clang
cxx=clang++
is_official_build=true
skia_enable_pdf=true
skia_use_fontconfig=true
skia_use_gl=true
skia_use_harfbuzz=false
skia_use_system_expat=false
skia_use_system_freetype2=false
skia_use_system_icu=false
skia_use_system_libjpeg_turbo=false
skia_use_system_libpng=false
skia_use_system_libwebp=false
skia_use_system_zlib=false
skia_use_vulkan=true
extra_cflags=[\"-fpic\", \"-fvisibility=default\"]
"
```
Or perhaps better to split this for readability:
```suggestion
ARGS="
target_cpu=$target_cpu
cc=clang
cxx=clang++
is_official_build=true
skia_enable_pdf=true
skia_use_fontconfig=true
skia_use_gl=true
skia_use_harfbuzz=false
skia_use_system_expat=false
skia_use_system_freetype2=false
skia_use_system_icu=false
skia_use_system_libjpeg_turbo=false
skia_use_system_libpng=false
skia_use_system_libwebp=false
skia_use_system_zlib=false
skia_use_vulkan=true
extra_cflags=[\"-fpic\", \"-fvisibility=default\"]
"

bin/gn gen out/Static --args="$ARGS"
```

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,87 @@
...
+   DirectorySource("./bundled"),
+]
+
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+filter!(p -> Sys.islinux(p) && libc(p) == "glibc" && call_abi(p) != "eabihf", platforms)
+
+
```
fingolfin: According to the auditor warnings in the build logs, you also need
```suggestion
platforms = expand_cxxstring_abis(platforms)
```

S/Skia/build_tarballs.jl
fingolfin: @stensmo wrote elsewhere:
> python3 tools/git-sync-deps is absolutely necessary to complete the build. This script fails for some reason, but still it syncs enough deps to complete the build (except for Android). I hope Google will fix the error in the next version of Skia, so the || true can be removed.


This is still unclear. *What exactly* fails? Looking at the build logs here, all I see is this:
```
[05:19:45]  ---> python3 tools/git-sync-deps
[05:19:45] fatal: remote transport reported error
[05:19:45] fatal: remote transport reported error
[05:19:45] fatal: remote transport reported error
[05:19:45] fatal: remote transport reported error
[05:19:46] fatal: remote transport reported error
[05:20:10] Skipping "bin".
[05:20:10] Skipping "infra/skia-infra".
[05:20:10] Skipping "task_drivers".
[... long list of stuff being synced ...]
[05:20:10] Traceback (most recent call last):
[05:20:10]   File "/workspace/srcdir/skia/tools/git-sync-deps", line 279, in <module>
[05:20:10]     exit(main(sys.argv[1:]))
[05:20:10]   File "/workspace/srcdir/skia/tools/git-sync-deps", line 267, in main
[05:20:10]     git_sync_deps(deps_file_path, argv, shallow, verbose)
[05:20:10]   File "/workspace/srcdir/skia/tools/git-sync-deps", line 237, in git_sync_deps
[05:20:10]     multithread(git_checkout_to_directory, list_of_arg_lists)
[05:20:10]   File "/workspace/srcdir/skia/tools/git-sync-deps", line 254, in multithread
[05:20:10]     raise Exception("Thread failure detected")
[05:20:10] Exception: Thread failure detected
[05:20:10]  ---> true
```

How do you deduce from this that "it syncs enough deps to complete the build (except for Android)" ?

Here is my worry: when this JLL gets rebuild in the future, a different subset of deps is synced. Maybe it still will build, but then perhaps with a different subset of features enabled (as those may depend on which deps are available). That will lead to reaaallly hard to understand issues, and doesn't sound like a stable situation to me.


I think the full list of things it tries to sync is given in https://github.com/google/skia/blob/main/DEPS (or rather [here](https://github.com/google/skia/blob/482de011c920d85fdbe21a81c45852655df6a809/DEPS) for the specific commit you are targeting in this recipe). There are three which reference an android repository.

But that also leads me to the other question I already asked elsewhere: this seems to download & build & link custom copies of e.g. zlib, libpng, libwebp and many others for which we have JLLs. Is that a good idea? Wondering what @giordano thinks about it. If he is OK with this, I'll not complain. But normally I would expect an attempt to try using e.g. `libpng_jll` etc. as dependencies...

That said, I can also see that it might be more useful for you to first get a version of this JLL so you can experiment with writing a wrapper. (Though I would point out that you can already do so by building the JLL locally and testing that...)

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,106 @@
...
+    target_cpu=powerpc64le
+elif [[ "${target}" == i686-* ]]; then
+    target_cpu=x86
+fi
+
+
+ARGS="
+target_cpu=$target_cpu
+cc=clang
+cxx=clang++
```
fingolfin: Oops
```suggestion
cxx=\"clang++\"
```

S/Skia/build_tarballs.jl
stensmo: > @stensmo wrote elsewhere:
> 
> > python3 tools/git-sync-deps is absolutely necessary to complete the build. This script fails for some reason, but still it syncs enough deps to complete the build (except for Android). I hope Google will fix the error in the next version of Skia, so the || true can be removed.
> 
> This is still unclear. _What exactly_ fails? Looking at the build logs here, all I see is this:
> 
> ```
> [05:19:45]  ---> python3 tools/git-sync-deps
> [05:19:45] fatal: remote transport reported error
> [05:19:45] fatal: remote transport reported error
> [05:19:45] fatal: remote transport reported error
> [05:19:45] fatal: remote transport reported error
> [05:19:46] fatal: remote transport reported error
> [05:20:10] Skipping "bin".
> [05:20:10] Skipping "infra/skia-infra".
> [05:20:10] Skipping "task_drivers".
> [... long list of stuff being synced ...]
> [05:20:10] Traceback (most recent call last):
> [05:20:10]   File "/workspace/srcdir/skia/tools/git-sync-deps", line 279, in <module>
> [05:20:10]     exit(main(sys.argv[1:]))
> [05:20:10]   File "/workspace/srcdir/skia/tools/git-sync-deps", line 267, in main
> [05:20:10]     git_sync_deps(deps_file_path, argv, shallow, verbose)
> [05:20:10]   File "/workspace/srcdir/skia/tools/git-sync-deps", line 237, in git_sync_deps
> [05:20:10]     multithread(git_checkout_to_directory, list_of_arg_lists)
> [05:20:10]   File "/workspace/srcdir/skia/tools/git-sync-deps", line 254, in multithread
> [05:20:10]     raise Exception("Thread failure detected")
> [05:20:10] Exception: Thread failure detected
> [05:20:10]  ---> true
> ```
> 
> How do you deduce from this that "it syncs enough deps to complete the build (except for Android)" ?
> 
> Here is my worry: when this JLL gets rebuild in the future, a different subset of deps is synced. Maybe it still will build, but then perhaps with a different subset of features enabled (as those may depend on which deps are available). That will lead to reaaallly hard to understand issues, and doesn't sound like a stable situation to me.
> 
> I think the full list of things it tries to sync is given in https://github.com/google/skia/blob/main/DEPS (or rather [here](https://github.com/google/skia/blob/482de011c920d85fdbe21a81c45852655df6a809/DEPS) for the specific commit you are targeting in this recipe). There are three which reference an android repository.
> 
> But that also leads me to the other question I already asked elsewhere: this seems to download & build & link custom copies of e.g. zlib, libpng, libwebp and many others for which we have JLLs. Is that a good idea? Wondering what @giordano thinks about it. If he is OK with this, I'll not complain. But normally I would expect an attempt to try using e.g. `libpng_jll` etc. as dependencies...
> 
> That said, I can also see that it might be more useful for you to first get a version of this JLL so you can experiment with writing a wrapper. (Though I would point out that you can already do so by building the JLL locally and testing that...)

If the deps fails to sync, Skia will not build. This is because Skia will look for header files etc. which will not exist if the sync fails. The included features are *not* controlled by which sources are avaiable, but through the line: bin/gn gen out/Static --args="..."

stensmo:
Your new (new) suggestion for multi line did not work either. I have tried several more combinations (before your review started), but could not get them to work. Now I have reverted back to a single line. The warning regarding platforms = expand_cxxstring_abis(platforms) is for GCC, but this build uses only Clang.

S/Skia/build_tarballs.jl
stensmo: But if I use the Skia version used and have to add to the C API they will be out of sync anyways. But to get this going I will set it to the Skia version

fingolfin:
I've now pushed a commit to implement the multiline args. Didn't realize that the inner quoting was `gn` syntax and thus had to be preserved in all places using it. (Then again it was obvious after reading the error messages `gn` produced. I couldn't read the error messages before @stensmo because you kept immediately tweaking and reverting, giving no chance to help).

Anyway, even with this: you still have not replied about the questions regarding the dependencies.

Also, it is still unclear to me why this builds a static library and then tries to turn that into a shared library, instead of building a shared library directly -- my guess is that these two questions are directly related, i.e. you might run into problems with the latter due to the former (carrying around duplicate shared libraries for zlib, libpng etc. etc. likely will lead to trouble down the road).

Anyway, I'll leave it to others to further review this

fingolfin:
(I *will* make sure the build works before I leave, please don't touch it until then to give me time to read the errors and fix them)

stensmo:
@fingolfin Many thanks for your contribution. Regarding the last question. The whole reason why it is built as an archive and then made an SO is that the linker discards the functions required for the Julia API. Right now, I have no idea why this is occuring. Thats why you see the --whole-archive flag. I added a test as well to check that (a representative one of the) symbols are indeed included.

stensmo:
Finally, I managed to fix the last issue you had with this build. Now I can build a dynamic library right away, and I have also tested that library, and everything works

stensmo:
@imciner2 Can you please take a look at this new version?

stensmo:
Regarding the dependencies. Packing all dependencies with Skia makes it easier to maintain. Google checkted that the version of the dependency is compatible with the current code. If the dependecies are not included with Skia, one must manually go through the dependencies and check the versions. 

If Julia is incompatible with large binaries, then we have to split it into smaller chunks. 

This is the recommended way to update Skia:
git pull
python3 tools/git-sync-deps
ninja -C out/Static

S/Skia/build_tarballs.jl
stensmo: This is now fixed

S/Skia/build_tarballs.jl
stensmo: The probable culprit is that the Python version you have to use when building Julia is very old. The version in use is 3.9.18. Python 3.9 was released  2020-10-05.

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,106 @@
...
+   GitSource("https://github.com/google/skia.git", "482de011c920d85fdbe21a81c45852655df6a809"),
+   GitSource("https://github.com/stensmo/cskia.git", "9a7c1f23ff12fbc94eed577ea366b0e51fedb119"),
+   DirectorySource("./bundled"),
+]
+
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+filter!(p -> Sys.islinux(p) && libc(p) == "glibc" && call_abi(p) != "eabihf", platforms)
```
giordano: I find the reference to the call ABI a bit obscure, what's more user-facing is the architecture
```suggestion
filter!(p -> Sys.islinux(p) && libc(p) == "glibc" && arch(p) ∉ ("armv6l", "armv7l"), platforms)
```

S/Skia/build_tarballs.jl
giordano: This was not addressed?

stensmo:
I reintroduced a Mac version.

S/Skia/build_tarballs.jl
giordano: @stensmo you keep marking this as resolved while not actually addressing it.  You must do that.

stensmo:
I have now removed the git-sync-deps call, and added more platforms.

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,162 @@
...
+   GitSource("https://chromium.googlesource.com/chromium/deps/libjpeg_turbo.git","e14cbfaa85529d47f9f55b0f104a579c1061f9ad"),
+   GitSource("https://chromium.googlesource.com/external/gitlab.com/wg1/jpeg-xl.git","a205468bc5d3a353fb15dae2398a101dff52f2d3"),
+   GitSource("https://skia.googlesource.com/third_party/libpng.git","ed217e3e601d8e462f7fd1e04bed43ac42212429"),
+   GitSource("https://chromium.googlesource.com/webm/libwebp.git","845d5476a866141ba35ac133f856fa62f0b7445f"),
+   GitSource("https://chromium.googlesource.com/libyuv/libyuv.git","d248929c059ff7629a85333699717d7a677d8d96"),
+   GitSource("https://android.googlesource.com/platform/external/piex.git","bb217acdca1cc0c16b704669dd6f91a1b509c406"),
+   GitSource("https://chromium.googlesource.com/external/github.com/unicode-org/unicodetools","66a3fa9dbdca3b67053a483d130564eabc5fe095"),
+   GitSource("https://skia.googlesource.com/external/github.com/google/wuffs-mirror-release-c.git","e3f919ccfe3ef542cfc983a82146070258fb57f8"),
+   GitSource("https://chromium.googlesource.com/chromium/src/third_party/zlib","646b7f569718921d7d4b5b8e22572ff6c76f2596"),
+
```
giordano: Why?  Besides the fact having this many source is a massive code smell and things should be split into multiple packages, we do have already vast majority of those packages as existing JLLs.

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,158 @@
...
+# Collection of sources required to complete build
+sources = [
+   GitSource("https://github.com/google/skia.git", "482de011c920d85fdbe21a81c45852655df6a809"),
+   GitSource("https://github.com/stensmo/cskia.git", "3438e6efd3a4f27f43457db675ceb33da30c60cf"),
+   DirectorySource("./bundled"),
+   #Missing headers for freetype2
+   GitSource("https://chromium.googlesource.com/chromium/src/third_party/freetype2.git","5d4e649f740c675426fbe4cdaffc53ee2a4cb954"),
+   GitSource("https://chromium.googlesource.com/libyuv/libyuv.git","d248929c059ff7629a85333699717d7a677d8d96"),
+   # These two have some kind of source dependency. 
+   GitSource("https://skia.googlesource.com/external/github.com/google/wuffs-mirror-release-c.git","e3f919ccfe3ef542cfc983a82146070258fb57f8"),
```
giordano, line 17: This is also added as `Dependency`, what it should be?

S/Skia/build_tarballs.jl
```diff
@@ -0,0 +1,158 @@
...
+name = "Skia"
+version = v"0.40.0"
+
+# Collection of sources required to complete build
+sources = [
+   GitSource("https://github.com/google/skia.git", "482de011c920d85fdbe21a81c45852655df6a809"),
+   GitSource("https://github.com/stensmo/cskia.git", "3438e6efd3a4f27f43457db675ceb33da30c60cf"),
+   DirectorySource("./bundled"),
+   #Missing headers for freetype2
+   GitSource("https://chromium.googlesource.com/chromium/src/third_party/freetype2.git","5d4e649f740c675426fbe4cdaffc53ee2a4cb954"),
```
giordano: What header is missing?

stensmo:
The reason for the change from using the git-sync-deps call to using git sources, was that many versions are built at the same time, and Google has implemented throttling (you cannot call git-sync-deps 25 times in a row quickly). 

Could you please specify why this pull request is still blocked?

PR #11325, opened 2025-05-30
[libjulia] bump 1.12 to beta3 and update 1.13-dev

benlorenz:
I guess we need this to work towards fixing GAP and CxxWrap errors on nightly. (And I think bumping 1.12 in the process is also reasonable)

cc: @fingolfin @lgoettgens @barche


lgoettgens:
Just as a note: The julia PR that made rebuilding for nightly necessary will get backported to 1.12. This should not block this PR whatsoever, but it will probably mean that we need another rebuild in the near future

giordano:
Llvm 20 needs zstd

fingolfin:
Unfortunately still fails with the same (?) error

barche:
Adding `-lz` to the `LDFLAGS` seems to fix it, but I'm not sure it's a good idea or just covering up the real cause.

PR #11338, opened 2025-06-01
ImageMagick: Update to 6.9.13-25 & 7.1.1-47

gnadt:


ViralBShah:
Perhaps needs a GCC bump since the arm64 build is segfaulting.

gnadt:
Thanks for the tip! Using (preferred) GCC v7 for the latest ImageMagick seems to work. Seemingly not necessary for the latest ImageMagick6 (Legacy). Unclear to me what sets the GCC compatibility (and not for a lack of trying).

Related, it's also unclear to me why the JLL compat works differently than normal Julia packages' compat. For example, `Dependency("Libtiff_jll"; compat="~4.7.0")` seemed to install `4.7.0`, not `4.7.1` as I would have expected. Note, this part of the changes is the main reason for this PR. Right now ImageMagick/Libtiff are holding back the ability to having the latest `Images` & `Plots` together.

ViralBShah:
`~4.7.0` fixes it to that exact version. 

The reason that was in place was because of some failing tests if a newer `libtiff_jll` was used. I can't seem to find it, but we'll find out if ImageMagick.jl tests pass.

gnadt:
@ViralBShah I noticed that the [ImageMagick_jll](https://github.com/JuliaBinaryWrappers/ImageMagick_jll.jl/commits/main) build numbers didn't increment, and I see now that the last release(s) in the GitHub repo were not registered due to some sort of [test failure(s)](https://github.com/JuliaRegistries/General/pull/115973), though I can't see the logs. I also potentially see what you're talking about with Libtiff_jll on the last release that was registered [here](https://github.com/JuliaRegistries/General/pull/105903#issuecomment-2120589263). Happy to help however I can, but I'm not sure what the next step would be.

ViralBShah:
I think the issue is basically that the version numbers are the same as before, and we can't change the version numbers of the dependencies. So we probably need to bump the version numbers here.

@gnadt We can use the version number scheme as follows: 7.1.10047 and 6.9.13025 and make a note about the scheme in the file. I'll delete the binaries from JuliaBinaryWrappers that did not get registered.

ViralBShah:
@giordano I suspect I deleted ImageMagick_jll 6.9.13 which was not the unregistered one from here, but an older registered version from JuliaBinaryWrappers. Is it possible to recover it? If not what can I do to fix it? I'm really sorry about the unnecessary trouble here.

EDIT: I think the assets I deleted (6.9.13+1) were not registered in the registry.

giordano:
> Is it possible to recover it?

Not as far as I know (maybe contacting github support can help, but in principle a deleted release asset is gone for good).

giordano:
I'm honestly missing what's the problem.

ViralBShah:
> > Is it possible to recover it?
> 
> Not as far as I know (maybe contacting github support can help, but in principle a deleted release asset is gone for good).

I was deleting new assets that I thought were created through this PR. But  I deleted an older asset which I thought was registered. However, I think I just got lucky that the thing I deleted is not in the registry (for whatever reason it might have failed - 6.9.13+1)

So I don't think there's anything to do.

gnadt:
@ViralBShah you _did_ delete new assets, kind of. `upload_to_github_releases()` is using the same tags (`ImageMagick-v6.9.13+1` & `ImageMagick-v7.1.1+2`) when calling [ghr](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/d10ad161d40260a8aa942b81c100e2bacc462ee9/src/AutoBuild.jl#L500), but since it's not set to use the `-delete` or `-replace` [options](https://github.com/tcnksm/ghr?tab=readme-ov-file#options), the assets aren't updated. This step is indeed being hit, though, because new assets for new platforms _do_ appear ([riscv64-linux](https://github.com/JuliaBinaryWrappers/ImageMagick_jll.jl/releases/tag/ImageMagick-v7.1.1%2B2), which you deleted for `6.9.13+1` along with the others).

Unless we manually modify the build number(s) in the `Project.toml` (or some other way I don't know about), it's just going to keep trying to use `ImageMagick-v6.9.13+1` & `ImageMagick-v7.1.1+2`, since it's [based on the last registered version(s)](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/d10ad161d40260a8aa942b81c100e2bacc462ee9/src/AutoBuild.jl#L546).

I'm not sure if this fully solves the issue, but probably should just delete the `ImageMagick-v6.9.13+1` & `ImageMagick-v7.1.1+2` assets _and_ tags (which were not registered: [6](https://github.com/JuliaRegistries/General/pull/115974), [7](https://github.com/JuliaRegistries/General/pull/115973)), and re-run the build if possible.

@giordano what I don't understand is why a registration PR wasn't created, if [`register_jll()` is called before `upload_to_github_releases()`](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/d10ad161d40260a8aa942b81c100e2bacc462ee9/src/AutoBuild.jl#L411-L421), the latter of which was hit.

ViralBShah:
We do need to use a new version number, and I suggest we map 6.9.13-25 to 6.9.13025 so that it will supercede the ones in the registry and allow us to release new upstream releases of 6.9.13 for example.

I am not an expert on the registration issues, but in general my understanding is that for the same release version, you can not change the versions of the dpendencies. If you change dependencies, you have to bump the version number and hence the scheme I suggested.

gnadt:
Yeah, I was just about to edit my comment that I understand what you mean (I think), also clearly not an expert in registration issues.

Hardcode `13025` instead of `upstream_version.patch` [here](https://github.com/JuliaPackaging/Yggdrasil/blob/b54d81fc341086de93ba9bbb9b6f7a119474ae23/I/ImageMagick/ImageMagick%406/build_tarballs.jl#L6), for example.

PR #11343, opened 2025-06-02
[libsingular_julia] Rebuild with libcxxwrap 0.14

lgoettgens:
This is needed to get the libcxxwrap that was build against the new libjula in our dependency tree.

The PR to Singular.jl that bumps the libsingular_julia compat to `~0.48` should also bump the CxxWrap compat to `~0.17`.

cc @fingolfin

### DIFF
--- a/L/libsingular_julia/build_tarballs.jl
+++ b/L/libsingular_julia/build_tarballs.jl
@@ -9,7 +9,7 @@ uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "libsingular_julia"
-version = v"0.47.2"
+version = v"0.48.0"
 
 # Collection of sources required to build libsingular-julia
 sources = [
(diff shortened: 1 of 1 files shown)


L/libsingular_julia/build_tarballs.jl
```diff
@@ -9,7 +9,7 @@ uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "libsingular_julia"
-version = v"0.47.2"
+version = v"0.48.0"
```
fingolfin: Why not
```suggestion
version = v"0.47.3"
```
?

lgoettgens: > This was done the last time when you guys updated the lobcxxwrap minor version.
And I think we want to enable Pkg using this new build together with bumping the CxxWrap.jl compat in Polymake.jl.
I have no idea what happens if it would try to resolve the old CxxWrap.jl 0.16.x with this jll build against a "newer" libcxxwrap.
>
> But you have done this for a longer time, so please just apply the patch of you are sure it doesn't break anything

My comment from libpolymake_julia. Just replace all Polymake by singular.

fingolfin:
Thanks a lot @lgoettgens

PR #11344, opened 2025-06-02
[libpolymake_julia] Switch to libcxxwrap 0.14

lgoettgens:
This is needed to get the libcxxwrap that was build against the new libjula in our dependency tree.

This libcxxwrap version corresponds to CxxWrap 0.17.x.

cc @fingolfin @benlorenz

### DIFF
--- a/L/libpolymake_julia/build_tarballs.jl
+++ b/L/libpolymake_julia/build_tarballs.jl
@@ -16,7 +16,7 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.13.0"), uuidopenssl)
 include("../../L/libjulia/common.jl")
 
 name = "libpolymake_julia"
-version = v"0.14.0"
+version = v"0.13.3"
 
 # reminder: change the above version when changing the supported julia versions
 # julia_versions is now taken from libjulia/common.jl


L/libpolymake_julia/build_tarballs.jl
```diff
@@ -16,7 +16,7 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.13.0"), uuidopenssl)
 include("../../L/libjulia/common.jl")
 
 name = "libpolymake_julia"
-version = v"0.13.2"
+version = v"0.14.0"
```
benlorenz:
```suggestion
version = v"0.13.3"
```
I don't see any need for bumping the minor version, it is just a coincidence that this is at 0.13.x.

lgoettgens: This was done the last time when you guys updated the lobcxxwrap minor version.
And I think we want to enable Pkg using this new build together with bumping the CxxWrap.jl compat in Polymake.jl.
I have no idea what happens if it would try to resolve the old CxxWrap.jl 0.16.x with this jll build against a "newer" libcxxwrap.

But you have done this for a longer time, so please just apply the patch of you are sure it doesn't break anything

PR #11352, opened 2025-06-04
[ROCmDeviceLibs] bump julia compat

sheevy:
LLVM 19 is required.
Julia 1.12 uses LLVM 18
Julia master uses LLVM 20

### DIFF
--- a/R/ROCmDeviceLibs/ROCmDeviceLibs@6.4.1/build_tarballs.jl
+++ b/R/ROCmDeviceLibs/ROCmDeviceLibs@6.4.1/build_tarballs.jl
@@ -37,4 +37,4 @@ build_tarballs(
     ARGS, configure_build(v"6.4.1")...;
     preferred_gcc_version=v"7",
     preferred_llvm_version=v"9",
-    julia_compat="1.12")
+    julia_compat="1.13")


sheevy:
@amontoison see https://github.com/JuliaPackaging/Yggdrasil/pull/11293#issuecomment-2939672127 to why it's needed

giordano:
@amontoison this PR shouldn't have been merged: changing compat bound always ***REQUIRES*** changing version number.

amontoison:
@giordano Sorry if I was mistaken.
I remembered that we had previously recompiled all dependencies with a compat bound on SCOTCH_jll.jl without bumping the version number.

If we publish a new version, rocm 6.4.1 can still be installed on Julia 1.12, which is problematic because it breaks users.
This PR also fixes the compat entry in the General registry to address that.

So we can do that for a dependency, but not when the incompatibility is with the Julia version itself?

- https://github.com/JuliaPackaging/Yggdrasil/pull/6574
- https://github.com/JuliaPackaging/Yggdrasil/pull/6575
- https://github.com/JuliaPackaging/Yggdrasil/pull/6572
- https://github.com/JuliaPackaging/Yggdrasil/pull/6571
- https://github.com/JuliaPackaging/Yggdrasil/pull/6576

giordano:
In the PRs you linked you ***added*** a compat bound not previously existing.  In this PR an _existing_ compat bound was ***changed***.  If there's a problem with a compat bound in the registry you fix the registry, not wreak it.

PR #11357, opened 2025-06-04
ImageMagick: Update versioning

gnadt:


ViralBShah:
@giordano If depending on Zlib_jll, is it necessary to have a compat entry for version 1.2.12? I have seen that in other recipes, and wondering if it is required.

ViralBShah:
I just marked it Ready for Review, and then all the builds died. Kicking off a rebuild.

ViralBShah:
This is something I have not seen before - a conflict in the Compat.toml. @giordano Can you help fix it?

https://github.com/JuliaRegistries/General/pull/132322

giordano:
I can try to have a look tomorrow, but registering two versions of the same package is bound to have conflicts, because the two registrations happen in separate independent PRs.

ViralBShah:
This is not the same version though. We changed the version number. Curious to see what you find.

giordano:
> This is not the same version though. We changed the version number.

I know, in fact I said "registering two versions" 🙂 That's normal, it has happened before, because, as I said, that's just bound to happen by construction: two different PRs are independently touching the same files in different ways.

gnadt:
Hopefully saved @giordano a bit of time, made a suggested change on the PR to (probably) resolve the conflict.

giordano:
#11364

PR #11359, opened 2025-06-04
[normaliz] update to 3.10.5

fingolfin:
One big question is: does this preserve binary compatibility... or do we need to mark it as breaking a


benlorenz:
I tried putting it into an override folder and I am getting a segfault for some Oscar.jl tests.
To prevent this from being merged until I can understand what is going on I changed this to draft.

benlorenz:
The crash I am seeing with the binaries from this PR is very likely caused by some normaliz changes, the backtrace looks like this:
```
#0  0x00007f2c44f11888 in libnormaliz::AutomorphismGroup<__gmp_expr<__mpz_struct [1], __mpz_struct [1]> >::~AutomorphismGroup() ()
   from /home/lorenz/.julia/scratchspaces/d720cf60-89b5-51f5-aff5-213f193123e7/polymake_11835203091876189151_1.11_depstree_v2/lib/polymake/lib/polytope.so
#1  0x00007f2c44f14967 in libnormaliz::Cone<__gmp_expr<__mpz_struct [1], __mpz_struct [1]> >::Cone<__gmp_expr<__mpz_struct [1], __mpz_struct [1]> >(std::map<libnormaliz::Type::InputType, std::vector<std::vector<__gmp_expr<__mpz_struct [1], __mpz_struct [1]>, std::allocator<__gmp_expr<__mpz_struct [1], __mpz_struct [1]> > >, std::allocator<std::vector<__gmp_expr<__mpz_struct [1], __mpz_struct [1]>, std::allocator<__gmp_expr<__mpz_struct [1], __mpz_struct [1]> > > > >, std::less<libnormaliz::Type::InputType>, std::allocator<std::pair<libnormaliz::Type::InputType const, std::vector<std::vector<__gmp_expr<__mpz_struct [1], __mpz_struct [1]>, std::allocator<__gmp_expr<__mpz_struct [1], __mpz_struct [1]> > >, std::allocator<std::vector<__gmp_expr<__mpz_struct [1], __mpz_struct [1]>, std::allocator<__gmp_expr<__mpz_struct [1], __mpz_struct [1]> > > > > > > > const&) ()
...
```
It reproduces both on macos (aarch64) and linux (x86_64).

Looking at the diff between 3.10.2 and 3.10.5 there are for example changes in the class members of C++ classes, also added entries in the middle of some enums. Both inside headers, so this looks very much _not_ ABI-compatible.

Since the only two packages wich depend on this have `~300.1000.200` (`polymake` and `GAP_pkg_normalizinterface`) I would suggest to just bump the minor offset?

N/normaliz/build_tarballs.jl
```diff
@@ -22,15 +22,13 @@ import Pkg.Types: VersionSpec
 # to all components.
 
 name = "normaliz"
-version = v"300.1000.201"
-upstream_version = v"3.10.2"
+version = v"300.1000.500"
```
benlorenz:
```suggestion
version = v"300.1001.500"
```

benlorenz: We could also add the offset to the major version:
```suggestion
version = v"301.1000.500"
```

fingolfin:
Thanks for the checks, @benlorenz I'd be fine with either version change. Do you have a preference?

benlorenz:
> Thanks for the checks, @benlorenz I'd be fine with either version change. Do you have a preference?

I prefer smaller increments so I just bumped the minor version offset.

PR #11362, opened 2025-06-05
[OpenPolicyAgent] Update to 1.5.1

lamdor:
Release notes: https://github.com/open-policy-agent/opa/releases/tag/v1.5.1


imciner2:
It appears that OpenPolicyAgent grew a dependency on `goversioninfo` (https://github.com/josephspurrier/goversioninfo) sometime after we packaged it. So in order to update this, you will need to get that program installed as a build dependency.

lamdor:
@imciner2 I've added that in.

imciner2:
Thanks, looks good now.

PR #11366, opened 2025-06-05
New Recipe: libbacnet_stack v1.4.1

divbyzerofordummies:
Recipe to build shared libraries from the [BACnet Protocol Stack library](https://github.com/bacnet-stack/bacnet-stack).


L/libbacnet_stack/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/bacnet-stack/bacnet-stack.git", "c111154993cce5dc0e3edd10c602bce51e2d2d61")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cmake -B build -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON bacnet-stack
```
imciner2:
```suggestion
cmake -B build \
    -DCMAKE_INSTALL_PREFIX=$prefix \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=ON \
    bacnet-stack
```
Cleaner for future diffs.

L/libbacnet_stack/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
...
+cmake --build build --parallel 4
```
imciner2:
```suggestion
cmake --build build --parallel ${nproc}
```
We tell the build how many processes to use through this variable.

L/libbacnet_stack/build_tarballs.jl
```diff
@@ -0,0 +1,49 @@
...
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("riscv64", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
```
imciner2:
```suggestion
platforms = supported_platforms(; exclude=!Sys.islinux)
```
You just want to build for all Linux platforms, right? If so, this is the nicer way to say that.

imciner2 (changes requested):
Here are a few cleanups to the script.

The only error right now is that no license is being installed, however this package appears to distribute all its license files (https://github.com/bacnet-stack/bacnet-stack/tree/master/license).

I think we need to copy that entire directory into the license folder (`${prefix}/share/licenses/libbacnet_stack`).

L/libbacnet_stack/build_tarballs.jl
divbyzerofordummies: Completely agree. Fixed in new commit.

L/libbacnet_stack/build_tarballs.jl
divbyzerofordummies: Thanks for the tip. I have now succeeded with the Windows build process, so I now do not exclude that either.

L/libbacnet_stack/build_tarballs.jl
divbyzerofordummies: Perfect. Fixed it.

divbyzerofordummies:
> Here are a few cleanups to the script.
> 
> The only error right now is that no license is being installed, however this package appears to distribute all its license files (https://github.com/bacnet-stack/bacnet-stack/tree/master/license).
> 
> I think we need to copy that entire directory into the license folder (`${prefix}/share/licenses/libbacnet_stack`).

I just copied the content of the `license` directory to the `${prefix}/share/licenses/bacnet_stack` directory. The build process no longer complains, so I guess that worked.

PR #11367, opened 2025-06-05
[ImageMagick@6] Revert Libtiff_jll to ~4.5.1

gnadt:
Latest `Libtiff_jll` can still not be used with ImageMagick.jl, CI error message(s):

"The value of field_readcount and field_writecount must be greater than or equal to -3 and not zero.. `TIFFMergeFieldInfo' @ error/tiff.c/TIFFErrors/574"

This stems from [this libtiff issue](https://gitlab.com/libtiff/libtiff/-/issues/532#note_2134433038), which has been resolved [here](https://gitlab.com/libtiff/libtiff/-/merge_requests/668) and should be fixed in the next libtiff release.

Making separate PRs for `ImageMagick@6` & `ImageMagick@7` to avoid registration conflicts.

### DIFF
--- a/I/ImageMagick/ImageMagick@6/build_tarballs.jl
+++ b/I/ImageMagick/ImageMagick@6/build_tarballs.jl
@@ -2,11 +2,11 @@
 # `julia build_tarballs.jl --help` to see a usage message.
 using BinaryBuilder
 name = "ImageMagick"
-upstream_version = v"6.9.13-25"
+upstream_version = v"6.9.13-25" # TODO: re-sync with upstream next release (+ 1 below was to update compat)
 version = VersionNumber(
     upstream_version.major,
     upstream_version.minor,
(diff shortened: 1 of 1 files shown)


giordano (changes requested):
Changing compat bounds requires changing the version number.

gnadt:
Fair enough, just increment by 1 with a comment saying that they are out of sync? (I saw that elsewhere)

giordano:
See the package which use the offset, and yes, document it with a comment.

gnadt:
Not sure if that is a question, but [this](https://github.com/JuliaPackaging/Yggdrasil/blob/442d1d93800afdb9ac3b3c6e994f9c83f338fc8a/L/Libtiff/build_tarballs.jl#L7) is what I was referring to.

gnadt:
This might just need to be re-run. Looks like it got hung up on 1 platform (unclear why), then timed out.

giordano (approved):

PR #11369, opened 2025-06-05
[GoogleCodeSearch] update to force Go toolchain update

lamdor:
The go version and stdlib this was compiled with 5 years ago has several reported vulnerabilties. Since JuliaHub uses this jll, we get pinged on this in our docker image scans all the time.

I'm not sure of the best way to force a rebuild, but it doesn't build as it was.  I have to set julia_compat to 1.6 to have it build for the experimental platforms.

Is this the best way to do force a rebuild? Will this cause a released jll as `1.2.0+1` when it's merged?

### DIFF
--- a/G/GoogleCodeSearch/build_tarballs.jl
+++ b/G/GoogleCodeSearch/build_tarballs.jl
@@ -3,11 +3,14 @@
 using BinaryBuilder, Pkg
 
 name = "GoogleCodeSearch"
-version = v"1.2.0"
+# This version is different than the source version as we need to rebuild this with an updated Go version
+# and update compat bounds
+# The original version is v1.2.0 which hasn't been updated since 2020
+version = v"1.2.1"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/google/codesearch.git", "8ba29bd255b740aee4eb4e4ddb5d7ec0b4d9f23e")
+    GitSource("https://github.com/google/codesearch.git", "8ba29bd255b740aee4eb4e4ddb5d7ec0b4d9f23e") # v1.2.0
 ]
 
 # Bash recipe for building across all platforms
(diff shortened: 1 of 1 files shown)


G/GoogleCodeSearch/build_tarballs.jl
```diff
@@ -38,4 +38,4 @@ dependencies = Dependency[
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; compilers = [:go, :c])
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", compilers = [:go, :c])
```
tanmaykm:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", compilers = [:go, :c], preferred_go_version = v"1.23")
```

Maybe specifying the go version will be better so that it is explicit... and subsequent updates can change that.

lamdor: Yeah, that sounds good to me. Added.

imciner2:
Because the last jll is so old, this will need a version bump because the compat bounds for the dependencies will have changed, and also changing the Julia compat requires that.

lamdor:
@imciner2 But the last release of the source was 1.2.0: https://github.com/google/codesearch/tags. Won't binary builder release it as 1.2.0+1 since 1.2.0+0 of the jll exists? And then the compat bounds can change? Or that can't happen for `+0` build changes in Julia?

giordano:
Unfortunately the registry can't handle differences of compat bounds between versions which only differ for the build number (what comes after the +). This means that the version number _must_ be changed, even if this means diverging from the upstream version number, as inconvenient as it is.

lamdor:
@giordano @imciner2 Sounds good. I've bumped the version to 1.2.1 and added a note. It doesn't appear that if/when GoogleCodeSearch actually does a release, they will use the 1.2.1 either. They have a 1.3.0-rc1, but that was a year ago. Can you have a look and approve please? Thanks.

PR #11370, opened 2025-06-05
[HiGHS] update to v1.11.0

odow:
cc @galabovaa

### DIFF
--- a/H/HiGHS/highs_common.jl
+++ b/H/HiGHS/highs_common.jl
@@ -4,12 +4,12 @@ using BinaryBuilder, Pkg
 
 name = "HiGHS"
 
-version = v"1.10.0"
+version = v"1.11.0"
 
 sources = [
     GitSource(
         "https://github.com/ERGO-Code/HiGHS.git",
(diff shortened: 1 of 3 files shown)


odow:
DNM: @galabovaa is going to update the release

H/HiGHS/highs_common.jl
```diff
@@ -4,12 +4,12 @@ using BinaryBuilder, Pkg
...
 name = "HiGHS"
 
-version = v"1.10.0"
+version = v"1.11.0"
 
 sources = [
     GitSource(
         "https://github.com/ERGO-Code/HiGHS.git",
-        "fd8665394edfd096c4f847c4a6fbc187364ef474",
+        "6eca8325a43b06942107e3fa88a826051c679912",
```
odow:
```suggestion
        "364c83a51e44ba6c27def9c8fc1a49b1daf5ad5c",
```

imciner2:
The last update to the `highs_common.jl` file didn't actually cause a rebuild with the new commit, so the 1.11.0 that is registered is actually still labelled 1.10 and not the upstream 1.11 release.

PR #11379, opened 2025-06-08
MAGEMin v1.8.0

NicolasRiel:
- Added function to compute the properties of metastable assemblages
- Added feasibility check when using initial guess
- Added vol% to dataframe -> csv export function


NicolasRiel:
@giordano Thank you!

PR #11387, opened 2025-06-09
[eudev] Update to 3.2.14

JanisErdmanis:
This is one of many packages whose JLL depends explicitly on `Pkg`. To remove it, we need to bump the version and rebuild the JLL.

### DIFF
--- a/E/eudev/build_tarballs.jl
+++ b/E/eudev/build_tarballs.jl
@@ -19,7 +19,6 @@ apk add libxslt-dev docbook-xsl
 
 # Only apply the patch for musl targets
 if [[ "${target}" == *"musl"* ]]; then
-    echo "Applying musl-specific thread_local fix"
     atomic_patch -p0 ../musl.patch
 fi
 


E/eudev/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "eudev"
-version = v"3.2.9"
+version = v"3.2.10" # upstream is 3.2.9, but we needed a version bump
```
giordano: There are new releases at https://github.com/eudev-project/eudev.  Also, I'm not sure the runtime dependency on gperf is right.

E/eudev/build_tarballs.jl
```diff
@@ -3,17 +3,25 @@
...
 sources = [
-    ArchiveSource("https://dev.gentoo.org/~blueness/eudev/eudev-3.2.9.tar.gz", "89618619084a19e1451d373c43f141b469c9fd09767973d73dd268b92074d4fc")
+    GitSource("https://github.com/eudev-project/eudev", "9e7c4e744b9e7813af9acee64b5e8549ea1fbaa3")
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/eudev*
-apk add gperf
+apk add gperf autoconf automake libtool libxslt-dev docbook-xsl
```
giordano: Half those tools are already present

E/eudev/build_tarballs.jl
```diff
@@ -3,17 +3,25 @@
...
+./autogen.sh
+
+# Only apply the patch for musl targets
+if [[ "${target}" == *"musl"* ]]; then
+    echo "Applying musl-specific thread_local fix"
+    find . -name "*.c" -exec grep -l "thread_local" {} \; | xargs -r sed -i 's/thread_local/__thread/g'
+fi
```
giordano: Can we get a real patch instead of some sed magics?

giordano: And please don't use [gen AI tools to generate a made up patch](https://github.com/JuliaLang/julia/pull/58628#discussion_r2125092176)

E/eudev/build_tarballs.jl
JanisErdmanis: I did add them one by one to get the compilation running. I will try to remove `autoconf`, `automake` and `libtool` in the next run.

E/eudev/build_tarballs.jl
JanisErdmanis: I prefer sed magics because they are parsable by AI and can be disabled individually afterwards.

E/eudev/build_tarballs.jl
JanisErdmanis: I removed gperf from `apk`. We are now getting strange errors for some platforms.

JanisErdmanis: Fixed the issue. `gperf` needed to be added as `HostBuildDependency`

E/eudev/build_tarballs.jl
```diff
@@ -3,27 +3,37 @@
...
 script = raw"""
 cd $WORKSPACE/srcdir/eudev*
-apk add gperf
+apk add libxslt-dev docbook-xsl
+./autogen.sh
+
+# Only apply the patch for musl targets
+if [[ "${target}" == *"musl"* ]]; then
+    echo "Applying musl-specific thread_local fix"
+    # find . -name "*.c" -exec grep -l "thread_local" {} \; | xargs -r sed -i 's/thread_local/__thread/g'
```
giordano:
```suggestion
```

E/eudev/build_tarballs.jl
```diff
@@ -3,27 +3,37 @@
...
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/eudev*
-apk add gperf
+apk add libxslt-dev docbook-xsl
+./autogen.sh
+
+# Only apply the patch for musl targets
+if [[ "${target}" == *"musl"* ]]; then
+    echo "Applying musl-specific thread_local fix"
```
giordano:
```suggestion
```

PR #11390, opened 2025-06-09
[xcb-util-wm] update to 0.4.2

JanisErdmanis:


X/Xorg_xcb_util_wm/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder
 
 name = "Xorg_xcb_util_wm"
-version = v"0.4.1"
+version = v"0.4.2"
```
giordano, line 6: Need to update the source

giordano:
This also depends on #11394

PR #11395, opened 2025-06-09
[mtdev] Update to 1.1.7

JanisErdmanis:


M/mtdev/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "mtdev"
-version = v"1.1.6"
+version = v"1.1.7"
```
giordano, line 6: Need to update the source

PR #11396, opened 2025-06-09
[Wayland_protocols] Update to 1.44.0

JanisErdmanis:


JanisErdmanis:
I am (or more precisely Claude is) stuck on this.

W/Wayland_protocols/build_tarballs.jl
```diff
@@ -3,19 +3,31 @@
...
 script = raw"""
 cd $WORKSPACE/srcdir/wayland-protocols*/
+
+# Find wayland-scanner and create symlink where Meson expects it
+WAYLAND_SCANNER=$(which wayland-scanner)
+if [ -n "$WAYLAND_SCANNER" ]; then
+    echo "Found wayland-scanner at: $WAYLAND_SCANNER"
+    
+    # Create the exact directory structure that Meson expects
+    EXPECTED_PATH="/workspace/destdir/workspace/x86_64-linux-musl-cxx11/destdir/lib/pkgconfig/../../bin"
```
giordano: That's simply host_bindir or am I missing something?

JanisErdmanis: The wayland-scanner was introduced in https://gitlab.freedesktop.org/wayland/wayland-protocols/-/commit/c9972235833b3586c05a6e6c4145aef6df3f4281 

Meson could not find it for us. The only thing that worked was adding a symlink in a location it is looking for it.

giordano: What I'm trying to say is that wayland_scanner is provided by Wayland_jll, which is a host build dependency of this package, and is such it should be present in host_bindir. So I don't quite understand what's going on.

Also, if you really want to to do this (which you still need to explain me why), please don't hardcoded absolute paths like that, they're fragile (just like magic sed commands), we have loads of environment variables to make things more flexible: https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables

JanisErdmanis: I found the issue to be with `PKG_CONFIG_SYSROOT_DIR`. Clearing it resolves paths normally without the unwanted prefix "/workspace/destdir".

W/Wayland_protocols/build_tarballs.jl
```diff
@@ -3,19 +3,24 @@
...
     ArchiveSource("https://gitlab.freedesktop.org/wayland/wayland-protocols/-/releases/$(version.major).$(version.minor)/downloads/wayland-protocols-$(version.major).$(version.minor).tar.xz",
-                  "71fd4de05e79f9a1ca559fac30c1f8365fa10346422f9fe795f74d77b9ef7e92"),
+                  "3df1107ecf8bfd6ee878aeca5d3b7afd81248a48031e14caf6ae01f14eebb50e"),
 ]
 
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/wayland-protocols*/
+
+# Clear PKG_CONFIG_SYSROOT_DIR to fix wayland-scanner path resolution
```
giordano, line 18: Ok, this makes a lot more sense to me now: by default [that variable points to the target prefix](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/45acaa5c78a091ef75e5daa5de7a8676600509a5/src/Runner.jl#L1317), but in this case you want to find an executable in the host prefix.

PR #11409, opened 2025-06-10
New Recipe: Mongoose v7.18.0

AbrJA:


M/Mongoose/build_tarballs.jl
```diff
@@ -0,0 +1,70 @@
...
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; ),
+    Platform("x86_64", "freebsd"; ),
+    Platform("aarch64", "freebsd"; )
+]
```
giordano:
```suggestion
platforms = supported_platforms(; exclude=Sys.iswindows)
```
Also, what's the issue on Windows?

M/Mongoose/build_tarballs.jl
```diff
@@ -0,0 +1,70 @@
...
+cd $WORKSPACE/srcdir
+cat > CMakeLists.txt <<EOF
+cmake_minimum_required(VERSION 3.10)
+project(mongoose C)
+
+add_library(mongoose SHARED mongoose/mongoose.c)
+install(TARGETS mongoose
+        LIBRARY DESTINATION lib
+        ARCHIVE DESTINATION lib
+        RUNTIME DESTINATION bin)
```
imciner2: These aren't going to be right on all platforms. You should instead use the provided variables `$libdir` and `$bindir` to get the appropriate directories for the platform.

M/Mongoose/build_tarballs.jl
```diff
@@ -0,0 +1,70 @@
...
+EOF
+
+cmake -B build  \
+    -DCMAKE_INSTALL_PREFIX=${prefix}  \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}  \
+    -DCMAKE_BUILD_TYPE=Release  \
+    -S .
+cmake --build build --parallel ${nproc}
+cmake --install build
+"""
```
imciner2, line 40:
```suggestion
install_license mongoose/LICENSE
"""
```
IMO, it is more consistent with the other recipes in Ygg to have the license install at the very end and not part of the build system.

M/Mongoose/build_tarballs.jl
giordano: To stay in CMake, there's the module [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) which would do the right thing.

AbrJA: Using this it works
```
install(TARGETS mongoose
        LIBRARY DESTINATION $libdir
        ARCHIVE DESTINATION $libdir
        RUNTIME DESTINATION $bindir)
```
But I don't know if using Cmake is better
```
install(TARGETS mongoose
        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```
Let me know and I can change it, thanks!

imciner2: If you want to use `${CMAKE_INSTALL_LIBDIR}`, then you need to also have `include(GNUInstallDirs)` in the CMake file in order for the paths to be correct. Although if you do that, there shouldn't even be a need to set the paths in the `install()` command, and it can just be `install(TARGETS mongoose)`.

PR #11412, opened 2025-06-11
[GAP_pkg_*] Add gap_platforms function

lgoettgens:
List of changes:
- split `setup_gap_package` into two functions, as it did two different things
- use these two functions in more places (where applicable)
- generate platforms for `juliainterface` from `libjulia_platforms` instead of expanding the julia version of `setup_gap_package` (and remove some more platforms), so that the platform list for `GAP_pkg_juliainterface_jll` is equal to the one for `GAP_jll`
- prepare rebuilding all `GAP_pkg_*_jll`s for the newly enabled platforms of `GAP_jll`
- adapt some comments that are clearly wrong and probably just an artifact from the past

The bullet-points are not in 1-to-1-correspondence with commits, but it should be rather easy to see which commits does what.

This PR does not change anything about any of the packages involved, it just re-organizes some code; thus I think we can skip building everything (and should to save some ressources).
The two added `TODO` comments would change the lists of platforms of some packages, so that is something I would like to approach in a future PR (that then actually builds things).

cc @fingolfin

[skip build] [skip ci]


G/GAP_pkg/GAP_pkg_kbmag/build_tarballs.jl
```diff
@@ -39,9 +38,7 @@ rm $prefix/lib/gap/sysinfo.gap
 
 name = gap_pkg_name(name)
 
-platforms = supported_platforms()
-filter!(p -> nbits(p) == 64, platforms) # we only care about 64bit builds
-filter!(!Sys.iswindows, platforms)      # Windows is not supported
+platforms = gap_platforms()
```
fingolfin, line 41: Like `nq` this package only produces an executable and does not need GAP for this at all so I am not sure it is beneficial to use `get_platforms` here?

G/GAP/common.jl
```diff
@@ -0,0 +1,29 @@
...
+
+        # we only care about 64bit builds
+        filter!(p -> nbits(p) == 64, platforms)
+
+        # Windows is not supported
+        filter!(!Sys.iswindows, platforms)
+
+        return platforms
+    else
+        # TODO: create the platforms from the above list coming from libjulia
```
fingolfin, line 15: I guess you could use `julia_supported_platforms(v"2")` or so now that PR #11444 is in...

G/GAP/common.jl
```diff
@@ -0,0 +1,29 @@
...
+        # by removing the julia_version tag
+        
+        platforms = supported_platforms()
+        filter!(p -> nbits(p) == 64, platforms) # we only care about 64bit builds
+        filter!(!Sys.iswindows, platforms)      # Windows is not supported
```
fingolfin, line 20: How about moving this common code out to the end of the function, i.e. after the `if/else/end` have
```
    filter!(p -> nbits(p) == 64, platforms) # we only care about 64bit builds
    filter!(!Sys.iswindows, platforms)      # Windows is not supported
    return platforms
```

G/GAP_pkg/GAP_pkg_kbmag/build_tarballs.jl
fingolfin: That said I see no real drawback, and it is shorter, so *shrug*

fingolfin:
@lgoettgens let me know if you'd like to address any of the things I mentioned, or if I should just merge this as-is (we can always tweak it later). 

Thanks for working on this!

G/GAP_pkg/GAP_pkg_kbmag/build_tarballs.jl
lgoettgens: Iirc, my motivation for this was to remove some code duplication (even if it is only 3 lines). And I think it doesn't make much sense to build GAP packages for platforms that are not supported by GAP_jll.
But no strong preference either way on this.

G/GAP/common.jl
lgoettgens: Indeed, thanks for the pointer. But as already mentioned, I would like to do this in a future PR that then also rebuilds GAP_pkg_juliainterface_jll and thus verifies that everything indeed works correctly.

G/GAP/common.jl
lgoettgens: Good point, I'll do that when I redo parts of this function. cf. https://github.com/JuliaPackaging/Yggdrasil/pull/11412#discussion_r2149401738

lgoettgens:
> @lgoettgens let me know if you'd like to address any of the things I mentioned, or if I should just merge this as-is (we can always tweak it later).

Please merge this as-is, Thanks! The purpose of this PR is mostly to do all the little changes to all recipes and then skip all builds. My next PR will further adapt the `common.jl` (that does not force rebuilds of everything) and then touch a small subset of recipes to verify everything still works (and to rebuild for extended platforms).

PR #11415, opened 2025-06-12
[Geant4] Added explicit dependency to Zlib_jll

peremato:


G/Geant4/build_tarballs.jl
```diff
@@ -110,6 +110,7 @@ dependencies = [
     Dependency("Xorg_libXmu_jll"),
     Dependency("Libglvnd_jll"),
     Dependency("Xerces_jll"),
+    Dependency("Zlib_jll"; platforms=filter(Sys.isapple, platforms)),
```
giordano:
```suggestion
    Dependency("Zlib_jll"; compat="1.2.12", platforms=filter(Sys.isapple, platforms)),
```

PR #11416, opened 2025-06-12
[FLINT] Upgrade to version 3.3.0

lgoettgens:
cf. https://github.com/flintlib/flint/releases/tag/v3.3.0

As mentioned in the release description, there have been some struct layout changes in matrix types (from list of row ptrs to strides). What is the opinion of people using FLINT via julia: Is this fine for a minor version bump of the jll (as it is right now) or should I bump the major offset?

cc @benlorenz @eschnett @fingolfin @Joel-Dahne @thofma @albinahlback

### DIFF
--- a/F/FLINT/build_tarballs.jl
+++ b/F/FLINT/build_tarballs.jl
@@ -26,7 +26,7 @@ using BinaryBuilder, Pkg
 # and possibly other packages.
 name = "FLINT"
 upstream_version = v"3.3.0"
-version_offset = v"0.0.0"
+version_offset = v"1.0.0"
 version = VersionNumber(upstream_version.major * 100 + version_offset.major,
                         upstream_version.minor * 100 + version_offset.minor,
                         upstream_version.patch * 100 + version_offset.patch)


thofma:
Since changes of layout have the potential to be disruptive for our wrappers, I would prefer a major version bump.

lgoettgens:
> Since changes of layout have the potential to be disruptive for our wrappers, I would prefer a major version bump.

Just to mention it here:
- all Yggdrasil recipes using a 3.x FLINT have a `~300.200.0` specifier
- Arblib.jl and ArbNumerics.jl also have a `~300.200.0` specifier
- Nemo.jl has a `~300.200.0` specifier since https://github.com/Nemocas/Nemo.jl/pull/2052, and `300.200.0` before

So the only difference between a minor and a major bump here is for Nemo.jl v0.49.1 and before, but only if used on its own and not as part of any environment that contains another dependent of FLINT_jll (and thus no changes for Oscar).

thofma:
No strong objection. I happily follow the majority.

Joel-Dahne:
A major version bump seems reasonable to me. As you mention most (all?) packages depending on FLINT_jll specify also the minor version, which means that it might not matter that much in practice. But I think a major version bump is a good signal that there have been important changes.

fingolfin:
I concur. This is a breaking change, semver suggest increasing the major version

lgoettgens:
> I concur. This is a breaking change, semver suggest increasing the major version

I disagree that this is API-breaking. Of course it is ABI-breaking, and it is also breaking for Nemo.jl but that is only because we rely on internals that are not guaranteed to be stable.

Nonetheless, I'll bump the version to get this merged

F/FLINT/build_tarballs.jl
```diff
@@ -25,28 +25,22 @@ using BinaryBuilder, Pkg
 # coordinated with corresponding changes to Singular_jll.jl, Nemo.jl and polymake_jll.jl
 # and possibly other packages.
 name = "FLINT"
-upstream_version = v"3.2.2"
-version_offset = v"0.0.1"
+upstream_version = v"3.3.0"
+version_offset = v"0.0.0"
```
lgoettgens:
```suggestion
version_offset = v"1.0.0"
```

lgoettgens:
Thanks for all of your messages. The major part of the version is now bumped.

This is good to go from my POV, just needs someone to merge

PR #11424, opened 2025-06-12
[LAPACK] Add the missing suffix _64 on new symbols

amontoison:
I don't understand the issue on `riscv64` :thinking:

### DIFF
--- a/L/LAPACK/LAPACK/build_tarballs.jl
+++ b/L/LAPACK/LAPACK/build_tarballs.jl
@@ -16,4 +16,4 @@ filter!(p -> !(arch(p) == "aarch64" && Sys.islinux(p) && libgfortran_version(p) 
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
                clang_use_lld=false, julia_compat="1.9", preferred_gcc_version=v"6")
 
-# Build Trigger: 9
+# Build Trigger: 10
--- a/L/LAPACK/LAPACK32/build_tarballs.jl
+++ b/L/LAPACK/LAPACK32/build_tarballs.jl
@@ -12,4 +12,4 @@ products = [
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
                clang_use_lld=false, julia_compat="1.9", preferred_gcc_version=v"6")
 
-# Build Trigger: 6
+# Build Trigger: 7
--- a/L/LAPACK/common.jl
+++ b/L/LAPACK/common.jl
@@ -382,9 +382,6 @@ end
 # platforms are passed in on the command line
 platforms = expand_gfortran_versions(supported_platforms())
 
-# Require LBT ≥ 5.12.0
-filter!(p -> arch(p) != "riscv64", platforms)
-
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae")),


imciner2:
Appears to be an LBT problem:

```
Warning: Dependency libblastrampoline_jll does not have a mapping for artifact libblastrampoline for platform riscv64-linux-gnu-libgfortran5-cxx11
```

which leads to

```
-- Looking for Fortran dgemm
-- Looking for Fortran dgemm - not found
ERROR--> BLAS supplied by user is not WORKING, CANNOT USE -L/workspace/destdir/lib -lblastrampoline.
ERROR-->     Will use REFERENCE BLAS (by default)
ERROR-->     Or Correct your BLAS_LIBRARIES entry
ERROR-->     Or Consider checking USE_OPTIMIZED_BLAS
-- Using supplied NETLIB BLAS implementation
```

giordano:
Either libblastrampoline_jll 5.40 or 5.12.0+ have a build for riscv64.  Here you're using 5.11.2

amontoison:
> Either libblastrampoline_jll 5.40 or 5.12.0+ have a build for riscv64. Here you're using 5.11.2

@giordano How is it possible that the previous build of LAPACK 3.12.1 passed on riscv64 in that case?
I found the tarballs for riscv64 in the repo of `LAPACK_jll.jl`.

giordano:
>  I found the tarballs for riscv64 in the repo of LAPACK_jll.jl.

Providing me links you already have would simplify my work.

amontoison:
https://github.com/JuliaBinaryWrappers/LAPACK_jll.jl/releases/tag/LAPACK-v3.12.1%2B0

Sorry answered from phone.

imciner2:
The last LAPACK build used 5.4:
```
+ libblastrampoline_jll v5.4.0+2
```

imciner2:
Or at least the first build use 5.4, but then it was updated to a 5.11 compat later. Maybe there was an issue with stale builds that led to keeping that 5.4 version instead of rebuilding with 5.11?

giordano:
Yes, it was built with 5.4, but the specification of the compat is in the `common.jl` file which is in an upper directory and hence not part of the [cache index](https://github.com/JuliaPackaging/Yggdrasil/blob/21ac74843913ae5e7eb15ec6f07051cfaf820cbe/CONTRIBUTING.md#understanding-build-cache-on-yggdrasil), so following changes in #10820 to `common.jl` to make other platforms work only actually rebuilt the failing platforms, the successful ones were hitting the cache and weren't rebuilt.

amontoison:
I recompiled `LAPACK_jll.jl` and `LAPACK32_jll.jl` without `riscv64` and kept the same compat entry on LBT (5.11.2).

ViralBShah:
Should we just bump the version dep to 5.12 so we can grab riscv64?

amontoison:
If we want to support this platform, it is probably easier to compile LBT 5.11.2 on riscv64.
We will not bump the compat entry like this.

ViralBShah:
If straightforward, it would be nice to add LBT for 5.11.2 for riscv64 (support was added for it in 5.11.1). That will just keep our build recipes cleaner for all architectures.

Also, ok to merge this and clean the riscv64 later - I don't have a strong preference on the order or timing.

ViralBShah:
Created #11480 to see if that works.

amontoison:
The compilation failed for LAPACK32, I don't understand why it works for LAPACK in that case.

giordano:
https://github.com/JuliaPackaging/Yggdrasil/pull/11479#issuecomment-2984115044

PR #11432, opened 2025-06-13
new version 0.3.3

awietek:


awietek:
Hi there, I am trying to compile this to work for the upcoming Julia 1.12 version, but it seems to fail (only for this version) due to the fact that CMake can't find julia. Is this supposed to work even though 1.12 is not officially released? Or is something special introduced in the 1.12 version? I'd appreciate any help.

giordano:
I don't know, maybe @benlorenz @fingolfin can give a clue

X/XDiag/build_tarballs.jl
```diff
@@ -89,10 +89,10 @@ dependencies = [
     BuildDependency(PackageSpec(;name="libjulia_jll", version=v"1.10.7")),
```
benlorenz:
```suggestion
    BuildDependency(PackageSpec(;name="libjulia_jll", version=v"1.10.17")),
```
Please try a more recent `libjulia_jll` build.
(Or remove the version entry to always use the latest build)

awietek:
@benlorenz Thank you, that worked, you're the best!

awietek:
Also another question, maybe @giordano . I had some troubles with OpenMP on apple with LLVM in the past, which I in the past"solved" by linking to the gnu OpenMP. However, then it never ran multithreaded. There was also this PR: https://github.com/JuliaPackaging/Yggdrasil/pull/9423 
where  a LLVM version was set to 13, that seemed to fix some problems there. I am trying to do the same here now (without knowing whether it'll fix my problem). Have you by any chance had any new insights what may be causing such issues? I would try out the current solution of this PR, but I was wondering whether some understanding of such problems has surfaced.

X/XDiag/build_tarballs.jl
```diff
@@ -75,8 +75,8 @@ filter!(p -> (
     (os(p) == "macos" && arch(p) == "aarch64") ||
     (os(p) == "windows" && arch(p) == "x86_64")) &&
     p.tags["julia_version"] !="1.6.3" &&
-    p.tags["julia_version"] !="1.12.0" &&
-    p.tags["julia_version"] !="1.13.0", platforms)
+    p.tags["julia_version"] !="1.13.0" &&
+    p.tags["julia_version"] !="1.14.0", platforms)
```
fingolfin: Why do you filter out some Julia versions here? (Not that this will do anything right now anyway)?

fingolfin: (except for filtering out 1.6.3, that will work, but not 1.13.0 and 1.14.0 which are not currently versions in `julia_versions`

X/XDiag/build_tarballs.jl
```diff
@@ -86,13 +86,13 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    BuildDependency(PackageSpec(;name="libjulia_jll", version=v"1.10.7")),
+    BuildDependency(PackageSpec(;name="libjulia_jll", version=v"1.10.17")),
     Dependency(PackageSpec(name="libcxxwrap_julia_jll", uuid="3eaa8342-bff7-56a5-9981-c04077f7cee7"); compat="0.13.2"),
```
fingolfin: you may also wish to switch to a newer `libcxxwrap_julia_jll`, this may even be required for actual run-time Julia 1.12 compat (but then your Julia package also needs a newer CxxWrap.jl in its dependencies)

X/XDiag/build_tarballs.jl
giordano: Could also be a good occasion for requiring Julia v1.10 since that's the current LTS (which would also reduce a lot the number of artifacts)

awietek: Thanks for the review. Previously, I had issues with 1.12, so I thought I also don't want to deal with possible issues in 1.13 and 1.14. But if that's not necessary then I can remove this.

X/XDiag/build_tarballs.jl
awietek: Thanks. Which one should I choose then, I am not so familiar with all of this?

awietek: @fingolfin I updated 'libcxxwrap_julia_jll' compat to 0.14.3 now. You say I then also need to update the CxxWrap dependency in the 'XDiag.jl' package. Should this then just be 0.17, do you know?

fingolfin: Yes, 0.17 is the latest CxxWrap release series and uses libcxxwrap 0.14.3

X/XDiag/build_tarballs.jl
```diff
@@ -74,9 +74,7 @@ filter!(p -> (
     (os(p) == "macos" && arch(p) == "x86_64") ||
     (os(p) == "macos" && arch(p) == "aarch64") ||
     (os(p) == "windows" && arch(p) == "x86_64")) &&
```
fingolfin, line 76: Optional comment, feel free to ignore:


Just as a further note, from my personal "best practices" list: you keep a "positive" list of platforms you support. Maybe consider switching that to a "negative" list which excludes platforms on which there are problems (ideally in each case with a brief comment saying as to what is the problem). E.g. here it seems you mainly want
```
filter!(p -> nbits(p) != 32, platforms)
```

The advantage:
- as new platforms are added to Julia (e.g. RISC-V), you may sometimes automatically gain support for them when rebuilding the JLL
- the comments help with maintainability: when a user asks "why isn't this working on my system?" it may be helpful to know

Disadvantage:
- as you rebuild this JLL, you may see some of the "new" platforms fail
  - however, you can then just filter them out (with a comment stating what the failure was, so that anybody who'd like to get those platforms supported has a head start)
  - well, or you could of course decide to look into the failure and possibly ask for help with those, but that's optional, i.e., only if you have the time and will, otherwise just disable it

fingolfin:
This looks pretty reasonably to me, great work, @awietek. Shall we merge it?

X/XDiag/build_tarballs.jl
awietek: That's a good idea, I'll try that in the future. Makes sense, one can always remove platforms if one can't figure out problems. I am quite sure I don't want to support 32bit architectures though.

awietek:
Yes, let's merge it. I will try to see whether setting the LLVM version helped with the OpenMP on the ARM MacOS systems; otherwise, I'll be back. Thanks for your help!

PR #11442, opened 2025-06-14
Add recipe for juliaup

Keno:
Of course, it's a bit backwards to have this, because if you can install julia packages, you probably already have julia, but in https://github.com/Keno/ClaudeBox.jl I'm using JLLs for the non-standard purpose of putting together a container image in which I'd like to have juliaup available.

### DIFF
--- a/J/juliaup/build_tarballs.jl
+++ b/J/juliaup/build_tarballs.jl
@@ -1,0 +1,50 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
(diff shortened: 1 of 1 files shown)


giordano:
Uhm, I presume it's not intentional to rebuild libjulia here?

Keno:
Well it touched the version number computation which is in that directory. I touched that file intentionally, but of course we can move the file (although that would still rebuild libjulia I guess). We could also just copy the platform list, but I figured might as well have it in one place

fingolfin:
Just was about to comment that we should put the libjulia mod into a separate PR where building is skipped, but @giordano already did just that, so... :-)

PR #11443, opened 2025-06-14
[LBT] Upgrade to v5.13.0

amontoison:


L/libblastrampoline/libblastrampoline@5.13/bundled/cmake/yggdrasilenv.cmake
giordano, line 1: I thought this was upstreamed?

amontoison: @imciner2 Can you confirm that we don't need anymore the patch?

ViralBShah: If I understand this right, it is not actually needed for building LBT, but it gets installed so that other cmake builds can find LBT. It doesn't seem upstreamed just yet, but should be straightforward.

ViralBShah: Does it make sense to upstream this patch? It seems like it is Yggdrasil specific.

imciner2: > If I understand this right, it is not actually needed for building LBT, but it gets installed so that other cmake builds can find LBT.

Yes, this is just for consumers.

> Does it make sense to upstream this patch? It seems like it is Yggdrasil specific.

The core functionality (for the BLAS and LAPACK parts) was upstream into CMake proper in 3.30 - so we only have to do `-DBLA_VENDOR=libblastrampoline` and it will be detected on all platforms (no more futzing with the library variables or with the platform-specific library naming in our build scripts).

The only thing that this would have extra now is the LAPACKE and CBLAS parts, because upstream CMake doesn't actually provide that capability - but I guess that isn't as important because projects expecting those should have their own find modules.

We can probably remove this from Ygg now.

PR #11449, opened 2025-06-16
[normaliz] Rebuild with FLINT 3.3.0

lgoettgens:
cc @fingolfin @benlorenz


lgoettgens:
I had a quick look through the normaliz source and could not find any places where the old internals of flint matrices are accessed. And the compiler does not complain about anything, so I think there should be no instances whatsoever that need to be adapted.

PR #11450, opened 2025-06-16
[Singular] Rebuild with FLINT 3.3.0

lgoettgens:
cc @fingolfin @hannes14

### DIFF
--- a/S/Singular/build_tarballs.jl
+++ b/S/Singular/build_tarballs.jl
@@ -35,9 +35,9 @@ version = VersionNumber(upstream_version.major * 100 + upstream_version.minor + 
                         upstream_version.patch * 100 + version_offset.minor,
                         Int(upstream_version.prerelease[1]) * 100 + version_offset.patch)
 
-# Collection of sources required to build normaliz
+# Collection of sources required to build Singular
 sources = [
-    GitSource("https://github.com/Singular/Singular.git", "8c434595a030dbabb7d02b3a1b16ef028a867078"),
+    GitSource("https://github.com/Singular/Singular.git", "595d7167e6e019d45d9a4f1e18ae741df1f3c41d"),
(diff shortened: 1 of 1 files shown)


lgoettgens:
I had a quick look through the singular source (in particular `factory/FLINTconvert.cc`) and could not find any places where the old internals of flint matrices are accessed.
Let's see if the compiler complains about anything. If not, I think there should be no instances that need to be adapted.

Edit: there are such places...

hannes14:
One needs
GitSource("https://github.com/Singular/Singular.git", "595d7167e6e019d45d9a4f1e18ae741df1f3c41d")

lgoettgens:
> I had a quick look through the singular source (in particular `factory/FLINTconvert.cc`) and could not find any places where the old internals of flint matrices are accessed. Let's see if the compiler complains about anything. If not, I think there should be no instances that need to be adapted.
> 
> Edit: there are such places...

It seems that these are already fixed upstream by https://github.com/Singular/Singular/pull/1278 and https://github.com/Singular/Singular/pull/1280.

I don't now how the Singular_jll versioning is usually done. I just moved the upstream singular from https://github.com/Singular/Singular/commit/8c434595a030dbabb7d02b3a1b16ef028a867078 to https://github.com/Singular/Singular/commit/595d7167e6e019d45d9a4f1e18ae741df1f3c41d, but didn't bump any part of the `upstream_version` variable. Is this fine this way? Alternatives I could think of:
1. bump the `upstream_version`. But to what?
2. don't just move the upstream version, but instead add these two required PRs to the recipes as `.patch` files.

Any opinions on this @fingolfin @hannes14 ?

fingolfin:
Don't worry about the upstream version (Singular uses anti-semantic versioning), just bump `version_offset` as you already did.

PR #11454, opened 2025-06-16
[f4ncgb] new package

benlorenz:
Adds a package for non-commutative Groebner bases (https://gitlab.sai.jku.at/f4ncgb/f4ncgb.git).

~~Currently as draft since we are working on getting the patches upstreamed.~~

Needs a very recent macos sdk and gcc for std::format and concepts from C++20.

cc: @Sequenzer


F/f4ncgb/build_tarballs.jl
```diff
@@ -0,0 +1,79 @@
...
+    Dependency("FLINT_jll", compat = "~301.300.101"),
+    Dependency("boost_jll", compat = "=1.87.0"),
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+# gcc 13 is needed for std::format and thus we cannot dlopen during audit
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               julia_compat="1.10",
+               preferred_gcc_version=v"13",
+               dont_dlopen=true)
```
lgoettgens, line 79: Somehow, the auditor still tries to dlopen the library and log an error (but doesn't fail), even though this line tells him not do...
See https://buildkite.com/julialang/yggdrasil/builds/22011/steps?sid=0198138c-53e9-4bbc-bb7a-445944c2a1bb#0198138c-5477-43d2-8e1b-90b7d9d68536/7-23227

benlorenz: The auditor does indeed check the file. We could disable this as well with `skip_audit=true` but I'm not sure if this is necessary, so I would tend to keep it and the warning can stay as a reminder (once Yggdrasil works with a more recent julia version the flag can be removed).

The `dont_dlopen` flag (in this location) is to make sure a library that cannot be loaded is still accepted as a `LibraryProduct` (and still dlopen'ed at runtime when `using f4ncgb_jll`). Without this option the build would fail like this:
```
[ Info: Could not locate libf4ncgb inside [".../Yggdrasil/F/f4ncgb/build/x86_64-linux-gnu-cxx11/2KilrFiz/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib64", ".../Yggdrasil/F/f4ncgb/build/x86_64-linux-gnu-cxx11/2KilrFiz/x86_64-linux-gnu-libgfortran5-cxx11/destdir/lib"]
┌ Error: Built f4ncgb but libf4ncgb still unsatisfied:
└ @ BinaryBuilder .../packages/BinaryBuilder/zpeEH/src/AutoBuild.jl:973
ERROR: LoadError: Cannot continue with unsatisfied build products!

```

lgoettgens: ah ok, thanks for the explanation

PR #11444, opened 2025-06-14
[libjulia] Split julia/libjulia supported platforms

giordano:
Taken out of #11442.

PR #11455, opened 2025-06-16
[GAP_pkg_*] Rebuild with extended platforms (part 1)

lgoettgens:
Changes:
- Resolves the other TODO from https://github.com/JuliaPackaging/Yggdrasil/pull/11412.
- Applies some comments from https://github.com/JuliaPackaging/Yggdrasil/pull/11412.
- Extends the platform list for gap package jlls to include aarch64-freebsd and riscv (each that fails will have them filtered again in the respective recipe)
- Small changes to `update.jl`:
  - Adapt the Project.toml to include all deps that are loaded, fixed executing it for me
  - Fix the setting of `offset` when the upstream version changes. This is not relevant for anything in this PR, but I noticed that the current code is just wrong. Consider `upstream_version = v"1.2.3"; offset = v"1.0.0"` (so `version = v"101.200.300") and there is now a patch release of upstream. Resetting offset to `v"0.0.0"` would result in losing the API-incompatibility, so only the offset parts strictly left of the change in the upstream version should better be kept.
- Run `update.jl` for a handful packages. Once this is merged, I'll make some more PRs with updates for small-ish groups of the other gap package jlls.  

cc @fingolfin

### DIFF
--- a/G/GAP/common.jl
+++ b/G/GAP/common.jl
@@ -5,11 +5,6 @@ function gap_platforms(; expand_julia_versions::Bool=false)
         platforms = vcat(libjulia_platforms.(julia_versions)...)
     else
         platforms = union(julia_supported_platforms.(julia_versions)...)
-        
-        filter!(p -> arch(p) != "riscv64", platforms) # riscv64 is not supported atm
-
-        # TODO: re-enable FreeBSD aarch64 support once GAP_jll supports it
-        filter!(p -> !(Sys.isfreebsd(p) && arch(p) == "aarch64"), platforms)
     end
 
     # we only care about 64bit builds
--- a/G/GAP/common.jl
+++ b/G/GAP/common.jl
@@ -3,25 +3,20 @@ include("../../L/libjulia/common.jl")
 function gap_platforms(; expand_julia_versions::Bool=false)
     if expand_julia_versions
         platforms = vcat(libjulia_platforms.(julia_versions)...)
-
-        # we only care about 64bit builds
-        filter!(p -> nbits(p) == 64, platforms)
-
-        # Windows is not supported
-        filter!(!Sys.iswindows, platforms)
-
-        return platforms
     else
         platforms = union(julia_supported_platforms.(julia_versions)...)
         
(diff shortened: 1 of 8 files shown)


G/GAP/common.jl
```diff
@@ -3,27 +3,15 @@ include("../../L/libjulia/common.jl")
...
-
-        return platforms
     else
-        # TODO: create the platforms from the above list coming from libjulia
-        # by removing the julia_version tag
-        
-        platforms = supported_platforms()
-        filter!(p -> nbits(p) == 64, platforms) # we only care about 64bit builds
-        filter!(!Sys.iswindows, platforms)      # Windows is not supported
+        platforms = union(julia_supported_platforms.(julia_versions)...)
```
fingolfin, line 7: I don't understand this. Won't this do the same as in the `expand_julia_versions` (just with `union` instead of `vcat`)?

It would make sense if you strip the `julia_version` from each platform before calling `union` I guess?

Instead of `union` it might also be OK to just use `julia_supported_platforms(v"2.0")` as a hack since usually the latest Julia version (and we get that by specifying `v"2.0."` aka `infinity` ;-) ) supports "all" architectures... but perhaps not always...

(So then another option: add `expand_julia_versions` kwarg to `julia_supported_platforms` so you don't need to strip in the first place... ;-)

lgoettgens: That is some magic introduced by https://github.com/JuliaPackaging/Yggdrasil/pull/11444/files.
There are two functions: `julia_supported_platforms` and `libjulia_platforms`. Both take a julia version and return a list of platforms for that julia version, but with one difference: the only the latter has `julia_version` tags in the platforms.
Since I use the former here, no stripping of anything is needed.


> Instead of `union` it might also be OK to just use `julia_supported_platforms(v"2.0")` as a hack since usually the latest Julia version (and we get that by specifying `v"2.0."` aka `infinity` ;-) ) supports "all" architectures... but perhaps not always...

I thought about that, but it seems rather hacky. And the current version should be correct, and is rather cheap. So I see no strong reason to use the hack here.

PR #11464, opened 2025-06-17
[GAP_pkg_*] Rebuild with extended platforms (part 4)

lgoettgens:
Touches some recipes to rebuild with the changes from https://github.com/JuliaPackaging/Yggdrasil/pull/11455.

> Extends the platform list for gap package jlls to include aarch64-freebsd and riscv (each that fails will have them filtered again in the respective recipe)

cc @fingolfin

### DIFF
--- a/G/GAP_pkg/GAP_pkg_semigroups/build_tarballs.jl
+++ b/G/GAP_pkg/GAP_pkg_semigroups/build_tarballs.jl
@@ -50,6 +50,9 @@ dependencies = gap_pkg_dependencies(gap_version)
 platforms = gap_platforms()
 platforms = expand_cxxstring_abis(platforms)
 
+# TODO: re-enable the below platforms
+filter!(p -> arch(p) != "riscv64", platforms)
+
 # The products that we will ensure are always built
 products = [
     FileProduct("lib/gap/semigroups.so", :semigroups),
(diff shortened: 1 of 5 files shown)


G/GAP_pkg/GAP_pkg_semigroups/build_tarballs.jl
```diff
@@ -50,6 +50,9 @@ dependencies = gap_pkg_dependencies(gap_version)
 platforms = gap_platforms()
 platforms = expand_cxxstring_abis(platforms)
 
+# TODO: re-enable the below platforms
+filter!(p -> arch(p) != "riscv64", platforms)
```
fingolfin, line 54: What is the failure for semigroups?

lgoettgens:
```
[14:22:57] src/froidure-pin-max-plus-mat.cpp:41:1: fatal error: error writing to /tmp/ccIbjhhI.s: No space left on device
[14:22:57]    41 | }
[14:22:57]       | ^
[14:22:57] compilation terminated.
```
could also be a fluke in the filesystem, but I don't wanna rebuild all the other things here over and over again, so this will be further examined in a new PR

PR #11467, opened 2025-06-17
[GAP_pkg_semigroups] Build for riscv

lgoettgens:
This failed in https://github.com/JuliaPackaging/Yggdrasil/pull/11464, and was thus moved to its own PR here.


lgoettgens:
it indeed seems that this was just a fluke in the runner, and does not reproduce here.

PR #11469, opened 2025-06-17
[cddlib] Rebuild to get riscv

lgoettgens:
This is required for the dependent GAP_pkg_cddinterface_jll to support all architectures (cf. https://github.com/JuliaPackaging/Yggdrasil/pull/11455)

I have no idea if this could have any effect on CDDLib.jl, thus pinging @odow @blegat

cc @benlorenz @fingolfin

### DIFF
--- a/C/cddlib/build_tarballs.jl
+++ b/C/cddlib/build_tarballs.jl
@@ -5,7 +5,7 @@ using BinaryBuilder, Pkg
 name = "cddlib"
 # We're using an upstream commit to get a bug fix that is not part of an
 # official release yet.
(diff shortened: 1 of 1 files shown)


blegat:
I assume bumping the patch version of GMP shouldn't be an issue but since we fix the version:
https://github.com/JuliaPolyhedra/CDDLib.jl/blob/7fa31a0797e9cf14691e2e1e7d329b0532dcd1ae/Project.toml#L18
it will give us time to upgrade if needed so no objection from our side, thanks for asking.

PR #11471, opened 2025-06-17
Update AcousticsToolbox to include Orca model

mchitre:
Version update to 2025.06.18 to include OALIB and ORCA

### DIFF
--- a/A/AcousticsToolbox/build_tarballs.jl
+++ b/A/AcousticsToolbox/build_tarballs.jl
@@ -4,12 +4,12 @@
 using BinaryBuilder, Pkg
 
 name = "AcousticsToolbox"
-version_string = "2024_12_25"
-version = VersionNumber(replace(version_string, "_" => "."))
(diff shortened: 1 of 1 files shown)


ViralBShah:
It looks like this only has Windows binaries. It should probably only be built on Windows if that is the case.

mchitre:
> It looks like this only has Windows binaries. It should probably only be built on Windows if that is the case.

I am not sure I understand. Why do you think it only has Windows binaries? I see all the builds for all architectures and OSs pass. And a quick sampling of the logs from the build shows that the executables were built fine on other OSs.

ViralBShah:
The ExecutableProducts having .exe extensions is what made me think that. Apologies for the noise. If otherwise ready to go, happy to merge.

mchitre:
They are named with `.exe` on all platforms for consistency. They are Linux and Mac ELF binaries on those platforms. Please do merge. Thanks!

PR #11472, opened 2025-06-17
[libpolymake_julia] Rebuild with FLINT 3.3.0

lgoettgens:
and also update libjulia

cc @benlorenz @fingolfin


fingolfin:
```
ERROR: LoadError: Unsatisfiable requirements detected for package polymake_jll [7c209550]:
  | polymake_jll [7c209550] log:
  | ├─possible versions are: 4.1.0-400.1300.3 or uninstalled
  | └─restricted to versions 400.1300.4 by an explicit requirement — no versions left
```


It seems `polymake_jll` is not yet merged in the registry, see <https://github.com/JuliaRegistries/General/pull/133106>. It looks like it might be a registry error, I'll try to get someone to unwedge it.

benlorenz:
Thanks for taking care of it, looks like it was caused by some github download error so nothing to worry about.

PR #11473, opened 2025-06-17
[hidapi] Update to 0.15.0

Klafyvel:
This drops support for `arch64-unknown-freebsd` and `riscv64-linux-gnu` because the script fails to find `libusb-1.0` (is it an issue with libusb_jll?)


H/hidapi/build_tarballs.jl
```diff
@@ -34,11 +34,16 @@ cmake  \
     ${linuxflags} \
     ..
 
-cmake --build . --target install
+cmake --build . --parallel ${nproc} --target install
 install_license ../hidapi/LICENSE*.txt
 """
 
-platforms = supported_platforms()
+broken_platforms = Set([
```
giordano: Updating libusb to a newer version should easily resolve this

Klafyvel: I must be doing something wrong: when I add `compat="^1.0.23"` (lowest available version of libusb_jll if I'm correct), I still get the error when I try to build `aarch64_unknown_freebsd`.

giordano: What I'm suggesting is to build a new version of libusb in a separate PR: that should automatically bring the missing platforms without more effort.

Klafyvel: ah! sorry I misunderstood you. It's a bit late for me now, I'll do it later.

PR #11475, opened 2025-06-18
[msolve] Rebuild with FLINT 3.3

lgoettgens:
I haven't looked into the msolve source at all if it accesses some of the changed FLINT internals. Let's wait for the compiler to determine that.

cc @ederc @fingolfin @benlorenz

### DIFF
--- a/M/msolve/build_tarballs.jl
+++ b/M/msolve/build_tarballs.jl
@@ -5,7 +5,7 @@ using BinaryBuilder, Pkg
 name = "msolve"
 upstream_version = v"0.8.0"
 
-version_offset = v"0.0.0"
+version_offset = v"0.0.1"
 version = VersionNumber(upstream_version.major*100+version_offset.major,
                         upstream_version.minor*100+version_offset.minor,
                         upstream_version.patch*100+version_offset.patch)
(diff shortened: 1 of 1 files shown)


lgoettgens:
The compiler seems to be happy, which really confuses me, as there are accesses to the no longer existing `rows` field e.g. in https://github.com/algebraic-solving/msolve/blob/280da550cecbc232e85418ecfb6f1d8a10d6dbe8/src/upolmat/nmod_mat_permute_rows.c#L62

ref: https://github.com/algebraic-solving/msolve/issues/220

lgoettgens:
> The compiler seems to be happy, which really confuses me, as there are accesses to the no longer existing `rows` field e.g. in [algebraic-solving/msolve@`280da55`/src/upolmat/nmod_mat_permute_rows.c#L62](https://github.com/algebraic-solving/msolve/blob/280da550cecbc232e85418ecfb6f1d8a10d6dbe8/src/upolmat/nmod_mat_permute_rows.c#L62)
> 
> ref: [algebraic-solving/msolve#220](https://github.com/algebraic-solving/msolve/issues/220)

ok, I am stupid: the occurrences I found are all guarded by a `#if __FLINT_VERSION < 3`...

so since the build succeeds, everything should be fine

PR #11479, opened 2025-06-18
[DuckDB] Bump to v1.3.1

c-herrewijn:


ViralBShah:
I suspect you need a newer GCC to get riscv64 to build.

giordano:
We only have gcc 14 for riscv64, newer gcc would be a version from one month ago.

giordano:
Also, the error was
```
[12:33:05]     ccache: error: execute_noreturn of /opt/riscv64-linux-gnu/bin/riscv64-linux-gnu-gcc failed: No such file or directory
```
which is a filesystem issue (we've been having these issues on the build machine for a while), I simply restarted the job.

ViralBShah:
> We only have gcc 14 for riscv64, newer gcc would be a version from one month ago.

Ah that makes sense. So the preferred GCC really has no meaning for riscv.

giordano:
Right now, no.  When we'll have newer builds of GCC that'd change.

PR #11480, opened 2025-06-18
[LBT] Update 5.11.2 to build risv64 

ViralBShah:


ViralBShah:
Ah cool! It is smarty enough to just build the new architecture.

PR #11488, opened 2025-06-19
Update OpenCL headers

VarLad:
They tagged a new version. Not a new release but would be better than using an arbitrary version in my opinion.


VarLad:
cc: @maleadt

maleadt (approved):

PR #11493, opened 2025-06-19
Legate.jl Wrapper

ejmeitz:
This is the CxxWrap code for Legate.jl.

This build uses headers form Legate which depend on CUDA, hence the expansion of the CUDA version. I do not love that I also have to depend on CUDA to actually get those headers in the build-env but I think that is required.


ejmeitz:
I realize the diff on the patch is too long to preview. Here's a [link ](https://github.com/JuliaLegate/Legate.jl/blob/main/scripts/legion_redop_patch.inl)to the whole file.

L/legate_jl_wrapper/build_tarballs.jl
```diff
@@ -0,0 +1,140 @@
...
+    # Put new CMake first on path
+    export PATH=${host_bindir}:$PATH
+
+    cd ${WORKSPACE}/srcdir/
+    
+    # Apply patch for complex half types...
+    current_header=${prefix}/include/legate/deps/legion/legion_redop.inl 
+    if [ -e "$current_header" ]; then
+        rm $current_header 
+    fi
```
giordano: Why `rm`?  `cp` overrides anyway.

L/legate_jl_wrapper/bundled/legion_redop_patch.inl
giordano, line 1: This is a huge file to carry around.  Also, it's an entire file, not just a diff.  If anything it should be a diff, to show what are the changes (and a patch is also slightly more future proof).  But also, upstreaming the changes would really be a lot better, we don't want to carry patches unnecessarily.

ejmeitz: Yeah I'll make it a diff. NVIDIA is aware we ran into this issue, but I don't have much power to effect change upstream.

ejmeitz:
@giordano Ready to go I believe. Only weird thing on this one is expanding both CUDA and Julia.

L/legate_jl_wrapper/build_tarballs.jl
```diff
@@ -0,0 +1,130 @@
...
+        -DJulia_PREFIX=${prefix} \
+        ../legate_jl_wrapper/
+
+    VERBOSE=ON cmake --build . --config Release --target install -- -j${nproc}
+    install_license $WORKSPACE/srcdir/legate_jl_wrapper*/LICENSE
+
+    if [[ "${target}" == aarch64-linux-* ]]; then
+        # ensure products directory is clean
+        rm -rf ${prefix}/cuda
+    fi
```
giordano: Why is this conditional?  The symlink above is created unconditionally.

ejmeitz: Yeah, to be honest this is just copied from the first script that used nvcc to cross compile libraries for arm. I never really understood its purpose. I assumed that the headers etc. in that folder were somehow for x86, but cannot confirm. 

Its also in the NCCL, legate and whatever script originally did the cross compilation.

ejmeitz: Ok I removed and re-triggered builds

ejmeitz:
@giordano Possible to merge whenever is convenient? I removed the code deleting the CUDA dir.

PR #11507, opened 2025-06-21
[Qt6Wayland] Restore init block

barche:
Wayland was not loaded because of this oversight in the last upgrade


barche:
@giordano Could you relaunch the failing arm7 build please? https://buildkite.com/julialang/yggdrasil/builds/21314#01979414-4094-4813-b265-da022c668e66

barche:
I have to pick this up again after investigating what happened on freeBSD.

PR #11522, opened 2025-06-24
[DuckDB] Explicitly set compatible platform targets

Mytherin:
Should fix issues where extensions can no longer be found in DuckDB.jl

### DIFF
--- a/D/DuckDB/build_tarballs.jl
+++ b/D/DuckDB/build_tarballs.jl
@@ -24,8 +24,7 @@ cmake -B build \
       -DENABLE_EXTENSION_AUTOLOADING=1 \
       -DENABLE_EXTENSION_AUTOINSTALL=1 \
       -DBUILD_UNITTESTS=FALSE \
-      -DBUILD_SHELL=TRUE \
-      -DDUCKDB_EXPLICIT_PLATFORM="${target}"
+      -DBUILD_SHELL=TRUE
 cmake --build build --parallel ${nproc}
 cmake --install build
 
--- a/D/DuckDB/build_tarballs.jl
+++ b/D/DuckDB/build_tarballs.jl
@@ -15,11 +15,11 @@ script = raw"""
 cd $WORKSPACE/srcdir/duckdb/
 
 export DUCKDB_TARGET="${target}"
-if [[ "${target}" == *86*-linux-gnu ]]; then
+if [[ "${target}" == "x86_64-linux-gnu" ]]; then
     export DUCKDB_TARGET="linux_amd64"
 elif [[ "${target}" == aarch64-linux-gnu ]]; then
     export DUCKDB_TARGET="linux_arm64"
-elif [[ "${target}" == *86*-linux-musl* ]]; then
+elif [[ "${target}" == "x86_64-linux-musl" ]]; then
     export DUCKDB_TARGET="linux_amd64_musl"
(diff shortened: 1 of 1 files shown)


D/DuckDB/build_tarballs.jl
```diff
@@ -14,6 +14,23 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/duckdb/
 
+export DUCKDB_TARGET="${target}"
+if [[ "${target}" == *86*-linux-gnu ]]; then
+    export DUCKDB_TARGET="linux_amd64"
```
giordano: Note that i686-linux-gnu would match this, and it's 32-bit platform.  Is that expected to be `DUCKDB_TARGET="linux_amd64"`?  Just checking.

Mytherin: Thanks - indeed that's not correct.

Mytherin: Is there any way of checking the `cxx` version? `linux_amd64` extensions are compiled against `cxx11`, the `cxx03` versions should target `linux_amd64_gcc4` (which we do not distribute anymore - but an error on fetching the extension is better than downloading an incompatible extension and crashing).

giordano: To be clear, cxx03/cxx11 is about the [stdlibc++ dual ABI](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html), not about the C++ standard: https://docs.binarybuilder.org/stable/building/#Expanding-C-string-ABIs-or-libgfortran-versions

Mytherin: Yes I'm aware, we use the platform to link to extensions (effectively shared libraries) that need to be ABI compatible otherwise weird things happen (generally the program crashes). We used to distribute binaries for the old C++ ABI under platform name `_gcc4` but have recently deprecated the old ABI and now only distribute the latest ABI. 

I've solved the issue now with the following check:

```bash
if [[ "${CMAKE_TARGET_TOOLCHAIN}" == *-cxx03* ]]; then
    export DUCKDB_TARGET="${DUCKDB_TARGET}_gcc4"
fi
```

Let me know if there's another way that is more direct - otherwise it seems like this is working in the builds I've manually checked.

D/DuckDB/build_tarballs.jl
```diff
@@ -14,6 +14,27 @@ sources = [
...
+    export DUCKDB_TARGET="linux_amd64_musl"
+elif [[ "${target}" == "x86_64-w64-mingw32" ]]; then
+    export DUCKDB_TARGET="windows_amd64_mingw"
+elif [[ "${target}" == "x86_64-apple-darwin" ]]; then
+    export DUCKDB_TARGET="osx_amd64"
+elif [[ "${target}" == "aarch64-apple-darwin" ]]; then
+    export DUCKDB_TARGET="osx_arm64"
+fi
+
+if [[ "${CMAKE_TARGET_TOOLCHAIN}" == *-cxx03* ]]; then
```
giordano: [`bb_full_target`](https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables) is more semantically accurate than relying on the path where the cmake toolchain file is:
```suggestion
if [[ "${bb_full_target}" == *-cxx03* ]]; then
```

Mytherin:
Thanks for the help!

PR #11526, opened 2025-06-24
Add Patch to Legate

ejmeitz:
Adds patch to legate_jll instead of the wrapper code.

The patch is applied to a dependency of legate that we do not wrap and do not have control over. It is applied after configuration which is responsible for cloning the source of the external library we patch. This also ensures the patch is shipped to all dependencies of legate_jll.


ejmeitz:
@giordano Could you please merge this when you have the time! Thanks

PR #11533, opened 2025-06-25
Update Quantum ESPRESSO to 7.4.1

Technici4n:
Not sure what to do with the version? I don't want to bother compiling 7.1, 7.2 and 7.3, but on the other hand the Julia package registry will probably reject 7.0 -> 7.4? Additionally, it seems that QE made a few "incompatible changes" (breaking changes) between 7.0 and 7.4.


lgoettgens:
> I don't want to bother compiling 7.1, 7.2 and 7.3, but on the other hand the Julia package registry will probably reject 7.0 -> 7.4?

No, it won't complain for jlls

Q/QuantumEspresso/build_tarballs.jl
```diff
@@ -4,28 +4,19 @@ const YGGDRASIL_DIR = "../.."
...
 name = "QuantumEspresso"
-version = v"7.0.1"
-quantumespresso_version = v"7.0.0"
-
-# ES 2022-08-01: I tried updating to 7.1, but I encountered build problems:
-# - v7.1 encounters ICEs in gfortran; requires at least GCC 10
-# - the build fails because the directory "../W90" is not found at some point
-# version = v"7.1"
-# quantumespresso_version = v"7.1"
+version = v"7.4.0"
```
lgoettgens:
```suggestion
version = v"7.4.1"
```
why don't follow the upstream version here?

Technici4n: Yeah you're right, no idea why I did this.

Technici4n:
Good for merge now I think!

Q/QuantumEspresso/build_tarballs.jl
```diff
@@ -4,28 +4,19 @@ const YGGDRASIL_DIR = "../.."
...
-# quantumespresso_version = v"7.1"
+version = v"7.4.1"
+quantumespresso_version = v"7.4.1"
 
 sources = [
-    ArchiveSource("https://gitlab.com/QEF/q-e/-/archive/qe-7.0/q-e-qe-7.0.tar.gz",
-                  "85beceb1aaa1678a49e774c085866d4612d9d64108e0ac49b23152c8622880ee"),
-    # ArchiveSource("https://gitlab.com/QEF/q-e/-/archive/qe-7.1/q-e-qe-7.1.tar.gz",
-    #               "d56dea096635808843bd5a9be2dee3d1f60407c01dbeeda03f8256a3bcfc4eb6"),
+    ArchiveSource("https://gitlab.com/QEF/q-e/-/archive/qe-7.4.1/q-e-qe-7.4.1.tar.gz",
```
giordano:
```suggestion
    ArchiveSource("https://gitlab.com/QEF/q-e/-/archive/qe-$(version)/q-e-qe-$(version).tar.gz",
```

Q/QuantumEspresso/build_tarballs.jl
```diff
@@ -57,7 +48,7 @@ else
 fi
 
 ./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} ${flags[@]}
-make all "${make_args[@]}" -j $nproc
```
giordano: Why removing nproc? That makes compilation a lot slower, no?

Technici4n: Yeah, it's a lot slower on CI at least. I did it because the QE make files seem not to support parallel compilation (I had spurious failures with `*** No rule to make target '../../PHonon/PH/libph.a', needed by 'kcw.x'.  Stop.`).

Technici4n: OK, the QE docs mention that parallel make should be supported, so I'll investigate a bit more.

Technici4n: Fixed!

PR #11536, opened 2025-06-25
Libxc: Add fortran .so product

Technici4n:
Forgot to do that in https://github.com/JuliaPackaging/Yggdrasil/pull/11418, so the QE JLL is still not happy.


Technici4n:
OK, tested and it works. (It requires a newly built QE, linking against `libxc.so.15` - the latest built JLL links against `libxc.so.9`).

PR #11543, opened 2025-06-26
[slope] Update to 5.1.1

jolars:
https://github.com/jolars/libslope/compare/v2.6.0...v5.1.1

### DIFF
--- a/S/slope/build_tarballs.jl
+++ b/S/slope/build_tarballs.jl
@@ -44,6 +44,7 @@ install_license $WORKSPACE/srcdir/libslope/LICENSE
 """
 
 include("../../L/libjulia/common.jl")
+julia_versions = filter(v -> v >= v"1.10", julia_versions)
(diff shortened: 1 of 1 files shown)


giordano:
Is it necessary to still build for all julia versions back to v1.6?  would it be possible to build only v1.10 (current LTS version) instead?  According to https://juliahub.com/ui/Packages/General/slope_jll/ there aren't even packages using it at the moment.

jolars:
No, that makes sense. I've updated the PR now. The SLOPE package uses it (<https://juliahub.com/ui/Packages/General/slope_jll#dependents>).

giordano:
Ah, the badge is misleading 😅

giordano:
Ah, you need libcxxwrap_julia_jll 0.14.3 for riscv64: https://github.com/JuliaPackaging/Yggdrasil/blob/f7bf6a01ab17ba0b9a2e424eb0654e9d3c3ac380/S/slope/build_tarballs.jl#L60

PR #11552, opened 2025-06-27
quickfixc build script

AlexKlo:


Q/quickfixc/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+cmake \
+    -DCMAKE_INSTALL_PREFIX=$prefix \
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    ..
+
+make -j${nproc} install
+"""
+
+platforms = [Platform("x86_64", "linux"; libc="glibc"), Platform("aarch64", "macos")]
```
giordano: Any reason for having only these two platforms?

Q/quickfixc/build_tarballs.jl
```diff
@@ -0,0 +1,47 @@
...
+
```
giordano, line 41: From the audit log
```suggestion
platforms = expand_cxxstring_abis(platforms)

```

Q/quickfixc/build_tarballs.jl
AlexKlo: I’m only using these two platforms at the moment and haven’t tested on others yet.

AlexKlo:
@giordano I have added more platforms.
Can we merge it?

PR #11565, opened 2025-06-27
[FLINT] Build for riscv

lgoettgens:
cc @fingolfin @benlorenz @Joel-Dahne

### DIFF
--- a/F/FLINT/build_tarballs.jl
+++ b/F/FLINT/build_tarballs.jl
@@ -72,7 +72,7 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-               julia_compat = "1.6", preferred_gcc_version=v"6",
+               julia_compat = "1.6", preferred_gcc_version=v"8",
                init_block = """
   if !Sys.iswindows() && !(get(ENV, "NEMO_THREADED", "") == "1")
     #to match the global gmp ones


lgoettgens:
Hmm, unfortunately, thereis no OpenBLAS32 0.3.29 for powerpc. I'll instead try to get a 0.3.28 build for riscv (cf https://github.com/JuliaPackaging/Yggdrasil/pull/11567) and then change the version here back.

giordano:
> Hmm, unfortunately, thereis no OpenBLAS32 0.3.29 for powerpc

There is, but only with libgfortran5

lgoettgens:
> > Hmm, unfortunately, thereis no OpenBLAS32 0.3.29 for powerpc
> 
> There is, but only with libgfortran5

How can I use that then? (without expanding libgfortran versions and filtering all !=5 out)

giordano:
Require gcc 8

F/FLINT/build_tarballs.jl
```diff
@@ -69,8 +66,8 @@ products = [
 # Dependencies that must be installed before this package can be built
 dependencies = [
     Dependency("GMP_jll", v"6.2.1"),
-    Dependency("MPFR_jll", v"4.1.1"),
-    Dependency("OpenBLAS32_jll", v"0.3.28"),
+    Dependency("MPFR_jll", v"4.2.1"),
```
benlorenz: julia 1.6 (which is claimed in the compat entry) has MPFR 4.1.1 and even julia 1.10 has a lower version (4.2.0). Using an older version at runtime seems risky.

lgoettgens: Yeah, good point. I opened https://github.com/JuliaPackaging/Yggdrasil/pull/11576 to rebuild MPFR 4.1.1 and 4.2.0 with updated platform support. Once that is in, I'll revert the MPFR version change here

lgoettgens:
```suggestion
    Dependency("MPFR_jll", v"4.1.1"),
```

lgoettgens:
This is now passing @fingolfin @benlorenz

PR #11570, opened 2025-06-28
[Gnuplot_jll] update `Gnuplot` to `6.0.3`

t-bltg:
For https://github.com/mbaz/Gaston.jl/issues/135#issuecomment-3011072655.

~~Needs https://github.com/JuliaPackaging/Yggdrasil/pull/11571 to get in first.~~

PR #11571, opened 2025-06-28
[Gnuplot_jll] fix `Gnuplot` 5 build

t-bltg:
Fix gnuplot 5 build (which currently fails to build).

### DIFF
--- a/G/Gnuplot/build_tarballs.jl
+++ b/G/Gnuplot/build_tarballs.jl
@@ -15,7 +15,6 @@ sources = [
 script = raw"""
 cd $WORKSPACE/srcdir/gnuplot-*/
 
-
 if [[ "${target}" == "${MACHTYPE}" ]]; then
     # Delete system libexpat to avoid confusion
     rm /usr/lib/libexpat.so*
@@ -26,10 +25,16 @@ elif [[ "${target}" == *-mingw* ]]; then
     autoreconf -fiv
 fi
 
+# FIXME: no Qt Tools artifacts available for these platforms
+# -> missing either uic or lrelease
+case "$target" in
+    aarch64-apple-darwin) args+=(--with-qt=no);;
+esac
+
 export CPPFLAGS="$(pkg-config --cflags glib-2.0) $(pkg-config --cflags cairo) $(pkg-config --cflags pango) -I$(realpath term)"
 export LIBS="-liconv"
 
-./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target}
+./configure --prefix=${prefix} --build=${MACHTYPE} --host=${target} ${args[@]}
 
 cd src
 make -j${nproc}
--- a/G/Gnuplot/build_tarballs.jl
+++ b/G/Gnuplot/build_tarballs.jl
@@ -1,7 +1,7 @@
 using BinaryBuilder, Pkg
 
 name = "Gnuplot"
-version = v"5.4.6"
+version = v"5.4.5"
 julia_build = 0  # NOTE: increment on rebuild
 
 # Collection of sources required to complete build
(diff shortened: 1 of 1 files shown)


t-bltg:
@giordano, this is ready to be merged :pray: .

G/Gnuplot/build_tarballs.jl
```diff
@@ -13,23 +12,27 @@ sources = [
...
 script = raw"""
-cd $WORKSPACE/srcdir/gnuplot-*/
-
+cd $WORKSPACE/srcdir/gnuplot-*
 
 if [[ "${target}" == "${MACHTYPE}" ]]; then
     # Delete system libexpat to avoid confusion
     rm /usr/lib/libexpat.so*
 elif [[ "${target}" == *-mingw* ]]; then
     # This is needed because otherwise we get unusable binaries (error "The specified executable is not a valid application for this OS platform"). These come from CompilerSupportLibraries_jll:
```
giordano: This comment doesn't make sense anymore without the code you deleted.

giordano (changes requested):
This was last built three years ago, the version number must change because of implicit changes in the compat bounds of all jlls since then.

G/Gnuplot/build_tarballs.jl
t-bltg: Fixed.

t-bltg:
> This was last built three years ago, the version number must change because of implicit changes in the compat bounds of all jlls since then.

This PR is merely a rebuild in order to prepare for the gnuplot `6.X` upgrade in https://github.com/JuliaPackaging/Yggdrasil/pull/11570.

What should I do there, use `5.4.6` ? This desynchronizes with the upstream gnuplot version ...

giordano:
If there isn't another upstream version to build, that's the only option, yes. This is a limitation of Pkg which can't handle changing compat bounds without changing the x.y.z version.

t-bltg:
~~I'm more enclined towards having https://github.com/JuliaPackaging/Yggdrasil/pull/11570 for gnuplot `6.x` go in then, and drop this PR ...~~

~~What do you think ?~~

Forget it, there is 5.4.6 :)

t-bltg:
Are version numbers using build number such as
```julia
julia> VersionNumber(5, 4, 0, ("build", 1))
v"5.4.0-build.1"
```
allowed in Yggdrasil ?

giordano:
No, you must stick to x.y.z only. Pkg really doesn't like anything out of that, don't try to trick it, it will end badly.

G/Gnuplot/build_tarballs.jl
```diff
@@ -1,35 +1,38 @@
-
 using BinaryBuilder, Pkg
 
 name = "Gnuplot"
-version = v"5.4.5"
+version = v"5.4.6"
```
giordano: If you're faking the version number anyway, why not sticking to 5.4.5?

t-bltg: I'm fine with either 5.4.5 or bump to the latest 5.4.10.

G/Gnuplot/build_tarballs.jl
```diff
@@ -40,21 +43,22 @@ make install
...
     Dependency(PackageSpec(name="Cairo_jll", uuid="83423d85-b0ee-5818-9007-b63ccbeb887a")),
     Dependency(PackageSpec(name="Libiconv_jll", uuid="94ce4f54-9a6c-5748-9c1c-f9c7231a4531")),
+    BuildDependency(PackageSpec(name="Qt5Tools_jll", uuid="a9c6e4b1-b2fb-56d5-96a9-25f276f13840")),
+    Dependency(PackageSpec(name="Qt5Svg_jll", uuid="3af4ccab-a251-578e-a514-ea85a0ba79ee")),
 ]
 
 # Build the tarballs, and possibly a `build.jl` as well.
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; julia_compat="1.6", preferred_gcc_version = v"8")
+build_tarballs(
+    ARGS, name, VersionNumber(version.major, version.minor, 1_000 * version.patch + julia_build),
```
giordano: Last thing: can you move the assignment of this version number to a variable to right after the definition of `version` and `julia_build`, so that it's easier to spot that there's some shenagians going on?

t-bltg: Sure, done in https://github.com/JuliaPackaging/Yggdrasil/pull/11571/commits/d9508ad2528be7caf24ab121b92df17f7e9aa475.

PR #11576, opened 2025-06-30
[MPFR] Split recipe into version-specific ones

lgoettgens:
As mentioned by @benlorenz in https://github.com/JuliaPackaging/Yggdrasil/pull/11565#discussion_r2172725802, yggdrasil recipes should use the MPFR version that is the one bundled with the julia_compat specified in the recipe.
This means that recipes still wanting to support julia 1.6 need MPFR 4.1.1, julia 1.10 needs 4.2.0.

This PR splits the recipe into a common part and a very compact version-specific part (very much inspired by OpenBLAS_jll) for MPFR 4.1.1, 4.2.0, and the more recent ones.
This makes it easily possible to rebuild other JLLs that still wanna use an older MPFR version due to their julia_compat, but with the whole list of platforms (e.g. https://github.com/JuliaPackaging/Yggdrasil/pull/11565), as this PR will rebuild all MPFR versions. The alternative would be something like https://github.com/JuliaPackaging/Yggdrasil/pull/9543 and https://github.com/JuliaPackaging/Yggdrasil/pull/9545, but I find these split-up recipes much clearer in these cases where one may want to rebuild some older version multiple times (and at least MPFR 4.2.0, which is in julia 1.10, will need rebuilds for any new platforms as long as 1.10 is lts).

ping @giordano @ararslan @benlorenz @fingolfin


giordano:
To be clear, this is only reorganisation of the recipes and nothing actually needs to be rebuilt, right?

M/MPFR/common.jl
```diff
@@ -0,0 +1,72 @@
...
+    # On Windows, make sure non-versioned filename exists...
+    if [[ ${target} == *mingw* ]]; then
+        cp -v ${libdir}/libmpfr-*.dll ${libdir}/libmpfr.dll
+    fi
+    """
+    return script
+end
+
+function mpfr_platforms(; experimental::Bool=true, kwargs...)
+    platforms = supported_platforms(; experimental)
```
imciner2:
```suggestion
    platforms = supported_platforms()
```
Not needed.

imciner2 (commented):
From an organizational point of view, it would be good to include (either as a comment or in a readme in the MPFR directory) what Julia version each is associated with.

> (and at least MPFR 4.2.0, which is in julia 1.10, will need rebuilds for any new platforms as long as 1.10 is lts)

I don't think new platforms would be backported to the LTS release, though, and I can't think of what new platforms we would have in the pipeline right now either.

lgoettgens:
> To be clear, this is only reorganisation of the recipes and nothing actually needs to be rebuilt, right?

The motivation for this whole PR is to have a rebuild of 4.1.1 and 4.2.0 for all current platforms. And as a side-effect it makes future rebuilds easier.

> From an organizational point of view, it would be good to include (either as a comment or in a readme in the MPFR directory) what Julia version each is associated with.

Good point, I'll add that tomorrow.



> > (and at least MPFR 4.2.0, which is in julia 1.10, will need rebuilds for any new platforms as long as 1.10 is lts)
> 
> I don't think new platforms would be backported to the LTS release, though, and I can't think of what new platforms we would have in the pipeline right now either.

I am not talking about platforms being backported but instead yggdrasil recipes that still want to support the LTS release (thus wanna depend on an older MPFR) but also want to build for all current platforms, e.g. FLINT_jll still supports julia 1.6 but I want to get rid of the platform filters.

M/MPFR/common.jl
lgoettgens: I know, but allowing to pass the kwarg seems like future-proofing for me (this is taken from the OpenBLAS `common.jl`). If you want to, I can get rid of this, but then I would also remove the kwarg in the line above.

imciner2: `experimental` was only used for the Apple aarch64 M1 bringup, and mainly because pre-1.6 versions choked on the idea of aarch64 darwin. The riscv platform that was just rolled out didn't use `experimental`, and I doubt it will be used anytime in the future since Julia 1.6+ gracefully supports unknown platforms: https://github.com/JuliaPackaging/Yggdrasil/pull/5057#discussion_r902929036.

And yep, remove the kwarg as well.

giordano: `experimental` simply doesn't do anything nowadays, the keyword is kept only for compatibility, not to make recipes using error, but maybe we should just remove it at some point.

lgoettgens:
```suggestion
function mpfr_platforms(; kwargs...)
    platforms = supported_platforms()
```

lgoettgens: Done, see https://github.com/JuliaPackaging/Yggdrasil/pull/11576/commits/aa186ca70ea45aa45abd2b0e87bbbe2576cf4779

M/MPFR/MPFR@4.1.1/build_tarballs.jl
```diff
@@ -0,0 +1,23 @@
...
+
+name = "MPFR"
+version = v"4.1.1"
+
+sources = mpfr_sources(version)
+script = mpfr_script()
+platforms = mpfr_platforms()
+products = mpfr_products()
+
+preferred_llvm_version = v"13.0.1"
```
fingolfin, line 15: This seems to be always the same? Even if not, how about just "inlining" it into the `build_tarballs` argument list directly?

M/MPFR/MPFR@4.1.1/build_tarballs.jl
```diff
@@ -0,0 +1,23 @@
...
+
+# Dependencies that must be installed before this package can be built
+dependencies = mpfr_dependencies(platforms; llvm_compilerrt_version=preferred_llvm_version)
```
fingolfin, line 18: ... and then herr use the default?
```suggestion
dependencies = mpfr_dependencies(platforms)
```

M/MPFR/MPFR@4.1.1/build_tarballs.jl
lgoettgens: I find it more consise to pass this explicitly to both places. IMO, this will help the future person that needs to change the llvm version for some future version of MPFR

giordano:
You can't have register multiple versions of the same package in one go: depending on the content of the Versions.toml and Compat.toml files in the registry, you're almost guaranteed to get conflicts in the PRs to General by construction.

lgoettgens:
> You can't have register multiple versions of the same package in one go: depending on the content of the Versions.toml and Compat.toml files in the registry, you're almost guaranteed to get conflicts in the PRs to General by construction.

ah, good point. Could you then merge this with a `[skip build]` or `[skip ci]` tag or whatever to block the registering of rebuilds? I will then provide separate follow-up PRs for each of the versions that just touch some whitespace or so.

lgoettgens:
> > You can't have register multiple versions of the same package in one go: depending on the content of the Versions.toml and Compat.toml files in the registry, you're almost guaranteed to get conflicts in the PRs to General by construction.
> 
> ah, good point. Could you then merge this with a `[skip build]` or `[skip ci]` tag or whatever to block the registering of rebuilds? I will then provide separate follow-up PRs for each of the versions that just touch some whitespace or so.

bump @giordano, would that be possible? or is there any other way forward you would prefer?

PR #11578, opened 2025-07-01
dist4-mri quick recipe 

Fe-r-oz:
This PR aims to build a executable recipe for unrelased and unpublished https://github.com/QEC-pages/dist-m4ri which we supposed to be 0.0.1. This is my first time building any recipe for I really appreciate the guidance! I used the earlier script as reference to write this script. Thank you!


D/dist-m4ri/build_tarballs.jl
```diff
@@ -0,0 +1,26 @@
+using BinaryBuilder, Pkg
+
+name = "dist-m4ri"
```
giordano: Can't use dashes in package names, won't be easily referenceable in Julia code. Can do something like
```suggestion
name = "dist_m4ri"
```
Note that the directory name must match exactly this name, so you'll need to rename that accordingly.

D/dist-m4ri/build_tarballs.jl
```diff
@@ -0,0 +1,26 @@
...
+sources = [
+    GitSource("https://github.com/QEC-pages/dist-m4ri.git",
+              "0c540cb766f5538d9a89152dc875ae5a357ee6c6")
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir/dist-m4ri/
+make -j${nproc} CC=$CC CXX=$CXX
+mkdir -p ${bindir}
+cp dist-m4ri ${bindir}/
```
giordano:
```suggestion
install -Dvm 755 "dist-m4ri${exeext}"  -t "${bindir}"
```

D/dist-m4ri/build_tarballs.jl
```diff
@@ -0,0 +1,26 @@
...
+"""
+
+platforms = filter(p -> Sys.islinux(p) || Sys.isapple(p), supported_platforms())
+
+products = [
+    ExecutableProduct("dist-m4ri", :dist_m4ri)
+]
+
+dependencies = []
```
giordano:
```suggestion
dependencies = Dependency[]
```

D/dist-m4ri/build_tarballs.jl
```diff
@@ -0,0 +1,26 @@
...
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir/dist-m4ri/
+make -j${nproc} CC=$CC CXX=$CXX
+mkdir -p ${bindir}
+cp dist-m4ri ${bindir}/
+"""
+
+platforms = filter(p -> Sys.islinux(p) || Sys.isapple(p), supported_platforms())
```
giordano: Any particular reason for excluding Windows and FreeBSD?

D/dist-m4ri/build_tarballs.jl
Fe-r-oz: Thank you!

D/dist-m4ri/build_tarballs.jl
Fe-r-oz: Thank you! Added Windows and FreeBSD

D/dist_m4ri/build_tarballs.jl
```diff
@@ -0,0 +1,25 @@
...
+              "0c540cb766f5538d9a89152dc875ae5a357ee6c6")
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir/dist-m4ri/src/
+make -j${nproc} CC=$CC CXX=$CXX
+install -Dvm 755 "dist-m4ri${exeext}"  -t "${bindir}"
+"""
+
+platforms = filter(p -> Sys.islinux(p) || Sys.isapple(p) || Sys.isfreebsd(p) || Sys.iswindows(p), supported_platforms())
```
Krastanov, line 17: @Fe-r-oz , while Mose is the one to really say how these things are set up, I usually try with all `supported_platforms()` directly, without filtering anything out. That way you can even detect issues with the build recipe that have not surfaced in previous builds.

Krastanov:
@Fe-r-oz, have you been able to build this locally before? This is usually how I debug the very first platform, and send the recipe for building here (especially if I work on a device where I can not download the many GB of platform-specific build containers).

For instance, once I have a draft build script, I would try to debug it locally as follows:

```
julia +1.7 --project=. build_tarballs.jl x86_64-linux-gnu --debug=error
```

If this fails, it will drop me in a shell (inside of the build container) in which I can start experimenting manually with the build command. If there was a missing file error (which is the case for you right now), I can look around, find the file, and modify the build command accordingly. Only after this build succeeds for me locally, I submit the build script as a PR here.

Lastly, it would probably make more sense to learn how to do these builds with the m4ri library first, as (1) it is a dependency that you need to build this anyway and (2) it is a better packaged library which will have way fewer warts when one tries to build it, so it is better learning experience.

Fe-r-oz:
@Krastanov, Thank you for your feedback and advice! I will build work on `m4ri` library first and follow the step you mentioned before submitting it here.

PR #11581, opened 2025-07-02
[libcint] Add windows platform

dnkats:
The current new version of libcint (6.1.2) seems to be compatible with windows.


giordano:
Thanks!

PR #11595, opened 2025-07-03
fix `gnuplot_qt` and `gnuplot_x11`

t-bltg:
For https://github.com/mbaz/Gaston.jl/pull/192.

~We need to figure out what to do when the qt terminal is unavailable on some platforms.~
I've added a fake executable in `libexec`, in order to determine `GNUPLOT_DRIVER_DIR` at runtime.

### DIFF
--- a/G/Gnuplot/build_tarballs.jl
+++ b/G/Gnuplot/build_tarballs.jl
@@ -12,6 +12,8 @@ sources = [
     DirectorySource("./bundled"),
 ]
 
+libexec_path = joinpath("libexec", "gnuplot", "$(version.major).$(version.minor)")
+
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/gnuplot-*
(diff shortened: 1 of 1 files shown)


G/Gnuplot/build_tarballs.jl
```diff
@@ -43,14 +45,30 @@ esac
 
 make -C src -j${nproc}
 make -C src install
+
+if [[ "${target}" == *-mingw* ]]; then
+    suffix='.exe'
```
giordano: exeext does this already: https://docs.binarybuilder.org/stable/build_tips/#Automatic-environment-variables

G/Gnuplot/build_tarballs.jl
```diff
@@ -43,14 +45,30 @@ esac
...
+chmod +x \$prefix/$libexec_path/gnuplot_fake\$suffix
 """
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = supported_platforms()
 
 # The products that we will ensure are always built
-products = [ExecutableProduct("gnuplot", :gnuplot)]
+# @show libexec_path
```
giordano: Dead code
```suggestion
```

G/Gnuplot/build_tarballs.jl
t-bltg: Damn, I've `grep`ed the whole repo and searched the docs, but missed it.
Thanks :pray:

PR #11598, opened 2025-07-04
[DynarePreprocessor] Bump version to 7.0.0

svillemot:
– New upstream snapshot
– Switch to meson build system

### DIFF
--- a/D/DynarePreprocessor/build_tarballs.jl
+++ b/D/DynarePreprocessor/build_tarballs.jl
@@ -1,43 +1,31 @@
 using BinaryBuilder
 
 name = "DynarePreprocessor"
-version = v"6.4.0"
+version = v"7.0.0"
 sources = [
-    GitSource("https://git.dynare.org/Dynare/preprocessor.git", "8d72527518523620050e819e9cd59afc76862ab9"),
-    ArchiveSource("https://github.com/phracker/MacOSX-SDKs/releases/download/11.3/MacOSX11.3.sdk.tar.xz",
-                  "cd4f08a75577145b8f05245a2975f7c81401d75e9535dcffbb879ee1deefcbf4"),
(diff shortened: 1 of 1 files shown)


D/DynarePreprocessor/build_tarballs.jl
```diff
@@ -1,39 +1,27 @@
...
-mkdir -p "${bindir}"
-strip "src/dynare-preprocessor${exeext}"
-cp "src/dynare-preprocessor${exeext}" "${bindir}"
+# Use flex from `flex_jll` instead of the one under /usr/bin,
+# because there is no /usr/include/FlexLexer.hh
+echo -e "[binaries]\nflex = '${host_bindir}/flex'\n" > flex.ini
+
+# We enforce GCC on all platforms (i.e. incl. macOS and FreeBSD), by using the GCC toolchain file
+# See https://docs.binarybuilder.org/stable/build_tips/#Using-GCC-on-macOS-and-FreeBSD
+meson setup --cross-file="${MESON_TARGET_TOOLCHAIN%.*}_gcc.meson" --cross-file=flex.ini --buildtype=release build
```
giordano, line 18: Just to understand: why using GCC also on macOS and FreeBSD?  If we use GCC also on those platforms we need to expand the C++ string ABI also there (i.e. use `expand_cxxstring_abis(...; skip=Returns(false))`)

svillemot: We require GCC because it is the only compiler supported by the Dynare Team. We use many C++20 features and this frequently creates problems with Clang because the two compilers have different feature sets for recent C++ standards.

svillemot: I’ve updated the PR so that the C++ string ABI gets expanded on macOS and FreeBSD.

PR #11601, opened 2025-07-04
DataEcon v0.3.2

bbejanov:


D/DataEcon/build_tarballs.jl
```diff
@@ -25,7 +25,9 @@ install -Dvm 644 "include/daec.h" "${includedir}/daec.h"
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
-platforms = supported_platforms()
+platforms = supported_platforms(;
+    exclude=[Platform("riscv64", "linux"; libc="glibc")]
+)
```
giordano: Why?  If you're referring to
```
[20:42:33] src/sqlite3/shell.c: In function ‘fileLinkStat’:
[20:42:33] src/sqlite3/shell.c:5771:10: error: implicit declaration of function ‘lstat’; did you mean ‘fstat’? [-Wimplicit-function-declaration]
[20:42:33]  5771 |   return lstat(zPath, pStatBuf);
[20:42:33]       |          ^~~~~
[20:42:33]       |          fstat
```
that's an upstream bug that must be fixed.  For the record, you're running into https://gcc.gnu.org/gcc-14/porting_to.html#implicit-function-declaration, but those warnings are already in all other platforms, it's just that in previous versions of GCC those warnings aren't fatal.

giordano: [`symlink`](https://linux.die.net/man/2/symlink) and [`lstat`](https://linux.die.net/man/2/lstat) are conditionally defined, with `-std=c99` they are ***not*** defined.  The solution is hinted in the docs linked above: use the GNU extension, which enables those functions: `-std=gnu99`.

giordano: Oh, you're also running into https://gcc.gnu.org/gcc-14/porting_to.html#int-conversion
```
[20:42:33] src/sqlite3/shell.c:23278:25: error: assignment to ‘char *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
[20:42:33] 23278 |         azName[nName*2] = strdup(zSchema);
[20:42:33]       |                         ^
[20:42:33] src/sqlite3/shell.c:23279:27: error: assignment to ‘char *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
[20:42:33] 23279 |         azName[nName*2+1] = strdup(zFile);
[20:42:33]       |                           ^
[20:42:33] src/sqlite3/shell.c: In function ‘main’:
[20:42:33] src/sqlite3/shell.c:27199:19: error: assignment to ‘char *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
[20:42:33] 27199 |       data.zNonce = strdup(argv[++i]);
[20:42:33]       |                   ^
[20:42:33] src/sqlite3/shell.c:27485:18: error: assignment to ‘char *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
[20:42:33] 27485 |         zHistory = strdup(zHistory);
[20:42:33]       |                  ^
```
This must all be fixed upstream.

bbejanov: Thanks for the review. Let me try to address some of the issues. I'll let you know when I'm ready for you to review again.

bbejanov:
@giordano thanks for your patience. I believe I have addressed all issues. Please take another look.

PR #11607, opened 2025-07-05
[libpolymake_julia] update to 0.14, build with polymake 4.14, drop julia < 1.10

benlorenz:


benlorenz:
All failing due to some weird dependency error:
```
ERROR: LoadError: Unsatisfiable requirements detected for package libblastrampoline_jll [8e850b90]:
 libblastrampoline_jll [8e850b90] log:
 ├─possible versions are: 3.0.4 or uninstalled
 └─found to have no compatible versions left with Ipopt_jll [9cc047cb]
   └─Ipopt_jll [9cc047cb] log:
     ├─possible versions are: 3.13.1-300.1400.1701 or uninstalled
     └─restricted by compatibility requirements with SCIP_jll [e5ac4fe4] to versions: 300.1400.1400-300.1400.1701
       └─SCIP_jll [e5ac4fe4] log:
         ├─possible versions are: 0.1.0-900.200.200 or uninstalled
         └─restricted by compatibility requirements with polymake_jll [7c209550] to versions: 900.200.200
           └─polymake_jll [7c209550] log:
             ├─possible versions are: 4.1.0-400.1400.0 or uninstalled
             └─restricted to versions 400.1400.0 by an explicit requirement, leaving only versions 400.1400.0
```
This seems to be due to the updated `SCIP_jll` dependency in `polymake_jll`, while it built fine there it fails here, probably due to the use of `julia_version`ed platforms.
There does not seem to be any other path to `libblastrampoline_jll` other than scip and as a stdlib. `Ipopt_jll` has compat `5.4.0` which conflicts with something requesting version 3.0.4.
Version 3.0.4 corresponds to julia 1.7.3, the version we use for building, not the target version.

benlorenz:
I removed some more entries from this stdlibs list
```
uuidblastramp = Base.UUID("8e850b90-86db-534c-a0d3-1478176c7d93")
delete!.(Pkg.Types.get_last_stdlibs.(julia_versions), uuidblastramp)
```
which is weird because the entries should be correct, at least once `HistoricalStdlibVersions` is loaded:
```
$ julia-1.7.3 --project=../../.ci -q
julia> Pkg.Types.get_last_stdlibs(v"1.10.0")[Base.UUID("8e850b90-86db-534c-a0d3-1478176c7d93")]
("libblastrampoline_jll", v"3.0.4+0")

julia> using BinaryBuilder

julia> Pkg.Types.get_last_stdlibs(v"1.10.0")[Base.UUID("8e850b90-86db-534c-a0d3-1478176c7d93")]
("libblastrampoline_jll", v"5.8.0+1")
```

So I am not sure why this helps but it does seem to work.


_Edit:_ 
- This does seem wrong as we are getting libblastrampoline 5.12 now for julia 1.10.
- In this case this would not hurt since libpolymake_julia does not link with libblastrampoline (it is only required by some other libraries in `polymake_jll` which are not used for linking).

I will try looking for a different way to work around this.

benlorenz:
> I will try looking for a different way to work around this.

I could not get BinaryBuilder/Pkg to install the correct libblastrampoline versions (depending on the julia versions). But with the `delete!` and an extra `BuildDependency` we can make sure to use the same version (5.4.0) that `Iptopt_jll` is using.
Not pretty but it seems to work.

fingolfin (approved):
Looks good to me. Of course the blastrampoline hack is a hack, but you gotta do what you gotta do. And it is commented

lgoettgens:
Do you want to get this in before or after a libjulia rebuild for 1.12.0-rc1 (https://github.com/JuliaPackaging/Yggdrasil/pull/11630)?

benlorenz:
> Do you want to get this in before or after a libjulia rebuild for 1.12.0-rc1 (#11630)?

It's not urgent, so I guess we could wait until the new libjulia (and probably another libcxxwrap rebuild).

lgoettgens:
I expected a rc release shortly after https://github.com/JuliaLang/julia/pull/58655 was merged, but now there is https://github.com/JuliaLang/julia/pull/58956 (another backport PR). Thus I would say that you have enough time to let this dripple downstream until the rc gets released, and we can already benefit from less build jobs when eventually rebuilding with a new libcxxwrap

lgoettgens:
Now there indeed is https://github.com/JuliaLang/julia/pull/58985 to release 1.12.0-rc1.

PR #11610, opened 2025-07-06
[MPFR@4.1.1] Rebuild after recipe splitup

lgoettgens:
See https://github.com/JuliaPackaging/Yggdrasil/pull/11576#issuecomment-3022771553. In particular, this should add a riscv build.

Also add a version overview as asked for by @imciner2 in https://github.com/JuliaPackaging/Yggdrasil/pull/11576#pullrequestreview-2972440476

cc @giordano

### DIFF
--- a/M/MPFR/common.jl
+++ b/M/MPFR/common.jl
@@ -2,6 +2,12 @@ using BinaryBuilder
 using BinaryBuilderBase: sanitize
 using Pkg
 
+# Version overview:
+# - 4.1.1: Bundled with Julia 1.6, 1.7, 1.8, 1.9
+# - 4.2.0: Bundled with Julia 1.10
+# - 4.2.1: Bundled with Julia 1.11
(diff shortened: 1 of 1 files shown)


fingolfin:
This and the sibling PRs seem sensible to me. I am tempted to merge it but since @giordano was explicitly tagged I'll wait a bit in case he is aware of discussions or complications that I am not.

giordano:
Do we actually need to rebuild all of them?

lgoettgens:
> Do we actually need to rebuild all of them?

At least the ones that haven't been build for all platforms (so at least 4.1.1 and 4.2.0, I would need to check for the other two).
But is there any downside to just doing it? The cpu time is already spent, and I see no downside in just doing it

lgoettgens:
The registration PR has registry inconsistency failures... *sigh*
https://github.com/JuliaRegistries/General/pull/134392

PR #11612, opened 2025-07-06
[MPFR@4.2.1] Rebuild after recipe splitup

lgoettgens:
See https://github.com/JuliaPackaging/Yggdrasil/pull/11576#issuecomment-3022771553. 

cc @giordano


lgoettgens:
This seems to be identical to the current build, thus not needed

PR #11628, opened 2025-07-09
Libwebsockets builded with SOCKS5 support

stepanplaunov:
Hello! Added an LWS_WITH_SOCKS5 argument to the jll build to enable Libwebsockets to work with SOCKS5.

LWS docs reference: https://github.com/warmcat/libwebsockets/blob/main/lib/secure-streams/README.md?plain=1#L831

### DIFF
--- a/L/libwebsockets/build_tarballs.jl
+++ b/L/libwebsockets/build_tarballs.jl
@@ -21,7 +21,8 @@ cmake -B build \
     -DZLIB_INCLUDE_DIR=${includedir} \
     -DLWS_WITH_ACCESS_LOG=ON \
     -DLWS_WITHOUT_EXTENSIONS=OFF \
-    -DLWS_WITHOUT_TESTAPPS=ON
+    -DLWS_WITHOUT_TESTAPPS=ON \
+    -DLWS_WITH_SOCKS5=ON
 cmake --build build --parallel ${nproc}
 cmake --install build
 """ 


L/libwebsockets/build_tarballs.jl
```diff
@@ -21,14 +21,17 @@ cmake -B build \
...
+    -DLWS_WITH_SOCKS5=ON
 cmake --build build --parallel ${nproc}
 cmake --install build
 """ 
 
 # These are the platforms we will build for by default, unless further
 # platforms are passed in on the command line
 platforms = supported_platforms(exclude = Sys.iswindows)
+# `OpenSSL_jll` is missing
+filter!(p -> arch(p) != "riscv64", platforms)
```
giordano: OpenSSL is available if you use version 3.0.15 below

giordano (commented):
Thanks for your contribution!  Please update to a newer upstream version as well.

L/libwebsockets/build_tarballs.jl
stepanplaunov: Upgraded OpenSSL_jll to the latest version 3.5.1 (3.0.15 have same issues)

L/libwebsockets/build_tarballs.jl
```diff
@@ -37,7 +38,7 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency("OpenSSL_jll"; compat="3.0.8"),
+    Dependency("OpenSSL_jll"; compat="3.5.1"),
```
giordano: 3.5.1 won't work in julia v1.12.  I think I was wrong, it was
```suggestion
    Dependency("OpenSSL_jll"; compat="3.0.16"),
```
not 3.0.15

stepanplaunov:
> This still needs to address
> 
> > Please update to a newer upstream version as well.

Version updated

L/libwebsockets/build_tarballs.jl
```diff
@@ -3,7 +3,7 @@
 using BinaryBuilder, Pkg
 
 name = "libwebsockets"
-version = v"4.3.3"
+version = v"4.3.4"
```
giordano, line 6: Need to change also the source, to point to the commit of the upstream v4.3.4

stepanplaunov: Now i changed the commit

PR #11641, opened 2025-07-10
[SCOTCH] Upgrade to v7.0.8

amontoison:
Update: I need to modify the patches.


fingolfin:
Marking this as draft since it needs work. Just mark it as "ready for release" once you are ready. Assuming you still plan to update it?

amontoison:
It needs to update at least one patch at each release, which I didn't do yet. It is a similar issue with PT-SCOTCH.

ViralBShah:
SCOTCH_jll is at 7.0.11

PR #11645, opened 2025-07-11
New recipe: git-lfs v3.7.0

cgarling:
Would like to hook this up with Git.jl so the git executable provided there calls the git-lfs we build here when we issue commands like ```run(`$(Git.git()) lfs ...`)``` -- not sure how to accomplish that yet ...

### DIFF
--- a/G/Git_LFS/build_tarballs.jl
+++ b/G/Git_LFS/build_tarballs.jl
@@ -1,0 +1,39 @@
+using BinaryBuilder
+
+name    = "GitLFS"
+version = v"3.7.0"
+
+sources = [
+    GitSource("https://github.com/git-lfs/git-lfs.git", "92dddf560e62ef7dd25877d87ce072f7595aa52d")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+# echo $GOPATH # /workspace/.gopath
+cd $WORKSPACE/srcdir/git-lfs/
+mkdir -p ${bindir}
+
+# Install goversioninfo if host is Windows
+case "$MACHTYPE" in
+  *mingw*|*cygwin*) go install github.com/josephspurrier/goversioninfo/cmd/goversioninfo ;;
+esac
+export PATH="$GOPATH/bin:$PATH"
+
+go build -o ${bindir}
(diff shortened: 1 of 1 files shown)


G/Git_LFS/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+  *mingw*|*cygwin*) go install github.com/josephspurrier/goversioninfo/cmd/goversioninfo ;;
+esac
+export PATH="$GOPATH/bin:$PATH"
+
+go build -o ${bindir}
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms(;experimental=true)
```
giordano:
```suggestion
platforms = supported_platforms()
```

G/Git_LFS/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+products = [
+    ExecutableProduct("git-lfs", :git_lfs),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; compilers=[:c, :go], julia_compat="1.6")
```
giordano:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; compilers=[:c, :go], julia_compat="1.6")

```

giordano (commented):
> not sure how to accomplish that yet

I don't know how git lfs works, but my guess would be that `git` searches for `git-lfs` in `PATH`, so you'd need to add `git-lfs`'s directory to the `PATH` in `Git.git()`'s environment, similar to https://github.com/JuliaVersionControl/Git.jl/pull/55.  If my guess was right, of course, which I don't know.

cgarling:
Thanks for the advice and the review! Will look into making a PR to Git.jl to add git-lfs support.

giordano:
Thanks for your contribution!

PR #11655, opened 2025-07-14
[libcxxwrap-julia] Rebuild against latest libjulia

lgoettgens:
cc @fingolfin @barche @benlorenz


fingolfin (approved):

fingolfin:
The one drawback of keeping the version the same that I can think of is that we can't force an upgrade for people who used a beta version of 1.12 before -- but those people should know to `Pkg.update` / we can tell them. So it seems fine to me

PR #11658, opened 2025-07-14
[libpolymake_julia] rebuild with new libjulia for older Oscar versions

benlorenz:
Since #11607 will be for future Oscar versions only, this rebuild would help get julia 1.12 working with Oscar 1.4.

cc: @fingolfin @lgoettgens

### DIFF
--- a/L/libpolymake_julia/build_tarballs.jl
+++ b/L/libpolymake_julia/build_tarballs.jl
@@ -16,7 +16,7 @@ delete!(Pkg.Types.get_last_stdlibs(v"1.13.0"), uuidopenssl)
 include("../../L/libjulia/common.jl")
 
 name = "libpolymake_julia"
-version = v"0.13.4"
+version = v"0.13.5"
 
 # reminder: change the above version when changing the supported julia versions
 # julia_versions is now taken from libjulia/common.jl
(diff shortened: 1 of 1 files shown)


lgoettgens:
You don't need to bump the version here as you just change a build dependency, see https://github.com/JuliaPackaging/Yggdrasil/pull/11553 and https://github.com/JuliaPackaging/Yggdrasil/pull/11561.
If you don't change the version numbers, you don't need to do another Polymake.jl release just to update the provided jll version.

benlorenz:
> You don't need to bump the version here as you just change a build dependency, see #11553 and #11561. If you don't change the version numbers, you don't need to do another Polymake.jl release just to update the provided jll version.

The compat in Polymake.jl allows a patch level bump without a new release.
With the version bumps it is slightly cleaner as it makes sure `libpolymake_julia` and `polymake_oscarnumber` are matching, and disallows a mix between an updated and an old version.

lgoettgens:
> > You don't need to bump the version here as you just change a build dependency, see #11553 and #11561. If you don't change the version numbers, you don't need to do another Polymake.jl release just to update the provided jll version.
> 
> The compat in Polymake.jl allows a patch level bump without a new release. With the version bumps it is slightly cleaner as it makes sure `libpolymake_julia` and `polymake_oscarnumber` are matching, and disallows a mix between an updated and an old version.

In my memory, polymake_oscarnumber_jll was specified with a `=version` qualifier. But that's not the case, so just ignore everything I wrote

PR #11663, opened 2025-07-15
0_RootFS bootstrap: Build libstdc++ to be compatible with both TLS and non-TLS ABI

Keno:
As reported in https://github.com/JuliaLang/julia/issues/57021, our Yggdrasil-built binaries no longer work out-of-the box with the msys2-provided mingw toolchain. This is not a good situation, because it basically means that partial BinaryBuilder-provided builds are off the table. The root cause of the issue is that at some point msys2 started building gcc (and libstdc++) with TLS support. Enabling TLS is an ABI-breaking change, because it removes symbols from libstdc++ that were referenced in the header. That said, I don't believe it is necessary for this to be ABI breaking. In particular, I think the included patch provides a correct ABI-compability implementation that supports applications linked against either the TLS enabled or disabled version. The net effect of this change would be that (after this is bumped on CSL) the new binaries we build would have TLS enabled and work on the new msys64 ABI as well as the new CSL, which any old binaries could continue to work on either the new CSL or the old CSL. Notably, new binaries would not work on the old CSL, so once we bump CSL, we'll have to put in new JLL compats for any newly built JLLs.

Note that this only changes the libstdc++ for GCC@14, which is the one we bundle in CSL. However, it does change the TLS flag for all versions of GCC, so that TLS will become enabled even when the new binaries are built with older GCC.

### DIFF
--- a/0_RootFS/GCCBootstrap@14/bundled/patches/gcc/gcc1420-libstdcpp-tls-abi.patch
+++ b/0_RootFS/GCCBootstrap@14/bundled/patches/gcc/gcc1420-libstdcpp-tls-abi.patch
@@ -1,0 +1,166 @@
+From 648157c34413e7ecbbd9dffe871617042e46e7e9 Mon Sep 17 00:00:00 2001
+From: Keno Fischer <keno@juliahub.com>
+Date: Tue, 15 Jul 2025 02:47:01 +0000
+Subject: [PATCH] libstdc++: Provide ABI compatibility when turning on
+ _GLIBCXX_HAVE_TLS
+
+The _GLIBCXX_HAVE_TLS flag is an ABI-breaking flag - setting it removes the
+definitions for `set_lock_ptr` (as well as the comptability symbols for
+even older ABIs). Try to improve this situation by changing the flag to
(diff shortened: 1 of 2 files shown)


fingolfin:
Apparently there was a bit of discussion on Slack. And @gbaraldi might need / want these changes so perhaps he can do a review?


I have a question:

> Notably, new binaries would not work on the old CSL, so once we bump CSL, we'll have to put in new JLL compats for any newly built JLLs.

Could you elaborate what that means? Is this something we'd do automatically (were?) or something that would require updating build recipes (but which ones? how do we find them?)

fingolfin:
Aha and @giordano points out to me that the patch was [proposed upstream](https://gcc.gnu.org/pipermail/gcc-patches/2025-October/697842.html), was meant to be reviewed there, but that did not yet happen.

fingolfin:
Seems @Keno [pinged them October 27](https://gcc.gnu.org/pipermail/gcc-patches/2025-October/698760.html) but unfortunately still no reaction :-(

Keno:
I'll ping them again, but I'm gonna go ahead and merge this - there are more ABI issues, so I think we just need to start taking ownership of this.

PR #11664, opened 2025-07-15
0_RootFS: Update MinGW sources to 14.0.0

### LINKED ISSUE #13445: Possible to update MinGW to verision 14?
I'm trying to build SimpleBLE for windows which requires header files in MinGW version 14 but Mingw_jll is only version 11.

I would like to try getting it to work but this [build_tarbals.jl](https://github.com/JuliaPackaging/Yggdrasil/blob/fcc59ab37948c3fb8bc689388d3e11630774c4c9/M/Mingw/build_tarballs.jl) is only building MinGW version 5 so I can't find out how the current version 11 was built.

Can anyone help? I'm way out of my depth but I'm willing to try if I have the latest build script.

Keno:
As noted in https://github.com/JuliaLang/julia/issues/56840, our CSL-distributed msvcrt is no longer compatible with mingw as distributed in msys2. This is because https://github.com/mingw-w64/mingw-w64/commit/d0cad4c27090246be25ecd4fddaefd71ba8c27af added a call to _initterm_e but the corresponding msvcrt.a emulation was added in https://github.com/mingw-w64/mingw-w64/commit/6c89679ccc51ea4f218683e61943881a7038a1c9 which is newer than what we ship in CSL.

That said, it is not clear to me why we ship msvcrt.a from mingw in CSL at all. It makes sense to ship gnd dlls, but if we're not shipping a toolchain it's odd to ship msvcrt.a. After all, we're not shipping glibc either.

### DIFF
--- a/0_RootFS/gcc_sources.jl
+++ b/0_RootFS/gcc_sources.jl
@@ -312,8 +312,8 @@ function gcc_sources(gcc_version::VersionNumber, compiler_target::Platform; kwar
         end
     elseif Sys.iswindows(compiler_target)
         libc_sources = [
-            ArchiveSource("https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v11.0.1.tar.bz2",
-                          "3f66bce069ee8bed7439a1a13da7cb91a5e67ea6170f21317ac7f5794625ee10"),
+            ArchiveSource("https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v13.0.0.tar.bz2",
+                          "5afe822af5c4edbf67daaf45eec61d538f49eef6b19524de64897c6b95828caf"),
         ]
     else
         error("Unknown libc mapping for platform $(compiler_target)")


Keno:
> It makes sense to ship gnd dlls, but if we're not shipping a toolchain it's odd to ship msvcrt.a.

To answer my own question, we link this for pkgimage creation: https://github.com/JuliaLang/julia/blob/master/base/linking.jl#L143. I guess we're not linking dllcrt, because we know that we don't have any initializers that we expect the C runtime to run?

ViralBShah:
Ok to merge?

giordano:
Haven't tested it, but looks reasonable.

giordano:
I'll try to work on this in the coming weeks (can't promise much though), we should take this as an occasion to fix #4067 as well because that's a mess.

ViralBShah:
Is it ok to merge this since #4067 seems to be fixed - or does it need to be redone?

giordano:
This still needs to be done still, it's orthogonal to #4067.

0_RootFS/gcc_sources.jl
```diff
@@ -312,8 +312,8 @@ function gcc_sources(gcc_version::VersionNumber, compiler_target::Platform; kwar
         end
     elseif Sys.iswindows(compiler_target)
         libc_sources = [
-            ArchiveSource("https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v11.0.1.tar.bz2",
-                          "3f66bce069ee8bed7439a1a13da7cb91a5e67ea6170f21317ac7f5794625ee10"),
+            ArchiveSource("https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v13.0.0.tar.bz2",
+                          "5afe822af5c4edbf67daaf45eec61d538f49eef6b19524de64897c6b95828caf"),
```
fingolfin, line 316:
```suggestion
            ArchiveSource("https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v14.0.0.tar.bz2",
                          "6eaf921d9eb987d3820b364ea9775bc19b965ec81490b6fdd716526c28e1995c"),
```

Keno:
I'll merge this. Our rebuilt base windows runners have newer mingw and fail to build without this.

0_RootFS/gcc_sources.jl
Keno: Sorry, I missed that this hadn't gotten applied. Since i've already built v13 now, let me push that through to make sure everything is working and we can do a v14 bump separately.

PR #11669, opened 2025-07-15
[GAP_pkg_juliainterface] Rebuild v0.1300.103 against new libjulia

lgoettgens:
The above is the version that Oscar v1.4.1 on julia 1.12.0-rc1 resolves to. With this rebuild, Oscar v1.4.1 should load again on julia 1.12 rc.
This PR resets the recipe back to https://github.com/lgoettgens/Yggdrasil/blob/86f0a6ac4aec5c17a88b1e05d13075e7e16d093b/G/GAP_pkg/GAP_pkg_juliainterface/build_tarballs.jl, with the single change of an updated `libjulia_jll` version number.

Immediately after this is merged, this PR should be reverted with `[skip ci]` and `[skip build]` so that we have the latest recipe on the yggy master again.

cc @fingolfin


lgoettgens:
Bump @fingolfin

PR #11672, opened 2025-07-15
Add more Quantum ESPRESSO executable products

Technici4n:
I noticed that `projwfc.x` was missing from the JLL. To avoid such issues in the future, I took all the products mentioned in https://www.quantum-espresso.org/documentation/input-data-description/ that were built with the current settings and added them.


Technici4n:
Thanks!

PR #11697, opened 2025-07-18
[GAP] Drop `jl_reinit_foreign_type` workaround for julia pre 1.10

lgoettgens:
I wouldn't want to release this right now, as the version change requires us to rebuild many other jlls as well, without a clear benefit. 
Instead, when at some point we do a minor change anyway (or e.g. something that changes the julia compat), we should then include this patch as well.

cc @fingolfin


fingolfin (approved):

lgoettgens:
@fingolfin can we expect a gap release that would warrant a minor version bump of the jll in the next few weeks? If yes, I would wait with merging this PR here until that is there. If not, I would say we should get this in, such that this and all of the required rebuilds of dependent jlls can still make it into GAP.jl 0.15.0

PR #11699, opened 2025-07-19
Build LibAMVW for all platforms

andreasnoack:


giordano:
Changing compat bounds requires changing the version number too (it's a requirement from Pkg).  Ideally if there's a newer upstream version you'd use that one.

andreasnoack:
There is not a new version. I only changed the required version because of the error in https://buildkite.com/julialang/yggdrasil/builds/22056/steps/canvas?sid=01982388-87c8-4da9-b860-2caa6001ab4f. I thought that experimental platforms were disabled by default so not sure why I got that error. Is there a way to exclude experimental versions such that I don't have to bump the required Julia version?

giordano:
> Is there a way to exclude experimental versions such that I don't have to bump the required Julia version?

There is, but there are also other (implicit) compat bounds which changed since 2020, so a change of the version number of the package is necessary anyway.

andreasnoack:
https://github.com/JuliaPackaging/Yggdrasil/pull/11701 supersedes this one. The source code was made part of eiscor so that library is essentially the new version.

PR #11701, opened 2025-07-19
Wizard recipe: Eiscor-v0.2.0

andreasnoack:

* Package name: Eiscor
* Version: v0.2.0

### DIFF
--- a/E/Eiscor/build_tarballs.jl
+++ b/E/Eiscor/build_tarballs.jl
@@ -13,7 +13,7 @@ sources = [
 # Bash recipe for building across all platforms
 script = raw"""
 cd $WORKSPACE/srcdir/eiscor/
-make
+make "-j$nproc"
 make LIBDIR="${libdir}" install
 """
 


E/Eiscor/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/eiscor/eiscor.git", "696b90afcd8d637d080fe29de5c3c799bf5eb8dd")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd eiscor-0.2.0/
+export HOME=$libdir
```
andreasnoack: This is just to set the install directory. See https://github.com/eiscor/eiscor/blob/849f56c64eb79a4fd75c46c9d2591045a6d5c889/make.inc#L8

andreasnoack:
Looks like there is a connectivity issue so please trigger a retry.

E/Eiscor/build_tarballs.jl
staticfloat: I think you should be able to instead do `make INSTALLDIR=$libdir ...`

E/Eiscor/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/eiscor/eiscor.git", "696b90afcd8d637d080fe29de5c3c799bf5eb8dd")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd eiscor-0.2.0/
```
staticfloat:
```suggestion
cd eiscor*/
```

To make this work for either tarballs or git checkouts

andreasnoack:
I could use some help here. What is the issue with the Mac and Windows builds? It looks like the check can find the product initially but then eventually concludes that
```julia
┌ Error: Built Eiscor but libeiscor still unsatisfied:
--
  | └ @ BinaryBuilder /cache/julia-buildkite-plugin/depots/e2fd9734-29d8-45cd-b0eb-59f7104f3131/packages/BinaryBuilder/3eUfg/src/AutoBuild.jl:987
  | ERROR: LoadError: Cannot continue with unsatisfied build products!
```
Why would that be?

giordano:
Wrong file extension.

E/Eiscor/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/eiscor/eiscor.git", "696b90afcd8d637d080fe29de5c3c799bf5eb8dd")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd eiscor*/
+make "-j$nproc"
```
giordano, line 16:
```suggestion
make -j${nproc} SLIB="${dlext}"
```

E/Eiscor/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+make INSTALLDIR="$libdir" install
```
giordano:
```suggestion
make INSTALLDIR="${libdir}" SLIB="${dlext}" install
```

E/Eiscor/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/eiscor/eiscor.git", "696b90afcd8d637d080fe29de5c3c799bf5eb8dd")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd eiscor*/
```
giordano:
```suggestion
```

E/Eiscor/build_tarballs.jl
```diff
@@ -0,0 +1,39 @@
...
+version = v"0.2.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/eiscor/eiscor.git", "696b90afcd8d637d080fe29de5c3c799bf5eb8dd")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
```
giordano:
```suggestion
cd $WORKSPACE/srcdir/eiscor*/
```

andreasnoack:
Thanks. Trying that now. I expected that https://github.com/eiscor/eiscor/blob/v0.2.0/.master.inc would have handled the extension.

andreasnoack:
There still seems to be an issue with Windows and macOS.

giordano:
Probably still because of the name. Putting version number after the extension isn't standard for non-ELF files, and the last bit of the version number would be detected as the extension.

andreasnoack:
Thanks. I see. I'll try to patch it.

giordano:
Specifically, you can see what's going on by reading the documentation: https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct.  For example, on Windows the file is called `libeiscor.dll.0.2.0`:
```julia-repl
julia> Base.BinaryPlatforms.parse_dl_name_version("libeiscor.dll.0.2.0", "windows")
ERROR: ArgumentError: Invalid dynamic library path 'libeiscor.dll.0.2.0'
Stacktrace:
 [1] parse_dl_name_version(path::String, os::String)
   @ Base.BinaryPlatforms ./binaryplatforms.jl:812
 [2] top-level scope
   @ REPL[31]:1
```

andreasnoack:
It looks like I finally managed to get this working. It is unclear to me why just `x86_64-linux-gnu-libgfortran4` would be failing here when all other targets are passing. @giordano any ideas to why this could be?

ViralBShah:
This looks funny, but why would it only mess up here:

https://buildkite.com/julialang/yggdrasil/builds/22094/steps/table?jid=0198307e-1306-4d06-b540-d69432ca0f17#0198307e-1306-4d06-b540-d69432ca0f17/691-943

andreasnoack:
I think we are good to go here now. After looking more carefully, the question shouldn't be why only the single job failed, but why the other ones passed. The symlinking was completely broken. Should be fixed now, though. @giordano anything else to do here or can this one be merged?

ViralBShah:
Worth re-enabling bsd and apple to try again?

andreasnoack:
Apple is already enabled here. I think we can wait for an interested BSD user to show up before enabling it here. I have no experience with BSD (except macOS) and I have no way of testing that the libraries actually works.

ViralBShah:
Oh I see the filter rule now - which makes apple go through. 

I think in general with BB, we build all the platforms that build fine. My suggestion would be that if freebsd builds (without needing any effort or attention), we should enable it. I would say that the majority of freebsd builds on BB do not get tested.

andreasnoack:
The library isn't using a build system that would BSD automatically (I guess). It was just a very basic makefile setup that produced wrongly named libraries for Windows and macOS. I tried to patch it based on Perplexity's suggestions. Can I just expand the Linux branch to also cover BSDs?

ViralBShah:
In that case let us just merge.

andreasnoack:
Is anything else needed here for this to get merged?

PR #11711, opened 2025-07-22
oneAPI: Bump headers to v1.22.4

michel2323:


michel2323:
@maleadt

michel2323:
@maleadt Like this?

PR #11714, opened 2025-07-23
  [openfhe_julia] and [openfhe_julia_int128] update to version v0.3.10 (2nd attempt)

sloede:


sloede:
@giordano This is not a solution to the root cause, but I hope we can proceed anyways.

I've documented the CMake issue in the [upstream repo](https://github.com/hpsc-lab/OpenFHE.jl/issues/86) for a proper fix of the wrong library/include paths.

Why the local build requires setting `-L$libdir` explicitly in the `LDFLAGS` is beyond my current understanding (I thought that this would already be set by specifying the `$CMAKE_TARGET_TOOLCHAIN` for CMake, but apparently it is not).

cc @ArseniyKholod

giordano:
This is _bad_: https://github.com/openfheorg/openfhe-development/blob/32362085fcfa8133f86a4e81d4440734c984b16d/CMakeLists.txt#L404-L455. Hardcoding absolute paths on the developers' machine, ignoring the possibility downstream users will have different setups which would work just fine if they didn't do this stuff, is a terrible practice: by setting those options they're forcing [`FindOpenMP`](https://cmake.org/cmake/help/latest/module/FindOpenMP.html) to look into the wrong place. This may have worked before because by accident libdir happened to be in the search path for cmake, but if that's not the case cmake has no way to guess where openmp is because the build system is forcibly giving it completely wrong information.

sloede:
> This is _bad_: https://github.com/openfheorg/openfhe-development/blob/32362085fcfa8133f86a4e81d4440734c984b16d/CMakeLists.txt#L404-L455. Hardcoding absolute paths on the developers' machine, ignoring the possibility downstream users will have different setups which would work just fine if they didn't do this stuff, is a terrible practice: by setting those options they're forcing [`FindOpenMP`](https://cmake.org/cmake/help/latest/module/FindOpenMP.html) to look into the wrong place. This may have worked before because by accident libdir happened to be in the search path for cmake, but if that's not the case cmake has no way to guess where openmp is because the build system is forcibly giving it completely wrong information.

I agree - this is not good SE practice :-/

PR #11716, opened 2025-07-23
Initial Agama recipe

langfzac:
Recipe for [Agama](https://github.com/GalacticDynamics-Oxford/Agama)

### DIFF
--- a/A/Agama/build_tarballs.jl
+++ b/A/Agama/build_tarballs.jl
@@ -16,7 +16,7 @@ products = [
 ]
 
 dependencies = [
-    Dependency("GSL_jll"),
+    Dependency("GSL_jll", compat="~2.8.1"),
     Dependency("LLVMOpenMP_jll"),
     Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
 ]


A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
+using BinaryBuilder, Pkg
+
+name = "Agama"
+repo = "https://github.com/GalacticDynamics-Oxford/Agama.git"
+
+version = v"1.0"
```
fingolfin, line 6: Where is this version coming from? The repository in question doesn't seem to carry any version. 

Food for thought: perhaps it would be better to start out with something like 0.1, in case they start using a version?

Also, I wonder if you have asked them if they would be willing to set a version?

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
...
+    GitSource(repo, "f392b9d6e59269d451b37b3ada2508a44bd3a80c"),
+]
+
+platforms = [
+    Platform("aarch64", "macos"),
+    Platform("x86_64", "linux")
+]
+
+products = [
+    LibraryProduct("agama.so", :libagama),
```
fingolfin:
```suggestion
    LibraryProduct("agama", :libagama),
```

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,60 @@
...
+
+version = v"1.0"
+sources = [
+    GitSource(repo, "f392b9d6e59269d451b37b3ada2508a44bd3a80c"),
+]
+
+platforms = [
+    Platform("aarch64", "macos"),
+    Platform("x86_64", "linux")
+]
```
fingolfin: Why so restrictive? Have you tried building for more platforms?

A/Agama/build_tarballs.jl
langfzac: The version is taken from the `setup.py` file -- they use v1.0.

A/Agama/build_tarballs.jl
langfzac: This was just so I could test the build more quickly. I'll change this to `supported_platforms()` once I figure out why the current build is failing.

A/Agama/build_tarballs.jl
langfzac: But, this is the python package version, not the c library version. So point taken. I could see a reason for matching the python version number, since everything is shipped together.

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+else
+LINK_FLAGS_EXE_SHARED += -Wl,-rpath,'\$\$ORIGIN'
+endif 
+" > Makefile.local
+
+# Generate the library
+make -j${nproc} lib
+
+# Install
+cp agama.so ${libdir}/agama.so
```
giordano:
```suggestion
install -Dvm 755 libagama.${dlext} -t ${libdir}
```

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+
+# Handle macOS install
+if [ "$(uname)" == "Darwin" ]; then
+    mv ${libdir}/agama.so ${libdir}/agama.dylib
+fi
```
giordano:
```suggestion
```

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+FC = gfortran
+endif
+LINK = \$(CXX)
+
+COMPILE_FLAGS_ALL += -fPIC -fopenmp -Wall -O2 
+LINK_FLAGS_ALL += -fopenmp
+COMPILE_FLAGS_ALL += -std=c++11
+
+COMPILE_FLAGS_LIB += -I${prefix}/include
+LINK_FLAGS_LIB_AND_EXE_STATIC += -L${prefix}/lib -lgsl -lgslcblas
```
giordano:
```suggestion
LINK_FLAGS_LIB_AND_EXE_STATIC += -L${libdir} -lgsl -lgslcblas
```

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+ifeq ($(origin FC),default)
+FC = gfortran
+endif
+LINK = \$(CXX)
+
+COMPILE_FLAGS_ALL += -fPIC -fopenmp -Wall -O2 
+LINK_FLAGS_ALL += -fopenmp
+COMPILE_FLAGS_ALL += -std=c++11
+
+COMPILE_FLAGS_LIB += -I${prefix}/include
```
giordano:
```suggestion
COMPILE_FLAGS_LIB += -I${includedir}
```

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+LINK_FLAGS_LIB_AND_EXE_STATIC += -L${prefix}/lib -lgsl -lgslcblas
+
+ifeq (\$(shell uname -s),Darwin)
+LINK_FLAGS_LIB += -Wl,-install_name,@executable_path/agama.so
+else
+LINK_FLAGS_EXE_SHARED += -Wl,-rpath,'\$\$ORIGIN'
+endif
```
giordano: The audit should handle this stuff automatically for you

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+
+ifeq (\$(shell uname -s),Darwin)
+LINK_FLAGS_LIB += -Wl,-install_name,@executable_path/agama.so
+else
+LINK_FLAGS_EXE_SHARED += -Wl,-rpath,'\$\$ORIGIN'
+endif 
+" > Makefile.local
+
+# Generate the library
+make -j${nproc} lib
```
giordano:
```suggestion
make -j${nproc} "libagama.${dlext}" LIBNAME_SHARED="libagama.${dlext}"
```

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+    Dependency(PackageSpec(name="CompilerSupportLibraries_jll", uuid="e66e0078-7015-5450-92f7-15fbd957f2ae"))
+]
+
+script = raw"""
+cd $WORKSPACE/srcdir/Agama
+
+# Make the Makefile.local
+echo "
+ifeq ($(origin CXX),default)
+FC = g++
```
giordano:
```suggestion
FC = c++
```
but I'm very confused by this at all, claiming that the fortran compiler is `g++` sounds very fishy

A/Agama/build_tarballs.jl
langfzac: Will this work if the output from running `make lib` is `agama.so`? I don't have the ability to change that.

langfzac: Oh, I see. This is set later on. Ignore me.

A/Agama/build_tarballs.jl
langfzac: This might be a typo... I'm copying from `Makefile.local.template` in the main repo

langfzac: I've just removed it, and things seem to work fine.

A/Agama/build_tarballs.jl
langfzac: Ok, I've added all linux (glibc) and macos. Windows doesn't compile, and I'm not too concerned about support for windows at the moment.

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+name = "Agama"
+repo = "https://github.com/GalacticDynamics-Oxford/Agama.git"
+
+version = v"1.0"
+sources = [
+    GitSource(repo, "f392b9d6e59269d451b37b3ada2508a44bd3a80c"),
+]
+
+
+all_platforms = supported_platforms()
```
giordano:
```julia
supported_platforms(; exclude=Sys.iswindows)
```
is simpler than all of this.

langfzac: Ah ok. I also need to exclude the `musl` linux (which aren't compiling), can I do that in a similar way? Or would I need to write a little anon-func for that keyword?

giordano: That'd be one way to do that, but what's the problem with Musl?  We have ***very*** few packages which are incompatible with Musl, so that's usually surprising.

langfzac: It looks like the code uses `exeinfo.h`, which it sounds like is only available from glibc? The folks who will be using this package right now will have already been using the c library (or python package) , so I don't think this will be an issue short term. But, I'll prod the developer to add support.

A/Agama/build_tarballs.jl
```diff
@@ -0,0 +1,46 @@
...
+
+platforms = supported_platforms(;exclude=x->Sys.iswindows(x) || Sys.isfreebsd(x) || (libc(x) == "musl"))
+platforms = expand_cxxstring_abis(platforms)
+
+products = [
+    LibraryProduct("libagama", :libagama),
+]
+
+dependencies = [
+    Dependency("GSL_jll"),
```
giordano: GSL always breaks the ABI when they change the minor version, at least when they remember to do that, so specifying the compat bound is very important
```suggestion
    Dependency("GSL_jll"; compat="~2.8.1"),
```

giordano:
Thanks for your contribution!

PR #11719, opened 2025-07-24
oneAPI: Bump loader to v1.22.4

michel2323:


michel2323:
> This requires changing the version number

It got a bump in `common.jl` when bumping the headers.

imciner2:
> It got a bump in common.jl when bumping the headers.

I'd suggest rebasing this on the current master of Yggdrasil and then doing bumping the text in the build_tarballs to ensure it builds a fresh version. We'd like to make sure the build worked with the new headers (the last used the 1.13.0 version).

michel2323:
Like this?

amontoison (approved):

amontoison:
Related PR: https://github.com/JuliaPackaging/Yggdrasil/pull/11711
LGTM :+1:

O/oneAPI_Level_Zero/oneAPI_Level_Zero_Loader/build_tarballs.jl
```diff
@@ -52,3 +52,4 @@ dependencies = [
 
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
                preferred_gcc_version=v"8")
```
giordano, line 54: This is for a future version, but it'd be good to require julia v1.6, so that the dependency on Pkg will be automatically dropped.

PR #11733, opened 2025-07-28
Add Sundials v7.4.0 build recipe

ChrisRackauckas:
This PR adds a build recipe for SUNDIALS v7.4.0, the latest release from Lawrence Livermore National Laboratory (LLNL).

## What is SUNDIALS?

SUNDIALS (SUite of Nonlinear and DIfferential/ALgebraic equation Solvers) provides a suite of time integrators and nonlinear solvers including:
- CVODE - for stiff and nonstiff ODE systems
- CVODES - CVODE with sensitivity analysis capabilities  
- IDA - for differential-algebraic equation (DAE) systems
- IDAS - IDA with sensitivity analysis capabilities
- ARKODE - for additive Runge-Kutta methods
- KINSOL - for nonlinear algebraic systems

## Changes in this PR

- Added new build recipe for Sundials v7.4.0 (latest release as of June 2024)
- Included patches from v6 build that are still needed:
  - `SundialsSetupCompilers.patch` - for 64-bit integer support
  - `Sundials_windows.patch` - Windows compatibility fix
  - `Sundials_findklu_suffixes.patch` - KLU library detection on Windows
- Added new library products that were introduced in v7:
  - `libsundials_generic`
  - `libsundials_core`

## Version Information

- Previous version in Yggdrasil: v6.7.0
- New version: v7.4.0
- Release date: June 23, 2024
- Source: https://github.com/LLNL/sundials/releases/tag/v7.4.0

## Testing

The build recipe follows the same structure as the existing v6 recipe with necessary updates for v7. All platforms that were supported in v6 should continue to work with v7.

## Related Issues/PRs

This addresses the need for Sundials v7 support in the Julia ecosystem, enabling packages that depend on Sundials to use the latest features and improvements.

cc @ChrisRackauckas


S/Sundials/Sundials@7/build_tarballs.jl
```diff
@@ -0,0 +1,100 @@
...
+]
+
+dependencies = [
+    Dependency("CompilerSupportLibraries_jll"),
+    Dependency("libblastrampoline_jll"),
+    Dependency("SuiteSparse_jll"),
+]
+
+# Build the tarballs
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"6", julia_compat="1.11")
```
ChrisRackauckas:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"6", julia_compat="1.10")
```

ChrisRackauckas:
I'm not sure what's making those specific architectures fail.

S/Sundials/Sundials@7/build_tarballs.jl
```diff
@@ -0,0 +1,100 @@
...
+]
+
+dependencies = [
+    Dependency("CompilerSupportLibraries_jll"),
+    Dependency("libblastrampoline_jll"),
+    Dependency("SuiteSparse_jll"),
+]
+
+# Build the tarballs
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"6", julia_compat="1.10")
```
lgoettgens:
```suggestion
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; preferred_gcc_version = v"6", julia_compat="1.10")

```
trailing newline is missing

lgoettgens:
> I'm not sure what's making those specific architectures fail.

I think all failing runs were on `amdci7`. And according to the description on the binarybuilder channel on slack, its filesystem is somewhat corrupt. So maybe just try to restart the jobs and hope for a different runner. (cc @giordano )

giordano:
Failure to apply patch on windows looks genuine. And Yggdrasil runs only on amdci7, there aren't different runners.

ChrisRackauckas:
Ahh I didn't catch that the theme was windows, thanks.

ChrisRackauckas:
Ehh couldn't figure out the windows part

ViralBShah:
The AI couldn't figure out how to update the Windows patch (and maybe it isn't even needed)

ChrisRackauckas:
Yeah I know from reading the release notes that the fortran interface is decoupled now, which means the name mangling doesn't need to be there. I don't know what that means for the windows patch though or if it means anything at all.

ViralBShah:
The patch is trivial - I'll try to recreate the environment and fix it. Shouldn't be too hard. But I thought this is the sort of thing AI should have known how to do.

It is possible that the patch isn't even needed anymore.

The other thing to do carefully is to use the right version of SuiteSparse that is included in Julia (but it probably already does that).

giordano:
One would hope so, but no: https://github.com/JuliaLang/julia/pull/58628#discussion_r2125088829, https://github.com/JuliaLang/julia/pull/58628#discussion_r2125092176, https://github.com/JuliaLang/julia/pull/58628#issuecomment-2945921323. It got that patch wrong multiple times in different ways.

S/Sundials/Sundials@7/build_tarballs.jl
```diff
@@ -0,0 +1,100 @@
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_TOOLCHAIN_FILE="${CMAKE_TARGET_TOOLCHAIN}" \
+    -DEXAMPLES_ENABLE_C=OFF \
+    -DENABLE_KLU=ON \
+    -DKLU_INCLUDE_DIR="${includedir}/suitesparse" \
+    -DKLU_LIBRARY_DIR="${libdir}" \
+    -DKLU_WORKS=ON \
+    -DENABLE_LAPACK=ON \
+    -DBLAS_LIBRARIES="${BLAS_NAME}" \
+    -DLAPACK_LIBRARIES="${LAPACK_NAME}" \
```
imciner2, line 46: Setting these variables directly is actually not supported by CMake (https://gitlab.kitware.com/cmake/cmake/-/issues/25913#note_1511382).

Instead, the build script should remove the image CMake version (`apk del cmake`) and then add a `HostBuildDependency("CMake_jll")` and use `-DBLA_VENDOR="libblastrampoline"` in the CMake command instead.

imciner2 (changes requested):

S/Sundials/Sundials@7/build_tarballs.jl
ViralBShah: We follow what the Sundials docs say: https://sundials.readthedocs.io/en/latest/sundials/Install_link.html

giordano: They probably don't know libblastrampoline and the fact that cmake now supports it.

imciner2: The Sundials docs are fairly generic, and they actually seem to suggest just leaving all those variables blank and letting CMake auto-detect the library (they don't preset any of those variables). Their docs should really be suggesting that people use the `BLA_VENDOR` CMake flag, because upstream CMake has said that overwriting those variables is actually a bad idea (we did see problems doing this in a Ygg recipe a while ago, hence why I added LBT to CMake upstream).

ViralBShah: Ok, so should I leave these existing variables as they are and also just add `BLA_VENDOR`?

ViralBShah:
@giordano @imciner2 The failing builds all are unable to find `cmake` which is a `HostBuildDependency`.  How do we fix that?

When running locally, there's no issue finding the `cmake`.

giordano:
That should be in PATH automatically.  Might be a filesystem issue.

ViralBShah:
Re-trying the failed builds got them to pass. This should be good to merge when the builders are fixed/rebooted.

PR #11736, opened 2025-07-29
[jlqml] Version 0.7.1

barche:
For improved Makie support.


giordano:
Is supporting julia v1.6 necessary at this point?  Any chance to require v1.10 instead and greatly reduce the number of artifacts (and the time required to build all of them)?

barche:
No, good catch. QML.jl itself requires Julia 1.10 so it makes no sense to build for 1.6. I assume I can patch this without rebuilding?

giordano:
No, this will trigger a rebuild because changing the content of the file invalidates the full build cache.  But the ccache should be warm now, so the actual compilation process should be faster.

PR #11740, opened 2025-07-29
Fix libblastrampoline julia compat

IanButterworth:


giordano:
I think we need the same for v5.13.1

PR #11753, opened 2025-07-31
[ReferenceBLAS] Upgrade to v3.12.1 (LP64 and ILP64)

amontoison:


amontoison:
@imciner2 I checked the artifacts with `nm -D` and we have both LP64 and ILP64 symbols when `INDEX64=ON`. Is it expected?!

imciner2:
> @imciner2 I checked the artifacts with `nm -D` and we have both LP64 and ILP64 symbols when `INDEX64=ON`. Is it expected?!

Based on the logic in the upstream CMake here https://github.com/Reference-LAPACK/lapack/blob/bfbef596de9c445075b809f127e85c379884e209/SRC/CMakeLists.txt#L546, it appears that the library is supposed to contain both the LP64 and ILP64 symbols when the `INDEX64` option is given. I don't see a reason why that would necessarily cause a problem either.

R/ReferenceBLAS/common.jl
```diff
@@ -16,297 +15,40 @@ function blas_script(;blas32::Bool=false)
...
+    sed -i 's/FortranCInterface_VERIFY/# FortranCInterface_VERIFY/g' ./CBLAS/CMakeLists.txt
+    sed -i 's/FortranCInterface_VERIFY/# FortranCInterface_VERIFY/g' ./LAPACKE/include/CMakeLists.txt
+
     mkdir build && cd build
-    cmake .. "${CMAKE_FLAGS[@]}" \
+    cmake .. \
        -DCMAKE_INSTALL_PREFIX="$prefix" \
        -DCMAKE_FIND_ROOT_PATH="$prefix" \
        -DCMAKE_TOOLCHAIN_FILE="${CMAKE_TARGET_TOOLCHAIN}" \
+       -DCMAKE_Fortran_FLAGS="-fno-optimize-sibling-calls" \
```
imciner2, line 37: What's the reason for disabling this optimization? It would be good to document that in a comment at least.

ViralBShah: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/92

PR #11754, opened 2025-07-31
[FLINT] provide dummy function to make `va_end` available from julia

lgoettgens:
This contains the changes on the C side to make https://github.com/Nemocas/Nemo.jl/pull/2128 work.

Please only merge once @fingolfin and @thofma had a chance to review it.


thofma:
Looks harmless, but I suggest we wait for @fingolfin to have a look.

lgoettgens:
> Looks harmless, but I suggest we wait for @fingolfin to have a look.

Agreed, I'll try to remember to ping them here again once they are back in the office.

F/FLINT/bundled/patches/Provide-va_end-macro-as-a-function-to-be-used-from-j.patch
```diff
@@ -0,0 +1,50 @@
...
++#include <stdarg.h>
++#include "flint.h"
++
++/* provide some macros as functions for use from julia ***********************/
++
++#ifdef FLINT_HAVE_VA_LIST
++void flint_va_end(va_list vlist)
++{
++    va_end(vlist);
++}
```
fingolfin, line 45: Long version: if the resulting binaries actually work for you, let's do it.


----

Longer version: I am somewhat surprised this works (or even compiles). Was it actually field tested? I am not sure it is generally portable -- but if it works for gcc and clang, and on Linux/macOS/BSD (AFAIK these macros are also ABI dependent), then of course that's "good enough" for us...

(In fact I am pretty sure for most OS / compiler combos `va_end` is a no-opp, but apparently in old TCC versions it invoked `free` on a memory block allocated by `va_start`...)

Plus it seems this is what FLINT already does, so perhaps it *is* safe if one reads the standard spec closely enough: the following code calls `va_end` in a function that did not call `vs_start`. OK...
```C
FLINT_NORETURN static void __flint_throw(flint_err_t exc, const char * msg, va_list ap)
{
...
    flint_vprintf(msg, ap);
    va_end(ap);
    flint_abort();
}

...

FLINT_NORETURN void flint_throw(flint_err_t exc, const char * msg, ...)
{
    va_list ap;
    va_start(ap, msg);
    (*throw_func)(exc, msg, ap);
}
```

---

To clarify: The `va_start/va_arg/va_end` macros are among the dark voodoo of C compilers and can do some really weird shit. Here is the definition from macOS / clang:

```
#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
/* C23 does not require the second parameter for va_start. */
#define va_start(ap, ...) __builtin_va_start(ap, 0)
#else
/* Versions before C23 do require the second parameter. */
#define va_start(ap, param) __builtin_va_start(ap, param)
#endif
#define va_end(ap) __builtin_va_end(ap)
#define va_arg(ap, type) __builtin_va_arg(ap, type)
```

So yeah, it "doesn't do fancy macro magic" (quoting @lgoettgens here). But to quote someone on the net:

> `__builtin_va_start` and `__builtin_va_end` specifically are weird because they're builtins which have a signature which can't be expressed in C.


The [clang implementation](https://github.com/llvm/llvm-project/blob/17dbb92612db6dcf9746cf494d37476b93e49a49/clang/include/clang/CIR/Dialect/IR/CIROps.td#L3683) does not really make it clear whether the above should be always safe or not either.

fingolfin:
@lgoettgens wrote on https://github.com/Nemocas/Nemo.jl/pull/2128 that he tested this PR locally, so why not... but it is marked as a "draft" ?

lgoettgens:
This was marked as a draft so that nobody accidentally merges it while you (@fingolfin) were on vacation.
 I think adding this here should break nothing. If it indeed works on systems other than my own will be tested in the Nemo PR once this here is released. If it turns out that it won't work, we can just revert this change here.

So please go ahead with merging

lgoettgens:
Bump @fingolfin, please merge

PR #11757, opened 2025-08-01
[LibCURL] Build with zstd support

Keno:
Since Zstd_jll is now a julia standard lib, there is no reason not to enable it in CURL (and in fact a source build currently enables it automatically). Ref https://github.com/JuliaLang/julia/issues/59171.

### DIFF
--- a/L/LibCURL/common.jl
+++ b/L/LibCURL/common.jl
@@ -22,7 +22,7 @@ const curl_hashes = Dict(
     v"8.15.0" => "d85cfc79dc505ff800cb1d321a320183035011fa08cb301356425d86be8fc53c",
 )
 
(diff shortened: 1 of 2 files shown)


L/LibCURL/common.jl
```diff
@@ -155,6 +155,7 @@ function build_libcurl(ARGS, name::String, version::VersionNumber)
     dependencies = [
         Dependency("LibSSH2_jll"),
         Dependency("Zlib_jll"),
+        Dependency("Zstd_jll"),
```
giordano: This should be dependent on the version, so that if necessary we can rebuild older versions consistently (without zstd).

Keno:
@giordano do you like this version better?

PR #11762, opened 2025-08-01
cuPyNumeric

ejmeitz:
Had to:
 - Use`dont_dlopen` cause ` version GLIBCXX_3.4.30 not found` in build env (same as legate_jll)
 - Added CUTENSOR_JLL as an `ArchiveSource`. CuTensor is only built for cuda+12.0 as it only depends on the driver which is backwards compatible. I expand the CUDA version which results in CUTENSOR_JLL not getting installed for any platform with CUDA newer than cuda+12.0.
 - Built tblis from source instead of using tblis_jll. Needed to set some flags not present in our recipe.

### DIFF
--- a/C/cupynumeric/build_tarballs.jl
+++ b/C/cupynumeric/build_tarballs.jl
@@ -1,0 +1,129 @@
+using BinaryBuilder
+import Pkg: PackageSpec
+using Base.BinaryPlatforms: arch, os, tags
+
+# needed for libjulia_platforms and julia_versions
+const YGGDRASIL_DIR = "../../"
+include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
+include(joinpath(YGGDRASIL_DIR, "platforms", "cuda.jl"))
+
+name = "cupynumeric"
+version = v"25.5" # cupynumeric has 05, but Julia doesn't like that
+sources = [
+    GitSource("https://github.com/nv-legate/cupynumeric.git","cbd9a098b32531d68f1b3007ef86bb8d3859174d"),
+]
+
+
+# These should match the legate_jll build_tarballs script
+MIN_CUDA_VERSION = v"12.2"
+MAX_CUDA_VERSION = v"12.8.999"
+
+script = raw"""
+
+    # Put new CMake first on path
+    export PATH=${host_bindir}:$PATH
+
+    cd ${WORKSPACE}/srcdir/cupynumeric
+
+    # Necessary operations to cross compile CUDA from x86_64 to aarch64
+    if [[ "${target}" == aarch64-linux-* ]]; then
+
+        # Add /usr/lib/csl-musl-x86_64 to LD_LIBRARY_PATH to be able to use host nvcc
+        export LD_LIBRARY_PATH="/usr/lib/csl-musl-x86_64:/usr/lib/csl-glibc-x86_64:${LD_LIBRARY_PATH}"
+        
+        # Make sure we use host CUDA executable by copying from the x86_64 CUDA redist
+        NVCC_DIR=(/workspace/srcdir/cuda_nvcc-*-archive)
+        rm -rf ${prefix}/cuda/bin
+        cp -r ${NVCC_DIR}/bin ${prefix}/cuda/bin
+        
+        rm -rf ${prefix}/cuda/nvvm/bin
+        cp -r ${NVCC_DIR}/nvvm/bin ${prefix}/cuda/nvvm/bin
+
+        export NVCC_PREPEND_FLAGS="-ccbin='${CXX}'"
+    fi
+
+    export CUDA_HOME=${prefix}/cuda;
+    export PATH=$PATH:$CUDA_HOME/bin
+    export CUDACXX=$CUDA_HOME/bin/nvcc
+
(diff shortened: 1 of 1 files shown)


C/cupynumeric/build_tarballs.jl
```diff
@@ -0,0 +1,178 @@
...
+
+    should_build_platform(triplet(platform)) || continue
+
+    cuda_deps = CUDA.required_dependencies(platform, static_sdk=true)
+
+    cuda_ver = platform["cuda"]
+
+    platform_sources = BinaryBuilder.AbstractSource[sources...]
+
+    # Add x86_64 CUDA_SDK, nvcc isn't actually used but CMake
```
krasow: Are you sure? There are a lot of .cu files and nvcc is used by cuPyNumeric's CMake. nvcc invokes ptxas when you assemble for a specific sm arch.

C/cupynumeric/build_tarballs.jl
```diff
@@ -0,0 +1,178 @@
...
+
+    install_license $WORKSPACE/srcdir/cupynumeric*/LICENSE
+    install_license $WORKSPACE/srcdir/share/licenses/CUTENSOR/LICENSE
+
+
+"""
+
+platforms = CUDA.supported_platforms(; min_version = MIN_CUDA_VERSION, max_version = MAX_CUDA_VERSION)
+
+# for now NO ARM support, tblis doesnt have docs on how to build for arm
```
krasow, line 119: cuPyNumeric depends on tblis (private fork) and can be built for ARM. It is unclear how with the public repo.

https://anaconda.org/legate/cupynumeric/25.07.00/download/linux-aarch64/cupynumeric-25.07.00-cuda12_py313_gpu_g6132d845_0.conda

C/cupynumeric/build_tarballs.jl
ejmeitz: oh that comment was just from what I copied the code from. This definitely uses nvcc :)

C/cupynumeric/build_tarballs.jl
ejmeitz: Yeah ARM is a later problem....

ejmeitz:
@giordano when you have time! This should be good, but the ArchiveSource might be a bit non standard.

C/cupynumeric/build_tarballs.jl
```diff
@@ -0,0 +1,174 @@
...
+        --enable-config=${BLI_CONFIG}
+
+    make -j ${nproc} && make install
+
+    cd ${WORKSPACE}/srcdir/cupynumeric
+
+    # COPIED FROM OpenBLAS_jll script
+    if [[ ${target} == aarch64-* ]] && [[ ${bb_full_target} != *-libgfortran3* ]]; then
+        export TARGET=ARMV8
+    fi
```
giordano: OpenBLAS recipe defines that because that's a [variable defined in OpenBLAS' build system](http://www.openmathlib.org/OpenBLAS/docs/install/), with specific values expected, but here you aren't compiling OpenBLAS, are you?  Why is this needed here?

C/cupynumeric/build_tarballs.jl
```diff
@@ -0,0 +1,174 @@
...
+        -Dlegate_ROOT:STRING=${prefix} \
+        -DCMAKE_PREFIX_PATH=${prefix} \
+        -DCMAKE_INSTALL_PREFIX=${prefix} \
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DNCCL_LIBRARY=${libdir}/libnccl.so \
+        -DNCCL_INCLUDE_DIR=${includedir} \
+        -Dcutensor_LIBRARY=${libdir}/libcutensor.so \
+        -Dcutensor_INCLUDE_DIR=${includedir} \
+        -DBLAS_LIBRARIES=${libdir}/libopenblas.so \
```
giordano, line 106: Why is OpenBLAS32 used here?  Don't you use tblis already?

C/cupynumeric/build_tarballs.jl
krasow: cuPyNumeric builds LAPACK/BLAS from source if a path isn't specified. This may not be necessary.

C/cupynumeric/build_tarballs.jl
krasow: https://docs.nvidia.com/cupynumeric/latest/developer/building.html#building-cupynumeric-from-source

Both are used. TBLIS -> tensor contraction and BLAS -> linear algebra

C/cupynumeric/build_tarballs.jl
ejmeitz: Yeah this can be removed, is a remnant from when I was letting cupynumeric build BLAS.

C/cupynumeric/build_tarballs.jl
ejmeitz: They list both as dependencies: https://docs.nvidia.com/cupynumeric/25.07/developer/building.html

I'm not super familiar with tblis, but I'm pretty sure they provide different routines. The build also definitely fails if both are not present.

ejmeitz:
Ok this should be good now! @giordano

PR #11767, opened 2025-08-02
[Uno] update to v2.0.1

cvanaret:
Uno v2.0.1 released on August 2nd, 2025 (documented [here](https://github.com/cvanaret/Uno/releases/tag/v2.0.1)).

cc @amontoison

### DIFF
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -73,6 +73,8 @@ filter!(p -> !(Sys.isfreebsd(p) && arch(p) == "aarch64"), platforms)
 filter!(p -> arch(p) != "riscv64", platforms)
 platforms = expand_cxxstring_abis(platforms)
 platforms = expand_gfortran_versions(platforms)
+platforms = filter(p -> libgfortran_version(p) != v"3", platforms)
+platforms = filter(p -> libgfortran_version(p) != v"4", platforms)
 
 products = [
     # We call this amplexe to match the convention of other JLL packages (like Ipopt_jll) that provide AMPL wrappers
--- a/U/Uno/build_tarballs.jl
+++ b/U/Uno/build_tarballs.jl
@@ -4,12 +4,12 @@ using BinaryBuilder, Pkg
 
 name = "Uno"
 
-version = v"2.0.0"
+version = v"2.0.1"
 
 sources = [
     GitSource(
         "https://github.com/cvanaret/Uno.git",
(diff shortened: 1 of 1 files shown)


U/Uno/build_tarballs.jl
```diff
@@ -77,12 +72,13 @@ platforms = supported_platforms()
 filter!(p -> !(Sys.isfreebsd(p) && arch(p) == "aarch64"), platforms)
 filter!(p -> arch(p) != "riscv64", platforms)
 platforms = expand_cxxstring_abis(platforms)
+platforms = expand_gfortran_versions(platforms)
```
amontoison, line 75: @cvanaret 
https://github.com/JuliaPackaging/Yggdrasil/blob/master/G/GALAHAD/build_tarballs.jl#L135-L136

amontoison: We also dropped `libgfortran3` and `libgfortran4` for the last release of `HSL_jll.jl`:
https://discourse.julialang.org/t/psa-intent-to-deprecate-cxx03-libgfortran3-libgfortran4-abis/130799

giordano: Why dropping libgfortran4? That's working, isn't it?

U/Uno/build_tarballs.jl
```diff
@@ -20,10 +20,10 @@ cd build
 
 if [[ "${target}" == *mingw* ]]; then
     LBT=blastrampoline-5
-    LIBHIGHS=${prefix}/lib/libhighs.dll.a
+    HIGHS_DIR=${prefix}/lib
```
amontoison, line 23: @cvanaret It seems to be wrong, can you check the logs to see if it is correctly detected / linked on Windows?

cvanaret: I'll have a look. That's what we have in Uno's CI script.

U/Uno/build_tarballs.jl
amontoison: We can keep it but who is still using it?  Uno also requires a recent GCC (10.x), I am scared that it will be  a pain to maintain for GCC 7, required for `libgfortran4`.

U/Uno/build_tarballs.jl
amontoison: You did this modification and not me :)

cvanaret: I like to live dangerously :) It does look ok though.

PR #11769, opened 2025-08-03
update LCIO_Julia_Wrapper to modern CxxWrap

Moelf:


barche:
Are you sure you don't need the 1.12-rc1 and nightly versions? It might be safer to just include all versions at LTS and higher, using e.g. `filter!(x -> x >= v"1.10.0",  julia_versions)`

Moelf:
I'm okay with this for now, also idk how to obtain that initial list of Julia versions

barche:
julia_versions is defined automatically when including the libjulia common.jl.

Moelf:
which is not available in this build script? I don't understand how this would work

Moelf:
https://github.com/JuliaPackaging/Yggdrasil/blob/1e903be5475b272f1137972eed9dc0f3d427923a/O/OpenFHE/common.jl#L3

I will copy this example

Moelf:
@barche apparently it doesn't support RC

L/LCIO_Julia_Wrapper/build_tarballs.jl
```diff
@@ -3,19 +3,20 @@
...
-uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
-delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "LCIO_Julia_Wrapper"
-version = v"0.13.3"
-julia_versions = [v"1.6.3", v"1.7", v"1.8", v"1.9", v"1.10"]
+version = v"0.13.4"
+julia_versions = filter(v -> !occursin("DEV", string(v)), julia_full_versions)
+julia_versions = filter(v -> !occursin("-rc", string(v)), julia_versions)
+julia_versions = filter(v -> v >= v"1.10", julia_versions)
```
benlorenz:
```suggestion
julia_versions = filter(v -> v >= v"1.10", julia_versions)
```
The platform tags don't like DEV or rc versions but if you just use the `julia_versions` variable (also provided by `common.jl`) instead of the `full` versions then this should be fine.

Moelf:
@barche I think we gotta drop dev

benlorenz:
That is weird, libcxxwrap_julia did build fine with -DEV: https://buildkite.com/julialang/yggdrasil/builds/21937/steps/table
And I used it to build libpolymake_julia also for -DEV: https://buildkite.com/julialang/yggdrasil/builds/21959/steps/table
I will try to have a look at the error.

L/LCIO_Julia_Wrapper/build_tarballs.jl
```diff
@@ -45,10 +44,10 @@ products = [
 
 # Dependencies that must be installed before this package can be built
 dependencies = [
-    Dependency(PackageSpec(name="libcxxwrap_julia_jll"), compat="0.13.4"),
+    Dependency(PackageSpec(name="libcxxwrap_julia_jll"), compat="0.14"),
```
benlorenz:
```suggestion
    Dependency(PackageSpec(name="libcxxwrap_julia_jll"), compat="0.14.3"),
```
We need a more recent patch version of libcxxwrap for 1.13 (Yggdrasil always builds with the oldest compatible version if a compat entry is given).

Moelf:
finally, I think we can merge now @giordano , whenever you have time no rush

PR #11770, opened 2025-08-03
new recipe: perfetto

simeonschaub:


P/perfetto/build_tarballs.jl
```diff
@@ -0,0 +1,56 @@
...
+name = "perfetto"
+version = v"51.2.0"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource(
+        "https://github.com/google/perfetto",
+        "7a9a6a0587348bffd1796b66a1da33cc1ea421d8"
+    ),
+    DirectorySource("./bundled"),
```
giordano, line 14: Can you please upstream all these patches? They don't seem specific to our setup, right? Being able to compile with mingw is a generally useful feature which other projects may benefit from (and fixes wrong msvc specialisation in upstream source code).

simeonschaub: Yes, the only specific one to our setup should be the meson.build one, I believe google always does static linking. Will try to upstream these patches

PR #9103, opened 2024-07-19
Add Powsybl julia binaries building script

obrix:
Hello,

This pull request is adding the build script build_tarballs.jl to build the binary part of powsybl julia bindings.

[PowSyBl ](https://www.powsybl.org/)(Power System Blocks) is an open source framework written in Java, dedicated to electrical grid modelling and simulation.

One thing that is a bit specific to the building process here is that we have a dependency to pre built binaries generated by GraalVM native image. it is used by [pypowsybl ](https://github.com/powsybl/pypowsybl) (powsybl python bindings) to provide a C++ interface to the java framework. Those binaries are also used by the julia bindings and need to be pre built elsewhere because native image does not support cross compilation yet.

This is a work in progress, feel free to comment if something seems off.

PR #11775, opened 2025-08-04
Update pypowsybl and powsybl.jl versions

obrix:
Update pypowsybl and powsybl.jl versions

### DIFF
--- a/P/Powsybl/build_tarballs.jl
+++ b/P/Powsybl/build_tarballs.jl
@@ -8,26 +8,23 @@ version = v"0.2.0"
 uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
-pypowsybl_version = "1.9.0"
+pypowsybl_version = "1.12.0"
 
 
 sources = [
     GitSource("https://github.com/powsybl/powsybl.jl.git", "19ec2f5aff90d42df5d90f85dda76f6007a13757"),
-    GitSource("https://github.com/powsybl/pypowsybl.git", "81d7e40a32d056a9945293238ce09cf207fc47fc"),
+    GitSource("https://github.com/powsybl/pypowsybl.git", "cfc5f6b15e31d11f1879ba01fbf9e9f8032efd0b"),
     ArchiveSource("https://github.com/powsybl/pypowsybl/releases/download/v$(pypowsybl_version)/binaries-v$(pypowsybl_version)-windows.zip",
-                  "a3d1912e9ed213a4ca157ffa8470e4c38cbce84fa1529c646cb32ec4562f8c13",
+                  "467d269c52de4a3bcc73dc351f7f777357c5dda0311962b605f7262e3bde639d",
                   "powsybl-java-x86_64-w64-mingw32"),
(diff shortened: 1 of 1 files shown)


P/Powsybl/build_tarballs.jl
```diff
@@ -1,33 +1,30 @@
...
 sources = [
-    GitSource("https://github.com/powsybl/powsybl.jl.git", "19ec2f5aff90d42df5d90f85dda76f6007a13757"),
-    GitSource("https://github.com/powsybl/pypowsybl.git", "81d7e40a32d056a9945293238ce09cf207fc47fc"),
+    GitSource("https://github.com/powsybl/powsybl.jl.git", "70503bf17751430d4c00f7345986bd574cbdb3ef"),
+    GitSource("https://github.com/powsybl/pypowsybl.git", "342fb354a7c9f9bdfdec66d5901005293848d64b"),
     ArchiveSource("https://github.com/powsybl/pypowsybl/releases/download/v$(pypowsybl_version)/binaries-v$(pypowsybl_version)-windows.zip",
-                  "a3d1912e9ed213a4ca157ffa8470e4c38cbce84fa1529c646cb32ec4562f8c13",
+                  "4b4a8c1b2bc9a210902773bda3f14a4131e48fb54a453857164c7d90fa8114e3",
                   "powsybl-java-x86_64-w64-mingw32"),
     ArchiveSource("https://github.com/powsybl/pypowsybl/releases/download/v$(pypowsybl_version)/binaries-v$(pypowsybl_version)-linux.zip",
```
fingolfin, line 21: Why does this have arch specific source archives? Do those contain generated code? Normally the goal is to have everything built from source. If that is not practical here, a comment explaining the rationale here would be good

P/Powsybl/build_tarballs.jl
```diff
@@ -73,7 +70,7 @@ install_license $WORKSPACE/srcdir/powsybl.jl/LICENSE.md
 include("../../L/libjulia/common.jl")
 platforms = vcat(libjulia_platforms.(julia_versions)...)
 
-filter!(p -> (arch(p) == "x86_64" && os(p) ∈ ("windows", "linux", "macos")) || (arch(p) == "aarch64" && Sys.isapple(p)), platforms)
```
fingolfin, line 76: This PR seems to drop macos support, but this is not mentioned or explained in the PR description.

P/Powsybl/build_tarballs.jl
obrix: This was explained in the initial pull request for the powsybl recipe here for reference :  https://github.com/JuliaPackaging/Yggdrasil/pull/9103 . This is only a version bump.
Powsybl julia (and pypowsybl) rely on graalVM native image to compile java code to a native binaries. This part cannot be done in a binary builder environment. This is why the recipe is fetching pre built binaries.

obrix: I just added a small comment in the recipe.

P/Powsybl/build_tarballs.jl
obrix: Dropping x86_64 support for macos has been in the pipe for some time in the powsybl project, but we still support it. 
This was a mistake, I just reverted the change, sorry for the noise.

PR #11776, opened 2025-08-04
[minio] Bump MinIO version

andrebsguedes:
The version in use currently has a flaky segfault that makes CloudTest tests spuriously fail.

### DIFF
--- a/M/minio/build_tarballs.jl
+++ b/M/minio/build_tarballs.jl
@@ -5,7 +5,8 @@ version = v"1.0.0"
 
 # Collection of sources
 sources = [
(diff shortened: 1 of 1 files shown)


andrebsguedes:
Do you have a suggestion? This is quite the version bump so maybe a major like `2.0.0`?

giordano:
Doesn't upstream have a proper version number? If not (sigh), did the API/ABI change in a breaking way? If so, then bumping to v2.0 would make sense.

andrebsguedes:
> Doesn't upstream have a proper version number? If not (sigh), did the API/ABI change in a breaking way? If so, then bumping to v2.0 would make sense.

They don't unfortunately, I have skimmed through the changelogs and saw some breaking releases between the versions in question, so 2.0 it is.

PR #11781, opened 2025-08-05
[cargo_c] New recipe

nhz2:


C/cargo_c/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+filter!(p -> !Sys.iswindows(p) || arch(p) != "i686", platforms)
+
+# Rust toolchain 1.87.0 not available on platform aarch64-unknown-freebsd or riscv64-linux-gnu
+filter!(p -> !(arch(p) == "aarch64" && Sys.isfreebsd(p)), platforms)
+filter!(p -> !(arch(p) == "riscv64"), platforms)
+
+# Errors with "ld: library not found for -lclang_rt.osx"
+filter!(p -> !Sys.isapple(p), platforms)
+
+# Audit warns: "Linked library bcryptprimitives.dll could not be resolved and could not be auto-mapped"
```
giordano: This library should be added to https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/d41ddc84157a8c4b490761b0fb5f66cf32746ce6/src/auditor/dynamic_linkage.jl#L299-L351, there's no reason to skip windows only for this.

C/cargo_c/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+sources = [
+    GitSource("https://github.com/lu-zero/cargo-c", "d1da3c27a0defdceba12c2b2762094cdde6c0337"),
+    FileSource("https://github.com/lu-zero/cargo-c/releases/download/v$(version)/Cargo.lock", "0af99496210d7e5cb70de6643b571a8882120148e628d7a75327937b64fd9841")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cp $WORKSPACE/srcdir/Cargo.lock $WORKSPACE/srcdir/cargo-c/
+cd $WORKSPACE/srcdir/cargo-c/
+cargo build --release --locked --features=vendored-openssl
```
giordano: Why not using our openssl?

C/cargo_c/build_tarballs.jl
```diff
@@ -0,0 +1,63 @@
...
+platforms = supported_platforms()
+
+# Rust toolchain for i686 Windows is unusable
+filter!(p -> !Sys.iswindows(p) || arch(p) != "i686", platforms)
+
+# Rust toolchain 1.87.0 not available on platform aarch64-unknown-freebsd or riscv64-linux-gnu
+filter!(p -> !(arch(p) == "aarch64" && Sys.isfreebsd(p)), platforms)
+filter!(p -> !(arch(p) == "riscv64"), platforms)
+
+# Errors with "ld: library not found for -lclang_rt.osx"
```
giordano: That's in llvm compiler rt

C/cargo_c/build_tarballs.jl
nhz2: I get an error 
```
  = note: /opt/x86_64-linux-musl/bin/../lib/gcc/x86_64-linux-musl/5.2.0/../../../../x86_64-linux-musl/bin/ld: cannot find -lssl
          /opt/x86_64-linux-musl/bin/../lib/gcc/x86_64-linux-musl/5.2.0/../../../../x86_64-linux-musl/bin/ld: cannot find -lcrypto
```

When compiling for x86_64-linux-musl

C/cargo_c/build_tarballs.jl
nhz2: For some reason, I still get this error even if I have `LLVMCompilerRT_jll`

nhz2:
Closing this because it seems better to download the built version from https://github.com/lu-zero/cargo-c/releases than struggle with the rust compiler and the large number of dependencies cargo-c is using.

PR #11788, opened 2025-08-05
Update various SuiteSparse component packages

imciner2:


ViralBShah:
I usually find it better to update the pieces individually.

imciner2:
Yea... I was hoping it was just going to work. Apparently something doesn't like the new CUDA 13 we have though.

I guess I can either split out the CUDA update, or just ignore v13 for now. I am tempted to say let's just ignore v13 for now, because it might also require upstream code changes since it is a new major version.

ViralBShah:
Yeah I think just stay with CUDA 12.

PR #11797, opened 2025-08-06
[GettextTiny] New recipe

nhz2:
From https://github.com/sabotage-linux/gettext-tiny
This provides an alternative complete implementation of `msgfmt` that is easier to compile compared to `Gettext_jll`.

### DIFF
--- a/G/GettextTiny/build_tarballs.jl
+++ b/G/GettextTiny/build_tarballs.jl
@@ -16,18 +16,12 @@ export LDFLAGS="-L${libdir}"
 
 make LIBINTL=NONE
 make LIBINTL=NONE DESTDIR=${prefix} prefix=/ install
-
-# Add .exe on Windows if it is missing
-if [[ -f ${bindir}/msgfmt ]]
-then
-    mv -n ${bindir}/msgfmt ${bindir}/msgfmt${exeext}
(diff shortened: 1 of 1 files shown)


G/GettextTiny/build_tarballs.jl
```diff
@@ -0,0 +1,43 @@
...
+
+make LIBINTL=NONE
+make LIBINTL=NONE DESTDIR=${prefix} prefix=/ install
+
+# Add .exe on Windows if it is missing
+if [[ -f ${bindir}/msgfmt ]]
+then
+    mv -n ${bindir}/msgfmt ${bindir}/msgfmt${exeext}
+    mv -n ${bindir}/msgmerge ${bindir}/msgmerge${exeext}
+fi
```
giordano: Using gcc 6 or so would avoid this.

nhz2: Yes, with gcc 8 I get the .exe, but then this breaks the install step in the Makefile, because I assume it is hardcoding the name of the executable somewhere.

nhz2: I'm removing Windows builds since they are unlikely to work and I don't have a good way to test them.

PR #11798, opened 2025-08-06
[power_grid_model_c]  version 1.12 new package

uwbanjoman:
https://lfenergy.org/projects/power-grid-model/


P/power_grid_model_c/build_tarballs.jl
```diff
@@ -0,0 +1,59 @@
...
+
+platforms = [
+             Platform("x86_64", "linux"; libc = "glibc", cpu_target="x86_64_v2", cxxstring_abi=:cxx11),
+             Platform("x86_64", "linux"; libc = "glibc", cpu_target="x86_64_v3", cxxstring_abi=:cxx11),
+             Platform("aarch64", "linux"; libc = "glibc", cxxstring_abi=:cxx11),
+             Platform("x86_64", "linux"; libc = "musl", cxxstring_abi=:cxx11),
+             Platform("aarch64", "linux"; libc = "musl", cxxstring_abi=:cxx11),
+             Platform("x86_64", "windows"; cpu_target="x86_64_v2", cxxstring_abi=:cxx11, march="avx2"),
+             Platform("x86_64", "windows"; cpu_target="x86_64_v3", cxxstring_abi=:cxx11, march="avx2"),
+             Platform("x86_64", "linux"; libc="musl", cxxstring_abi=:cxx11),
```
imciner2:
```suggestion
```
Duplicated.

P/power_grid_model_c/build_tarballs.jl
```diff
@@ -0,0 +1,59 @@
...
+cd power_grid_model-1.12.0/power_grid_model_c
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_CXX_STANDARD=20 -DCMAKE_CXX_FLAGS="$CXXFLAGS" -DCMAKE_BUILD_TYPE=Release
+cmake --build build --parallel ${nproc}
+cmake --install build
+install_license $WORKSPACE/srcdir/power_grid_model-1.12.0/LICENSE
+"""
+
+platforms = [
+             Platform("x86_64", "linux"; libc = "glibc", cpu_target="x86_64_v2", cxxstring_abi=:cxx11),
+             Platform("x86_64", "linux"; libc = "glibc", cpu_target="x86_64_v3", cxxstring_abi=:cxx11),
```
imciner2: What are you trying to accomplish with all these `cpu_target` flags added to the platform? That flag is not provided by BinaryBuilder, and it doesn't appear you use them in the above build recipe at all, so how it supposed to be used?

uwbanjoman:
Thanks for your suggestions, I have cleaned up the platforms. and also am in learning mode.

P/power_grid_model_c/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+
+
+# The products that we will ensure are always built.
+products = [
+            LibraryProduct("libpower_grid_model_c", :libpower_grid_model_c; dont_dlopen=true)
+           ]
+
+# Dependencies that must be installed before this package can be built
+dependencies = [
+                Dependency(PackageSpec(name="boost_jll", uuid="28df3c45-c428-5900-9ff8-a3135698ca75"))
```
imciner2: You need to add a compat entry for boost_jll: https://github.com/JuliaPackaging/Yggdrasil/issues/3024#issue-894813654

P/power_grid_model_c/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+    atomic_patch -p1 ${f}
+done
+# On intel processors, enable SSE2 and AVX2
+if [[ "${target}" == x86_64-*  ]]; then
+    FLAGS+=( --enable-sse2 --enable-avx2 )
+fi
+# Enable NEON on Aarch64
+if [[ "${target}" == aarch64-* ]]; then
+    FLAGS+=( --enable-neon )
+fi
```
imciner2: When building packages in Ygg, we must target a common subset of ISAs, because otherwise if this gets loaded on a computer that is older than what is expected, then it can cause Julia to crash (https://docs.binarybuilder.org/stable/tricksy_gotchas/, "Instruction Set Differences" section). So you shouldn't be setting these flags in the build script.

P/power_grid_model_c/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+cd power_grid_model-1.12.0/power_grid_model_c
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_CXX_STANDARD=20 -DCMAKE_CXX_FLAGS="$CXXFLAGS" -DCMAKE_BUILD_TYPE=Release
```
imciner2:
```suggestion
cmake -B build \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_CXX_STANDARD=20 \
    -DCMAKE_CXX_FLAGS="$CXXFLAGS" \
    -DCMAKE_BUILD_TYPE=Release
```
For readability.

You already are setting the CXX standard in the patch, why are you passing it as a command line option here?

You reference `CXXFLAGS` here, but you are setting `FLAGS` above. However, given my comment about not using those flags, I think you can just remove the setting of `CMAKE_CXX_FLAGS` from here.

P/power_grid_model_c/build_tarballs.jl
```diff
@@ -0,0 +1,61 @@
...
+install_license $WORKSPACE/srcdir/power_grid_model-1.12.0/LICENSE
+"""
+
+platforms = [
+             Platform("x86_64", "linux"; libc = "glibc", cxxstring_abi=:cxx11),
+             Platform("x86_64", "linux"; libc = "musl", cxxstring_abi=:cxx11),
+             Platform("aarch64", "linux"; libc = "glibc", cxxstring_abi=:cxx11),
+             Platform("aarch64", "linux"; libc = "musl", cxxstring_abi=:cxx11),
+             Platform("x86_64", "windows"; cxxstring_abi=:cxx11),
+            ]
```
imciner2:
```suggestion
platforms = [
             Platform("x86_64", "linux"; libc = "glibc"),
             Platform("x86_64", "linux"; libc = "musl"),
             Platform("aarch64", "linux"; libc = "glibc"),
             Platform("aarch64", "linux"; libc = "musl"),
             Platform("x86_64", "windows"),
            ]
platforms = expand_cxxstring_abis(platforms)
```
We should be building with both the CXX string ABIs (https://docs.binarybuilder.org/stable/tricksy_gotchas/, `cxx11 string ABI` section).

P/power_grid_model_c/bundled/patches/cmake-patch.patch
imciner2, line 1: What is the purpose of this patch? It looks like the contents of this patch are just the contents of the CMake in the root of the repository, so it would be better if you just called that CMake file instead of changing to the `power_grid_model_c` directory and using this CMake directly.

imciner2 (changes requested):

P/power_grid_model_c/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+
+# cmake reprts "Could NOT find Boost (missing: Boost_INCLUDE_DIR)",
+# on aarch64-unknown-freebsd, remove
+platforms = filter!(p -> !(os(p) == "freebsd"), platforms)
+
+# cmake reports "note: parameter passing for argument of type ‘struct format_args’ changed in GCC 9.1", remove
+platforms = filter!(p -> cxxstring_abi(p) != "cxx03" && arch(p) != "armv6l", platforms)
+platforms = filter!(p -> cxxstring_abi(p) != "cxx03" && (arch(p) != "armv7l"), platforms)
+
+platforms = expand_cxxstring_abis(platforms)
```
imciner2, line 56: Are these out of order? You seem to use the CXX ABI in the conditional, but this is before the ABI is actually expanded. Did you just want to filter out all the arm6 and arm7 platforms, or do you actually want to keep the `cxx11` ABI ones?

(right now all arm6 and arm7 platforms are getting removed).

P/power_grid_model_c/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+platforms = filter!(p -> !(arch(p) == "riscv64"), platforms)
+
+# Apple build reports "fatal error: 'concepts' file not found"
+# it needs a higher version of boost_jll, 
+# whhich in turn needs an upgrade of msgpack_jll. maybe later, remove.
+platforms = filter(p -> !Sys.isapple(p), platforms)
+
+# cmake reprts "Could NOT find Boost (missing: Boost_INCLUDE_DIR)",
+# on aarch64-unknown-freebsd, remove
+platforms = filter!(p -> !(os(p) == "freebsd"), platforms)
```
imciner2:
```suggestion
platforms = filter!(p -> !Sys.isfreebsd(p), platforms)
```
We have a helper function for detecting the FreeBSD OS.

giordano: For what is worth, that can also be written as
```suggestion
platforms = filter!(!Sys.isfreebsd, platforms)
```

P/power_grid_model_c/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+# Collection of sources required to complete build
+sources = [
+           ArchiveSource("https://github.com/PowerGridModel/power-grid-model/releases/download/v$(version)/power_grid_model-$(version).tar.gz", 
+                      "b38be158af11541759b7b2c01e1baab099f8f22fe453237d7814a8698bb67745")
+          ]
+
+# Bash recipe for building across all platforms.
+script = raw"""
+apk del cmake
+cd $WORKSPACE/srcdir/power_grid_model-1.12.0
```
fingolfin:
```suggestion
cd $WORKSPACE/srcdir/power_grid_model*
```

P/power_grid_model_c/build_tarballs.jl
```diff
@@ -0,0 +1,65 @@
...
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_CXX_STANDARD=20 -DCMAKE_BUILD_TYPE=Release
+cmake --build build --parallel ${nproc}
+cmake --install build
+install_license $WORKSPACE/srcdir/power_grid_model-1.12.0/LICENSE
```
fingolfin:
```suggestion
install_license $WORKSPACE/srcdir/power_grid_model*/LICENSE
```

P/power_grid_model_c/build_tarballs.jl
fingolfin: Indeed this seems out of order

fingolfin:
@uwbanjoman hi there, do you still plan to finish this? There have been a couple suggestions for changing that you then should consider addressing.

fingolfin:
@uwbanjoman ping

PR #11813, opened 2025-08-11
[tree_sitter_modelica] New recipe

m-c-marius:


T/tree_sitter_modelica/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+make -j${nproc}
+make install LANGUAGE_NAME=libtreesitter_modelica
+
+cd ..
+if [ -d tree-sitter/queries ]; then
+   cp -r tree-sitter/queries $WORKSPACE/destdir/
+fi
+if [ -f tree-sitter/LICENSE ]; then
+    install_license tree-sitter/LICENSE
+fi
```
imciner2: Why does this need to be guarded here? If the license goes missing, that is something we would want to know about through an error.

T/tree_sitter_modelica/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+BUILD_FLAGS=(-DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN})
+mkdir build && cd build
+cmake .. "${BUILD_FLAGS[@]}"
+make -j${nproc}
+make install LANGUAGE_NAME=libtreesitter_modelica
+
+cd ..
+if [ -d tree-sitter/queries ]; then
+   cp -r tree-sitter/queries $WORKSPACE/destdir/
+fi
```
imciner2: What is this supposed to be doing? There doesn't appear to be any directory called `queries` in the upstream repo, so this is not getting used according the the build logs.

T/tree_sitter_modelica/bundled/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.13)
+project(treesitter_modelica)
+set(CMAKE_C_STANDARD 99)
+
+include_directories(tree-sitter/src)
+add_library(treesitter_modelica SHARED tree-sitter/src/parser.c)
+
+install(TARGETS treesitter_modelica DESTINATION lib CONFIGURATIONS Release)
```
imciner2:
```suggestion
include(GNUInstallDirs)
install(TARGETS treesitter_modelica DESTINATION "${CMAKE_INSTALL_LIBDIR}" CONFIGURATIONS Release)
```
Windows doesn't put libraries into the `lib` folder, it actually wants them inside `bin`. The best way of handling this is to use the CMake module that sets the install directories and then use this in the `install` command.

T/tree_sitter_modelica/build_tarballs.jl
m-c-marius: It is the same logic as in most of the other tree sitter libraries. I'll remove the guard if it is an issue. Should I make another PR removing this guard in the other tree sitter libraries as well?

T/tree_sitter_modelica/build_tarballs.jl
m-c-marius: Same reason as above. I removed it.

T/tree_sitter_modelica/bundled/CMakeLists.txt
m-c-marius: Thanks! I saw the issue in the logs as well, but didn't know how `GNUInstallDirs` works on Windows.

T/tree_sitter_modelica/bundled/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+cmake_minimum_required(VERSION 3.13)
+project(treesitter_modelica)
+set(CMAKE_C_STANDARD 99)
+
+include_directories(tree-sitter/src)
+add_library(treesitter_modelica SHARED tree-sitter/src/parser.c)
+
+include(GNUInstallDirs)
+install(TARGETS treesitter_modelica DESTINATION "${CMAKE_INSTALL_LIBDIR}" CONFIGURATIONS Release)
```
imciner2:
```suggestion
install(TARGETS treesitter_modelica CONFIGURATIONS Release
    LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
    ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
    RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}")
```
Sorry, forgot we have to list these out separately.

T/tree_sitter_modelica/build_tarballs.jl
imciner2: The other recipes are older, so I guess at that time people wanted to do that. Now we prefer to have the script break if it doesn't find things so we can see it didn't work.

There is no need to go into the other scripts to change this right now. If they ever get updated again, we can touch them then.

imciner2 (approved):

imciner2:
Looks good. Thanks for your work on this. The package will be added to the General registry soon and then be installable as normal.

PR #11816, opened 2025-08-11
Enable broadway backend

VarLad:
Enables HTML backend for GTK4


VarLad:
The broadway backend is to be deprecated at a later date.
I'll be closing this PR.

PR #11822, opened 2025-08-12
SCS: update to version 3.2.8

araujoms:


araujoms:
Build is failing because SCS now uses the compiler flag `Werror=incompatible-pointer-types`. It is supported by GCC 5 upwards, but the build script is using GCC 4.

giordano:
That's easy to fix by adding `preferred_gcc_version=v"5"` to `build_tarballs(...)`

araujoms:
Thanks a lot, I was going through the docs to find where to change this.

araujoms:
There's three failing architectures; I think RISC-V and FreeBSD on aarch64 can be safely ignored, but I don't know about CUDA on Linux. I see that all versions other than 13 succeed.

giordano:
https://github.com/JuliaPackaging/Yggdrasil/blob/b670a072e4be75f2f79d7a6bb586ecfed84f9ea3/S/SCS/build_tarballs.jl#L37 should be changed to
```julia
 Dependency("OpenBLAS32_jll"),
```
and probably that will need to use gcc 8 to make powerpc64le happy.

araujoms:
Indeed, that worked. CUDA 13 still fails, though. Is that relevant?

giordano:
We can't merge PRs without any failing status, so it's relevant. The error is
```
[16:48:34] In file included from /workspace/destdir/cuda/include/cublas_api.h:74:0,
[16:48:34]                  from /workspace/destdir/cuda/include/cublas_v2.h:69,
[16:48:34]                  from linsys/gpu/gpu.h:14,
[16:48:34]                  from linsys/gpu/indirect/private.h:10,
[16:48:34]                  from linsys/gpu/indirect/private.c:1:
[16:48:34] /workspace/destdir/cuda/include/driver_types.h:59:30: fatal error: crt/host_defines.h: No such file or directory
[16:48:34] compilation terminated.
```
I don't know if @maleadt has some ideas, otherwise the easiest would be to drop cuda 13 for the time being (it was released like last week).

imciner2:
IMO, drop CUDA 13 for now. It is a major update and so it might require more work to get things building with it. I would suggest opening an issue upstream to let them know there is an error there.

araujoms:
Thanks, done. All green now.

giordano:
Thanks for your contribution!

odow:
I wonder if we should have an OWNERS file (or a comment header) for JLLs.

@kalmarek wanted some other changes but we didn't want to split the JLL/upstream version numbers: https://github.com/JuliaPackaging/Yggdrasil/pull/9608. I guess we can wait until the next upstream release.

giordano:
I'm not opposed to CODEOWNERS, but I seem to remember that requires the users to be committers and I don't think we want to write access to all contributors, only few people would be able to be added as owners.

odow:
Even just a convention for comment at the top of the file to ping when making a PR.

odow:
See https://github.com/JuliaPackaging/Yggdrasil/pull/11916

PR #11833, opened 2025-08-13
[CI] Downgrade HistoricalStdlibVersions to v2.0.2

lgoettgens:
This is a short-term fix for building recipes with libjulia as a dependency for `julia_version+1.13.0`, that currently error with 
```julia
ERROR: LoadError: KeyError: key UUID("3161d3a3-bdf6-5164-811a-617609db77b4") not found
```
e.g. in https://github.com/JuliaPackaging/Yggdrasil/pull/11832

cc @giordano

### DIFF
--- a/.ci/Project.toml
+++ b/.ci/Project.toml
@@ -3,6 +3,7 @@ BinaryBuilder = "12aac903-9f7c-5d81-afc2-d9565ea332ae"
 BinaryBuilderBase = "7f725544-6523-48cd-82d1-3fa08ff4056e"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
+HistoricalStdlibVersions = "6df8b67a-e8a0-4029-b4b7-ac196fe72102"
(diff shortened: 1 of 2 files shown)


giordano (changes requested):
You need to change the compat bound in Project.toml, otherwise this will be overridden again next time we update this.

.ci/Project.toml
```diff
@@ -11,5 +11,9 @@ TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 ghr_jll = "07c12ed4-43bc-5495-8a2a-d5838ef8d533"
 squashfs_tools_jll = "eed32e3e-a7c5-5bf9-9121-5cf3ab653887"
 
+[weakdeps]
+HistoricalStdlibVersions = "6df8b67a-e8a0-4029-b4b7-ac196fe72102"
```
giordano: Why weakdep?  It doesn't even exist in v1.7.

lgoettgens: I didn't want to introduce a new dep just for setting a compat bound. But of course I forgot that we are stuck with an ancient julia version here (as I always do :sweat_smile: )

giordano (approved):

PR #11850, opened 2025-08-15
Add iceberg_rust_ffi package

vustef:
Setting up a new JLL build for `RelationalAI/iceberg_rust_ffi`, which is going to be an FFI for Apache Iceberg Rust library.

### DIFF
--- a/I/iceberg_rust_ffi/build_tarballs.jl
+++ b/I/iceberg_rust_ffi/build_tarballs.jl
@@ -1,0 +1,45 @@
+using BinaryBuilder
+
+name = "iceberg_rust_ffi"
+version = v"0.1.0"
+
+sources = [
+    # GitSource("https://github.com/RelationalAI/iceberg_rust_ffi.git", "bebf0f4244b522e3b7412f895d28e00675bbedb2"),
+    DirectorySource("/Users/vstefanovic/ws/iceberg_rust_ffi/"; target="iceberg_rust_ffi"),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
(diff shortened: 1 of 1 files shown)


I/iceberg_rust_ffi/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+    LibraryProduct("libiceberg_rust_ffi", :libiceberg_rust_ffi),
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[]
+
+# Build the tarballs
+build_tarballs(
+    ARGS, name, version, sources, script, platforms, products, dependencies;
+    compilers=[:c, :rust], julia_compat="1.10", preferred_gcc_version=v"5", dont_dlopen=true,
```
giordano, line 41: Any reason for not testing `dlopen` of the library during audit?  Do you expect it to fail?

andrebsguedes: We require Julia v1.10 so it will complain about missing symbols like `jl_adopt_thread`. Is CI running on v1.10 now?

giordano: > We require Julia v1.10 so it will complain about missing symbols like jl_adopt_thread.

Ah, ok.

> Is CI running on v1.10 now?

Not yet, but we may switch to 1.12 at some point.

vustef:
> Thanks for your contribution! Looks good apart from the `dont_dlopen` option

Thank you for the prompt merge @giordano!

PR #11851, opened 2025-08-15
[Singular] update to 4.4.1p1

hannes14:
@fingolfin

### DIFF
--- a/S/Singular/build_tarballs.jl
+++ b/S/Singular/build_tarballs.jl
@@ -28,8 +28,8 @@ import Pkg.Types: VersionSpec
 #
 name = "Singular"
 
-upstream_version = v"4.4.1-1" # 4.4.1
-version_offset = v"0.0.9"
+upstream_version = v"4.4.1-3" # 4.4.1
+version_offset = v"0.0.10"
 
 version = VersionNumber(upstream_version.major * 100 + upstream_version.minor + version_offset.major,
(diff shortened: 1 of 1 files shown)


S/Singular/build_tarballs.jl
```diff
@@ -28,16 +28,16 @@ import Pkg.Types: VersionSpec
 #
 name = "Singular"
 
-upstream_version = v"4.4.1-1" # 4.4.1
-version_offset = v"0.0.9"
+upstream_version = v"4.4.1-3" # 4.4.1
+version_offset = v"0.0.10"
```
lgoettgens:
```suggestion
version_offset = v"0.0.0"
```
I think after bumping the upstream version, you can reset the offset

S/Singular/build_tarballs.jl
```diff
@@ -116,4 +116,4 @@ dependencies = [
 
 # Build the tarballs, and possibly a `build.jl` as well.
 build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
-    preferred_gcc_version=v"6", julia_compat = "1.6")
+    preferred_gcc_version=v"6", julia_compat = "1.10")
```
fingolfin, line 119: Just to make it explicit: after discussion with @lgoettgens I bumped the `julia_compat` here to 1.10, as that's where we want to use `Singular_jll` going forward.

PR #11852, opened 2025-08-15
[libsingular_julia] Increase julia compat to 1.10

lgoettgens:
Needs https://github.com/JuliaPackaging/Yggdrasil/pull/11851 to be released first.

cc @hannes14 @fingolfin


lgoettgens:
This is ready to merge now

PR #11857, opened 2025-08-15
[GAP_pkg_juliainterface] Rebuild with GAP_jll 400.1401.x, julia compat 1.10

lgoettgens:
cc @fingolfin


G/GAP_pkg/GAP_pkg_juliainterface/build_tarballs.jl
```diff
@@ -9,9 +9,9 @@ using Pkg
 uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
-gap_version = v"400.1400.005"
+gap_version = v"400.1401.5"
 name = "JuliaInterface"
-upstream_version = "0.14.2" # when you increment this, reset offset to v"0.0.0"
+upstream_version = "0.15.0" # when you increment this, reset offset to v"0.0.0"
```
lgoettgens, line 14: I fixed a version difference here. This is the build that is supposed to land in GAP.jl v0.15.0, so I chose that version here as well. It does not really matter, as GAP.jl uses a `=` specifier

PR #11861, opened 2025-08-15
[Singular] Revert "update to 4.4.1p1 (#11851)"

lgoettgens:
#11851 seems to cause a lot of trouble downstream (see https://github.com/oscar-system/Oscar.jl/issues/5201 and the further failures of the form
```
      From worker 3:	  wrong range[1] in ideal/module _(0)
      From worker 3:	  error occurred in or before primdec.lib::primdecGTZ_i line 5743: `   return(convList(decomp_i(patchPrimaryDecomposition,i,#)));`
      From worker 3:	  leaving primdec.lib::primdecGTZ_i (5685)
``` in https://github.com/oscar-system/Oscar.jl/actions/runs/16991011239/job/48170353920?pr=5202#step:10:4645.

Thus, I would suggest that we immediately revert that change (but still bump the version number), so that Oscar users get basically the previous version. We can then re-land #11851 with a minor version change instead, so that we don't get this to Oscar users immediately but only via a manual SIngular.jl release.

### DIFF
--- a/S/Singular/build_tarballs.jl
+++ b/S/Singular/build_tarballs.jl
@@ -29,7 +29,7 @@ import Pkg.Types: VersionSpec
 name = "Singular"
 
 upstream_version = v"4.4.1-3" # 4.4.1
-version_offset = v"0.0.0"
+version_offset = v"0.0.1"
 
 version = VersionNumber(upstream_version.major * 100 + upstream_version.minor + version_offset.major,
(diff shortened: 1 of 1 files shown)


lgoettgens:
Even the Singular.jl testsuite fails: https://github.com/oscar-system/Singular.jl/actions/runs/16991268344/job/48170975497?pr=876#step:7:490. One more reason for me to not have this as a patch release, but a minor one

PR #11756, opened 2025-07-31
bump LCIO version and pray

Moelf:

PR #11865, opened 2025-08-15
[Ghostscript] Rebuild to pull all new architectures

ViralBShah:


giordano (changes requested):
Previous build is from over 2 years ago, this must change the version number because compat bound of jllwrapper changed in the meantime (late December 2024)

ViralBShah:
Any idea how to fix the failing builds?

giordano:
Error is
```
[22:52:01] In file included from ./zlib/zutil.c:10:
[22:52:01] In file included from ./zlib/gzguts.h:21:
[22:52:01] In file included from /opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root/usr/include/c++/v1/stdio.h:108:
[22:52:01] /opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root/usr/include/stdio.h:210:7: error: expected identifier or '('
[22:52:01]   210 | FILE    *fdopen(int, const char *) __DARWIN_ALIAS_STARTING(__MAC_10_6, __IPHONE_2_0, __DARWIN_ALIAS(fdopen));
[22:52:01]       |          ^
[22:52:01] ./zlib/zutil.h:140:33: note: expanded from macro 'fdopen'
[22:52:01]   140 | #        define fdopen(fd,mode) NULL /* No fdopen() */
[22:52:01]       |                                 ^
```
This is the same error as in #11756, where someone was vendoring a zlib from last decade: https://github.com/iLCSoft/LCIO/issues/208. Solution was not to use the ancient vendored zlib and instead use the one from zlib_jll.

ViralBShah:
How does one fix this one?

```
In file included from ./libpng/png.c:14:
./libpng/pngpriv.h:527:16: fatal error: 'fp.h' file not found
  527 | #      include <fp.h>
      |                ^~~~~~
1 error generated.
make: *** [base/png.mak:94: obj/png.o] Error 1
```

ViralBShah:
I need to patch that to just be `math.h`. Tryign to see if we can just use our own libpng.

ViralBShah:
@giordano This is good to merge. Uses our zlib now everywhere.

ViralBShah:
The ImageMagick build could not find the gs libraries. Should I move the libraries on windows to a different directory (I think we do that in some packages)?

imciner2:
Windows should have the libraries in `bin`, and it looks like they already are here. (I think the recipes moving the libraries are moving them from `lib` to `bin`).

As for why imagemagik can't find it, that is weird, and I didn't see anything in the logs there. The configure script seems to find it, but the linker doesn't see it.

ViralBShah:
I think I need an approval by anyone with write access here to be able to merge.

PR #11869, opened 2025-08-16
cuNumeric.jl C/C++ wrapper

ejmeitz:
JLL for the C/C++ wrappers for cuNumeric.jl

### DIFF
--- a/C/cunumeric_jl_wrapper/build_tarballs.jl
+++ b/C/cunumeric_jl_wrapper/build_tarballs.jl
@@ -1,0 +1,115 @@
+using BinaryBuilder, Pkg
+
+# needed for libjulia_platforms and julia_versions
+const YGGDRASIL_DIR = "../../"
+include(joinpath(YGGDRASIL_DIR, "L", "libjulia", "common.jl"))
+include(joinpath(YGGDRASIL_DIR, "platforms", "cuda.jl"))
+
+name = "cunumeric_jl_wrapper"
+version = v"25.5" # cupynumeric has 05, but Julia doesn't like that
+sources = [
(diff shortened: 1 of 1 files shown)


ejmeitz:
Welp I'm a bit confused, the thing literally installed and the `.so.0.0.1` is opened by the checking script so how is the library not found....

barche:
> Welp I'm a bit confused, the thing literally installed and the `.so.0.0.1` is opened by the checking script so how is the library not found....

I think that on Linux BinaryBuilder also tries to dlopen the library, maybe that is failing for some reason?

C/cunumeric_jl_wrapper/build_tarballs.jl
```diff
@@ -0,0 +1,118 @@
...
+MIN_JULIA_VERSION = v"1.10"
+MAX_JULIA_VERSION = v"1.11.999"
+
+# These should match the cupynumeric_jll build_tarballs script
+MIN_CUDA_VERSION = v"12.2"
+MAX_CUDA_VERSION = v"12.8.999"
+
+script = raw"""
+    # Put new CMake first on path
+    export PATH=${host_bindir}:$PATH
```
imciner2, line 24: Generally, we recommend just doing `apk del cmake` to remove the host CMake version to prevent any confusion between the two.

ejmeitz:
@giordano Ready to merge I believe. This script is more or less a duplicate of the [legate_jl_wrapper ](https://github.com/JuliaPackaging/Yggdrasil/blob/master/L/legate_jl_wrapper/build_tarballs.jl) one.

PR #11897, opened 2025-08-20
[OpenBLAS] Backport patch to 0.3.23 to fix ASUM error

imciner2:
This backports the upstream commit that fixes the ASUM computation error reported in https://github.com/JuliaLang/LinearAlgebra.jl/issues/1406 to 0.3.23 so it can be included in Julia 1.10.


imciner2:
I have backported an additional commit: https://github.com/OpenMathLib/OpenBLAS/commit/f8ad5344c210960fc399ca5b0ad8559ab5ca253e. This actually fixes the error in the `casum` function that was being used originally. The first one just masked the error because it switched to using the optimized AVX512 kernel instead.

ViralBShah (approved):

PR #11898, opened 2025-08-20
make qdrant server available via jll

bdklahn:


bdklahn:
I'm not sure what the issue is.

This builds fine from BinaryBuilder running in Julia 1.7, installed via juliaup, on my OpenSUSE Tumbleweed host running on a Dell Precision 64-bit.

giordano:
Any reason for using prebuilt binaries instead of compiling it from source?

bdklahn:
> Any reason for using prebuilt binaries instead of compiling it from source?

Yes, realize this is not really a "Builder".
I started down that road and it seemed like it was going to take a long time to work through all the gotchas, with Vulkan, mold, etc.
So I thought I'd start with this, first.
Also, I wondered if this was going to be very taxing to the build service. It seemed a lot of things (deps), with Rust, end up getting fully compiled from  scratch when building "release". Upstream apparently uses something called chef to work around that, and cache some parts in Docker layers.
https://github.com/qdrant/qdrant/blob/master/Dockerfile

But it would be nice to, in the future, be able to mostly just bump the version number, rather have to update all the hashes, etc. -and I think it would be cool if we could even pick up another platform or two (but I'm doubtful).

Idk. I wasn't sure we'd actually start using it, with most of our devs narrowly-focused on Python-only.
I was playing with BB locally, and figured I'd just share.

No problem if you don't want to pull, and wait until someone can contribute a proper working Rust build.

giordano:
> But it would be nice to, in the future, be able to mostly just bump the version number, rather have to update all the hashes.

No, completely ignoring any checksumming is a terrible idea from a security point of view.

Q/qdrant/build_tarballs.jl
```diff
@@ -0,0 +1,48 @@
+using BinaryBuilder, Pkg
+
+# Collection of pre-build quarto binaries
+name = "qdrant"
```
giordano: Seems to be spelled
```suggestion
name = "Qdrant"
```
upstream: https://github.com/qdrant/qdrant/blob/a7d21f168097494acc2cbfb2489cfd2da40581ac/README.md.  You also need to rename the directory accordingly

bdklahn:
> > But it would be nice to, in the future, be able to mostly just bump the version number, rather have to update all the hashes.
> 
> No, completely ignoring any checksumming is a terrible idea from a security point of view.

Of course I/we should always use the hashes to verify each tarball.

I'm just saying that, with this "pull already-built binaries" pattern, I have to copy/paste a bunch of new hashes on update.
If I did a proper build, using a single GitSource line, I'd only have to update the version and the gitish vs. like 6 different things. :-)

Q/qdrant/build_tarballs.jl
bdklahn: I do see that their branding/logo is capitalized.
I was just going by the fact that they chose to make their repo name lowercase.

I suppose, then, the Quarto one was also named wrongly.

PR #11916, opened 2025-08-21
[SCS] add CODEOWNERS comment

odow:
I pointed out in https://github.com/JuliaPackaging/Yggdrasil/pull/11822 that it would be useful to have some sort of standardized mechanism (even if just by convention) for who to contact when making a PR.

Here's one option. If we like it, I can add some others, and add a sentence about it to the docs


odow:
Closing in favor of #11917

PR #11917, opened 2025-08-22
Add .github/workflows/component_owners.yml GitHub action

odow:
An alternative to https://github.com/JuliaPackaging/Yggdrasil/pull/11916

Suggested by @imciner2 

image

### DIFF
--- a/.github/workflows/component_owners.yml
+++ b/.github/workflows/component_owners.yml
@@ -1,0 +1,28 @@
+# The purpose of this workflow is to automaticallly assign reviewers for PRs
+# which modify different parts of Yggdrasil.
+#
+# The file that controls the reviewers is at `.github/component_owners.yml`
+#
+# See https://github.com/dyladan/component-owners for more details of this
+# action.
+name: 'Component Owners'
+on:
+  # pull_request_target is suggested for projects where pull requests will be
+  # made from forked repositories. If pull_request is used in these cases,
+  # the github token will not have sufficient permission to update the PR.
+  pull_request_target:
+permissions:
+  contents: read          # to read changed files
+  issues: write           # to read/write issue assignees
+  pull-requests: write    # to read/write PR reviewers
+jobs:
+  run_self:
+    runs-on: ubuntu-latest
+    name: Auto Assign Owners
+    steps:
+      - uses: dyladan/component-owners@main
+        with:
+          config-file: .github/component_owners.yml
+          repo-token: ${{ github.token }} 
+          assign-owners: "true"
+          request-owner-reviews: "true"
--- a/.github/component_owners.yml
+++ b/.github/component_owners.yml
@@ -1,0 +1,31 @@
+# The purpose of this file is to control which reviewers are assigned as part of
+# the `.github/workflows/component_owners.yml` workflow.
+#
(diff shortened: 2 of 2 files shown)


.github/workflows/component_owners.yml
```diff
@@ -0,0 +1,28 @@
...
+permissions:
+  contents: read          # to read changed files
+  issues: write           # to read/write issue assignees
+  pull-requests: write    # to read/write PR reviewers
+jobs:
+  run_self:
+    runs-on: ubuntu-latest
+    name: Auto Assign Owners
+    steps:
+      - uses: dyladan/component-owners@main
```
giordano: Use a commit hash

.github/workflows/component_owners.yml
```diff
@@ -0,0 +1,28 @@
...
+# action.
+name: 'Component Owners'
+on:
+  # pull_request_target is suggested for projects where pull requests will be
+  # made from forked repositories. If pull_request is used in these cases,
+  # the github token will not have sufficient permission to update the PR.
+  pull_request_target:
+permissions:
+  contents: read          # to read changed files
+  issues: write           # to read/write issue assignees
```
giordano: What do we need to do with issues?

.github/workflows/component_owners.yml
odow:
```suggestion
      - uses: dyladan/component-owners@ 58bd86e9814d23f1525d0a970682cead459fa783
```

.github/workflows/component_owners.yml
odow:
```suggestion
```

Not sure. I just copied the template: https://github.com/dyladan/component-owners

.github/workflows/component_owners.yml
giordano:
```suggestion
      - uses: dyladan/component-owners@ 58bd86e9814d23f1525d0a970682cead459fa783 # v0.1.0
```

giordano:
I'm not familiar with this workflow, but we can see how it goes.

odow:
> I'm not familiar with this workflow, but we can see how it goes.

Agreed. We can revert if it's a pain

.github/workflows/component_owners.yml
```diff
@@ -0,0 +1,27 @@
...
+  run_self:
+    runs-on: ubuntu-latest
+    name: Auto Assign Owners
+    steps:
+      - uses: dyladan/component-owners@ 58bd86e9814d23f1525d0a970682cead459fa783 # v0.1.0
+        with:
+          config-file: .github/component_owners.yml
+          repo-token: ${{ github.token }} 
+          assign-owners: "true"
+          request-owner-reviews: "true"
```
imciner2, line 27: I think with how we want to use this, we don't need to have it request the reviews. This can more serve as a notification system by just assigning the people, I don't think there is a need to have it actively request the review.

Thoughts?

odow: Not sure. I just kept the defaults.

giordano: I agree the main point should be pinging the maintainers and assignment is sufficient for that.

PR #11921, opened 2025-08-22
[GAP_pkg_semigroups] Enable riscv again

lgoettgens:
This had been disabled in https://github.com/JuliaPackaging/Yggdrasil/pull/11856 due to a build failure, but trying it once again locally it just seems to build fine.

cc @fingolfin


lgoettgens:
This now works (thanks to Mosè's help on slack)

PR #11925, opened 2025-08-22
[Sundials] Build 7.4 with 32 bit LAPACK And BLAS

ViralBShah:
@imciner2 This is the recipe I need to get it to build.

### DIFF
--- a/S/Sundials/Sundials@7/build_tarballs.jl
+++ b/S/Sundials/Sundials@7/build_tarballs.jl
@@ -38,7 +38,9 @@ cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DKLU_WORKS=ON \
     -DENABLE_LAPACK=ON \
     -DLAPACK_WORKS=ON \
-    -DBLA_VENDOR="libblastrampoline" \
+    -DBLA_VENDOR="libopenblas" \
+    -DBLAS_LIBRARIES="${libdir}/libopenblas.${dlext}" \
+    -DLAPACK_LIBRARIES="${libdir}/libopenblas.${dlext}" \
     ..
 
 cmake --build . --parallel ${nproc}
@@ -80,8 +82,8 @@ products = [
 dependencies = [
     HostBuildDependency("CMake_jll"),
     Dependency("CompilerSupportLibraries_jll"),
-    Dependency("libblastrampoline_jll"),
-    Dependency("SuiteSparse_jll"),
+    Dependency("OpenBLAS32_jll"),
+    Dependency("SuiteSparse32_jll"),
 ]
 
 # Build the tarballs


imciner2:
I haven't had a chance to look fully into this, so this is just some initial comments.

1) When changing the `julia_compat` I believe we will need to also bump the version to make the registry happy. (I don't know if that might also be needed because the dependencies are changing?)

2) I think we need to tell SUNDIALS to build in 32-bit mode as well, because it is defaulting to 64-bit indexes, and so I think there will be an issue with interfacing SUNDIALS built with 64-bit indices and OpenBLAS with 32-bit indices. I think you just need to add a CMake config variable for `SUNDIALS_INDEX_SIZE=32` to get it to build with 32-bit indices.

ViralBShah:
> 2. I think we need to tell SUNDIALS to build in 32-bit mode as well, because it is defaulting to 64-bit indexes, and so I think there will be an issue with interfacing SUNDIALS built with 64-bit indices and OpenBLAS with 32-bit indices. I think you just need to add a CMake config variable for `SUNDIALS_INDEX_SIZE=32` to get it to build with 32-bit indices.

That did not work. Using this:
```
index 0c7c8c53f..e69935d77 100644
--- a/S/Sundials/Sundials@7/build_tarballs.jl
+++ b/S/Sundials/Sundials@7/build_tarballs.jl
@@ -2,6 +2,7 @@ using BinaryBuilder
 
 name = "Sundials"
 version = v"7.4.0"
+ygg_version = v"7.4.1" # Fake version since we are changing deps
 
 # Collection of sources required to build Sundials
 sources = [
@@ -39,8 +40,7 @@ cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DENABLE_LAPACK=ON \
     -DLAPACK_WORKS=ON \
     -DBLA_VENDOR="libopenblas" \
-    -DBLAS_LIBRARIES="${libdir}/libopenblas.${dlext}" \
-    -DLAPACK_LIBRARIES="${libdir}/libopenblas.${dlext}" \
+    -DSUNDIALS_INDEX_SIZE=32 \
     ..
 
 cmake --build . --parallel ${nproc}
@@ -87,5 +87,5 @@ dependencies = [
 ]
 
 # Build the tarballs
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies; 
+build_tarballs(ARGS, name, ygg_version, sources, script, platforms, products, dependencies; 
                preferred_gcc_version = v"6", julia_compat="1.6")

```

produces

```
=FALSE to (re)run compatibility test.
[23:20:03] CMake Error at /workspace/x86_64-linux-musl-cxx11/artifacts/db67336af9bf24e2b9519fcc67d9b934e5647add/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
[23:20:03]   Could NOT find BLAS (missing: BLAS_LIBRARIES)
[23:20:03] Call Stack (most recent call first):
[23:20:03]   /workspace/x86_64-linux-musl-cxx11/artifacts/db67336af9bf24e2b9519fcc67d9b934e5647add/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
```

imciner2:
> That did not work. Using this:

The CMake BLAS name needs to be the exact one from this list: https://cmake.org/cmake/help/latest/module/FindBLAS.html#blas-lapack-vendors. In this case, for openblas, we actually want it to be
```
-DBLA_VENDOR="OpenBLAS"
```

ViralBShah:
Looks good to merge now.

PR #11928, opened 2025-08-23
[coveralls] Add recipe

IanButterworth:
A recipe for https://github.com/coverallsapp/coverage-reporter which we need a macOS binary for, and they only serve it via homebrew.

### DIFF
--- a/C/coveralls/build_tarballs.jl
+++ b/C/coveralls/build_tarballs.jl
@@ -34,8 +34,7 @@ apk add --no-cache \
     libevent-dev \
     libevent-static \
     gc-dev \
-    libxml2-dev \
-    libxml2-static
+    libxml2-dev
 
 # Set environment variables for static linking
 export CRYSTAL_CACHE_DIR=${WORKSPACE}/crystal_cache


C/coveralls/build_tarballs.jl
```diff
@@ -0,0 +1,89 @@
...
+# Bash recipe for building across all platforms
+script = raw"""
+cd ${WORKSPACE}/srcdir/coverage-reporter
+
+# Update package database and install Crystal compiler
+apk update
+apk add --no-cache \
+  -X https://dl-cdn.alpinelinux.org/alpine/edge/main \
+  -X https://dl-cdn.alpinelinux.org/alpine/edge/community \
+  'crystal=1.8.0-r0' shards
```
IanButterworth:
```suggestion
apk add --no-cache \
  -X https://dl-cdn.alpinelinux.org/alpine/v3.18/main \
  -X https://dl-cdn.alpinelinux.org/alpine/v3.18/community \
  crystal=1.8.2-r0 shards
```

IanButterworth:
@vtjnash this yack shave isn't progressing easily here either..

vtjnash:
Good point, seems like there isn't really a supported build environment at all for building coveralls on macos. It sort of hacks around with some zig, docker, and bash scripts to download homebrew stuff, but seems quite a mess. New theory: we should steal this script to run here and set up a new repo to run this as a github action on linux: https://github.com/luislavena/crystal-xbuild-container/blob/main/scripts/homebrew-downloader.cr to get it to download everything (it is where coverage-reporter upstream got their unsupported cross-compile for macos docker script), without ever installing homebrew, and just re-upload the required files as a single tarball.

PR #11932, opened 2025-08-25
[CUDA_Compiler_jll] Add a runtime dependency on CUDA_Runtime_jll

vchuravy:
As discussed in https://github.com/JuliaLang/julia/issues/59257 we can't declare compile time preferences on preferences not owned by the package.
We must therefore declare a RuntimeDependency (like we do on MPIPreference) on CUDA_Runtime_jll so that CUDA_Compiler_jll get's invalidated on preference changes.

### DIFF
--- a/C/CUDA/CUDA_Compiler/build_tarballs.jl
+++ b/C/CUDA/CUDA_Compiler/build_tarballs.jl
@@ -59,7 +59,8 @@ fi
 """
 
 dependencies = [
-    Dependency("CUDA_Driver_jll", v"13.0"; compat="13")
+    Dependency("CUDA_Driver_jll", v"13.0"; compat="13"),
+    RuntimeDependency("CUDA_Runtime_jll"),
 ]
 
 products = [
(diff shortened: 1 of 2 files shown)


vchuravy:
Should fix https://github.com/JuliaGPU/CUDA.jl/issues/2852

C/CUDA/CUDA_Compiler/build_tarballs.jl
```diff
@@ -59,7 +59,8 @@ fi
 """
 
 dependencies = [
-    Dependency("CUDA_Driver_jll", v"13.0"; compat="13")
+    Dependency("CUDA_Driver_jll", v"13.0"; compat="13"),
```
giordano: For what is worth, specifying the build version is redundant, since that's inferred automatically from the lowest version matching the range specified by `compat` (and removing the extra arguments reduces the place where the same information has to be consistently set)
```suggestion
    Dependency("CUDA_Driver_jll"; compat="13"),
```

PR #11930, opened 2025-08-23
Do not merge: a test for #11917

odow:
Don't merge this. I want to test the workflow we added in https://github.com/JuliaPackaging/Yggdrasil/pull/11917

PR #11934, opened 2025-08-25
Clarify that component_owners must be collaborators

odow:
x-ref https://github.com/JuliaPackaging/Yggdrasil/pull/11930

### DIFF
--- a/.github/component_owners.yml
+++ b/.github/component_owners.yml
@@ -10,6 +10,8 @@
 components:
   C/Coin-OR/:
     - odow
+  C/Coin-OR/Ipopt/:
+    - amontoison
   G/GLPK/:
     - odow
   H/HiGHS/:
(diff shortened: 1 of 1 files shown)


.github/component_owners.yml
```diff
@@ -4,11 +4,12 @@
...
 #
 # Please keep the list of components and reviewers in alphabetical order.
+#
+# Note that users MUST be collaborators on the JuliaPackaging/Yggdrasil
+# repository, but they do not need write access.
 components:
   C/Coin-OR/:
     - odow
-  C/Coin-OR/Ipopt/:
-    - amontoison
```
amontoison: You can keep me here @odow

.github/component_owners.yml
```diff
@@ -18,14 +21,10 @@ components:
   S/SCIP_PaPILO/:
     - odow
   S/SCS/:
-    - kalmarek
```
giordano, line 21: I believe @kalmarek received and accepted the invite to be a collaborator

kalmarek: still it would be nice to be pinged when something happens here ;)

giordano: Please add you back for the packages you're interested in.

PR #11940, opened 2025-08-26
Add lgoettgens, fingolfin, benlorenz to component_owners.yml

lgoettgens:
I took the liberty to add @fingolfin and @benlorenz to some recipes as well (you two, please check if you indeed want to be notified for these).


lgoettgens:
@giordano could you please also invite me as a read-only collaborator to yggdrasil? Thanks

giordano:
Done

PR #11941, opened 2025-08-26
Update collaborators file for imciner2

imciner2:


giordano:
Can't use skip ci in a pull request, otherwise the mandatory CI job never runs, that can only be used on merge commits.

PR #11967, opened 2025-08-28
[RootFs/Go] Update to 1.25.0

lamdor:
Like https://github.com/JuliaPackaging/Yggdrasil/pull/9267, but update to the newest.


lamdor:
This will need a companion PR to update BinaryBuilderBase as well after the artifacts are published.

PR #11975, opened 2025-08-29
legate_jl_wrapper_jll v25.5.1

krasow:
Bug fixes for registering custom tasks with Legate.jl.

Disclaimer: This is prerelease software. We are currently activately developing legate_jl_wrapper_jll and Legate.jl.


ejmeitz:
lgtm

PR #11982, opened 2025-08-31
Wizard recipe: libmxml-v4.0.4

vlad-circadia:

* Package name: libmxml
* Version: v4.0.4


L/libmxml/build_tarballs.jl
```diff
@@ -0,0 +1,53 @@
...
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("riscv64", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("x86_64", "macos"; ),
+    Platform("aarch64", "macos"; )
+]
```
imciner2:
```suggestion
platforms = supported_platforms()
filter!(p -> (Sys.islinux(p) || Sys.isapple(p)), platforms)
```
You only seem to be building for Linux and Apple, so this is the cleaner way of representing this.

Also, what would be the issues with building on the other platforms? Does upstream not support them?

imciner2 (changes requested):

L/libmxml/build_tarballs.jl
vlad-circadia: Fair point. It's my first JLL and I wasn't thinking much about other platforms I'm not using. 
Let me fix that.

imciner2:
Actually, the only last thing would be the name of this. We generally like to follow the upstream naming, so since it is known as `mxml` upstream, it is best called `mxml` here instead of `libmxml` (otherwise we would have a lot of things with `lib*` in the name).

vlad-circadia:
True. For some reason I thought the `lib` prefix would be the convention for packages providing libraries. 
I'll change the name then.

vlad-circadia:
I believe in order to change the package name, it's also required to rename the libmxml folder to mxml. Am I right?

giordano:
Correct

imciner2:
You need to move the file to the `M` directory as well.

vlad-circadia:
of course!

imciner2:
Thanks for your contribution!

PR #11985, opened 2025-09-01
FINUFFT v2.4.1

ludvigak:


C/cufinufft/build_tarballs.jl
```diff
@@ -53,9 +53,12 @@ filter!(p -> arch(p)=="x86_64", platforms)
 # cuFINUFFT does not compile with CUDA 12.5, so exclude
 filter!(p -> VersionNumber(p["cuda"]) != v"12.5", platforms)
 
-# CUDA 12.9 doesn't seem to build yet
+# CUDA 12.9 doesn't seem to build
 filter!(p -> VersionNumber(p["cuda"]) != v"12.9", platforms)
```
imciner2: Was this an upstream problem, or Ygg problem? I recall that when 12.9 was first added to Ygg, it didn't work with CMake due to the static libraries being missing, but that should be fixed now.

ludvigak: Good point, I haven't checked since then to be honest. Doing that right away...

ludvigak: Turns out it works now!

PR #11987, opened 2025-09-01
[GAP] Update to 4.15.0

lgoettgens:
This should not be merged, but instead wait for the actual 4.15.0 release. But I wanted to build a local copy already to test with Oscar, and wanted to make all changes to patches etc. available.

The update form the beta1 to the release should then be a minimal change to this PR once the release is out.

### DIFF
--- a/G/GAP/build_tarballs.jl
+++ b/G/GAP/build_tarballs.jl
@@ -26,13 +26,13 @@ uuid = Base.UUID("a83860b7-747b-57cf-bf1f-3e79990d037f")
 delete!(Pkg.Types.get_last_stdlibs(v"1.6.3"), uuid)
 
 name = "GAP"
-upstream_version = v"4.15.0-beta3"
+upstream_version = v"4.15.0"
 version = v"400.1500.000"
 
 # Collection of sources required to complete build
 sources = [
     ArchiveSource("https://github.com/gap-system/gap/releases/download/v$(upstream_version)/gap-$(upstream_version)-core.tar.gz",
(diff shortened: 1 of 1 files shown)


G/GAP/bundled/patches/sysinfo.patch
```diff
@@ -1,25 +1,24 @@
...
 --- a/Makefile.rules
 +++ b/Makefile.rules
-@@ -670,7 +670,7 @@ install-gaproot: CITATION
+@@ -671,7 +671,7 @@ install-gaproot: install-dirs CITATION
  
  # the following lines adjust variables for the installed sysinfo.gap
- install-sysinfo: SYSINFO_CPPFLAGS = -I${includedir}/gap -I${includedir} $(GAP_DEFINES)
+ install-sysinfo: SYSINFO_CPPFLAGS = -I${includedir}/gap/extra $(GAP_DEFINES)
 -install-sysinfo: SYSINFO_LDFLAGS = $(ABI_CFLAGS)
 +install-sysinfo: SYSINFO_LDFLAGS = $(ABI_CFLAGS) -L$(libdir) -lgap
```
lgoettgens, line 19: @fingolfin Is this patch something that could be upstreamed to gap? Or is this specific to the way GAP.jl compiles kernel extensions?

fingolfin: For now this is GAP.jl specific. Upstreaming this cleanly is a bit hairy, see https://github.com/gap-system/gap/pull/5780

lgoettgens:
@fingolfin this should be good to go now

PR #11988, opened 2025-09-01
[GAP_lib] Update to 4.15.0

lgoettgens:
Companion to https://github.com/JuliaPackaging/Yggdrasil/pull/11987.

This should not be merged, but instead wait for the actual 4.15.0 release. But I wanted to build a local copy already to test with Oscar, and wanted to make all changes to patches etc. available.

The update form the beta1 to the release should then be a minimal change to this PR once the release is out.


lgoettgens:
@fingolfin This should be good to go now (once CI finishes)

PR #11990, opened 2025-09-01
Wizard recipe: QOCO-v0.1.6

langestefan:

* Package name: QOCO
* Version: v0.1.6

### DIFF
--- a/Q/QOCO/build_tarballs.jl
+++ b/Q/QOCO/build_tarballs.jl
@@ -7,7 +7,7 @@ version = v"0.1.6"
 
 # Collection of sources required to complete build
 sources = [
-    GitSource("https://github.com/qoco-org/qoco.git", "bcd90673e9d5db372152f8ec726ae80d99831550")
+    GitSource("https://github.com/qoco-org/qoco.git", "df7192e38b09d5b952b2a15e071801354f7a5dfd")
 ]
 
 # Bash recipe for building across all platforms


langestefan:
@odow 

Here's the full log from step 4 onwards:

[yggdrasil_log.txt](https://github.com/user-attachments/files/22084635/yggdrasil_log.txt)

Q/QOCO/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Pkg
+
+name = "QOCO"
+version = v"0.1.6"
+
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/qoco-org/qoco.git", "6b38149481dffb431077396244403f3077686df1")
```
odow:
```suggestion
    GitSource("https://github.com/qoco-org/qoco.git", "c1d988da619fbbb32d24f3c66295c20cdc75d650")
```

The v0.1.6 commit is https://github.com/qoco-org/qoco/commit/c1d988da619fbbb32d24f3c66295c20cdc75d650.

It really helps if things are kept in sync. It'd be better also if upstream actually tagged releases?

Q/QOCO/build_tarballs.jl
```diff
@@ -0,0 +1,52 @@
...
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "glibc"),
+    Platform("powerpc64le", "linux"; libc = "glibc"),
+    Platform("riscv64", "linux"; libc = "glibc"),
+    Platform("i686", "linux"; libc = "musl"),
+    Platform("x86_64", "linux"; libc = "musl"),
+    Platform("aarch64", "linux"; libc = "musl"),
+    Platform("armv6l", "linux"; call_abi = "eabihf", libc = "musl"),
+    Platform("armv7l", "linux"; call_abi = "eabihf", libc = "musl")
+]
```
odow: This can probably just be
```
```

odow: Let's see all the failing builds. Not just the passing linux.

odow:
The error was further up from what you posted on slack. Looks like there are some missing imports upstream:

```
In file included from /workspace/srcdir/qoco/include/qoco_linalg.h:18:
/workspace/srcdir/qoco/include/structs.h:251:3: error: unknown type name 'QOCOTimer'
  251 |   QOCOTimer solve_timer;
      |   ^
In file included from /workspace/srcdir/qoco/src/input_validation.c:11:
In file included from /workspace/srcdir/qoco/include/input_validation.h:20:
In file included from /workspace/srcdir/qoco/include/structs.h:19:
/workspace/srcdir/qoco/include/timer.h:52:18: error: unknown type name 'QOCOTimer'
```

Q/QOCO/build_tarballs.jl
langestefan: Should be clear now

Q/QOCO/build_tarballs.jl
langestefan: Just released: https://github.com/qoco-org/qoco/releases/tag/v0.1.6

`6b38149481dffb431077396244403f3077686df1`

odow:
FreeBSD failures are because there's no "else" block here for other operating systems:
https://github.com/qoco-org/qoco/blob/6b38149481dffb431077396244403f3077686df1/include/timer.h#L20-L45

odow:
Windows failure is because of 
https://github.com/qoco-org/qoco/blob/6b38149481dffb431077396244403f3077686df1/CMakeLists.txt#L36
It's specific to MSVC. We can probably just remove it in favour of the existing `-O3`.

cc @govindchari

Q/QOCO/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+exit
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libqdldl", :qdldl),
```
imciner2: Is there a need to expose QDLDL in this recipe? We have the `QDLDL_jll` package already that is used by consumers who want the library (even though it is 0.1.5, nothing has actually changed in the algorithm part since then).

Q/QOCO/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+# Collection of sources required to complete build
+sources = [
+    GitSource("https://github.com/qoco-org/qoco.git", "6b38149481dffb431077396244403f3077686df1")
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cd qoco
+cmake -B build -DCMAKE_INSTALL_PREFIX=$prefix -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_STANDARD=99
```
imciner2:
```suggestion
cmake -B build \
    -DCMAKE_INSTALL_PREFIX=$prefix \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_C_STANDARD=99
```
Makes for more readable diffs in the future when options change.

Q/QOCO/build_tarballs.jl
```diff
@@ -0,0 +1,38 @@
...
+cmake --build build --parallel ${nproc}
+cmake --install build
+exit
```
imciner2:
```suggestion
```

Q/QOCO/build_tarballs.jl
langestefan: Apologies, I wasn't aware we already had QDLDL available. Seems okay with me: https://github.com/osqp/qdldl/compare/v0.1.5...v0.1.6

langestefan: The version of QDLDL in QOCO was modified to have dynamic regularization: https://github.com/qoco-org/qoco/commit/c90b3e62b77f15dbcd565b05b89e5acb1d59a665

If this is an overal improvement to the algorithm (I'm not sure how to judge that) perhaps we could merge that change upstream, release a new binary for QDLDL and then use it instead of v0.1.5? What do you think?

imciner2: You can keep using the bundled QDLDL, this comment was just about not exposing the version coming from here - which is what the `LibraryProduct` does. (actually, IMO QDLDL should be used as a static library instead of a shared library because it is basically just 5 functions and isn't worth the overhead of the shared library lookup and dispatch).

As for the dynamic regularization. I had heard feedback from ECOS a few years ago about wanting to switch to QDLDL and wanting dynamic regularization, but their switching stalled and so we never actually implemented it in QDLDL. I guess if there is more interest in having it in QDLDL proper I can look into it in the future.

langestefan: > You can keep using the bundled QDLDL, this comment was just about not exposing the version coming from here - which is what the LibraryProduct does.

Thanks! Got it, then I'll remove it.

> (actually, IMO QDLDL should be used as a static library instead of a shared library because it is basically just 5 functions and isn't worth the overhead of the shared library lookup and dispatch).

As in QOCO should make this change in its `CMakelists.txt?`

govindchari: qoco links against the qdldl static library when creating the qoco static library and links against the qdldl .so when creating the qoco .so

imciner2: > qoco links against the qdldl static library when creating the qoco static library and links against the qdldl .so when creating the qoco .so

@govindchari ideally, QDLDL should always be statically linked into OSQP no matter what OSQP's linkage is to your application. So in this case, since you are statically linking OSQP into the solver, you should be statically linking QDLDL as well.

Q/QOCO/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_STANDARD=99
+cmake --build build --parallel ${nproc}
+cmake --install build
+"""
+
+# These are the platforms we will build for by default, unless further
+# platforms are passed in on the command line
+platforms = supported_platforms()
```
odow, line 28:
```suggestion
platforms = supported_platforms()
platforms = filter!(!Sys.isfreebsd, platforms)
```

Q/QOCO/build_tarballs.jl
```diff
@@ -0,0 +1,41 @@
...
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct("libqdldl", :qdldl),
```
odow:
```suggestion
```

Q/QOCO/build_tarballs.jl
govindchari: Is this the change you are suggesting: https://github.com/qoco-org/qoco/pull/12?

Should I also be creating a static version of amd and linking against that?

imciner2: Yes, that is what I meant (sorry for the confusion above with OSQP - I was context switching too much last night and got confused :)). It looks like AMD is an `OBJECT` library, which means it will be similar enough to a static library and won't have this issue.

langestefan:
Comments have been addressed and all checks pass, so I think this is in a state to be merged now.

imciner2:
Looks good. Thanks for your patience while we worked out the linking.

langestefan:
Thanks for the review and answering questions!

PR #11996, opened 2025-09-03
Separate XGBoost_jll into regular and GPU package

BenCurran98:
This PR is to fix the issue referenced [here](https://github.com/JuliaPackaging/Yggdrasil/issues/11918) where users with CUDA v13 are unable to load XGBoost artifacts. The solution here is to separate XGBoost_jll into two packages: one with the "standard" (cpu only) package that will get loaded by default, and an optional GPU-compatible package XGBoost_GPU_jll which gets loaded if the parent XGBoost package finds the required software is available (CUDA, etc.)


BenCurran98:
@maleadt @tylerjthomas9 most of the builds are working, just stuck on a couple that I'm not sure how to fix. In particular, they seem broken on [CUDA v11.4](https://buildkite.com/julialang/yggdrasil/builds/23136/steps/canvas?sid=01990dca-868b-422a-8963-729783030cf3) where I get this error: 
image

Also for[ MacOS builds](https://buildkite.com/julialang/yggdrasil/builds/23136/steps/canvas?sid=01990dca-caea-4f06-a555-a271fe9dd1e1) it seems to have trouble linking to libomp? Again, not entirely familiar with this error so if you have any ideas I'd be happy to hear them :)
image

BenCurran98:
It looks like builds are only broken or CUDA v11.4 but work for 11.8, so I'll up the compatibility here. I feel like this is probably a restriction coming from XGBoost directly as they seem to have a PR in flux currently to backport some logic surrounding libcudacxx etc.

ViralBShah:
It would be nice to refactor some of the common parts into a common.jl to make it easier to upgrade versions in the future.

X/XGBoost/XGBoost/build_tarballs.jl
```diff
@@ -0,0 +1,42 @@
...
+script = raw"""
+cd ${WORKSPACE}/srcdir/xgboost
+git submodule update --init
+
+# builds on MacOS seem to fail with Clang - use gcc instead
+# see https://github.com/dmlc/xgboost/issues/11676
+export TOOLCHAIN_FILE="${CMAKE_TARGET_TOOLCHAIN}"
+if [[ "${target}" == *-apple-* ]]; then
+    export TOOLCHAIN_FILE="${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake"
+fi
```
giordano, line 19: Is this simply a matter of using a newer SDK?  Also, this broke x86_64 macOS: #12048

BenCurran98: I wasn't exactly sure. It looks like support for the latest compilers is only available on the next major release at the moment, which we are planning on releasing a build for soon

BenCurran98: [This](https://github.com/dmlc/xgboost/issues/11676) was the issue that brought me to try this change, although I'm not super confident with C/C++ so any insight would be great :)

PR #12003, opened 2025-09-03
Git: Update last Windows 32-bit version

mbauman:
This installer was made available "after warranty" to address a number of vulnerability advisories.

https://github.com/git-for-windows/git/releases/tag/v2.50.1.windows.1

Should this bump the patch version of the JLL itself?

### DIFF
--- a/G/Git/build_tarballs.jl
+++ b/G/Git/build_tarballs.jl
@@ -3,7 +3,7 @@
 using BinaryBuilder
 
 name = "Git"
-version = v"2.51.0"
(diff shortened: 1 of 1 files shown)


giordano:
Your change as-is will only bump the build number, is that OK?

eschnett:
I would update the patch number. This allows using all the version-number-specific tools we have, such as yanking, or avoiding versions, or requiring versions, etc.

mbauman:
Ohhh the irony!

image

PR #12004, opened 2025-09-03
 [LLVM_full] Add LLVM_full, LLVM_full_assert 21 recipes

Zentrik:


Zentrik:
Looks like lld on windows doesn't like --build-id=sha1 introduced in https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/435/files

Zentrik:
looks like lld mingw and lld coff don't support specifying the type, https://github.com/llvm/llvm-project/blob/e90e76e15dee4b835b9b2cfd55c0c3f047bb6d76/lld/MinGW/Driver.cpp#L329-L338, https://github.com/llvm/llvm-project/blob/e90e76e15dee4b835b9b2cfd55c0c3f047bb6d76/lld/COFF/Options.td#L328-L331.

giordano:
Should we update our fork to upstream v21.1.2 while we're here?

L/LLVM/common.jl
```diff
@@ -22,6 +22,7 @@ const llvm_tags = Dict(
     v"18.1.7" => "32719222d3ea71ed0b19c2cb75fa6f76713fda20", # julia-18.1.7-4
     v"19.1.7" => "ccda9ec62497d9de88ca7090a749e52a89f62132", # julia-19.1.7-2
     v"20.1.8" => "5b9f96366ce26dfc8ca91697ef0a57894791d95e", # julia-20.1.8-0
+    v"21.1.2" => "e01e3e96a51b18afc66b6b4ef358b3d72b51dc68", # julia-21.1.2-0
```
Zentrik, line 25: TODO: tag this

giordano:
Good to go (after tagging the commit)?

Zentrik:
Yh, I guess we'll see if we need any patches once someone builds Julia with
this.

L/LLVM/common.jl
giordano: Done: https://github.com/JuliaLang/llvm-project/releases/tag/julia-21.1.2-0

PR #12013, opened 2025-09-04
`AcousticsToolbox`: 2025.09.05

mchitre:
New version 2025.09.05:
- Patches `ORCA` to write modes and acoustic pressures to a file.

### DIFF
--- a/A/AcousticsToolbox/build_tarballs.jl
+++ b/A/AcousticsToolbox/build_tarballs.jl
@@ -23,7 +23,8 @@ make
 mkdir -p $bindir
 find . -name *.exe -exec cp {} $bindir \;
 cd $WORKSPACE/srcdir/ORCA_Mode_modelling_gfortran/src
-atomic_patch -p1 -l $WORKSPACE/srcdir/patches/cw_modes.patch
+perl -p -i -e 's/\r\n/\n/g;' cw_modes.f
+atomic_patch -p1 $WORKSPACE/srcdir/patches/cw_modes.patch
 rm -f *.o *.mod ../bin/*
 make
 cp ../bin/orca90* $bindir/orca90.exe


A/AcousticsToolbox/build_tarballs.jl
```diff
@@ -22,6 +22,17 @@ make
...
 cd $WORKSPACE/srcdir/ORCA_Mode_modelling_gfortran/src
+cat > patch << EOF
+c: HACK: dump mode functions and acoustic pressure to a file
+               open(67,file='modes_tlc.bin',form='unformatted')
+               write(67) nzsr, nmode, nrec, nsrc
+               write(67) ((phi(j,jm), j=1,nzsr), jm=1,nmode)
+               write(67) ((tlc(jrec,jsrc), jrec=1,nrec), jsrc=1,nsrc)
+               close(67)
+EOF
+head -106 cw_modes.f > tmp1
```
giordano: This is extremely brittle, when the lines of this file will shift this won't make any sense. Having a real patch, with context, is much more robust.

mchitre: Yes, I understand that, but the patch is for a specific version of the code. I will have to maintain it for newer versions - or hopefully the original author will fix the problem (I have reported it) and the patch won't be necessary.

The alternative is for me to host a copy of the code, which I prefer not to.

Do you have other suggestions?

giordano: Yes, apply a patch, which is what I suggested above. Search for atomic_patch in this repo, there are hundreds of examples.

mchitre: Ok, will take a look and update shortly. Thanks.

mchitre: Moved the changes to a `cw_modes.patch` file and applied it using `atomic_patch`.

mchitre:
@giordano I believe the review concerns are now addressed. Can you please help review/merge when you have a chance? Thanks.

giordano:
I was too quick merging, the license file is missing, [please install it](https://docs.binarybuilder.org/stable/build_tips/#Installing-the-license-file).

mchitre:
> I was too quick merging, the license file is missing, [please install it](https://docs.binarybuilder.org/stable/build_tips/#Installing-the-license-file).

Hopefully https://github.com/JuliaPackaging/Yggdrasil/pull/12030 addresses this.

PR #12018, opened 2025-09-05
[GAP] Rebuild with latest libjulia

lgoettgens:


benlorenz:
This might also be unnecessary, only Polymake.jl is failing on nightly and probably causing Oscar.jl to fail on nightly. So we could wait until the polymake jlls are done.

benlorenz:
not needed: https://github.com/oscar-system/Oscar.jl/actions/runs/17514492783

PR #12029, opened 2025-09-06
[odrpack] update to 3.0.1; no closures

HugoMVale:
This library version has no closures, which removes the need for an executable stack (required by GCC), and can be invoked from the Julia REPL with raising a stackoverflow error.


HugoMVale:
The build error with "x86_64-apple-darwin-libgfortran5" makes no sense, because it was working perfectly fine before adding the musl builds.

HugoMVale:
As far as I am concerned, we're good to merge. Thanks in advance.

PR #12030, opened 2025-09-06
`AcousticsToolbox`: add license information

mchitre:
The last PR (https://github.com/JuliaPackaging/Yggdrasil/pull/12013) was missing installation of LICENSE file into the tarball. This PR fixes that.

### DIFF
--- a/A/AcousticsToolbox/build_tarballs.jl
+++ b/A/AcousticsToolbox/build_tarballs.jl
@@ -26,7 +26,7 @@ cd $WORKSPACE/srcdir/ORCA_Mode_modelling_gfortran/src
 perl -p -i -e 's/\r\n/\n/g;' cw_modes.f
 atomic_patch -p1 $WORKSPACE/srcdir/patches/cw_modes.patch
 rm -f *.o *.mod ../bin/*
-make
+make -j${nproc}
 install -Dvm 755 "../bin/orca90${exeext}" "${bindir}/orca90.exe"
 install_license $WORKSPACE/srcdir/at/LICENSE
 install_license $WORKSPACE/srcdir/licenses/LICENSE-orca.txt
--- a/A/AcousticsToolbox/build_tarballs.jl
+++ b/A/AcousticsToolbox/build_tarballs.jl
@@ -28,6 +28,8 @@ atomic_patch -p1 $WORKSPACE/srcdir/patches/cw_modes.patch
 rm -f *.o *.mod ../bin/*
 make
 cp ../bin/orca90* $bindir/orca90.exe
+install_license $WORKSPACE/srcdir/at/LICENSE
+install_license $WORKSPACE/srcdir/bundled/licenses/LICENSE-orca.txt
 """
 
 # These are the platforms we will build for by default, unless further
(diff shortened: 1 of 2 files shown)


A/AcousticsToolbox/build_tarballs.jl
```diff
@@ -28,6 +28,8 @@ atomic_patch -p1 $WORKSPACE/srcdir/patches/cw_modes.patch
 rm -f *.o *.mod ../bin/*
 make
 cp ../bin/orca90* $bindir/orca90.exe
+install_license $WORKSPACE/srcdir/at/LICENSE
+install_license $WORKSPACE/srcdir/bundled/licenses/LICENSE-orca.txt
```
giordano:
```suggestion
install_license $WORKSPACE/srcdir/licenses/LICENSE-orca.txt
```

A/AcousticsToolbox/build_tarballs.jl
```diff
@@ -28,6 +28,8 @@ atomic_patch -p1 $WORKSPACE/srcdir/patches/cw_modes.patch
 rm -f *.o *.mod ../bin/*
 make
 cp ../bin/orca90* $bindir/orca90.exe
```
giordano:
```suggestion
install -Dvm 755 "../bin/orca90${exeext}" "${bindir}/orca90.exe"
```

A/AcousticsToolbox/build_tarballs.jl
```diff
@@ -27,7 +27,9 @@ perl -p -i -e 's/\r\n/\n/g;' cw_modes.f
 atomic_patch -p1 $WORKSPACE/srcdir/patches/cw_modes.patch
 rm -f *.o *.mod ../bin/*
 make
```
giordano, line 30: Why not making this a little bit faster?
```suggestion
make -j${nproc}
```

giordano: I guess the answer is "because the makefile is bad and doesn't track dependencies correctly".  But a comment to explain this would be necessary, because otherwise I'll make the same suggestion every single time I read that line.

mchitre: Ok, will add a comment shortly and revert the change.

mchitre: Reverted and added comment

A/AcousticsToolbox/build_tarballs.jl
mchitre: Reverting, as this fails on `x86_64-w64-mingw32-libgfortran3`, `x86_64-w64-mingw32-libgfortran4`, `i686-w64-mingw32-libgfortran3` and `i686-w64-mingw32-libgfortran4` where the `.exe` extension is not added during compilation but the `${exeext}` is set to `.exe`.

giordano: The more I look at this project, the more I realise the makefile is 💩

giordano:
```suggestion
install -Dvm 755 ../bin/orca90* "${bindir}/orca90.exe"
```

mchitre: > The more I look at this project, the more I realise the makefile is 💩

Agree!

PR #12036, opened 2025-09-08
[CUDA_Driver] Update init.jl

kshyatt:
Try to broadcast over readlines to unbreak things


giordano:
I can't test it right now, but at a quick glance the change looks sensible: `readlines(...)` returns a vector, so broadcasting is indeed necessary.

PR #12041, opened 2025-09-09
[GLPK] Rebuild to get rid of Pkg dependency

lgoettgens:
cc @fingolfin

### DIFF
--- a/G/GLPK/build_tarballs.jl
+++ b/G/GLPK/build_tarballs.jl
@@ -1,11 +1,12 @@
-using BinaryBuilder, Pkg
+using BinaryBuilder
 
 name = "GLPK"
-version = v"5.0.1" # <-- This is a lie, we're bumping from 5.0 to 5.0.1 to create a Julia v1.6+ release with experimental platforms
+upstream_version = v"5.0"
+version = v"5.0.2"
 
 # Collection of sources required to build GLPK
 sources = [
-    ArchiveSource("http://ftpmirror.gnu.org/gnu/glpk/glpk-$(version.major).$(version.minor).tar.gz",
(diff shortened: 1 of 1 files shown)


odow:
Why is it necessary to remove `Pkg` here?

odow:
Off topic: but hey! The CI script to ping me works well.

giordano:
> Why is it necessary to remove Pkg here?

I guess with the idea of making loading of libraries faster, since Pkg isn't in the sysimage anymore loading it adds latency to downstream packages. But in practice Pkg isn't actually loaded in Julia v1.6+: https://github.com/JuliaPackaging/JLLWrappers.jl/blob/5f026fe3dff5ce7d7830540cc84bcf07167e146d/src/toplevel_generators.jl#L27-L42. So I'm not entirely sure this achieves much, apart from nominally removing Pkg from the environment.

odow:
Which libraries though? Does `using Pkg` add it to `GLPK_jll`?

lgoettgens:
> > Why is it necessary to remove Pkg here?
> 
> I guess with the idea of making loading of libraries faster, since Pkg isn't in the sysimage anymore loading it adds latency to downstream packages. But in practice Pkg isn't actually loaded in Julia v1.6+: https://github.com/JuliaPackaging/JLLWrappers.jl/blob/5f026fe3dff5ce7d7830540cc84bcf07167e146d/src/toplevel_generators.jl#L27-L42. So I'm not entirely sure this achieves much, apart from nominally removing Pkg from the environment.

Ah ok, I didn't know that.
I was just working through `] why Pkg` with essentially the motivation to eliminate Pkg from our dependency tree (ref https://github.com/JuliaLang/julia/issues/57970#issuecomment-3271252988)

giordano:
> I was just working through ] why Pkg with essentially the motivation to eliminate Pkg from our dependency tree

I figured. This would reduce the noise, but not actually remove Pkg from your running session.

PR #11970, opened 2025-08-29
Reactant_jll v0.0.237

avik-pal:

PR #12049, opened 2025-09-10
Reactant_jll v0.0.240

avik-pal:


avik-pal:
cuda 12.4 has become unhappy

```
[17:45:48] # Configuration: 4524842b552e2fd6beb4fe7e1c338e2d4fefde2d4f342cf3e2cf298d856b37f9
[17:45:48] # Execution platform: @@local_execution_config_platform//:platform
[17:45:48] 
[17:45:48] Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
[17:45:48] external/xla/xla/tuple_tree.h:305:36: error: use of undeclared identifier '__T4'
[17:45:48]   305 | (*node) = Node::Leaf(std::forward< __T4> (value)); 
[17:45:48]       |                                    ^
[17:45:48] 1 error generated.
```

giordano:
https://github.com/JuliaPackaging/Yggdrasil/pull/11970#issuecomment-3239412674

giordano:
To elaborate, this patch https://github.com/EnzymeAD/Reactant.jl/commit/ce9246a1501676c133652eeee16e33e369dd8d3a was broken by https://github.com/openxla/xla/commit/5c637c945d40ab2439f7257cdbc4f18dd61165fc, so it needs to be adapted (in an ideal world we'd understand why the build system is doing some bad preprocessing).

wsmoses:
the answer is essentially older nvcc's are bad sadly, but the patch should be an easily adaptable one (and also a reason why getting the full jll ci to enzyme-jax for xla updates would help catch these earlier)

giordano:
https://github.com/EnzymeAD/Enzyme-JAX/pull/1369 is pretty much ready to use as is right now, except it'll be broken right now as this PR is. But https://github.com/EnzymeAD/ReactantBuilder needs more work to be usable for full deployment (specifically, registration is the bit which won't work at the moment).

Issue #12048, opened 2025-09-10
XGBoost_jll 2.1.4 doesn't load on x86 macOS


JackDunnNZ:
The new release 2.1.4 (after https://github.com/JuliaPackaging/Yggdrasil/pull/11996) is giving an error when trying to load libgomp on x86 macOS:

```
(@v1.11) pkg> activate --temp
  Activating new project at `/var/folders/0b/ql7ss_5s1d5082x_3zcgszn80000gn/T/jl_Tr7wMl`

(jl_Tr7wMl) pkg> add XGBoost_jll
   Resolving package versions...
  Downloaded artifact: LLVMOpenMP
  Downloaded artifact: XGBoost
    Updating `/private/var/folders/0b/ql7ss_5s1d5082x_3zcgszn80000gn/T/jl_Tr7wMl/Project.toml`
  [a5c6f535] + XGBoost_jll v2.1.4+0
    Updating `/private/var/folders/0b/ql7ss_5s1d5082x_3zcgszn80000gn/T/jl_Tr7wMl/Manifest.toml`
  [692b3bcd] + JLLWrappers v1.7.1
  [21216c6a] + Preferences v1.5.0
  [1d63c593] + LLVMOpenMP_jll v18.1.8+0
  [a5c6f535] + XGBoost_jll v2.1.4+0
  [56f22d72] + Artifacts v1.11.0
  [ade2ca70] + Dates v1.11.0
  [8f399da3] + Libdl v1.11.0
  [de0858da] + Printf v1.11.0
  [fa267f1f] + TOML v1.0.3
  [4ec0a83e] + Unicode v1.11.0
  [e66e0078] + CompilerSupportLibraries_jll v1.1.1+0
Precompiling project...
  4 dependencies successfully precompiled in 4 seconds. 5 already precompiled.

julia> using XGBoost_jll
ERROR: InitError: could not load library "/Users/jack/.julia/artifacts/b9dd5f0106845ece6565d536edd87ce429c4b3b8/lib/libxgboost.dylib"
dlopen(/Users/jack/.julia/artifacts/b9dd5f0106845ece6565d536edd87ce429c4b3b8/lib/libxgboost.dylib, 0x0001): Library not loaded: /workspace/destdir/x86_64-apple-darwin14/lib/libgomp.1.dylib
[...]

PR #12062, opened 2025-09-12
[XGBoost] Builds for Apple Systems and more CUDA platforms

BenCurran98:
This is a patch fix for #12048. For some reason, building from source on Mac OS systems doesn't seem to work with AppleClang, so instead I've set these systems to exclusively use the pre-built binaries from XGBoost itself. Any thoughts/suggestions welcome :slightly_smiling_face:

### DIFF
--- a/X/XGBoost/XGBoost/build_tarballs.jl
+++ b/X/XGBoost/XGBoost/build_tarballs.jl
@@ -8,23 +8,13 @@ sources = get_sources()
 
 # Bash recipe for building across all platforms
 script = raw"""
-cd ${WORKSPACE}/srcdir/
-
-# we can't seem to build from source for Apple, so install the pre-built binaries directly instead
-# see https://github.com/dmlc/xgboost/issues/11676 for details
-if [[ "${target}" == *apple-darwin* ]]; then
-    unzip -d xgboost-${target} xgboost-${target}.whl
-    cd xgboost-${target}/xgboost
-    cp ../../xgboost/LICENSE .
-else
-    cd xgboost
-    git submodule update --init
-    mkdir build && cd build
-    cmake .. -DCMAKE_INSTALL_PREFIX=${prefix} \
-            -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}
-    make -j${nproc}
-    cd ..
-fi
+cd ${WORKSPACE}/srcdir/xgboost
+git submodule update --init
+mkdir build && cd build
+cmake .. -DCMAKE_INSTALL_PREFIX=${prefix} \
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}
+make -j${nproc}
+cd ..
 """ * install_script
 
 # The products that we will ensure are always built
--- a/X/XGBoost/XGBoost/build_tarballs.jl
+++ b/X/XGBoost/XGBoost/build_tarballs.jl
@@ -8,12 +8,28 @@ sources = get_sources()
 
 # Bash recipe for building across all platforms
 script = raw"""
+# apple builds require a newer SDK (at least v12 to cover both x86_64 and aarch64) based on build info here:
+# https://github.com/dmlc/xgboost/blob/910c34b971b06861e66d3850714540b0697001e1/ops/pipeline/build-python-wheels-macos.sh#L14-L24
+if [[ $target == *"apple-darwin"* ]]; then
+    apple_sdk_root=$WORKSPACE/srcdir/MacOSX14.0.sdk
+    sed -i "s!/opt/$bb_target/$bb_target/sys-root!$apple_sdk_root!" $CMAKE_TARGET_TOOLCHAIN
+    sed -i "s!/opt/$bb_target/$bb_target/sys-root!$apple_sdk_root!" /opt/bin/$bb_full_target/$target-clang++
+    export MACOSX_DEPLOYMENT_TARGET=12
+fi
+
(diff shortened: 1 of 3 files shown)


X/XGBoost/XGBoost/build_tarballs.jl
```diff
@@ -1,28 +1,30 @@
...
+    cd xgboost-${target}/xgboost
+    cp ../../xgboost/LICENSE .
+else
+    cd xgboost
+    git submodule update --init
+    mkdir build && cd build
+    cmake .. -DCMAKE_INSTALL_PREFIX=${prefix} \
+            -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN}
+    make -j${nproc}
+    cd ..
```
giordano: That doesn't sound right.  I already mentioned at https://github.com/JuliaPackaging/Yggdrasil/pull/11996#discussion_r2337094032 that I suspect this needs a newer SDK.

Anyway,  #12048 is about libgomp not found, which should simply be solved by adding CompilerSupportLibraries for the platforms which need it (I don't know which they are though), so this PR as-is is not right anyway.

X/XGBoost/XGBoost_GPU/build_tarballs.jl
```diff
@@ -27,7 +27,7 @@ cmake .. -DCMAKE_INSTALL_PREFIX=${prefix} \
         -DUSE_CUDA=ON \
         -DBUILD_WITH_CUDA_CUB=ON
 make -j${nproc}
-
+cd ..
```
giordano, line 39: ?

ViralBShah:
I also think the GPU builds need to be built for all the CUDA platforms - like in https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SuiteSparse/SuiteSparse_GPU%407/build_tarballs.jl

X/XGBoost/XGBoost/build_tarballs.jl
BenCurran98: Yep sorry, you're completely right! I got a bit confused, but had a look at some other builds for Apple systems and think I got it right now building with v14 of the SDK. I also had to update the CMake version to match what XGBoost was using, and now the builds from source seem to work. Let me know if anything in my method doesn't make sense though, and thanks for your help and patience :slightly_smiling_face:

X/XGBoost/XGBoost_GPU/build_tarballs.jl
BenCurran98: This is just because the common "install" script located in `common.jl` operates on the level of `/workspace/srcdir/xgboost` rather than in the build directory. I can move it there if you think it makes more sense though

BenCurran98:
> I also think the GPU builds need to be built for all the CUDA platforms - like in https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SuiteSparse/SuiteSparse_GPU%407/build_tarballs.jl

I'll give this a go, although I had some issues getting things to build with earlier versions of CUDA like 11.4 (see the first part of [this comment](https://github.com/JuliaPackaging/Yggdrasil/pull/11996#issuecomment-3251341811)) but I'll take another crack!

BenCurran98:
> I also think the GPU builds need to be built for all the CUDA platforms - like in https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SuiteSparse/SuiteSparse_GPU%407/build_tarballs.jl

Ok I've had a go at this, but as far as I can tell, XGBoost doesn't seem to build for CUDA versions < 11.8 since it expects the libcudacxx library to be available at a certain path that doesn't exist for these earlier versions. We're also limited to CUDA v12.2 by compat constraints from CUDA_full_jll, but I've extended the builds to include all compatible versions between 11.8 and 12.2 

* Edit: Looks like GPU support is only there for linux x86_64 at the moment: https://github.com/dmlc/xgboost/blob/910c34b971b06861e66d3850714540b0697001e1/doc/install.rst

ViralBShah:
See my PR. You should use CUDA_SDK. CUDA_full stopped at 12.2. After that we should filter out any cuda versions that are problematic.

X/XGBoost/common.jl
```diff
@@ -8,7 +8,9 @@ include(joinpath(YGGDRASIL_DIR, "fancy_toys.jl"))
 # Collection of sources required to build XGBoost
 function get_sources()
     return [
-        GitSource("https://github.com/dmlc/xgboost.git","62e7923619352c4079b24303b367134486b1c84f") # v2.1.4
+        GitSource("https://github.com/dmlc/xgboost.git","62e7923619352c4079b24303b367134486b1c84f"), # v2.1.4
+        ArchiveSource("https://github.com/roblabla/MacOSX-SDKs/releases/download/macosx14.0/MacOSX14.0.sdk.tar.xz",
+                  "4a31565fd2644d1aec23da3829977f83632a20985561a2038e198681e7e7bf49")
```
ViralBShah, line 13: Maybe remove this comment here with the version number, so we do not need to worry about keeping it in sync.

ViralBShah:
@BenCurran98 This broadly looks good. Do you believe this also fixes Apple x86? I couldn't tell from the discussion above. It seems like it should.

I suggest we leave this open for a couple of days for any other reviews and then merge.

BenCurran98:
> @BenCurran98 This broadly looks good. Do you believe this also fixes Apple x86? I couldn't tell from the discussion above. It seems like it should.
> 
> I suggest we leave this open for a couple of days for any other reviews and then merge.

@ViralBShah  awesome, thanks :)
This should fix the Apple x86 issue, but I haven't got a device available to test it

